PR #29027, opened 2024-05-13
feat(datasource/cpan): return `sourceUrl`

RahulGautamSingh:
## Changes
- `homepage` -> `sourceUrl`

## Context
The url being returned as `homepage` is actually the `sourceUrl`, and needed a swap.


rarkins (changes requested):
This not a sourceUrl though. i.e. it's not a git repository

RahulGautamSingh:
> This not a sourceUrl though. i.e. it's not a git repository

Oh! I thought `metacpan.org` will be the homepage and `metacpan.org/<path-to-library>` will the the source url. Didn't know it had to be a git repo (which does makes sense now, since it will have the source code).

I will close this then.

PR #29044, opened 2024-05-13
docs: update regex to custom managers

setchy:


rarkins (dismissed):

lib/modules/manager/custom/regex/readme.md
```diff
@@ -1,4 +1,4 @@
-With the `regex` manager you can configure Renovate so it finds dependencies that are not detected by its other built-in package managers.
+With `customManagers` you can configure Renovate so it finds dependencies that are not detected by its other built-in package managers.
```
viceice, line 1: are we shure about this? this doc page is specific to the regex type of custom manager.

setchy: thoughts on adding a new category to https://docs.renovatebot.com/modules/manager/#supported-managers for example

| Group 	| Category ID|	Managers|
|-|-|-|
| Custom Managers | `custom` | `regex` |

viceice (approved):

lib/modules/manager/custom/regex/readme.md
```diff
@@ -1,4 +1,4 @@
-With the `regex` manager you can configure Renovate so it finds dependencies that are not detected by its other built-in package managers.
+With `customManagers` using `regex` you can configure Renovate so it finds dependencies that are not detected by its other built-in package managers.
```
HonkingGoose, line 1:
```suggestion
Renovate's built-in package managers may not find all your dependencies.
But you can tell Renovate how to find a missing dependency.
In short: write your own regex, and put the regex in the `customManagers` field.
Read the rest of the page to learn more.
```

How about something like this? Check if this is technically correct before applying. :smile:

setchy: 🤷‍♂️

RahulGautamSingh, line 1: Since we are already in the `regex` manager's readme its fine to call it rege manager directly, I think.

RahulGautamSingh: Since we are already in the regex manager's readme its fine to call it rege manager directly, I think.

PR #29065, opened 2024-05-14
fix(datasource/docker): use digest from header if available

viceice:
## Changes
Use digest from response header if it's a valid manifest, so we don't fallback to body computation.
Added todo's for another issue, whiuch needs more refactoring.
Added headers and body when manifest parsing fails.

## Context
- #28970
- https://github.com/renovate-reproductions/28970-renovate-docker-datasource


viceice:
compare without whitespace for less changes

rarkins (approved):

PR #29121, opened 2024-05-16
docs(config): warn about spaces in `schedule`

HonkingGoose:
## Changes

- Warn users that in the `schedule` config option:
  - `5am` is correct
  - `5 am` (with space) is wrong

## Context

`@rarkins` fixed a repository configuration error, they were using a space between `4` and `am`:

- https://github.com/slsa-framework/slsa-github-generator/issues/404#issue-1286144500
- https://github.com/slsa-framework/slsa-github-generator/pull/3638

The `schedule` docs only show examples of the _correct_ format, but it's easy to assume that `4 am` will work. Or to _accidentally_ add a space, because that feels natural.

I think we should warn louder about this possible gotcha. :smile: 

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -3739,6 +3739,12 @@ every 3 months on the first day of the month
 
 <!-- prettier-ignore -->
 !!! warning
+    You _must_ keep the number and the `am`/`pm` part _together_!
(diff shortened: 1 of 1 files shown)


viceice (approved):

viceice:
> Could we massage this instead?

I'm not sure, it can probably break some other scenarios

PR #29136, opened 2024-05-17
docs(gitlab): add note about group access token rotation

### LINKED ISSUE #21121: Document GitLab Group Access Token usage

We should add the `Group Access Token`[^1] support to the GitLab platform docs.
It's confirmed working here[^2].

**Limitations:** (needs validation)
- needs manual `RENOVATE_X_PLATFORM_VERSION` configuration
- no autodiscover

[^1]: https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html
[^2]: https://gitlab.com/renovate-bot/renovate-runner/-/issues/53

mschoettle:
## Changes

This PR adds a note about rotating the group access token to keep the same bot user.

Also updated the required scopes (`api` allows one to `read_user`, and autodiscover uses the API so `read_api` is sufficient).

## Context

See the discussion in https://github.com/renovatebot/renovate/discussions/28736#discussioncomment-9437710 

Closes #21121 

### DIFF
--- a/lib/modules/platform/gitlab/readme.md
+++ b/lib/modules/platform/gitlab/readme.md
@@ -9,8 +9,9 @@ To start, create either:
 - a [Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account
 - or a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups) for the bot account
 
-The bot account must have at least the Developer role in order to create issues and merge requests.
-The bot account must have the Maintainer role in order to perform automerge.
+The bot account must have at least the Developer role in order to [create issues and merge requests](https://docs.gitlab.com/ee/user/permissions.html#project-members-permissions).
+If you are using automerge, the bot account must have the appropriate "Allowed to merge" [permission on the protected branch](https://docs.gitlab.com/ee/user/project/protected_branches.html#require-everyone-to-submit-merge-requests-for-a-protected-branch) of your projects.
+If this is restricted to Maintainers, the bot account must have the Maintainer role.
 
 If you are using a group access token, to keep using the same GitLab-generated bot user you must [rotate/refresh the Group Access Token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) _before_ the token's expiry date.
 
--- a/lib/modules/platform/gitlab/readme.md
+++ b/lib/modules/platform/gitlab/readme.md
@@ -11,7 +11,7 @@ To start, create either:
 
 If you are using a Group access token, the token must have Developer role or higher permissions in order to create issues and merge requests.
 The token must have Maintainer permissions in order to perform Automerge.
-Ensure that you are [rotating the group access token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) before it expires in order to keep the same GitLab-generated bot user.
+To keep using the same GitLab-generated bot user you must [rotate/refresh the Group Access Token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) _before_ the token's expiry date.
 
 For real runs, give the access token these scopes:
 
--- a/lib/modules/platform/gitlab/readme.md
+++ b/lib/modules/platform/gitlab/readme.md
@@ -2,31 +2,29 @@
 
 ## Authentication
 
-First, [create a Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account.
+First, [create a Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account, or use a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups).
 
 If you are using a Group access token, the token must have Developer role or higher permissions in order to create issues and merge requests.
 The token must have Maintainer permissions in order to perform Automerge.
+Ensure that you are [rotating the group access token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) before it expires in order to keep the same GitLab-generated bot user.
 
-For real runs, give the PAT these scopes:
+For real runs, give the access token these scopes:
 
-- `read_user`
 - `api`
 - `write_repository`
 - `read_registry` (only if Renovate needs to access the [GitLab Container registry](https://docs.gitlab.com/ee/user/packages/container_registry/))
 
-For dry runs, give the PAT these scopes:
+For dry runs, give the access token these scopes:
 
-- `read_user`
 - `read_api`
 - `read_repository`
-- `write_repository` (when using autodiscover)
 - `read_registry` (only if Renovate needs to access the [GitLab Container registry](https://docs.gitlab.com/ee/user/packages/container_registry/))
 
-Let Renovate use your PAT by doing _one_ of the following:
+Let Renovate use your access token by doing _one_ of the following:
 
-- Set your PAT as a `token` in your `config.js` file
-- Set your PAT as an environment variable `RENOVATE_TOKEN`
-- Set your PAT when you run Renovate in the CLI with `--token=`
+- Set your access token as a `token` in your `config.js` file
+- Set your access token as an environment variable `RENOVATE_TOKEN`
+- Set your access token when you run Renovate in the CLI with `--token=`
 
 Remember to set `platform=gitlab` somewhere in your Renovate config file.
 
--- a/lib/modules/platform/gitlab/readme.md
+++ b/lib/modules/platform/gitlab/readme.md
@@ -9,9 +9,10 @@ To start, create either:
 - a [Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account
 - or a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups) for the bot account
 
-If you are using a Group access token, the token must have Developer role or higher permissions in order to create issues and merge requests.
-The token must have Maintainer permissions in order to perform Automerge.
-To keep using the same GitLab-generated bot user you must [rotate/refresh the Group Access Token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) _before_ the token's expiry date.
+The bot account must have at least the Developer role in order to create issues and merge requests.
+The bot account must have the Maintainer role in order to perform automerge.
+
+If you are using a group access token, to keep using the same GitLab-generated bot user you must [rotate/refresh the Group Access Token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) _before_ the token's expiry date.
 
 We refer to personal access tokens and group access tokens as _access tokens_ in the instructions that follow.
 
--- a/lib/modules/platform/gitlab/readme.md
+++ b/lib/modules/platform/gitlab/readme.md
@@ -13,6 +13,8 @@ If you are using a Group access token, the token must have Developer role or hig
 The token must have Maintainer permissions in order to perform Automerge.
 To keep using the same GitLab-generated bot user you must [rotate/refresh the Group Access Token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) _before_ the token's expiry date.
 
+We refer to personal access tokens and group access tokens as _access tokens_ in the instructions that follow.
+
 For real runs, give the access token these scopes:
 
 - `api`
@@ -36,8 +38,8 @@ Remember to set `platform=gitlab` somewhere in your Renovate config file.
 If you're using a private [GitLab container registry](https://docs.gitlab.com/ee/user/packages/container_registry/), you must:
 
 - Set the `RENOVATE_HOST_RULES` CI variable to `[{"matchHost": "${CI_REGISTRY}","username": "${GITLAB_USER_NAME}","password": "${RENOVATE_TOKEN}", "hostType": "docker"}]`.
-- Make sure the user that owns the `RENOVATE_TOKEN` PAT is a member of the corresponding GitLab projects/groups with the right permissions.
-- Make sure the `RENOVATE_TOKEN` PAT has the `read_registry` scope.
+- Make sure the user that owns the access token is a member of the corresponding GitLab projects/groups with the right permissions.
+- Make sure the access token has the `read_registry` scope.
 
 You may want to set `FORCE_COLOR: 3` or `TERM: ansi` to the job, in order to get colored output.
 [GitLab Runner runs the container’s shell in non-interactive mode, so the shell’s `TERM` environment variable is set to `dumb`.](https://docs.gitlab.com/ee/ci/yaml/script.html#job-log-output-is-not-formatted-as-expected-or-contains-unexpected-characters)
--- a/lib/modules/platform/gitlab/readme.md
+++ b/lib/modules/platform/gitlab/readme.md
@@ -2,7 +2,12 @@
 
 ## Authentication
 
-First, [create a Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account, or use a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups).
+You can authenticate Renovate to GitLab, with a Personal Access Token, or Group Access Token.
+
+To start, create either:
+
+- a [Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account
+- or a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups) for the bot account
 
 If you are using a Group access token, the token must have Developer role or higher permissions in order to create issues and merge requests.
 The token must have Maintainer permissions in order to perform Automerge.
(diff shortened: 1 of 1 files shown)


lib/modules/platform/gitlab/readme.md
```diff
@@ -2,31 +2,29 @@
...
+First, [create a Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account, or use a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups).
 
 If you are using a Group access token, the token must have Developer role or higher permissions in order to create issues and merge requests.
 The token must have Maintainer permissions in order to perform Automerge.
+Ensure that you are [rotating the group access token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) before it expires in order to keep the same GitLab-generated bot user.
 
-For real runs, give the PAT these scopes:
+For real runs, give the access token these scopes:
 
-- `read_user`
```
rarkins, line 12: I'm not sure this should be completely dropped, and I suspect that there should be alternative text added.

For a regular PAT, read_user allows Renovate to retrieve the user's account name and email so that they don't need to configure username and gitAuthor. For a group access token, maybe this isn't possible?

lib/modules/platform/gitlab/readme.md
```diff
@@ -2,31 +2,29 @@
 
 ## Authentication
 
-First, [create a Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account.
+First, [create a Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account, or use a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups).
```
HonkingGoose, line 5:
```suggestion
You can authenticate Renovate to GitLab, with a Personal Access Token, or Group Access Token.

To start, create either:

- a [Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account
- or a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups) for the bot account
```

How about this?

lib/modules/platform/gitlab/readme.md
```diff
@@ -2,31 +2,29 @@
...
 
 If you are using a Group access token, the token must have Developer role or higher permissions in order to create issues and merge requests.
 The token must have Maintainer permissions in order to perform Automerge.
+Ensure that you are [rotating the group access token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) before it expires in order to keep the same GitLab-generated bot user.
```
HonkingGoose, line 9:
```suggestion
To keep using the same GitLab-generated bot user you must [rotate/refresh the Group Access Token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) _before_ the token's expiry date.
```

Rewriting this a bit:

- Focus on the user
- Use the word _must_ to indicate requirements
- Use rotate/refresh, instead of just refresh
- Emphasize _before_ to make clear that the rotate/refresh must happen before the token's expiry date

lib/modules/platform/gitlab/readme.md
```diff
@@ -2,31 +2,29 @@
...
 
-For real runs, give the PAT these scopes:
+For real runs, give the access token these scopes:
```
HonkingGoose, line 11: Do personal access and group access tokens need the same permissions? It's not immediately obvious if the instructions/rules are the same for both PATs and GATs to me.

If the permissions/rules are different for PATs, than for GATs, we should have two sections:

- one section for PAT
- one section for GAT

But if both tokens have the same instructions/rules, we need a introductory sentence that explains that we will call both of these "access tokens" in the following text.

HonkingGoose (changes requested):

lib/modules/platform/gitlab/readme.md
mschoettle: I did a test with both PAT (`api`/`read_api`) and group access token (`api`/`read_api`; there is no `read_user` for those) and in both cases Renovate is able to retrieve the user's name and email and sets it as `gitAuthor`. I.e., it seems that the `read_api` and `api` scopes include the ability to read the current user's profile (I could not find any GitLab documentation stating this explicitly).

lib/modules/platform/gitlab/readme.md
mschoettle: @HonkingGoose Thanks for the suggestions above and the detailed explanations on why 👍 

If we are ok with removing `read_user` from the list (as it is not required, see above), then the permissions are the same. I changed it to "access token" because of that. It's a good point that an introductory sentence should explain this. I added one and replaced two other occurrences of "PAT" later in the instructions.

lib/modules/platform/gitlab/readme.md
```diff
@@ -2,39 +2,44 @@
...
+You can authenticate Renovate to GitLab, with a Personal Access Token, or Group Access Token.
+
+To start, create either:
+
+- a [Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account
+- or a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups) for the bot account
 
 If you are using a Group access token, the token must have Developer role or higher permissions in order to create issues and merge requests.
 The token must have Maintainer permissions in order to perform Automerge.
+To keep using the same GitLab-generated bot user you must [rotate/refresh the Group Access Token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) _before_ the token's expiry date.
```
HonkingGoose, line 14: Is this whole part about group access tokens? Right now we have this structure:

- One line about group access tokens needing "developer role or higher permissions"
- One line about tokens needing Maintainer permission to automerge
- One line about refreshing/rotating tokens

It's not clear to me when we're talking about Group access tokens, and when about generic instructions that apply to _both_ Group access tokens and personal access tokens.

Can we split up the text into clearer paragraphs?

HonkingGoose (changes requested):

lib/modules/platform/gitlab/readme.md
rarkins: I'd like to double check this via a `curl` command just to be sure. If so then I agree we can remove it

lib/modules/platform/gitlab/readme.md
mschoettle: You are right. It is actually mixed up right now. The role refers to the role of the bot account (so it affects both cases).

How about the following:

> The bot account must have at least the Developer role in order to create issues and merge requests.
> The bot account must have the Maintainer role in order to perform automerge.
>
> If you are using a group access token, to keep using the same GitLab-generated bot user you must [rotate/refresh the Group Access Token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) _before_ the token's expiry date.

HonkingGoose: The new text looks good to me!

lib/modules/platform/gitlab/readme.md
```diff
@@ -2,39 +2,45 @@
...
-If you are using a Group access token, the token must have Developer role or higher permissions in order to create issues and merge requests.
-The token must have Maintainer permissions in order to perform Automerge.
+To start, create either:
 
-For real runs, give the PAT these scopes:
+- a [Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account
+- or a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups) for the bot account
+
+The bot account must have at least the Developer role in order to create issues and merge requests.
+The bot account must have the Maintainer role in order to perform automerge.
```
HonkingGoose, line 13:
```suggestion
The bot account must have the Developer and Maintainer roles to work.

| Permission | Required | Needed to                        |
| ---------- | -------- | -------------------------------- |
| Developer  | Yes      | Create issues and Merge Requests |
| Maintainer | Yes      | Perform automerge                |
```

How about putting the info in a table? Optionally we can hammer home the need for the permissions by putting a Required=yes column in the table.

It's not clear to me if Renovate needs the _full permissions_ of the Developer/Maintainer role, or only _parts_ of it.

HonkingGoose (changes requested):
Nearly there!

lib/modules/platform/gitlab/readme.md
mschoettle: Hm. I am not sure about this.

An account can only have one role. The role has more permissions than are actually needed. Developer is required to create merge requests and branches, and push to them (see the table here: https://docs.gitlab.com/ee/user/permissions.html#project-members-permissions; for example: "Merge Requests: Create" and "Repository: Create new branches").

The "automerge needs Maintainer" actually depends on how the default branch is setup. It is assumed here that it is a protected branch I think. If I am not mistaken, the default permissions for a protected branch are that to push to it directly you need the Maintainer role. I assume that automerge merges via `git` (not via merge through the GitLab API)? There is a separate rule on protected branches for "allow to merge".

How about changing the second sentence to clarify that if automerge is used the role has to be in line with the protected branch rules (since it is project specific)?

gaeljw: I confirm that the Maintainer role is required if the branch protection rules say so. I have the same token without Maintainer role that successfully automerge in some projects but not in others because all my projects don't have the same branch protection rules.

> I assume that automerge merges via git (not via merge through the GitLab API)? 
I don't think so. Automerge enables a "flag" on the MR. It's Gitlab that does the automerge once MR checks are successful.

HonkingGoose: The [Renovate docs, `platformAutomerge`](https://docs.renovatebot.com/configuration-options/#platformautomerge) basically says:

- `platformAutomerge` has a `default` value of `true`
- `platformAutomerge` speeds up merging as it uses GitLab's native automerge functionality
- On GitLab Renovate will re-enable the PR for platform-native automerging after the PR is rebased
- `platformAutomerge` will configure PRs to be merged after the branch policies are met. Works on GitLab.
- Read the platform specific docs for version requirements

So I think this means that Renovate merges "however GitLab merges things".

mschoettle: Thanks to you both for the details. I replaced the second sentence with a clarification about the protect branch rules to make it clear that it depends on the "Allowed to merge" setting on the protected branch.

lib/modules/platform/gitlab/readme.md
```diff
@@ -2,39 +2,46 @@
...
-The token must have Maintainer permissions in order to perform Automerge.
+To start, create either:
 
-For real runs, give the PAT these scopes:
+- a [Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account
+- or a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups) for the bot account
+
+The bot account must have at least the Developer role in order to [create issues and merge requests](https://docs.gitlab.com/ee/user/permissions.html#project-members-permissions).
+If you are using automerge, the bot account must have the appropriate "Allowed to merge" [permission on the protected branch](https://docs.gitlab.com/ee/user/project/protected_branches.html#require-everyone-to-submit-merge-requests-for-a-protected-branch) of your projects.
+If this is restricted to Maintainers, the bot account must have the Maintainer role.
```
HonkingGoose, line 14: It's not clear to me what this bit refers to: `If this is restricted to Maintainers,`.

Do you maybe mean something like this?

> If automerge is restricted to Maintainers, then the bot account must have the Maintainer role.

HonkingGoose (approved):
I'm happy with the docs style and sentences now. :partying_face: 

I'll let a maintainer review the links to GitLab, and our new instructions for technical accuracy.

lib/modules/platform/gitlab/readme.md
```diff
@@ -2,39 +2,46 @@
...
 
-For real runs, give the PAT these scopes:
+- a [Personal Access Token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) for the bot account
+- or a [Group Access Token](https://docs.gitlab.com/ee/user/group/settings/group_access_tokens.html#bot-users-for-groups) for the bot account
+
+The bot account must have at least the Developer role in order to [create issues and merge requests](https://docs.gitlab.com/ee/user/permissions.html#project-members-permissions).
+If you are using automerge, the bot account must have the appropriate ["Allowed to merge" permission on the protected branch](https://docs.gitlab.com/ee/user/project/protected_branches.html#require-everyone-to-submit-merge-requests-for-a-protected-branch) of your projects.
+If merging is restricted to Maintainers, the bot account must have the Maintainer role.
+
+If you are using a group access token, to keep using the same GitLab-generated bot user you must [rotate/refresh the Group Access Token](https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token) _before_ the token's expiry date.
```
rarkins, line 16: Should we document about ignorePrAuthor if they _don't_ do this?

HonkingGoose: > Should we document about ignorePrAuthor if they _don't_ do this?

Is this what you're referring to?

1. User creates and uses token for `renovate-bot-1`
2. Token expires
3. User creates _new_ token for the same `renovate-bot-1` account
4. User now needs to set `ignorePrAuthor=true` to keep things working?

I'm not sure _why_ you need to set `ignorePrAuthor=true` when you only change the token? It's probably related to renaming-attacks, where people use a trusted name on a bad bot. Probably the token has the ID/authorization for the user's bot?

## `ignorePrAuthor` docs a bit confusing

The `ignorePrAuthor` docs are confusing me:

- One line says: "usecase GitLab users who _rotate tokens_"
- Two lines after says: "Only want to enable if you are _changing_ the bot account by _renaming_ the bot"

## Full quote of `ignorePrAuthor` docs:

> ## ignorePrAuthor
> 
> Set to `true` to fetch the entire list of PRs instead of only those authored by the Renovate user.
> 
> | Name | value |
> | --- | --- |
> | type | boolean |
> | default | false |
> | cli |	--ignore-pr-author| 
> | env |	RENOVATE_IGNORE_PR_AUTHOR
>
> This is usually needed if someone needs to migrate bot accounts, including from the Mend Renovate App to self-hosted.
> **An additional use case is for GitLab users of project or group access tokens who need to rotate them.**
> 
> If `ignorePrAuthor` is configured to true, it means Renovate will fetch the entire list of repository PRs instead of optimizing to fetch only those PRs which it created itself.
> **You should only want to enable this if you are changing the bot account** (e.g. from `@old-bot` to `@new-bot`) and want `@new-bot` to find and update any existing PRs created by `@old-bot`.
> 
> Setting this field to `true` in GitLab will also mean that all Issues will be fetched instead of only those by the bot itself.

rarkins: It could be renamed ignorePlatformAuthor

mschoettle: Based on my understanding (from the discussion that lead to this PR) the problem when changing the access token (I.e., a new bot user) is that new dashboard issues and PRs are created since the user does not match anymore. `ignorePrAuthor` solves this. 

It could be helpful to mention this and refer to the relevant docs.

HonkingGoose: If the picture in the user's brain should be closer to "ignore the platform author" than "ignore PR author" a rename seems to `ignorePlatformAuthor` seems good.

> Should we document about ignorePrAuthor if they don't do this?

I'm still confused by this whole thing, so I can only respond in general terms... :upside_down_face: The docs should make it easier for a user to:

- get the right mental picture, or concept of the config option and it's uses
- do the _right_ thing, the _first_ time
- be done reading quickly, and get on with work

If we need to explain better/more to help users, we should do that!

lib/modules/platform/gitlab/readme.md
```diff
@@ -2,39 +2,46 @@
 
 ## Authentication
```
HonkingGoose, line 3: I found this issue:

- #21121

Should this PR automatically close the issue above?

rarkins (approved):

lib/modules/platform/gitlab/readme.md
```diff
@@ -2,39 +2,46 @@
...
 - `write_repository`
 - `read_registry` (only if Renovate needs to access the [GitLab Container registry](https://docs.gitlab.com/ee/user/packages/container_registry/))
 
-For dry runs, give the PAT these scopes:
+For dry runs, give the access token these scopes:
 
-- `read_user`
 - `read_api`
 - `read_repository`
-- `write_repository` (when using autodiscover)
```
viceice, line 22: why this is removed? autodiscover will check for write access and filter repo without this permission!

https://github.com/renovatebot/renovate/blob/f21efd3c31c44367181d69f38a6e29674a653afd/lib/modules/platform/gitlab/index.ts#L168

viceice (changes requested):

lib/modules/platform/gitlab/readme.md
mschoettle: I tested autodiscover with `read_api` and `read_repository` and it was able to discover repositories.

`min_access_level = 30` refers to the Developer role (https://docs.gitlab.com/ee/api/members.html#roles) which is the minimum requirement for the bot account.

viceice: so the user or group has write access, but the token is limited to read-only? strange behavior 🤔

mschoettle: The way I understand it is that the role gives you [access/permissions to certain features](https://docs.gitlab.com/ee/user/permissions.html#project-members-permissions). If the token has the `read_api` scope then anything that the role has access to can be read.

Issue #27879, opened 2024-03-12
Convert experimental env into globalOnly `experimentalFlags`


rarkins:
### Describe the proposed change(s).

Remove all `RENOVATE_X_` logic in code, convert instead to a new admin option `experimentalFlags`. Convert docs accordingly (all should be documented and it enforced with type checking).

RahulGautamSingh:
There are some variables in this list which don't have the prefix `RENOVATE_X`, should they be included in this refactor too?

RahulGautamSingh:
Do you mean we store all these variables and values in `experimentalFlags` object? 
ie.,
_old_
```env
# .env
RENOVATE_X_HARD_EXIT=false
```

_new_
```js
// config.js
"exprimentalFlags": {
  "HARD_EXIT": false
}
```

rarkins:
I was going to keep them as an array. If they need a value then it could be like abc=10. One like this could be turned around such as "softExit" or "exitCodeZero"

RahulGautamSingh:
There are some options such as `RENOVATE_X_IGNORE_RE2` which are used before `GlobalConfig` is initialized, what should be do about them?

rarkins:
How many of those are there? I guess we need to keep them

RahulGautamSingh:
Two with the prefix `RENOVATE_X`:

 `RENOVATE_X_HARD_EXIT`          https://github.com/renovatebot/renovate/blob/bf91e94b4b95a2a9274319aeaae8e03f05605258/lib/renovate.ts#L22        
 
 `RENOVATE_X_IGNORE_RE2`         https://github.com/renovatebot/renovate/blob/bf91e94b4b95a2a9274319aeaae8e03f05605258/lib/util/regex.ts#L18

rarkins:
Ok keep them as is for now

RahulGautamSingh:
I've put together [this ](https://github.com/renovatebot/renovate/pull/28187/)PR as a starting point. I've tried to cover some of the major aspects:

1. Sorting out the validation logic
2. Adding documentation for things like experimentalFlags and a few other flags. The idea is to set up a template for handling similar cases down the line.
3. I've also included a utility function. It helps with pulling values out of the experimentalFlags array. 
For example, if we have the following config  this function makes sure we can easily get the value when needed.
``` experimentalFlags = ['dockerHubTags', 'dockerMaxPages=10'] ```

[...]

PR #29137, opened 2024-05-17
feat(self-hosted): convert experimental vars to config options

RahulGautamSingh:
## Changes
- Converted the following experimental env vars to self-hosted config options

| Env var | Config option |
|-|-|
| `RENOVATE_X_DOCKER_MAX_PAGES`  | `dockerMaxPages`  |

## Context
Part of: https://github.com/renovatebot/renovate/issues/27879#issuecomment-2116866024


docs/usage/self-hosted-configuration.md
```diff
@@ -501,6 +501,11 @@ For example, `{"dockerCliOptions": "--memory=4g"}` will add a CLI flag to the `d
 
 Read the [Docker Docs, configure runtime resource constraints](https://docs.docker.com/config/containers/resource_constraints/) to learn more.
 
+## dockerMaxPages
```
HonkingGoose, line 504: There's a typo in your PR title.

docs/usage/self-hosted-configuration.md
```diff
@@ -501,6 +501,11 @@ For example, `{"dockerCliOptions": "--memory=4g"}` will add a CLI flag to the `d
...
+
+If set to an integer, Renovate will use this as max page number for docker tags lookup on docker registries, instead of the default 20 pages.
```
HonkingGoose, line 506:
```suggestion
By default, Renovate will fetch a maximum of 20 pages when looking up Docker tags on Docker registries.

If set to an integer, Renovate will use this value as the maximum page number.
```

docs/usage/self-hosted-configuration.md
```diff
@@ -501,6 +501,11 @@ For example, `{"dockerCliOptions": "--memory=4g"}` will add a CLI flag to the `d
...
+This is useful for registries which ignores the `n` parameter in the query string and only return 50 tags per page.
```
HonkingGoose, line 507:
```suggestion
Setting a different limit is useful for registries that ignore the `n` parameter in Renovate's query string and thus only return 50 tags per page.
```

docs/usage/self-hosted-configuration.md
```diff
@@ -665,6 +670,27 @@ By default, Renovate logs and displays a warning when the `GITHUB_COM_TOKEN` is
 By setting `githubTokenWarn` to `false`, Renovate suppresses these warnings on Pull Requests, etc.
 Disabling the warning is helpful for self-hosted environments that can't access the `github.com` domain, because the warning is useless in these environments.
 
+## gitlabAutoMergeableCheckAttempts
+
+If set to an positive integer, Renovate will use this as the number of attempts to check if a merge request on GitLab is mergeable before trying to automerge.
```
HonkingGoose, line 675:
```suggestion
If set to a positive integer, Renovate tries this many times to check if a merge request on GitLab is mergeable, before trying to automerge.
```

docs/usage/self-hosted-configuration.md
```diff
@@ -665,6 +670,27 @@ By default, Renovate logs and displays a warning when the `GITHUB_COM_TOKEN` is
...
+The formula for the delay between attempts is `gitlabAutoMergeableCheckAttempts * attempt * attempt` milliseconds.
```
HonkingGoose, line 676:
```suggestion
Renovate calculates the delay between attempts with this formula: `gitlabAutoMergeableCheckAttempts * attempt * attempt` milliseconds.
```

docs/usage/self-hosted-configuration.md
```diff
@@ -665,6 +670,27 @@ By default, Renovate logs and displays a warning when the `GITHUB_COM_TOKEN` is
...
+
+Adjust default time (in milliseconds) given to GitLab to create pipelines for a commit pushed by Renovate.
+
+Can be useful for slow-running, self-hosted GitLab instances that don't react fast enough for the default delay to help.
+
+Default value: `1000` (milliseconds).
+
+## gitlabMergeRequestDelay
+
+If set, Renovate will use this as a delay to proceed with an automerge.
```
HonkingGoose, line 690:
```suggestion
If set, Renovate will wait this extra time before an automerge.
```

Maybe something like this is easier to read.

docs/usage/self-hosted-configuration.md
```diff
@@ -852,6 +878,13 @@ It also may mean that ignored directories like `node_modules` can be preserved a
 
 ## platform
 
+## platformVersion
+
+Specify this string for Renovate to skip API checks and provide GitLab/Bitbucket server version directly.
```
HonkingGoose, line 883:
```suggestion
Set a string here to let Renovate skip some API checks.
Renovate will use this value as the GitLab or Bitbucket server version.
```

Maybe we should explain which API check(s) Renovate will skip?

I think the key point is that renovate skips a "fetch version" API call?

lib/config/options/index.ts
```diff
@@ -3067,6 +3067,45 @@ const options: RenovateOptions[] = [
     default: 90,
     globalOnly: true,
   },
+  {
+    name: 'dockerMaxPages',
+    description:
```
HonkingGoose, line 3072: Please update the descriptions so they match my suggestions above.

docs/usage/self-hosted-configuration.md
RahulGautamSingh: I think delay is more commonly used.

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -109,9 +110,9 @@ export async function initPlatform({
   };
   try {
     let bitbucketServerVersion: string;
-    // istanbul ignore if: experimental feature
-    if (process.env.RENOVATE_X_PLATFORM_VERSION) {
-      bitbucketServerVersion = process.env.RENOVATE_X_PLATFORM_VERSION;
+    const platformVersion = GlobalConfig.get('platformVersion');
+    if (platformVersion) {
+      bitbucketServerVersion = platformVersion;
```
rarkins, line 115: pass this as an initPlatform() parameter instead of reading from GlobalConfig.get()

lib/modules/platform/gitlab/index.ts
```diff
@@ -130,9 +130,10 @@ export async function initPlatform({
         user.commit_email ?? user.email
       }>`;
     }
+    const platformVersion = GlobalConfig.get('platformVersion');
     // istanbul ignore if: experimental feature
-    if (process.env.RENOVATE_X_PLATFORM_VERSION) {
-      gitlabVersion = process.env.RENOVATE_X_PLATFORM_VERSION;
+    if (platformVersion) {
+      gitlabVersion = platformVersion;
```
rarkins, line 136: pass this as an initPlatform() parameter instead of reading from GlobalConfig.get()

lib/modules/platform/gitlab/index.ts
```diff
@@ -666,15 +667,15 @@ async function tryPrAutomerge(
       ];
       const desiredStatus = 'can_be_merged';
       // The default value of 5 attempts results in max. 13.75 seconds timeout if no pipeline created.
-      const retryTimes = parseInteger(
-        process.env.RENOVATE_X_GITLAB_AUTO_MERGEABLE_CHECK_ATTEMPS,
+      const retryTimes = GlobalConfig.get(
+        'gitlabAutoMergeableCheckAttempts',
```
rarkins, line 671: pass this as an initPlatform() parameter and save to local config instead of reading from GlobalConfig.get()

lib/modules/platform/gitlab/index.ts
```diff
@@ -666,15 +667,15 @@ async function tryPrAutomerge(
...
       );
 
-      const mergeDelay = parseInteger(
-        process.env.RENOVATE_X_GITLAB_MERGE_REQUEST_DELAY,
-        250,
-      );
+      const mergeDelay = GlobalConfig.get('gitlabMergeRequestDelay', 250);
+
+      // eslint-disable-next-line
+      console.log(mergeDelay, retryTimes);
```
rarkins, line 678: pass this as an initPlatform() parameter and save to local config instead of reading from GlobalConfig.get()

lib/modules/platform/gitlab/index.ts
```diff
@@ -1042,9 +1043,7 @@ export async function setBranchStatus({
 
   try {
     // give gitlab some time to create pipelines for the sha
-    await setTimeout(
-      parseInteger(process.env.RENOVATE_X_GITLAB_BRANCH_STATUS_DELAY, 1000),
-    );
+    await setTimeout(GlobalConfig.get('gitlabBranchStatusDelay', 1000));
```
rarkins, line 1046: pass this as an initPlatform() parameter and save to local config instead of reading from GlobalConfig.get()

rarkins (changes requested):

lib/config/options/index.ts
```diff
@@ -3067,6 +3067,46 @@ const options: RenovateOptions[] = [
...
+      'By default, Renovate fetches up to 20 pages of Docker tags from registries, but this maximum can be adjusted by setting this option.',
+    type: 'integer',
+    default: 20,
+    globalOnly: true,
+  },
+  {
+    name: 'gitlabAutoMergeableCheckAttempts',
+    description:
+      'If set to a positive integer, Renovate tries this many times to check if a merge request on GitLab is mergeable, before trying to automerge.',
```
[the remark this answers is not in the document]
RahulGautamSingh: Umm, no -ve integers won't be caught during validation. However I don't think anyone will use a negative integer as delay, cause it could cause time to move faster :)

lib/config/options/index.ts
```diff
@@ -3067,6 +3067,46 @@ const options: RenovateOptions[] = [
...
+    description:
+      'If set, Renovate will use this as a delay to proceed with an automerge.',
+    type: 'integer',
+    default: 250,
+    globalOnly: true,
+  },
+  {
+    name: 'platformVersion',
+    description:
+      'Set a string here to let Renovate skip some API calls to fetch the platform version.',
```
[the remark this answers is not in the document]
RahulGautamSingh: How about this
```suggestion
      'Set a version string here to let Renovate skip some API calls to fetch the platform version.',
```

lib/config/options/index.ts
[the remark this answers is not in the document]
rarkins: Good idea to validate though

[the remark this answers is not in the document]
RahulGautamSingh: On it: https://github.com/renovatebot/renovate/pull/29178

lib/modules/platform/gitlab/index.ts
```diff
@@ -1042,9 +1043,7 @@ export async function setBranchStatus({
 
   try {
     // give gitlab some time to create pipelines for the sha
```
HonkingGoose, line 1045:
```suggestion
    // give GitLab some time to create pipelines for the SHA
```

Let's capitalize GitLab.

Question: do you mean SHA, the algorithm here? Should you say which variant of SHA, like:

- SHA128
- SHA256
- SHA512

HonkingGoose (changes requested):

lib/modules/platform/gitlab/index.ts
RahulGautamSingh: I think here we mean the branch commit instead of the algorithm.

HonkingGoose: Okay! Let's leave the SHA part as is then. 🙂

HonkingGoose (approved):

lib/modules/platform/gitlab/index.ts
```diff
@@ -666,15 +673,9 @@ async function tryPrAutomerge(
       ];
       const desiredStatus = 'can_be_merged';
       // The default value of 5 attempts results in max. 13.75 seconds timeout if no pipeline created.
-      const retryTimes = parseInteger(
-        process.env.RENOVATE_X_GITLAB_AUTO_MERGEABLE_CHECK_ATTEMPS,
-        5,
-      );
+      const retryTimes = config.gitlabAutoMergeableCheckAttempts ?? 5;
```
viceice, line 676:
```suggestion
      const retryTimes = config.gitlabAutoMergeableCheckAttempts!;
```

they will be set from config at runtime, so please also set them in tests.

lib/modules/platform/gitlab/index.ts
```diff
@@ -666,15 +673,9 @@ async function tryPrAutomerge(
...
 
-      const mergeDelay = parseInteger(
-        process.env.RENOVATE_X_GITLAB_MERGE_REQUEST_DELAY,
-        250,
-      );
+      const mergeDelay = config.gitlabMergeRequestDelay ?? 250;
```
viceice, line 678:
```suggestion
      const mergeDelay = config.gitlabMergeRequestDelay!;
```

lib/modules/platform/gitlab/index.ts
```diff
@@ -1041,10 +1042,8 @@ export async function setBranchStatus({
   }
 
   try {
-    // give gitlab some time to create pipelines for the sha
-    await setTimeout(
-      parseInteger(process.env.RENOVATE_X_GITLAB_BRANCH_STATUS_DELAY, 1000),
-    );
+    // give GitLab some time to create pipelines for the SHA
+    await setTimeout(config.gitlabBranchStatusDelay ?? 1000);
```
viceice, line 1046:
```suggestion
    await setTimeout(config.gitlabBranchStatusDelay!);
```

lib/modules/platform/gitlab/index.ts
```diff
@@ -102,6 +104,8 @@ export async function initPlatform({
   endpoint,
   token,
   gitAuthor,
+  platformVersion,
+  gitlabPlatformParams,
```
viceice, line 108: it seems you forgot to pass those from worker code?

viceice (changes requested):

RahulGautamSingh:
Verified each migration locally, by logging the values of each option at runtime.
 check 1:  default values passed to functions ☑️
 check 2:  custom values passed to functions ☑️
 check 3:  old env vars parsed and then overwrite the default values which are passed to functions ☑️

lib/config/options/index.ts
```diff
@@ -3095,6 +3095,46 @@ const options: RenovateOptions[] = [
...
+  {
+    name: 'dockerMaxPages',
+    description:
+      'By default, Renovate fetches up to 20 pages of Docker tags from registries, but this maximum can be adjusted by setting this option.',
+    type: 'integer',
+    default: 20,
+    globalOnly: true,
+  },
+  {
+    name: 'gitlabAutoMergeableCheckAttempts',
```
viceice, line 3107: @rarkins @secustor I think we should unify this under a generic platform options object, so we've a single config option with known values. not is best time to do it.

WDYT?

rarkins: I'm not sure I understand. Do you mean nesting it inside a `platformOptions` object? Does it give any advantage other than making sure all possible ones are located together within config?

viceice: those are also easier findable in docs, as they're under same parent headline and not mixed all over the config options.

RahulGautamSingh: Create another issue: https://github.com/renovatebot/renovate/issues/29716

viceice (commented):
Why is this PR blocking #29716? Shouldn't we do #29716 before this? 

If we do #29716 after this then we need to also migrate things again.

docs/usage/self-hosted-configuration.md
```diff
@@ -895,6 +923,15 @@ It also may mean that ignored directories like `node_modules` can be preserved a
...
 ## platform
 
+## platformVersion
+
+Set a string here to let Renovate skip some API calls to fetch the platform version.
+Renovate will use this value as the GitLab or Bitbucket server version.
+
+Particularly useful with GitLab's `CI_JOB_TOKEN` to authenticate Renovate or to reduce API calls for Bitbucket.
+
+Read [platform details](modules/platform/gitlab/index.md) to learn why we need the server version on GitLab.
```
HonkingGoose, line 933: Is this the correct format for the link? I usually mess this up, but often the links have `./` or `../` in front of them.

lib/config/options/index.ts
```diff
@@ -3118,6 +3118,46 @@ const options: RenovateOptions[] = [
     default: 90,
     globalOnly: true,
   },
+  {
+    name: 'dockerMaxPages',
+    description:
+      'By default, Renovate fetches up to 20 pages of Docker tags from registries, but this maximum can be adjusted by setting this option.',
```
HonkingGoose, line 3124:
```suggestion
      'By default, Renovate fetches up to 20 pages of Docker tags from registries. But you can set your own limit with this config option.',
```

lib/config/options/index.ts
```diff
@@ -3118,6 +3118,46 @@ const options: RenovateOptions[] = [
...
+    description:
+      'Adjust default time (in milliseconds) given to GitLab to create pipelines for a commit pushed by Renovate.',
+    type: 'integer',
+    default: 1000,
+    globalOnly: true,
+  },
+  {
+    name: 'gitlabMergeRequestDelay',
+    description:
+      'If set, Renovate will use this as a delay to proceed with an automerge.',
```
HonkingGoose, line 3148:
```suggestion
      'If set, Renovate will use this as a delay before automerging.',
```

`delay to proceed` is hard to understand the first time you read it. How about this instead?

lib/config/options/index.ts
```diff
@@ -3118,6 +3118,46 @@ const options: RenovateOptions[] = [
...
+    type: 'integer',
+    default: 250,
+    globalOnly: true,
+  },
+  {
+    name: 'platformVersion',
+    description:
+      'Set a version string here to let Renovate skip some API calls to fetch the platform version.',
```
HonkingGoose, line 3156: Do you mean:

- some API calls related to fetching?
- all API calls related to fetching?

HonkingGoose (changes requested):

lib/config/options/index.ts
RahulGautamSingh: one api call, that is required to fetch version

HonkingGoose:
```suggestion
      'Set a version string here to let Renovate skip the API call to fetch the platform version.',
```

RahulGautamSingh:
Just realized, this does is not blocked by `platfromOptions`.

docs/usage/self-hosted-configuration.md
RahulGautamSingh: It works fine in the [live docs](https://docs.renovatebot.com/self-hosted-experimental/#renovate_x_platform_version)

RahulGautamSingh:
TODO: Test again

rarkins (changes requested):
Needs deconflicting

docs/usage/self-hosted-configuration.md
```diff
@@ -895,6 +902,15 @@ It also may mean that ignored directories like `node_modules` can be preserved a
 
 ## platform
 
+## platformVersion
```
viceice, line 905: Should be inside `platformOptions`. It's only used inside `initPlatform`

viceice (approved):

lib/workers/global/config/parse/env.spec.ts
```diff
@@ -281,6 +281,7 @@ describe('workers/global/config/parse/env', () => {
         mergeConfidenceDatasources: ['docker'],
         autodiscoverRepoSort: 'alpha',
         autodiscoverRepoOrder: 'desc',
+        dockerMaxPages: 10,
```
HonkingGoose, line 284: The docs say the _default_ maximum of Docker pages to fetch is `20`. But this test has `dockerMaxPages` set to `10`. Is that correct?

RahulGautamSingh: Yes for this test its fine.

viceice (dismissed):

rarkins (dismissed):

rarkins:
Needs deconflicting

viceice (approved):

PR #29138, opened 2024-05-17
fix(homebrew): handle new github archive url format

nikaro:
## Changes

This change fix the GitHub archive tag URL path parsing for Hombrew manager.

## Context

Old archive urls were in form: `https://github.com/<owner>/<repo>/archive/<tag>.tar.gz`
New ones are in form: `https://github.com/<owner>/<repo>/archive/refs/tags/<tag>.tar.gz`

So with the current parsing `currentValue` ends up being filled with the `refs` string instead of the tag value.

Also, `brew style` command complains about the old url format, cf.: https://github.com/nikaro/homebrew-tap/actions/runs/9131845923/job/25111949942#step:7:15


nikaro:
@rarkins i modified some forgotten urls in tests and force pushed.

rarkins (dismissed):

nikaro:
Ah my bad, i didn't saw about these snapshots things.

nikaro:
This time it should be good, i ran the `manager/homebrew` tests locally:

```
➤ pnpm jest manager/homebrew

> renovate@0.0.0-semantic-release jest /Users/nicolas/Developer/github.com/renovatebot/renovate
> GIT_ALLOW_PROTOCOL=file LOG_LEVEL=fatal node --experimental-vm-modules node_modules/jest/bin/jest.js --logHeapUsage "manager/homebrew"

Host stats:
    Cpus:      8
    Memory:    8.00 GB
    HeapLimit: 2.05 GB
 PASS  lib/modules/manager/homebrew/util.spec.ts (61 MB heap size)
 PASS  lib/modules/manager/homebrew/extract.spec.ts (88 MB heap size)
 PASS  lib/modules/manager/homebrew/update.spec.ts (84 MB heap size)

=============================== Coverage summary ===============================
Statements   : 2.45% ( 1021/41618 )
Branches     : 0.94% ( 125/13171 )
Functions    : 0.9% ( 47/5203 )
Lines        : 2.44% ( 1005/41032 )
================================================================================

Test Suites: 3 passed, 3 total
Tests:       25 passed, 25 total
Snapshots:   14 passed, 14 total
Time:        5.981 s
Ran all test suites matching /manager\/homebrew/i.
```

rarkins (dismissed):

nikaro:
I may need some help here, or someone taking over the PR. As i don't really know what to do to fix the coverage decrease. Is it just a matter of adding a case in the `extract.spec.ts` file? I'm not really familiar with TypeScript and even less with the tooling around (Jest or whatever).

Or maybe to keep things simple i can remove the handling of old URLs format, what do you think?

rarkins:
I'd prefer to keep it, unless we know that no GitHub instance (including self hosted) that anyone uses ever generates the old style any more.

Most likely it's 1-2 additional test cases. Maybe one which was changed can be reverted to the old format.

nikaro:
I added a test case, "`handles old "archive" github url format`", i hope that's enough.

rarkins:
There is update coverage missed: https://app.codecov.io/gh/renovatebot/renovate/pull/29138/blob/lib/modules/manager/homebrew/update.ts#L195

image

If this is not new code then it implies that it used to be reachable via a test, so again my advice would be to locate that test and duplicate it instead of replace it fully

nikaro:
I've simplified the way how i check for the new URL format, i guess the "issue" had to do with the null return for unhandled "archive" URLs. This should pass now. The tests and code coverage are good locally.

rarkins (approved):

PR #29145, opened 2024-05-18
feat(manager/pip-compile): extract Python version from lock files

amezin:
## Changes

Extract Python version from the header the same way as pip-compile command line.

## Context

Otherwise, you have to specify the constraint manually in renovate.json (and make sure it doesn't become out of date).

This change prevents unnecessary modification of the header line without additional configuration.

https://github.com/amezin/renovate-pip-compile-test


lib/modules/manager/pip-compile/common.ts
```diff
@@ -22,6 +23,11 @@ export function getPythonVersionConstraint(
     return python;
   }
 
+  if (extractedPythonVersion) {
+    logger.debug('Using python constraint extracted from the lock file');
+    return `==${extractedPythonVersion}`;
+  }
```
rarkins, line 29: pip-compile seems to embed a Python "stream" (e.g. 3.11) and not a version (e.g. 3.11.3). Shouldn't it be `==3.11.*` in such cases?

amezin: I'm not sure how exactly Renovate resolves these constraints, but:

1) https://docs.renovatebot.com/modules/manager/pip-compile/#configuration-of-python-version shows `==3.7` as an example

2) In my repository I currently have `"constraints": { "python": "==3.12" }`. Then in the log:

```
DEBUG: Resolved stable matching version (branch="renovate/pytest-8.x")
{
  "toolName": "python"
  "constraint": "==3.12"
  "resolvedVersion": "3.12.3"
}
```

Also, one of the added tests verifies that for `Python 3.11` in the header, the latest version from 3.11 "stream" gets installed (`getPkgReleases` mocked to return `3.11.0` and `3.11.1`).

amezin: No, I don't. Can create one if required - although tests in `artifacts.spec.ts` are almost "end to end".

rarkins: Thanks, I can confirm it worked for me in a fork of your reproduction

rarkins (approved):

amezin:
@rarkins Excuse me for the noise. This PR has been approved, but it's not in the merge queue. Does it mean "delayed till v38", or something else?

not7cd:
I have already added a comment to the squashed commit, so I'm leaving a paper trail here: https://github.com/renovatebot/renovate/commit/77524af19f8f21eca30329ebb6b57b3e34f84905#r143091151

PR #29154, opened 2024-05-19
feat(self-hosted): convert experimental env vars to config options

RahulGautamSingh:
## Changes
- Converted the following experimental env vars to self-hosted config options

| Env var | Config option |
|-|-|
| `RENOVATE_X_DELETE_CONFIG_FILE`  | `deleteConfigFile`  |
| `RENOVATE_X_S3_ENDPOINT`|  `s3Endpoint` |
| `RENOVATE_X_S3_PATH_STYLE` | `s3PathStyle` |

There are some boolean types here. They could not be converted to `experimentalFlags` because `GlobalConfig` hasn't been initialized yet, when they are used, hence converting them to self-hosted config options instead.

## Context
Last of: https://github.com/renovatebot/renovate/issues/27879#issuecomment-2116866024

- [x] Both unit tests + ran on a real repository (locally)


docs/usage/self-hosted-configuration.md
```diff
@@ -384,6 +384,13 @@ The above configuration approach will mean the values are redacted in logs like
          "customEnvVariables": {"SECRET_TOKEN": "{{ secrets.SECRET_TOKEN }}"},
 ```
 
+## deleteConfigFile
+
+If set to `true` Renovate tries to delete the self-hosted config file after reading it.
+You can set the config file Renovate should read with the `RENOVATE_CONFIG_FILE` environment variable.
+
+The process that runs Renovate must have the correct permissions to delete the config file.
```
HonkingGoose, line 392:
```suggestion
If set to `true` Renovate tries to delete the self-hosted config file after reading it.

The process that runs Renovate must have the correct permissions to delete the config file.

<!-- prettier-ignore -->
!!! tip
    You can tell Renovate where to find your config file with the `RENOVATE_CONFIG_FILE` environment variable.
```

I think we should:

- first explain how the `deleteConfigFile` works
- then give the tip about the `RENOVATE_CONFIG_FILE` environment variable

So I've moved the paragraphs around, and added a new tip admonition.

docs/usage/self-hosted-configuration.md
```diff
@@ -553,6 +560,10 @@ You can choose from the following behaviors for the `dryRun` config option:
 
 Information provided mainly in debug log level.
 
+## eagerGlobalExtends
+
+Resolve and merge `globalExtends` presets before other global config, instead of after.
```
HonkingGoose, line 565:
```suggestion
By default, Renovate resolves and merges `globalExtends` presets _after_ other global config.

But if you set `eagerGlobalExtends` to ..., then Renovate will resolve and merge `globalExtends` presets _before_ the global config.
```

Before applying my suggestion, please:

- [ ] Double check if the sentences have the correct logic!
- [ ] Replace the `...` bit with the correct instruction

docs/usage/self-hosted-configuration.md
```diff
@@ -688,6 +699,12 @@ Value of `0` means no caching.
 !!! warning
     When you set `httpCacheTtlDays` to `0`, Renovate will remove the cached HTTP data.
 
+## ignoreNodeWarn
+
+Using this option, you can suppress the default warning when a deprecated version of Node.js is used to run Renovate.
+
+By default, Renovate logs a WARN debug message if this option is set to true, Renovate will log an INFO debug message instead.
```
HonkingGoose, line 706:
```suggestion
By default, Renovate logs a `WARN` debug message if this option is set to true, Renovate will log an `INFO` debug message instead.
```

docs/usage/self-hosted-configuration.md
```diff
@@ -1093,6 +1110,18 @@ The combinations of `requireConfig` and `onboarding` are:
 | `requireConfig=optional` | An onboarding PR will be created if no config file exists. If the onboarding PR is closed and there's no config file, the repository will be processed. | Repository is processed regardless of config file presence.   |
 | `requireConfig=ignored`  | No onboarding PR will be created and repo will be processed while ignoring any config file present.                                                     | Repository is processed, any config file is ignored.          |
 
+## s3Endpoint
+
+If set, Renovate will use this string as the `endpoint` when instantiating the AWS S3 client.
```
HonkingGoose, line 1115: Is there a simpler word than `instantiate` that we can use here?

Maybe `when starting the AWS S3 client`?

docs/usage/self-hosted-configuration.md
```diff
@@ -1093,6 +1110,18 @@ The combinations of `requireConfig` and `onboarding` are:
...
+
+## s3PathStyle
+
+If set, Renovate will enable `forcePathStyle` when instantiating the AWS S3 client.
```
HonkingGoose, line 1119: Same comment about instantiate here.

docs/usage/self-hosted-configuration.md
```diff
@@ -1093,6 +1110,18 @@ The combinations of `requireConfig` and `onboarding` are:
...
+
+> Whether to force path style URLs for S3 objects (e.g., `https://s3.amazonaws.com//` instead of `https://.s3.amazonaws.com/`)
```
HonkingGoose, line 1121:
```suggestion
> Whether to force path-style URLs for S3 objects (e.g., `https://s3.amazonaws.com//` instead of `https://.s3.amazonaws.com/`)
```

It feels like we need ` -` here, but this is a direct quote, I think?

lib/config/options/index.ts
```diff
@@ -3067,6 +3067,41 @@ const options: RenovateOptions[] = [
     default: 90,
     globalOnly: true,
   },
+  {
+    name: 'deleteConfigFile',
+    description:
+      'If set to true, Renovate tries to delete the self-hosted config file atfer reading it.',
```
HonkingGoose, line 3073:
```suggestion
      'If set to `true`, Renovate tries to delete the self-hosted config file after reading it.',
```

lib/config/options/index.ts
```diff
@@ -3067,6 +3067,41 @@ const options: RenovateOptions[] = [
...
+    type: 'boolean',
+    globalOnly: true,
+  },
+  {
+    name: 'eagerGlobalExtends',
+    description:
+      'If set to true, Renovate resolves and merges `globalExtends` presets before other global config, instead of after.',
```
HonkingGoose, line 3080:
```suggestion
      'If set to `true`, Renovate resolves and merges `globalExtends` presets _before_ other global config, instead of _after_.',
```

lib/config/options/index.ts
```diff
@@ -3067,6 +3067,41 @@ const options: RenovateOptions[] = [
...
+    type: 'boolean',
+    globalOnly: true,
+  },
+  {
+    name: 'ignoreNodeWarn',
+    description:
+      'If set to true, you can suppress the default warning when a deprecated version of Node.js is used to run Renovate.',
```
HonkingGoose, line 3087:
```suggestion
      'When set to `true`, suppresses the default warning when a deprecated version of Node.js is used to run Renovate.',
```

lib/config/options/index.ts
```diff
@@ -3067,6 +3067,41 @@ const options: RenovateOptions[] = [
...
+    type: 'boolean',
+    globalOnly: true,
+  },
+  {
+    name: 's3Endpoint',
+    description:
+      'If set, Renovate will use this string as the `endpoint` when instantiating the AWS S3 client.',
```
HonkingGoose, line 3094: Same instantiate replacement thing here.

lib/config/options/index.ts
```diff
@@ -3067,6 +3067,41 @@ const options: RenovateOptions[] = [
...
+    type: 'string',
+    globalOnly: true,
+  },
+  {
+    name: 's3PathStyle',
+    description:
+      'If set, Renovate will enable `forcePathStyle` when instantiating the AWS S3 client.',
```
HonkingGoose, line 3101: Same instantiate replacement thing here.

HonkingGoose (changes requested):

lib/workers/global/config/parse/file.ts
```diff
@@ -79,21 +79,22 @@ export async function getConfig(env: NodeJS.ProcessEnv): Promise<AllConfig> {
     logger.debug('No config file found on disk - skipping');
   }
 
-  await deleteNonDefaultConfig(env); // Try deletion only if RENOVATE_CONFIG_FILE is specified
+  await deleteNonDefaultConfig(env, !!config.deleteConfigFile); // Try deletion only if RENOVATE_CONFIG_FILE is specified
```
viceice, line 82: this is a problem. We can't set `RENOVATE_DELETE_CONFIG_FILE` from env, because we don't pass it to the file `getConfig`. so the config value isn't yet available and causes a breaking change.
also cli flags are not yet available, so wer probably need to move the delete code to a later stage.

@rarkins WDYT?

viceice (changes requested):

lib/workers/global/config/parse/file.ts
rarkins: @nabeelsaabna can we move this to a later stage and still achieve the goal?

RahulGautamSingh: In that case, we could move it to the `config/parse/index` file. We will have the env vars, and flags available there.

RahulGautamSingh:
Verified each migration locally, by either logging the values of each option at runtime or confirming that they are working as expected
 check 1:  default values passed to functions ☑️
 check 2:  custom values (from cli, env, file), overwrites default values ☑️
 check 3:  old env vars parsed and then overwrite the default values ☑️

lib/config/options/index.ts
```diff
@@ -3097,6 +3097,45 @@ const options: RenovateOptions[] = [
...
+  {
+    name: 'deleteConfigFile',
+    description:
+      'If set to `true`, Renovate tries to delete the self-hosted config file after reading it.',
+    type: 'boolean',
+    default: false,
+    globalOnly: true,
+  },
+  {
+    name: 'eagerGlobalExtends',
```
rarkins, line 3109:
```suggestion
    name: 'globalExtendsEager',
```

Let's rename it so that it sorts next to globalExtends

rarkins (changes requested):

docs/usage/self-hosted-configuration.md
```diff
@@ -684,6 +694,12 @@ Use the `extends` field instead of this if, for example, you need the ability fo
     When Renovate resolves `globalExtends` it does not fully process the configuration.
     This means that Renovate does not have the authentication it needs to fetch private things.
 
+## globalExtendsEager
+
+By default, Renovate resolves and merges `globalExtends` presets _after_ other global config.
+
+But if you set `globalExtendsEager` to `true`, then Renovate will resolve and merge `globalExtends` presets _before_ the global config.
+
```
rarkins, line 702: I just realized that we should reverse this default and drop this option in v38.

Maybe better to remove this from the migration?

docs/usage/self-hosted-configuration.md
```diff
@@ -694,6 +710,12 @@ Value of `0` means no caching.
 !!! warning
     When you set `httpCacheTtlDays` to `0`, Renovate will remove the cached HTTP data.
 
+## ignoreNodeWarn
+
+Using this option, you can suppress the default warning when a deprecated version of Node.js is used to run Renovate.
+
+By default, Renovate logs a `WARN` debug message if this option is set to true, Renovate will log an `INFO` debug message instead.
+
```
rarkins, line 718: This should be removed too, because how logLevelRemap can be used instead

docs/usage/self-hosted-configuration.md
```diff
@@ -1132,6 +1154,18 @@ The combinations of `requireConfig` and `onboarding` are:
...
+If set, Renovate will use this string as the `endpoint` when creating the AWS S3 client instance.
+
+## s3PathStyle
+
+If set, Renovate will enable `forcePathStyle` when creating the AWS S3 client instance.
+
+> Whether to force path-style URLs for S3 objects (e.g., `https://s3.amazonaws.com//` instead of `https://.s3.amazonaws.com/`)
+
+Source: [AWS S3 documentation - Interface BucketEndpointInputConfig](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/bucketendpointinputconfig.html)
+
```
rarkins, line 1168: @Gabriel-Ladzaretti why is this necessary? Isn't it the same administrator who's configuring the S3?

Gabriel-Ladzaretti: This is mainly used for s3 compatible solutions that dont support or are not configured to use `virtual-hosted–style` requests. It essentially alters the way the aws client constructs and resolves urls.
[path-style-access](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#path-style-access) 
e.g, `https://s3.amazonaws.com/mybucket/myfile`
vs
[virtual-hosted-style-access](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#virtual-hosted-style-access) 
e,g, `https://mybucket.s3.amazonaws.com/myfile`

i dont think there is another way to support these s3 instances.

docs/usage/self-hosted-configuration.md
RahulGautamSingh: https://github.com/renovatebot/renovate/pull/29835

docs/usage/self-hosted-configuration.md
viceice: I use it for my Minio instance

lib/instrumentation/reporting.ts
```diff
@@ -89,7 +89,7 @@ export async function exportStats(config: RenovateConfig): Promise<void> {
         ContentType: 'application/json',
       };
 
-      const client = getS3Client();
+      const client = getS3Client(config.s3Endpoint, config.s3PathStyle);
```
viceice, line 92: Is this required because we didn't initialized the global config at that time?

RahulGautamSingh: Yes, we do not have global config class initialized.

docs/usage/self-hosted-experimental.md
```diff
@@ -32,6 +32,25 @@ Skipping the check will speed things up, but may result in versions being return
...
+## `RENOVATE_X_DELETE_CONFIG_FILE`
+
+If `true` Renovate tries to delete the self-hosted config file after reading it.
+You can set the config file Renovate should read with the `RENOVATE_CONFIG_FILE` environment variable.
+
+The process that runs Renovate must have the correct permissions to delete the config file.
+
+## `RENOVATE_X_DOCKER_HUB_DISABLE_LABEL_LOOKUP`
+
+If set to any value, Renovate will skip attempting to get release labels (e.g. gitRef, sourceUrl) from manifest annotations for `https://index.docker.io`.
```
HonkingGoose, line 44:
```suggestion
If set to any value, Renovate will not fetch any release labels from manifest annotations.

For example, when set to any value, Renovate skips getting the `gitRef` and/or `sourceUrl` release labels from the `https://index.docker.io` manifest annotations.
```

Let's split the sentence into two parts. One sentence for the basic idea. And one sentence with the example bits.

docs/usage/self-hosted-experimental.md
```diff
@@ -32,6 +32,25 @@ Skipping the check will speed things up, but may result in versions being return
...
+
+Due to the missing label information like sourceUrl, Renovate will not be able to perform certain actions dependent on these information for the images.
+
+This includes the following:
+
+- Generating changelogs
+- Applying package rules dependent on the labels
+- Including the sourceUrls in PR bodies
```
HonkingGoose, line 52:
```suggestion
If Renovate does not find the label information that it needs, for example from `sourceUrl`, then Renovate will:

- Not generate changelogs
- Not apply package rules that rely on label information
- Not include the `sourceUrls` in the PR body text
```

HonkingGoose (changes requested):

rarkins (dismissed):

docs/usage/self-hosted-configuration.md
```diff
@@ -1140,6 +1150,18 @@ The combinations of `requireConfig` and `onboarding` are:
...
+If set, Renovate will use this string as the `endpoint` when creating the AWS S3 client instance.
+
+## s3PathStyle
+
+If set, Renovate will enable `forcePathStyle` when creating the AWS S3 client instance.
+
+> Whether to force path-style URLs for S3 objects (e.g., `https://s3.amazonaws.com//` instead of `https://.s3.amazonaws.com/`)
+
+Source: [AWS S3 documentation - Interface BucketEndpointInputConfig](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/bucketendpointinputconfig.html)
```
HonkingGoose, line 1163:
```suggestion
If set, Renovate will enable `forcePathStyle` when creating the AWS S3 client instance.

Quote from [AWS S3 documentation - Interface BucketEndpointInputConfig](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/bucketendpointinputconfig.html):

> Whether to force path-style URLs for S3 objects (e.g., `https://s3.amazonaws.com//` instead of `https://.s3.amazonaws.com/`)
```

Switching around the order, with a small rewrite.

lib/util/s3.spec.ts
```diff
@@ -28,10 +26,13 @@ describe('util/s3', () => {
     expect(client1).toBe(client2);
   });
 
-  it('is uses experimental env', async () => {
-    process.env.RENOVATE_X_S3_ENDPOINT = 'https://minio.domain.test';
-    process.env.RENOVATE_X_S3_PATH_STYLE = 'true';
+  it('uses user configured s3 values', async () => {
```
HonkingGoose, line 29:
```suggestion
  it('uses user-configured s3 values', async () => {
```

Adding a `-` here makes it easier to see the difference between `uses` and `user`. 😉

lib/util/s3.ts
```diff
@@ -1,12 +1,20 @@
 // Singleton S3 instance initialized on-demand.
 import { S3Client } from '@aws-sdk/client-s3';
+import is from '@sindresorhus/is';
+import { GlobalConfig } from '../config/global';
 import { parseUrl } from './url';
 
 let s3Instance: S3Client | undefined;
-export function getS3Client(): S3Client {
+export function getS3Client(
+  // not necessary unless GlobalConfig is not initialized due to some error
```
HonkingGoose, line 9:
```suggestion
  // Only needed if GlobalConfig is not initialized due to some error
```

lib/workers/global/config/parse/index.ts
```diff
@@ -113,6 +113,9 @@ export async function parseConfigs(
     config.forkProcessing = 'enabled';
   }
 
+  // Try deletion only if RENOVATE_CONFIG_FILE is specified
```
HonkingGoose, line 116:
```suggestion
  // Only try deletion if RENOVATE_CONFIG_FILE is set
```

HonkingGoose (changes requested):
Some small things, then good to go.

HonkingGoose (approved):
Good to go for me. But the maintainers should do the final review. :wink:

RahulGautamSingh:
It's been a while since I have tested these, so I will test once and open afterwards.

rarkins (changes requested):
Needs conflict resolution

docs/usage/self-hosted-configuration.md
```diff
@@ -1140,6 +1150,18 @@ The combinations of `requireConfig` and `onboarding` are:
...
+
+If set, Renovate will use this string as the `endpoint` when creating the AWS S3 client instance.
+
+## s3PathStyle
+
+If set, Renovate will enable `forcePathStyle` when creating the AWS S3 client instance.
+
+Quote from [AWS S3 documentation - Interface BucketEndpointInputConfig](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/bucketendpointinputconfig.html):
+
+> Whether to force path-style URLs for S3 objects (e.g., `https://s3.amazonaws.com//` instead of `https://.s3.amazonaws.com/`)
```
viceice, line 1163:
```suggestion
> Whether to force path-style URLs for S3 objects (e.g., `https://s3.amazonaws.com/bucket/` instead of `https://bucket.s3.amazonaws.com/`)
```

i think this is a more clear sample

lib/workers/global/config/parse/index.ts
```diff
@@ -98,6 +98,13 @@ export async function parseConfigs(
     config.forkProcessing = 'enabled';
   }
 
+  // Only try deletion if RENOVATE_CONFIG_FILE is set
+  await fileParser.deleteNonDefaultConfig(env, !!config.deleteConfigFile);
+
+  // Remove log file entries
+  delete config.logFile;
+  delete config.logFileLevel;
+
```
viceice, line 107:
```suggestion
```

Why? Those are now env vars.

RahulGautamSingh: maybe left while deconficting

docs/usage/self-hosted-configuration.md
HonkingGoose, line 1163:
```suggestion
For example:

| `s3PathStyle` | Path                               |
| ------------- | ---------------------------------- |
| `off`         | `https://bucket.s3.amazonaws.com/` |
| `on`          | `https://s3.amazonaws.com/bucket/` |

Read the [AWS S3 docs, Interface BucketEndpointInputConfig](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/bucketendpointinputconfig.html) to learn more about path-style URLs.
```

I like @viceice's suggestion above. But quotes must match the actual quote.

How about this instead?

- Use a table to show the difference between config option on/off
- Drop the AWS docs quote
- Rewrite sentence to mention the AWS docs

My suggestion only gives you the _layout_. Please update the table so it shows the _correct_ information!

lib/config/types.ts
```diff
@@ -166,6 +166,9 @@ export interface RepoGlobalConfig {
   autodiscoverRepoSort?: RepoSortMethod;
   autodiscoverRepoOrder?: SortMethod;
   userAgent?: string;
+  deleteConfigFile?: boolean;
+  s3Endpoint?: string;
+  s3PathStyle?: boolean;
```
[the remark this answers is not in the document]
RahulGautamSingh: No, as we need to access them before `GlobalConfig` is set.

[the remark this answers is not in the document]
RahulGautamSingh: We can remove `deleteConfigFile`

lib/instrumentation/reporting.ts
RahulGautamSingh: Yes it is intialized incase everything runs fine. But incase some error occurs and renovateRepository fn does not get called the GlobalConfig.set will not be called.

Therefore, I passed the values directly from global config incase that happens, to be safe.

RahulGautamSingh: https://github.com/renovatebot/renovate/blob/daf48c7f06f85e8797d9ba6ae9ae7027b22d885f/lib/workers/global/index.ts#L217

need it here

viceice (approved):

rarkins (approved):

rarkins:
@nabeelsaabna please review

nabeelsaabna (approved):
We are only using `RENOVATE_X_DELETE_CONFIG_FILE`

PR #29164, opened 2024-05-20
fix: Correct digest resolution when the replacementName and replacementVersion options are defined

### LINKED ISSUE #20304: docker: Replacement causes existing dependencies to use wrong digest

Mend Renovate hosted app on github.com

### If you're self-hosting Renovate, tell us what version of Renovate you run.

34.125.1

### If you're self-hosting Renovate, select which platform you are using.

Bitbucket Server

### If you're self-hosting Renovate, tell us what version of the platform you run.

N/A

### Was this something which used to work for you, and then stopped?

I never saw this working

### Describe the bug

We're replacing an old image with a new one, but the digest for the new image is different.

This is fine and the replacement PRs look fine too, but Renovate seems to assume the replacements are also aliases.

This means that Renovate is now opening PRs for the old image with the digest for the new image...

To help illustrate this, here are our two images:

- legacy, digest: 44dea30fe85df1f3d554cee602fa69d553da7ab746c8c9f20defb1f1fc3e03b8
- current, digest: e742813303ad00bbe1409ea01eb909cb67818a8fbaae8c07df79036a7dea3e1d

Renovate is opening PRs which look like:

```diff
-FROM legacy@sha256:44dea30fe85df1f3d554cee602fa69d553da7ab746c8c9f20defb1f1fc3e03b8
+FROM legacy@sha256:e742813303ad00bbe1409ea01eb909cb67818a8fbaae8c07df79036a7dea3e1d
```

This is wrong and unexpected. I've read these issues, but they're not completely the same:

- https://github.com/renovatebot/renovate/issues/20128
- https://github.com/renovatebot/renovate/issues/19999

### Relevant debug logs

```
Copy/paste the relevant log(s) here, between the starting and ending backticks
```

</details>

### Have you created a minimal reproduction repository?

<https://github.com/renovate-reproductions/27728-docker-replacement-digest>

ssegato81:
## Changes

The lookupName and currentDigest values are set as undefined when a replacementName update is running because the new image may differ from the original one.
Added new test case that validates the correct getDigest function call

## Context

Discussion https://github.com/renovatebot/renovate/discussions/27728
Closes https://github.com/renovatebot/renovate/issues/20304


rarkins:
Do you have an example/reproduction repo to demonstrate this?

ssegato81:
> Do you have an example/reproduction repo to demonstrate this?

I included all the information in the Issue created a year ago. I don't know if that would be enough. 

https://github.com/renovatebot/renovate/issues/20304#issuecomment-1936464731

If you need a repository that reproduces the problem I'd need to make a different one for this PR because the original one includes company information.

ssegato81:
> Do you have an example/reproduction repo to demonstrate this?

Would this be a good example? 

https://github.com/ssegato81/minimalReproduction

lib/workers/repository/process/lookup/index.ts
```diff
@@ -537,7 +537,13 @@ export async function lookupUpdates(
           const getDigestConfig: GetDigestInputConfig = {
             ...config,
             registryUrl: update.registryUrl ?? res.registryUrl,
-            lookupName: res.lookupName,
+            // TODO #27728
+            lookupName:
+              update.updateType === 'replacement' ? undefined : res.lookupName,
```
rarkins, line 542: Should we also check whether there's a replacement name? e.g. if no replacement name then is it ok to keep using res.lookupName?

ssegato81: A replacement update is created when a replacementName, replacementNameTemplate or replacementVersion is defined. 

The replacement update is created here:
https://github.com/renovatebot/renovate/blob/25255596d63a03a312885aba1b25fdfd7b76c7a4/lib/workers/repository/process/lookup/index.ts#L472

And this function validates if it's needed:
https://github.com/renovatebot/renovate/blob/25255596d63a03a312885aba1b25fdfd7b76c7a4/lib/workers/repository/process/lookup/utils.ts#L28

The only case when it may make sense to keep the lookupName if it's only the version is updated, but despite that it seems that the value is calculated again in the getDigest function because if the lookupName doesn't exist uses the packageName (the lookupName is based on the packageName)

https://github.com/renovatebot/renovate/blob/25255596d63a03a312885aba1b25fdfd7b76c7a4/lib/modules/datasource/docker/index.ts#L811

viceice (dismissed):

lib/workers/repository/process/lookup/index.ts
```diff
@@ -538,7 +538,23 @@ export async function lookupUpdates(
             ...config,
             registryUrl: update.registryUrl ?? res.registryUrl,
             lookupName: res.lookupName,
+            currentDigest: config.currentDigest,
           };
+
+          // #20304 only pass it for replacement updates, otherwise we get wrong or invalid digest
+          if (update.updateType !== 'replacement') {
+            delete getDigestConfig.replacementName;
+          }
```
viceice, line 547: @rarkins this is the second fix. we currently always use the `replacementName` for all update types, so the `getDigest` is called with wrong package name for nomal updates and causes digest lookup warnings (especially for docker)

viceice (approved):

lib/workers/repository/process/lookup/index.ts
rarkins: Where was it being used after this though?

rarkins (approved):

PR #29168, opened 2024-05-20
feat(util/package-rules): allow glob pattens in match{Current,New}Value

amezin:
## Changes

Allow glob patterns in `packageRules.match{Current,New}Value`

## Context

- #29167

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2749,14 +2749,15 @@ Use this field to restrict rules to a particular datasource. e.g.
 
 This option is matched against the `currentValue` field of a dependency.
 
-`matchCurrentValue` supports Regular Expressions and glob patterns. For example, the following enforces that only `1.*` versions will be used:
+`matchCurrentValue` supports Regular Expressions and glob patterns. For example, the following enforces that updates from `1.*` versions will be merged automatically:
 
 ```json
 {
   "packageRules": [
     {
       "matchPackagePatterns": ["io.github.resilience4j"],
-      "matchCurrentValue": "1.*"
+      "matchCurrentValue": "1.*",
+      "automerge": true
     }
   ]
 }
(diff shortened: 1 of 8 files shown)


docs/usage/configuration-options.md
```diff
@@ -2749,8 +2749,20 @@ Use this field to restrict rules to a particular datasource. e.g.
...
-For example, the following enforces that only `1.*` versions will be used:
+`matchCurrentValue` supports Regular Expressions and glob patterns. For example, the following enforces that only `1.*` versions will be used:
+
+```json
+{
+  "packageRules": [
+    {
+      "matchPackagePatterns": ["io.github.resilience4j"],
+      "matchCurrentValue": "1.*"
+    }
```
rarkins, line 2760: Add a non match argument to this rule so it's valid, eg automerge=true

rarkins (changes requested):

lib/config/validation.spec.ts
```diff
@@ -163,10 +163,15 @@ describe('config/validation', () => {
...
+          {
+            matchPackageNames: ['bad'],
+            matchNewValue: '/^2(/',
+            enabled: true,
+          },
         ],
       };
       const { errors } = await configValidation.validateConfig('repo', config);
-      expect(errors).toHaveLength(2);
+      expect(errors).toHaveLength(1);
```
rarkins, line 174: Why did it go down?

amezin: All previous "invalid" - non-regex - values became valid

viceice (commented):
i think this should target v38 because of probably breaking behavior

rarkins:
@viceice I think this is "adding a feature" (glob and exact match) while existing functionality will remain the same, or have I missed something?

viceice:
ok

rarkins (approved):

PR #29178, opened 2024-05-21
feat(config/validation): add validation for negative numbers

RahulGautamSingh:
## Changes
- Validate negative numbers when they are set for integer types.

## Context
Ref this discussion: https://github.com/renovatebot/renovate/pull/29137#discussion_r1606402087


rarkins:
Are you sure we never use a negative number like -1 intentionally?

RahulGautamSingh:
> Are you sure we never use a negative number like -1 intentionally?

We do but only for `prPriority`. I am still to add that change here.

rarkins:
I think that this PR is maybe oversimplifying things. Not all numbers need to be positive, even if most are. A broad rule doesn't work.

Also yes, prPriority should always be part of packageRules

RahulGautamSingh:
We can add `allowNegative` field for config options of type: `integer`. That will help differentiate them better.

viceice (approved):

rarkins (changes requested):
Please test against a real repo and show us the results.

RahulGautamSingh:
Repro: https://github.com/Rahul-renovate-testing/feat-29178/issues/1

rarkins (approved):

PR #29205, opened 2024-05-22
feat(pr): add an indication to PRs that a warning was raised if there was in logs/dashboard(repo problems section)

PhilipAbed:


rarkins:
@justo-mend @PhilipAbed I think we need to go back to a Discussion about this, make sure we are designing it right and thinking of the edge cases. For example if there's a warning, do we go and update every open PR body with the warning? Are there types of warnings we _don't_ want to include, etc.

I suggest that you do not tie the encryptedWarning capability to this change. Instead, get it added to dashboard only, so that it's not delayed.

PhilipAbed:
closed as per rhys's last comment.

PR #29216, opened 2024-05-22
feat(gomod): Support go work vendor

chancez:
## Changes

Run `go work vendor` instead of `go mod vendor` if a `go.work` is present in the module. As of Go 1.22 (https://github.com/golang/go/issues/60056) Go supports vendoring with workspaces. If a `go.work` is located in the module along with the `vendor` directory, `go mod vendor` will fail, so `go work vendor` is needed instead. 

I also run `go work sync` since pretty likely users want their dependencies synced across all their modules in the repository. If this is undesirable, an option could be added to skip this step as well.

## Context

Relates to https://github.com/renovatebot/renovate/discussions/29107. I wanted an option to disable the vendoring behavior entirely, but I figured making renovate run `go work vendor` would help more people, so I opted to do this instead.

### DIFF
--- a/lib/modules/manager/gomod/artifacts.ts
+++ b/lib/modules/manager/gomod/artifacts.ts
@@ -237,6 +237,7 @@ export async function updateArtifacts({
         throw new Error('Invalid goGetDirs');
       }
     }
+
     let args = `get -d -t ${goGetDirs ?? './...'}`;
     logger.trace({ cmd, args }, 'go get command included');
     execCommands.push(`${cmd} ${args}`);
@@ -276,7 +277,6 @@ export async function updateArtifacts({
         config.postUpdateOptions?.includes('gomodTidy1.17') === true ||
         config.postUpdateOptions?.includes('gomodTidyE') === true ||
         (config.updateType === 'major' && isImportPathUpdateRequired));
-
     if (isGoModTidyRequired) {
       args = 'mod tidy' + tidyOpts;
       logger.debug('go mod tidy command included');
@@ -290,7 +290,7 @@ export async function updateArtifacts({
       const goWorkEnv = await exec(`${cmd} env GOWORK`, execOptions);
       const goWorkFile = goWorkEnv?.stdout?.trim() || '';
       const useGoWork = goWorkFile.length && ((await readLocalFile(goWorkFile)) !== null);
-      if (useGoWork) {
+      if (!useGoWork) {
         logger.debug('No go.work found');
       }
 
@@ -308,6 +308,7 @@ export async function updateArtifacts({
         execCommands.push(`${cmd} ${args}`);
       }
     }
+
     // We tidy one more time as a solution for #6795
     if (isGoModTidyRequired) {
       args = 'mod tidy' + tidyOpts;
--- a/lib/modules/manager/gomod/artifacts.ts
+++ b/lib/modules/manager/gomod/artifacts.ts
@@ -13,6 +13,7 @@ import {
   isValidLocalPath,
   readLocalFile,
   writeLocalFile,
+  findLocalSiblingOrParent,
 } from '../../../util/fs';
 import { getRepoStatus } from '../../../util/git';
 import { getGitEnvironmentVariables } from '../../../util/git/auth';
@@ -287,15 +288,15 @@ export async function updateArtifacts({
     if (useVendor) {
       // If go env GOWORK returns a non-empty path, check that it exists and if
       // it does, then use go workspace vendoring.
-      const goWorkEnv = await exec(`${cmd} env GOWORK`, execOptions);
-      const goWorkFile = goWorkEnv?.stdout?.trim() || '';
-      const useGoWork =
-        goWorkFile.length && (await readLocalFile(goWorkFile)) !== null;
-      if (!useGoWork) {
+      const goWorkFile = await findLocalSiblingOrParent(
+        goModFileName,
+        'go.work',
+      );
+      if (!goWorkFile) {
         logger.debug('No go.work found');
       }
 
-      if (useGoWork) {
+      if (goWorkFile) {
         args = 'work vendor';
         logger.debug('using go work vendor');
         execCommands.push(`${cmd} ${args}`);
--- a/lib/modules/manager/gomod/artifacts.ts
+++ b/lib/modules/manager/gomod/artifacts.ts
@@ -286,15 +286,11 @@ export async function updateArtifacts({
 
     const goWorkSumFileName = upath.join(goModDir, 'go.work.sum');
     if (useVendor) {
-      // If go env GOWORK returns a non-empty path, check that it exists and if
-      // it does, then use go workspace vendoring.
+      // If we find a go.work, then use go workspace vendoring.
       const goWorkFile = await findLocalSiblingOrParent(
         goModFileName,
         'go.work',
       );
-      if (!goWorkFile) {
-        logger.debug('No go.work found');
-      }
 
       if (goWorkFile) {
         args = 'work vendor';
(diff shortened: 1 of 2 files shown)


lib/modules/manager/gomod/artifacts.ts
```diff
@@ -235,7 +237,6 @@ export async function updateArtifacts({
         throw new Error('Invalid goGetDirs');
       }
     }
-
```
viceice, line 238: revert

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -275,23 +276,38 @@ export async function updateArtifacts({
...
-      args = 'mod vendor';
-      logger.debug('go mod tidy command included');
-      execCommands.push(`${cmd} ${args}`);
-      if (isGoModTidyRequired) {
-        args = 'mod tidy' + tidyOpts;
-        logger.debug('go mod tidy command included');
+      // If go env GOWORK returns a non-empty path, check that it exists and if
+      // it does, then use go workspace vendoring.
+      const goWorkEnv = await exec(`${cmd} env GOWORK`, execOptions);
+      const goWorkFile = goWorkEnv?.stdout?.trim() || '';
```
viceice, line 291: why not simply checking file exists? it's much faster.

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -275,23 +276,38 @@ export async function updateArtifacts({
...
+      const useGoWork = goWorkFile.length && ((await readLocalFile(goWorkFile)) !== null);
+      if (useGoWork) {
+        logger.debug('No go.work found');
+      }
```
viceice, line 295: wrong condition?

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -275,23 +276,38 @@ export async function updateArtifacts({
...
+        args = 'work sync';
+        logger.debug('using go work sync');
+        execCommands.push(`${cmd} ${args}`);
+      } else {
+        args = 'mod vendor';
+        logger.debug('using go mod vendor');
         execCommands.push(`${cmd} ${args}`);
       }
     }
-
```
viceice, line 294: avoid those changes

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -275,23 +276,38 @@ export async function updateArtifacts({
...
+      const goWorkEnv = await exec(`${cmd} env GOWORK`, execOptions);
+      const goWorkFile = goWorkEnv?.stdout?.trim() || '';
+      const useGoWork = goWorkFile.length && ((await readLocalFile(goWorkFile)) !== null);
+      if (useGoWork) {
+        logger.debug('No go.work found');
+      }
+
+      if (useGoWork) {
+        args = 'work vendor';
+        logger.debug('using go work vendor');
```
viceice, line 299: use trace log level

viceice (changes requested):

lib/modules/manager/gomod/artifacts.ts
chancez: I started there, but it doesn't seem robust to when `go.work` isn't located directly near the `go.mod` that the manager is operating on.

lib/modules/manager/gomod/artifacts.ts
chancez: The existing log messages and pretty much everything else is using debug, using trace here would be inconsistent, are you sure you want trace here? What about the rest of the logs, or existing logs?

chancez: Also; these debug logs are really useful for debugging what it's doing, idk how to get trace logs but debug logs here are useful for end-users troubleshooting.

lib/modules/manager/gomod/artifacts.ts
viceice: the `go.work` is a silbling or in parent folder? then use this

https://github.com/renovatebot/renovate/blob/8e6f89b0da7e722fe82c878c7afe35687a539d4b/lib/util/fs/index.ts#L136-L142

lib/modules/manager/gomod/artifacts.ts
viceice: you can remap log level at runtime, current logging is too verbose, we should lower almost all messages

https://docs.renovatebot.com/configuration-options/#loglevelremap

viceice (commented):
see comments

lib/modules/manager/gomod/artifacts.ts
rarkins: I'm ok to include this at debug level, particularly considering we have existing statements like `logger.debug('go mod tidy command included');`

chancez:
@viceice PTAL

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -281,13 +284,29 @@ export async function updateArtifacts({
...
 
+    const goWorkSumFileName = upath.join(goModDir, 'go.work.sum');
     if (useVendor) {
-      args = 'mod vendor';
-      logger.debug('go mod tidy command included');
-      execCommands.push(`${cmd} ${args}`);
-      if (isGoModTidyRequired) {
-        args = 'mod tidy' + tidyOpts;
-        logger.debug('go mod tidy command included');
+      // If go env GOWORK returns a non-empty path, check that it exists and if
```
viceice, line 289: please update this comment, it doesn't describe the current behavior

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -281,13 +284,29 @@ export async function updateArtifacts({
...
+      // it does, then use go workspace vendoring.
+      const goWorkFile = await findLocalSiblingOrParent(
+        goModFileName,
+        'go.work',
+      );
+      if (!goWorkFile) {
+        logger.debug('No go.work found');
+      }
```
viceice, line 297: remove,no benefit

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -281,13 +284,29 @@ export async function updateArtifacts({
       execCommands.push(`${cmd} ${args}`);
     }
 
+    const goWorkSumFileName = upath.join(goModDir, 'go.work.sum');
     if (useVendor) {
-      args = 'mod vendor';
-      logger.debug('go mod tidy command included');
-      execCommands.push(`${cmd} ${args}`);
-      if (isGoModTidyRequired) {
-        args = 'mod tidy' + tidyOpts;
```
viceice, line 289: you completely lost this part. is that expected? 🤔

chancez: Yes, this was removed in my first commit, intentionally as it's redundant. There's a `goModTidy`  above that should handle this since the condition above is also on `isGoModTidyRequired` so it was just running `go mod tidy` twice in a row when it only needs to run once.

chancez:
@viceice updated, PTAL

lib/modules/manager/gomod/artifacts.ts
rarkins: This was intentional actually. Somehow the tidying wasn't complete after one so we intentionally ran it twice. In theory it shouldn't be needed

chancez: So I thought the same when I started on this, but when I looked at the commits and comments, it looks like what you're talking about is running `go mod tidy` _after_ vendoring (03b3f09d37). I think once before, once after should be sufficient for all cases.

chancez: I just added it back to simplify review.

chancez:
@viceice I think this is ready to merge, let me know if you have any other changes you would like to see. Thanks!

viceice (approved):

PR #29223, opened 2024-05-23
fix(pr): derive precise rebase description for `rebaseWhen: auto`

sisp:
## Changes

I've fixed the rebase description in the PR body for `rebaseWhen: auto` to derive the precise message according to Renovate's automerge setting and the platform/project setting that determines whether a branch must be up-to-date.

For this to work, I extended the function `getPrConfigDescription(...)` to consider the automerge config value and, if needed, call `platform.getBranchForceRebase(...)` in order to resolve `rebaseWhen: auto` to either `behind-base-branch` or `conflicted`. `getPrConfigDescription(...)` is now asynchronous because `platform.getBranchForceRebase(...)` is asynchronous, and consequently also `getPrBody(...)`, which calls `getPrConfigDescription(...)`, is now asynchronous as well. Also a number of tests needed to be made asynchronous and I mocked the resolved value of `platform.getBranchForceRebase(...)` for testing.

## Context

Before, `rebaseWhen: auto` always resulted in the rebase description for `rebaseWhen: conflicted` even when `rebaseWhen: auto` resolved to `rebaseWhen: behind-base-branch`. Now, the rebase description is more precise, depending on what `rebaseWhen: auto` resolves to.


lib/workers/repository/update/pr/body/config-description.ts
```diff
@@ -23,9 +26,16 @@ export function getPrConfigDescription(config: BranchConfig): string {
...
   prBody += '\n\n';
   prBody += emojify(':recycle: **Rebasing**: ');
-  if (config.rebaseWhen === 'behind-base-branch') {
+  const rebaseWhen =
+    config.rebaseWhen === 'auto'
+      ? config.automerge === true ||
+        (await platform.getBranchForceRebase?.(config.baseBranch))
+        ? 'behind-base-branch'
+        : 'conflicted'
+      : config.rebaseWhen;
```
rarkins, line 35: Is this intended to to reflect the same result as from https://github.com/renovatebot/renovate/blob/2d2d0b9e735625a46568234d8924dff40784a273/lib/workers/repository/update/branch/reuse.ts#L46-L52 ?

sisp: Yes, I duplicated the logic from there. We could factor it out into a helper function if you prefer.

rarkins: Is it possible to _store_ that result and pass it through? Instead of calling it twice.

sisp: I think we could memoize the resolved value of the returned promise of `platform.getBranchForceRebase(...)`, so the first time this function is called and the promise is resolved, the value is cached and subsequent calls will take the cached value instead of recomputing it. To avoid implementing this memoization for each platform implementation, I could try adding it in [`platform.initPlatform(...)`](https://github.com/renovatebot/renovate/blob/7392dbe63179ae6aaad0b9dc1b582a1beb277b06/lib/modules/platform/index.ts#L43). WDYT?

rarkins: I meant that I prefer to save the result into the config which is passed through

sisp: Would you mind sketching a bit more concretely what you envision the implementation to look like? `platform.getBranchForceRebase(...)` has only the base branch name as argument, so this function cannot update the config without changing its signature. Also, each platform implements this function, so a naive implementation would duplicate the config updating logic, which violates the DRY principle. Alternatively, we could introduce a helper function which calls `platform.getBranchForceRebase(...)` and updates the config, but we'd need to make sure that always the helper function and not `platform.getBrancheForceRebase(...)` is called –this feels a bit error-prone IMO. That's why I thought to add a transparent memoization wrapper in `platform.initPlatform(...)`.

rarkins: Current code is this:

```
  if (
    config.rebaseWhen === 'behind-base-branch' ||
    (await shouldKeepUpdated(config, baseBranch, branchName)) ||
    (config.rebaseWhen === 'auto' &&
      (config.automerge === true ||
        (await platform.getBranchForceRebase?.(config.baseBranch))))
  ) {
```

Could it instead be changed to something like this?

```
  if (config.rebaseWhen === 'auto') {
    if (config.automerge === true) {
      logger.debug('Converting rebaseWhen=auto to rebaseWhen=behind-base-branch because automerge=true');
      config.rebaseWhen = 'behind-base-branch';
    }
    if (await platform.getBranchForceRebase?.(config.baseBranch)) {
      logger.debug('Converting rebaseWhen=auto to rebaseWhen=behind-base-branch because platform is configured to require up-to-date branches');
      config.rebaseWhen = 'behind-base-branch';
    }
  }
  if (
    config.rebaseWhen === 'behind-base-branch' ||
    (await shouldKeepUpdated(config, baseBranch, branchName))
  ) {
```

sisp: I see. Yes, I believe that could work. _But_ it makes a precise rebase description highly dependent on the execution order, i.e. it will only be precise when `shouldReuseExistingBranch(...)` is called before `getPrConfigDescription(...)`, which doesn't feel too robust. If you prefer this implementation over some kind of memoization of `platform.getBranchForceRebase(...)`, I can of course do it.

rarkins: This is part of branch logic, which always comes before PR logic, so we should make sure it's deterministic

sisp: Okay, I've updated `shouldReuseExistingBranch(...)` along the lines of what you suggested: 856d36b WDYT?

rarkins: Have you tested it on your example repo(s) to see if it satisfies your requirements still?

sisp: No, and as it turned out: good point! I've pushed another commit that fixes retaining the updated config. Now, I've tested it on a real repo and the rebase description is properly derived.

lib/workers/repository/update/branch/index.ts
```diff
@@ -463,7 +463,8 @@ export async function processBranch(
       );
       config.reuseExistingBranch = false;
     } else {
-      config = { ...config, ...(await shouldReuseExistingBranch(config)) };
+      const parentConfig = await shouldReuseExistingBranch(config);
+      config = { ...config, ...parentConfig };
```
rarkins, line 467: How is this different?

sisp: Before, `{ ...config, /* ... */ }` was first included in the new object before `await shouldReuseExistingBranch(config)` had resolved and internally updated `config`, so the updated `config.rebaseWhen` value was not included in the new object and thus not passed through. Now, `await shouldReuseExistingBranch(config)` gets resolved first and `config.rebaseWhen` gets updated, then the two configs `{ ...config, ...parentConfig }` get merged where `config` has already been updated.

sisp: See also [what `config = { ...config, ...parentConfig };` is transpiled to](https://www.typescriptlang.org/play/?#code/MYewdgZglg5gBAXjgbzgOg6SsA06MAOAhgE4CmYALgMLjTwC+A3EA), this makes it clear:

```js
config = Object.assign(Object.assign({}, config), parentConfig);
```

rarkins: OK, so this is because we added a mutation in shouldReuseExistingBranch()? I hadn't realized this was what I was proposing. Can we instead _return_ the changed rebaseWhen?

sisp: Not without extending the [`ParentBranch` type](https://github.com/renovatebot/renovate/blob/26337ac124f4dd1885a44ed65822f2944ad69f3d/lib/workers/repository/update/branch/reuse.ts#L7-L11). `shouldReuseExistingBranch(...)` resolves to a `ParentBranch` object which is not a `BranchConfig` and does not have the `rebaseWhen` key.

rarkins: I think this is preferable and more easier to maintain in future than the current mutation

sisp: Shall I simply add `rebaseWhen` to the `ParentBranch` type although it doesn't seem semantically related? Or shall I merge the config with the current return value of `shouldReuseExistingBranch(...)` and change its return type to `BranchConfig`?

viceice: yes, merge inside that function and return the new config. change return type too of cause

lib/workers/repository/update/branch/reuse.ts
```diff
@@ -34,21 +34,44 @@ async function shouldKeepUpdated(
 
 export async function shouldReuseExistingBranch(
   config: BranchConfig,
-): Promise<ParentBranch> {
+): Promise<BranchConfig> {
   const { baseBranch, branchName } = config;
+  // eslint-disable-next-line no-param-reassign
+  config = { ...config };
   const result: ParentBranch = { reuseExistingBranch: false };
```
viceice, line 41:
```suggestion
   const result: BranchConfig = { ...config, reuseExistingBranch: false };
```
then return result again, so less changes.

sisp: It needs a bit more than that: Since we mutate the `rebaseWhen` property of the config, we need to perform this mutation on `result` and not on `config` after this config merge because the change it wouldn't be part of the returned `result` object otherwise. And for consistency, all further references to the config in this function should use `result` because that's the up-to-date config object. But the failing tests from before are still failing and I can't think of an explanation.

See 2831d2f2e632d5384f5c75c81789a1805ea2eb2d.

viceice (dismissed):
LGTM

viceice:
needs some more test fixes

sisp:
I don't think the tests need fixing because they were passing for f335126f0bf16624e469b34e3df3f959161c2f8a and the current state of this PR should be only a refactoring, but now some tests are failing and I don't know why. Something is fishy ...

viceice:
> I don't think the tests need fixing because they were passing for f335126f0bf16624e469b34e3df3f959161c2f8a and the current state of this PR should be only a refactoring, but now some tests are failing and I don't know why. Something is fishy ...

maybe because the changed function is mocked in tests, so it doesn't return the full config.

sisp:
@viceice You're exactly right, the mocked function caused the tests to fail. :+1: Before, the line

```ts
config = { ...config, ...(await shouldReuseExistingBranch(config)) };
```

was effectively

```ts
config = Object.assign(Object.assign({}, config), undefined);
```

when the function `shouldReuseExistingBranch()` was mocked in the tests which is equivalent to:

```ts
config = { ...config }
```

With the new implementation

```ts
config = await shouldReuseExistingBranch(config))
```

it is effectively

```ts
config = undefined
```

when the function `shouldReuseExistingBranch()` is mocked in the tests

To restore the previous behavior, I've added a mock implementation for `shouldReuseExistingBranch()` that simply returns the config: 4831975b49911318fd3165f6f48909b377ee2291

viceice (approved):

lib/workers/repository/update/branch/reuse.ts
rarkins: @viceice was this resolved?

viceice: yes

viceice (approved):

PR #29291, opened 2024-05-28
feat(sbt): Support Scala 3 dependency resolution

### LINKED ISSUE #29087: Support Scala 3 dependency resolution in sbt manager
### Discussed in https://github.com/renovatebot/renovate/discussions/29082

Starting with scala 3, dependencies are published with major version.
Here is an example:

```
https://mvnrepository.com/artifact/io.circe/circe-core_2.13/0.15.0-M1
https://mvnrepository.com/artifact/io.circe/circe-core_3/0.15.0-M1
```

As you may see, for scala 3 there is only a major version. However if we have scalaVersion defined in build.sbt:

```
ThisBuild / scalaVersion := "3.4.1"
```

Renovate attempts to resolve dependency with _3.4 prefix, can not find it via sbt manager and fallbacks to maven (which does not work in all cases, e.g. when library pushed without maven support).

george-wilson-rea:
## Changes

Scala 2 packages have their artefact name suffixed by the major and minor version, eg. `_2.13`.
Scala 3 packages are different - their names are suffixed by only the major version, `_3` rather than being suffixed by the major and minor version, eg. `_3.4`

This change detects Scala 3 code and uses the correct shorter suffix.

## Context

Closes #29087

- [x] I'm not sure, but I'm happy to update documentation if necessary

### DIFF
--- a/lib/modules/manager/sbt/extract.ts
+++ b/lib/modules/manager/sbt/extract.ts
@@ -201,9 +201,9 @@ function depHandler(ctx: Ctx): Ctx {
   delete ctx.variableName;
 
   const depName = `${groupId!}:${artifactId!}`;
-  const isScala3 = scalaVersion !== undefined && scalaVersion.length > 0 && scalaVersion.charAt(0) === '3';
+  const isScala3 = scalaVersion?.[0] === '3';
   const scalaVersionForPackageName = isScala3
-    ? scalaVersion.substring(0, scalaVersion.lastIndexOf('.'))
+    ? '3'
     : scalaVersion;
 
   const dep: PackageDependency = {
--- a/lib/modules/manager/sbt/extract.ts
+++ b/lib/modules/manager/sbt/extract.ts
@@ -202,15 +202,15 @@ function depHandler(ctx: Ctx): Ctx {
 
   const depName = `${groupId!}:${artifactId!}`;
   const isScala3 = scalaVersion?.[0] === '3';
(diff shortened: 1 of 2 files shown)


lib/modules/manager/sbt/extract.ts
```diff
@@ -201,12 +201,16 @@ function depHandler(ctx: Ctx): Ctx {
   delete ctx.variableName;
 
   const depName = `${groupId!}:${artifactId!}`;
+  const isScala3 = scalaVersion !== undefined && scalaVersion.length > 0 && scalaVersion.charAt(0) === '3';
```
rarkins, line 204:
```suggestion
  const isScala3 = scalaVersion?.[0] === '3';
```

lib/modules/manager/sbt/extract.ts
```diff
@@ -201,12 +201,16 @@ function depHandler(ctx: Ctx): Ctx {
...
+  const scalaVersionForPackageName = isScala3
+    ? scalaVersion.substring(0, scalaVersion.lastIndexOf('.'))
```
rarkins, line 206:
```suggestion
    ? '3'
```

Isn't this simpler and always correct?

rarkins (changes requested):

rarkins (dismissed):

george-wilson-rea:
Sorry for the churn; I needed to run `prettier`.

rarkins (approved):

PR #29329, opened 2024-05-29
feat(replacements): add airbnb-prop-types-to-prop-types-tools

ljharb:
## Changes

Replace `airbnb-prop-types` with its new name, `prop-types-tools`

## Context

Airbnb gave me the repo, but made me rename the npm package to not have "airbnb" in it.

### DIFF
--- a/lib/config/presets/internal/replacements.ts
+++ b/lib/config/presets/internal/replacements.ts
@@ -53,7 +53,7 @@ export const presets: Record<string, Preset> = {
     ignoreDeps: [], // Hack to improve onboarding PR description
   },
 // eslint-disable-next-line sort-keys
-  'airbnb-prop-types-to-prop-types-tools': { // eslint-disable-line sort-keys
+  'airbnb-prop-types-to-prop-types-tools': {
     description: '`airbnb-prop-types` was given to a new maintainer and renamed to `prop-types-tools`.',
     packageRules: [
       {
--- a/lib/config/presets/internal/replacements.ts
+++ b/lib/config/presets/internal/replacements.ts
@@ -10,6 +10,7 @@ export const presets: Record<string, Preset> = {
   all: {
     description: 'Apply crowd-sourced package replacement rules.',
     extends: [
+      'replacements:airbnb-prop-types-to-prop-types-tools',
       'replacements:apollo-server-to-scoped',
       'replacements:babel-eslint-to-eslint-parser',
       'replacements:containerbase',
@@ -50,6 +51,18 @@ export const presets: Record<string, Preset> = {
       'replacements:zap',
     ],
     ignoreDeps: [], // Hack to improve onboarding PR description
+  },
+  'airbnb-prop-types-to-prop-types-tools': {
+    description: '`airbnb-prop-types` was given to a new maintainer and renamed to `prop-types-tools`.',
+    packageRules: [
+      {
+        matchCurrentVersion: '^2',
+        matchDatasources: ['npm'],
+        matchPackageNames: ['airbnb-prop-types'],
+        replacementName: 'prop-types-tools',
+        replacementVersion: '2.17.0',
+      },
+    ],
   },
   'apollo-server-to-scoped': {
     description: '`apollo-server` packages became scoped.',
--- a/lib/config/presets/internal/replacements.ts
+++ b/lib/config/presets/internal/replacements.ts
@@ -52,7 +52,7 @@ export const presets: Record<string, Preset> = {
     ],
     ignoreDeps: [], // Hack to improve onboarding PR description
   },
-  'airbnb-prop-types-to-prop-types-tools': {
+  'airbnb-prop-types-to-prop-types-tools': { // eslint-disable-line sort-keys
     description: '`airbnb-prop-types` was given to a new maintainer and renamed to `prop-types-tools`.',
     packageRules: [
       {
--- a/lib/config/presets/internal/replacements.ts
+++ b/lib/config/presets/internal/replacements.ts
@@ -52,6 +52,7 @@ export const presets: Record<string, Preset> = {
     ],
     ignoreDeps: [], // Hack to improve onboarding PR description
   },
+// eslint-disable-next-line sort-keys
   'airbnb-prop-types-to-prop-types-tools': { // eslint-disable-line sort-keys
     description: '`airbnb-prop-types` was given to a new maintainer and renamed to `prop-types-tools`.',
     packageRules: [
--- a/lib/config/presets/internal/replacements.ts
+++ b/lib/config/presets/internal/replacements.ts
@@ -52,9 +52,10 @@ export const presets: Record<string, Preset> = {
     ],
     ignoreDeps: [], // Hack to improve onboarding PR description
   },
-// eslint-disable-next-line sort-keys
+  // eslint-disable-next-line sort-keys
   'airbnb-prop-types-to-prop-types-tools': {
-    description: '`airbnb-prop-types` was given to a new maintainer and renamed to `prop-types-tools`.',
+    description:
+      '`airbnb-prop-types` was given to a new maintainer and renamed to `prop-types-tools`.',
     packageRules: [
       {
         matchCurrentVersion: '^2',


lib/config/presets/internal/replacements.ts
```diff
@@ -51,6 +52,18 @@ export const presets: Record<string, Preset> = {
...
   },
+  'airbnb-prop-types-to-prop-types-tools': {
+    '`airbnb-prop-types` was given to a new maintainer and renamed',
+    packageRules: [
+      {
+        matchCurrentVersion: '^2',
+        matchDatasources: ['npm'],
+        matchPackageNames: ['airbnb-prop-types'],
+        replacementName: 'prop-types-tools',
+        replacementVersion: '^2.17.0',
```
rarkins, line 63: I think this needs to be a version and not range?

rarkins (changes requested):

lib/config/presets/internal/replacements.ts
secustor: Yes that is correct

HonkingGoose, line 63: The [Renovate developer docs, replacement presets](https://github.com/renovatebot/renovate/blob/main/docs/development/creating-editing-renovate-presets.md#replacement-presets) says:

> 3. Replacements should update the user to the first recommended version of the new dependency and not include any new changes - whether breaking or not - if they can be avoided

I assume that `2.17.0` is the first recommended version? Currently the `replacementVersion` uses the range syntax `^2.17.0`. The way I'm reading the rule means we should _pin_ the version instead:

```suggestion
        replacementVersion: '2.17.0',
```

lib/config/presets/internal/replacements.ts
```diff
@@ -51,6 +52,18 @@ export const presets: Record<string, Preset> = {
     ],
     ignoreDeps: [], // Hack to improve onboarding PR description
   },
+  'airbnb-prop-types-to-prop-types-tools': {
+    '`airbnb-prop-types` was given to a new maintainer and renamed',
```
HonkingGoose, line 56:
```suggestion
    description: '`airbnb-prop-types` was given to a new maintainer and renamed to `prop-types-tools`.',
```

Changes:

- Add `description:` part before the description
- End description sentence with a full stop: `.`
- Add name of replacement package to the description

HonkingGoose (changes requested):

lib/config/presets/internal/replacements.ts
HonkingGoose, line 56: I was kinda expecting a test to shout at us for missing the `description:` field. Should we test for this?

lib/config/presets/internal/replacements.ts
ljharb: ah ok, but if the latest is 2.17.1 or 2.18.0 i'd want that to be what's used :-/ why would the replacement version need to be pegged?

ljharb: yes, that's the first one, but the semantics of a range are what users will want

lib/config/presets/internal/replacements.ts
ljharb: oh duh thanks

lib/config/presets/internal/replacements.ts
ljharb: or do i have to come back here and make a PR every time i publish a v2 update to prop-types-tools?

HonkingGoose: I think this is the sequence that happens:

1. At the time of merging the replacement preset `2.17.0` is latest
2. @ljharb releases new version, let's say `2.50.0`
3. Renovate end-users gets the replacement PR to the pinnede `2.17.0` version, and merges it.
4. Renovate creates a PR to update to the latest version `2.50.0`, when it meets the other requirements (passed the `minimumReleaseAge` days, for example)
5. @ljharb releases new major `v3` version
6. Renovate creates a update PR for the new major

So you do _not_ need to update the "replacement preset" after you're done with this PR. :wink:

HonkingGoose: I'll let a maintainer respond to your point. :wink:

secustor: The idea is that, the replacement is identical to the last version under the old naming schema. 

Later upgrades of dependencies are than provided separatly, after an user merges this. 

The reasoning is that way you should not get any CI failures which are actually reasoned on actual code changes, but rather only references to the old package. This makes it easier and more predictable changing packages.

rarkins (dismissed):

lib/config/presets/internal/replacements.ts
```diff
@@ -51,6 +52,18 @@ export const presets: Record<string, Preset> = {
     ],
     ignoreDeps: [], // Hack to improve onboarding PR description
   },
+  'airbnb-prop-types-to-prop-types-tools': {
```
HonkingGoose, line 55: `@rarkins` enabled auto-merge but ESLint complains about the sorting order of the new preset name:

> **ESLint failure message:**
> Expected object keys to be in natural insensitive ascending order. 'airbnb-prop-types-to-prop-types-tools' should be before 'all'.

rarkins: That's funny/annoying. Wonder if we should add an ignore so that we can keep all at the top

HonkingGoose: I searched the ESLint docs, but I can't seem to find any block ignore syntax like: `<--- eslint-ignore-this-block --->` and then `<--- eslint-start-checking-again -->`. With Prettier you can ignore a specific portion of the code, but I don't know if ESLint has that.

We can tell ESLint to ignore the whole file, but that may cause bigger problems later.

- [ESLint docs for `sort-keys` rule](https://eslint.org/docs/latest/rules/sort-keys)
- [ESLint docs, Ignore Files](https://eslint.org/docs/latest/use/configure/ignore)

lib/config/presets/internal/replacements.ts
ljharb: sure. but semver means that they won't get CI failures from any v2 upgrade :-)

ljharb:
oops, sorry, i forgot that you don't want me to force push :-/ i do it everywhere else on github so it's a habit.

lib/config/presets/internal/replacements.ts
HonkingGoose: The new syntax is:

> ## Disabling Rules
> 
> ### Using configuration comments
> 
> * **Use with Caution.** Disabling ESLint rules inline should be restricted and used only in situations with a clear and
>       valid reason for doing so. Disabling rules inline should not be the default solution to resolve linting errors.
> * **Document the Reason.** Provide a comment explaining the reason for disabling a particular rule after the `--` section of the comment. This
>       documentation should clarify why the rule is being disabled and why it is necessary in that specific situation.
> * **Temporary Solutions.** If a disable comment is added as a temporary measure to address a pressing issue, create a follow-up task to address the underlying problem adequately. This ensures that the
>       disable comment is revisited and resolved at a later stage.
> * **Code Reviews and Pair Programming.** Encourage team members to review each other's code regularly. Code reviews can help
>       identify the reasons behind disable comments and ensure that they are used appropriately.
> * **Configurations.** Whenever possible, prefer using ESLint configuration files over disable comments. Configuration
>       files allow for consistent and project-wide rule handling.
> 
> To disable rule warnings in a part of a file, use block comments in the following format:
> 
> ```js
> /* eslint-disable */
> 
> alert('foo');
> 
> /* eslint-enable */
> ```
> 
> You can also disable or enable warnings for specific rules:
> 
> ```js
> /* eslint-disable no-alert, no-console */
> 
> alert('foo');
> console.log('bar');
> 
> /* eslint-enable no-alert, no-console */
> ```
> 
> ::: warning
> `/* eslint-enable */` without any specific rules listed causes all disabled rules to be re-enabled.
> :::
> 
> To disable rule warnings in an entire file, put a `/* eslint-disable */` block comment at the top of the file:
> 
> ```js
> /* eslint-disable */
> 
> alert('foo');
> ```
> 
> You can also disable or enable specific rules for an entire file:
> 
> ```js
> /* eslint-disable no-alert */
> 
> alert('foo');
> ```
> 
> To ensure that a rule is never applied (regardless of any future enable/disable lines):
> 
> ```js
> /* eslint no-alert: "off" */
> 
> alert('foo');
> ```
> 
> To disable all rules on a specific line, use a line or block comment in one of the following formats:
> 
> ```js
> alert('foo'); // eslint-disable-line
> 
> // eslint-disable-next-line
> alert('foo');
> 
> /* eslint-disable-next-line */
> alert('foo');
> 
> alert('foo'); /* eslint-disable-line */
> ```
> 
> To disable a specific rule on a specific line:
> 
> ```js
> alert('foo'); // eslint-disable-line no-alert
> 
> // eslint-disable-next-line no-alert
> alert('foo');
> 
> alert('foo'); /* eslint-disable-line no-alert */
> 
> /* eslint-disable-next-line no-alert */
> alert('foo');
> ```
> 
> To disable multiple rules on a specific line:
> 
> ```js
> alert('foo'); // eslint-disable-line no-alert, quotes, semi
> 
> // eslint-disable-next-line no-alert, quotes, semi
> alert('foo');
> 
> alert('foo'); /* eslint-disable-line no-alert, quotes, semi */
> 
> /* eslint-disable-next-line no-alert, quotes, semi */
> alert('foo');
> 
> /* eslint-disable-next-line
>   no-alert,
>   quotes,
>   semi
> */
> alert('foo');
> ```
> 
> All of the above methods also work for plugin rules. For example, to disable `eslint-plugin-example`'s `rule-name` rule, combine the plugin's name (`example`) and the rule's name (`rule-name`) into `example/rule-name`:
> 
> ```js
> foo(); // eslint-disable-line example/rule-name
> foo(); /* eslint-disable-line example/rule-name */
> ```
> 
> ::: tip
> Comments that disable warnings for a portion of a file tell ESLint not to report rule violations for the disabled code. ESLint still parses the entire file, however, so disabled code still needs to be syntactically valid JavaScript.
> :::

[ESLint docs, Disabling rules](https://eslint.org/docs/latest/use/configure/rules#disabling-rules)

lib/config/presets/internal/replacements.ts
secustor: That would be the ideal scenario, though that is sadly often not the case.

lib/config/presets/internal/replacements.ts
rarkins: Maybe just a disable next line of possible?

HonkingGoose, line 55:
```suggestion
  'airbnb-prop-types-to-prop-types-tools': { // eslint-disable-line sort-keys
```

I hope the above correctly disables the `sort-keys` rule for just the `'airbnb-prop-types-to-prop-types-tools'` object.

If the above does not work, maybe we can try this format with the _next line_ variant:

```ts
// eslint-disable-next-line sort-keys
  'airbnb-prop-types-to-prop-types-tools': { 
```

Read the [ESLint docs, disabling rules](https://eslint.org/docs/latest/use/configure/rules#disabling-rules) to learn more.

lib/config/presets/internal/replacements.ts
```diff
@@ -51,6 +52,18 @@ export const presets: Record<string, Preset> = {
     ],
     ignoreDeps: [], // Hack to improve onboarding PR description
   },
+  'airbnb-prop-types-to-prop-types-tools': { // eslint-disable-line sort-keys
```
HonkingGoose, line 55:
```suggestion
  'airbnb-prop-types-to-prop-types-tools': {
    // eslint-disable-line sort-keys
```

I think this should fix the Prettier lint error? I hope that preserves the ESLint ignore though...

ljharb: it won’t preserve the ignore. I think you need disable-next-line, on the previous line.

HonkingGoose:
```suggestion
// eslint-disable-next-line sort-keys
  'airbnb-prop-types-to-prop-types-tools': { // eslint-disable-line sort-keys
```

Let's try this then... :upside_down_face:

lib/config/presets/internal/replacements.ts
```diff
@@ -51,6 +52,19 @@ export const presets: Record<string, Preset> = {
     ],
     ignoreDeps: [], // Hack to improve onboarding PR description
   },
+// eslint-disable-next-line sort-keys
+  'airbnb-prop-types-to-prop-types-tools': { // eslint-disable-line sort-keys
```
HonkingGoose, line 56:
```suggestion
  'airbnb-prop-types-to-prop-types-tools': {
```

Whoops, we don't need this then. :see_no_evil:

lib/config/presets/internal/replacements.ts
HonkingGoose: @rarkins do you think we should test for missing `description` fields in our presets?

The docs use the `description` field to display a short summary of the preset. So it's best if it's always filled in.

rarkins: For replacements perhaps

HonkingGoose: Do you want me to create an issue to check the `description` for _replacement_ presets only?

viceice (dismissed):

viceice:
now prettier complains😂

HonkingGoose:
This is getting too much like Whack-A-Mole:

1. ESLint complains
2. Add ESLint ignore rule
3. Prettier complains
4. Try again with different format
5. Prettier complains again

Silly idea incoming: maybe we need a _Prettier_ ignore for the _ESLint_ ignore... :upside_down_face:

rarkins (approved):

HonkingGoose:
@ljharb congratulations on getting your first PR merged. 🥳 Thank you for sticking with us during the review process! ❤️ 

I'm happy we finally managed to get our linter ducks in a row. 🦆 🦆 🦆

ljharb:
Thanks :-p

Issue #29148, opened 2024-05-18
Improve accessibility for screenreaders


HonkingGoose:
### Describe the proposed change(s).

Todo:

- [ ] #29022
- [ ] Improve docs homepage/index file, based on the readme improvements
- [ ] Check that all images have a good `alt` text, add/edit if needed

PR #29369, opened 2024-05-31
docs(homepage): convert Markdown list to plain HTML list

HonkingGoose:
## Changes

- Convert Markdown list to plain HTML list

## Context

Helps with #29148, but does not close it.

I recently added a Markdown list to the homepage wrapped in a `<details>` tag, that lists the companies and projects that use Renovate. The goal is for screenreader users to expand the `<details>` element and then "have the screenreader read the list". With a proper list you can step through the list items, and go backwards and forwards in the list. Here's the PR that added this change:

- #29345

The changes from the PR above are now live on the published docs, and the `<details>` admonition expands into a long string: `- Prisma - Netlify - Envoy - Condé Nast - Microsoft - Atlassian - Sourcegraph - Mozilla - Deloitte - Telus - Yarn - HashiCorp - Automattic - Algolia - eBay - Cypress - Red Hat - Financial Times - Uber - Buildkite`. It should be a proper list! :melting_face: 

Renovate homepage, shows expanded admonition with badly formatted list

Maybe this problem is caused by my Markdown formatting? Or maybe I should force the correct rendering with `<ul`> and `<li`> HTML elements.

I'll try plain HTML elements first, to see if MkDocs and Material for MkDocs can render this properly.

### DIFF
--- a/docs/usage/index.md
+++ b/docs/usage/index.md
@@ -80,28 +80,28 @@ Renovate is used by:
 
 <details>
 <summary>List of companies and projects that use Renovate</summary>
-
-- Prisma
-- Netlify
-- Envoy
-- Condé Nast
-- Microsoft
-- Atlassian
-- Sourcegraph
-- Mozilla
-- Deloitte
-- Telus
-- Yarn
-- HashiCorp
-- Automattic
-- Algolia
-- eBay
-- Cypress
-- Red Hat
-- Financial Times
-- Uber
-- Buildkite
-
+<ul>
+  <li>Prisma</li>
+  <li>Netlify</li>
+  <li>Envoy</li>
+  <li>Condé Nast</li>
+  <li>Microsoft</li>
+  <li>Atlassian</li>
+  <li>Sourcegraph</li>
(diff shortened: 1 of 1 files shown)


rarkins (approved):

viceice:
I'll try to check the root cause the next days. not enough time now. sorry.

rarkins:
Shall we merge though?

HonkingGoose:
I think we're safe to merge, as plain HTML elements _should_ work, fingers crossed.

If plain HTML elements _still_ render badly, we'll have to dive deeper for sure.

I look forward to your analysis later @viceice! :heart:

HonkingGoose:
The plain HTML list works well on the published docs. 🥳

I guess we need to use a specific number of spaces, or tabs, or other formatting for a _Markdown_-based list to work. We likely will need to tell Prettier to stay away from the custom formatting in the collapsible admonition.

I'll let `@viceice` figure out the correct Markdown, or root cause for why it does _not_ work in Markdown. 🙂

PR #29388, opened 2024-06-02
fix(cargo): Correct bump of non-range values

zharinov:
## Changes

I've noticed Renovate doesn't perform updates for some dependencies released during last month.

The reason is the regression introduced recently, which led to incorrect `getNewValue(...)` calculation.

This PR makes sure `update-lockfile` just bumps the input value, if it isn't range.

## Context

- Ref: #28826


lib/modules/versioning/cargo/index.spec.ts
```diff
@@ -99,58 +99,59 @@ describe('modules/versioning/cargo/index', () => {
...
+    ${'<=1.3.4'}             | ${'replace'}         | ${'1.2.3'}     | ${'1.5.0'}      | ${'<=1.5.0'}
+    ${'<= 1.3.4'}            | ${'replace'}         | ${'1.2.3'}     | ${'1.5.0'}      | ${'<= 1.5.0'}
+    ${'<=   1.3.4'}          | ${'replace'}         | ${'1.2.3'}     | ${'1.5.0'}      | ${'<= 1.5.0'}
+    ${'>= 0.1.21, < 0.2.0'}  | ${'bump'}            | ${'0.1.21'}    | ${'0.1.24'}     | ${'>= 0.1.24, < 0.2.0'}
+    ${'>= 0.1.21, <= 0.2.0'} | ${'bump'}            | ${'0.1.21'}    | ${'0.1.24'}     | ${'>= 0.1.24, <= 0.2.0'}
+    ${'>= 0.0.1, <= 0.1'}    | ${'bump'}            | ${'0.0.1'}     | ${'0.0.2'}      | ${'>= 0.0.2, <= 0.1'}
+    ${'>= 1.2.3, <= 1'}      | ${'bump'}            | ${'1.2.3'}     | ${'1.2.4'}      | ${'>= 1.2.4, <= 1'}
+    ${'>= 1.2.3, <= 1.0'}    | ${'bump'}            | ${'1.2.3'}     | ${'1.2.4'}      | ${'>= 1.2.4, <= 1.2'}
+    ${'>= 0.0.1, < 0.1'}     | ${'bump'}            | ${'0.1.0'}     | ${'0.2.1'}      | ${'>= 0.2.1, < 0.3'}
+    ${'0.6.7'}               | ${'update-lockfile'} | ${'0.6.7'}     | ${'0.6.9'}      | ${'0.6.9'}
```
zharinov, line 154: This single line was added

rarkins (commented):
If we're updating lock file then shouldn't the package file value remain unchanged if the update is "in range"?

rarkins:
In Cargo 0.6.7 means ~0.6.7, which covers 0.6.9

zharinov:
Hm, I see, but also I expect the package [`winnow`](https://crates.io/crates/winnow/versions) to be updated here in the project [without lockfile](https://github.com/zharinov/verver/blob/main/Cargo.toml#L11) 🤔

zharinov:
I think it's correct for projects with lockfiles, but without lockfile it isn't obvious why there is no updates. If it was bump for such cases, maybe it would be better, though now I'm not sure.

viceice (approved):

rarkins (changes requested):
When updating lock file I don't think we should narrow ranges in the cargo toml

Issue #14756, opened 2022-03-22
Cannot handle yarn v3 and private registry


negibokken:
### How are you running Renovate?

WhiteSource Renovate hosted app on github.com

### Was this something which used to work for you, and then stopped?

I never saw this working

### Describe the bug

Hi, thank you for developing such great software!

I met a problem with yarn v3 and private registry (registry.npmjs.org). 
The problem is that when I specify `npmScopes` in `.yarnrc.yml`, I got an error message (`Error: Environment variable not found (NPM_TOKEN)`) on the lockfile update step in the Docker process (Please see for the detail logs in Relevant debug logs section).

I specified `hostRules` and encrypted token in `renovate.json` so Renovate can fetch the version information but Renovate cannot generate lockfile (In [this PR](https://github.com/negibokken/renovate-private-package-repro1/pull/2), only `package.json` file is updated).

The minimum repro repo is below.

https://github.com/negibokken/renovate-private-package-repro1

## etc.

When I don't specify `npmScopes` in `.yarnrc.yml` then Renovate can update lockfile (The error message goes away). But then we can't install it in the local machine because we don't specify `${NPM_TOKEN}` for the private registry.

The minimum repro repo for the above case is below. (We can see the lockfile is updated in [the PR](https://github.com/negibokken/renovate-private-package-repro3/pull/2))

https://github.com/negibokken/renovate-private-package-repro3

So Renovate doesn't require the `npmScopes` because `renovate.json` has a private registry setting and [Renovate generates](https://github.com/renovatebot/renovate/blob/c14732416960159843253772989afcc2553c8a84/lib/modules/manager/npm/post-update/rules.ts#L12) `npmRegistries` from `hostRules` while processing lockfile.(*1)

## Ideas

Currently, I have two ideas to handle this issue.

1.  Pass the environment variables to Docker process
    - We can specify the environment variable [here](https://github.com/renovatebot/renovate/blob/c14732416960159843253772989afcc2553c8a84/lib/modules/manager/npm/post-update/yarn.ts#L95)?
2. Parse `.yarnrc.yml` and remove `npmScopes` that includes environment variables (like `${NPM_TOKEN}`)
[...]

PR #29417, opened 2024-06-03
docs(yarn): docs to clarify expected setup for yarn 2+

Cynocracy:
This documents what the local .yarnrc.yml should look like, and that it CANNOT contain a top level reference to NPM_TOKEN (unlike .npmrc)

## Changes

Doc-only change

## Context

Clarify yarn 2+ private registry support https://github.com/renovatebot/renovate/issues/14756 

### DIFF
--- a/docs/usage/getting-started/private-packages.md
+++ b/docs/usage/getting-started/private-packages.md
@@ -399,13 +399,13 @@ npmRegistries:
     npmAuthToken: <Decrypted PAT Token>
 ```
 
-Thus, to allow for private registries to be used locally via the `NPM_TOKEN` environment variable AND with renovate, your local `.yarnrc.yml` should contain
+Thus, to allow for private registries to be used locally via the `NPM_TOKEN` environment variable _and_ with Renovate, your local `.yarnrc.yml` must contain:
+
+
 ```yaml
 npmRegistries:
    //npm.pkg.github.com/:
     npmAuthToken: "${NPM_TOKEN}"
-```
-which will be automatically updated by renovate via the hostRules. You *cannot* set a top level `npmAuthToken` in your `.yarnrc.yml` file, as it is not overriden.
 
 ### maven
 
--- a/docs/usage/getting-started/private-packages.md
+++ b/docs/usage/getting-started/private-packages.md
@@ -409,12 +409,6 @@ npmRegistries:
     npmAuthToken: "${NPM_TOKEN}"
 ```
 
-or
-
-```yaml
-npmAuthToke: "${NPM_TOKEN:-}"
-```
-
 
 ### maven
 
--- a/docs/usage/getting-started/private-packages.md
+++ b/docs/usage/getting-started/private-packages.md
@@ -360,9 +360,10 @@ The end-result looks like this:
 
 #### Yarn 2+
 
-Renovate doesn't support reading `npmRegistries` and `npmScopes` from `.yarnrc.yml`, so `hostRules` (or `npmToken`) and `npmrc` should be configured like above. The `${NPM_TOKEN}` substitution that takes place for `.npmrc` will also not occur for `.yarnrc.yml`.
+In Yarn 2+, a [.yarnrc.yml](https://yarnpkg.com/configuration/yarnrc) file can be used with to configure npm authentication for private registries via `npmRegistries` or via a top level `npmAuthToken`.
 
-Renovate updates `npmRegistries` in `.yarnrc.yml` with resolved `hostRules` before running Yarn.
+Renovate updates `npmRegistries` in `.yarnrc.yml` with resolved `hostRules` before running Yarn, but does not overwrite `npmAuthToken`.
+
 For Renovate to overwrite existing `npmRegistries` entry, the key should match the `matchHost` minus the protocol (`http:` or `https:`) plus the trailing slash.
 
 For example, the Renovate configuration:
@@ -406,6 +407,14 @@ Thus, to allow for private registries to be used locally via the `NPM_TOKEN` env
 npmRegistries:
    //npm.pkg.github.com/:
     npmAuthToken: "${NPM_TOKEN}"
+```
+
+or
+
+```yaml
+npmAuthToke: "${NPM_TOKEN:-}"
+```
+
 
 ### maven
 
--- a/docs/usage/getting-started/private-packages.md
+++ b/docs/usage/getting-started/private-packages.md
@@ -360,7 +360,8 @@ The end-result looks like this:
 
 #### Yarn 2+
 
-Renovate doesn't support reading `npmRegistries` and `npmScopes` from `.yarnrc.yml`, so `hostRules` (or `npmToken`) and `npmrc` should be configured like above.
+Renovate doesn't support reading `npmRegistries` and `npmScopes` from `.yarnrc.yml`, so `hostRules` (or `npmToken`) and `npmrc` should be configured like above. The `${NPM_TOKEN}` substitution that takes place for `.npmrc` will also not occur for `.yarnrc.yml`.
+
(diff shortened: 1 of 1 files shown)


docs/usage/getting-started/private-packages.md
```diff
@@ -360,7 +360,8 @@ The end-result looks like this:
 
 #### Yarn 2+
 
-Renovate doesn't support reading `npmRegistries` and `npmScopes` from `.yarnrc.yml`, so `hostRules` (or `npmToken`) and `npmrc` should be configured like above.
+Renovate doesn't support reading `npmRegistries` and `npmScopes` from `.yarnrc.yml`, so `hostRules` (or `npmToken`) and `npmrc` should be configured like above. The `${NPM_TOKEN}` substitution that takes place for `.npmrc` will also not occur for `.yarnrc.yml`.
```
rarkins, line 363:
```suggestion
Renovate doesn't support reading `npmRegistries` and `npmScopes` from `.yarnrc.yml`, so `hostRules` (or `npmToken`) and `npmrc` should be configured like above.
The `${NPM_TOKEN}` substitution that takes place for `.npmrc` will also not occur for `.yarnrc.yml`.
```

docs/usage/getting-started/private-packages.md
```diff
@@ -398,6 +399,14 @@ npmRegistries:
     npmAuthToken: <Decrypted PAT Token>
 ```
 
+Thus, to allow for private registries to be used locally via the `NPM_TOKEN` environment variable AND with renovate, your local `.yarnrc.yml` should contain
+```yaml
+npmRegistries:
+   //npm.pkg.github.com/:
+    npmAuthToken: "${NPM_TOKEN}"
+```
+which will be automatically updated by renovate via the hostRules. You *cannot* set a top level `npmAuthToken` in your `.yarnrc.yml` file, as it is not overriden.
```
rarkins, line 408:
```suggestion

```yaml
npmRegistries:
   //npm.pkg.github.com/:
    npmAuthToken: "${NPM_TOKEN}"
```

The registry will be automatically updated by renovate via the hostRules.
Setting a top level `npmAuthToken` in your `.yarnrc.yml` file is not supported by Renovate, as it is not overriden.
```

**Review by rarkins — CHANGES_REQUESTED** [review]

File: docs/usage/getting-started/private-packages.md [hunk 3]
```diff
@@ -398,6 +399,14 @@ npmRegistries:
     npmAuthToken: <Decrypted PAT Token>
 ```
 
+Thus, to allow for private registries to be used locally via the `NPM_TOKEN` environment variable AND with renovate, your local `.yarnrc.yml` should contain
```
HonkingGoose, line 402:
```suggestion
Thus, to allow for private registries to be used locally via the `NPM_TOKEN` environment variable _and_ with Renovate, your local `.yarnrc.yml` must contain:

```

Changes:

- Use _emphasis_ styling, instead of CAPSLOCK text
- Capitalize proper nouns or brandnames, like Renovate
- Lead-in text for a codeblock usually needs to end with a `:`
- If a user must do something, for things to work properly, we use the keyword must
- Use a newline between the lead-in text and the codeblock to match Prettier linter formatting

docs/usage/getting-started/private-packages.md
HonkingGoose: @rarkins your suggestion is not displaying properly on GitHub.

I think GitHub does not like it when you have backticks within backticks, like in a review suggestion. I can't get those kind of review suggestion to work properly either. :upside_down_face:

docs/usage/getting-started/private-packages.md
HonkingGoose: This text is hard to read and understand. How about something like this?

> #### Yarn 2+
>
> Introductory sentence that explains the default Yarn 2+ configuration.
> If we recommend one method of Yarn use above another, say so here as well.
> I suppose the default Yarn 2+ config works well with Renovate?
> 
> ##### When you use a `.yarnrc.yml` file
>
> Renovate can _not_ read these these config options from `.yarnrc.yml` files:
>
> - `npmRegistries`
> - `npmScopes`
> 
> When you use a `.yarnrc.yml` file Renovate will _not_ substitute the `${NPM_TOKEN}`, like it does when you use a `.npmrc` file.
> 
> This all means you must configure your `hostRules` (or `npmToken`) and `npmrc` like ...insert proper reference to correct config here...

docs/usage/getting-started/private-packages.md
```diff
@@ -360,8 +360,10 @@ The end-result looks like this:
 
 #### Yarn 2+
 
-Renovate doesn't support reading `npmRegistries` and `npmScopes` from `.yarnrc.yml`, so `hostRules` (or `npmToken`) and `npmrc` should be configured like above.
```
rarkins, line 363: Why have you removed this information?

docs/usage/getting-started/private-packages.md
```diff
@@ -398,6 +400,16 @@ npmRegistries:
     npmAuthToken: <Decrypted PAT Token>
 ```
 
+Thus, to allow for private registries to be used locally via the `NPM_TOKEN` environment variable _and_ with Renovate, your local `.yarnrc.yml` must contain:
```
rarkins, line 403: I am not sure this works, and this does not seem to be the recommendation from the linked discussion

rarkins (changes requested):

docs/usage/getting-started/private-packages.md
HonkingGoose: That's my fault! I suggested changes/edits to the line, as it was very hard to understand.

How about this then?

> #### Yarn 2+
>
> ##### `.yarnrc.yml` file limitations
>
> Renovate can _not_ read the following configuration items from `.yarnrc.yml` files:
> 
> - `npmRegistries`
> - `npmScopes`
>
> Use the following configuration options instead, to get a similar result:
> 
> - `hostRules` (or `npmToken`)
> - `npmrc`
>
> ...Rest of docs...

docs/usage/getting-started/private-packages.md
Cynocracy: The discussion is confused, so I can understand why it wouldn't be clear.

https://github.com/renovatebot/renovate/issues/14756#issuecomment-1578651075

my experience is similar to the one link here, in an internal repo we have


.yarnrc.yml

```
nodeLinker: node-modules

npmRegistries:
  "//registry.yarnpkg.com/":
    npmAuthToken: "${NPM_TOKEN}"

```

and renovate.json

```

  "hostRules": [
    {
      "matchHost": "https://registry.yarnpkg.com/",
      "hostType": "npm",
      "encrypted": {
        "token": "[REDACTED]"
      }
    }
  ]
```

yarn locally works because the .yarnrc.yml, and it is able to pull the private registries fine in renovate.

The ${NPM_TOKEN:-} issue discussed is only relevant if you have a use of the NPM_TOKEN environment variable that is not replaced by the host matching rules, so I did not include it in the text.

This all works as long as the fuzzy matched merging performed at https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/npm/post-update/index.ts#L553-L559 correctly overwrites the required env var.

Does that make sense? Would you like me to try to publish a repro? (I'm not sure exactly what I would use a private registry)

docs/usage/getting-started/private-packages.md
Cynocracy: Ah, yes, I think the text here is a little misleading still, it says "Renovate can not read [...] npmRegistries"
but that is the field that is written, so while it doesn't 'read' it, it's not totally unawares of it.

What about

```
Yarn 2+
.yarnrc.yml file limitations
Renovate can be configured to update the following items from .yarnrc.yml files:

npmRegistries
npmScopes
Use the following configuration options to update them:

hostRules (or npmToken)
npmrc
...Rest of docs...
```

Issue #29322, opened 2024-05-29
Glasskube Support


secustor:
### Discussed in https://github.com/renovatebot/renovate/discussions/29206

<div type='discussions-op-text'>

<sup>Originally posted by **kosmoz** May 22, 2024</sup>
### Tell us more.

[Glasskube](https://github.com/glasskube/glasskube/) is a package manager for Kubernetes clusters. Glasskube and Glasskube packages can be managed via GitOps solutions, such as Flux or ArgoCD, in which case our users would love to get notified about updates directly in their repositories.</div>

kosmoz:
Hi, we (Glasskube) would love to contribute this feature! 

Do you have any suggestions about where to best get started?

rarkins:
Hi @kosmoz, that's great to hear! We recommend you implement in this order:
- Datasource
- Versioning
- Manager

Datasource or Datasource+versioning won't be particularly useful on their own, although potentially someone could already use our custom manager with them for some value. However it simplifies things to work on them and merge incrementally. The final validation would be a real e2e repo test once the manager PR is ready.

For datasource, check out [the readme](https://github.com/renovatebot/renovate/blob/main/lib/modules/datasource/readme.md) and then it can be easiest if you pick an existing datasource which is similar and start by copy/pasting. Here was the most recent datasource addition for reference: https://github.com/renovatebot/renovate/pull/27971

kosmoz:
I already started working on the versioning scheme in parallel to the datasource, but I'm not sure where to push my changes to best fit your workflow. Should I create a new PR based on `main`?

rarkins:
Yea there should be no overlaps

PR #29430, opened 2024-06-04
feat(datasource): add glasskube packages datasource

kosmoz:
## Changes

This PR adds a new datasource for updating packages used by the Glasskube package manager.

## Context

See #29322

Disclaimer: I added unit tests to fulfill the coverage goal, but did not test on a real repository because I could not find out how to do that without a matching manager.


viceice (changes requested):

lib/modules/datasource/glasskube-packages/index.ts
```diff
@@ -0,0 +1,77 @@
+import urljoin from 'url-join';
```
secustor, line 1: Use https://github.com/renovatebot/renovate/blob/main/lib/util/url.ts#L8 instead

lib/modules/datasource/glasskube-packages/index.ts
```diff
@@ -0,0 +1,77 @@
...
+      logger.error(
+        { error, registryUrl },
+        'Failed to request releases from Glasskube packages datasource',
+      );
+      return null;
+    }
+
+    result.releases = versions.versions.map((it) => ({
+      version: it.version,
+      isStable: it.version === versions.latestVersion,
```
secustor, line 47: Are all versions except the latest one considered pre releases? 

Think `-SNAPSHOT` for Maven/Gradle for example

secustor (changes requested):

lib/modules/datasource/glasskube-packages/index.ts
kosmoz: No, the `latestVersion` typically represents the newest version that has undergone an additional manual vetting step after being added to the repository. We generally do not include prerelease versions.

Should I change this so that `isStable` is always false?

secustor: You can simply remove it as the system will assume that the version is stable

lib/modules/datasource/glasskube-packages/index.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  override defaultRegistryUrls = [
+    GlasskubePackagesDatasource.defaultRegistryUrl,
+  ];
+
+  constructor() {
+    super(GlasskubePackagesDatasource.id);
+  }
+
+  override async getReleases({
```
secustor, line 24:
```suggestion
  @cache({
    namespace: `datasource-${GlasskubePackagesDatasource.id}`,
    key: ({ packageName }: GetReleasesConfig) => `${packageName}`,
  })
  override async getReleases({
```
Let's cache the results

secustor (changes requested):

lib/modules/datasource/glasskube-packages/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    let versions: GlasskubePackageVersions;
+    const result: ReleaseResult = { releases: [] };
+
+    try {
+      const response = await this.http.get(
+        joinUrlParts(registryUrl!, packageName, 'versions.yaml'),
+      );
+      versions = GlasskubePackageVersionsYaml.parse(response.body);
+    } catch (error) {
+      logger.error(
```
viceice, line 37: do we log this as errors in other datasources? 🤔 I think we just pass most errors up.

lib/modules/datasource/glasskube-packages/schema.ts
```diff
@@ -0,0 +1,23 @@
+import { z } from 'zod';
+import { Yaml } from '../../../util/schema-utils';
+
+export const GlasskubePackageVersions = z.object({
```
viceice, line 4:
```suggestion
const GlasskubePackageVersions = z.object({
```
we don't use it directly

lib/modules/datasource/glasskube-packages/schema.ts
```diff
@@ -0,0 +1,23 @@
...
+  latestVersion: z.string(),
+  versions: z.array(z.object({ version: z.string() })),
+});
+
+export const GlasskubePackageManifest = z.object({
```
viceice, line 9:
```suggestion
const GlasskubePackageManifest = z.object({
```
same

lib/modules/datasource/glasskube-packages/index.ts
kosmoz: I took the Unity3D datasource as reference and it also logs this kind of error. But I can remove the logging if desired, just let me know.

kosmoz:
I added the default versioning, as suggested by @secustor (https://github.com/renovatebot/renovate/issues/29322#issuecomment-2153349999).

Unfortunately, I had to force push the branch because GitHub didn't like the way I previously merged the upstream changes. Hope this didn't mess up the reviewer history, since it didn't affect the old commits… :upside_down_face:

lib/modules/datasource/glasskube-packages/index.ts
viceice: I think at least some errors need to be rethrown, like `ExternalHostError`.

viceice: use this datasource as a better sample 

https://github.com/renovatebot/renovate/blob/49b1402058400cc2fc8d0c9ebe8cfba5765f781c/lib/modules/datasource/ruby-version/index.ts#L67

viceice (commented):
see comments

secustor (dismissed):

lib/modules/datasource/glasskube-packages/index.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+  it('should handle error response on versions request', async () => {
+    httpMock
+      .scope(versionsUrl.origin)
+      .get(versionsUrl.pathname)
+      .reply(500, 'internal server error');
+    await expect(
+      getPkgReleases({
+        datasource: GlasskubePackagesDatasource.id,
+        packageName: 'cloudnative-pg',
+        registryUrls: [GlasskubePackagesDatasource.defaultRegistryUrl],
```
viceice, line 28: I think this should be removed, the datasource doesn't seem to support custom registry urls.

if it supports then, then use different urls for testing.

otherwise remove registry urls from tests.

kosmoz: The datasource should be able to support custom registries. Therefore, the tests should use different URLs. Did I understand that correctly?

secustor: Your interpretation is correct, yeah.

lib/modules/datasource/glasskube-packages/index.spec.ts
secustor, line 1: There are no test with default registry url anymore.

lib/modules/datasource/glasskube-packages/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  GlasskubePackageManifestYaml,
+  GlasskubePackageVersions,
+  GlasskubePackageVersionsYaml,
+} from './schema';
+
+export class GlasskubePackagesDatasource extends Datasource {
+  static readonly id = 'glasskube-packages';
+  static readonly defaultRegistryUrl =
+    'https://packages.dl.glasskube.dev/packages';
+
```
secustor, line 16:
```suggestion
  override readonly customRegistrySupport = true;
```
Activate custom registry support.

lib/modules/datasource/glasskube-packages/index.spec.ts
viceice: add tests for both cases, default and non-default registry

viceice (commented):
see comments

lib/modules/datasource/glasskube-packages/index.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+      .get(defaultVersionUrl.pathname)
+      .reply(200, versionsYaml);
+    httpMock
+      .scope(defaultPackageManifestUrl.origin)
+      .get(defaultPackageManifestUrl.pathname)
+      .reply(200, packageManifestYaml);
+    const response = await getPkgReleases({
+      datasource: GlasskubePackagesDatasource.id,
+      packageName: 'cloudnative-pg',
+      registryUrls: [],
```
viceice, line 129:
```suggestion
```

lib/modules/datasource/glasskube-packages/index.ts
```diff
@@ -0,0 +1,75 @@
...
+        joinUrlParts(registryUrl!, packageName, 'versions.yaml'),
+      );
+      versions = GlasskubePackageVersionsYaml.parse(response.body);
+    } catch (err) {
+      this.handleGenericErrors(err);
+    }
+
+    result.releases = versions.versions.map((it) => ({
+      version: it.version,
+      registryUrl,
```
viceice, line 49:
```suggestion
```
no need to put it in every release

lib/modules/datasource/glasskube-packages/__fixtures__/versions.yaml
```diff
@@ -0,0 +1,4 @@
+latestVersion: v1.23.1+1
```
viceice, line 1: is this always the highest version? if not, then we should return it as latest tag, so users can use follow tag config option.

kosmoz: No, as previously discussed (https://github.com/renovatebot/renovate/pull/29430#discussion_r1626325139), there might be higher versions than this.

viceice: add it as tags.latest on release result. so users can opt-in to follow that tag to only allow version lower or equal to this version.

kosmoz:
Hey folks, just wanted to ask, is there anything for me left to do for this PR?  :slightly_smiling_face:

lib/modules/datasource/glasskube-packages/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    GlasskubePackagesDatasource.defaultRegistryUrl,
+  ];
+
+  constructor() {
+    super(GlasskubePackagesDatasource.id);
+  }
+
+  @cache({
+    namespace: `datasource-${GlasskubePackagesDatasource.id}`,
+    key: ({ packageName }: GetReleasesConfig) => `${packageName}`,
```
viceice, line 29: this should include the registry url, otherwise we get caching issues with different registries.

lib/modules/datasource/glasskube-packages/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  ];
+
+  constructor() {
+    super(GlasskubePackagesDatasource.id);
+  }
+
+  @cache({
+    namespace: `datasource-${GlasskubePackagesDatasource.id}`,
+    key: ({ registryUrl, packageName }: GetReleasesConfig) =>
+      [registryUrl, packageName].filter((s) => s).join('-'),
```
secustor, line 30: Why the noop filter with a join?

```suggestion
      `${registryUrl}-${packageName}`
```

kosmoz: Because `registryUrl` can be undefined, at least according to the type definition (which would result in `"-foo"`). But if that doesn't matter, your proposed change makes sense.

secustor: `registryUrl` can not be `undefined` as long there is a `defaultRegistryUrl` defined.

lib/modules/datasource/glasskube-packages/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  ];
+
+  constructor() {
+    super(GlasskubePackagesDatasource.id);
+  }
+
+  @cache({
+    namespace: `datasource-${GlasskubePackagesDatasource.id}`,
+    key: ({ registryUrl, packageName }: GetReleasesConfig) =>
+      `${registryUrl!}-${packageName}`
```
viceice, line 30:
```suggestion
      `${registryUrl!}:${packageName}`
```
we prefer the colon

viceice (dismissed):

secustor:
Needs prettier run

lib/modules/datasource/glasskube-packages/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  ];
+
+  constructor() {
+    super(GlasskubePackagesDatasource.id);
+  }
+
+  @cache({
+    namespace: `datasource-${GlasskubePackagesDatasource.id}`,
+    key: ({ registryUrl, packageName }: GetReleasesConfig) =>
+      `${registryUrl!}:${packageName}`,
```
viceice, line 30:
```suggestion
      `${registryUrl}:${packageName}`,
```
do we really need this?

viceice (dismissed):

secustor (approved):

PR #29438, opened 2024-06-04
chore(lookup): increase log level of "Found no results ..." message

amezin:
## Changes

Add a warning when no valid package versions can be found in the datasource.

## Context

This is almost always a configuration error. It happened to me multiple times when I configured versioning or extractVersion incorrectly. And the issue is hard to notice when there are no updates available immediately, because logs and reports only show newer versions from lookup results.

There is a debug-level log message already, which I currently remap to warning. But I'm a bit afraid that it'll change or disappear.

### DIFF
--- a/lib/workers/repository/process/lookup/index.spec.ts
+++ b/lib/workers/repository/process/lookup/index.spec.ts
@@ -3749,7 +3749,12 @@ describe('workers/repository/process/lookup/index', () => {
         sourceUrl: 'https://github.com/nodejs/node',
         updates: [],
         versioning: 'node',
-        warnings: [],
+        warnings: [
+          {
+            message: 'Found no versions of package node in datasource docker',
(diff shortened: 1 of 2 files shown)


secustor:
@nabeelsaabna Do you use the warn level for telemetry in the App? 
This will potentially raise a lot of additional warnings.

rarkins:
My instinct is that although this seems like a good idea, similar things have bitten us in the past. Let's keep in mind that users can use logLevelRemap to raise logs to WARN if they want

nabeelsaabna:
> @nabeelsaabna Do you use the warn level for telemetry in the App? This will potentially raise a lot of additional warnings.

nope

secustor (commented):
Based on the discussion above please reduce it to an info log message

secustor (approved):

PR #29485, opened 2024-06-06
fix(github): fetch all open issue if ignorePrAuthor=true

### LINKED ISSUE #28999: ignorePrAuthor=true should mean all issues are fetched on GitHub

When ignorePrAuthor=true, Renovate should fetch all issues, not just the ones by the bot itself.

This might also mean that the initRepo() issues query should have more issues fetched in such cases

Equivalent for GitLab: https://github.com/renovatebot/renovate/pull/28996

RahulGautamSingh:
## Changes
- Add username to the api call options only if `ignorePrAuthor` is false.

## Context
Closes: https://github.com/renovatebot/renovate/issues/28999

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2065,7 +2065,7 @@ An additional use case is for GitLab users of project or group access tokens who
 If `ignorePrAuthor` is configured to true, it means Renovate will fetch the entire list of repository PRs instead of optimizing to fetch only those PRs which it created itself.
 You should only want to enable this if you are changing the bot account (e.g. from `@old-bot` to `@new-bot`) and want `@new-bot` to find and update any existing PRs created by `@old-bot`.
 
-Setting this field to `true` in Github, GitLab will also mean that all Issues will be fetched instead of only those by the bot itself.
+Setting this field to `true` in Github or GitLab will also mean that all Issues will be fetched instead of only those by the bot itself.
 
 ## ignorePresets
 
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2065,7 +2065,7 @@ An additional use case is for GitLab users of project or group access tokens who
 If `ignorePrAuthor` is configured to true, it means Renovate will fetch the entire list of repository PRs instead of optimizing to fetch only those PRs which it created itself.
 You should only want to enable this if you are changing the bot account (e.g. from `@old-bot` to `@new-bot`) and want `@new-bot` to find and update any existing PRs created by `@old-bot`.
 
-Setting this field to `true` in GitLab will also mean that all Issues will be fetched instead of only those by the bot itself.
+Setting this field to `true` in Github, GitLab will also mean that all Issues will be fetched instead of only those by the bot itself.
 
 ## ignorePresets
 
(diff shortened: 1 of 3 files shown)


docs/usage/configuration-options.md
```diff
@@ -2065,7 +2065,7 @@ An additional use case is for GitLab users of project or group access tokens who
 If `ignorePrAuthor` is configured to true, it means Renovate will fetch the entire list of repository PRs instead of optimizing to fetch only those PRs which it created itself.
 You should only want to enable this if you are changing the bot account (e.g. from `@old-bot` to `@new-bot`) and want `@new-bot` to find and update any existing PRs created by `@old-bot`.
 
-Setting this field to `true` in GitLab will also mean that all Issues will be fetched instead of only those by the bot itself.
+Setting this field to `true` in Github, GitLab will also mean that all Issues will be fetched instead of only those by the bot itself.
```
rarkins, line 2068:
```suggestion
Setting this field to `true` in Github or GitLab will also mean that all Issues will be fetched instead of only those by the bot itself.
```

lib/modules/platform/github/index.ts
```diff
@@ -1206,15 +1206,15 @@ export async function setBranchStatus({
 
 // Issue
 
-async function getIssues(): Promise<Issue[]> {
+export async function getIssues(): Promise<Issue[]> {
```
rarkins, line 1209: Necessary?

rarkins (changes requested):

lib/modules/platform/github/index.ts
RahulGautamSingh: NO

RahulGautamSingh, line 1209:
```suggestion
async function getIssues(): Promise<Issue[]> {
```

rarkins (dismissed):

lib/modules/platform/github/index.ts
```diff
@@ -1214,7 +1214,7 @@ async function getIssues(): Promise<Issue[]> {
       variables: {
         owner: config.repositoryOwner,
         name: config.repositoryName,
-        user: config.renovateUsername,
+        ...(!config.ignorePrAuthor && { user: config.renovateUsername }),
```
rarkins, line 1217: Isn't user a mandatory field of this graphql query?

rarkins: Tested?

RahulGautamSingh: Yes and it passed :)

But actually the test was faulty because on re-test I found that the `ignorePrAuthor` value I set in my `renovate.json` was not being used. Instead the default `false` was still being passed to the `getIssues` fn, thus the `user` was always being set.

I will draft it for now and check why the correct `ignorePrAuthor` is not being passed to the `getIssues` fn. I doubt that gitlab might have the same issue.

RahulGautamSingh: Is `ignorePrAuthor` supposed to be self hosted config option?

RahulGautamSingh: Tested again locally ☑️

lib/modules/platform/github/graphql.ts
```diff
@@ -41,7 +41,7 @@ export const getIssuesQuery = `
 query(
   $owner: String!,
   $name: String!,
-  $user: String!,
+  $user: String,
```
rarkins, line 44: Does it work to just leave it undefined?

RahulGautamSingh: Yes

viceice (approved):

PR #29487, opened 2024-06-06
feat(hostRules/matchHost): massage and validate 

### LINKED ISSUE #29425: hostRules: improve matchHost validation and parsing

Goal: massage "incorrect" matchHost values with ports or paths, for example `gitlab.company.com:8080` or `registry.company.com/some/path`.

Improve validation: if a matchHost value contains `"` or `/` then we should add `https://` as the prefix and check if it's a valid URL. If it's not then I think it's safe for us to raise a config validation error. If it's valid, then we don't need to migrate it, because some config.js configurations may not be possible to migrate (the value comes from env).

Also improve validation of empty matchHost. If the `matchHost` key is defined, but the _value_ is `undefined`, then we should warn. This is commonly caused by empty process.env.SOMETHING being used. We should similar do for empty string `""`.

Finally, we should enhance the addRule function to do similar logic and prefix the matchHost with `https://` if it contains `:` or `/`.

RahulGautamSingh:
## Changes
- Prefix `matchHost` url which contain `:` or `/` with `htps://` (migrated as well)
- Validate `matchHost` urls
   1. if its an empty string
   2. check is it is a valid url if the url includes `://` (ie. only check for valid if url has scheme)
   
   

## Context
Closes: https://github.com/renovatebot/renovate/issues/29425


lib/config/validation.spec.ts
```diff
@@ -1134,6 +1134,44 @@ describe('config/validation', () => {
       ]);
     });
 
+    it('errors if invalid matchHost values in hostRules', async () => {
+      GlobalConfig.set({ allowedHeaders: ['X-*'] });
+
+      const config = {
+        hostRules: [
```
rarkins, line 1141: Please add one with a missing matchHost to make sure that _doesn't_ error. e.g. have hostType and token only

lib/config/migrations/custom/host-rules-migration.spec.ts
```diff
@@ -27,6 +27,7 @@ describe('config/migrations/custom/host-rules-migration', () => {
           { hostName: 'some.domain.com', token: '123test' },
           { endpoint: 'domain.com/', token: '123test' },
           { host: 'some.domain.com', token: '123test' },
+          { host: 'some.domain.com:8080', token: '123test' },
```
rarkins, line 30:
```suggestion
          { matchHost: 'some.domain.com:8080', token: '123test' },
```

lib/config/validation.ts
```diff
@@ -814,6 +814,25 @@ export async function validateConfig(
...
+        if (is.nonEmptyString(rule.matchHost)) {
+          if (rule.matchHost.includes('://')) {
+            try {
+              new URL(rule.matchHost);
+            } catch (err) {
+              errors.push({
+                topic: 'Configuration Error',
+                message: `hostRules matchHost \`${rule.matchHost}\` is not a valid URL.`,
+              });
+            }
```
rarkins, line 826: Use parseUrl() from utils instead, then no need to try/catch

rarkins (changes requested):

rarkins:
Please test for both global and repo config

RahulGautamSingh:
> Please test for both global and repo config

Done ☑️

rarkins (dismissed):

lib/config/validation.ts
```diff
@@ -814,6 +815,23 @@ export async function validateConfig(
           ? (config.allowedHeaders as string[]) ?? []
           : GlobalConfig.get('allowedHeaders', []);
       for (const rule of val as HostRule[]) {
+        if (is.nonEmptyString(rule.matchHost)) {
+          if (rule.matchHost.includes('://')) {
+            if (is.null_(parseUrl(rule.matchHost))) {
```
viceice, line 820:
```suggestion
            if (null === parseUrl(rule.matchHost)) {
```
looks simpler than the `null_` function 🤔

viceice (commented):
otherwise LGTM

lib/config/validation.ts
RahulGautamSingh, line 820:
```suggestion
            if (parseUrl(rule.matchHost) === null) {
```

lib/config/migrations/custom/host-rules-migration.ts
```diff
@@ -94,6 +94,8 @@ function validateHostRule(rule: LegacyHostRule & HostRule): void {
 function massageUrl(url: string): string {
```
viceice, line 94: we should use a single shared function here and inside host rules

viceice (commented):
otherwise LGTM

RahulGautamSingh:
When I export the `massageHostUrl` fn from `util/host-rules` and import it in `host-rules-migration.ts` the `util/sanitize` module doesn't get mocked properly in the `datasource/npm/npmrc.spec.ts` file. I am not sure what is the cause & the fix here.

If I only export the `massageHostUrl` fn from `util/host-rules` but do not import it in `host-rules-migration.ts` the `util/sanitize` module is mocked.

@viceice any suggestions?

rarkins (approved):

viceice (approved):

PR #29519, opened 2024-06-07
feat(self-hosted)!: convert `ignorePrAuthor` to self-hosted option

RahulGautamSingh:
## Changes
- Convert `ignorePrAuthor` option to self-hosted option

## Context
Ref: https://github.com/renovatebot/renovate/discussions/29497


RahulGautamSingh:
I am not sure about the breaking label. Because even if someone may have added the option inside repo config, it would be meaningless.

rarkins:
There is probably a use case for repo config ignorePrAuthor (e.g. someone was running their own bot, then migrated to the hosted app), however if it _doesn't work_ today then I think we can make it global only for now

viceice (commented):
if you think we should merge it as breaking, then please target v38 branch

RahulGautamSingh:
I will do it in a new PR, targeting v38

PR #29523, opened 2024-06-07
feat(self-hosted)!: convert `ignorePrAuthor` to global only option

RahulGautamSingh:
## Changes
- Convert `ignorePrAuthor` to global only option

- [x] Both unit tests + ran on a real repository (tested locally)


lib/modules/platform/bitbucket/index.ts
```diff
@@ -195,7 +194,6 @@ export async function initRepo({
   });
   config = {
     repository,
-    ignorePrAuthor,
```
RahulGautamSingh, line 198: I removed it from bitbucket as it is not used in this platform yet.

rarkins (commented):
When you say it didn't work previously if in repo config, does that apply to both GitLab and github?

RahulGautamSingh:
> When you say it didn't work previously if in repo config, does that apply to both GitLab and github?

Yes

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -335,7 +334,10 @@ export async function getPrList(refreshCache?: boolean): Promise<Pr[]> {
     const searchParams: Record<string, string> = {
       state: 'ALL',
     };
-    if (!config.ignorePrAuthor && config.username !== undefined) {
+    if (
+      !GlobalConfig.get('ignorePrAuthor', false) &&
+      config.username !== undefined
+    ) {
```
rarkins, line 340: I don't think that these changes to use GlobalConfig are necessary. We only use this for _security_ reasons to prevent user config from overriding global config, but that can't happen here as we haven't read user credentials before initRepo().

rarkins (changes requested):

lib/modules/platform/bitbucket-server/index.ts
RahulGautamSingh: Based on these comments:

https://github.com/renovatebot/renovate/blob/a9a17787c57f06ab096cfef31fb4a3dfcb9812c2/lib/config/types.ts#L99-L101


https://github.com/renovatebot/renovate/blob/a9a17787c57f06ab096cfef31fb4a3dfcb9812c2/lib/config/types.ts#L131-L133

rarkins: Good point, although in reality it's "Config options used prior to reading in repo config" or not

rarkins: But I guess we can keep as-is so there's a clean separation

rarkins (approved):

viceice:
is it really breaking when it never worked as expected?

rarkins:
It will now create a config error, whereas it didn't before

RahulGautamSingh:
Too conflicted!

PR #29525, opened 2024-06-08
feat(bitbucket-server)!: autodetect gitAuthor if possible

amezin:
## Changes

Try to detect `gitAuthor` on Bitbucket Server during initialization.

## Context

Unfortunately, there seems to be no API endpoint to get the information about the currently authenticated user. So I'm using an endpoint that depends on username. Actually, the endpoint depends on username slug, but the rules to convert username to slug aren't clear. So I'm just trying to pass the username as is, and hope it works. For all real usernames that I tried - it works. If it fails, the failure isn't fatal.

### DIFF
--- a/lib/modules/platform/bitbucket-server/index.ts
+++ b/lib/modules/platform/bitbucket-server/index.ts
@@ -11,7 +11,7 @@ import type { BranchStatus } from '../../../types';
 import type { FileData } from '../../../types/platform/bitbucket-server';
 import { parseJson } from '../../../util/common';
 import * as git from '../../../util/git';
-import { deleteBranch } from '../../../util/git';
+import { FALLBACK_GIT_AUTHOR, deleteBranch } from '../../../util/git';
 import * as hostRules from '../../../util/host-rules';
 import {
   BitbucketServerHttp,
@@ -157,8 +157,9 @@ export async function initPlatform({
       ).body;
 
       platformConfig.gitAuthor = `${displayName} <${emailAddress}>`;
+      platformConfig.gitIgnoredAuthors = [FALLBACK_GIT_AUTHOR];
 
-      logger.debug(`Detected gitAuthor: ${platformConfig.gitAuthor});
+      logger.debug(`Detected gitAuthor: ${platformConfig.gitAuthor}`);
     } catch (err) {
       logger.debug(
         { err },
--- a/lib/modules/platform/bitbucket-server/index.ts
+++ b/lib/modules/platform/bitbucket-server/index.ts
@@ -135,6 +135,7 @@ export async function initPlatform({
   }
 
   if (!gitAuthor && username) {
+    logger.debug(`Attempting to confirm gitAuthor from username`);
     const options: HttpOptions = {
       memCache: false,
     };
@@ -157,14 +158,11 @@ export async function initPlatform({
 
       platformConfig.gitAuthor = `${displayName} <${emailAddress}>`;
 
-      logger.debug(
-        { gitAuthor: platformConfig.gitAuthor },
-        'Detected gitAuthor',
-      );
+      logger.debug(`Detected gitAuthor: ${platformConfig.gitAuthor});
     } catch (err) {
       logger.debug(
         { err },
-        'Failed to get user info, please configure gitAuthor manually',
+        'Failed to get user info, fallback gitAuthor will be used',
       );
     }
   }
--- a/lib/modules/platform/bitbucket-server/index.ts
+++ b/lib/modules/platform/bitbucket-server/index.ts
@@ -40,6 +40,7 @@ import type {
 } from '../types';
 import { getNewBranchName, repoFingerprint } from '../util';
 import { smartTruncate } from '../utils/pr-body';
+import { UserSchema } from './schema';
 import type {
   BbsConfig,
   BbsPr,
@@ -147,10 +148,11 @@ export async function initPlatform({
 
     try {
       const { displayName, emailAddress } = (
-        await bitbucketServerHttp.getJson<{
-          displayName: string;
-          emailAddress: string;
-        }>(`./rest/api/1.0/users/${username}`, options)
+        await bitbucketServerHttp.getJson(
+          `./rest/api/1.0/users/${username}`,
+          options,
+          UserSchema,
+        )
       ).body;
 
       platformConfig.gitAuthor = `${displayName} <${emailAddress}>`;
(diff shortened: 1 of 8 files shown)


lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -132,6 +133,40 @@ export async function initPlatform({
...
+
+    try {
+      const { displayName, emailAddress } = (
+        await bitbucketServerHttp.getJson<{
+          displayName: string;
+          emailAddress: string;
+        }>(`./rest/api/1.0/users/${username}`, options)
+      ).body;
+
+      platformConfig.gitAuthor = `${displayName} <${emailAddress}>`;
```
rarkins, line 156: Check first if they're defined in the response (we prefer zod)

rarkins (changes requested):
This is a breaking change so should be part of v38

rarkins:
I was under the impression that today it falls back to bot@renovateapp.com for Bitbucket Server if not configured, but is that the case? I'm not seeing evidence of that in the code

amezin:
> I was under the impression that today it falls back to [bot@renovateapp.com](mailto:bot@renovateapp.com) for Bitbucket Server if not configured, but is that the case? I'm not seeing evidence of that in the code

https://github.com/renovatebot/renovate/blob/a9a17787c57f06ab096cfef31fb4a3dfcb9812c2/lib/util/git/index.ts#L277 ?

rarkins:
> > I was under the impression that today it falls back to [bot@renovateapp.com](mailto:bot@renovateapp.com) for Bitbucket Server if not configured, but is that the case? I'm not seeing evidence of that in the code
> 
> https://github.com/renovatebot/renovate/blob/a9a17787c57f06ab096cfef31fb4a3dfcb9812c2/lib/util/git/index.ts#L277
> 
> ? This function (`setGitAuthor`) currently receives `null` on bitbucket-server if not configured manually

That's the one - I had misremembered the default bot value. I think we need to have that as gitIgnoredAuthors in such cases, to assist bitbucket server users who ran prior to v38

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -132,6 +134,41 @@ export async function initPlatform({
     );
   }
 
+  if (!gitAuthor && username) {
```
HonkingGoose, line 137: This PR still targets the `main` branch! I think you want the author to target the `v38` branch?

Also if this is a breaking change it should have the `!` character in the PR title, in the right place. So that Semantic Release groups it under the breaking changes section of the final 38.0.0 changelog.

amezin:
> I think we need to have that as gitIgnoredAuthors in such cases, to assist bitbucket server users who ran prior to v38

Do you mean "add it automatically to the config after successful autodetect on bitbucket-server"?

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -132,6 +134,41 @@ export async function initPlatform({
...
+      platformConfig.gitAuthor = `${displayName} <${emailAddress}>`;
+
+      logger.debug(
+        { gitAuthor: platformConfig.gitAuthor },
+        'Detected gitAuthor',
+      );
+    } catch (err) {
+      logger.debug(
+        { err },
+        'Failed to get user info, please configure gitAuthor manually',
```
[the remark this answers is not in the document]
rarkins: Shouldn't it fall back to bot@renovateapp.com like it used to have?

rarkins (changes requested):
Still needs gitIgnoredAuthors set too

lib/modules/platform/bitbucket-server/index.ts
rarkins, line 167:
```suggestion
        'Failed to get user info, fallback gitAuthor will be used',
```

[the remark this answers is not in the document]
rarkins: It still needs `gitIgnoredAuthors` added

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -132,6 +134,40 @@ export async function initPlatform({
...
+      const { displayName, emailAddress } = (
+        await bitbucketServerHttp.getJson(
+          `./rest/api/1.0/users/${username}`,
+          options,
+          UserSchema,
+        )
+      ).body;
+
+      platformConfig.gitAuthor = `${displayName} <${emailAddress}>`;
+      platformConfig.gitIgnoredAuthors = [FALLBACK_GIT_AUTHOR];
```
rarkins, line 160: Unfortunately I think this might override any configured `gitIgnoredAuthors` by the user. Instead I think we'd be better to configure `renovate@whitesourcesoftware.com` as the default value for `gitIgnoredAuthors`. It's ok to have that for all platforms

rarkins (changes requested):

lib/modules/platform/bitbucket-server/index.ts
amezin: No, it doesn't override. `gitIgnoredAuthors` from `platformConfig` is merged with `gitIgnoredAuthors` from config. See [lib/modules/platform/index.ts](https://github.com/renovatebot/renovate/pull/29525/files#diff-62ca4fa20e113c91022a1baadcef7616e79f6f500b1d31b3d19b892dfa9f3750).

amezin: Or do you think it's unnecessary complexity?

rarkins: I don't have a strong opinion about either approach so if you've already written this one then I'm ok with it

rarkins (approved):

lib/modules/platform/bitbucket-server/index.ts
viceice: This shouldn't be set here. It should be set for all platforms or not at all.
At least it needs to be documented!

viceice (changes requested):
See reopened conversation

lib/modules/platform/bitbucket-server/index.ts
rarkins: OK, let's add `renovate@whitesourcesoftware.com` and `bot@renovateapp.com` as default `gitIgnoredAuthors`

amezin: #30313

viceice (approved):
@rarkins I don't see any breaking changes. 🤔

rarkins:
@amezin can you try undoing your last commit and trying again?

amezin:
> @amezin can you try undoing your last commit and trying again?

I don't get how it should help (and what exactly you mean).

I can rebase this PR on top of new v38 and force-push.

viceice:
needs interactive rebase onto v38

PR #29549, opened 2024-06-09
feat(datasource/custom): remove content limiter for plain

secustor:
## Changes
Removes the header check for the `plain` format of the custom manager

## Context
https://github.com/renovatebot/renovate/discussions/29410

### DIFF
--- a/lib/modules/datasource/custom/formats/plain.ts
+++ b/lib/modules/datasource/custom/formats/plain.ts
@@ -20,11 +20,6 @@ export class PlainFetcher implements CustomDatasourceFetcher {
   async fetch(http: Http, registryURL: string): Promise<unknown> {
     const response = await http.getPlain(registryURL);
 
-    const contentType = response.headers['content-type'];
-    if (!contentType?.startsWith('text/')) {
-      return null;
-    }
-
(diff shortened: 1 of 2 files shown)


lib/modules/datasource/custom/formats/plain.ts
```diff
@@ -20,11 +20,6 @@ export class PlainFetcher implements CustomDatasourceFetcher {
   async fetch(http: Http, registryURL: string): Promise<unknown> {
     const response = await http.getPlain(registryURL);
 
-    const contentType = response.headers['content-type'];
-    if (!contentType?.startsWith('text/')) {
-      return null;
-    }
-
```
viceice, line 27: shouldn't we check for string? otherwise we try to parse blobs and that maybe throws somewhere.

viceice (commented):
otherwise LGTM

lib/modules/datasource/custom/formats/plain.ts
secustor: Based on the types getPlain only can return strings. 
So if we need to add checks, IMO it should be there.

viceice (approved):

PR #29550, opened 2024-06-09
feat(git): add support for SSH-based commit signing

### LINKED ISSUE #18197: SSH based commit and push signing

Add support for ssh based commit and push signing

Pretty much the same way as the existing gnupg signing, but with ssh instead

Maybe

sisp:
## Changes

I've added support for SSH-based commit signing.

## Context

Closes #18197.

The implementation includes a workaround for a quirk in `ssh-keygen` which requires the public SSH key to exist for signing even when the private key is provided directly, which is technically not necessary (see https://github.com/renovatebot/renovate/issues/18197#issuecomment-2152333710).


lib/util/git/private-key.ts
```diff
@@ -5,59 +5,132 @@ import upath from 'upath';
...
 
-export async function writePrivateKey(): Promise<void> {
-  if (!gitPrivateKey) {
-    return;
+class SSHKey extends PrivateKey {
+  protected readonly gpgFormat = 'ssh';
+
+  protected async importKey(): Promise<string | undefined> {
+    if (await this.hasPassphrase()) {
+      throw new Error('SSH key must have an empty passhprase');
```
sisp, line 78: Why is this line not covered? There is a [test case](https://github.com/renovatebot/renovate/pull/29550/files#diff-7c56595536556edce447a7b810a258d538dfb69b20aa61212cd7edd2148e727aR76-R90) ...

sisp: Never mind, my bad. I've fixed it and force-pushed (since there hasn't been a review yet).

sisp:
I'm not sure what to do about the type error. 🤔 Any advice?

rarkins:
Try refactoring the existing part first (separate PR, to be merged first). This is our preferred approach to such changes regardless of any type challenges

sisp:
I think the problem is caused by `@types/jest-when`: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/68605#issue-2130731211

viceice (commented):
I would prefer to not use jest when and stick to current test dependencies

sisp:
Okay, I noticed that [`jest-mock-extended` also has a `calledWith()` extension](https://github.com/marchaos/jest-mock-extended?tab=readme-ov-file#calledwith-extension), so I'll try to modify the tests to use it instead.

lib/util/git/private-key.spec.ts
```diff
@@ -10,39 +12,121 @@ jest.mock('fs-extra', () =>
...
 );
-jest.mock('../exec');
+jest.mock('../exec', () =>
+  jest
+    .requireActual<typeof import('jest-mock-extended')>('jest-mock-extended')
+    .mock<MockProxy<typeof execModule>>(),
+);
 
-const exec = mocked(exec_);
+const exec = (execModule as MockProxy<typeof execModule>).exec;
```
sisp, line 21: See https://github.com/marchaos/jest-mock-extended/issues/87#issuecomment-1419801660 for the source of inspiration.

sisp (commented):
I've refactored the tests to use `jest-mock-extended` instead of `jest-when`. Rewriting the tests also revealed a bug which I've fixed, too. If the CI checks pass, I think this PR is ready for a proper review.

lib/util/git/private-key.spec.ts
```diff
@@ -10,39 +12,121 @@ jest.mock('fs-extra', () =>
...
 );
-jest.mock('../exec');
+jest.mock('../exec', () =>
+  jest
+    .requireActual<typeof import('jest-mock-extended')>('jest-mock-extended')
+    .mock<MockProxy<typeof execModule>>(),
+);
 
-const exec = mocked(exec_);
+const exec = execModule as MockProxy<typeof execModule>;
```
sisp, line 21: See https://github.com/marchaos/jest-mock-extended/issues/87#issuecomment-1419801660 for the source of inspiration.

lib/util/git/private-key.spec.ts
```diff
@@ -10,39 +12,121 @@ jest.mock('fs-extra', () =>
     >('../../../test/fixtures')
     .fsExtra(),
 );
-jest.mock('../exec');
+jest.mock('../exec', () =>
+  jest
+    .requireActual<typeof import('jest-mock-extended')>('jest-mock-extended')
+    .mock<MockProxy<typeof execModule>>(),
```
viceice, line 18:
```suggestion
    .mock(),
```
I think the type here doesn't matter?

why not use `mockDeep` as we use in other tests?

lib/util/git/private-key.spec.ts
```diff
@@ -1,5 +1,7 @@
-import { mocked } from '../../../test/util';
-import * as exec_ from '../exec';
+import fs from 'fs-extra';
+import { MockProxy, any, mockReset } from 'jest-mock-extended';
+import { Fixtures } from '../../../test/fixtures';
+import * as execModule from '../exec';
```
viceice, line 4: please don't rename for no benefit

lib/util/git/private-key.spec.ts
viceice: should we create another helper for this?

lib/util/git/private-key.ts
```diff
@@ -5,59 +5,132 @@ import upath from 'upath';
...
-export function setPrivateKey(key: string | undefined): void {
-  if (!is.nonEmptyStringAndNotWhitespace(key)) {
-    return;
+const sshKeyRegex = regEx(
+  /-----BEGIN ([A-Z ]+ )?PRIVATE KEY-----(.|\n)*?-----END ([A-Z]+ )?PRIVATE KEY-----/,
+);
+
+let gitPrivateKey: PrivateKey | undefined;
+
+abstract class PrivateKey {
```
viceice, line 20: please do the class refactoring in a separate PR to be merged before ssh support.

viceice (changes requested):

lib/util/git/private-key.ts
sisp: Done, see #29875.

lib/util/git/private-key.spec.ts
sisp: Sounds good. I've create a helper: 2f115add610c5d7d9c8678de0407284d6d5ce3a4

lib/util/git/private-key.spec.ts
sisp: Reverted: b2e11ca77d84226013bf28aaa097817df8f6c12f

lib/util/git/private-key.spec.ts
sisp: > I think the type here doesn't matter?

You're right. I've removed the type: 0beb5e0c240520cfc715fde367768b445f328062

> why not use `mockDeep` as we use in other tests?

No specific reason, it just wasn't necessary. For better consistency with other tests, I've migrated to `mockDeep`: f7fa05de65e3e859e5092a3267779b7b949bd8f9

lib/util/git/private-key.ts
sisp: #29875 has been merged and released.

lib/util/git/private-key.spec.ts
```diff
@@ -1,4 +1,7 @@
-import { mocked } from '../../../test/util';
+import fs from 'fs-extra';
+import { any, mockReset } from 'jest-mock-extended';
+import { Fixtures } from '../../../test/fixtures';
+import { mockedDeep } from '../../../test/util';
```
sisp, line 4: Moving this import from the top to here is necessary to satisfy ESLint, otherwise it raises the following error:

```
Error: `../../../test/util` import should occur after import of `../../../test/fixtures`
```

sisp (commented):
@viceice I've updated this PR to include the factored out preparatory work from #29875 and addressed your review feedback. WDYT?

lib/util/git/private-key.spec.ts
```diff
@@ -10,39 +13,121 @@ jest.mock('fs-extra', () =>
     >('../../../test/fixtures')
     .fsExtra(),
 );
-jest.mock('../exec');
+jest.mock('../exec', () =>
+  jest
+    .requireActual<typeof import('jest-mock-extended')>('jest-mock-extended')
+    .mockDeep(),
+);
```
viceice, line 20:
```suggestion
jest.mock('../exec', () => mockDeep());
```

This works on other tests

lib/util/git/private-key.spec.ts
```diff
@@ -10,39 +13,121 @@ jest.mock('fs-extra', () =>
...
 
-const exec = mocked(exec_);
+const exec = mockedDeep(exec_);
 
 describe('util/git/private-key', () => {
   describe('writePrivateKey()', () => {
+    beforeEach(() => {
+      jest.resetAllMocks();
```
viceice, line 27:
```suggestion
```

Done globally

https://github.com/renovatebot/renovate/blob/d65effaa77b53ef93312474e81fdbd0efeff0498/jest.config.ts#L231

test/util.ts
```diff
@@ -20,6 +21,14 @@ export function mocked<T extends object>(module: T): jest.Mocked<T> {
   return jest.mocked(module);
 }
 
+/**
+ * Simple wrapper for getting mocked version of a module
+ * @param module module which is mocked by `jest-mock-extended.mockDeep`
+ */
+export function mockedDeep<T extends object>(module: T): DeepMockProxy<T> {
```
viceice, line 28: Use different name to not clash naming with `mockDeep` from `jest-mock-extended`

lib/util/git/private-key.spec.ts
```diff
@@ -10,39 +13,121 @@ jest.mock('fs-extra', () =>
...
+      mockReset(exec);
```
viceice, line 28: i think the reset above will already reset it?
```suggestion
```

lib/util/git/private-key.ts
```diff
@@ -5,14 +5,22 @@ import upath from 'upath';
...
-import { newlineRegex } from '../regex';
+import type { ExecResult } from '../exec/types';
+import { newlineRegex, regEx } from '../regex';
 import { addSecretForSanitizing } from '../sanitize';
 
+type PrivateKeyFormat = 'gpg' | 'ssh';
+
+const sshKeyRegex = regEx(
+  /-----BEGIN ([A-Z ]+ )?PRIVATE KEY-----(.|\n)*?-----END ([A-Z]+ )?PRIVATE KEY-----/,
+);
```
viceice, line 16:
```suggestion
const sshKeyRegex = regEx(
  /-----BEGIN ([A-Z ]+ )?PRIVATE KEY-----.*?-----END ([A-Z]+ )?PRIVATE KEY-----/,
  's'
);
```

`s` causes `.` to match newlines 😉

lib/util/git/private-key.ts
```diff
@@ -60,11 +71,60 @@ class GPGKey extends PrivateKey {
...
+    await fs.chmod(keyFileName, 0o600);
+    // HACK: `git` calls `ssh-keygen -Y sign ...` internally for SSH-based
+    // commit signing. Technically, only the private key is needed for signing,
+    // but `ssh-keygen` has an implementation quirk which requires also the
+    // public key file to exist. Therefore, we derive the public key from the
+    // private key just to satisfy `ssh-keygen` until the problem has been
+    // resolved.
+    // https://github.com/renovatebot/renovate/issues/18197#issuecomment-2152333710
+    const { stdout } = await exec(`ssh-keygen -y -P "" -f ${keyFileName}`);
+    await fs.outputFile(`${keyFileName}.pub`, stdout);
```
viceice, line 92: private key file needs to be removed when renovate exits.

lib/util/git/private-key.spec.ts
sisp: Done: adb96b8a563553ae2e404e71a45781e933cce1f5

test/util.ts
sisp: How about `mockedExtended`, alluding to the fact that this function uses the `mockDeep` function from `jest-mock-**extended**`? I also realized that `mocked()` always uses a deep mock, so there's no need to add `deep` to the name here either. See 34fae8521fae897d27118cce12f90289fb435c02.

lib/util/git/private-key.spec.ts
sisp: I think you're right, done: f4ec3f5a6c667a3378891921ce187bbd395ca97f

lib/util/git/private-key.ts
sisp: Nice! :ok_hand: Done: 275d0592bb583bc5144c11cc1d7b6c54d20fcaf5

lib/util/git/private-key.ts
sisp: Good point! :ok_hand: I've added exit handlers to delete both private and the public key files and amended a test to check for their successful removal: 9ee2ebca2f42e17fb0ed4345c2185c16df5d6ed8

sisp:
Thanks for the thorough and valuable feedback, @viceice! :bow: I've addressed your remarks. WDYT? :slightly_smiling_face:

lib/util/git/private-key.spec.ts
```diff
@@ -10,39 +13,119 @@ jest.mock('fs-extra', () =>
...
-        stderr: `something wrong`,
-        stdout: '',
-      });
+      exec.exec.calledWith(any()).mockResolvedValue({ stdout: '', stderr: '' });
+      exec.exec
+        .calledWith('gpg --import /tmp/git-private-gpg.key')
+        .mockRejectedValueOnce({
+          stderr: `something wrong`,
+          stdout: '',
+        });
```
viceice, line 39: please do these refactorings in a separate PR. Ideallay this PR doesn't change any existing test to ensure no regressions

viceice:
Please don't force-push. Just merge main if required, should only be required on conflicts

lib/util/git/private-key.spec.ts
sisp: Done, see #30396.

sisp:
I didn't force-push to rebase onto the latest state of `main` but only amended and force-pushed the most recent commit before your subsequent review because it caused the CI pipeline to fail. But if you prefer, I can push new commits for such iterations between reviews.

viceice:
yes, please just push new changes. we squash merge PRs, so don't worry about git history

sisp:
I've merged `main` into here and fixed the temporary directory path like in https://github.com/renovatebot/renovate/commit/1f0ce9abcde5ff90bea2d6e21633d8159d49b597. WDYT, @viceice?

viceice (approved):

philippe-granet:
Hi @sisp , I try to use your feature, but I can't make it work, see https://github.com/renovatebot/renovate/discussions/30999
Can you help us understand what's the problem?

sisp:
It's a bug indeed. :bow: See https://github.com/renovatebot/renovate/discussions/30999#discussioncomment-10442102 and #31005.

PR #29575, opened 2024-06-10
feat(gomod): add support for package replacements

jamietanna:
## Changes

Modifies the `gomod` manager to add support for managing package replacements, if present.

## Context

As part of #24883, towards being able to solve https://github.com/renovatebot/renovate/discussions/29567, we want to add support for managing package replacement with Renovate.

To make it possible for folks using Go modules to be able to take
advantage of Renovate's package replacements, we should add support to
the `gomod` manager.

This adds support for managing the following `updateType`s:

- minor/patch
- major
- digest

For `major` and `digest` updates, we need to support the case that we're
performing a replacement, so add an additional check for the
`updateType=replacement` and whether there's a `newMajor` or
`newDigest`.

Additionally, as we're replacing the dependency's name, we want to have
a separate regex capture group for `depName`, which we can also use for
determining whether there are quotes around the dependency.

As part of #24883.

Example PR: https://github.com/jamietanna/renovate-iss-oapi/pull/2

**Note**: should the `Package` column be updated in the PR description to be the new package?


lib/modules/manager/gomod/update.ts
```diff
@@ -36,11 +35,14 @@ export function updateDependency({
     const lineToChange = lines[upgrade.managerData.lineNumber];
     logger.trace({ upgrade, lineToChange }, 'go.mod current line');
     if (
-      !lineToChange.includes(depNameNoVersion) &&
+      !lineToChange.includes(fromPackageNameNoVersion) &&
```
jamietanna, line 38: Should we also check `toPackageNameNoVersion` here?

lib/modules/manager/gomod/update.ts
```diff
@@ -17,16 +17,15 @@ export function updateDependency({
...
-      logger.warn('gomod manager does not support replacement updates yet');
-      return null;
-    }
+    const { depType, newName } = upgrade;
+    const fromPackageName = upgrade.depName;
+    // newName will be available for replacement
+    const toPackageName = newName ?? fromPackageName;
     // istanbul ignore if: should never happen
-    if (!depName || !upgrade.managerData) {
+    if (!fromPackageName || !toPackageName || !upgrade.managerData) {
```
rarkins, line 25:
```suggestion
    if (!fromPackageName || !upgrade.managerData) {
```

toPackageName will be same as fromPackageName if it's falsey

lib/modules/manager/gomod/update.ts
rarkins: It depends if this code gets executed the "second" time (next run) after a replacement. If so it looks like you need to check both

lib/modules/manager/gomod/update.ts
```diff
@@ -36,11 +35,11 @@ export function updateDependency({
     const lineToChange = lines[upgrade.managerData.lineNumber];
     logger.trace({ upgrade, lineToChange }, 'go.mod current line');
     if (
-      !lineToChange.includes(depNameNoVersion) &&
+      !lineToChange.includes(fromPackageNameNoVersion) &&
       !lineToChange.includes('rethinkdb/rethinkdb-go.v5')
     ) {
       logger.debug(
-        { lineToChange, depName },
+        { lineToChange, depName: toPackageName },
```
rarkins, line 42:
```suggestion
        { lineToChange, depName: fromPackageName },
```

To be consistent with line 38

lib/modules/manager/gomod/update.ts
```diff
@@ -55,19 +54,21 @@ export function updateDependency({
...
     } else if (depType === 'require' || depType === 'indirect') {
       if (upgrade.managerData.multiLine) {
-        updateLineExp = regEx(/^(?<depPart>\s+[^\s]+)(?<divider>\s+)[^\s]+/);
+        updateLineExp = regEx(
+          /^(?<depPart>\s+)(?<depName>[^\s]+)(?<divider>\s+)[^\s]+/,
+        );
       } else {
         updateLineExp = regEx(
-          /^(?<depPart>require\s+[^\s]+)(?<divider>\s+)[^\s]+/,
+          /^(?<depPart>require\s+)(?<depName>[^\s]+)(?<divider>\s+)[^\s]+/,
```
rarkins, line 71: I have concerns about these regex changes breaking something. Unfortunately the existing regex isn't documented

rarkins (changes requested):
I haven't reviewed the second half yet

lib/modules/manager/gomod/update.ts
```diff
@@ -22,7 +22,7 @@ export function updateDependency({
     // newName will be available for replacement
     const toPackageName = newName ?? fromPackageName;
     // istanbul ignore if: should never happen
-    if (!fromPackageName || !toPackageName || !upgrade.managerData) {
+    if (!fromPackageName || !upgrade.managerData) {
```
jamietanna, line 25: we may need to undo this check to ensure that `eslint` knows that we've checked this

jamietanna:
Sorry, re-requested review a bit early - will fix up

jamietanna:
While testing this - [here](https://gitlab.com/jamietanna/oapi-codegen-example-project/-/merge_requests/2) - I've noticed two things:

The package updates table presents the old package name with a `replacement` noted, like so:

| Package | Type | Update | Change |
|---|---|---|---|
| [github.com/deepmap/oapi-codegen](https://github.com/deepmap/oapi-codegen) | require | replacement | `v1.12.4` -> `v2.3.0` |

Should we instead show the new name?

| Package | Type | Update | Change |
|---|---|---|---|
| [github.com/deepmap/oapi-codegen](https://github.com/deepmap/oapi-codegen) -> [github.com/oapi-codegen/oapi-codegen](https://github.com/oapi-codegen/oapi-codegen/v2) | require | replacement | `v1.12.4` -> `v2.3.0` |

This is likely a wider change, as it probably impacts other cases where we're doing replacements, so may be worth a separate PR.

Also, I've noticed that the `gomodUpdateImportPaths` doesn't seem to be run - it looks like this is [because of this line](https://github.com/jamietanna/renovate/blob/e9dff6a877a73b9019afe10b427216298cad57e9/lib/modules/manager/gomod/artifacts.ts#L250-L253) in which we only perform this check on a `major` update, but in this case, we're performing a major version bump.

Thoughts on whether these are things we should try and close off here before this PR goes in?

lib/modules/manager/gomod/update.ts
```diff
@@ -17,14 +17,15 @@ export function updateDependency({
...
+    const { depType } = upgrade;
     const currentName = upgrade.depName;
-    if (updateType === 'replacement') {
-      logger.warn('gomod manager does not support replacement updates yet');
-      return null;
+    let newName = currentName;
+    // newName will be available for replacement
+    if (upgrade.newName) {
+      newName = upgrade.newName;
     }
```
rarkins, line 26:
```suggestion
    const newName = upgrade.newName ?? currentName;
```

lib/modules/manager/gomod/update.ts
```diff
@@ -17,14 +17,15 @@ export function updateDependency({
...
     // istanbul ignore if: should never happen
-    if (!currentName || !upgrade.managerData) {
+    if (!currentName || !newName || !upgrade.managerData) {
```
rarkins, line 28:
```suggestion
    if (!currentName || !upgrade.managerData) {
```

Impossible to happen

lib/modules/manager/gomod/update.ts
rarkins: Also consider moving this down closer to where it's used

rarkins (changes requested):

jamietanna:
Hmm I need to retry https://github.com/jamietanna/renovate-iss-oapi/pull/2 and check it also works when we update the module import path

jamietanna:
OK, so retrying with an update to my test repo, I see https://github.com/jamietanna/renovate-iss-oapi/pull/5 raised which does _not_ take the `postUpdateOptions` into account :thinking:

jamietanna:
Adding support for that looks like it could be a little more involved - separate PR?

jamietanna:
(This isn't _yet_ working - I can't seem to trigger the `postUpdateOptions` for https://github.com/jamietanna/renovate-iss-oapi

viceice (dismissed):

jamietanna:
Any thoughts on why we may not be able to see `postUpdateOptions` working? 🤔

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -31,7 +31,7 @@ import { getExtraDepsNotice } from './artifacts-extra';
 const { major, valid } = semver;
 
 function getUpdateImportPathCmds(
-  updatedDeps: PackageDependency[],
+  updatedDeps: Upgrade[],
```
jamietanna, line 34: this isn't quite correct, and why things aren't working

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -31,7 +31,7 @@ import { getExtraDepsNotice } from './artifacts-extra';
...
   { constraints }: UpdateArtifactsConfig,
 ): string[] {
   // Check if we fail to parse any major versions and log that they're skipped
```
jamietanna, line 37: For some reason, when we're seeing a replacement, we get the wrong data here :thinking:

jamietanna: As:

```
 WARN: Ignoring dependency: Could not get major version (repository=jamietanna/renovate-iss-oapi, branch=renovate/github.com-deepmap-oapi-codegen-replacement)
       "depName": "github.com/deepmap/oapi-codegen",
       "newVersion": undefined,
       "newValue": "v2.3.0",
       "updateType": "replacement"
```

Will modify this for handling a `replacement`

jamietanna:
https://github.com/renovatebot/renovate/compare/main...xNok:renovate:replacement-for-gomod and https://github.com/marwan-at-work/mod/pull/26 (cc @xNok) are relevant here, as it looks like `mod` can't be used as-is :thinking:

jamietanna:
S'all good, will follow up with this in the future when I get a chance

PR #29586, opened 2024-06-10
fix(vulnerabilities): do not force exact patch version for PyPI datasource in GitHub alerts

Churro:
## Changes

GitHub vulnerability alerts may suggest a particular patch version for PyPI dependencies that could be retracted, currently resulting in no PR being created at all. This change relaxes `==` to `>=`, so that the first available version is picked.

## Context

- Fixes https://github.com/renovatebot/renovate/discussions/29572

Test repo: https://github.com/renovate-demo/29572-renovate-python-requests/pull/2

### DIFF
--- a/lib/workers/repository/init/vulnerability.ts
+++ b/lib/workers/repository/init/vulnerability.ts
@@ -207,7 +207,7 @@ export async function detectVulnerabilityAlerts(
           // TODO: types (#22198)
           const allowedVersions =
             datasource === PypiDatasource.id
-              ? `==${val.firstPatchedVersion!}`
+              ? `>=${val.firstPatchedVersion!}`
               : val.firstPatchedVersion;
           const matchFileNames =
             datasource === GoDatasource.id
--- a/lib/workers/repository/init/__snapshots__/vulnerability.spec.ts.snap
+++ b/lib/workers/repository/init/__snapshots__/vulnerability.spec.ts.snap
@@ -108,7 +108,7 @@ An issue was discovered in FasterXML jackson-databind prior to 2.7.9.4, 2.8.11.2
 exports[`workers/repository/init/vulnerability detectVulnerabilityAlerts() returns pip alerts 1`] = `
 [
   {
-    "allowedVersions": "==2.2.1.0",
+    "allowedVersions": ">=2.2.1.0",
     "force": {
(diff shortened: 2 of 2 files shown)


lib/workers/repository/init/vulnerability.ts
```diff
@@ -207,7 +207,7 @@ export async function detectVulnerabilityAlerts(
           // TODO: types (#22198)
           const allowedVersions =
             datasource === PypiDatasource.id
-              ? `==${val.firstPatchedVersion!}`
+              ? `>=${val.firstPatchedVersion!}`
```
rarkins, line 210: What about the line below it too?

Churro: Hmm, then overall this should work:

```suggestion
          const allowedVersions =
            datasource === MavenDatasource.id
              ? `[${val.firstPatchedVersion!},]`
              : `>=${val.firstPatchedVersion!}`;
```

I'm goint to run it on a larger test repo to verify, just in case.

Churro: **Update**: Looks good -> https://github.com/renovate-demo/29586-vulndemo/pulls

The only issue is that this breaks `nuget` security updates since https://github.com/renovatebot/renovate/pull/26150 is missing. Hence, `allowedVersions` for nuget would need to stay at `val.firstPatchedVersion!` as an interim solution.

I'd prefer doing the change to all managers in a separate PR. This PR related to PyPI could be merged for now and would do no harm but if you prefer only one change, feel free to close here.

viceice (approved):

rarkins (commented):
What about the change to the worker so that it picks the minimal instead of maximum matching version?

Churro:
> What about the change to the worker so that it picks the minimal instead of maximum matching version?

Should be warranted for GitHub alerts with this snippet:
https://github.com/renovatebot/renovate/blob/e6b04dab1b48097c96f89661ef03a1d1bef3f0b4/lib/workers/repository/process/lookup/index.ts#L364-L366

For OSV, this is a thing but I wanted to address this in a separate PR - see https://github.com/renovatebot/renovate/discussions/29280#discussioncomment-9641300

rarkins:
Yes you're right, I forgot that this is for non-OSV

rarkins:
Follow-up issue: #29600

Issue #6508, opened 2020-06-13
Julia support


waghanza:
Hi,

It will be awesome if `renovatebot`, could parse `Project.toml` in **Julia** Project.

Dependencies in **julia** could be specified in various ways :

+ `^1.2.3 `  => [1.2.3, 2.0.0)
+ `^1.2`     => [1.2.0, 2.0.0)
+ `^1`       => [1.0.0, 2.0.0)
+ `^0.2.3`   => [0.2.3, 0.3.0)
+ `^0.0.3`   => [0.0.3, 0.0.4)
+ `^0.0`     => [0.0.0, 0.1.0)
+ `^0`       => [0.0.0, 1.0.0)
+ `~1.2.3`   => [1.2.3, 1.3.0)
+ `~1.2`     => [1.2.0, 1.3.0)
+ `~1`       => [1.0.0, 2.0.0)
+ `~0.2.3`   => [0.2.3, 0.3.0)
+ `~0.0.3`   => [0.0.3, 0.0.4)
+ `~0.0`     => [0.0.0, 0.1.0)
+ ~0`       => [0.0.0, 1.0.0)
+ `>= 1.2.3` => [1.2.3,  ∞)
+ `≥ 1.2.3`  => [1.2.3,  ∞)
+ `= 1.2.3`  => [1.2.3, 1.2.3]
+ `< 1.2.3`  => [0.0.0, 1.2.2]

@codeneomatrix Do I forget anything according `julia` and **deps** management

@rarkins You can find a example in https://github.com/waghanza/http-benchmark/blob/julia/deps/julia/merly/Project.toml

Regards,

rarkins:
First thing is: do we need a new `datasource`? Seems like we do, is it https://github.com/JuliaRegistries/General? Example versions list for a random package: https://raw.githubusercontent.com/JuliaRegistries/General/master/G/GaussianProcesses/Versions.toml

After `datasource` we need to verify `versioning`. I checked the full list of ranges above and npm's `semver` implementation actually follows the same results, so looks like we can reuse it. Maybe we alias it in case one day it needs to diverge.

Finally there is the `manager`. We'd need to extract dependencies from `Project.toml` files, which should be pretty easy. I don't see any lock file in Julia, so that should be all we need.

codeneomatrix:
Hi.

Starting with Julia version 1.4, it is also possible to use hyphens in the version specification of dependencies. 

So it is recomended to also add 
```
[compat]
julia = "1.4"
```
[...]

PR #29623, opened 2024-06-12
feat(datasource/julia-pkg-server): add Julia `PkgServer`

bauglir:
## Changes

[`PkgServer`s](https://github.com/JuliaPackaging/PkgServer.jl/) are the primary package distribution system for Julia (the alternative being operating on the underlying Git repositories directly, which I might contribute later but seemed like a more daunting task given the number of possible hosting platforms). This datasource enables retrieval of packages from registries through these `PkgServer`s. It supports multiple registries across `PkgServer`s (a single server can host multiple registries and the same registry can be hosted across different `PkgServer`s).

Julia-specific versioning is not yet implemented, I will implement that as a follow-up.

## Context

- #6508.

## Documentation

## How I've tested my work


lib/modules/datasource/julia-pkg-server/__fixtures__/ExtraHTTP.tar
bauglir, line 1: These are binary files, but given the `PkgServer` behavior there is (unfortunately) a need for these and I did not see an easier way of including these apart from possibly creating them in-memory. I intentionally tried to keep these as small as possible.

lib/modules/datasource/julia-pkg-server/registry.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  // Block until the content of all files has been (asynchronously) read from
+  // the tarball
+  return await new Promise<FileContents<string> | null>((resolve, reject) => {
+    // File content may be streamed in chunks which need to be aggregated
+    // TODO: The tests for this functionality only require a single "chunk" to
+    // be read from the tarball to succeed. Hence, it would be sufficient to
+    // just process a single chunk directly as a string and return it. This
+    // does not hold in general and a (minimal) test requiring multiple chunks
+    // should be implemented
```
bauglir, line 77: I have a tarball where this fails (which was obtained from the official `PkgServer), but that tarball is over 7MB large and I did not want to include that as a fixture. I have tried to create a more minimal example where the simpler approach fails, but have so far been unable to do so.

package.json
```diff
@@ -239,6 +239,7 @@
     "simple-git": "3.24.0",
     "slugify": "1.6.6",
     "source-map-support": "0.5.21",
+    "tar": "6.2.1",
```
bauglir, line 242: I have moved the `tar` dependency from `devDependencies` to `dependencies` as `PkgServer`s serve tarballs that need to be extracted. I hope this is okay. (Re)using this dependency seemed the most straight-forward as it was already included as a development dependency and therefore likely has gone through some scrutiny.

lib/modules/datasource/julia-pkg-server/registry.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    tarballParser.on('entry', (entry) => {
+      entry.on('data', (data) => {
+        fileContentChunks[entry.path].push(data);
+      });
+
+      entry.on('error', reject);
+    });
+
+    tarballParser.end(tarball);
```
bauglir, line 109: This extracts the necessary files in memory. It might be more straightforward to unpack do disk instead and work with some form of caching. I'd be happy to change it if that's better suited to Renovate's overall design.

HonkingGoose (changes requested):
Most of my changes are for the docs or code comments.

In general the changes are:

- One sentence per line for docs files.
- Simpler words and phrases.
- Shorter sentences.
- Fix some typos.

This matches the [Renovate Style Guide](https://github.com/renovatebot/renovate/blob/main/docs/development/style-guide.md).

I'll let the maintainers do a technical review of your code and tests. :wink:

lib/modules/datasource/julia-pkg-server/registry.ts
```diff
@@ -0,0 +1,246 @@
...
+/**
+ * URLs referencing a registry served by a PkgServer take the form:
+ * `<PkgServer host>/registry/<registry UUID>/<registry state>`.
+ *
+ * Although required when querying the state of a registry, the state in the
+ * URL is not mandatory when configuring a `registryUrl` passed into the
+ * datasource. Therefore that section is marked optional and should be ensured
+ * to be filled before further processing.
+ */
+const PKG_SERVER_REGISTRY_URL_FORMAT = new RegExp(
```
rarkins, line 37:
```suggestion
const PKG_SERVER_REGISTRY_URL_FORMAT = regEx(
```

Use regEx, which is re2 under the hood

lib/modules/datasource/julia-pkg-server/registry.ts
rarkins: How often do we need to fetch tarballs, if they can be large? And for how long can we cache them?

lib/modules/datasource/julia-pkg-server/registry.ts
```diff
@@ -0,0 +1,246 @@
...
+
+      entry.on('error', reject);
+    });
+
+    tarballParser.end(tarball);
+  });
+}
+
+/**
+ * Parses the provided `registryURL` into an object containing all the
```
rarkins, line 114:
```suggestion
 * Parses the provided `registryUrl` into an object containing all the
```

lib/modules/datasource/julia-pkg-server/index.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  override readonly customRegistrySupport = true;
+
+  // Multiple package servers can contain the same package, but contain
+  // different versions of that package
+  override readonly registryStrategy = 'merge';
+
+  @cache({
+    namespace: `datasource-${juliaPkgServerDatasourceId}-releases`,
+    key: ({ packageName }: GetReleasesConfig) => packageName,
```
rarkins, line 45: Shouldn't caching be unique per registry url?

rarkins (changes requested):

lib/modules/datasource/julia-pkg-server/registry.ts
bauglir: That is a very good question and one I'm not entirely sure about myself.

The tarballs depend on the state of the registry, so they change whenever a new package is published to a registry. This is done by issuing a PR to the repository hosting the registry. So a suitable cache duration highly depends on the expected rate of package releases, when the corresponding PRs are merged, and how soon new releases should be available to Renovate. Given that these variables will very much depend on the registry from which dependencies are pulled, I'm not entirely sure there will be a uniformly suitable answer.

That being said, the `General` registry (which is where all public, open-source packages tend to be published) sees a couple of releases per hour and "automatically merges" suitable PRs every 8 minutes. So at a minimum a cache duration of 8 minutes can be used, although in practice it is probably fine to cache these up to an hour, maybe even a bit longer.

lib/modules/datasource/julia-pkg-server/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  })
+  async downloadRegistry(registryUrl: string): Promise<Buffer> {
+    const { val, err } = await Result.wrap(
+      this.http.getBuffer(registryUrl, { headers: PKG_SERVER_REQUEST_HEADERS }),
+    )
+      .transform(({ body }) => body)
+      .onError((error) =>
+        logger.warn(
+          { datasource: JuliaPkgServerDatasource.id, error, registryUrl },
+          'An error occurred fetching the registry',
```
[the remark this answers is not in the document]
bauglir: These URLs hit (public) Cloudflare buckets directly (in the case of the default `PkgServer` at least) and produce regular HTTP status codes which are included through the `error` in the first argument to `warn`. There are no Julia `PkgServer` specific error codes to display.

That being said, it turns out that including the full error may not be the best idea after all as the response contains HTML as the body which gets printed byte-by-byte in the logs.

Perhaps just logging the "message" is more appropriate/sufficient? That would result in messages in the logs like

```
 WARN: An error occurred fetching the registry (repository=bauglir/renovate-test)
       "datasource": "julia-pkg-server",
       "error": "Response code 404 (Not Found)",
       "registryUrl": "https://pkg.julialang.org/registry/23338594-aafe-5451-b93e-139f81909106/ea7560e220b6569307594b617dd2869b0804a4df"
```

I also just noticed more or less the same information gets logged to the `DEBUG` stream, so perhaps it's better to forego this warning altogether?

```
DEBUG: Datasource 404 (repository=bauglir/renovate-test)
       "datasource": "julia-pkg-server",
       "packageName": "HTTP",
       "url": "https://pkg.julialang.org/registry/23338594-aafe-5451-b93e-139f81909106/ea7560e220b6569307594b617dd2869b0804a4dc"
```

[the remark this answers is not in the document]
HonkingGoose: I'll let the maintainer decide!

lib/modules/datasource/julia-pkg-server/registry.ts
rarkins: If it's a single tarball per registry and not per package then I think something like a 15-30 minute cache time is fine. If it was common for the tarball to be unchanged for longer periods of time then we'd want to retain and use etag or last-modified headers for subsequent requests, but if registries typically get updated frequently then there's little gain from that

lib/modules/datasource/julia-pkg-server/registry.ts
bauglir: I had noticed that existed while starting to work on the versioning 😅

lib/modules/datasource/julia-pkg-server/readme.md
```diff
@@ -0,0 +1,15 @@
...
+
+Custom registries are supported by specifying the full URL of the registry,
+i.e. `https://<pkg-server>/registry/<uuid>`. Optionally a "state", which
+corresponds to a Git tree SHA from the regitries' repository, may be appended
+to the URL to retrieve releases for a package at a particular moment in time.
+If the state is not provided it will be automatically determined by querying
+the `PkgServer`.
+
+When the same package is registered through multiple registries, the
+corresponding releases will be merged.
```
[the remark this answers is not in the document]
bauglir: Included in d1b01307e555f77157adc8bc898a3a5b8d28a971.

lib/modules/datasource/julia-pkg-server/readme.md
```diff
@@ -0,0 +1,15 @@
...
+[the General registry](https://github.com/JuliaRegistries/General/) hosted
+through [https://pkg.julialang.org](https://pkg.julialang.org) (which is the
+default registry used by this datasource).
+
+Custom registries are supported by specifying the full URL of the registry,
+i.e. `https://<pkg-server>/registry/<uuid>`. Optionally a "state", which
+corresponds to a Git tree SHA from the regitries' repository, may be appended
+to the URL to retrieve releases for a package at a particular moment in time.
+If the state is not provided it will be automatically determined by querying
+the `PkgServer`.
```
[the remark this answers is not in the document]
bauglir: Included in d1b01307e555f77157adc8bc898a3a5b8d28a971.

lib/modules/datasource/julia-pkg-server/readme.md
```diff
@@ -0,0 +1,15 @@
+This datasource returns releases for Julia packages from registries hosted
+through [`PkgServer`](https://github.com/JuliaPackaging/PkgServer.jl/)s, e.g.
+[the General registry](https://github.com/JuliaRegistries/General/) hosted
+through [https://pkg.julialang.org](https://pkg.julialang.org) (which is the
+default registry used by this datasource).
```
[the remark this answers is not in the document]
bauglir: Included in d1b01307e555f77157adc8bc898a3a5b8d28a971.

lib/modules/datasource/julia-pkg-server/index.ts
bauglir: Resolved in 77d0ec6e99c38afd9f5cb551bb6471c88496c46b.

lib/modules/datasource/julia-pkg-server/__fixtures__/ExtraHTTP.tar
viceice: we've a node-tar dependency, can we use that to generate this on the fly?

how big are those tar files? can't see it on mobile.

lib/modules/datasource/julia-pkg-server/index.ts
```diff
@@ -0,0 +1,113 @@
...
+  registryPathForPackage,
+} from './registry';
+import { Package, Versions } from './schema';
+
+/** The files in a registry defining the metadata for a package. */
+type PackageMetadataPath = `${string}/Package.toml` | `${string}/Versions.toml`;
+
+export class JuliaPkgServerDatasource extends Datasource {
+  // This URL will redirect to a suitable, geographically near, mirror
+  static readonly generalRegistryUrl =
```
viceice, line 24: I think it should be named `defaultRegistryUrl`, that's the terminology on other datasources

lib/modules/datasource/julia-pkg-server/__fixtures__/ExtraHTTP.tar
bauglir: This uses the `node-tar` dependency in the actual implementation, but not (yet) for tests. The files are less than 4kB small, so they end up being 4kB. The tests just need a few very small files.

I had considered generating them on the fly, but I figured the trade-off of the more complex test code and runtime cost outweighed adding the tarballs  to the repository directly.

That being said, I'm fine with either way. The upside of generating them on-demand would be that the contents of the test files would be easier to inspect.

lib/modules/datasource/julia-pkg-server/registry.ts
viceice: it depends on how big these tarballs can be. ~10 MB should be ok, but bigger files can cause memory issues (oom)

lib/modules/datasource/julia-pkg-server/registry.ts
```diff
@@ -0,0 +1,246 @@
...
+      headers: PKG_SERVER_REQUEST_HEADERS,
+    }),
+  )
+    // The response contains the paths to all registries the package server
+    // knows formatted as /registry/<registry uuid>/<registry state sha> on
+    // separate lines, including a final newline
+    .transform(({ body }) => {
+      const registryPaths = body.split('\n').slice(0, -1);
+
+      const registryStates: { [key: string]: string } = {};
```
viceice, line 209: use `Record<string,string>`

lib/modules/datasource/julia-pkg-server/registry.ts
```diff
@@ -0,0 +1,246 @@
...
+    const fileContentChunks: FileContents<Buffer[]> = {};
+    for (const path of pathsToExtract) {
+      fileContentChunks[path] = [];
+    }
+
+    const tarballParser = new tarParse({
+      filter: (path) => pathsToExtract.includes(path),
+    });
+
+    tarballParser.on('end', () => {
```
viceice, line 86: doesn't it support some kind of async functions? maybe we need v7 of tar. 🤔

viceice (commented):
is there no way to only download the required files instead of the full tarballs? 🤔

bauglir:
> is there no way to only download the required files instead of the full tarballs? 🤔

Unfortunately not 😞 At least not when relying on package servers. That would require direct interaction with the underlying Git repositories. This is an alternative datasource that I was planning on creating as well, but that supports different use cases (e.g. package servers enable access to registries to which one does not have access to the underlying repository of a registry) and this one seemed more straightforward to implement.

I'm open to starting a discussion around whether package servers could start to provide direct access to these files, but that will likely be a long discussion and might not be possible given the design of the package server infrastructure.

lib/modules/datasource/julia-pkg-server/__fixtures__/ExtraHTTP.tar
bauglir: This ended up being not too cumbersome to implement in 813fab435d386e0cf9dd562dee0a1394c4a7b063. The one thing I wasn't sure about was how I could best add the test helper as ESLint either complains about not allowing exports from spec files or about a dependency being used exclusively for tests not being a proper dependency.

viceice:
> > is there no way to only download the required files instead of the full tarballs? 🤔
> 
> Unfortunately not 😞 At least not when relying on package servers. That would require direct interaction with the underlying Git repositories. This is an alternative datasource that I was planning on creating as well, but that supports different use cases (e.g. package servers enable access to registries to which one does not have access to the underlying repository of a registry) and this one seemed more straightforward to implement.
> 
> I'm open to starting a discussion around whether package servers could start to provide direct access to these files, but that will likely be a long discussion and might not be possible given the design of the package server infrastructure.

we should start that discussion on the package server repo, so they hopefully consider it. it would be a huge win for renovate.

lib/modules/datasource/julia-pkg-server/registry.ts
```diff
@@ -0,0 +1,246 @@
...
+ * particular moment in time.
+ *
+ * If the `registryUrl` does not specify the state of the registry, the state
+ * will be determined by querying the associated PkgServer.
+ */
+export async function parseRegistryUrl(
+  http: Http,
+  registryUrl?: string,
+): Promise<JuliaPkgServerRegistry | null> {
+  const parsedRegistryUrl = registryUrl?.match(PKG_SERVER_REGISTRY_URL_FORMAT);
```
viceice, line 125:
```suggestion
  const parsedRegistryUrl = registryUrl?.match(PKG_SERVER_REGISTRY_URL_FORMAT)?.groups;
```

lib/modules/datasource/julia-pkg-server/registry.ts
```diff
@@ -0,0 +1,246 @@
...
+      },
+      'An invalid registry URL was specified',
+    );
+
+    return null;
+  }
+
+  // The groups will always be specified, as otherwise the URL wouldn't have
+  // parsed and the function returned above
+  const { pkgServer, state, uuid } = parsedRegistryUrl.groups!;
```
viceice, line 141:
```suggestion
  const { pkgServer, state, uuid } = parsedRegistryUrl;
```
simplify 😉

lib/modules/datasource/julia-pkg-server/index.ts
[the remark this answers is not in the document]
viceice: I would prefer to remove this logging in favor of the general logging of errors in base class or upper datasource layer

lib/modules/datasource/julia-pkg-server/registry.ts
viceice: @rarkins we need to think about those big blobs. we cache every request in memory, so we've this blob twice.

https://github.com/renovatebot/renovate/blob/26337ac124f4dd1885a44ed65822f2944ad69f3d/lib/util/http/index.ts#L186

we probably should pass `memCache:false` for those files and only cache the smaller transformed data to package cache.

@bauglir I think we should extract all deps from that file and add them to the package cache, so we can speed up additional lookups. we do something similar at `rubygems` datasource.

viceice (changes requested):

lib/modules/datasource/julia-pkg-server/registry.ts
```diff
@@ -0,0 +1,246 @@
...
+const PKG_SERVER_REGISTRY_URL_FORMAT = new RegExp(
+  '^(?<pkgServer>.+)/registry/(?<uuid>\\p{Hex_Digit}{8}(-\\p{Hex_Digit}{4}){3}-\\p{Hex_Digit}{12})(/(?<state>\\p{Hex_Digit}{40}))?/?$',
+  'u',
+);
+
+/**
+ * The final part of the path of a (fully specified) registry URL references
+ * the tree SHA of the state of of the registry repository the package server
+ * is serving. Given the content-addressable nature of this SHA, it can be used
+ * as a caching mechanism.
```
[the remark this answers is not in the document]
HonkingGoose: ## Check input matches expectations

> I think this raises a good question as to whether we should be enforcing strict formatting on the state.

It's probably best to to check what Renovate "pulls in". So in general terms:

- Check input is valid
- Check input matches expected format
- Sanitize dangerous stuff
- Only then proceed with processing the input

## SHA state check as general utility/function to be used by other Renovate code?

Checking if the state is SHA-1 vs SHA-256 vs SHA-512 sounds like something that other Renovate code could re-use? Maybe that should be a new file/function in a suitable place?

## To check or not to check?

### I vote check

> Either we should probably check it everywhere, or we can forego the check altogether

I would vote check everywhere, but I'm not a maintainer. :smile:

### If we decide to not check, show good error message

> ... or we can forego the check altogether which will then just result in 404 errors when retrieving the registry ...

I guess that would look like this then?

1. Renovate tries to create PR for SHA-256 registry package.
2. Renovate shows generic 404 error.
3. User thinks Renovate is broken.

I would prefer something like this:

1. Renovate tries to create PR for SHA-256 registry package.
2. Renovate shows 404 error, and explains that the SHA-256 bit is not supported yet.
3. This probably prompts the user to ask the Renovate maintainers to add code to support the new hash.

## Your upside is my downside

> ... with the upside that it would not require changes to Renovate if the package server starts supporting different state formats.

I don't agree with this. If the package server supports a new format, we probably need to account for that in the Renovate code. So it's best if we state explicitly what's supported, what's not, and have a way to get nudged when things break.

## Maintainers make the decision here

I'll let the maintainers decide what to do with all this info above, and with your idea! :wink:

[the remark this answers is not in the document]
viceice: can we simply allow sha256? so we're future prove when the package server becomes support for it? I think in near future more and more new repos will use sha256 by default.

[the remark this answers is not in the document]
bauglir: Thanks for the extensive input.

The state verification currently only applies to user-provided input. Given that a package server will directly report what URLs to use when asking it for the state of a registry, the URLs/states it provides are assumed to be correct (note that the package server hosts these URLs itself). If the package server turns out to provide wrong URLs that will indeed result in 404s.

Currently, no validation is done on those returned URLs (but I do have a note locally on whether this is something we'd want to do which I think is what you are suggesting). So I'm open to changing that.

> can we simply allow sha256?

I have added this in 1ffa7aee56078a9c57d0df50ddfa9af62c2a1c90.

bauglir:
> we should start that discussion on the package server repo, so they hopefully consider it. it would be a huge win for renovate.

I have started the discussion in JuliaPackaging/PkgServer.jl#210 and will be happy to follow up on simplifying things once/if that gets resolved.

rarkins:
Seems like no progress in the upstream issue

rarkins:
@bauglir @viceice are we stuck on this PR if the upstream issue is unresolved?

viceice:
> @bauglir @viceice are we stuck on this PR if the upstream issue is unresolved?

i think no, but we need to think about good caching to not load the big packages too often.

bauglir:
> are we stuck on this PR if the upstream issue is unresolved

I don't think we are. I also suspect it to be likely that even if the ask upstream would be resolved, that it'd still be a while.

This has been primarily stuck on me. I got stuck a bit and haven't had the bandwidth over the last month or two to get unstuck. But I was planning on getting back to this over the next week or two. I'm sorry about the delay.

> but we need to think about good caching to not load the big packages too often.

I have already done some experiments here and I'm relatively sure I can make this work. Some rudimental tests seem to suggest a much smaller memory/storage footprint implementing a cache for just the absolutely necessary information. Basically I'm just doing the extraction once and manually caching just the package version information instead of the tarballs and network requests.

I did run into one situation I'll probably need some help with, but I can follow up with that on Slack. The gist of the issue is that the current way I have implemented the cache works properly for subsequent runs of Renovate, but if multiple packages need to be updated the full cache update mechanism gets triggered for every package.

PR #29644, opened 2024-06-13
build(deps): Revert "build(deps): update dependency re2 to v1.21.1"

rarkins:
Reverts renovatebot/renovate#29634

1.21.1 still had memory problems
1.21.2 fails our tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -252,7 +252,7 @@
   "optionalDependencies": {
     "better-sqlite3": "9.6.0",
     "openpgp": "5.11.1",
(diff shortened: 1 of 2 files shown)


viceice (commented):
are the memory issues so big? is my user complaining?

rarkins:
This has not been published long enough for any users to use it and complain. Needs reverting immediately before that happens

viceice:
https://github.com/uhop/node-re2/issues/212

viceice (approved):

PR #29664, opened 2024-06-13
fix(datasource/galaxy): support multiple results

rarkins:


lib/modules/datasource/galaxy/index.ts
```diff
@@ -50,13 +50,19 @@ export class GalaxyDatasource extends Datasource {
...
     }
 
-    // istanbul ignore if
     if (body.results.length > 1) {
-      logger.warn(
-        { dependency: packageName },
-        `Received multiple results from ${galaxyAPIUrl}`,
-      );
-      return null;
+      const githubUsernames = body.results.map((result) => result.github_user);
```
viceice, line 54: I would filter here and warn if the filtered result is empty. that way we only have to iterate once.

lib/modules/datasource/galaxy/index.ts
```diff
@@ -50,13 +50,19 @@ export class GalaxyDatasource extends Datasource {
       throw this.handleGenericErrors(err);
     }
```
secustor, line 52:
```suggestion
     body.results = body.results.filter((result) => result.github_user === username)
```

lib/modules/datasource/galaxy/index.ts
```diff
@@ -50,13 +50,19 @@ export class GalaxyDatasource extends Datasource {
...
+        body.results = body.results.filter(
+          (result) => result.github_user === userName,
+        );
+      } else {
+        logger.warn(
+          { dependency: packageName, githubUsernames },
+          `Multiple results from galaxy for ${packageName}, none match`,
+        );
+        return null;
+      }
```
secustor, line 65:
```suggestion
      logger.warn(
        { dependency: packageName },
        `Received multiple results from ${galaxyAPIUrl}`,
      );
      return null;
```

secustor (dismissed):

lib/modules/datasource/galaxy/index.ts
```diff
@@ -49,8 +49,7 @@ export class GalaxyDatasource extends Datasource {
     } catch (err) {
       this.handleGenericErrors(err);
     }
-
-    // istanbul ignore if
+     body.results = body.results.filter((result) => result.github_user === username)
```
rarkins, line 52:
```suggestion
    body.results = body.results.filter((result) => result.github_user === username)
```

viceice (approved):

Issue #29600, opened 2024-06-11
GitHub vulnerability alerts: do not limit to single version


rarkins:
### Describe the proposed change(s).

Following on from #29586, we should support "greater than or equal" for non-Python ecosystems too. i.e. pick the lowest which exists.

For ecosystems which support Semver, it can be expressed as `>=`. For Maven it would be `[version,)`. It's perhaps not supported for Nuget yet.

PR #27363, opened 2024-02-16
fix(vulnerabilities): describe last_affected version as ecosystem-specific version constraint

Churro:
## Changes

Ensures that Maven version ranges are used in `allowedVersions` constraints when OSV vulnerabilities specify an ecosystem-specific `last_affected` version. This prevents an `Aggregate error` exception that was thrown when advisories specify non-semver compatible versions, such as `1.2.1.2-jre17` (see discussion).

Of the OSV-supported ecosystems, this change is only required for `maven` and `nuget`. Support for `nuget` is missing for now as it depends on support for nuget version ranges being available.

## Context

- https://github.com/renovatebot/renovate/discussions/27221
- https://github.com/renovatebot/renovate/pull/26150

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

Test repo: https://github.com/renovate-demo/renovate-osv-ecosystem-lastaffected/pull/1

PR #29666, opened 2024-06-13
fix(vulnerabilities): do not force exact patch version in OSV alerts

### LINKED ISSUE #29342: Verify OSV alert logic (and maybe make code changes as needed)

Check if the OSV alert logic is correct, and behaving like we want. `@churro` wanted to check this for us, so here's the issue to track that work.

#### Relevant discussion

Quotes come from:

- #29280

`@rarkins` said:

> Renovate's other vulnerability handling (GitHub alerts, or Mend's SCA products) use `>=` for this reason plus this logic: https://github.com/renovatebot/renovate/blob/9e2ca6b152a81b3e7e19a06fe5e54c7823844a94/lib/workers/repository/process/lookup/index.ts#L364-L366
> 
> In other words it becomes "take the first release which exists greater than or equal to x.y.z". Usually it's x.y.z but sometimes it can be higher due to reasons like this.
> 
> I wasn't involved in the OSV implementation although you can see from the code I linked to that `@Churro` was aware of it when implementing

Response from `@Churro`

> I don't remember why back then I thought it's necessary to treat OSV alerts differently but I can verify this over the next few days.

> [!TIP]
> You may want to check out [our fork of `@astellingwerf`'s reproduction for Discussion 29280 and Issue 29342](https://github.com/renovate-reproductions/29280-and-29342) to confirm if the OSV logic is working as intended.

Churro:
## Changes

OSV vulnerability alerts may suggest a particular patch version that could be retracted, currently resulting in no PR being created at all. This change relaxes `==` to `>=`, so that the first available version is picked.

This PR also removes an extra condition for OSV that caused advisories with `last_affected` field set to suggest the newest version, rather than the minimal patched version. Advisories usually populate the `fixed` version, not the `last_affected` field, so this was infrequently used. Now this the change to `>=` e.g., for pypi, the security fix PR would suggest version 2.32.3 to address CVE-2024-35195 in `requests <= 2.32.0`, whereas the minimal patched version is 2.32.2 (see test repo).

## Context

- Closes https://github.com/renovatebot/renovate/issues/29342
- Fixes https://github.com/renovatebot/renovate/discussions/29280
- https://github.com/renovatebot/renovate/issues/29600
- https://github.com/renovatebot/renovate/discussions/29572
- https://github.com/renovatebot/renovate/pull/27363#pullrequestreview-1885643289

* Test repo with requests: https://github.com/renovate-demo/29280-osv-requests/pull/1
* Test repo with different managers: https://github.com/renovate-demo/29666-osv/pulls

### DIFF
--- a/lib/workers/repository/process/vulnerabilities.ts
+++ b/lib/workers/repository/process/vulnerabilities.ts
@@ -409,7 +409,7 @@ export class Vulnerabilities {
       this.isVersionGt(version, depVersion, versioningApi),
     );
     if (fixedVersion) {
-      return ecosystem === 'PyPI' ? `==${fixedVersion}` : fixedVersion;
+      return ecosystem === 'PyPI' ? `>=${fixedVersion}` : fixedVersion;
     }
 
     lastAffectedVersions.sort((a, b) => versioningApi.sortVersions(a, b));
--- a/lib/workers/repository/process/vulnerabilities.ts
+++ b/lib/workers/repository/process/vulnerabilities.ts
@@ -409,7 +409,7 @@ export class Vulnerabilities {
       this.isVersionGt(version, depVersion, versioningApi),
     );
     if (fixedVersion) {
-      return ecosystem === 'PyPI' ? `>=${fixedVersion}` : fixedVersion;
+      return this.getFixedVersionByEcosystem(fixedVersion, ecosystem);
     }
 
     lastAffectedVersions.sort((a, b) => versioningApi.sortVersions(a, b));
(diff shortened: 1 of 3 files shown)


lib/workers/repository/process/vulnerabilities.ts
```diff
@@ -409,7 +409,7 @@ export class Vulnerabilities {
       this.isVersionGt(version, depVersion, versioningApi),
     );
     if (fixedVersion) {
-      return ecosystem === 'PyPI' ? `==${fixedVersion}` : fixedVersion;
+      return ecosystem === 'PyPI' ? `>=${fixedVersion}` : fixedVersion;
```
rarkins, line 412: Sure we can't fix all ecosystems at once?

viceice (dismissed):

lib/workers/repository/process/vulnerabilities.ts
Churro: Sure, we can 👍

Churro:
Update: As requested, I've now extended the change to cover not only pypi but all managers.

New test repo: https://github.com/renovate-demo/29666-osv/pulls

rarkins (approved):

viceice (approved):

PR #29707, opened 2024-06-16
fix(manager/gitlabci): support alias with a path in component URL

marcovmun:
## Changes

Support aliases that include a path, because replacement was happening too late in the function.

## Context

Components with path like this were not supported:
```
include:
  - component: $COMPONENT_REGISTRY/a-project/a-component@1.0
```


lib/modules/manager/gitlabci/extract.ts
```diff
@@ -31,6 +31,7 @@ import { getGitlabDep, replaceReferenceTags } from './utils';
 const componentReferenceRegex = regEx(
   /(?<fqdn>[^/]+)\/(?<projectPath>.+)\/(?:.+)@(?<specificVersion>.+)/,
 );
+const componentFQDNRegex = regEx(/[^/]+/);
```
viceice, line 34: I don't understand why we need this simple regex? what do you check with that? isn't a simple string check enough?

lib/modules/manager/gitlabci/common.ts
```diff
@@ -37,5 +37,8 @@ export function isGitlabIncludeLocal(
 export function isGitlabIncludeComponent(
   include: GitlabInclude,
 ): include is GitlabIncludeComponent {
-  return !is.undefined((include as GitlabIncludeComponent).component);
+  return (
+    !is.undefined((include as GitlabIncludeComponent).component) &&
+    typeof (include as any).component === 'string'
+  );
```
rarkins, line 43: Is `is.string(.....)` an alternative here?

lib/modules/manager/gitlabci/extract.ts
```diff
@@ -122,9 +122,16 @@ function extractDepFromIncludeComponent(
   includeComponent: GitlabIncludeComponent,
   registryAliases?: Record<string, string>,
 ): PackageDependency | null {
-  const componentReference = componentReferenceRegex.exec(
-    includeComponent.component,
-  )?.groups;
+  let componentPath = includeComponent.component;
+  const componentFQDN = componentPath?.substring(0, componentPath.indexOf('/'));
```
viceice, line 126:
```suggestion
  const componentFQDN = componentPath.substring(0, componentPath.indexOf('/'));
```
I think component path can't be empty with your changed check above.

lib/modules/manager/gitlabci/common.ts
viceice: no any cast.

lib/modules/manager/gitlabci/extract.ts
```diff
@@ -122,9 +122,16 @@ function extractDepFromIncludeComponent(
   includeComponent: GitlabIncludeComponent,
   registryAliases?: Record<string, string>,
 ): PackageDependency | null {
-  const componentReference = componentReferenceRegex.exec(
-    includeComponent.component,
-  )?.groups;
+  let componentPath = includeComponent.component;
+  const componentFQDN = componentPath.substring(0, componentPath.indexOf('/'));
```
viceice, line 126: why this substring? makes no sense to me. used should be able to alias more than the first part.

```suggestion
  const componentFQDN = componentPath;
```

lib/modules/manager/gitlabci/extract.ts
```diff
@@ -122,9 +122,16 @@ function extractDepFromIncludeComponent(
...
+  if (componentFQDN && registryAliases?.[componentFQDN]) {
```
viceice, line 127:
```suggestion
  if (registryAliases?.[componentFQDN]) {
```

viceice (commented):
looks like we can drop the `componentFQDN` variable

lib/modules/manager/gitlabci/extract.ts
marcovmun: I don't get it, `componentPath` is always combination of a variable and a path to the specifiek component and version. The reason for the aliases to replace the first path of the string with the absolute URL.
Like in the test:
```
- component: $COMPONENT_REGISTRY/a-project/a-component@1.0
```
You want to get `$COMPONENT_REGISTRY` from the componentPath to check if variable can be found in the `registryAliases` array.

If you want to be more dynamic, then we need to loop through all aliases and check if alias can be found in the component path and replace that. But that was not how the old implementation was working, but can be added.

```
// dummy code
for (registryAliases as key => value} {
     componentPath = componentPath.replace(key, value)
}
```

But maybe I don't understand your suggestion.

viceice (approved):

PR #29727, opened 2024-06-17
feat(manager/pip_requirements): allow = as extra-url separator

Finkregh:
See also: https://github.com/renovatebot/renovate/discussions/29726

### manager/pip_requirements

Before:

only `--extra-index-url http...` matched

After:

also `--extra-index-url=http...` matches

## Context

These lines reflect CLI arguments, normal or expected behaviour is that `--foo bar` and `--foo=bar` lead to the same result.

This is not explicitly documented, but inferred from behaviour of many other tools: https://pip.pypa.io/en/stable/reference/requirements-file-format/

Please comment if the docs should be updated, and where.


Finkregh:
The CI seems to be stuck :(

Finkregh:
I don't know how to fix the issues the CI is showing 🙈

rarkins:
Focus on coverage first

lib/modules/manager/pip_requirements/common.ts
```diff
@@ -33,13 +33,23 @@ export function extractPackageFileFlags(
...
-      registryUrls = [line.split(' ')[1]];
-    } else if (line.startsWith('--extra-index-url ')) {
-      const extraUrl = line
-        .substring('--extra-index-url '.length)
-        .split(' ')[0];
-      additionalRegistryUrls.push(extraUrl);
+    if (line.startsWith('-i ') || line.startsWith('--index-url')) {
+      // parameters can be separated by space or =, same below for --extra-index-url
+      if (line.includes('=')) {
+        registryUrls = [line.split('=')[1]];
```
RahulGautamSingh, line 39: Due to split on `=` the flags will be included in the registryUrl too 
ie. for
```
--index-url=https://artifactory.company.com/artifactory/api/pypi/python/simple --trusted-host artifactory.company.com --default-timeout 600
```

registryUrl will be:
`
https://artifactory.company.com/artifactory/api/pypi/python/simple --trusted-host artifactory.company.com --default-timeout 600
`

whereas we want:
`
https://artifactory.company.com/artifactory/api/pypi/python/simple
`

This is also the cause of the failing tests in CI.

Finkregh: Ah, good point. I'll look into it as as find time.

PR #28760, opened 2024-04-30
feat(datasource/custom): Support digest in custom datasource

acortelyou:
## Changes

To support digests a datasource has to implement the getDigest method.
This PR implements the method returning a null promise.
The value itself can be provided in the getReleases method.

PR #29730, opened 2024-06-17
fix(datasource/custom): do not run digest update on version updates

marcoesters:
## Changes

Add an exception for custom data sources to disallow digest updates when other updates are already in place.

## Context

Relevant discussion: #28343

PR #28760 introduced the ability to perform digest updates, i.e., updating the digest without a version change. This had an unintended consequence under the following circumstances: if a version update also updates a digest, `renovate` will additionally try a digest update for the *current* version (not the updated version). If an API returns only the latest version (as shown in the [k3s example](https://docs.renovatebot.com/modules/datasource/custom/#k3s)), this will result in a misleading warning that a digest could not be found.

The [following lines](https://github.com/renovatebot/renovate/blob/main/lib/workers/repository/process/lookup/index.ts#L502-L510) are the culprit:

```typescript
    if (supportsDigests(config.datasource)) {
      if (config.currentDigest) {
        if (!config.digestOneAndOnly || !res.updates.length) {
          // digest update
          res.updates.push({
            updateType: 'digest',
            newValue: config.currentValue,
          });
        }
```

`!res.updates.length` is supposed to prevent such an extra update from happening. However, `config.digestOneAndOnly` is `undefined` for anything but Go and (sometimes) Bazel managers, so `!config.digestOneAndOnly` always evaluates to `true`.

I admit that this is a hack, but I don't have enough overview of the code base to replace `digestOneAndOnly` or find a way to change the condition without guaranteeing any fallout for other managers. This may need to be reevaluated comprehensively in the future.

A test repository can be found here: https://github.com/marcoesters/renovate-miniconda-digest-reproducer

~It contains the original update, but~ I added some simplified json files. Expected behavior would be:

* `major.json`, `minor.json`, `digest.json`: update, no warnings
* `warning.json`: no update, warning issued

Currently, `renovate` does the following:
* `digest.json`: update, no warnings
* `major.json`, `minor.json`: update, warning issued
* `warning.json`: no update, warning issued

This PR results in the expected behavior


rarkins:
@marcoesters thanks for this PR.

Could you:
1. Add a test (which would have failed prior to this PR)
2. Add a minimal reproduction repo demonstrating the new functionality?

marcoesters:
> @marcoesters thanks for this PR.
> 
> Could you:
> 
>     1. Add a test (which would have failed prior to this PR)
> 
>     2. Add a minimal reproduction repo demonstrating the new functionality?

I will try and work on the test (typescript isn't a language I'm super familiar with). I simplified the repository I mentioned in my PR description to have simple data sources and covert the scenarios mentioned in the discussion: https://github.com/marcoesters/renovate-miniconda-digest-reproducer

Do I need to add it somewhere into the code base?

lib/workers/repository/process/lookup/index.ts
```diff
@@ -541,7 +541,14 @@ export async function lookupUpdates(
...
       if (config.currentDigest) {
-        if (!config.digestOneAndOnly || !res.updates.length) {
+        // Custom datasources should not run a digest update
+        // on the current version if it already runs a version update.
+        if (
+          !(
+            config.digestOneAndOnly ?? config.datasource.startsWith('custom.')
+          ) ||
+          !res.updates.length
+        ) {
```
rarkins, line 551: I'd like to convert this logic to an if/else to reduce the amount of negating. The `if` path can have logger.trace() inside it and the else then will contain the res.updates.push()

rarkins (changes requested):

lib/workers/repository/process/lookup/index.ts
rarkins: (for future readability - the proposed change has definitely crossed a threshold for readability, maybe already had)

marcoesters: I tried a few logic combinations and they are all very much unreadable, so I resorted to a boolean that will hopefully add some more to the semantics.

This is, unfortunately, still all very hacky, but I think a more sustainable solution will require deeper insights into the renovate code bases and use cases than I currently possess.

lib/workers/repository/process/lookup/index.ts
```diff
@@ -541,7 +541,15 @@ export async function lookupUpdates(
     // Add digests if necessary
     if (supportsDigests(config.datasource)) {
       if (config.currentDigest) {
-        if (!config.digestOneAndOnly || !res.updates.length) {
+        let alwaysUpdateDigest = true;
+        if (config.digestOneAndOnly === true) {
+          alwaysUpdateDigest = false;
+        } else if (config.datasource.startsWith('custom.')) {
+          // Custom datasources should not run a digest update
+          // on the current version if it already runs a version update.
```
rarkins, line 549: If it "already ran a version update" then wouldn't res.updates.length be non-zero length anyway?

viceice: then this isn't the right solution, because it should be possible to do digest only updates with regex manager. I do that heavily on jenkinsfiles for docker images

marcoesters: My solution here still allows for digest-only updates. My example repo covers scenarios for digest-only updates. The unit tests pass as well, which contain digest-only updates.

The problem was that renovate will do a digest-only additionally to a version + digest update for the same dependency. Under the circumstances above, this will result in an incorrect warning that renovate cannot look up a digest.

PR #29745, opened 2024-06-18
feat(manager): runtime version

setchy:
## Changes

Add new manager for runtime versions.

Initial focus is on python runtimes used by
- https://devcenter.heroku.com/articles/python-runtimes
- https://www.koyeb.com/docs/build-and-deploy/build-from-git/python
- https://docs.cloudfoundry.org/buildpacks/python/index.html
- etc

Example repository: https://github.com/setchy/renovate-heroku-runtime/pull/3


lib/modules/manager/heroku-runtime/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import { DockerDatasource } from '../../datasource/docker';
+import * as dockerVersioning from '../../versioning/docker';
+
+export { extractPackageFile } from './extract';
+
+export const supportedDatasources = [DockerDatasource.id];
+
+export const defaultConfig = {
+  fileMatch: ['^runtime.txt$'],
+  versioning: dockerVersioning.id,
```
viceice, line 11:
```suggestion
```
default for docker datasource

lib/modules/manager/heroku-runtime/index.ts
```diff
@@ -0,0 +1,15 @@
+import type { Category } from '../../../constants';
+import { DockerDatasource } from '../../datasource/docker';
+import * as dockerVersioning from '../../versioning/docker';
+
+export { extractPackageFile } from './extract';
+
+export const supportedDatasources = [DockerDatasource.id];
```
viceice, line 7: please export some urls as reference links

lib/modules/manager/heroku-runtime/extract.ts
```diff
@@ -0,0 +1,12 @@
+import { DockerDatasource } from '../../datasource/docker';
+import type { PackageDependency, PackageFileContent } from '../types';
+
+export function extractPackageFile(content: string): PackageFileContent {
+  const dep: PackageDependency = {
+    depName: 'python',
+    commitMessageTopic: 'Python',
+    currentValue: content.replaceAll('python-', '').trim(),
```
viceice, line 8: is there only one line in this file? can there be other values? we should validate and document that only python is yet supported.

maybe also add a preview state note.

setchy: changed to use a regex match group

secustor (commented):
Not sure about the name as this file and content seemed to be used by other systems too. 

https://www.koyeb.com/docs/build-and-deploy/build-from-git/python
https://docs.cloudfoundry.org/buildpacks/python/index.html

Is this maybe based on a common system?

setchy:
How interesting @secustor - I didn't realize that.

So perhaps this is better off as a `python-runtime` manager

rarkins:
Can we make it a "runtime" manager more generically? If the python version always has a prefix, then in future there might be others, e.g. ruby-1.2.3 and we can then immediately accommodate them without a rename.

lib/modules/manager/runtime-version/extract.ts
```diff
@@ -0,0 +1,25 @@
...
+);
+
+export function extractPackageFile(content: string): PackageFileContent | null {
+  const regexResult = pythonRuntimeRegex.exec(content);
+  const runtimeVersion = regexResult?.groups?.version;
+
+  if (runtimeVersion) {
+    const dep: PackageDependency = {
+      depName: 'python',
+      commitMessageTopic: 'Runtime Version',
```
secustor, line 16: Why the specific message topic?

lib/modules/manager/runtime-version/index.ts
```diff
@@ -0,0 +1,15 @@
+import type { Category } from '../../../constants';
+import { DockerDatasource } from '../../datasource/docker';
+
+export { extractPackageFile } from './extract';
+
+export const displayName = 'Runtime Version';
+
+export const defaultConfig = {
+  fileMatch: ['^runtime.txt$'],
```
secustor, line 9:
```suggestion
  fileMatch: ['(^|/)runtime.txt$'],
```
Can this really be only at the root of repos?

lib/modules/manager/runtime-version/extract.ts
setchy: Copied the structure from one of the existing managers.  Easily removed

lib/modules/manager/runtime-version/index.ts
setchy: Heroku
> To specify a Python runtime, add a runtime.txt file to your app’s root directory that declares the exact version number to use.

Koyeb
> To specify the Python version to use, add a runtime.txt file in the root of your project directory with one of the versions mentioned above

CloudFoundry
> You can specify a version of the Python runtime by including it within a runtime.txt file. For example:

setchy: I'm not fussed if we relax this though

secustor (approved):

rarkins (approved):

Issue #29622, opened 2024-06-12
Support hex v2 registry


rarkins:
### Describe the proposed change(s).

Discussed in https://github.com/renovatebot/renovate/discussions/24764

In order to support private registries, we need the hex datasource to support the hex v2 registry format described here: https://github.com/hexpm/specifications/blob/main/registry-v2.md

It seems that we should switch to using this registry protocol by default, including for the public registry.

bryannaegele:
How are the elixir/erlang versions set? And is it possible to include an elixir script to fetch the package info? That would make this very straightforward.

rarkins:
No, we don't permit the calling of CLI tools as part of lookup

bryannaegele:
Ok. The suggestion here is to allow a script which is executed with `mix run` that is just fetching and decoding the list of releases from a registry. Everything needed is already available when installing elixir which is allowed in the [mix manager](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/mix/artifacts.ts#L116-L123) but not currently done in the hex datasource. Hex is shipped as part of Elixir, so there are no other cli requirements.

The current methodology fetches json from an API and performs no validations. Using the registry, responses are protobuf encoded and include other features such as package signing verification. So our options for hex are to either leverage the functionality that ships with elixir for auth, fetching, decoding, and package signing verification or reimplement all of that functionality in TS.

I wasn't able to find any other datasources that are processing protobufs or protobuf related packages in core, so that would be a new addition.

So the proposal here is to let hex interact with hex registries in the hex datasource. Are there concerns or certain pieces of other functionality that are part of the current policy? We're trying to understand the reason for the policy if it pertains only to package lookup.

bryannaegele:
This is a very basic demo of the proposal.

https://gist.github.com/bryannaegele/4b9d11bc927d3887267028d1ea9f64f9

rarkins:
We want datasources to be in TS so they can be used in the browser.  Not other datasource requires binaries for lookups. If it's extremely hard it perhaps indicates that the registry should be made more accessible in future? Most are REST-based.

viceice:
running external code for datasources will slow down the lookup phase massively, because we need to ensure to install a proper version of erlang and elixir. that is usually only done when we need to update lockfiles.

so I'm strictly against calling any binaries inside datasources.

There are official JavaScript libraries to decode protobuff messages easily.

bryannaegele:
Yep. Understandable. I'll get working on it

[...]

PR #29756, opened 2024-06-19
feat(hex): use registry v2 for packages rather than hex.pm http API

bryannaegele:
## Changes

- use hex registry v2 for package information
- use hex.pm package API for supplemental package metadata when package is hosted there.
- support `registryAliases` for private repo/registry mapping
- run `mix.repo add` in mix manager for private repos

## Context

https://github.com/renovatebot/renovate/issues/29622

This changes the `hex` datasource to use hex registry v2 for package management. Hex registries are not required to implement an http API like `hex.pm` has, so the prior implementation prevented updating of packages in private repos such as oban pro. Additionally, due to the way mix works when updating dependencies, `mix.lock` files could not be updated when such a dependency existed and the mix manager ran `mix deps.update <<updated packages>>`. 

In order to keep as much parity with the prior implementation, hex.pm is used for supplemental package metadata where possible.


bryannaegele:
Leaving this as a draft until there's an answer for this https://github.com/renovatebot/renovate/issues/29622#issuecomment-2177278783

I am no TS expert, so please feel free to jump in and fix up whatever can be done better.

lib/modules/datasource/hex/package.spec.ts
```diff
@@ -0,0 +1,17 @@
+// import { mockDeep } from 'jest-mock-extended';
```
viceice, line 1: remove

lib/modules/datasource/hex/package.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+import { gunzipSync } from 'zlib';
```
viceice, line 2: use async functions

lib/modules/manager/mix/artifacts.ts
```diff
@@ -11,8 +11,8 @@ import {
 } from '../../../util/fs';
 import * as hostRules from '../../../util/host-rules';
 import { regEx } from '../../../util/regex';
-
 import type { UpdateArtifact, UpdateArtifactsResult } from '../types';
+import { aliasRecordToRepositories } from './utils';
```
viceice, line 15: split datasource and manager changes to separate PR

lib/modules/datasource/hex/signed.ts
```diff
@@ -0,0 +1,154 @@
+// Code generated by protoc-gen-ts_proto. DO NOT EDIT.
```
viceice, line 1: this should be generated on build and not checked into git. or fix it to a readable format for proper linting.

viceice (commented):
reduce fixtures, we don't need all versions for testing, ~3 are enough. ensure the tgz file is at minimum too

lib/modules/datasource/hex/signed.ts
bryannaegele: Can you help with how you would go about that? I am not a TS developer and this is what was generated by https://www.npmjs.com/package/ts-proto#usage, unedited in any way. There are a lot of options that can be passed to it but it's all foreign to me. The comment and eslint ignore are output by ts-proto.

I'm not sure how generating the files on build ends up with a different result. It would still be the same code, just lint ignored? And the code has to be there for unit tests, so I'm not clear on how a build step improves things.

bryannaegele:
> reduce fixtures, we don't need all versions for testing, ~3 are enough. ensure the tgz file is at minimum too

Is this strictly necessary? The tgz has the registry version of the package wrapped in a signed wrapper. I don't have any way to create a version with less releases in it that would have a valid signature.

I can try to find a different package as an example but this is one of the rare ones with retirements.

bryannaegele:
Mix changes split to #29775

lib/modules/manager/mix/artifacts.ts
bryannaegele: #29775

lib/modules/datasource/hex/signed.ts
viceice: you need to add that package to dev dependencies, then add a generate step to package.json scripts. there are already some generation steps as samples.

bryannaegele:
Ok. I think everything has been addressed. This and #29775 are ready to go @viceice and @rarkins

.github/workflows/build.yml
```diff
@@ -207,6 +207,9 @@ jobs:
           node-version: ${{ needs.setup-build.outputs.node-version }}
           os: ${{ runner.os }}
 
+      - name: Generate protos
+        run: pnpm create-protos
+
```
viceice, line 212: revert those changes and do it in generate script, see package.json for other generate scripts

package.json
```diff
@@ -8,14 +8,15 @@
...
-    "build": "run-s clean 'generate:*' 'compile:*' create-json-schema",
+    "build": "run-s clean 'generate:*' create-protos 'compile:*' create-json-schema",
     "build:docker": "ts-node tools/docker.ts",
     "build:docs": "ts-node tools/generate-docs.ts",
     "clean": "rimraf dist tmp",
     "clean-cache": "node tools/clean-cache.mjs",
     "compile:ts": "tsc -p tsconfig.app.json",
     "config-validator": "ts-node lib/config-validator.ts",
     "create-json-schema": "ts-node tools/generate-schema.ts && prettier --write --cache 'renovate-schema.json'",
+    "create-protos": "ts-node tools/generate-protos.ts",
```
viceice, line 19:
```suggestion
    "generate:create-protos": "ts-node tools/generate-protos.ts",
```
then all other changes can be reverted

tools/generate-protos.ts
```diff
@@ -0,0 +1,98 @@
...
+import { getProblems, logger } from '../lib/logger';
+import { exec } from './utils/exec';
+
+process.on('unhandledRejection', (err) => {
+  // Will print "unhandledRejection err is not defined"
+  logger.error({ err }, 'unhandledRejection');
+  process.exit(-1);
+});
+
+const moveFiles = async (sourceDir: string, destDir: string): Promise<void> => {
```
viceice, line 12: use normal function declaration

tools/generate-protos.ts
```diff
@@ -0,0 +1,98 @@
...
+
+function istanbulIgnoreFile(file_path: string): string {
+  fs.appendFile(file_path, '\n// istanbul ignore file', function (err) {
+    if (err) {
+      throw err;
+    }
+  });
+
+  return '';
+}
```
viceice, line 56: exclude in jest.config.ts instead

tools/generate-protos.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  const protos_path = './lib/modules/datasource/hex/protos';
+  await generateProto(protos_path, 'package.proto');
+  await generateProto(protos_path, 'signed.proto');
+  istanbulIgnoreFile(`${protos_path}/package.ts`);
+  istanbulIgnoreFile(`${protos_path}/signed.ts`);
+  await moveFiles(
+    `${process.cwd()}/lib/modules/datasource/hex/protos`,
+    `${process.cwd()}/lib/modules/datasource/hex`,
+  );
```
viceice, line 96: why moving? why not directly generate on right path?

.github/workflows/build.yml
bryannaegele: That was my first approach. I tried doing it in generate but that creates a dependency issue and requires modifying a number of test commands and likely the github workflows to run generate where they currently do not. One example is linting which does not have a dependency on `setup` currently. This causes the files to not be present when there are references to them.

image

Introducing a generic protos creation step was the least intrusive solution I could get to.

If you can assist in getting to a working solution with generate, it would be much appreciated.

package.json
bryannaegele: See note in the other comment about generate not running in the lint pipeline.

tools/generate-protos.ts
bryannaegele: It is only able to generate the file to the current directory where the proto is located. Currently, the proto files are separated for organizational purposes so it would just mean having the proto files interspersed with the ts files, e.g. `signed.proto` and `signed.ts` in the same directory. Either approach works so happy to go whichever way you'd like.

bryannaegele:
Ok. I am at a point now where trying to get the protos to generate on build in `generate:` and putting them to the correct place for the sake of not committing the generated files is at a breaking point for me. There is a dependency chain issue here that I cannot find a way resolve without extensive amount of changes to the workflows and/or package scripts.

* the generated files must live in the datasource, so `lib/modules/datasource/hex` to compile
* the `generate:` command runs on a sparse checkout that does not include `lib`
* adding that path to the sparse checkout results in error:
```. prepare: generating hashes
. prepare: ERROR: ENOENT: no such file or directory, scandir 'lib/modules/manager'```
* subsequent steps such as lint do a full checkout
* any generated file would be overwritten by the full checkout, so generation would have to happen again after every step that does that

I understand the desire to not add the `create-protos` step I added which was working fine [here](https://github.com/renovatebot/renovate/pull/29756/commits/a433ab7e0604ddc7c004f2b7e969d9c42272e7c4), but trying to shoehorn it into `generate:` is infeasible imo. I've attempted it two times now and it just doesn't work. It also can't go into `build:` like the docs because `compile` runs before that and the generated files must be there.

If the files must be generated and not committed, can we please use the new workflow I had introduced?

I feel the simplest solution is still to commit the generated files which we're ignoring for coverage and linting in jest config now anyways and documenting the instructions for generating them should they ever need to be updated. It would be great to reconsider that approach again.

Thanks

**Comment by viceice** [comment]
> Ok. I am at a point now where trying to get the protos to generate on build in `generate:` and putting them to the correct place for the sake of not committing the generated files is at a breaking point for me. There is a dependency chain issue here that I cannot find a way resolve without extensive amount of changes to the workflows and/or package scripts.
> 
> * the generated files must live in the datasource, so `lib/modules/datasource/hex` to compile
> * the `generate:` command runs on a sparse checkout that does not include `lib`
> * adding that path to the sparse checkout results in error:
> ```. prepare: generating hashes
> . prepare: ERROR: ENOENT: no such file or directory, scandir 'lib/modules/manager'```
> * subsequent steps such as lint do a full checkout
> * any generated file would be overwritten by the full checkout, so generation would have to happen again after every step that does that
> 
> I understand the desire to not add the `create-protos` step I added which was working fine [here](https://github.com/renovatebot/renovate/pull/29756/commits/a433ab7e0604ddc7c004f2b7e969d9c42272e7c4), but trying to shoehorn it into `generate:` is infeasible imo. I've attempted it two times now and it just doesn't work. It also can't go into `build:` like the docs because `compile` runs before that and the generated files must be there.
> 
> If the files must be generated and not committed, can we please use the new workflow I had introduced?
> 
> I feel the simplest solution is still to commit the generated files which we're ignoring for coverage and linting in jest config now anyways and documenting the instructions for generating them should they ever need to be updated. It would be great to reconsider that approach again.
> 
> Thanks

@zharinov can you help with fixing the CI in an elegant way?

**Comment by zharinov** [comment]
Since we don't expect generated files to change frequently, why not commit them? At the CI level we could check the generated code is identical to the committed one, and well... sometimes we have to replace it manually.

I don't know how how bad consequences are, but if we ignore `signed.ts` and `package.ts`, our typescript code is broken unless the `pnpm install` command is executed.

**Comment by viceice** [comment]
> Since we don't expect generated files to change frequently, why not commit them? At the CI level we could check the generated code is identical to the committed one, and well... sometimes we have to replace it manually.
> 
> I don't know how bad consequences are (probably, the language server will suffer), but if we ignore `signed.ts` and `package.ts`, our typescript project is broken unless the `pnpm install` command is executed.

it's already broken without install

**Comment by zharinov** [comment]
> it's already broken without install

Well, fair enough. I think generate phase may just quit early if there is no `hex` subdirectory, and it should avoid usage of `utils/exec` in favor of the Node's basic one.

**Comment by viceice** [comment]
> > it's already broken without install
> 
> Well, fair enough. I think generate phase may just quit early if there is no `hex` subdirectory, and it should avoid usage of `utils/exec` in favor of the Node's basic one.

sounds good. only use node build-in or deps, but no code from src inside prepare steps?

**Comment by zharinov** [comment]
Yes, because if the `lib/` directory is absent for this particular CI task, the module import attempt will fail.

**Comment by bryannaegele** [comment]
If it helps on frequency of update with regard to committing the generated code, these protos are [5 and 6 years old](https://github.com/hexpm/specifications/tree/main/registry) at this point. They may never be updated again. I suppose the generated code may need an update at some point due to changes in TS or a bug but the code seems pretty simple, so low risk there.

**Comment by viceice** [comment]
ok, how big are the generated files? maybe we can check them in and ignore in linting

**Comment by viceice** [comment]
or move the proto files to a different root folder which is checked out

**Comment by bryannaegele** [comment]
> ok, how big are the generated files? maybe we can check them in and ignore in linting

| file | generated | formatted | formatted and encoding functions manually deleted |
| --- | --- | --- | --- |
| signed.ts | 132 LOC | 155 LOC | 119 LOC |
| package.ts | 570 LOC | 638 LOC | 479 LOC |

**Comment by zharinov** [comment]
> ok, how big are the generated files? maybe we can check them in and ignore in linting

Yeah, I think having the task like `codegen:hex-datasource-protobuf` is enough

**Comment by bryannaegele** [comment]
> or move the proto files to a different root folder which is checked out
Missed this comment. The issue isn't really where the proto files live, it's that the generated files have to exist in the hex datasource folder and can't be overwritten by a subsequent checkout that a lot of the steps do. So that would introduce having to generate and copy the files from a different location anytime a checkout happens in CI.

**Comment by bryannaegele** [comment]
@zharinov and @viceice - have we landed on just committing the generated files or is someone taking over to get codegen how y'all want it?

**Comment by bryannaegele** [comment]
Thanks. I'll get that done.

One thing to note is the base functionality has not changed, it's all enhancements, so it shouldn't be breaking. And we probably want to get #29775 reviewed to be merged in conjunction with this one. They are dependent on each other.

**Comment by bryannaegele** [comment]
Alright. Made the following requested changes:
- remove codegen from CI and setup
- commit generated files
- add readme with info on how to generate the proto files
- renamed script name in package per @zharinov's suggestion

I have kept the generation script in tools so that if anyone should need to regenerate the files, it's as simple as possible.

**Comment by esambo** [comment]
@bryannaegele FYI, you have merge conflicts

**Comment by bryannaegele** [comment]
> @bryannaegele FYI, you have merge conflicts

Are y'all ready to merge this and the other PR? I'd like to not keep resolving the conflicts unless it's going to get merged.

File: lib/modules/datasource/hex/index.ts [hunk 10]
```diff
@@ -38,31 +43,111 @@ export class HexDatasource extends Datasource {
...
+    const resp = await this.http.getBuffer(hexRegistryUrl);
+
+    // istanbul ignore else
+    if (resp.statusCode === 200) {
+      await decompressBuffer(resp.body)
+        .then((signedPackage) => {
+          const { payload: payload } = Signed.decode(signedPackage);
+          const registryPackage = Package.decode(payload);
+
+          const releases: Release[] = registryPackage.releases.map(
```
viceice, line 83: use for-of loop

lib/modules/datasource/hex/index.ts
```diff
@@ -38,31 +43,111 @@ export class HexDatasource extends Datasource {
...
+    const releaseResult: ReleaseResult = { releases: [] };
+
+    if (packageName.startsWith('org:')) {
+      const [, organizationName, hexPackageName] = packageName.split(':');
+      urlPath = `/repos/${organizationName}/packages/${hexPackageName}`;
+      releaseResult.isPrivate = true;
+    } else if (packageName.startsWith('repo:')) {
+      const [, repoName, hexPackageName] = packageName.split(':');
+      urlPath = `/packages/${hexPackageName}`;
+      this.hexRepoName = repoName;
```
viceice, line 65: no, we've concurrent calls, don't store state in datasource . use local function variable

lib/modules/datasource/hex/index.ts
```diff
@@ -38,31 +43,111 @@ export class HexDatasource extends Datasource {
...
+        .catch(
+          /* istanbul ignore next */ (err) => {
+            return null;
+          },
+        );
+
+      if (this.hexRepoName === 'hexpm' && releaseResult.releases.length > 0) {
+        const metadataUrl = joinUrlParts(this.hexAPIBaseUrl, urlPath);
+
+        logger.debug(`Package metadata url: ${metadataUrl}`);
```
viceice, line 108:
```suggestion
        logger.trace(`Package metadata url: ${metadataUrl}`);
```

lib/modules/datasource/hex/index.ts
```diff
@@ -38,31 +43,111 @@ export class HexDatasource extends Datasource {
...
+
+        const { val: packageMetadata, err } = await this.http
+          .getJsonSafe(metadataUrl, HexAPIPackageMetadata)
+          .unwrap();
+
+        // istanbul ignore if
```
viceice, line 114: no, add test

lib/modules/datasource/hex/index.ts
```diff
@@ -38,31 +43,111 @@ export class HexDatasource extends Datasource {
...
+        if (err) {
+          this.handleGenericErrors(err);
+        } else {
+          releaseResult.changelogUrl = packageMetadata.meta?.links.Changelog;
+          releaseResult.sourceUrl = packageMetadata.meta?.links.Github;
+          releaseResult.homepage = packageMetadata.html_url;
+
+          const releasesWithMeta = releaseResult.releases.map((rel) => {
```
viceice, line 122: use for-of loop

tools/generate-hex-protos.ts
```diff
@@ -0,0 +1,86 @@
...
+    console.log('Unexpected error');
+    console.log(err);
+    process.exit(1);
+  }
+})();
+
+function generateProto(protos_path: string, file: string): Promise<string> {
+  return new Promise((resolve, reject) => {
+    console.log(`Current directory: ${process.cwd()}`);
+    fs.readdirSync(process.cwd()).forEach((file) => {
```
viceice, line 47: use `fs/promises` api

viceice (changes requested):

lib/modules/datasource/hex/index.ts
```diff
@@ -1,28 +1,33 @@
...
 
-  override readonly defaultRegistryUrls = ['https://hex.pm'];
+  override readonly defaultRegistryUrls = ['https://repo.hex.pm'];
 
   override readonly defaultVersioning = hexVersioning.id;
 
-  override readonly releaseTimestampSupport = true;
-  override readonly releaseTimestampNote =
-    'The release timestamp is determined the `inserted_at` field in the results.';
+  override readonly releaseTimestampSupport = false;
```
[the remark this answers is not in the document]
bryannaegele: Timestamps are not a piece of metadata in the registry metadata. They are only a piece of metadata that hex.pm supports via its HTTP API which is not a requirement to run a registry.

[Available fields](https://github.com/hexpm/specifications/blob/main/registry/package.proto).

`inserted_at` field [is preserved](https://github.com/renovatebot/renovate/pull/29756/files#diff-99307b00eaeb9c01bb1a10333bde71e11f13a5c3eab4c19a1c1101726d4f1d37R18) but that field is only available for a package served from hexpm.

Can you explain why it's a showstopper? I understand it's a potentially breaking change but the original implementation was technically incorrect. I'm not sure how the `releaseTimestampSupport` is used. If we set it to `true` and the value isn't available for some packages, what would happen?

lib/modules/datasource/hex/index.ts
```diff
@@ -38,31 +43,111 @@ export class HexDatasource extends Datasource {
...
+
+          releaseResult.releases = releasesWithMeta;
+
+          return releaseResult;
+        }
+      }
+
+      return releaseResult;
+    } else {
+      // not sure how to handle error here since get or getBuffer doesn't return an error
```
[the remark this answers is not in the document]
bryannaegele: I don't understand what this means. Can you point to an example or provide a suggestion?

lib/modules/datasource/hex/index.ts
```diff
@@ -38,31 +43,111 @@ export class HexDatasource extends Datasource {
...
-    return result;
+    const hexRegistryUrl = joinUrlParts(registryUrl, urlPath);
+    logger.debug(`Package registry url: ${hexRegistryUrl}`);
+
+    const resp = await this.http.getBuffer(hexRegistryUrl);
+
+    // istanbul ignore else
+    if (resp.statusCode === 200) {
+      await decompressBuffer(resp.body)
+        .then((signedPackage) => {
```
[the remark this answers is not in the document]
bryannaegele: Can I ask why? await -> then -> else is used extensively in the codebase. This feels like a nit.

bryannaegele:
**Handing to Maintainers to Finish**

I leave this PR and the Mix PR #29775 to the maintainers to complete. My limited time is better spent on my maintainer duties elsewhere.

These PRs are required to Elixir/Hex support for companies to adopt Renovate in non-trivial setups. First-order support for the hex registry is necessary to support the Hex package manager. This was unfortunately missed during original implementation in the good faith effort to support Hex/Mix.

Reasons for the handoff:
* compatibility concerns and solutions are best left to maintainers
* maintainer expertise on testing techniques employed in the project will provide better coverage
* strong opinions on code style, organization, and nits are best resolved by codeowners

Three months and >100 hours greatly surpasses the level of effort I thought this would take when I volunteered to investigate and take a stab.

Renovate is a great project. I hope you can find resources to complete the work @rarkins so we can consider using it for Elixir projects again someday.

Thanks

#29622

rarkins:
@bryannaegele thanks for your efforts. I've closed this if you're unable to continue and hope someone can pick it up in future.

zharinov:
@bryannaegele Thanks for your effort, it definitely isn't easy one 🤝 

I'll take a look if I could reuse it as series of smaller changes.

PR #29775, opened 2024-06-20
feat(mix): Support private hex repos/registries

bryannaegele:
## Changes

This adds support to the mix manager to add private hex repos/registries via `registryAliases` for mapping and `hostRules` for authentication.

## Context

This is the companion to #29756 and dependent on that change. Splitting the manager and datasource work per request in #29756. When combined, this will enable support for private repos/registries such as oban pro.

Tested in combination with #29756. This will not fully work without that code.


epinault:
Any update on this? can we get this merged?

bryannaegele:
Feedback should be addressed 👍🏻

lib/modules/manager/mix/artifacts.spec.ts
```diff
@@ -291,6 +303,45 @@ describe('modules/manager/mix/artifacts', () => {
...
+      newPackageFileContent: '{}',
+      config,
+    });
+
+    expect(result).toEqual([
+      {
+        file: { type: 'addition', path: 'mix.lock', contents: 'New mix.lock' },
+      },
+    ]);
+    expect(execSnapshots).toMatchObject([
```
viceice, line 338: i don't see a test for the `mix repo.add ...` command

lib/modules/manager/mix/artifacts.ts
```diff
@@ -75,26 +78,50 @@ export async function updateArtifacts({
 
   for (const { packageName } of updatedDeps) {
     if (packageName) {
-      const [, organization] = packageName.split(':');
+      const [, organization, ,] = packageName.split(':');
```
viceice, line 81:
```suggestion
      const [, organization] = packageName.split(':');
```
not needed

lib/modules/manager/mix/artifacts.ts
```diff
@@ -75,26 +78,50 @@ export async function updateArtifacts({
...
+    // regex match anything hex.pm to an org
+    // if not hex, do repo.add
+    const repoAliases = aliasRecordToRepositories(config.registryAliases);
+
+    for (const { name, registry } of repoAliases) {
+      const hostRule = hostRules.find({ url: registry });
 
-    return acc;
-  }, [] as string[]);
+      // istanbul ignore if: no good way to test
```
viceice, line 112: no, why? you can mock hostrule find properly

lib/modules/manager/mix/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export function resolveAlias(
+  packageName: string,
+  registryAliases: Record<string, string>,
+): string | null {
+  if (!isAlias(packageName)) {
+    return packageName;
+  }
+
+  const [, repoName, ,] = packageName.split(':');
```
viceice, line 19:
```suggestion
  const [, repoName] = packageName.split(':');
```

lib/modules/manager/mix/readme.md
```diff
@@ -1,3 +1,55 @@
...
+  hostRules: [
+    {
+      matchHost: 'https://getoban.pro/repo',
+      token: process.env.RENOVATE_OBAN_AUTH_TOKEN,
+      authType: 'Token-Only',
+    },
+  ],
+  registryAliases: {
+    oban: 'https://getoban.pro/repo',
+  },
```
viceice, line 53: this should be scoped to `mix` manager, otherwise it will interferre with other managers using registry aliases

lib/modules/manager/mix/index.ts
```diff
@@ -8,6 +8,9 @@ export { extractPackageFile } from './extract';
 export { updateArtifacts } from './artifacts';
 
 export const defaultConfig = {
+  registryAliases: {
+    hexpm: 'https://hex.pm',
+  },
```
viceice, line 13: why is this needed

lib/modules/manager/mix/extract.ts
```diff
@@ -60,11 +69,41 @@ export async function extractPackageFile(
...
-            packageName: organization ? `${app}:${organization}` : app,
+            packageName: app,
           };
+
           if (requirement?.startsWith('==')) {
             dep.currentVersion = requirement.replace(regEx(/^==\s*/), '');
           }
+
+          if (organization) {
+            dep.packageName = `org:${organization}:${app}`;
```
viceice, line 80: this is used by the hex datasource, so needs changes there too

https://github.com/renovatebot/renovate/blob/302718f38e54c6c7590f488d749fb4da8362f36f/lib/modules/datasource/hex/index.ts#L45

bryannaegele: That is in the other PR. You requested to split them so neither PR will work on its own

lib/modules/manager/mix/utils.ts
bryannaegele: I don't understand the suggestion. This wants to extract the second value in whatever is returned.

lib/modules/manager/mix/artifacts.spec.ts
bryannaegele: We aren't taking credentials and logging them in. The user provides the auth token generated by doing that command locally and which is what is passed in the auth header. 

`mix repo.add` takes a set of credentials and other arguments and stores an auth token on the local machine. When a dependency with a repo that isn't `hexpm` is encountered, it looks up the token and URL from that config. Rather than inventing and introducing a new auth system to renovate, we use the package rule to look up the URL and the host rule then adds the token to the request.

The readme lays out the configuration for the user pretty clearly. Let me know if something is missing there

lib/modules/manager/mix/index.ts
bryannaegele: We need the URL for the hexpm registry, which is the default, so we put it into the default config. This is merged with anything the user has supplied, giving us a complete lookup of the URLs. This has an added benefit that if a company wished to host an internal proxy registry, no updates are needed to their mix files; they have a way to point someplace else.

lib/modules/manager/mix/artifacts.ts
bryannaegele: I don't understand the suggestion. This wants to extract the second value in whatever is returned. The packageName when split can be 2 or 3 elements in length. The suggestion only works when array of length 2 is returned.

viceice: no, the split returns all elements, but the array destruction will only return the first two and ignore the rest.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

lib/modules/manager/mix/extract.ts
viceice: then don't change package name in this PR, do the package name change in a separate refactor PR

lib/modules/manager/mix/index.ts
viceice: you should not add a default url, the datasource has it already, so only set registry url if it's not the default

lib/modules/manager/mix/artifacts.spec.ts
viceice: then add a test for it

lib/modules/manager/mix/extract.ts
bryannaegele: So neither PR is backwards compatible with the other. They are both dependent on the format change. If we don't change this here then your proposed refactor PR just pushes the change both need to make into a single PR that touches the datasource and manager.

lib/modules/manager/mix/readme.md
1oglop1: @bryannaegele are you still working on this? 
I have the same/very similar use case and I'd love renovate to support elixir correctly.

this particular change can be observed here https://github.com/renovatebot/renovate/blob/main/docs/usage/configuration-options.md#registryaliases 
IMO this is what viceice is asking about.
```suggestion
  {
  "mix": {
   registryAliases: {
    oban: 'https://getoban.pro/repo',
   },
  }
}
```

epinault:
this ticket is closed, what s the plan? is there a solution merged? what s the end result?

rarkins:
The original author was not able to continue the work. You are welcome to pick it up where he left off and submit a new PR

PR #29787, opened 2024-06-20
feat(pipenv): Use `@renovatebot/detect-tools` for constraints detection

### LINKED ISSUE #29000: Move tool detection to renovatebot/detect-tools

We should move tool detection to a separate repo/module so that it can be shared outside Renovate, particularly with some Containerbase tooling we are building.

Start with Poetry and Pipenv.

Example from: https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/pipenv/artifacts.ts

Instead of a `getPythonConstraint()` function where we pass _content_ to the function, we should instead call our new external library and pass it a fully qualified _path_/directory.

zharinov:
## Changes

Use drop-in library replacement for pypi constraint detection.

## Context

- Closes: #29000
- https://github.com/renovatebot/detect-tools
- https://www.npmjs.com/package/@renovatebot/detect-tools?activeTab=code


lib/modules/manager/pipenv/artifacts.ts
```diff
@@ -1,5 +1,5 @@
+import * as detectTools from '@renovatebot/detect-tools';
```
viceice, line 1: why the import star is required? can't we use a default export?

viceice (commented):
new code looks promising, please also let me review changes to the detect tools. I need them later for containerbase.

lib/modules/manager/pipenv/schema.ts
viceice, line 1: Please don't delete, it should be used to extract locked version later

lib/modules/manager/pipenv/artifacts.ts
```diff
@@ -289,16 +139,11 @@ export async function updateArtifacts({
...
-      newPipfileContent,
-      existingLockFileContent,
-      config,
-    );
-    const pipenvConstraint = getPipenvConstraint(
-      existingLockFileContent,
-      config,
-    );
+    const pipfileDir = getParentDir(ensureLocalPath(pipfileName));
+    const tagConstraint = await pipenvDetect.getPythonConstraint(pipfileDir);
```
viceice, line 143:
```suggestion
    const tagConstraint = config.constraints?.python ?? await pipenvDetect.getPythonConstraint(pipfileDir);
```
we should allow configure python version via constraints too

lib/modules/manager/pipenv/artifacts.ts
```diff
@@ -289,16 +139,11 @@ export async function updateArtifacts({
...
+    const pipenvConstraint =
+      config?.constraints?.pipenv ??
```
viceice, line 145:
```suggestion
      config.constraints?.pipenv ??
```

lib/modules/manager/pipenv/artifacts.spec.ts
```diff
@@ -845,19 +880,22 @@ describe('modules/manager/pipenv/artifacts', () => {
...
 
   it('uses pipenv version from config', async () => {
     GlobalConfig.set(dockerAdminConfig);
+    fsExtra.stat.mockResolvedValueOnce({} as never);
     fsExtra.ensureDir.mockResolvedValue(undefined as never);
 
     const oldLock = JSON.stringify({
       default: { pipenv: { version: '==2020.8.13' } },
-    } satisfies PipfileLockSchema) as never;
+    }) as never;
```
viceice, line 898: revert

lib/modules/manager/pipenv/artifacts.spec.ts
```diff
@@ -24,7 +24,6 @@ import {
   extractEnvironmentVariableName,
   getMatchingHostRule,
 } from './artifacts';
-import type { PipfileLockSchema } from './schema';
```
viceice, line 27: revert, we should still use it see other comments

lib/modules/manager/pipenv/artifacts.ts
zharinov: Yep, I don't remember why I removed it, seems like by error

lib/modules/manager/pipenv/artifacts.spec.ts
zharinov: It doesn't exist anymore

zharinov: It was heavily refactored in the `detect-tools` package

lib/modules/manager/pipenv/artifacts.ts
```diff
@@ -289,16 +139,13 @@ export async function updateArtifacts({
...
-      existingLockFileContent,
-      config,
-    );
-    const pipenvConstraint = getPipenvConstraint(
-      existingLockFileContent,
-      config,
-    );
+    const pipfileDir = getParentDir(ensureLocalPath(pipfileName));
+    const tagConstraint =
+      config?.constraints?.python ??
```
viceice, line 144:
```suggestion
      config.constraints?.python ??
```
config is never null or undefined here

lib/modules/manager/pipenv/artifacts.ts
```diff
@@ -289,16 +139,13 @@ export async function updateArtifacts({
...
+      (await pipenvDetect.getPythonConstraint(pipfileDir));
+    const pipenvConstraint =
+      config?.constraints?.pipenv ??
```
viceice, line 147:
```suggestion
      config.constraints?.pipenv ??
```

viceice (commented):
https://github.com/renovatebot/renovate/blob/607d7e3e1bc973a65a9a964fbdd7d050dc3a6239/lib/modules/manager/pipenv/extract.ts#L171

this should be replaced too

lib/modules/manager/pipenv/extract.spec.ts
```diff
@@ -99,7 +114,7 @@ describe('modules/manager/pipenv/extract', () => {
           },
         ],
         extractedConstraints: {
-          python: '== 3.6.*',
+          python: '== 3.6.2',
```
viceice, line 117: why this change? 🤔

viceice (commented):
otherwise LGTM

lib/modules/manager/pipenv/extract.spec.ts
zharinov: This is how artifact extraction logic works, it turned out to be different than `extract.ts`

viceice (approved):

PR #29791, opened 2024-06-21
chore(regex): add additional logging for `isValidDependency` failures

jamietanna:
## Changes

Adds TRACE level logging, as per discussion on Renovate Slack.

## Context

When writing regexes to match custom file formats, it may not be obvious
why a valid regex may not be resulting in a match.

This is often because we have written a perfectly correct regex which
has matched the line(s), but then our `isValidDependency` check fails.

To make it easier to discover this, we can add a TRACE log, and note
that we will discard the dependency, giving suitable context for what is
being rejected.

### DIFF
--- a/lib/modules/manager/custom/regex/strategies.ts
+++ b/lib/modules/manager/custom/regex/strategies.ts
@@ -31,19 +31,9 @@ export function handleAny(
       ),
     )
     .filter(is.truthy)
-    .filter((dep: PackageDependency) => {
-      const isValid = isValidDependency(dep)
-      if (!isValid) {
-        const meta = {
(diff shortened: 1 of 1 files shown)


lib/modules/manager/custom/regex/strategies.ts
```diff
@@ -30,12 +31,24 @@ export function handleAny(
       ),
     )
     .filter(is.truthy)
-    .filter(isValidDependency);
+    .filter((dep: PackageDependency) => {
```
viceice, line 34: create a util function and reuse

rarkins (approved):

PR #29802, opened 2024-06-21
feat(gerrit): use commit message footers to store source branch name

felipecrs:
## Changes

This changes the approach used to store and retrieve the source branch information in Gerrit. Instead of relying on hashtags, we now store the source branch information in a commit message footer named `Renovate-Branch`.

I believe the new approach is a lot friendlier for the end user. Less hacky, more intuitive, and better looking. Kind of aligns with the Gerrit way of doing things (like how they store and retrieve the change id using the `Change-Id` footer).

Here is an example of how it was:

https://review.gerrithub.io/c/felipecrs/gerrit-jenkins/+/1196601/1//COMMIT_MSG

image

And here is an example on how it is now:

https://review.gerrithub.io/c/felipecrs/gerrit-jenkins/+/1196702

image

## Context

I had proposed this in https://github.com/renovatebot/renovate/pull/18961#issuecomment-1633281364, but we concluded to do this post-merge of the Gerrit support given for how long that PR was being dragged on already.

So, I finally got around to doing this.


felipecrs:
@NiasSt90, @viceice, @valodzka, @kerby82, found your names from previous Gerrit-related PRs. I am just pinging you in case you'd like to opine about/review this PR.

elsisi33 (approved):

lib/modules/platform/gerrit/scm.ts
```diff
@@ -109,7 +109,7 @@ export class GerritScm extends DefaultGitScm {
       typeof commit.message === 'string' ? [commit.message] : commit.message;
     commit.message = [
       ...origMsg,
-      `Change-Id: ${existingChange?.change_id ?? generateChangeId()}`,
+      `Renovate-Source-Branch: ${commit.branchName}\nChange-Id: ${existingChange?.change_id ?? generateChangeId()}`,
```
felipecrs, line 112: Here is one of my own notes:

If we make this shorter, there is a lesser chance that this line will offend line length rules of commit message linters.

For example: `Source-Branch` or `Renovate-Branch`.

rarkins: I think it's important to decide/finalize this now?

NiasSt90:
I don't like the idea of doing this as a breaking change.

We can/should search for the old hashtag too. (syntax untested)

    filters.push(`hashtag:sourceBranch-${searchConfig.branchName} OR footer:.....`);

lib/modules/platform/gerrit/scm.ts
felipecrs: Absolutely.

felipecrs:
@NiasSt90, yeah, keeping backwards compatibility is doable. But it would be considerably more difficult to do:

- Search for existing issues (your comment)
- Extract source branch from hashtags and from the commit message footer
- Add some kind of migration function to migrate from hashtags to commit message footer
- Add tests for all variants of the points above

I'm just not sure if it's worth it. As I mentioned, Gerrit support is still [marked as _experimental_](https://docs.renovatebot.com/modules/platform/gerrit/).

NiasSt90:
> @NiasSt90, yeah, keeping backwards compatibility is doable. But it would be considerably more difficult to do:
> 
> * Search for existing issues (your comment)
> * Extract source branch from hashtags and from the commit message footer

use the existing `extractSourceBranch` code and append your new search to the end...if first search found nothing.

> * Add some kind of migration function to migrate existing changes from hashtags to commit message footer

migration? 
You search/extract from both (extract from hashtags first) sources and write only to the new (commit-message) destination....nothing more needed or?

> * Add tests for all variants of the points above

combine the old and the new tests....instead of change the old ones.

> 
> I'm just not sure if it's worth it. As I mentioned, Gerrit support is still [marked as _experimental_](https://docs.renovatebot.com/modules/platform/gerrit/).

because it's easy doable....

felipecrs:
> migration?
> You search/extract from both (extract from hashtags first) sources and write only to the new (commit-message) destination....nothing more needed or?

Does Renovate always recreate changes from scratch and only retain their source branch and commit ids? I thought Renovate would sometimes just rebase changes, in which case I would expect to write code to migrate such change from hashtag to footer.

Also, hashtags won't be gone from Gerrit once a new patchset is published with the footer. We would need to write code for it.

felipecrs:
Or maybe you mean we should search for both kind of changes and let it be. Someday users would end up merging or closing all changes and then there won't be more changes with hashtags.

We don't necessarily need to do housekeeping of removing the hashtags.

Is that what you meant?

NiasSt90:
> Or maybe you mean we should search for both kind of changes and let it be. Someday users would end up merging or closing all changes and then there won't be more changes with hashtags.
> 
> We don't necessarily need to do housekeeping of removing the hashtags.
> 
> Is that what you meant?

Yes, we search and extract from both sources (hashtags and commit-msg) but we only write the `sourceBranch` to the new location inside the commit message.

We don't lost/forget old changes (even closed ones) and accidentally reopen them but all new created changes have the new location for their sourceBranch.

lib/modules/platform/gerrit/scm.ts
felipecrs: @NiasSt90 what's your opinion about this?

felipecrs:
Alright. I can work on this. @rarkins no objections I suppose?

felipecrs:
Done:

image

Here is an example of a change that was "migrated". Note that the hashtag was not removed.

felipecrs:
Note that I updated the documentation to mention that support for hashtags will be removed once support for Gerrit is graduated from experimental. I don't like technical debts, and I think that's a fair compromise.

lib/modules/platform/gerrit/scm.ts
felipecrs: For now I renamed to `Renovate-Branch`.

lib/modules/platform/gerrit/readme.md
```diff
@@ -3,11 +3,12 @@
 ## Supported Gerrit versions
 
 Renovate supports all Gerrit 3.x versions.
+
 Support for Gerrit is currently _experimental_, meaning that it _might_ still have some undiscovered bugs or design limitations, and that we _might_ need to change functionality in a non-backwards compatible manner in a non-major release.
 
-The current implementation uses Gerrit's "hashtags" feature.
-Therefore you must use a Gerrit version that uses the [NoteDB](https://gerrit-review.googlesource.com/Documentation/note-db.html) backend.
-We did not test Gerrit `2.x` with NoteDB (only in `2.15` and `2.16`), but could work.
+The current implementation uses the _commit message footer_ to store the Renovate metadata.
```
HonkingGoose, line 9:
```suggestion
Renovate stores its metadata in the _commit message footer_.
```

Let's simplify.

lib/modules/platform/gerrit/readme.md
```diff
@@ -3,11 +3,12 @@
...
+
+Previous implementations used to store such metadata using Gerrit's _hashtags_. For backwards compatibility, Renovate will still read metadata from hashtags if it finds them, but it will always write metadata to the commit message footer. Once Gerrit support is taken out of experimental status, Renovate will stop reading metadata from hashtags. This means changes without metadata in the commit message footer will be "forgotten" by Renovate.
```
HonkingGoose, line 11:
```suggestion
Previously Renovate stored metadata in Gerrit's _hashtags_.
To keep backwards compatibility, Renovate still reads metadata from hashtags.
But Renovate _always_ puts its metadata in the _commit message footer_!
When the Renovate maintainers mark Gerrit support as stable, the maintainers will remove the "read metadata from hashtags" feature.
This means changes without metadata in the commit message footer will be "forgotten" by Renovate.
```

Let's put each sentence on its own line in the source code. In the published docs it will show as one paragraph.

And simplify and clarify.

HonkingGoose (changes requested):

HonkingGoose (approved):
The docs look good to me.

lib/modules/platform/gerrit/scm.spec.ts
```diff
@@ -300,13 +300,20 @@ describe('modules/platform/gerrit/scm', () => {
         baseBranch: 'main',
         branchName: 'renovate/dependency-1.x',
         files: [],
-        message: ['commit msg', expect.stringMatching(/Change-Id: I.{32}/)],
+        message: [
+          'commit msg',
+          expect.stringMatching(
+            new RegExp(
+              '^Renovate-Branch: renovate/dependency-1\\.x\\nChange-Id: I[a-z0-9]{40}$',
+            ),
```
viceice, line 308:
```suggestion
            /^Renovate-Branch: renovate/dependency-1\\.x\\nChange-Id: I[a-z0-9]{40}$/,
```

simplify

lib/modules/platform/gerrit/types.ts
```diff
@@ -34,7 +34,7 @@ export type GerritReviewersType = 'REVIEWER' | 'CC' | 'REMOVED';
 
 export interface GerritChange {
   branch: string;
-  hashtags?: string[];
+  hashtags?: string[]; // for backwards compatibility
```
viceice, line 37:
```suggestion
  /**
   * for backwards compatibility
   */
  hashtags?: string[];
```

Use tsdoc comment style

lib/modules/platform/gerrit/utils.ts
```diff
@@ -90,9 +90,22 @@ export function mapGerritChangeStateToPrState(
...
 export function extractSourceBranch(change: GerritChange): string | undefined {
-  return change.hashtags
-    ?.find((tag) => tag.startsWith('sourceBranch-'))
-    ?.replace('sourceBranch-', '');
+  let sourceBranch: string | undefined = undefined;
+
+  if (change.current_revision) {
+    sourceBranch = change.revisions[
+      change.current_revision
+    ]?.commit?.message?.match(/^Renovate-Branch: (.+)$/m)?.[1];
```
viceice, line 98:
```suggestion
    const re = regEx(/^Renovate-Branch: (.+)$/m);
    sourceBranch = re.exec(change.revisions[
      change.current_revision
    ]?.commit?.message ?? '')?.[1];
```

use regex helper

rarkins:
@NiasSt90 can you review/approve?

NiasSt90:
> @NiasSt90 can you review/approve?

Yes, but not until Monday

viceice (dismissed):

felipecrs:
@NiasSt90 just a friendly reminder. :)

felipecrs:
FWIIW I have been self-hosting this PR since its creation, no issues.

lib/modules/platform/gerrit/client.ts
```diff
@@ -231,8 +231,15 @@ class GerritClient {
...
     const filters = ['owner:self', 'project:' + repository, filterState];
-    if (searchConfig.branchName !== '') {
-      filters.push(`hashtag:sourceBranch-${searchConfig.branchName}`);
+    if (searchConfig.branchName) {
+      filters.push(
+        ...[
+          `footer:Renovate-Branch=${searchConfig.branchName}`,
+          // for backwards compatibility
+          'OR',
+          `hashtag:sourceBranch-${searchConfig.branchName}`,
```
NiasSt90, line 240: This will not work correctly without braces around the OR expression.
All filters after this OR will be optional......

NiasSt90 (changes requested):

lib/modules/platform/gerrit/client.ts
felipecrs: Nice catch!

felipecrs: I tried reproducing it since I didn't notice anything wrong with it during my self-hosting, and indeed all filters after the OR ends up optional unless I add the braces.

felipecrs: Thanks a lot, @NiasSt90!

NiasSt90 (approved):

lib/modules/platform/gerrit/client.ts
NiasSt90: > I tried reproducing it since I didn't notice anything wrong with it during my self-hosting, and indeed all filters after the OR ends up optional unless I add the braces.

then you don't have autoapprove/automerge enabled... ;)

rarkins:
Needs 7 conversations resolved

felipecrs:
@rarkins, they are all resolved to me:

image

rarkins (approved):

PR #29823, opened 2024-06-24
fix(datasource/go): support go proxy with abortOnError

ahippler:
## Changes

Do not fail goproxy version lookup with `abortOnError: true`.

## Context

The goproxy registry api only support listing versions per major version. Renovate tries to fetch the versions for the next major versions to be able to detect major version updated. This is repeated until the major version does not exist and the last requests fails with a 404. With abortOnError enabled the whole lookup throws an error for the last request. The logic to ignore the error does not handle the wrapper error.


lib/modules/datasource/go/releases-goproxy.spec.ts
```diff
@@ -195,46 +205,50 @@ describe('modules/datasource/go/releases-goproxy', () => {
...
-        .get('/@v/list')
-        .replyWithError('unknown');
-
-      httpMock
-        .scope(`${baseUrl}/github.com/google/btree`)
-        .get('/@v/list')
-        .reply(
-          200,
-          codeBlock`
+    it.each([true, false])(
```
viceice, line 208: please use string template syntax

lib/modules/datasource/go/releases-goproxy.spec.ts
```diff
@@ -406,53 +420,57 @@ describe('modules/datasource/go/releases-goproxy', () => {
...
-    it('handles major releases', async () => {
-      process.env.GOPROXY = baseUrl;
-
-      httpMock
-        .scope(`${baseUrl}/github.com/google/btree`)
-        .get('/@v/list')
-        .reply(
-          200,
-          codeBlock`
+    it.each([true, false])(
```
viceice, line 423: same here

lib/modules/datasource/go/releases-goproxy.spec.ts
ahippler: maybe its better readable with multiple columns but you loose the type support

lib/modules/datasource/go/releases-goproxy.spec.ts
```diff
@@ -195,46 +205,54 @@ describe('modules/datasource/go/releases-goproxy', () => {
...
-        .get('/@v/list')
-        .replyWithError('unknown');
-
-      httpMock
-        .scope(`${baseUrl}/github.com/google/btree`)
-        .get('/@v/list')
-        .reply(
-          200,
-          codeBlock`
+    it.each`
```
viceice, line 208: you can still have type support

https://github.com/renovatebot/renovate/blob/d65effaa77b53ef93312474e81fdbd0efeff0498/lib/modules/datasource/nuget/common.spec.ts#L4

viceice (approved):

PR #29839, opened 2024-06-25
fix(pypi): support additional file name extensions

rarkins:
## Changes

Improves PyPI simple logic to support additional archive file formats

### DIFF
--- a/lib/modules/datasource/pypi/index.ts
+++ b/lib/modules/datasource/pypi/index.ts
@@ -182,9 +182,21 @@ export class PypiDatasource extends Datasource {
     // source packages
     const srcText = PypiDatasource.normalizeName(text);
     const srcPrefix = `${packageName}-`;
-    const srcSuffix = '.tar.gz';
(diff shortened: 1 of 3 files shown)


lib/modules/datasource/pypi/index.ts
```diff
@@ -182,9 +182,21 @@ export class PypiDatasource extends Datasource {
...
+      let res = srcText.replace(srcPrefix, '');
+      for (const suffix of srcSuffixes) {
+        if (res.endsWith(suffix)) {
+          // strip off the suffix using character length
+          res = res.slice(0, -suffix.length);
+          // strip off only one prefix
+          break;
+        }
+      }
+      return res;
```
secustor, line 199:
```suggestion
      const res = srcText.replace(srcPrefix, '');
      for (const suffix of srcSuffixes) {
        if (res.endsWith(suffix)) {
          // strip off the suffix using character length
          return res.slice(0, -suffix.length);
        }
      }
```
Simplified

secustor (approved):

PR #29847, opened 2024-06-25
fix: Only use GitHub releases to update Bazelisk

mark-thm:
## Changes

Change Bazelisk to use GitHub Releases datasource instead of GitHub Tags datasource.

## Context

Bazel's release process is to first create a tag, and then only to create a GitHub release once all artifacts have been built and are ready/available. The GitHub tag will exist before the real Bazel release artifacts are available, which causes Renovate to create PRs that may fail or only partially succed.


rarkins (changes requested):
I think the extract function also need changing

rarkins (approved):

PR #29860, opened 2024-06-26
refactor(cache): use `rm.content` to remove expired content

### LINKED ISSUE #29795: Cleaning file-based package cache is slow, research if/when we need `verify` step
### Intro

Discussion #29767 describes how this PR may have a regression:

- #28275

The PR appears to cause a major slowdown on self-hosted users, especially for those using long living containers (CE/EE).

### Reproduction

- [Reproduction repository](https://github.com/ladzaretti/package_cache_npm)

### Quotes from the discussion:

The bug reporter said:

> This performance hit is pretty consistent for all scan. I even have an example where cleaning the file-based package cache took almost 40 seconds for 1200~ cache entries, whereas prior to the fix, the time taken was much more reasonable.
> 
> Now this issue becomes even bigger for long living containers such as Renovate CE/EE.
> One workaround is to use the redis/sqlite cache.
> 
> Main question is: is this considered a bug? should we change the default package cache to sqlite based? Any other suggestions?

`@rarkins` responded:

> [#28169 (comment)](https://github.com/renovatebot/renovate/discussions/28169#discussioncomment-9220529) could be relevant.
> 
> If we `rm.content` then we shouldn't need to do a `verify`. It might be the `verify` which takes all the time

### Todos:

- [x] Confirm the `verify` step is slower now
- [x] Research if we _need_ to run a `verify` after `rm.content`?
- [x] Research if we still want to do a `verify` from time to time to stop the cache growing continuously?
- [x] Open PR to speed up behavior, or close issue if there's no good solution.

RahulGautamSingh:
## Changes
- Use the combination of `rm.content` and `rm.entry` to remove the content from the disk and the entry from the index
- Remove `cacache.verify` as it took longer than desriable to confirm that content is removed
> we give up the guarantee that we clear any and all unwanted entries from the index/disk 
> but we do delete the expired content from the disk with certainity, which was the original intention I believe

## Context
Closes: https://github.com/renovatebot/renovate/issues/29795


rarkins:
I tested this locally and got:

```
DEBUG: Checking file package cache for expired items
DEBUG: Deleted 113 of 192 file cached entries in 302ms
```

I don't have a "before" to compare it to

viceice (approved):

PR #29950, opened 2024-06-30
feat(manager): add mise package manager

james0209:
## Changes

Add support for the `mise` package manager: https://mise.jdx.dev/getting-started.html

`mise` has it's own TOML file - `.mise.toml` - which is an extension of `asdf`'s `.tool-versions`

## Context

Closes https://github.com/renovatebot/renovate/issues/28918


lib/modules/manager/mise/__fixtures__/Mise.1.toml
rarkins, line 1: Please inline these small fixtures using `codeBlock`

rarkins (changes requested):

lib/modules/manager/mise/extract.ts
```diff
@@ -0,0 +1,103 @@
...
+  miseTooling,
+} from './upgradeable-tooling';
+
+export function extractPackageFile(content: string): PackageFileContent | null {
+  logger.trace(`mise.extractPackageFile()`);
+
+  let misefile: MiseFile;
+
+  try {
+    misefile = parseToml(content) as MiseFile;
```
secustor, line 18: Use a zod schema instead of `as`

lib/modules/manager/mise/extract.ts
```diff
@@ -0,0 +1,103 @@
...
+    typeof toolData === 'object' &&
+    'version' in toolData &&
+    typeof toolData.version === 'string'
+  ) {
+    // Handle the object case with a string version
+    // e.g. 'python = { version = "3.11.2" }'
+    return toolData.version;
+  } else if (Array.isArray(toolData)) {
+    // Handle the array case
+    // e.g. 'erlang = ["23.3", "24.0"]'
```
secustor, line 57: What does it mean, if multiple tool versions are defined here?

lib/modules/manager/mise/extract.ts
```diff
@@ -0,0 +1,103 @@
...
+    return toolData[0]; // Get the first version in the array
+  }
+  return undefined; // Return undefined if no version is found
```
secustor, line 60: use `null`, if manually set.

lib/modules/manager/mise/extract.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+function getToolConfig(
+  name: string,
+  version: string | undefined,
+): ToolingConfig | undefined {
```
secustor, line 66:
```suggestion
): ToolingConfig | null {
```
Here too.

secustor (changes requested):

lib/modules/manager/mise/extract.ts
james0209: I believe it follows the same logic as `asdf`'s `.tool-versions` ([ref](https://asdf-vm.com/manage/configuration.html#tool-versions))

> Multiple versions can be set by separating them with a space. For example, to use Python 3.7.2, fallback to Python 2.7.15 and finally to the system Python, the following line can be added to .tool-versions.
> 
> 
> `python 3.7.2 2.7.15 system`

Example in mise documentation [here](https://mise.jdx.dev/configuration.html#mise-toml)

```
[tools]
# specify single or multiple versions
terraform = '1.0.0'
erlang = ['23.3', '24.0']

# supports everything you can do with .tool-versions currently
node = ['16', 'prefix:20', 'ref:master', 'path:~/.nodes/14']
```

The logic used for `asdf` is to only update the first version if multiple are listed, as the rest are fallbacks ([ref](https://github.com/renovatebot/renovate/blob/f19afb4dee5e5bf5b7ef6a67d6d817f2d38f6773/lib/modules/manager/asdf/readme.md?plain=1#L11)) - so I've used the same logic in this PR.

lib/modules/manager/mise/index.ts
```diff
@@ -0,0 +1,12 @@
+import { supportedDatasources as asdfSupportedDatasources } from '../asdf/index';
```
viceice, line 1:
```suggestion
import { supportedDatasources as asdfSupportedDatasources } from '../asdf';
```

lib/modules/manager/mise/upgradeable-tooling.ts
```diff
@@ -0,0 +1,148 @@
...
+    misePluginUrl: 'https://mise.jdx.dev/lang/go.html',
+    config: {
+      packageName: 'golang/go',
+      datasource: GithubTagsDatasource.id,
+      extractVersion: '^go(?<version>\\S+)',
+    },
+  },
+  java: {
+    misePluginUrl: 'https://mise.jdx.dev/lang/java.html',
+    config: (version) => {
```
viceice, line 53: can we reuse the logic from asdf?

lib/modules/manager/mise/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { logger } from '../../../logger';
+import { MiseFileSchema, MiseFileSchemaToml } from './schema';
+
+export function parseTomlFile(content: string): MiseFileSchema | null {
+  const res = MiseFileSchemaToml.safeParse(content);
+  if (res.success) {
+    return res.data;
+  } else {
+    logger.debug({ err: res.error }, 'Error parsing Mise file.');
```
viceice, line 9: pass the packafilename and log it here

lib/modules/manager/mise/upgradeable-tooling.ts
james0209: Theoretically yes, but there's some issues that I think makes it not worthwhile.

The `asdf` Java plugin does not support all the same types of Java as the `mise` one. i.e. in the `asdf` manager logic, there is no pathways for Zulu Java (`zulu-`), Corretto Java (`corretto-`), or even just a version number (i.e. `java = "17.0.2"`)

- This becomes an issue as `mise` supports having no prefix - defaulting to OpenJDK for no prefix.
- There are plenty of example of these different types of Java / no prefix being used in `.mise.toml` files. You can see on Github Search [here](https://github.com/search?q=mise.toml+java+language%3ATOML&type=code&l=TOML).

lib/modules/manager/mise/extract.ts
```diff
@@ -0,0 +1,108 @@
...
+    typeof toolData.version === 'string'
+  ) {
+    // Handle the object case with a string version
+    // e.g. 'python = { version = "3.11.2" }'
+    return toolData.version;
+  } else if (Array.isArray(toolData)) {
+    // Handle the array case
+    // e.g. 'erlang = ["23.3", "24.0"]'
+    return toolData[0]; // Get the first version in the array
+  }
```
secustor, line 56:
```suggestion
  if (is.nonEmptyString(toolData)) {
    // Handle the string case
    // e.g. 'erlang = "23.3"'
    return toolData;
  }
    if (is.array(toolData, is.string)) {
    // Handle the array case
    // e.g. 'erlang = ["23.3", "24.0"]'
    return toolData[0]; // Get the first version in the array
  }
  if (is.nonEmptyString(toolData.version)) {
    // Handle the object case with a string version
    // e.g. 'python = { version = "3.11.2" }'
    return toolData.version;
  } 
```
Simplify

secustor (changes requested):

lib/modules/manager/mise/extract.ts
james0209: Nice suggestion, thanks!

lib/modules/manager/mise/extract.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (is.object(toolData)) {
+    // Handle the object case with a string version
+    // e.g. 'python = { version = "3.11.2" }'
+    if (is.string(toolData.version)) {
+      return toolData.version;
+    }
+  }
```
secustor, line 57:
```suggestion
  if (is.nonEmptyString(toolData.version)) {
    // Handle the object case with a string version
    // e.g. 'python = { version = "3.11.2" }'
    return toolData.version;
  }
```
Because of type narrowing we already know that it is a object and based on your implementation this should return null for empty string

lib/modules/manager/mise/extract.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+          },
+        ],
+      });
+    });
+
+    it('complete example with skip', () => {
+      const content = codeBlock`
+      [tools]
+      java = '21.0.2'
+      erlang = ['23.3', '24.0']
```
secustor, line 174:
```suggestion
      erlang = []
```
Add a test that empty arrays are handled gracefully.

lib/modules/manager/mise/extract.ts
```diff
@@ -0,0 +1,109 @@
...
+function parseVersion(toolData: MiseToolSchema): string | null {
+  if (is.nonEmptyString(toolData)) {
+    // Handle the string case
+    // e.g. 'erlang = "23.3"'
+    return toolData;
+  }
+  if (is.array(toolData, is.string)) {
+    // Handle the array case
+    // e.g. 'erlang = ["23.3", "24.0"]'
+    return toolData[0]; // Get the first version in the array
```
secustor, line 49:
```suggestion
    return toolData.length ? toolData[0] : null; // Get the first version in the array
```

secustor (changes requested):

lib/modules/manager/mise/readme.md
```diff
@@ -0,0 +1,13 @@
...
+Therefore, specific tool support must be added one-by-one.
+
+The following tools are currently supported:
+
+<!-- Autogenerate in https://github.com/renovatebot/renovate -->
+<!-- Autogenerate end -->
+
+<!-- prettier-ignore -->
+!!! note
+    Only the first version entry for each supported tool is managed, due to `.mise.toml` supporting fallback versions.
```
HonkingGoose, line 13:
```suggestion
Renovate's `mise` manager can version these tools:

<!-- Autogenerate in https://github.com/renovatebot/renovate -->
<!-- Autogenerate end -->

<!-- prettier-ignore -->
!!! warning
    Only the _first_ version entry for each supported tool is managed!
    This is due to `.mise.toml` supporting fallback versions.

To understand the versioning, Renovate re-uses:

- [mise's plugins](https://github.com/jdx/mise/tree/main/src/plugins/core)
- and [asdf's plugins](https://mise.jdx.dev/registry.html)

This also means that support for a new tool's versioning must be added one-by-one.
```

Changes:

- Put the important info for the reader up front, so the list of supported tools goes on top
- Bump the alert level of the admonition to a `warning` to grab attention better
- Make a list of the plugins Renovate re-uses
- Write short sentences
- Use simple words and phrases
- Split long sentences into two

HonkingGoose (changes requested):

lib/modules/manager/mise/readme.md
```diff
@@ -0,0 +1,18 @@
...
+
+Renovate's `mise` manager can version these tools:
+
+<!-- Autogenerate in https://github.com/renovatebot/renovate -->
+<!-- Autogenerate end -->
+
+<!-- prettier-ignore -->
+!!! warning
+    Only the _first_ version entry for each supported tool is managed!
+    This is due to `.mise.toml` supporting fallback versions.
```
HonkingGoose, line 11: It feels we're missing a step in the explanation. Here's what I think when I read the current message:

1. Okay, so only the _first_ version for each tool is managed.
2. Looks like `.mise.toml` fallback versions cause this behavior.
3. But "why/how" does the fallback version "break" the way Renovate "sees/upgrades" versions?

Can you please improve the explanation, so the logical leap from "fallback version" to "broken Renovate" is clear? :wink:

lib/modules/manager/mise/readme.md
```diff
@@ -0,0 +1,21 @@
...
+<!-- Autogenerate in https://github.com/renovatebot/renovate -->
+<!-- Autogenerate end -->
+
+<!-- prettier-ignore -->
+!!! warning
+    Only the _first_ version entry for each supported tool is managed by Renovate. This limitation arises from the way `.mise.toml` is designed to handle fallback versions.
+    Specifically, when multiple versions are specified for a tool, `.mise.toml` allows for fallback to secondary versions if the primary version is unavailable.
+    Example: 'erlang = ["23.3", "22.0"]'
+    Fallback versions provide a safety net in case the primary version is not accessible - the first version provided is the primary version and should be used where possible.
+    Renovate will only update the primary version - it is the users' responsibility to maintain functioning fallback versions.
```
HonkingGoose, line 14: Thank you for the extra explanation! :heart: The text is now too big for a admonition. So let's give the text some space, and nice headings. I also attempted another rewrite, to use simpler words and phrases. Feel free to correct any technical errors. :wink:

Do you think we should recommend users only use a _single_ version? Or will that cause more problems?

Here's my suggested text, for you to copy/paste/edit:

```markdown
### Renovate only updates primary versions

Renovate only updates the _first_ version it finds for each tool listed in the `mise.toml` file.
You must manually update any fallback versions!

For example if you had this snippet in your `mise.toml` file:

`'erlang = ["23.3", "22.0"]'`

Then Renovate would update the first, primary version: `"23.3"` but ignore the fallback `"22.0"` version.

#### Why can Renovate only update primary versions?

Usually the mise package manager can access the primary version, and skips any fallback versions.

But when the mise package manager accesses a fallback version, Renovate gets confused.

To be safe, Renovate only updates the _first_ version it sees.

## Renovate re-uses these plugins
```

HonkingGoose (changes requested):

secustor (dismissed):
Waiting for a review `@HonkingGoose` before merging.

HonkingGoose (changes requested):
The docs are nearly good to go, some small style fixes.

Changes:

- Swap some difficult words for simple words
- Drop `.` at the end of list items
- Rewrite some sentences
- Use `monospace` styling for some things

By the way, we have a [Style Guide](https://github.com/renovatebot/renovate/blob/main/docs/development/style-guide.md) for user-facing text. That explains why I'm making some changes, that may look weird at first, like the `.` at the end of lists. :wink:

lib/modules/manager/mise/readme.md
```diff
@@ -6,16 +6,47 @@ Renovate's `mise` manager can version these tools:
...
-- and [asdf's plugins](https://mise.jdx.dev/registry.html)
+- [asdf's plugins](https://mise.jdx.dev/registry.html)
+
+for understanding and managing tool versioning.
+
+Support for new tools/plugins needs to be _manually_ added to Renovate's logic.
+
+#### Adding new tool support
+
+Adding support for a new tool involves either contributing to Mise Plugins or extending Renovate's asdf manager, which indirectly benefits mise management.
```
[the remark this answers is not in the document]
james0209: Rewritten it a lil' bit - mise uses asdf's plugin ecosystem under the hood, so adding support for `asdf-x` plugins is arguably one of the easiest ways to add support for tooling into `mise`.

`mise` does have core plugins which "are new and will improve over time. ([ref](https://mise.jdx.dev/plugins.html#core-plugins)), but I'm not sure about user contribution to them (i.e. it looks like it's only for main language versions rather than tools generically)

lib/modules/manager/mise/readme.md
```diff
@@ -0,0 +1,55 @@
...
+
+#### Why can Renovate only update primary versions?
+
+To maintain consistency and reliability, Renovate opts to only manage the _first_ listed version.
+
+- Fallback versions can often be older versions of a tool that are known to work and are there as a backup.
+
+This follows the same workflow that Renovate's `asdf` manager uses.
+
+### Plugin/Tool support
```
HonkingGoose, line 33:
```suggestion
### Plugin/tool support
```

lib/modules/manager/mise/readme.md
```diff
@@ -0,0 +1,55 @@
...
+Support for new tools/plugins needs to be _manually_ added to Renovate's logic.
+
+#### Adding new tool support
+
+There are 2 ways to integrate versioning for a new tool:
+
+- Renovate's `mise` manager: ensure upstream `mise` supports the tool, then add support to the `mise` manager in Renovate
+- Renovate's `asdf` manager: improve the `asdf` manager in Renovate, which automatically extends support to `mise`
+
+If `mise` adds support for more tools via it's own [core plugins](https://mise.jdx.dev/plugins.html#core-plugins), you can create a PR to extend Renovate's `mise` manager to add support for the new tooling.
```
HonkingGoose, line 51:
```suggestion
If `mise` adds support for more tools via its own [core plugins](https://mise.jdx.dev/plugins.html#core-plugins), you can create a PR to extend Renovate's `mise` manager to add support for the new tooling.
```

lib/modules/manager/mise/readme.md
```diff
@@ -0,0 +1,55 @@
...
+
+- You may be able to add support for new tooling upstream in the core plugins - create an issue and see if the community agrees whether it belongs there, or if it would be better as an `asdf-` plugin.
```
HonkingGoose, line 53:
```suggestion
You may be able to add support for new tooling upstream in the core plugins - create an issue and see if the community agrees whether it belongs there, or if it would be better as an `asdf-` plugin.
```

Let's drop the bullet point here, and keep it as plain text.

HonkingGoose (changes requested):
Final review, I promise! :wink:

lib/modules/manager/mise/readme.md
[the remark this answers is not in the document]
HonkingGoose: I'm happy with your text. I don't know enough to decide which method is best, so I'll let you judge what's best. :smile: 

Readers can now use the information in the docs to decide how they want to add support for new tools to Renovate.

HonkingGoose (approved):

secustor (approved):

james0209:
@rarkins can you give this PR some 👀 please? It can't be merged until you re-review it - thanks

rarkins (approved):

PR #29951, opened 2024-06-30
fix(git/auth): fix `insteadOf` for bitbucket-server

amezin:
## Changes

Fix git submodule update on Bitbucket Server/Data Center.

## Context

https://github.com/renovatebot/renovate/pull/29527 fixed dependency extraction, but update still fails - Renovate tries to clone submodules over SSH, because none of generated `url.*.instanceOf` matches.

### DIFF
--- a/lib/util/git/auth.spec.ts
+++ b/lib/util/git/auth.spec.ts
@@ -350,6 +350,27 @@ describe('util/git/auth', () => {
         GIT_CONFIG_VALUE_2: 'https://github.com:89/org/repo.git',
       });
     });
+
+    it('returns url with token for bitbucket-server', () => {
+      expect(
+        getGitAuthenticatedEnvironmentVariables('https://git.mycompany.com/', {
+          token: 'token1234',
+          hostType: 'bitbucket-server',
+          matchHost: 'git.mycompay.com',
+        }),
+      ).toStrictEqual({
+        GIT_CONFIG_COUNT: '3',
+        GIT_CONFIG_KEY_0:
+          'url.https://ssh:token1234@git.mycompany.com/scm/.insteadOf',
+        GIT_CONFIG_KEY_1:
+          'url.https://git:token1234@git.mycompany.com/scm/.insteadOf',
+        GIT_CONFIG_KEY_2:
+          'url.https://token1234@git.mycompany.com/scm/.insteadOf',
+        GIT_CONFIG_VALUE_0: 'ssh://git@git.mycompany.com:7999/',
+        GIT_CONFIG_VALUE_1: 'ssh://git@git.mycompany.com:7999/',
+        GIT_CONFIG_VALUE_2: 'https://git.mycompany.com/scm/',
+      });
+    });
   });
 
   describe('getGitEnvironmentVariables()', () => {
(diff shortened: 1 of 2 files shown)


amezin:
This works only if default ports are used.

A more correct solution is possible: almost all (if not all) platforms return both HTTP(S) and SSH git URLs in API. These URLs can be used to figure out base URLs for both protocols, and then perform translation (both in `url.*.insteadOf` and in `getHttpUrl()`). This would also handle the case when HTTP(S) repository access is disabled (possible according to comments in bitbucket-server module). But it'll be a non-trivial change, I'm not sure how exactly to implement it.

lib/util/git/auth.ts
```diff
@@ -115,33 +116,45 @@ function getAuthenticationRulesWithToken(
...
   gitUrl: string,
+  hostType: string | undefined | null,
   token: string,
 ): AuthenticationRule[] {
   const authenticationRules = [];
   const hasUser = token.split(':').length > 1;
   const insteadUrl = parseGitUrl(gitUrl);
 
+  if (hostType === 'bitbucket-server') {
+    insteadUrl.source = 'bitbucket-server';
```
rarkins, line 127: How is this field used?

amezin: https://github.com/IonicaBizau/git-url-parse/blob/master/lib/index.js#L304

amezin: Already had to set it in `getHttpUrl()` previously: https://github.com/renovatebot/renovate/blob/main/lib/util/git/url.ts#L47 (#29527)

lib/util/git/auth.ts
```diff
@@ -115,33 +116,45 @@ function getAuthenticationRulesWithToken(
...
+  }
+
   const url = { ...insteadUrl };
   const protocol = regEx(/^https?$/).test(url.protocol)
     ? url.protocol
     : 'https';
 
+  const sshUrl = { ...insteadUrl };
```
rarkins, line 135: What's the need/benefit of creating `sshUrl` instead of reusing `insteadUrl`?

amezin: Because `insteadUrl` is used (by the existing code) later to generate a `https://` url

rarkins: I think this function has now crossed the line of being too complex, due to `insteadUrl`, `url` and `sshUrl` all being similar but the relationships undocumented. Do you think a refactor can be done to reduce or simplify this, or we just need to document it?

amezin: What if I rewrite it like this? https://github.com/renovatebot/renovate/pull/29951/commits/cd401892d17f9ef71b6f72977abb90e9a6f47b46

rarkins: Yes, that's better. But please take the opportunity to document this logic so it's easier to maintain in future (including any confusing parts which were there already)

amezin: I'm unable to write any useful documentation because I don't fully understand this code. For example, is this function expected to take only https urls, or ssh urls too. On one hand, it seems to accept non-http on line 144. On the other hand the comment on line 166 suggests `protocol` is expected to be `https` and nothing else. Also from the surrounding code and tests it looks like only https is expected.

Plus half of the logic (including special cases/workarounds) is in git-url-parse, and half is in Renovate. For example, see [`buildToken`](https://github.com/IonicaBizau/git-url-parse/blob/28828546c148d58bbcff61409915a4e1e8f7eb11/lib/index.js#L293) in git-url-parse and `getAuthenticationRulesWithToken` here.

rarkins (approved):

lib/util/git/auth.spec.ts
```diff
@@ -350,6 +350,27 @@ describe('util/git/auth', () => {
         GIT_CONFIG_VALUE_2: 'https://github.com:89/org/repo.git',
       });
     });
```
alupula-vk, line 352: get data info and report bugs

Tripp1984: Ian

Tripp1984: > Iam

PR #29955, opened 2024-07-01
feat(templates): allow `upgrades` in `commitMessage` and `prTitle`

amezin:
## Changes

Allow access to `upgrades` in `commitMessage` and `prTitle` templates.

## Context

https://github.com/renovatebot/renovate/discussions/29546


lib/workers/repository/updates/generate.ts
```diff
@@ -13,6 +13,33 @@ import { uniq } from '../../../util/uniq';
 import type { BranchConfig, BranchUpgradeConfig } from '../../types';
 import { CommitMessage } from '../model/commit-message';
 
+function mergeProxy(
```
rarkins, line 16: How come no existing merge logic such as mergeChildConfig() wasn't suitable?

amezin: `templateInput` is `upgrade` + `{ upgrades: ... }`. One of the "inputs" (`upgrade`) changes when `templateInput` already exists. So it's not enough to just merge once - either need a proxy, or have to merge multiple times, after any value in `upgrade` gets set/updated.

rarkins: At this point I feel that this extra complexity is not justified by the benefits. Let's go back to the Discussion instead

amezin: https://github.com/renovatebot/renovate/discussions/29546

But I guess the only alternative is "you can't do that"?

rarkins: That might be the case. I know it can be hard to hear, but at times we have to say "Sorry, the complexity you are suggesting is not justified by the benefit to one or a small number of users". Let's discuss there.

rarkins (dismissed):
Can you refactor this into a function first (with no functional changes), and then after apply this change?

amezin:
It seems that existing tests didn't verify `hasBaseBranches: true` and `separateMajorMinor: true` cases (there are such updates in the tests, but their generated title was always thrown away and not checked)

rarkins:
Can you link to the repository you tested against? I think we need to validate both the group and non-group cases

amezin:
> Can you link to the repository you tested against? I think we need to validate both the group and non-group cases

https://github.com/amezin/renovate-template-issues

rarkins:
Awaitinv v38

amezin:
Do you consider this to be a breaking change?

Should I switch the target branch to v38?

viceice (approved):

PR #29983, opened 2024-07-02
build(docker): use cross platform builds

viceice:
## Changes
- more verbose builds to see sqlite build errors
- fix docker file directive casing
- speedup arm64 builds

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -542,6 +542,11 @@ jobs:
         with:
           show-progress: false
 
+      - name: docker-config
+        uses: containerbase/internal-tools@0b4162e48b68b3521953c3c663376ca22a1f769b # v3.3.2
+        with:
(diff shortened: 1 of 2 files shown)


viceice:
@rarkins @secustor WDYT about this in general? Build needs now ~15min instead of ~3. I maybe can speed this up a little more.

We can also make the docker build a seperate job, so it don't block other jobs

rarkins:
I'm not sure I want to _support_ ARM. i.e. accept bug reports any time it's not working. Right now I don't have a personal or commercial reason to do so

viceice:
> I'm not sure I want to _support_ ARM. i.e. accept bug reports any time it's not working. Right now I don't have a personal or commercial reason to do so

OK. will drop arm test builds again

rarkins (approved):

Issue #9271, opened 2021-03-24
Post-upgrade modifications can be dropped


croachrose:
**What Renovate type, platform and version are you using?**

v24.17.2
self-hosted renovate in GitHub actions

**Describe the bug**

1. Use go vendor to upgrade a go dependency and copy to `vendor/`
2. `go vendor` deletes all BUILD.bazel files from `vendor/`
3. Our post-upgrade script regenerates build files
4. renovate ignores the undeleted files because they are now identical in content to how they were before the upgrade

This same thing happens if the script unmodifies a file from the upgrade

**Relevant debug logs**

**Have you created a minimal reproduction repository?**

Please read the [minimal reproductions documentation](https://github.com/renovatebot/renovate/blob/master/docs/development/minimal-reproductions.md) to learn how to make a good minimal reproduction repository.

- [ ] This is a really small bug, it does not need a reproduction (think small typo)
- [ ] I have provided a minimal reproduction repository
- [ ] I don't have time for that, but it happens in a public repository I have linked to
- [ ] I don't have time for that, and cannot share my private repository
- [ ] The nature of this bug means it's impossible to reproduce publicly

**Additional context**

I have a fork with a fix and test https://github.com/renovatebot/renovate/compare/master...evenco:patrick/reset-artifacts

HonkingGoose:
@croachrose Can you try a newer version of Renovate? Your version (24.17.2) is from 14 January 2021. Maybe the problem is fixed already in the latest release.

- Please copy and paste your logs (remove private stuff!), as we have no output from the logs to help us debug the problem. 
- Please tick the relevant box in the "Have you created a minimal reproduction repository?"

croachrose:
@HonkingGoose I have reproduced the issue on master

Here is a fix and test https://github.com/renovatebot/renovate/compare/master...evenco:patrick/reset-artifacts

HonkingGoose:
@rarkins Can you take a look at the fix and test that the original poster provided, and tell us what you need/want next?

1. Pull request with fix and test.
1. Reproduction repository.
1. Retry with newer Renovate version.

rarkins:
I'm not certain that the suggested fix is safe enough not to cause problems with other use cases. It needs a reproduction to start with.

croachrose:
@rarkins ok, I'll work on a minimal reproduction

croachrose:
https://github.com/croachrose/renovate9271

[...]

PR #30025, opened 2024-07-04
feat(gomod): Add `gomodSkipVendor` postUpdateOption

r0bobo:
## Changes

This PR adds a postUpdateOption `gomodSkipVendor` which disables automatic vendoring functionality in the `gomod` manager.

## Context

I'm creating this PR because I've had the issue described in:
- #9271

Files that are removed by `go mod vendor` but are then restored in a `postUpgradeTask` are not added to the final commit.
After finding [this comment](https://github.com/renovatebot/renovate/discussions/29107#discussioncomment-9459541) in a discussion about go vendoring trouble I took a stab at adding the suggested option.


rarkins (approved):

rarkins (changes requested):
Add a test for this skip case

lib/config/options/index.ts
```diff
@@ -2508,6 +2508,7 @@ const options: RenovateOptions[] = [
       'gomodTidy1.17',
       'gomodTidyE',
       'gomodUpdateImportPaths',
+      'gomodVendorSkip',
```
viceice, line 2511:
```suggestion
      'gomodSkipVendor',
```
sounds better to me

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -133,7 +133,9 @@ export async function updateArtifacts({
 
   const vendorDir = upath.join(goModDir, 'vendor/');
   const vendorModulesFileName = upath.join(vendorDir, 'modules.txt');
-  const useVendor = (await readLocalFile(vendorModulesFileName)) !== null;
+  const useVendor =
+    (await readLocalFile(vendorModulesFileName)) !== null &&
+    !config.postUpdateOptions?.includes('gomodVendorSkip');
```
viceice, line 138: invert condition

r0bobo: Do you mean as in first checking if the option is not set and then checking the file, or as in flipping the meaning of the variable to something like `skipVendor`?

viceice: yes, first checking the option. that should be faster and we don't need to read the file if we skip anyways

r0bobo: Right, I'll fix it.

r0bobo:
> Add a test for this skip case

I added a test that copies the `it('supports vendor directory update', ...)` case but passes the `gomodSkipVendor` as well. I started editing the expected output so that the diff would disappear, but I kinda feel like I don't know what I'm doing as I don't understand the testing framework that well. Do you have any pointers on how the test case should look?

rarkins (dismissed):

r0bobo:
I've fixed the test so it is now green.

rarkins (approved):

PR #30030, opened 2024-07-04
refactor(gomod): rename `depName`

jamietanna:
## Changes

Rename the `depName` variable, as in #29575 we will be adding replacements for `gomod` and so
we'll have both a `fromPackageName` and a `toPackageName`.

## Context

As part of a prefactor for #29575, we should make this change
independently, to make the changes in #29575 easier to review.

This is required as we will be adding replacements for `gomod` and so
we'll have both a `fromPackageName` and a `toPackageName`.

### DIFF
--- a/lib/modules/manager/gomod/update.ts
+++ b/lib/modules/manager/gomod/update.ts
@@ -17,16 +17,17 @@ export function updateDependency({
 }: UpdateDependencyConfig): string | null {
   try {
     logger.debug(`gomod.updateDependency: ${upgrade.newValue}`);
-    const { depName, depType, updateType } = upgrade;
+    const { depType, updateType } = upgrade;
+    const fromPackageName = upgrade.depName;
     if (updateType === 'replacement') {
       logger.warn('gomod manager does not support replacement updates yet');
       return null;
     }
     // istanbul ignore if: should never happen
-    if (!depName || !upgrade.managerData) {
+    if (!fromPackageName || !upgrade.managerData) {
       return null;
     }
-    const depNameNoVersion = getDepNameWithNoVersion(depName);
+    const depNameNoVersion = getDepNameWithNoVersion(fromPackageName);
     const lines = fileContent.split(newlineRegex);
     // istanbul ignore if: hard to test
     if (lines.length <= upgrade.managerData.lineNumber) {
@@ -40,7 +41,7 @@ export function updateDependency({
       !lineToChange.includes('rethinkdb/rethinkdb-go.v5')
     ) {
       logger.debug(
-        { lineToChange, depName },
+        { lineToChange, depName: fromPackageName },
         "go.mod current line doesn't contain dependency",
       );
       return null;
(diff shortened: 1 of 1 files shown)


lib/modules/manager/gomod/update.ts
```diff
@@ -17,16 +17,17 @@ export function updateDependency({
...
       logger.warn('gomod manager does not support replacement updates yet');
       return null;
     }
     // istanbul ignore if: should never happen
-    if (!depName || !upgrade.managerData) {
+    if (!fromPackageName || !upgrade.managerData) {
       return null;
     }
-    const depNameNoVersion = getDepNameWithNoVersion(depName);
+    const depNameNoVersion = getDepNameWithNoVersion(fromPackageName);
```
rarkins, line 30: Did you also want to rename this depNameNoVersion at the same time?

viceice (dismissed):

lib/modules/manager/gomod/update.ts
```diff
@@ -17,16 +17,17 @@ export function updateDependency({
 }: UpdateDependencyConfig): string | null {
   try {
     logger.debug(`gomod.updateDependency: ${upgrade.newValue}`);
-    const { depName, depType, updateType } = upgrade;
+    const { depType, updateType } = upgrade;
+    const fromPackageName = upgrade.depName;
```
rarkins, line 21: I have a little concern about readability here. packageName is a formal concept in Renovate, separate from depName. This use of packageName in fromPackageName is not really in line with how we use the packageName concept. Does a term like "currentDepName" fit instead?

Regarding your replacement one, do you think it would be semantically wrong to do something like `let depName = upgrade.depName` and then reassign it later if needed?

viceice: or use `fromDepName`?

jamietanna: I was sure that I'd copied this pattern from somewhere else in the codebase, but I now can't see it, so I must have been mistaken.

+1 to the use of `packageName` not being the same as what we're doing here, and happy to update it to `fromDepName`.

In terms of:

>  do you think it would be semantically wrong to do something like `let depName = upgrade.depName` and then reassign it later if needed?

My intent was to make it possible to surface the to/from data so you're aware i.e. in the logs that it's a replacement.

However, that's something we're not doing elsewhere, and so I'm happy to replace with a `let` :+1: 

In which case, I believe this PR is no longer needed, and we can do it all in #29575?

jamietanna:
I've re-opened as it appears that yes actually we do need to have the renaming of variables, as we do some pre-emptive checks i.e. https://github.com/jamietanna/renovate/blob/33a20bcc2a8ad9b35d63d1e99530eac776778bdc/lib/modules/manager/gomod/update.ts#L45 to make sure that the dependency is there, and if we reuse the `depName` then those checks will fail

rarkins (approved):

jamietanna:
Thank you! :bow:

PR #30058, opened 2024-07-05
feat(managers/npm): add entries with protocol prefix in temporary .yarnrc.yml file

secustor:
## Changes
Add entries with protocol prefix, it supplied to temporary`.yarnrc.yml`. 
I have done also some small refactoring. 

## Context
Yarn allows to pin to add the protocol prefix to the host, though does merge configuration if only the host is supplied. 
This is similar to our host rules logic. 

### DIFF
--- a/lib/modules/manager/npm/post-update/rules.ts
+++ b/lib/modules/manager/npm/post-update/rules.ts
@@ -32,7 +32,7 @@ export function processHostRules(): HostRulesResult {
 
     const uri = `//${matchedHost}/`;
     let cleanedUri = uri;
-    if (is.string(matchedHost) && isHttpUrl(matchedHost)) {
+    if (isHttpUrl(matchedHost)) {
       cleanedUri = matchedHost.replace(regEx(/^https?:/), '');
     }
 
@@ -74,11 +74,11 @@ export function processHostRules(): HostRulesResult {
     }
   }
 
-  const hasYarnRCNPMRegistries =
+  const hasYarnRcNpmRegistries =
     Object.keys(additionalYarnRcYml.npmRegistries).length > 0;
   return {
     additionalNpmrcContent,
-    additionalYarnRcYml: hasYarnRCNPMRegistries
+    additionalYarnRcYml: hasYarnRcNpmRegistries
       ? additionalYarnRcYml
       : undefined,
   };
--- a/lib/modules/manager/npm/post-update/rules.ts
+++ b/lib/modules/manager/npm/post-update/rules.ts
@@ -3,6 +3,7 @@ import * as hostRules from '../../../../util/host-rules';
 import { regEx } from '../../../../util/regex';
 import { toBase64 } from '../../../../util/string';
 import { isHttpUrl } from '../../../../util/url';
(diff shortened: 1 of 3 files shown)


lib/modules/manager/npm/post-update/rules.ts
```diff
@@ -3,51 +3,83 @@ import * as hostRules from '../../../../util/host-rules';
...
+    const matchedHost = hostRule.matchHost;
+    // Should never be needed as if we have a resolvedHost there has to be a matchHost
+    // istanbul ignore next
+    if (!matchedHost) {
+      continue;
+    }
+
+    const uri = `//${matchedHost}/`;
+    let cleanedUri = uri;
+    if (is.string(matchedHost) && isHttpUrl(matchedHost)) {
```
viceice, line 35: why the additional is.string check?

lib/modules/manager/npm/post-update/rules.ts
```diff
@@ -3,51 +3,83 @@ import * as hostRules from '../../../../util/host-rules';
...
+      const registries = {
+        npmAuthIdent: `${hostRule.username}:${hostRule.password}`,
+      };
+      additionalYarnRcYml.npmRegistries[cleanedUri] = registries;
+      additionalYarnRcYml.npmRegistries[uri] = registries;
     }
   }
-  return { additionalNpmrcContent, additionalYarnRcYml };
+
+  const hasYarnRCNPMRegistries =
```
viceice, line 77:
```suggestion
  const hasYarnRcNpmRegistries =
```
be consistent with interface

lib/modules/manager/npm/post-update/rules.ts
```diff
@@ -3,51 +3,83 @@ import * as hostRules from '../../../../util/host-rules';
...
+    Object.keys(additionalYarnRcYml.npmRegistries).length > 0;
+  return {
+    additionalNpmrcContent,
+    additionalYarnRcYml: hasYarnRCNPMRegistries
```
viceice, line 81:
```suggestion
    additionalYarnRcYml: hasYarnRcNpmRegistries
```

lib/modules/manager/npm/post-update/rules.ts
```diff
@@ -3,51 +3,83 @@ import * as hostRules from '../../../../util/host-rules';
...
+      const registry = {
+        npmAuthToken: hostRule.token,
+      };
+      additionalYarnRcYml.npmRegistries[cleanedUri] = registry;
+      additionalYarnRcYml.npmRegistries[uri] = registry;
+
+      continue;
+    }
+
+    if (is.string(hostRule.username) && is.string(hostRule.password)) {
```
viceice, line 62: use else if, so we don't need continue above? less code.

secustor: That's hurting readability as we get a lot of indentations wrapping here.

lib/modules/manager/npm/post-update/rules.ts
secustor: left over from before the falsely check has been added.

rarkins (approved):

PR #30066, opened 2024-07-06
fix(azure): getRawFile not handling 404 for Azure DevOps

sindrekroknes:
## Changes

Replaced the getItemContent call in getRawFile with getItem

## Context

When getting a file that doesn't exist, the getItemContent function in azure-devops-node-api will still return a stream with json content. But it's a json describing the error (that the file doesn't exist). When Renovate tries to read this json as a config file it fails with validation errors. 
This happens for repositories that are not onboarded, when repositoryCache is enabled. Renovate tries to read the renovate.json file from the default branch, where it doesn't exist yet, and because it gets a non-empty response it tries to use the error response as a config file.
(Not sure if it's a bug that Renovate tries to read the config file from the default branch when the repository hasn't been onboarded)

Switching to getItem will fetch more data, since it includes some metadata. But it makes error handling easier, since azure-devops-node-api then returns null for 404 and throws for errors. Instead of just returning the raw stream.

### DIFF
--- a/lib/modules/platform/azure/index.spec.ts
+++ b/lib/modules/platform/azure/index.spec.ts
@@ -1886,12 +1886,11 @@ describe('modules/platform/azure/index', () => {
     it('returns file content', async () => {
       const data = { foo: 'bar' };
       azureApi.gitApi.mockImplementationOnce(
-        () =>
-          ({
-            getItem: jest.fn(() =>
-              Promise.resolve({ content: JSON.stringify(data) }),
-            ),
-          }) as any,
+        jest.fn().mockImplementationOnce(() => ({
+          getItem: jest.fn(() =>
+            Promise.resolve({ content: JSON.stringify(data) }),
+          ),
+        })),
       );
       const res = await azure.getJsonFile('file.json');
       expect(res).toEqual(data);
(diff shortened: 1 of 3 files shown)


lib/modules/platform/azure/index.spec.ts
```diff
@@ -1888,15 +1888,26 @@ describe('modules/platform/azure/index', () => {
...
       const res = await azure.getJsonFile('file.json');
       expect(res).toEqual(data);
     });
 
+    it('returns null when file not found', async () => {
+      azureApi.gitApi.mockImplementationOnce(
+        () =>
+          ({
+            getItem: jest.fn(() => Promise.resolve(null)),
+          }) as any,
```
viceice, line 1905: please use our partial helper function instead of using the any cast.

lib/modules/platform/azure/index.ts
```diff
@@ -156,10 +155,10 @@ export async function getRawFile(
       undefined,
       undefined,
       branchOrTag ? versionDescriptor : undefined,
+      true,
```
viceice, line 158: what is this flag changing?

sindrekroknes: it's the "includeContent" parameter, for including the content when getting the full item.

viceice (approved):

Issue #7041, opened 2020-08-20
Add apt datasource for private repositories


micheelengronne:
**What would you like Renovate to be able to do?**

https://github.com/renovatebot/renovate/issues/3722 was closed because we can manage packages pining with public apt repositories throug the repology datasource. It does not work for private repositories and any repository that is not indexed by repology.

**Are there any workarounds or alternative ideas you've tried to avoid needing this feature?**

I create Docker layers with the apt packages installed in it to ensure immutability through the Docker datasource. But that is not an optimized solution at all.

HonkingGoose:
@rarkins I've tagged this issue with the labels I think it needs.

I noticed this feature needs more discussion before it can be coded, so I moved it into the **Needs requirements** column. Was that the right thing to do?

Can you tell me if I should move issues like this to the **Needs Requirements** column or not? Does moving a card into the Project board signify that the Renovate team is going to implement this? I don't want to add obligations on your end by accident. :smile:

rarkins:
There's no formality to worry about with the Project board and `Needs requirements` is the appropriate column until someone has marked it as otherwise

nejch:
Does anyone else here have a use case to manage the dependency versions of actual debian packages themselves, in `debian/control` files via Depends/Pre-Depends and so on? https://debian-handbook.info/browse/stable/sect.package-meta-information.html

nejch:
So I looked into this a bit more and I think some of the discussion in the previous issue (https://github.com/renovatebot/renovate/issues/3722) is still relevant:
1. renovate should not rely on any kind of [API](https://sources.debian.org/doc/api/), as it might not be available in other implementations like aptly, Artifactory, Nexus, and soon GitLab.
1. renovate should not try to run `apt list`, because it cannot know the architecture of the target host (e.g. a package may already be available for `amd64` but not yet for `armhf` or so).

IMO the best way forward here is:
1. define private repo(s) for the datasource with enough information to replicate an entry* in `sources.list`:
    1. `registryUrl` / uri
    1. distribution
    1. component(s)
    1. architecture (since renovate cannot know what arch the target host is running on)
1. renovate fetches the `Packages.gz` index for the given registry/dist/component/arch, and extracts the package versions, e.g.:
    - `${registryUrl}/dists/${dist}/${component}/binary-${arch}/Packages.gz` (arch-specific packages), and also
[...]

PR #30071, opened 2024-07-07
feat(datasource): add debian datasource

oxdev03:
## Changes

- Built upon @Ka0o0 implementation of the Debian datasource.
- Includes minor improvements.
- Addresses coverage issues identified in the previous review of the Debian datasource.
- Fixes earlier review findings (except the caching issue).

## Context

**Resolves**:  #7041, #24906 

### Concept

Based on the discussion and the reused implementation (see the related issue for more details):

1. Evaluation of Provided or Default Registry URL
    - Example: The URL `https://ftp.debian.org/debian?release=bullseye&components=main,contrib&binaryArch=amd64` is evaluated to:
        - `.../debian/dists/bullseye/main/binary-amd64`
        - `.../debian/dists/bullseye/contrib/binary-amd64`
    - Supports multiple component URLs.

2. Check for Existing Compressed Package
    - If the compressed `Packages.gz` file has already been downloaded and extracted:
        - Send a HEAD request to check if the upstream file has been updated.
    - If not already downloaded:
        - Download the compressed file.
        - Extract the contents.
        - Delete the compressed file.
    - Currently, only gzip compression is supported.

3. Iterate through Package Index
    - Gather relevant information for the requested package.
    - Stop iteration, if found

4. Return relevant release information

included documentation for new datasource

[Real Test](https://github.com/oxdev03/renovate-test/pull/1)

## Open Points

- Does the current caching implementation for the datasource need to be reworked? Can the Renovate team provide a concept/details for this? (e.g., invalidation of the compressed file based on its timestamp)

- Is the current approach generic enough? Are there any points that were not considered?

- If significant rework is required, please let me know in advance.


lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,440 @@
...
+  }
+
+  /**
+   * Constructs the component URLs from the given registry URL.
+   *
+   * @param registryUrl - The base URL of the registry.
+   * @returns An array of component URLs.
+   * @throws Will throw an error if required parameters are missing from the URL.
+   */
+  constructComponentUrls(registryUrl: string): string[] {
```
secustor, line 320: Please use functions and methods where possible. Also move functions into separate files.

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,440 @@
...
+
+    const validateUrlAndParams = (url: URL): void => {
+      REQUIRED_PARAMS.forEach((param) => {
+        if (!url.searchParams.has(param)) {
+          throw new Error(`Missing required query parameter '${param}'`);
+        }
+      });
+    };
+
+    const getReleaseParam = (url: URL): string => {
```
secustor, line 332: No local functions, please.

secustor (changes requested):
Thanks for your contribution, and the well documented functionality!

This is my first pass as there is a lot going on here and we have to be sure that the caching is setup efficiently. 

Would this work for every APT repository?

secustor:
> Does the current caching implementation for the datasource need to be reworked? Can the Renovate team provide a concept/details for this? (e.g., invalidation of the compressed file based on its timestamp)

We should start simple and then start to optimize. 
The `Package.gz` will be cached non transparently by the http layer, so the file will be downloaded only when necessary. 
I suggest to simply cache the extracted file on a time base ( extract it to a different function and use the annotation ).

oxdev03:
> This is my first pass as there is a lot going on here and we have to be sure that the caching is setup efficiently.

Thx for the initial review, I fix the findings soon.

> Would this work for every APT repository?

should work with every apt repository, which follows the https://wiki.debian.org/DebianRepository standard.

lib/modules/datasource/deb/index.spec.ts
```diff
@@ -0,0 +1,311 @@
...
+        debDatasource.checkIfModified(
+          getPackageUrl(debBaseUrl, 'stable', 'non-free', 'amd64'),
+          new Date(),
+        ),
+      ).resolves.toBe(true);
+    });
+  });
+});
+
+const getComponentUrl = (
```
viceice, line 291: use proper function declarations

lib/modules/datasource/deb/readme.md
```diff
@@ -0,0 +1,70 @@
...
+    gcc-11="${GCC_VERSION}" && \
+    apt-get clean
+```
+
+When the apt package for `gcc` is updated, Renovate updates the environment variable.
+
+<!-- prettier-ignore -->
+!!! tip
+    We recommend you try `loose` versioning for distribution packages first.
+    This is because the version number usually doesn't match Renovate's default `semver-coerced` specification.
```
viceice, line 70: No don't set any versioning, the datasource should set `deb` versioning as default, which should work in those cases.

lib/modules/datasource/repology/readme.md
```diff
@@ -33,7 +33,7 @@ First you would set a custom manager in your `renovate.json` file for `Dockerfil
 
 Then you would put comments in your Dockerfile, to tell Renovate where to find the updates:
 
-```docker
+```dockerfile
```
viceice, line 36: - #30097

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,347 @@
...
+      `${extractedFile}:${packageName}:${lastTimestamp.getTime()}`,
+    ttlMinutes: 24 * 60,
+  })
+  async parseExtractedPackage(
+    extractedFile: string,
+    packageName: string,
+    lastTimestamp: Date,
+  ): Promise<ReleaseResult | null> {
+    // read line by line to avoid high memory consumption as the extracted Packages
+    // files can be multiple MBs in size
```
secustor, line 263: How big are we talking here? 

Do you have considered capsuling componentUrl processing into a single function and cache that instead?
```typescript
export function processComponentUrl(url: string): Record<string, ReleaseResult> {
```
That way subsequent lookups of other packages do not result in additional requests to the registry.

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,347 @@
...
+   * - components: comma separated list of components
+   * - suite: stable, oldstable or other alias for a release, either this or release must be given
+   * - release: buster, etc.
+   * - binaryArch: e.g. amd64 resolves to http://ftp.debian.org/debian/dists/stable/non-free/binary-amd64/
+   */
+  override readonly defaultRegistryUrls = [
+    'https://ftp.debian.org/debian?suite=stable&components=main,contrib,non-free&binaryArch=amd64',
+  ];
+
+  override readonly caching = true;
```
secustor, line 58:
```suggestion
```
Standard caching will not work efficiently as the registryUrl as each parameter combination will result in a different cache key. This should be handled internally. See the componentURL caching idea. 

https://github.com/renovatebot/renovate/blob/283cda0f29d16715d3c4cc8129cdb3d67c89595d/lib/modules/datasource/index.ts#L66

secustor (changes requested):

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,347 @@
...
+
+  /**
+   * Not all Debian packages follow Semver, so it's wise to keep this loose but make sure to
+   * have enough tests in your application.
```
viceice, line 62: seems wrong, the value is `deb` versioning, not `loose`

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,347 @@
...
+    if (lastDownloadTimestamp) {
+      needsToDownload = await this.checkIfModified(
+        packageUrl,
+        lastDownloadTimestamp,
+      );
+    }
+
+    if (needsToDownload) {
+      try {
+        const readStream = this.http.stream(packageUrl);
```
viceice, line 194: we should also download the releases file and parse for checksums, so we can validate the packages file

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,347 @@
...
+              logger.warn(
+                { packageName },
+                'Package occurred in more than one repository with different meta information. Aggregating releases anyway.',
+              );
+            }
+            aggregatedRelease.releases.push(...newRelease.releases);
+          }
+        }
+      } catch (error) {
+        logger.warn(
```
viceice, line 338:
```suggestion
        logger.debug(
```
we usually don't warn here

lib/modules/datasource/deb/readme.md
```diff
@@ -0,0 +1,65 @@
...
+    }
+  ],
+  "packageRules": [
+    {
+      "matchDatasources": ["deb"],
+      "registryUrls": [
+        "https://ftp.debian.org/debian?suite=stable&components=main,contrib,non-free&binaryArch=amd64"
+      ]
+    }
+  ]
```
viceice, line 47: remove, this is already the default registry url and will be added to docs

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,347 @@
...
+          if (line.startsWith(`${key}:`)) {
+            currentPackage[key] = line.substring(key.length + 1).trim();
+            break;
+          }
+        }
+      }
+    }
+
+    // Check the last package after file reading is complete
+    if (currentPackage.Package === packageName) {
```
viceice, line 289: there can be multiple versions of a single package, we should extract all versions

lib/util/fs/index.ts
```diff
@@ -176,6 +176,11 @@ export function createCacheWriteStream(path: string): fs.WriteStream {
   return fs.createWriteStream(fullPath);
 }
 
+export function createCacheReadStream(path: string): fs.ReadStream {
```
viceice, line 179: please create a separate PR for this changes

lib/modules/datasource/deb/url.ts
```diff
@@ -0,0 +1,76 @@
...
+ */
+function getReleaseParam(url: URL, optionalParams: string[]): string {
+  for (const param of optionalParams) {
+    const paramValue = url.searchParams.get(param);
+    if (paramValue !== null) {
+      return paramValue;
+    }
+  }
+  throw new Error(
+    `Missing one of ${optionalParams.join(', ')} query parameter`,
```
viceice, line 74: why it's optional when you validate existence here? simply make them mandatory too

lib/util/fs/index.ts
oxdev03: done, #30600

lib/modules/datasource/deb/url.ts
oxdev03: Atleast one of the parameter is required 'release' or 'suite'

lib/modules/datasource/deb/index.ts
oxdev03: The [debian doc](https://wiki.debian.org/DebianRepository/Format#Version) only mentions their can be different Package for a source. 
e.g 
Package: 0ad-data-common
Source: 0ad-data

But I think the Package field is unique and only includes the latest version. I even checked the debian Package Index and only found single occurrence with their latest version.

lib/modules/datasource/deb/index.ts
oxdev03: > How big are we talking here?
It really depends on the used registry. With public registry like https://ftp.debian.org/debian/dists/bullseye/main/binary-amd64/ the compressed file is usually around 1-15MB. After extraction the size is around 10-100MB.

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,347 @@
...
+          await fs.rmCache(compressedFile);
+        }
+      }
+
+      if (!lastTimestamp) {
+        //extracting went wrong
+        break;
+      }
+
+      return { extractedFile, lastTimestamp };
```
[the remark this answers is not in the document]
oxdev03: good point, fixed it.
The idea was to use the next supported compression (incase the default first fails), this happens in line 150. But currently it only supports gz.

lib/modules/datasource/deb/index.ts
oxdev03: I updated the parsing of the extracted package to parse all packages instead, to prevent lookups for the same package index under different package names.

There is already a caching mechanism. If everything is already cached, such as from a previous run, it will only lead to an HTTP HEAD call to the registry to check whether the upstream has a newer package index.

It's not possible to use cache decorators for downloadPackageFile since there is no function to invalidate the cache.

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,387 @@
...
+      terminal: false,
+    });
+
+    let currentPackage: PackageDescription = {};
+    const allPackages: PackageDescription[] = [];
+
+    for await (const line of rl) {
+      if (line === '') {
+        // All information of the package are available, add to the list of packages
+        if (Object.keys(currentPackage).length > 0) {
```
secustor, line 311: Should this really return if there is any key? 

I would expect that this should only return, if there are all expected keys 
```suggestion
        if (!DebDatasource.requiredPackageKeys.some((key) => !(key in currentPackage))) {
```

secustor (changes requested):

lib/modules/datasource/deb/types.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * A package file contains multiple package descriptions which are each separated by an completely empty line.
+ * A package description contains meta data properties in the form of
+ *
+ * PropertyName: value
+ */
```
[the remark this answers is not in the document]
oxdev03: better?

lib/modules/datasource/deb/index.ts
oxdev03: Good point, didn't know about that usecase. I already applied your suggestion. Is their something missing for this?

oxdev03:
fyi I resolved some old threads.

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,396 @@
...
+        { url: packageUrl, targetFile: compressedFile },
+        'Downloading Debian package file',
+      );
+
+      const actualChecksum = await computeFileChecksum(compressedFile);
+
+      if (actualChecksum !== expectedChecksum) {
+        await fs.rmCache(compressedFile);
+        throw new Error('SHA256 checksum validation failed');
+      }
```
secustor, line 224:
```suggestion
      const readStream = this.http.stream(packageUrl);
      const writeStream = fs.createCacheWriteStream(compressedFile);
      await fs.pipeline(readStream, writeStream);
      logger.debug(
        { url: packageUrl, targetFile: compressedFile },
        'Downloading Debian package file',
      );

      const actualChecksum = await computeFileChecksum(compressedFile);

      try {
        const inReleaseContent = await this.fetchInReleaseFile(baseReleaseUrl);
        const expectedChecksum = parseChecksumsFromInRelease(
          inReleaseContent,
          // path to the Package.gz file
          packageUrl.replace(`${baseReleaseUrl}/`, ''),
        );
        if (actualChecksum !== expectedChecksum) {
          logger.debug(
            { url: baseReleaseUrl },
            'SHA256 checksum validation failed',
          );
          return false;
        }
      } catch (error) {
        // This is expected to fail for Artifactory if GPG verification is not enabled
        logger.debug({ url: baseReleaseUrl }, 'Could not fetch InRelease file');
      }
```
I have tested against my Artifactory instance and it seems like that one has no InRelease files if GPG verification is not enabled.

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,396 @@
...
+    const inReleaseUrl = `${baseReleaseUrl}/InRelease`;
+    try {
+      const response = await this.http.get(inReleaseUrl);
+      return response.body;
+    } catch (error) {
+      logger.error(
+        `Failed to download InRelease file from ${inReleaseUrl}: ${error.message}`,
+      );
+      throw new Error('InRelease download failed');
+    }
```
secustor, line 249:
```suggestion
    const inReleaseUrl = joinUrlParts(baseReleaseUrl, 'InRelease');
    const response = await this.http.get(inReleaseUrl);
    return response.body;
```
You are capturing the exception already outside of this function and you are logging again with less information

lib/modules/datasource/deb/index.ts
secustor: Caching should now work as expected. 

I have run it against my local dev Artifactory deb repo and it works now with the suggested changes.

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,396 @@
...
+   * @returns True if the file was downloaded, otherwise false.
+   */
+  async downloadPackageFile(
+    basePackageUrl: string,
+    compression: string,
+    compressedFile: string,
+    lastDownloadTimestamp?: Date,
+  ): Promise<boolean> {
+    const baseReleaseUrl = getBaseReleaseUrl(basePackageUrl);
+    const packageUrl = `${basePackageUrl}/Packages.${compression}`;
```
secustor, line 193:
```suggestion
    const packageUrl = joinUrlParts(basePackageUrl, `Packages.${compression}`);
```
This makes it more forgiving if the baseURL does include a trailing slash

secustor (changes requested):
We are getting close, I could confirm that it works locally and with the patches also with Artifactory repos

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,396 @@
...
+    let needsToDownload = true;
+
+    if (lastDownloadTimestamp) {
+      needsToDownload = await this.checkIfModified(
+        packageUrl,
+        lastDownloadTimestamp,
+      );
+    }
+
+    if (needsToDownload) {
```
secustor, line 203: This can also be made easier readable by using early returns

```suggestion

    if (!needsToDownload) {
      logger.debug(`No need to download ${packageUrl}, file is up to date.`);
      return false;
    }
```

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,396 @@
...
+      const actualChecksum = await computeFileChecksum(compressedFile);
+
+      if (actualChecksum !== expectedChecksum) {
+        await fs.rmCache(compressedFile);
+        throw new Error('SHA256 checksum validation failed');
+      }
+    } else {
+      logger.debug(`No need to download ${packageUrl}, file is up to date.`);
+    }
+
```
secustor, line 228:
```suggestion
```
This has been replaced by the early return

secustor (changes requested):

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,399 @@
...
+    if (inReleaseContent) {
+      const expectedChecksum = parseChecksumsFromInRelease(
+        inReleaseContent,
+        // path to the Package.gz file
+        packageUrl.replace(`${baseReleaseUrl}/`, ''),
+      );
+      if (actualChecksum !== expectedChecksum) {
+        await fs.rmCache(compressedFile);
+        throw new Error('SHA256 checksum validation failed');
```
oxdev03, line 235: @secustor your suggestion removed this, I added it back.

Otherwise if the checksum check fails and the function returns false , it will continue processing in downloadAndExtractPackage

with the exception, it will prevent this,  since the downloadPackageFile returns on errors

```
try {
        wasUpdated = await this.downloadPackageFile(
          componentUrl,
          compression,
          compressedFile,
          lastTimestamp,
        );
} catch (error) {
       logger.error(
          {
            compression,
            error: error.message,
          },
          `Failed to download package file from ${componentUrl}`,
        );

        continue;
}
```

secustor: There is a race condition in the datasource if different packages are looked up from the same index. 

In that case you are writing and deleting them at the same time the archives. 
Simply generate a random string to fix that. 

I will work on a locking mechanism to only run it one soon. 

```
         {
           "name": "renovate",
           "level": 50,
           "logContext": "T8GoJkZ_AFuxx0_pwD5_9",
           "repository": "secustor/renovate-deb-test",
           "compression": "gz",
           "error": "ENOENT: no such file or directory, open '/var/folders/sb/c6sv5dy16rjdlflwkf5q2kfr0000gn/T/renovate/cache/others/deb/296879bddc438f33b24e312f2e2fd99a34f41459d3d3c4237a9cfa4d5aa21f2c.gz'",
           "msg": "Failed to download package file from http://localhost:8082/artifactory/debian/dists/noble/restricted/binary-amd64"
         },
```

secustor (changes requested):

lib/modules/datasource/deb/index.ts
secustor: With https://github.com/renovatebot/renovate/pull/30815

the race condition should not occure anymore. Please create dynamic paths anyway

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,399 @@
...
+   * valid URLs which is why the parameters are encoded in the URL.
+   *
+   * The following query parameters are required:
+   * - components: comma separated list of components
+   * - suite: stable, oldstable or other alias for a release, either this or release must be given
+   * - release: buster, etc.
+   * - binaryArch: e.g. amd64 resolves to http://ftp.debian.org/debian/dists/stable/non-free/binary-amd64/
+   */
+  override readonly defaultRegistryUrls = [
+    'https://ftp.debian.org/debian?suite=stable&components=main,contrib,non-free&binaryArch=amd64',
```
Finkregh, line 57: In the end it might not make such a big difference, but https://ftp.debian.org/ tells you to use `deb.deian.org` :)

Thanks for your work so far!

lib/modules/datasource/deb/index.ts
oxdev03: Sure, will adjust it. Did you mean with dynamic path, unique cache path for every renovate lookup?

secustor: Yeah exactly. You can use the `nanoid` package generate an unique string.

oxdev03: thx, also added a testcase for this.

secustor (dismissed):

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,416 @@
...
+        }
+      } else {
+        for (const key of DebDatasource.packagesKeys) {
+          if (line.startsWith(`${key}:`)) {
+            currentPackage[key] = line.substring(key.length + 1).trim();
+            break;
+          }
+        }
+      }
+    }
```
oxdev03, line 342: With private Artifactory like JFrog multiple version of a package can exist

oxdev03:
> This will merge results from different registries. Is it really what you want here? This is only used if there is a a second registry which has more metadata.

The `registryStrategy=merge` is required for the following use case: when there are multiple apt packages, some of which are only included in the remote repository and others in the internal Artifactory. With the current registryStrategy (first), it would require two separate configurations and additional labeling of internal dependencies

lib/modules/datasource/deb/release.ts
```diff
@@ -18,14 +18,14 @@ export function releaseMetaInformationMatches(
...
  */
 export function formatReleaseResult(
-  packageDesc: PackageDescription,
+  packagesDesc: PackageDescription[],
 ): ReleaseResult {
   return {
-    releases: [{ version: packageDesc.Version! }],
-    homepage: packageDesc.Homepage,
+    releases: packagesDesc.map((p) => ({ version: p.Version! })),
+    homepage: packagesDesc[0].Homepage,
```
secustor, line 29:
```suggestion
    releases: packagesDesc.map((p) => ({ version: p.Version! })),
    homepage: packagesDesc[0]?.Homepage,
```
This will else throw if the array is empty

secustor (dismissed):

oxdev03:
> > This will merge results from different registries. Is it really what you want here? This is only used if there is a a second registry which has more metadata.
> 
> The `registryStrategy=merge` is required for the following use case: when there are multiple apt packages, some of which are only included in the remote repository and others in the internal Artifactory. With the current registryStrategy (first), it would require two separate configurations and additional labeling of internal dependencies

Can we add this though or something speaking against it?

secustor:
> > > This will merge results from different registries. Is it really what you want here? This is only used if there is a a second registry which has more metadata.
> > 
> > 
> > The `registryStrategy=merge` is required for the following use case: when there are multiple apt packages, some of which are only included in the remote repository and others in the internal Artifactory. With the current registryStrategy (first), it would require two separate configurations and additional labeling of internal dependencies
> 
> Can we add this though or something speaking against 

Add it, the only downside is for scenarios in which the metadata conflicts. 
Though in that case we are not extracting much of them.

secustor (dismissed):

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,423 @@
...
+   * @param componentUrl - The URL of the component.
+   * @returns The path to the extracted file and the last modification timestamp.
+   * @throws Will throw an error if no valid compression method is found.
+   */
+  async downloadAndExtractPackage(
+    componentUrl: string,
+  ): Promise<{ extractedFile: string; lastTimestamp: Date }> {
+    const packageUrlHash = createHash('sha256')
+      .update(componentUrl)
+      .digest('hex');
```
viceice, line 125:
```suggestion
    const packageUrlHash = toSha256(componentUrl);
```

use our utils

https://github.com/renovatebot/renovate/blob/302718f38e54c6c7590f488d749fb4da8362f36f/lib/util/hash.ts#L19

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,423 @@
...
+  /**
+   * This is just an internal list of compressions that are supported and tried to be downloaded from the remote
+   */
+  static readonly compressions = ['gz'];
+
+  /**
+   * This specifies the directory where the extracted and downloaded packages files are stored relative to cacheDir.
+   * The folder will be created automatically if it doesn't exist.
+   */
+  static readonly cacheSubDir: string = 'deb';
```
viceice, line 34: move to utils

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,423 @@
...
+    componentUrl: string,
+  ): Promise<{ extractedFile: string; lastTimestamp: Date }> {
+    const packageUrlHash = createHash('sha256')
+      .update(componentUrl)
+      .digest('hex');
+    const fullCacheDir = await fs.ensureCacheDir(DebDatasource.cacheSubDir);
+    const extractedFile = upath.join(fullCacheDir, `${packageUrlHash}.txt`);
+    let lastTimestamp = await this.getFileCreationTime(extractedFile);
+
+    for (const compression of DebDatasource.compressions) {
```
viceice, line 130: we should remove the loop, we only support one compression. makes code cleaner

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,423 @@
...
+    const actualChecksum = await computeFileChecksum(compressedFile);
+
+    let inReleaseContent = '';
+
+    try {
+      inReleaseContent = await this.fetchInReleaseFile(baseReleaseUrl);
+    } catch (error) {
+      // This is expected to fail for Artifactory if GPG verification is not enabled
+      logger.debug(
+        { url: baseReleaseUrl, error },
```
viceice, line 237:
```suggestion
        { url: baseReleaseUrl, err: error },
```

we've a special log formatter for `err` error properties

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,423 @@
...
+    }
+    const readStream = this.http.stream(packageUrl);
+    const writeStream = fs.createCacheWriteStream(compressedFile);
+    await fs.pipeline(readStream, writeStream);
+    logger.debug(
+      { url: packageUrl, targetFile: compressedFile },
+      'Downloading Debian package file',
+    );
+
+    const actualChecksum = await computeFileChecksum(compressedFile);
```
viceice, line 228: only compute if inReleaseContent != ''

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,423 @@
...
+    // read line by line to avoid high memory consumption as the extracted Packages
+    // files can be multiple MBs in size
+    const rl = readline.createInterface({
+      input: fs.createCacheReadStream(extractedFile),
+      terminal: false,
+    });
+
+    let currentPackage: PackageDescription = {};
+    // A Package Index can contain multiple Versions of the package on private Artifactory (e.g. Jfrog)
+    const allPackages: Record<string, PackageDescription[]> = {};
```
viceice, line 325: hopefully this doesn't cause to much memory and package cache pressure

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,423 @@
...
+    return await this.parseExtractedPackageIndex(extractedFile, lastTimestamp);
+  }
+
+  /**
+   * Fetches the release information for a given package from the registry URL.
+   *
+   * @param config - Configuration for fetching releases.
+   * @returns The release result if the package is found, otherwise null.
+   */
+  async getReleases({
```
viceice, line 382: @zharinov should we enable the cachable property to true? so this call is also cached?

oxdev03: it is already cached in getPackageIndex, so it would be redundant

lib/modules/datasource/deb/index.ts
oxdev03: depends if you are using a public or private artifactory. For a extracted package (public artifactory) index (public artifactory), which is around 100MB. After the parsing the size would be around ~5mb.

lib/modules/datasource/deb/index.ts
zharinov: The `@cache` decorator is the more functional alternative, e.g. you can set the TTL

secustor: We will get cache bust for each configuration option and their order in the registryUrl though. 
The current implementation checks cache after breaking down this options.

oxdev03: I meant its already done here.

```
  @cache({
    namespace: `datasource-${DebDatasource.id}`,
    key: (componentUrl: string) => componentUrl,
  })
  async getPackageIndex(
    componentUrl: string,
  ): Promise<Record<string, PackageDescription[]>> {
    const { extractedFile, lastTimestamp } =
      await this.downloadAndExtractPackage(componentUrl);
    return await this.parseExtractedPackageIndex(extractedFile, lastTimestamp);
  }
```

an no async operations happen in getReleases if the cache of getPackageIndex is not stale.

but I still added this.

oxdev03: let me revert the change again, github seems bugging and not loading latest the conversations 😅.

lib/modules/datasource/deb/index.spec.ts
```diff
@@ -0,0 +1,564 @@
...
+/**
+ * Computes the SHA256 checksum of a specified file.
+ *
+ * @param filePath - path of the file
+ * @returns resolves to the SHA256 checksum
+ */
+function computeFileChecksum(filePath: string): Promise<string> {
+  const stream = createReadStream(filePath);
+  return hashStream(stream, 'sha256');
+}
```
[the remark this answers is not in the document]
oxdev03: that only works if it is in the cache dir, since it uses createCacheReadStream and the fixtures are not located their.

lib/modules/datasource/deb/index.spec.ts
```diff
@@ -0,0 +1,564 @@
...
+        registryUrl:
+          'http://deb.debian.org/debian?suite=stable&components=non-free&binaryArch=amd64',
+        releases: [
+          {
+            version: '4.15-1',
+          },
+        ],
+      });
+
+      const modifiedTs = httpMock.getTrace()[0].headers['if-modified-since'];
```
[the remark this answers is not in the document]
oxdev03: So I need to replace it with the current mock?  
Couldn't find any example in existing tests.

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,345 @@
...
+
+  /**
+   * Fetches the release information for a given package from the registry URL.
+   *
+   * @param config - Configuration for fetching releases.
+   * @returns The release result if the package is found, otherwise null.
+   */
+  @cache({
+    namespace: `datasource-${DebDatasource.id}-releases`,
+    key: (registryUrl: string, packageName: string) =>
```
viceice, line 301:
```suggestion
    key: ({registryUrl, packageName}: GetReleasesConfig) =>
```
needs to match function below

lib/util/cache/package/types.ts
```diff
@@ -42,6 +42,8 @@ export type PackageCacheNamespace =
   | 'datasource-cpan'
   | 'datasource-crate-metadata'
   | 'datasource-crate'
+  | 'datasource-deb'
+  | 'datasource-deb-releases'
```
viceice, line 46:
```suggestion
```

lib/modules/datasource/deb/index.ts
```diff
@@ -0,0 +1,345 @@
...
+      `${registryUrl}-${packageName}`,
```
viceice, line 302:
```suggestion
      `${registryUrl}:${packageName}`,
```

secustor (approved):

secustor:
@oxdev03 Thanks for the contribution and pushing this through! 🚀

PR #30105, opened 2024-07-09
fix: Restrict downgrade checks only to Docker versioning scheme

zharinov:


rarkins (approved):

viceice:
do we now know it's only happening on docker versioning?

zharinov:
> do we now know it's only happening on docker versioning?

It's not proven, but so far, only Docker-based versions were affected. And we have bunch of weird versioning schemes, which may be affected by our patch, so for now it's reasonable to restrict it only for Docker ones.

viceice:
> > do we now know it's only happening on docker versioning?
> 
> It's not proven, but so far, only Docker-based versions were affected. And we have bunch of weird versioning schemes, which may be affected by our patch, so for now it's reasonable to restrict it only for Docker ones.

ok, next time please add this information to the PR description or link to relevant issue or discussion 🙃

PR #30119, opened 2024-07-10
feat(asdf): Add gleam to asdf manager

paulo-ferraz-oliveira:
## Changes

This pull request adds support to bumping `gleam` entries in `.tool-versions` files, as supported by https://asdf-vm.com/.

## Context

`gleam` is not supported (by Renovate) in `.tool-versions` files, at this moment.

### DIFF
--- a/lib/modules/manager/asdf/upgradeable-tooling.ts
+++ b/lib/modules/manager/asdf/upgradeable-tooling.ts
@@ -220,6 +220,14 @@ export const upgradeableTooling: Record<string, ToolingDefinition> = {
       extractVersion: '^v(?<version>\\S+)',
     },
   },
+  gleam: {
+    asdfPluginUrl: 'https://github.com/asdf-community/asdf-gleam.git',
+    config: {
+      datasource: GithubTagsDatasource.id,
+      packageName: 'gleam-lang/gleam',
+      extractVersion: '^v(?<version>.+)',
+    },
+  },
   gohugo: hugoDefinition,
   golang: {
     asdfPluginUrl: 'https://github.com/kennyp/asdf-golang',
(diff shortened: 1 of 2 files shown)


rarkins (dismissed):

lib/modules/manager/asdf/upgradeable-tooling.ts
```diff
@@ -220,6 +220,14 @@ export const upgradeableTooling: Record<string, ToolingDefinition> = {
       extractVersion: '^v(?<version>\\S+)',
     },
   },
+  gleam: {
+    asdfPluginUrl: 'https://github.com/asdf-community/asdf-gleam.git',
+    config: {
+      datasource: GithubTagsDatasource.id,
```
viceice, line 226: test failed. should it use releases or tags?

paulo-ferraz-oliveira: Lemme re-check. Thanks.

paulo-ferraz-oliveira: Both are populated. I think I see an inconsistency, but why would it matter if they both have the same values (not that it wouldn't semantically, though)?

paulo-ferraz-oliveira: 35373e325

paulo-ferraz-oliveira:
Tests are now passing; is this Ok for re-review? Thanks.

lib/modules/manager/asdf/upgradeable-tooling.ts
rarkins: This may be wrong, but I think quite a few asdf would be wrong too. Let's look another time

rarkins (approved):

Issue #21957, opened 2023-05-03
Add Bitrise ( CI provider ) manager


secustor:
### What would you like Renovate to be able to do?

[Bitrise](https://bitrise.io/) is a platform which provides CI services for mobile apps. 

Bitrise uses step definitions which are defined in yaml files with the name `bitrise.yaml`. 

Users of Bitrise are: 
- reddit
- duolingo
- equinox
- ... 

### If you have any ideas on how this should be implemented, please tell us here.

Steps are hosted and managed by repos with following pattern: `https://github.com/bitrise-steplib/steps-{{step reference}}`.

```yaml
format_version: 11
default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
project_type: android
app:
  envs:
  - MY_NAME: My Name
workflows:
  test:
    steps:
    - script@1.1.5:
        inputs:
        - content: echo "Hello ${MY_NAME}!"
```
Should result in a dependency with `github-release` datasource and `bitrise-steplib/steps-script` as `packageName`.

If `default_step_lib_source` deviates from `https://github.com/bitrise-io/bitrise-steplib.git`, `skipReason` should be set as we have no datasource to extract the folders in a generic way. 

### Is this a feature you are interested in implementing yourself?

Maybe

secustor:
@ubiratansoares As discussed could you fill out the questionnaire? Thx!

ubiratansoares:
@secustor On my list! 🔥

ubiratansoares:
@secustor : Let me know if I missed anything!

---

**Did you read our documentation on adding a package manager?**

- [x] I've read the [adding a package manager](adding-a-package-manager.md) documentation.

## Basics

### Name of package manager

- bitrise (https://bitrise.io/integrations)

### What language does this support?

- Not a programming language per se, it would be an equivalent of Github Actions or Circle Orbs (hence, YAML)

### How popular is this package manager?

- Every Bitrise user leverages one of more [steps provided by Bitrise](https://bitrise.io/integrations/steps), hence, Renovate would serve an entire fleet of iOS/Android/RN/Flutter Engineers building on top of this CI/CD provider

### Does this language have other (competing?) package managers?

- [ ] Yes (give names)
- [x] No

---

## Package File Detection

### What type of package files and names does it use?

[...]

PR #30138, opened 2024-07-11
feat(datasources): add bitrise datasource

secustor:
## Changes
Adds a Bitrise datasource

which uses https://github.com/bitrise-io/bitrise-steplib as source similar to the BitriseCLI
~~Added a helper function to `getRawFile` to `GithubHttp` and zod schemas for API calls to the `/content` endpoint.~~
 

## Context
Allows implementation of https://github.com/renovatebot/renovate/issues/21957


lib/util/http/github.ts
```diff
@@ -493,4 +493,12 @@ export class GithubHttp extends Http<GithubHttpOptions> {
 
     return result;
   }
+
+  public getRawFile(url: string): Promise<HttpResponse> {
```
viceice, line 497: can we split this to a separate PR?

lib/modules/platform/github/schema.ts
```diff
@@ -0,0 +1,31 @@
+import { z } from 'zod';
```
viceice, line 1: can we split this to a separate PR?

lib/modules/datasource/bitrise/index.ts
```diff
@@ -0,0 +1,89 @@
...
+  async getReleases({
+    packageName,
+    registryUrl,
+  }: GetReleasesConfig): Promise<ReleaseResult | null> {
+    // istanbul ignore if
+    if (!registryUrl) {
+      return null;
+    }
+
+    const client = new GithubHttp(this.id);
```
viceice, line 49: do in constructor and override default http property

lib/modules/datasource/bitrise/index.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    const parsedUrl = parseGitUrl(registryUrl);
+    if (detectPlatform(registryUrl) !== 'github') {
+      logger.warn(
+        `${parsedUrl.source} is not a supported Git hoster for this datasource`,
```
viceice, line 54: I think we should validate GitHub here, as the code below only works for GitHub not for ghes

lib/modules/datasource/bitrise/index.ts
```diff
@@ -0,0 +1,138 @@
...
+    registryUrl,
+  }: GetReleasesConfig): Promise<ReleaseResult | null> {
+    // istanbul ignore if
+    if (!registryUrl) {
+      return null;
+    }
+
+    const parsedUrl = parseGitUrl(registryUrl);
+    if (detectPlatform(registryUrl) !== 'github') {
+      logger.warn(
```
rarkins, line 56:
```suggestion
      logger.once.warn(
```

lib/modules/datasource/bitrise/index.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    // if we have no releases return null
+    if (!result.releases.length) {
+      return null;
+    }
+
+    return {
+      ...result,
+      homepage: `https://bitrise.io/integrations/steps/${packageName}`,
+    };
```
rarkins, line 136:
```suggestion
    result.homepage = `https://bitrise.io/integrations/steps/${packageName}`;
    return result;
```

Isn't this more efficient, or does it need this shallow copy?

rarkins (changes requested):

lib/modules/datasource/bitrise/index.ts
secustor: As the objects in question are minimal in size, the impact is nearly not existing. 
Doing that with RenovateConfig in a loop would be a different story.

rarkins (approved):

PR #30142, opened 2024-07-11
refactor(manager/circleci): replace regex system with YAML parsing

secustor:
## Changes
Replace regex based parsing with a YAML parser based one including schema. 

## Context
Fixes this issue too: https://github.com/renovatebot/renovate/pull/29902

### DIFF
--- a/lib/config/migration.ts
+++ b/lib/config/migration.ts
@@ -70,7 +70,7 @@ export function migrateConfig(config: RenovateConfig): MigratedConfig {
       } else if (is.array(val)) {
         if (is.array(migratedConfig?.[key])) {
           const newArray = [];
-          for (const item of migratedConfig[key] as unknown[]) {
+          for (const item of migratedConfig[key]) {
             if (is.object(item) && !is.array(item)) {
               const arrMigrate = migrateConfig(item as RenovateConfig);
               newArray.push(arrMigrate.migratedConfig);
--- a/lib/config/migration.ts
+++ b/lib/config/migration.ts
@@ -70,7 +70,7 @@ export function migrateConfig(config: RenovateConfig): MigratedConfig {
       } else if (is.array(val)) {
         if (is.array(migratedConfig?.[key])) {
           const newArray = [];
-          for (const item of migratedConfig[key]) {
+          for (const item of migratedConfig[key] as unknown[]) {
             if (is.object(item) && !is.array(item)) {
               const arrMigrate = migrateConfig(item as RenovateConfig);
               newArray.push(arrMigrate.migratedConfig);
--- a/lib/modules/datasource/npm/npmrc.ts
+++ b/lib/modules/datasource/npm/npmrc.ts
@@ -29,7 +29,7 @@ function envReplace(value: any, env = process.env): any {
       logger.warn('Failed to replace env in config: ' + match);
       throw new Error('env-replace');
     }
-    return env[envVarName]!;
+    return env[envVarName];
   });
 }
 
--- a/lib/modules/datasource/npm/npmrc.ts
+++ b/lib/modules/datasource/npm/npmrc.ts
@@ -29,7 +29,7 @@ function envReplace(value: any, env = process.env): any {
       logger.warn('Failed to replace env in config: ' + match);
       throw new Error('env-replace');
     }
-    return env[envVarName];
+    return env[envVarName]!;
   });
 }
 
--- a/lib/modules/manager/circleci/__fixtures__/config3.yml
+++ b/lib/modules/manager/circleci/__fixtures__/config3.yml
@@ -1,18 +1,16 @@
-aliases:  
+aliases:
   - &nodejs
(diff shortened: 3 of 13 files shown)


lib/config/migration.ts
```diff
@@ -70,7 +70,7 @@ export function migrateConfig(config: RenovateConfig): MigratedConfig {
       } else if (is.array(val)) {
         if (is.array(migratedConfig?.[key])) {
           const newArray = [];
-          for (const item of migratedConfig[key] as unknown[]) {
+          for (const item of migratedConfig[key]) {
```
viceice, line 73: revert

lib/modules/datasource/npm/npmrc.ts
```diff
@@ -29,7 +29,7 @@ function envReplace(value: any, env = process.env): any {
       logger.warn('Failed to replace env in config: ' + match);
       throw new Error('env-replace');
     }
-    return env[envVarName]!;
+    return env[envVarName];
```
viceice, line 32: revert

lib/modules/manager/circleci/__fixtures__/config3.yml
```diff
@@ -1,18 +1,16 @@
-aliases:  
+aliases:
   - &nodejs
     image: cimg/node:14.8.0
 
 version: 2
 jobs:
-  checkout: 
-    <<: *defaults
```
viceice, line 8: why remove?

lib/modules/manager/gradle/parser/handlers.ts
```diff
@@ -371,7 +371,7 @@ export function handleApplyFrom(ctx: Ctx): Ctx {
 
   const matchResult = parseGradle(
     // TODO #22198
-    ctx.fileContents[scriptFilePath]!,
+    ctx.fileContents[scriptFilePath],
```
viceice, line 374: revert

lib/modules/manager/npm/post-update/npm.ts
```diff
@@ -234,7 +234,7 @@ export async function generateLockFile(
           if (
             lockFileParsed.packages?.['']?.[depType]?.[lockUpdate.packageName!]
           ) {
-            lockFileParsed.packages[''][depType]![lockUpdate.packageName!] =
+            lockFileParsed.packages[''][depType][lockUpdate.packageName!] =
```
viceice, line 237: revert

lib/modules/manager/nuget/artifacts.ts
```diff
@@ -163,7 +163,7 @@ export async function updateArtifacts({
           file: {
             type: 'addition',
             path: lockFileName,
-            contents: newLockFileContentMap[lockFileName]!,
+            contents: newLockFileContentMap[lockFileName],
```
viceice, line 166: revert

lib/modules/manager/circleci/__fixtures__/config3.yml
secustor: YAML parsing fails because there there is no definition for the `defaults` anchor.

viceice: ok, but is it a valid circleci file? do they provide an implicit default anchor?

we could simply strip such anchors before parsing, like we do in gitlab manager

viceice (commented):
see comment

lib/modules/manager/circleci/__fixtures__/config3.yml
secustor: I don't thinks so, I can't find any reference to something like default anchor which gets injected into the file before parsing. 

See https://discuss.circleci.com/t/using-defaults-syntax-in-config-yaml-aka-yaml-anchors/16168 for example.

The other test files contain anchors too and parsing fine. 
Actually the fact anchors are now resolved is the reason I had to change the tests.

lib/modules/manager/circleci/extract.ts
```diff
@@ -1,80 +1,50 @@
...
-          },
-          'CircleCI docker image',
-        );
-        dep.depType = 'docker';
-        dep.versioning = 'docker';
-        if (
-          !dep.depName?.startsWith('ubuntu-') &&
-          !dep.depName?.startsWith('windows-server-') &&
-          !dep.depName?.startsWith('android-') &&
-          dep.depName !== 'android'
```
viceice, line 72: you dropped this condition, it should stay because those values are not resolvable docker images

secustor: This on purpose this is not needed anymore. 

Before that it has been used to cover `jobs.release_docker.machine.image` as the regexes looked for `image:`

https://github.com/renovatebot/renovate/blob/34c56e3fe1c4048cfadab6a706553d488c1ef0a0/lib/modules/manager/circleci/__fixtures__/config3.yml#L15-L17

viceice (approved):

PR #30155, opened 2024-07-12
chore(http/github): add utility function to fetch raw files

secustor:
## Changes
add `getRawFile` which will set the necessary accept header to get a raw file. 
If the requested API path is not a single file it will default to the same behavior as `application/json`

## Context
Enables https://github.com/renovatebot/renovate/pull/30155


lib/util/http/github.spec.ts
```diff
@@ -786,4 +786,24 @@ describe('util/http/github', () => {
...
+        .scope(githubApiHost)
+        .get('/foo/bar/contents/lore/ipsum.txt')
+        .matchHeader(
+          'accept',
+          'application/vnd.github.raw+json, application/vnd.github.v3+json',
+        )
+        .reply(200, 'foo');
+      await expect(
+        githubApi.getRawFile(
+          `${githubApiHost}/foo/bar/contents/lore/ipsum.txt`,
```
viceice, line 802: will it work with relative paths?

viceice (commented):
otherwise LGTM

lib/util/http/github.spec.ts
secustor: Yes, it does.

lib/util/http/github.spec.ts
```diff
@@ -786,4 +786,38 @@ describe('util/http/github', () => {
...
+    });
+
+    it('support relative path', async () => {
+      httpMock
+        .scope(githubApiHost)
+        .get('/foo/bar/contents/lore/ipsum.txt')
+        .reply(200, 'foo');
+      await expect(
+        githubApi.getRawFile(
+          `${githubApiHost}/foo/bar/contents/foo/../lore/ipsum.txt`,
```
viceice, line 816:
```suggestion
          `foo/bar/contents/foo/../lore/ipsum.txt`,
```
I mean this

secustor: No that isn't supported

secustor: Actually, it has worked already. 

I have added additionally logic to allow simply querying by providing the path to a file and the repository as option.

lib/util/http/github.ts
```diff
@@ -493,4 +493,24 @@ export class GithubHttp extends Http<GithubHttpOptions> {
...
+      headers: {
+        accept: 'application/vnd.github.raw+json',
+      },
+    };
+
+    let newURL = url;
+    const httpRegex = regEx(/^https?:\/\//);
+    if (options.repository && !httpRegex.test(options.repository)) {
+      newURL = joinUrlParts(options.repository, 'contents', url);
+    }
```
viceice, line 512: why do we need this? only baseUrl should be added if url isn't absolute. 🤔

the repo field is used for auth only, see request method

secustor: Or do you simply mean `repository` should be removed from `newOptions` so it is not fed into `request`?

viceice: No, my idea with relative paths where to allow `foo/bar/contents/lore/ipsum.txt` urls, which should just work

lib/util/http/github.ts
```diff
@@ -493,4 +493,24 @@ export class GithubHttp extends Http<GithubHttpOptions> {
 
     return result;
   }
+
+  public getRawFile(
```
viceice, line 497: Add a tsdoc comment and a note that only text files are allowed. The return type indicates a string body, so when you pass a binary file url, then you'll get a Buffer as body which is higly unexpected at runtime.

So maybe rename to explpict mention and validate in function that returned body is a string.
```suggestion
  public getRawTextFile(
```

lib/util/http/github.ts
secustor: Yeah, they work already. I have added some tests to prevent regressions.

lib/util/http/github.ts
secustor: I haven't managed to force got to return a Buffer without setting the `responseType`, though I have added a check and a test for it just to be sure.

viceice (approved):

PR #30207, opened 2024-07-16
feat(http): Force default rate limits for some known hosts

### LINKED ISSUE #27802: Apply default rate limits for hosts/registries with known rate limit needs

For example, the npmjs registry or GOPROXY does not need rate limiting applied, while the following are known to need:
- Repology
- crates.io
- Rubygems API

Let's identify all such hosts in a single file/location, in a file like `lib/data/host-limits.ts`. I think we can reuse `matchHost` syntax, it can maybe even be `hostRules` format but is not necessary (the `matchHost` value can be a key for the Record).

Importantly, if a host has a default limit in this file then we should make sure that users cannot _increase_ the rate limit using `hostRules`. Ideally we can catch and warn if such an attempt is made. This protects fragile registries from falling victim to flooding from Renovate users unless they really go out of their way to overcome this.

zharinov:
## Changes

- Default limits can be overriden only towards more restriction
- 8 requests per second for `rubygems.org` 
- 1 request per second for `https://crates.io/api`
- 16 concurrent requests for each host by default

## Context

- Ref: #27369
- Closes: #27802


HonkingGoose (commented):
Do you have a source for the new rate limit numbers?

Maybe the project has documentation that says: "please limit to `x` requests per `y` time"?

lib/util/http/queue.ts
```diff
@@ -16,7 +17,11 @@ export function getQueue(url: string): PQueue | null {
   let queue = hostQueues.get(host);
   if (queue === undefined) {
     queue = null; // null represents "no queue", as opposed to undefined
-    const concurrency = getConcurrentRequestsLimit(url);
+    const hostConcurrency = getConcurrentRequestsLimit(url);
+    const defaultConcurrency = getDefaultConcurrentRequestsLimit(url);
+    const concurrency = Math.min(
+      ...[hostConcurrency, defaultConcurrency].filter((x) => x !== null),
```
viceice, line 23:
```suggestion
      ...[hostConcurrency, defaultConcurrency].filter(is.notNull),
```

lib/util/http/throttle.ts
```diff
@@ -33,9 +34,13 @@ export function getThrottle(url: string): Throttle | null {
   let throttle = hostThrottles.get(host);
   if (throttle === undefined) {
     throttle = null; // null represents "no throttle", as opposed to undefined
-    const throttleOptions = getThrottleIntervalMs(url);
-    if (throttleOptions) {
-      const intervalMs = throttleOptions;
+    const hostThrottleMs = getThrottleIntervalMs(url);
+    const defaultThrottleMs = getDefaultThrottleIntervalMs(url);
+    const throttleMs = Math.max(
+      ...[hostThrottleMs, defaultThrottleMs].filter((x) => x !== null),
```
viceice, line 40:
```suggestion
      ...[hostThrottleMs, defaultThrottleMs].filter(is.notNull),
```

viceice (commented):
otherwise LGTM

lib/util/http/rate-limit.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+      ]);
+      expect(getConcurrentRequestsLimit('https://index.crates.io')).toBeNull();
+    });
+
+    it('obtains the limit from the host rules', () => {
+      hostRules.add({ matchHost: 'example.com', concurrentRequestLimit: 123 });
+      expect(getConcurrentRequestsLimit('https://example.com')).toBe(123);
+    });
+
+    it('selects default value is host rule is greater', () => {
```
HonkingGoose, line 31:
```suggestion
    it('selects default value if host rule is greater', () => {
```

lib/util/http/rate-limit.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    });
+
+    it('returns null if host does not match', () => {
+      setHttpRateLimits([
+        { matchHost: 'https://crates.io/api/', concurrency: 123 },
+      ]);
+      expect(getThrottleIntervalMs('https://index.crates.io')).toBeNull();
+    });
+
+    it('obtains the limit from the host rules', () => {
```
HonkingGoose, line 61:
```suggestion
    it('gets the limit from the host rules', () => {
```

HonkingGoose (changes requested):

HonkingGoose (approved):
I'm happy with the test descriptions now. I'll let the maintainers review the rest.

lib/util/http/rate-limits.ts
```diff
@@ -0,0 +1,87 @@
...
+    throttleMs: 125,
+  },
+  {
+    // https://crates.io/data-access#api
+    matchHost: 'https://crates.io/api/',
+    throttleMs: 1000,
+  },
+  {
+    matchHost: '*',
+    concurrency: 16,
```
viceice, line 19: @rarkins do we really want this? 🤔

lib/util/http/rate-limits.ts
```diff
@@ -0,0 +1,87 @@
...
+  },
+];
+
+let limits: RateLimitRule[] = [];
+
+export function setHttpRateLimits(rules: RateLimitRule[] = defaults): void {
+  limits = rules;
+}
```
viceice, line 27: this is not fail save against runtime null passing

zharinov: In the runtime, it's meant to be called with no arguments, hence set `defaults`.
For tests, it remains empty everywhere except `rate-limit.spec.ts`.

What would you like to change?

viceice:
```suggestion
export function setHttpRateLimits(rules?: RateLimitRule[]): void {
  limits = rules ?? defaults;
```
so it's null and undefined safe for any invalid code ussge

lib/util/http/rate-limits.ts
rarkins: Probably a good idea. For example npm v9 had a default of 15 concurrent requests, which was lost in v10, and apparently it caused some hosts load problems.

zharinov:
This PR seems to be the only blocker for #27369

rarkins (approved):

viceice (approved):

PR #30215, opened 2024-07-17
fix(helmfile): support case with oci repository in different document

PhilipAbed:
## Changes

reading the `oci:true` flag in the `helmfile.yaml`

before: it was reading only in the same document
after: it will be read as part of the repositoryData, and then used later in the releases

## Context

https://github.com/renovatebot/renovate/discussions/30168


lib/modules/manager/helmfile/extract.ts
```diff
@@ -106,23 +110,21 @@ export async function extractPackageFile(
...
-      );
+
       if (isOCIRegistry(dep.chart)) {
         res.datasource = DockerDatasource.id;
         res.packageName = repoName + '/' + depName;
-      } else if (repository?.oci) {
+      } else if (registryData[repoName]?.oci) {
         res.datasource = DockerDatasource.id;
-        res.packageName = registryAliases[repoName] + '/' + depName;
+        res.packageName = registryData[repoName]?.url
```
viceice: use string template or urlJoin helper

lib/modules/manager/helmfile/extract.ts
```diff
@@ -12,7 +12,7 @@ import type {
   PackageFileContent,
 } from '../types';
 import type { Doc } from './schema';
-import { Doc as documentSchema } from './schema';
+import {Doc as documentSchema, HelmRepository} from './schema';
```
viceice:
```suggestion
import {Doc as documentSchema, type HelmRepository} from './schema';
```

lib/modules/manager/helmfile/extract.ts
```diff
@@ -51,11 +51,15 @@ export async function extractPackageFile(
...
     return null;
   }
+
   for (const doc of docs) {
     // Always check for repositories in the current document and override the existing ones if any (as YAML does)
     if (doc.repositories) {
-      registryAliases = {};
+      registryData = {};
+      // used for printing only
+      let registryAliases: Record<string, string> = {};
```
viceice: why not printing the new object to logs?

lib/modules/manager/helmfile/extract.ts
```diff
@@ -1,4 +1,5 @@
 import is from '@sindresorhus/is';
+import urlJoin from 'url-join';
```
secustor, line 2: use our own abstraction
```suggestion
import { joinUrlParts } from '../../../util/url';
```

lib/modules/manager/helmfile/extract.ts
```diff
@@ -51,15 +52,16 @@ export async function extractPackageFile(
     );
     return null;
   }
+
   for (const doc of docs) {
     // Always check for repositories in the current document and override the existing ones if any (as YAML does)
     if (doc.repositories) {
-      registryAliases = {};
+      registryData = {};
```
secustor, line 59:
```suggestion
```
You are resetting `registryData` iteration, that way you will have only the data for that specific document again. 🤔

viceice: expected if the current document is defining it's own repos

secustor: But can't a document define repos and reference ones in a other document at the same time?

viceice: I don't think so. if I remember correctly it's always overriding the previous.

PhilipAbed:
```
or (const doc of docs) {
    // Always check for repositories in the current document and override the existing ones if any (as YAML does)
```
it says in the comment

lib/modules/manager/helmfile/extract.spec.ts
```diff
@@ -363,6 +363,36 @@ describe('modules/manager/helmfile/extract', () => {
...
+      const content =
+        'repositories:\n' +
+        '  - name: oci-repo\n' +
+        '    url: ghcr.io/example/oci-repo\n' +
+        '    oci: true\n' +
+        '---\n' +
+        'releases:\n' +
+        '  - name: example\n' +
+        '    version: 0.1.0\n' +
+        '    chart: oci-repo/example';
```
secustor, line 376: use a codeBlock instead
https://github.com/renovatebot/renovate/blob/e2a7586466df0370ac71659332b5fb459a483d4b/lib/util/toml.spec.ts#L6-L10

secustor (commented):
small nitpick, else LGTM

viceice (dismissed):

lib/modules/manager/helmfile/extract.spec.ts
PhilipAbed: eh, last time i tried that, it failed on indentations :/ ill try

lib/modules/manager/helmfile/extract.spec.ts
```diff
@@ -363,6 +363,37 @@ describe('modules/manager/helmfile/extract', () => {
...
+repositories:
+  - name: oci-repo
+    url: ghcr.io/example/oci-repo
+    oci: true
+---
+releases:
+  - name: example
+    version: 0.1.0
+    chart: oci-repo/example
+`;
```
secustor, line 377:
```suggestion
      const content = codeBlock`
      repositories:
        - name: oci-repo
          url: ghcr.io/example/oci-repo
          oci: true
      ---
      releases:
        - name: example
          version: 0.1.0
          chart: oci-repo/example
      `;
```

secustor (approved):

PR #30227, opened 2024-07-18
fix(nuget): always coalesce sourceUrl and homepage if homepage available

NatMarchand:
## Changes

As suggested in #24843 when renovate is unable to fetch metadata from nuget, we don't want to early return in order to set sourceUrl with homepage if available.

## Context

When using github packages as nuget source, the call to fetch the nuspec always returns 404 as it is not implemented by github. Therefore, we're unable to get a value for sourceUrl and then cannot group our packages with `matchSourceUrls`

- Closes #24843 


NatMarchand:
Please note that it changes current tests behavior, I can change the expected results accordingly.

lib/modules/datasource/nuget/__snapshots__/index.spec.ts.snap
```diff
@@ -1497,6 +1497,7 @@ exports[`modules/datasource/nuget/index getReleases processes real data (v3) nus
       "version": "3.12.0",
     },
   ],
+  "sourceUrl": "https://nunit.org/",
```
rarkins, line 1500: This is better suited for homepage  and not sourceUrl

rarkins (changes requested):

lib/modules/datasource/nuget/__snapshots__/index.spec.ts.snap
NatMarchand: It's the same behavior as in 
`modules/datasource/nuget/index getReleases processes real data (v2) 1`
`modules/datasource/nuget/index getReleases processes real data (v3) feed is not a nuget.org 1`
`modules/datasource/nuget/index getReleases processes real data with no github project url (v2) 1`

rarkins (approved):

NatMarchand:
Hello!
Thanks for your approval, however there's still a conversation that needs to be resolved if it's ok for you !
Thanks

PR #30235, opened 2024-07-18
fix(cargo): Only use `cargo update --precise` in case of lockfile updates

### LINKED ISSUE #29620: Use cargo fetch for updating lock files

Discussed in https://github.com/renovatebot/renovate/discussions/29539

Per https://doc.rust-lang.org/cargo/commands/cargo-fetch.html, "cargo fetch" will update a lock file to match the Cargo.toml

This command would be better to use than our existing `cargo update` one. I think we can adopt this algorithm:

- If there are any non-lockfile-only updates (i.e. Cargo.toml is updated) then we run a single `cargo fetch`
- If there are any lockfile-only updates then we run `cargo update` targeting only those dependencies (similar to today)

We _might_ need an additional `cargo fetch` after the `cargo update` commands, but we can leave that out for now if we're not sure.

davidkna:
## Changes

`cargo update --workspace` has better support for duplicate dependency versions, which `cargo update --precise` can fail to update. To avoid issues, only use it for `lockfile`-style updates.

The initial approach followed the proposal in #29620:
> * If there are any non-lockfile-only updates (i.e. Cargo.toml is updated) then we run a single `cargo fetch`
> * If there are any lockfile-only updates then we run `cargo update` targeting only those dependencies (similar to today)

~~In addition, I also moved the missing-dep case to `cargo fetch` because it is the less invasive approach.~~ (Reverted)

## Context

Closes #29620
Discussed in #29539

I modified the unit test to run, but I did not extend them. I confirmed that my issue was solved, but did not do any more in-depth testing of different test scenarios.

### DIFF
--- a/lib/modules/manager/cargo/artifacts.spec.ts
+++ b/lib/modules/manager/cargo/artifacts.spec.ts
@@ -132,7 +132,7 @@ describe('modules/manager/cargo/artifacts', () => {
         packageName: 'dep1',
         lockedVersion: '1.0.0',
         newVersion: '1.0.1',
-        updateType: 'lockfileUpdate' as UpdateType,
+        updateType: 'lockfileUpdate' satisfies UpdateType as UpdateType,
       },
     ];
     expect(
@@ -291,21 +291,21 @@ describe('modules/manager/cargo/artifacts', () => {
         packageName: 'dep1',
         lockedVersion: '1.0.0',
         newVersion: '1.0.1',
-        updateType: 'lockfileUpdate' as UpdateType,
+        updateType: 'lockfileUpdate' satisfies UpdateType as UpdateType,
       },
       {
         depName: 'dep2',
         packageName: 'dep2',
         lockedVersion: '1.0.0',
         newVersion: '1.0.2',
-        updateType: 'lockfileUpdate' as UpdateType,
+        updateType: 'lockfileUpdate' satisfies UpdateType as UpdateType,
       },
       {
         depName: 'dep3',
         packageName: 'dep3',
         lockedVersion: '1.0.0',
         newVersion: '1.0.3',
-        updateType: 'lockfileUpdate' as UpdateType,
+        updateType: 'lockfileUpdate' satisfies UpdateType as UpdateType,
       },
     ];
 
@@ -343,21 +343,21 @@ describe('modules/manager/cargo/artifacts', () => {
       {
         depName: 'dep1',
         packageName: 'dep1',
-        updateType: 'minorUpdate' as UpdateType,
+        updateType: 'minor' satisfies UpdateType as UpdateType,
         lockedVersion: '1.0.0',
         newVersion: '1.0.1',
       },
       {
         depName: 'dep2',
         packageName: 'dep2',
-        updateType: 'minorUpdate' as UpdateType,
+        updateType: 'minor' satisfies UpdateType as UpdateType,
         lockedVersion: '1.0.0',
         newVersion: '1.0.2',
       },
       {
         depName: 'dep3',
         packageName: 'dep3',
-        updateType: 'lockfileUpdate' as UpdateType,
+        updateType: 'lockfileUpdate' satisfies UpdateType as UpdateType,
         lockedVersion: '1.0.0',
         newVersion: '1.0.3',
       },
--- a/lib/modules/manager/cargo/artifacts.spec.ts
+++ b/lib/modules/manager/cargo/artifacts.spec.ts
@@ -332,7 +332,7 @@ describe('modules/manager/cargo/artifacts', () => {
     ]);
   });
 
-  it('runs cargo fetch if there is any non-lockfile update', async () => {
+  it('runs cargo update precise only for lockfile update', async () => {
     fs.statLocalFile.mockResolvedValueOnce({ name: 'Cargo.lock' } as any);
     fs.findLocalSiblingOrParent.mockResolvedValueOnce('Cargo.lock');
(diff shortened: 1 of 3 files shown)


lib/modules/manager/cargo/artifacts.ts
```diff
@@ -144,13 +161,32 @@ async function updateArtifactsImpl(
...
+          (dep) => dep.updateType !== 'lockfileUpdate',
         );
+        if (anyNonLockfileUpdate) {
+          // Cargo fetch is safer to use for non-lockfile updates, because it
+          // has fewer issues with duplicate dependencies.
+          await cargoFetch(
+            packageFileName,
+            config.env ?? {},
+            config.constraints?.rust,
+          );
```
rarkins, line 175: We prefer to have a single exec with all commands, so that if there's a failure we can capture it more easily

rarkins (changes requested):

lib/modules/manager/cargo/artifacts.ts
```diff
@@ -51,9 +51,23 @@ async function cargoUpdatePrecise(
       ` --manifest-path ${quote(manifestPath)} --workspace`,
   ];
 
+  const anyNonLockfileUpdate = updatedDeps.some(
+    (dep) => dep.updateType !== 'lockfileUpdate',
+  );
+  if (anyNonLockfileUpdate) {
+    cmds.push(
+      `cargo fetch --config net.git-fetch-with-cli=true` +
```
rarkins, line 59: Should we do cargo fetch _instead_ of cargo update?

davidkna: I changed the implementation to run `cargo update --workspace` since it appears to be updating the lockfile in the same way without actually needing to fetch the dependencies.

rarkins (dismissed):

mariobris:
what is still blocking to merge the PR ? Does anyone know how long it will take ?

simonzkl:
Iirc `--precise` was used so Renovate has control over which version to update to (the expected `dep.newVersion`). I'm not sure what happens if `dep.newVersion` and the actual version you get out differ.

davidkna:
@simonzkl Assuming that `cargo update --workspace` already updates the lockfile to be consistent with the updated `Cargo.toml`, the `lockedVersion` will no longer be present in the lockfile, and `cargo update --precise` will not match it.
The error handling in the module will then apparently exempt all those dependencies from being updated precisely and will not use the targeted new version.

simonzkl:
> The error handling in the module will then apparently exempt all those dependencies from being updated precisely and already will not use the targeted new version.

The question then is, will Renovate correctly report the actual new version? Or will there be a missmatch between what Renovate is reporting and what is actually in `Cargo.lock`?

lib/modules/manager/cargo/artifacts.spec.ts
```diff
@@ -328,6 +332,60 @@ describe('modules/manager/cargo/artifacts', () => {
...
+      },
+    ];
+    expect(
+      await cargo.updateArtifacts({
+        packageFileName: 'Cargo.toml',
+        updatedDeps,
+        newPackageFileContent: '{}',
+        config,
+      }),
+    ).not.toBeNull();
```
viceice, line 372: use toEqual, otherwise we can silently fail in future

viceice (commented):
it seems we need real tests to validate new behavior

lib/modules/manager/cargo/artifacts.spec.ts
```diff
@@ -328,6 +332,62 @@ describe('modules/manager/cargo/artifacts', () => {
...
+    fs.findLocalSiblingOrParent.mockResolvedValueOnce('Cargo.lock');
+    git.getFile.mockResolvedValueOnce('Old Cargo.lock');
+    const execSnapshots = mockExecAll();
+    fs.findLocalSiblingOrParent.mockResolvedValueOnce('Cargo.lock');
+    fs.readLocalFile.mockResolvedValueOnce('New Cargo.lock');
+    const updatedDeps = [
+      {
+        depName: 'dep1',
+        packageName: 'dep1',
+        updateType: 'minorUpdate' as UpdateType,
```
viceice, line 346: wrong 

https://github.com/renovatebot/renovate/blob/49c0e6a2a1b058d47e79246cd66862fb65674d91/lib/config/types.ts#L336

lib/modules/manager/cargo/artifacts.spec.ts
```diff
@@ -328,6 +332,62 @@ describe('modules/manager/cargo/artifacts', () => {
...
+        lockedVersion: '1.0.0',
+        newVersion: '1.0.1',
+      },
+      {
+        depName: 'dep2',
+        packageName: 'dep2',
+        updateType: 'minorUpdate' as UpdateType,
```
viceice, line 353: wrong

lib/modules/manager/cargo/artifacts.spec.ts
```diff
@@ -132,6 +132,7 @@ describe('modules/manager/cargo/artifacts', () => {
         packageName: 'dep1',
         lockedVersion: '1.0.0',
         newVersion: '1.0.1',
+        updateType: 'lockfileUpdate' as UpdateType,
```
viceice, line 135:
```suggestion
        updateType: 'lockfileUpdate' satisfies UpdateType,
```
use satisfies

davidkna: I ended up needing to use both `as` and `satisfies` to satisfy the type checker.

viceice (approved):

PR #30285, opened 2024-07-21
ci(docs): Check broken links

zharinov:
## Changes

Leverage [lychee CLI tool](https://github.com/lycheeverse/lychee) to check internal and external links in the documentation.


rarkins (dismissed):

lib/modules/manager/tekton/readme.md
```diff
@@ -17,7 +17,7 @@ Read the [Tekton Pipeline remote resolution docs](https://tekton.dev/docs/pipeli
 
 ### Using a PipelinesAsCode remote URL reference
 
-By specifying the annotation with a remote task or a remote pipeline based on the recommended way using [git based versioning](https://github.com/tektoncd/community/blob/main/teps/0115-tekton-catalog-git-based-versioning/index.md). How this can be used can be seen in the example below.
+By specifying the annotation with a remote task or a remote pipeline based on the recommended way using [git based versioning](https://github.com/tektoncd/community/blob/main/teps/0115-tekton-catalog-git-based-versioning.md). How this can be used can be seen in the example below.
```
HonkingGoose, line 20:
```suggestion
By specifying the annotation with a remote task or a remote pipeline based on the recommended way using [git based versioning](https://github.com/tektoncd/community/blob/main/teps/0115-tekton-catalog-git-based-versioning.md).
How this can be used can be seen in the example below.
```

Let's make this one sentence per line.

.github/lychee.toml
```diff
@@ -0,0 +1,21 @@
...
+  "localhost",
+  "codepen.io",
+  "repology.org",
+  "https://github.com/foo/bar/",
+  "https://raw.githubusercontent.com/foo/bar/",
+  "https://developer.mend.io/api/mc/badges",
+  "https://hub.docker.com/r/renovate/renovate/",
+  "^https\\:\\/\\/github\\.com\\/renovatebot\\/renovate\\/issues\\?q\\=is\\:issue\\+is\\:open\\+sort\\:updated\\-desc\\+label\\:manager\\:pre\\-commit$",
+]
+insecure = true
```
HonkingGoose, line 16: According to the docs:

> `  -i, --insecure
          Proceed for server connections considered insecure (invalid TLS)
`

Do we need this flag?

HonkingGoose (changes requested):

.github/workflows/build.yml
```diff
@@ -596,6 +596,65 @@ jobs:
           name: docs
           path: tmp/docs/
 
+  check-doc-links:
+    needs: [build-docs]
```
viceice, line 600: use list style

docs/usage/examples/opentelemetry.md
```diff
@@ -130,7 +130,7 @@ You should be able to see now the full trace view which shows each HTTP request
 ### Metrics
 
 Additional to the received traces some metrics are calculated.
-This is achieved using the [spanmetricsprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanmetricsprocessor).
+This is achieved using the [spanmetricsprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/3cd5b15466fe4065ef76ea5c557dab435a7d7318/processor/spanmetricsprocessor).
```
viceice, line 133: do we really want to pin this?

lib/modules/versioning/ruby/index.ts
```diff
@@ -19,7 +19,6 @@ export const id = 'ruby';
 export const displayName = 'Ruby';
 export const urls = [
   'https://guides.rubygems.org/patterns/',
-  'https://bundler.io/v1.5/gemfile.html',
```
viceice, line 22: why removed?

.github/workflows/build.yml
```diff
@@ -596,6 +596,65 @@ jobs:
...
+        uses: actions/cache/restore@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+        with:
+          path: .lycheecache
+          key: lychee-cache-main
+
+      - name: Check links
+        uses: lycheeverse/lychee-action@2b973e86fc7b1f6b36a93795fe2c9c6ae1118621 # v1.10.0
+        with:
+          args: --config .github/lychee.toml --include-fragments tmp/docs
+          fail: true
```
viceice, line 635: this should not accidentally fail if external urls change. will be really bad ci experience

docs/usage/examples/opentelemetry.md
zharinov: The current link is broken, `spanmetricprocessor` was deprecated and removed

.github/workflows/build.yml
zharinov: I agree, it should be annoying enough for us to fix links ASAP, but shouldn't prevent us from :shipit:

lib/modules/versioning/ruby/index.ts
rarkins: This link keeps changing, let's leave it off

zharinov:
Now it will fail, but this won't block merge because none of required tasks depend on it

viceice (dismissed):

lib/modules/versioning/ruby/index.ts
```diff
@@ -19,7 +19,6 @@ export const id = 'ruby';
 export const displayName = 'Ruby';
 export const urls = [
   'https://guides.rubygems.org/patterns/',
```
zharinov, line 21:
```suggestion
  'https://guides.rubygems.org/patterns/',
  'https://bundler.io/v2.5/man/gemfile.5.html',
```

.github/lychee.toml
```diff
@@ -0,0 +1,20 @@
...
+  "codepen.io",
+  "repology.org",
+  "https://github.com/foo/bar/",
+  "https://raw.githubusercontent.com/foo/bar/",
+  "https://developer.mend.io/api/mc/badges",
+  "https://hub.docker.com/r/renovate/renovate/",
+  "^https\\:\\/\\/github\\.com\\/renovatebot\\/renovate\\/issues\\?q\\=is\\:issue\\+is\\:open\\+sort\\:updated\\-desc\\+label\\:manager\\:pre\\-commit$",
+]
+
+accept = ["200", "429"]
```
HonkingGoose, line 17: The [lychee docs, rate limiting](https://lychee.cli.rs/troubleshooting/rate-limits/) list steps to reduce the number of 429 responses. Accepting `429` rate limited responses is the last resort. So maybe we should try the other things the docs list?

| Method | Do we use it? | Notes |
| :---- | :--- | :--- | 
| Do not retry a failed link, with `--max-retries 0` | No | |
| Set lower `--max-concurrency` | No | Lower concurrency makes the action slower! |
| Use `GITHUB_TOKEN` | No | A personal token with no extra permissions will work to check public links. |
| Caching | Yes, 1 day | |
| Ignore certain sites or paths | Yes | | 
| Accept `429` responses | Yes | Last resort according to the lychee docs. |

.github/workflows/build.yml
```diff
@@ -595,8 +595,69 @@ jobs:
           name: docs
           path: tmp/docs/
 
+  check-doc-links:
+    needs:
+      - build-docs
+
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
```
HonkingGoose, line 603: The [GitHub Docs, `timeout-minutes`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) says:

> The maximum number of minutes to let a job run before GitHub automatically cancels it. Default: 360

> **Note:** The `GITHUB_TOKEN` expires when a job finishes or after a maximum of 24 hours.

Are you sure that 2 minutes is enough for the action to run succesfully? This timeout seems very low to me.

.github/lychee.toml
```diff
@@ -0,0 +1,20 @@
+verbose = "info"
```
HonkingGoose, line 1: I found these security related issues. I'll let you review them:

- https://github.com/lycheeverse/lychee/issues/1297
- https://github.com/lycheeverse/lychee/issues/1298

.github/lychee.toml
zharinov: We have an additional step to remove `429` responses from the cache, so the next run they will be retried. Eventually, all the rate limited links will be       checked (during 24 hours).

.github/workflows/build.yml
zharinov: It's okay, in my laptop runs for about 10 seconds

.github/lychee.toml
HonkingGoose: #### My summary of your comment

Here's how I understood your comment:

- We get lots of PRs per day. Either from Renovate bot, or us, or contributors.
- This means the link check runs at least once for each incoming PR.
- Because the link check runs often: we can swallow the `429` and "hope" that on the next run(s) the link checker gets a good response.
- Over the course of a day, we checked all rate-limited links.

#### One day cache long enough?

Is a day enough time to be reasonably sure we checked all the links? I mean, is it worth it to have a longer time to live for the cache? Say 2 days?

zharinov: 1 day is ok

.github/lychee.toml
zharinov: 🤷‍♂️

HonkingGoose:
We just merged this PR, that makes it so we use Material for MkDocs to build our docs. The publish step stays in the Renovate docs repository!

- #30735

I asked @viceice what we want to do with your PR:

> > What do you want to do with this PR?
> > 
> > * [ci(docs): Check broken links #30285](https://github.com/renovatebot/renovate/pull/30285)
> 
> I think it's obsolete then 🤷‍♂️

Because we're building with Material for MkDocs here now, we probably can rely more on their built-in link checks.

We should test if the docs build on `renovatebot/renovate` works as we expect. And then decide if we want to use the Lychee stuff.

In any case, if we decide to try Lychee, you should merge in the latest changes from `main` to make sure we're not having any logical conflicts with the docs build on `main`. 😉

PR #30305, opened 2024-07-22
docs: fix `changelogUrl` description

RahulGautamSingh:
## Changes
- rectify the description for `changelogUrl` 

## Context
Ref: https://github.com/renovatebot/renovate/discussions/30299
I think this was missed during migration from `customChangelogUrl` -> `changelogUrl`

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2558,8 +2558,7 @@ You must use the `!/ /` syntax, like this:
 
 ### changelogUrl
 
-Use this field to set the changelog URL for a package, including overriding any existing one.
-Using this field we can specify the exact URL to fetch changelogs from.
+This doesn't help with _fetching_ the changelogs, but if you configure it then Renovate will include a link to this URL in the PR body, so users can click through to read the changelog.
 
 ```json title="Setting the changelog URL for the dummy package"
 {
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2558,7 +2558,12 @@ You must use the `!/ /` syntax, like this:
 
 ### changelogUrl
 
-This doesn't help with _fetching_ the changelogs, but if you configure it then Renovate will include a link to this URL in the PR body, so users can click through to read the changelog.
+Sometimes Renovate does not show the correct changelog for a package.
+As a workaround for this problem, you can give Renovate the URL to the changelog with the `changelogUrl` config option.
+When set, Renovate will put a _link_ to the changelogs in the Renovate PR body.
+
+Renovate does _not_ show the complete changelogs from the `changelogUrl` in its PR body text, you only get the URL from Renovate.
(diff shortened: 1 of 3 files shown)


rarkins:
Do we fetch it though?

RahulGautamSingh:
> Do we fetch it though?

no, just add a link like we do for `source`. reversed the docs.

viceice (dismissed):

RahulGautamSingh:
> Do we fetch it though?

no, only link it like we do for source. fixed

lib/config/options/index.ts
```diff
@@ -1706,7 +1706,7 @@ const options: RenovateOptions[] = [
   {
     name: 'changelogUrl',
     description:
-      'If set, Renovate will use this URL to fetch changelogs for a matched dependency. Valid only within a `packageRules` object.',
+      'Allows you to specify a custom URL for the changelog, which Renovate will include in PR descriptions for easy access.',
```
HonkingGoose, line 1709:
```suggestion
      'Set a custom URL for the changelog. Renovate will put this URL in the PR body text.',
```

docs/usage/configuration-options.md
```diff
@@ -2558,8 +2558,7 @@ You must use the `!/ /` syntax, like this:
 
 ### changelogUrl
 
-Use this field to set the changelog URL for a package, including overriding any existing one.
-Using this field we can specify the exact URL to fetch changelogs from.
+This doesn't help with _fetching_ the changelogs, but if you configure it then Renovate will include a link to this URL in the PR body, so users can click through to read the changelog.
```
HonkingGoose, line 2561:
```suggestion
Sometimes Renovate does not show the correct changelog for a package.
As a workaround for this problem, you can give Renovate the URL to the changelog with the `changelogUrl` config option.
When set, Renovate will put a _link_ to the changelogs in the Renovate PR body.

Renovate does _not_ show the complete changelogs from the `changelogUrl` in its PR body text, you only get the URL from Renovate.
To read the changelogs you must use the link.
```

My goals with this text are to explain that:

- `changelogUrl` is a workaround for missing changelogs
- you only get a _link_ to the changelog, not the full changelogs

We already have a note admonition below after the codeblock to explain that setting `changelogUrl` does not help Renovate _fetch_ the changelogs. So I'm dropping that bit of info in my rewrite.

HonkingGoose (changes requested):

lib/config/options/index.ts
RahulGautamSingh:
```suggestion
      'Set a custom URL for the changelog. Renovate will put this URL in the PR body text.',
```

RahulGautamSingh, line 1709:
```suggestion
      'Set a custom URL for the changelog. Renovate will put this URL in the PR body text.',
```

rarkins (approved):

PR #30350, opened 2024-07-24
fix: check for explicit preset during onboarding

LukasAuerbeck:
## Changes
Updates the onboarding handling to ensure explicitly set `onboardingConfig.extends` in the renovate `config.js` don't get overwritten by the default renovate-config repo lookup. See https://github.com/renovatebot/renovate/discussions/30202#discussioncomment-10064964 for more detailed discussion about the expected behaviour.

### DIFF
--- a/lib/workers/repository/onboarding/branch/config.ts
+++ b/lib/workers/repository/onboarding/branch/config.ts
@@ -14,20 +14,26 @@ async function getOnboardingConfig(
 ): Promise<RenovateSharedConfig | undefined> {
   let onboardingConfig = clone(config.onboardingConfig);
 
-  // TODO #22198 fix types
-  const foundPreset = await searchDefaultOnboardingPreset(config.repository!);
+  const hasPredefinedPreset = onboardingConfig?.extends !== undefined;
 
-  if (foundPreset) {
-    logger.debug(`Found preset ${foundPreset} - using it in onboarding config`);
-    onboardingConfig = {
-      $schema: 'https://docs.renovatebot.com/renovate-schema.json',
-      extends: [foundPreset],
-    };
-  } else {
-    // Organization preset did not exist
-    logger.debug(
-      'No default org/owner preset found, so the default onboarding config will be used instead.',
-    );
+  if (!hasPredefinedPreset) {
+    // TODO #22198 fix types
+    const foundPreset = await searchDefaultOnboardingPreset(config.repository!);
+
+    if (foundPreset) {
+      logger.debug(
+        `Found preset ${foundPreset} - using it in onboarding config`,
+      );
+      onboardingConfig = {
+        $schema: 'https://docs.renovatebot.com/renovate-schema.json',
+        extends: [foundPreset],
+      };
+    } else {
+      // Organization preset did not exist
+      logger.debug(
+        'No default org/owner preset found, so the default onboarding config will be used instead.',
+      );
+    }
   }
 
   logger.debug({ config: onboardingConfig }, 'onboarding config');
--- a/lib/workers/repository/onboarding/branch/config.ts
+++ b/lib/workers/repository/onboarding/branch/config.ts
@@ -16,9 +16,10 @@ async function getOnboardingConfig(
 
   const hasPredefinedPreset = onboardingConfig?.extends !== undefined;
   if (hasPredefinedPreset) {
-    logger.debug('Predefined onboardingConfig.extends found, so default org presets won't be used');
-  }
-  if (!hasPredefinedPreset) {
+    logger.debug(
+      "Predefined onboardingConfig.extends found, so default org presets won't be used",
+    );
+  } else {
     // TODO #22198 fix types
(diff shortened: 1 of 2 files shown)


rarkins:
Why not make the logic higher-level?

In my opinion, if any `onboardingConfig` is set at all (even empty), then we shouldn't hunt for any default extends

LukasAuerbeck:
Yea, I agree it makes most sense to skip the lookups. Initially opted for the easiest option to prevent the overwriting.

LukasAuerbeck:
Is there anything open for this PR to be ready to merge/anything left on my side to do? In my last commit I changed the implementation to check for a user-defined preset before renovate attempts to find a repo that can be used as a default.

lib/workers/repository/onboarding/branch/config.ts
```diff
@@ -14,12 +14,39 @@ async function getOnboardingConfig(
...
+        'No default org/owner preset found, so the default onboarding config will be used instead.',
+      );
+    }
+  }
+
+  logger.debug({ config: onboardingConfig }, 'onboarding config');
+  return onboardingConfig;
+}
 
+async function searchDefaultOnboardingPreset(
```
viceice, line 42: please do this refactor in a separate PR

viceice (changes requested):
@rarkins my expectation is the opposite. I like to use the default org config to be preferred over the onboarding config from admin config as it is now. if this will be changed, then we should have a flag to flip.

LukasAuerbeck:
Moved the `searchDefaultOnboardingPreset` refactor into a dedicated PR and had to do some force-pushing/rebasing to clean my fork's main branch.

I think adding an additional option for admins to configure `onboardingRepositoryLookup` (or similar) also makes sense and enables a few more complex configurations.

What comes to my mind is a situation in an org with these repos:
- `org/org-unit-1/renovate-config`
- `org/org-unit-2/renovate-config`

Where `org-unit-1` operates the renovate for the whole `org` group. They may have a config.js like this:

```js
{
  onboardingConfig: {
    extends: [
      "local>org/org-unit-1/renovate-config#v1.23.2"
    ]
  }
}
```

## PR in org-unit-1 group

With option `onboardingRepositoryLookup` set to false, my assumption would be that renovate skips the repo lookups and generates this config:

```json
{
  "extends": [
    "local>org/org-unit-1/renovate-config#v1.23.2"
  ]
}
```

With `onboardingRepositoryLookup` set to true, I would assume renovate attempts to do the repo lookups, finds the `org/org-unit-1/renovate-config` repo, but since the admin configured preset is more specific (since it contains the pinned version) I would still assume that the onboarding PR has this config:
```json
{
  "extends": [
    "local>org/org-unit-1/renovate-config#v1.23.2"
  ]
}
```

## PR in org-unit-2 group

With `onboardingRepositoryLookup` false I would assume the same onboarding config as in `org-unit-1` with an extends to `local>org/org-unit-1/renovate-config#v1.23.2`

With `onboardingRepositoryLookup` set to true, I would assume renovate attempts to do the repo lookups, finds the `org/org-unit-2/renovate-config` repo and creates this PR:
```json
{
  "extends": [
    "local>org/org-unit-2/renovate-config"
  ]
}
```

viceice:
that's why you should not use main branch as source for a PR. we've a precommit check which should error out and disallow main

LukasAuerbeck:
For future PRs I will use a dedicated branch in the forked repo, since I can't change the PR and I don't think it's that important to close the PR and create a new one. For now I will just manually rebase and force push the rebased commits. If I change anything in the actual implementation I will use a new commit to make re-reviewing easier.

What are your opinions on my suggestion in https://github.com/renovatebot/renovate/pull/30350#issuecomment-2264806790 to make the handling more flexible by adding an additional flag? Would handling it this way be in-line with both of your expectations @rarkins, @viceice?

lib/workers/repository/onboarding/branch/config.ts
```diff
@@ -14,20 +14,26 @@ async function getOnboardingConfig(
...
-      $schema: 'https://docs.renovatebot.com/renovate-schema.json',
-      extends: [foundPreset],
-    };
-  } else {
-    // Organization preset did not exist
-    logger.debug(
-      'No default org/owner preset found, so the default onboarding config will be used instead.',
-    );
+  const hasPredefinedPreset = onboardingConfig?.extends !== undefined;
+
```
rarkins, line 18:
```suggestion
  if (hasPredefinedPreset) {
    logger.debug('Predefined onboardingConfig.extends found, so default org presets won't be used');
  }
```

rarkins (approved):

viceice (changes requested):
i still need a config option to reverse the logic change because our org presets have higher prio than our default onboarding config.

rarkins:
@viceice so your use case is "if an org preset exists, then add it to onboarding config" and you run on multiple orgs, some which don't have one?

viceice:
> @viceice so your use case is "if an org preset exists, then add it to onboarding config" and you run on multiple orgs, some which don't have one?

yes

PR #30359, opened 2024-07-24
feat(package-rules): matchReleaseAge

mzglinski:
## Changes

Implement and add `matchReleaseAge` matcher to the packageRules, equivalent to `matchCurrentAge`.

## Context

For noise impression reasons. My organization requires all PRs not to fall behind a default branch. In a situation where the repository owner ignores the PR for too long, it is updated every time there is a change pushed to the default branch, resulting in constant notifications and unnecessary CI resource consumption. With this rule, we can bundle "forgotten" dependencies into a single PR to be dealt with at a convenient time for the owner. 

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2412,7 +2412,7 @@ The matching process for a package rule:
 - Multiple values within a single matcher will be evaluated independently (they're OR-ed together).
 - Combining multiple matchers will restrict the resulting matches (they're AND-ed together):
   `matchCurrentVersion`, `matchCurrentValue`, `matchNewValue`, `matchConfidence`, `matchCurrentAge`,
-  `matchManagers`, `matchDatasources`, `matchCategories`, `matchDepTypes`, `matchUpdateTypes`,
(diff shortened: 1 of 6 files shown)


RahulGautamSingh:
https://docs.renovatebot.com/configuration-options/#minimumreleaseage, same?

mzglinski:
@RahulGautamSingh I do not believe it is the same. `minimumReleaseAge` allows to wait for `X` amount of time before creating a PR. My suggestion allows (for example) to group new releases of packages of which the releaseTimestamp is older than one month. 

Example scenario: A few major packages are released, and the owner of the repository is not ready to update the project as of this moment, unfortunately, they decide to do nothing with the PR (no new rules, to hold the version, no stop-updating label etc), and just let it hang. those PRs are updated after every change made to the default branch. Generating noise and consuming CI resources. With the above matcher, I can specify (in the default config that our org uses), to group such updates into single PR.

RahulGautamSingh:
Got it. Please create a discussion with you points first, as the policy dictates to discuss before creating a PR. 
Once the maintainers comment/approve your idea, you can reopen this PR.

PR #30395, opened 2024-07-26
feat(changelogs): Include `depName` in release name patterns

### LINKED ISSUE #30297: Include depName in changelog release name patterns

Currently we look for github releases either by version (e.g. `1.0.0`) or `packageName-release`, e.g. (`foo-1.0.0`). We should include `depName` in this search too.

Discussion with reproduction: https://github.com/renovatebot/renovate/discussions/29943 (test-dep update to v1.2.0)

usmonster:
## Changes

Update regular expressions and interfaces to allow filtering releases by `depName` (in addition to `packageName`), thus enabling broader support for changelog retrieval.

## Context

Closes #30297.


lib/workers/repository/update/pr/changelog/release-notes.spec.ts
```diff
@@ -644,6 +644,51 @@ describe('workers/repository/update/pr/changelog/release-notes', () => {
...
+            'some body #123, [#124](https://github.com/some/yet-other-repository/issues/124)',
+        },
+      ]);
+
+      const res = await getReleaseNotes(
+        {
+          ...githubProject,
+          repository: 'some/other-repository',
+          packageName: 'some.registry/some/other',
+          depName: 'other',
```
usmonster, line 673: (☝️ The only difference with the previous test.)

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -166,10 +173,13 @@ export async function getReleaseNotes(
 
 function getExactReleaseMatch(
   packageName: string,
+  depName: string,
   version: string,
   releases: ChangeLogNotes[],
 ): ChangeLogNotes | undefined {
-  const exactReleaseReg = regEx(`${packageName}[@_-]v?${version}`);
+  const exactReleaseReg = regEx(
+    `(?:${depName}|${packageName})[@_-]v?${version}$`,
```
usmonster, line 181: > [!NOTE]
> Terminated the regex with `$` to make it "more exact" as a scout habit; feel free to revert if desired.

rarkins: Best to revert this because we don't know the impact, plus we are more likely to want "loose" matching than to be extra strict

lib/workers/repository/update/pr/changelog/source.ts
```diff
@@ -185,11 +188,19 @@ export abstract class ChangeLogSource {
   private findTagOfRelease(
     version: allVersioning.VersioningApi,
     packageName: string,
+    depName: string,
     depNewVersion: string,
     tags: string[],
   ): string | undefined {
-    const regex = regEx(`(?:${packageName}|release)[@-]`, undefined, false);
-    const exactReleaseRegex = regEx(`${packageName}[@\\-_]v?${depNewVersion}`);
+    const releaseNameRegexPrefix = `(?:${depName}|${packageName}|release)[@_-]`
```
usmonster, line 195: Simplified a bit here. Had 2 questions:

1. Should/can this prefix include the optional `v?`?
1. Naming is hard—do you suggest a better one? (`releaseRegexPrefix`?)

Example:
```suggestion
    const releaseRegexPrefix = `(?:${depName}|${packageName}|release)[@_-]v?`
```

This shouldn't negatively effect tag filtering.

lib/workers/repository/update/pr/changelog/source.ts
```diff
@@ -185,11 +188,19 @@ export abstract class ChangeLogSource {
...
+    const regex = regEx(
+      releaseNameRegexPrefix,
+      undefined,
+      false,
+    );
+    const exactReleaseRegex = regEx(
+      `${releaseNameRegexPrefix}v?${depNewVersion}$`,
```
usmonster, line 202: (Same as https://github.com/renovatebot/renovate/pull/30395/files#r1692674786.)

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -166,10 +173,13 @@ export async function getReleaseNotes(
...
+  );
   const candidateReleases = releases.filter((r) => r.tag?.endsWith(version));
   const matchedRelease = candidateReleases.find((r) =>
     exactReleaseReg.test(r.tag!),
```
usmonster, line 185: Maybe slightly off-topic, but I wonder if the regex for `other` would match both `other-1.0.1` and `not-other-1.0.1`. 🤔

I'm tempted to prefix the regex with `(?:^|\/)` to avoid this kind of substring overmatching... maybe for another PR, but maybe worth doing here since `depName` is often shorter than `packageName` and could lead to more false partial matches.

WDYT?

lib/workers/repository/update/pr/changelog/release-notes.ts
usmonster: Following [my comment below](https://github.com/renovatebot/renovate/pull/30395/files#r1692724285), I'd like to further improve this to avoid unintentional partial matches (while still allowing for tag prefixes):
```suggestion
    `(?:^|\/)(?:${depName}|${packageName})[@_-]v?${version}$`,
```

usmonster: > Best to revert this because we don't know the impact, plus we are more likely to want "loose" matching than to be extra strict

After some code inspection, it looks like there's no impact here since the first filter to collect `candidateReleases` already ensures that the corresponding tag will strictly end with the `version`. Similarly, in the context of `ChangeLogSource`'s `findTagOfRelease()`, there's an equality check in the last filter to ensure the version matches exactly. So no problem to revert the change to those regex patterns. 🙂

I'm still interested to add the prefix referenced in my pervious comment as part of this PR in order to reduce the chance of partial matching issues. WDYT?

lib/workers/repository/update/pr/changelog/release-notes.ts
rarkins: OK to change if existing tests pass

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -166,10 +173,13 @@ export async function getReleaseNotes(
 
 function getExactReleaseMatch(
   packageName: string,
+  depName: string,
   version: string,
   releases: ChangeLogNotes[],
 ): ChangeLogNotes | undefined {
-  const exactReleaseReg = regEx(`${packageName}[@_-]v?${version}`);
+  const exactReleaseReg = regEx(
+    `(?:^|/)(?:${packageName}|${depName})[@_-]v?${version}`,
```
usmonster, line 181: `/` included here since release tag may have a prefix (delimited by `/`) which is not part of the `packageName` (cf. [related test](https://github.com/renovatebot/renovate/blob/70cc17fc397f8e652a6ff8e85dbbd63c3fa6aa07/lib/workers/repository/update/pr/changelog/release-notes.spec.ts#L862-L911)).

lib/workers/repository/update/pr/changelog/source.ts
```diff
@@ -185,11 +188,13 @@ export abstract class ChangeLogSource {
   private findTagOfRelease(
     version: allVersioning.VersioningApi,
     packageName: string,
+    depName: string,
     depNewVersion: string,
     tags: string[],
   ): string | undefined {
-    const regex = regEx(`(?:${packageName}|release)[@-]`, undefined, false);
-    const exactReleaseRegex = regEx(`${packageName}[@\\-_]v?${depNewVersion}`);
+    const releaseRegexPrefix = `^(?:${packageName}|${depName}|release)[@_-]v?`;
```
usmonster, line 195: Unlike [in release-notes](https://github.com/renovatebot/renovate/pull/30395/files#r1694023245), no `/` here since the regex is later used to strip the release name in order to test if only the new version remains. We already know the test will fail if there's a prefix, so this doesn't change that behavior. (If it's not desired behavior, we can properly address it in a separate pull request.)

usmonster:
Gentle nudge—seems like this is ready for final review before merging? Please let me know if someone can (update the branch and then) approve this or if something's missing—thanks!

viceice (approved):

usmonster:
Thanks for the approval @viceice, but I don't have access to merge it. Can you please reapprove and maybe enable automerge if possible? Thanks again!

PR #30472, opened 2024-07-30
feat(schedule): Log descriptive cron schedules

Sukhdev2002:

### DIFF
--- a/lib/workers/repository/update/branch/schedule.spec.ts
+++ b/lib/workers/repository/update/branch/schedule.spec.ts
@@ -426,11 +426,10 @@ describe('workers/repository/update/branch/schedule', () => {
     });
 
     it('should throw an error for an invalid cron expression "* * */2 6#1"', () => {
-      expect(() => {
-        cronstrue.toString('* * */2 6#1');
-      }).toThrow(
-        'Error: Expression has only 4 parts. At least 5 parts are required.',
-      );
+      const result = cronstrue.toString("* * */2 6#1", {
+        throwExceptionOnParseError: false,
+      });
+      expect(result).toBe("An error occured when generating the expression description.  Check the cron expression syntax.");
     });
   });
 });
--- a/lib/workers/repository/update/branch/schedule.spec.ts
+++ b/lib/workers/repository/update/branch/schedule.spec.ts
@@ -1,3 +1,4 @@
+import cronstrue from 'cronstrue';
 import type { RenovateConfig } from '../../../../config/types';
 import * as schedule from './schedule';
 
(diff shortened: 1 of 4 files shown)


rarkins (changes requested):
Needs tests, including if an invalid cron string is passed

Sukhdev2002:
> Needs tests, including if an invalid cron string is passed

test case already written for invalid string types

rarkins:
You had tests before but they're missing now

Sukhdev2002:
However, all the scenarios have already been covered in the previous test cases. Additionally, the test cases I added earlier will not run for the function where I made the changes.

rarkins:
We don't add features without tests to verify that the feature works

Sukhdev2002:
> We don't add features without tests to verify that the feature works

added please check

lib/workers/repository/update/branch/schedule.spec.ts
```diff
@@ -407,4 +408,21 @@ describe('workers/repository/update/branch/schedule', () => {
       });
     });
   });
+
+  describe('cronstrue', () => {
```
rarkins, line 412: Needs a test for an invalid value

rarkins (changes requested):

lib/workers/repository/update/branch/schedule.spec.ts
Sukhdev2002: ok

Sukhdev2002: > Needs a test for an invalid value

Invalid test case is added

rarkins (changes requested):

lib/workers/repository/update/branch/schedule.spec.ts
```diff
@@ -407,4 +408,28 @@ describe('workers/repository/update/branch/schedule', () => {
...
+    });
+
+    it('should correctly convert "0 23 * * *" to human-readable format', () => {
+      const result = cronstrue.toString('0 23 * * *');
+      expect(result).toBe('At 11:00 PM');
+    });
+    
+    it('should throw an error for an invalid cron expression "* * */2 6#1"', () => {
+      expect(() => {
+        cronstrue.toString("* * */2 6#1");
```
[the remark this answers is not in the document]
Sukhdev2002: > Use throwExceptionOnParseError=false like in the actual code

added

rarkins (approved):

lib/workers/repository/update/branch/schedule.ts
```diff
@@ -25,6 +26,10 @@ function parseCron(
   timezone?: string,
 ): CronExpression | undefined {
   try {
+    const cronScheduleSummary = cronstrue.toString(scheduleText, {
+      throwExceptionOnParseError: false,
+    });
+    logger.debug(`Human-readable summary for cron:: ${cronScheduleSummary}`);
```
viceice, line 32: this causes a weird debug message for all later based strings as we try to parse as cron first 😕

viceice: @rarkins we should refactor to only log if it's a cron expression and not a later expression.

viceice: this message is also logged multiple times (for each branch)

PR #30490, opened 2024-07-30
feat: auto inject Bitrise default host rule

secustor:
## Changes
Injects new hostRules scoped to the `bitrise` hostType and matchHost `https://api.github.com/repos/bitrise-io/bitrise-steplib/contents`.

## Context
Currently,  by default no authorization header are added while looking up Bitrise data, which results in rate limiting for bigger runs.

This is an alternative to setting the hostType to `github`.

 


docs/usage/getting-started/running.md
```diff
@@ -209,11 +209,11 @@ Read the platform-specific docs to learn how to setup authentication on your pla
 - [github.com and GitHub Enterprise Server](../modules/platform/github/index.md)
 - [GitLab](../modules/platform/gitlab/index.md)
 
-### GitHub.com token for changelogs
+### GitHub.com token for meta data
 
 If you are running on any platform except github.com, you should also set the environment variable `GITHUB_COM_TOKEN` and put the Personal Access Token for github.com in it.
 This account can be _any_ account on GitHub, and needs only `read-only` access.
-It's used when fetching changelogs for repositories in order to increase the hourly API limit.
+It's used to increase the hourly API limit when fetching changelogs or other contents from repositories.
```
rarkins, line 216: I think it's also mandatory for GraphQL queries

rarkins, line 216:
```suggestion
It's used to increase the hourly API limit when fetching changelogs or other contents from repositories and in required in some cases because GitHub does not permit anonymous GraphQL queries.
```

lib/workers/global/config/parse/env.ts
```diff
@@ -211,11 +212,30 @@ export async function getConfig(
...
+    const token = env.GITHUB_COM_TOKEN;
     logger.debug(`Converting GITHUB_COM_TOKEN into a global host rule`);
     config.hostRules.push({
       hostType: 'github',
       matchHost: 'github.com',
-      token: env.GITHUB_COM_TOKEN,
+      token,
+    });
+
+    // allow accessing the central bitrise repo via renovate token
```
rarkins, line 223: How come Bitrise is the only datasource which requires this type of special treatment?

secustor: It is currently the only datasource which uses the Github REST Api. `github-tags` and `github-releases` use GraphQL.

viceice: why do we use the API for raw contents instead of the public raw files endpoint?

secustor: This is not only for fetching RAW files. 
`bitrise` also fetches folder contents to gather the available versions. 
see for example https://github.com/bitrise-io/bitrise-steplib/tree/master/steps/activate-ssh-key

https://github.com/renovatebot/renovate/blob/f0c20677986914d0497e2366fb6e2fc48e05a5a5/lib/modules/datasource/bitrise/index.ts#L79

I use the contents API as that way not an additional call is necessary to get the default branch. 

https://raw.githubusercontent.com/bitrise-io/bitrise-steplib/master/steps/activate-ssh-key/3.0.2/step.yml.  
vs.  
https://api.github.com/repos/bitrise-io/bitrise-steplib/contents/steps/activate-ssh-key/3.0.2/step.yml

docs/usage/getting-started/running.md
```diff
@@ -209,11 +209,11 @@ Read the platform-specific docs to learn how to setup authentication on your pla
 - [github.com and GitHub Enterprise Server](../modules/platform/github/index.md)
 - [GitLab](../modules/platform/gitlab/index.md)
 
-### GitHub.com token for changelogs
+### GitHub.com token for meta data
 
 If you are running on any platform except github.com, you should also set the environment variable `GITHUB_COM_TOKEN` and put the Personal Access Token for github.com in it.
 This account can be _any_ account on GitHub, and needs only `read-only` access.
-It's used when fetching changelogs for repositories in order to increase the hourly API limit.
+It's used to increase the hourly API limit when fetching changelogs or other contents from repositories and in required in some cases because GitHub does not permit anonymous GraphQL queries.
```
secustor, line 216:
```suggestion
It's used to increase the hourly API limit when fetching changelogs or other contents from repositories and is required in some cases because GitHub does not permit anonymous GraphQL queries.
```

lib/workers/global/config/parse/env.ts
viceice: you can replace default branch with `HEAD` 🙃

secustor: TIL, though still the raw endpoint does not work for folder contents.

secustor: Also not, sure if this raw endpoints exist on GHE setups.

viceice (dismissed):
I'm still concerned that it will break exising users

lib/modules/datasource/bitrise/readme.md
```diff
@@ -24,3 +24,19 @@ workflows:
...
+```json title="Host Rule which matches the Bitrise step lib repository and datasource"
+{
+  "hostRules": [
+    {
+      "hostType": "bitrise",
+      "matchHost": "https://api.github.com/repos/my-org/my-repo/contents"
+    }
+  ]
+}
+```
```
rarkins, line 42: I don't like having a host rule in docs which "does nothing". Would adding a dummy token here be appropriate?

viceice (dismissed):
I still think this can cause harm 🤷‍♂️

secustor:
What do you expect to break / cause harm? 🤔

The token is scoped to the repository.

viceice:
> What do you expect to break / cause harm? 🤔
> 
> The token is scoped to the repository.

I'm not sure. it's a feeling.

effektsvk:
@secustor Why was this closed? Is it not getting merged?

secustor:
I have accidentally closed deleted the branch, while cleaning up my fork.

jamietanna:
For anyone interested, this will be replaced by https://github.com/renovatebot/renovate/pull/38733 which auto-wires the authentication for any `api.github.com` API calls, if there is a `hostRule` setting it up

PR #30491, opened 2024-07-30
feat(docker): Reconcile deleted releases in cache

### LINKED ISSUE #30442: Docker Hub tags cache: reconcile should prune removed tags

Likely the root cause of https://github.com/renovatebot/renovate/issues/30441

If a tag is removed, our reconcile function will never remove it, and it will be persisted in the cached results forever.

Quick win for images with <=1000 tags (single page of results): overwrite this.cache.items completely. The only thing is that we ideally don't "churn" the cache unnecessarily so we should do an intelligent compare. But ultimately if there's a single page of results then we should _synchronize_ them and not _reconcile_ - the new items should be the source of truth.

For images with >1000 tags, could we use the updated_at field to determine ones which have been removed, even if they are in pages 2+?

Here's my idea:
- Pass through the `expectedCount` to cache.reconcile
- At the end of each reconcile:
  - Build a set of `id` numbers from the new items
  - Calculate the newest and oldest `updated_at` dates from the new items
  - Iterate through the cache items and if their date falls between the newest/oldest dates from new items but their ID is not found in the set then remove it from the cache
  - If the expectedCount still doesn't match then return needNewMatch=true

Example starting code: https://github.com/renovatebot/renovate/compare/feat/docker-hub-cache-mismatch?expand=1

zharinov:
## Context

- Closes: #30442


lib/modules/datasource/docker/dockerhub-cache.ts
```diff
@@ -32,32 +32,64 @@ export class DockerHubCache {
...
 
+    const ids = new Set<number>();
+
     for (const newItem of items) {
       const id = newItem.id;
+      ids.add(id);
+
       const oldItem = this.cache.items[id];
 
+      const itemDate = DateTime.fromISO(newItem.last_updated);
```
rarkins, line 51: Do we need validation here? Maybe DateTime.fromISO could fail?

zharinov: Doesn't happen in practice it seems, but maybe in separate PR we could do it just in case (on schema level)

rarkins: Are you validating the date using zod instead?

zharinov: No, we don't, so far it goes without validation

rarkins: In that case please wrap it in try/catch because I think it would throw if encountering an invalid string

zharinov: It doesn't throw, it returns special invalid `DateTime` instance

zharinov: Ah, yes, we did the string validation at the schema level

rarkins (dismissed):

lib/modules/datasource/docker/dockerhub-cache.ts
```diff
@@ -32,32 +33,66 @@ export class DockerHubCache {
...
-        latestDate = newItemDate;
-      }
-
       this.isChanged = true;
     }
 
     this.cache.updatedAt = updatedAt;
+
+    if (earliestDate && latestDate) {
+      for (const key of Object.keys(this.cache.items)) {
```
viceice, line 73:
```suggestion
      for (const [key, item] of Object.entries(this.cache.items)) {
```

lib/modules/datasource/docker/dockerhub-cache.ts
```diff
@@ -32,32 +33,66 @@ export class DockerHubCache {
...
+        const id = parseInt(key, 10);
+        const item = this.cache.items[id];
```
viceice, line 75:
```suggestion
```

viceice (commented):
otherwise LGTM

lib/modules/datasource/docker/dockerhub-cache.ts
```diff
@@ -32,32 +33,65 @@ export class DockerHubCache {
...
-      }
-
       this.isChanged = true;
     }
 
     this.cache.updatedAt = updatedAt;
+
+    if (earliestDate && latestDate) {
+      for (const [key, item] of Object.entries(this.cache.items)) {
+        const id = parseInt(key, 10);
```
viceice, line 74:
```suggestion
        const id = item.id;
```
does this work?

zharinov: Type error: `string` to `number`

viceice (approved):

secustor (approved):

PR #30556, opened 2024-08-02
fix: disable docker:pinDigests for ArgoCD

morremeyer:
## Changes

This disables Docker digest pinning for the ArgoCD manager in the `docker:pinDigests` preset.

## Context

The `argocd` manager uses the Docker datasource [only for helm chart target versions](https://github.com/renovatebot/renovate/blob/d7989b0de840074d596fc1784e14e57467cf8a99/lib/modules/manager/argocd/extract.ts#L59-L70).

However, helm does not yet support digest pinning, see

- https://github.com/helm/helm/issues/10312#issuecomment-961396434
- https://github.com/helm/helm/issues/10678

Trying to pin them anyway leads to `WARN: Error updating branch: update failure` warnings from renovate.

The same applies to the `helmv3` manager, when it tries to pin the chart version of an OCI chart in a `Chart.yaml` file.

Because pinning is not supported, pinning digests for helm charts should be disabled in these managers.

I updated the Anaconda base configuration in https://github.com/anaconda/renovate-config/pull/101 to have the same configuration as this PR, which was successful for our repositories and resolved all `WARN: Error updating branch: update failure` errors.

### DIFF
--- a/lib/config/presets/internal/docker.ts
+++ b/lib/config/presets/internal/docker.ts
@@ -43,7 +43,7 @@ export const presets: Record<string, Preset> = {
         pinDigests: true,
       },
       {
-        matchManagers: ['devcontainer', 'pyenv'],
+        matchManagers: ['argocd', 'devcontainer', 'helmv3', 'pyenv'],
         pinDigests: false,
       },
     ],


morremeyer:
Drafting this again since I may have found additional managers impacted by this.

morremeyer:
Updated the PR to also not pin digests for the `helmv3` manager, since this is also not supported by helm.

rarkins (approved):

PR #30559, opened 2024-08-02
Feat/manager circleci executors parsing

tlbraams:
## Changes

Add executors to the circleci manager schema to return dependency detection. Docker images defined in executors were detected before a refactor in #30142. This returns support for these dependencies.

## Context

closes #30558 


tlbraams:
Just saw that you also created a PR. Since I can't use this account to sign the CLA, I have multiple email's associated with this account and it's not clear to me which of those will be used/linked to the CLA, I was about to recreate these changes under a different account which doesn't have those issues.

To me it looks to be identical, so your PR would be fine.

PR #30569, opened 2024-08-03
docs(manager/gleam): fix readme formatting

SnakeDoc:
## Changes

Correct a formatting error in the gleam manager readme, caused by prettier, which resulted in the example sub-lists rendering as parent list items.

<details>
<summary>Current Render:</summary>

image
</details> 

## Context

Fix markdown formatting and have prettier ignore the block. Tested using mkdocs locally.

<details>
<summary>Fixed Render:</summary>

image
</details>

### DIFF
--- a/lib/modules/manager/gleam/readme.md
+++ b/lib/modules/manager/gleam/readme.md
@@ -15,12 +15,13 @@ The `gleam` manager, however, uses `gleam` itself to keep `manifest.toml` up-to-
 
 Renovate's `"auto"` strategy defaults to `"widen"` and works like this for `gleam`:
 
+<!-- prettier-ignore -->
 1. If an existing range is a complex range (contains multiple range specifications), Renovate widens it to include the new version.
-   - Example: `>= 0.14.0 and < 0.15.0` becomes `>= 0.14.0 and < 0.16.1` for a new `0.16.0` version.
+    - Example: `>= 0.14.0 and < 0.15.0` becomes `>= 0.14.0 and < 0.16.1` for a new `0.16.0` version.
 1. For simple ranges, if the update is outside the existing range, Renovate widens the range to include the new version.
-   - Example: `<= 0.38.0` becomes `<= 0.39.0` for a new `0.39.0` version.
+    - Example: `<= 0.38.0` becomes `<= 0.39.0` for a new `0.39.0` version.
 1. For exact version constraints, Renovate replaces the version with the new one.
-   - Example: `== 0.12.0` becomes `== 0.13.0` for a new `0.13.0` version.
+    - Example: `== 0.12.0` becomes `== 0.13.0` for a new `0.13.0` version.
 
 <!-- prettier-ignore -->
 !!! warning


rarkins (approved):

rarkins:
Is this a known `prettier` problem? I dislike needing to have the `prettier-ignore` permanently

SnakeDoc:
I was surprised myself, actually. Without the ignore directive, it removes 1 space char in the sub-list (leaving 3 instead of 4 spaces of indentation). It still renders as expected in whatever markdown renderer Github uses as well as my ide, but somehow mkdocs loses it in translation.

rarkins:
Waiting in case @HonkingGoose has any insights

SnakeDoc:
Looking at it a bit more, it does look like prettier has a default [`tabWidth`](https://prettier.io/docs/en/options.html#tab-width) of 2 space chars, which seems to cause this issue with embedded sub-lists for a reason I don't yet understand.

When I add `"tabWidth": 4` to `.prettierrc.json`, then the `prettier-ignore` is unnecessary. However, setting the tab width to anything but default might have a significant impact. Perhaps HonkingGoose will know more.

HonkingGoose:
## Summary

We use MkDocs plus the Material for MkDocs extension to build our docs. MkDocs uses Python-Markdown, which mostly matches the Markdown spec.

But Python-Markdown is _different_ in:

- Middle-Word Emphasis
- Indentation/Tab Length (MkDocs forces `4 spaces`, instead of letting you do whatever)
- Consecutive Lists (MkDocs ignores list marker changes, which is _good_ as Prettier forces it all to `1.` or `-` anyways)

The problem with setting `tabWidth` to `4 spaces` is that ordered lists seem to need 4 and 2 spaces! At least that's what the Material for MkDocs examples use.

I don't know if setting `tabWidth` to `4 spaces` is better than what we do now. We should try it in a new PR though, to see what changes, and then decide what's best. :wink: 

## MkDocs uses Python-Markdown

Quote from [MkDocs, writing with Markdown](https://www.mkdocs.org/user-guide/writing-your-docs/#writing-with-markdown):

> MkDocs uses the [Python-Markdown](https://python-markdown.github.io/) library to render Markdown documents to HTML. Python-Markdown is almost completely compliant with the [reference implementation](https://daringfireball.net/projects/markdown/), although there are a few very minor [differences](https://python-markdown.github.io/#differences).

## Differences in Python-Markdown

Quote from [Python-Markdown, differences](https://python-markdown.github.io/#differences):

> #### Differences
> 
> While Python-Markdown strives to fully implement markdown as described in the [syntax rules](https://daringfireball.net/projects/markdown/syntax), the rules can be interpreted in different ways and different implementations occasionally vary in their behavior (see the [Babelmark FAQ](https://johnmacfarlane.net/babelmark2/faq.html#what-are-some-examples-of-interesting-divergences-between-implementations) for some examples). Known and intentional differences found in Python-Markdown are summarized below:
> 
> ##### Middle-Word Emphasis
> 
> Python-Markdown defaults to ignoring middle-word emphasis (and strong emphasis). In other words, some_long_filename.txt will not become some<em>long</em>filename.txt. This can be switched off if desired. See the [Legacy EM Extension](https://python-markdown.github.io/extensions/legacy_em/) for details.
> 
> ##### Indentation/Tab Length
> 
> The [syntax rules](https://daringfireball.net/projects/markdown/syntax#list) clearly state that when a list item consists of multiple paragraphs, “each subsequent paragraph in a list item must be indented by either 4 spaces or one tab” (emphasis added). However, many implementations do not enforce this rule and allow less than 4 spaces of indentation. The implementers of Python-Markdown consider it a bug to not enforce this rule.
> 
> This applies to any block level elements nested in a list, including paragraphs, sub-lists, blockquotes, code blocks, etc. They must always be indented by at least four spaces (or one tab) for each level of nesting.
> 
> In the event that one would prefer different behavior, [tab_length](https://python-markdown.github.io/reference/#tab_length) can be set to whatever length is desired. Be warned however, as this will affect indentation for all aspects of the syntax (including root level code blocks). Alternatively, a [third party extension](https://github.com/Python-Markdown/markdown/wiki/Third-Party-Extensions) may offer a solution that meets your needs.
> 
> ##### Consecutive Lists
> 
> While the syntax rules are not clear on this, many implementations (including the original) do not end one list and start a second list when the list marker (asterisks, pluses, hyphens, and numbers) changes. For consistency, Python-Markdown maintains the same behavior with no plans to change in the foreseeable future. That said, the [Sane List Extension](https://python-markdown.github.io/extensions/sane_lists/) is available to provide a less surprising behavior.

## Material for MkDocs reference guide, lists

Examples from the [Material for MkDocs, reference, lists](https://squidfunk.github.io/mkdocs-material/reference/lists/).

### Ordered lists

Syntax:

- Start bullet at position `0`
- Sub list uses `4 spaces`

```markdown
- Nulla et rhoncus turpis. Mauris ultricies elementum leo. Duis efficitur
  accumsan nibh eu mattis. Vivamus tempus velit eros, porttitor placerat nibh
  lacinia sed. Aenean in finibus diam.

    * Duis mollis est eget nibh volutpat, fermentum aliquet dui mollis.
    * Nam vulputate tincidunt fringilla.
    * Nullam dignissim ultrices urna non auctor.
```

### Ordered lists

Syntax:

1. Start with list marker `1.` at position `0`
1. Sub-list starts with `4 spaces` then the `1.` list marker, then `2 spaces, then the list text

```markdown
1.  Vivamus id mi enim. Integer id turpis sapien. Ut condimentum lobortis
    sagittis. Aliquam purus tellus, faucibus eget urna at, iaculis venenatis
    nulla. Vivamus a pharetra leo.

    1.  Vivamus venenatis porttitor tortor sit amet rutrum. Pellentesque aliquet
        quam enim, eu volutpat urna rutrum a. Nam vehicula nunc mauris, a
        ultricies libero efficitur sed.

    2.  Morbi eget dapibus felis. Vivamus venenatis porttitor tortor sit amet
        rutrum. Pellentesque aliquet quam enim, eu volutpat urna rutrum a.

        1.  Mauris dictum mi lacus
        2.  Ut sit amet placerat ante
        3.  Suspendisse ac eros arcu
```

lib/modules/manager/gleam/readme.md
```diff
@@ -15,12 +15,13 @@ The `gleam` manager, however, uses `gleam` itself to keep `manifest.toml` up-to-
 
 Renovate's `"auto"` strategy defaults to `"widen"` and works like this for `gleam`:
 
+<!-- prettier-ignore -->
```
HonkingGoose, line 18: I'm okay with this as hack-job.

But we could try setting `4 spaces` as `tabwidth` for Prettier, in a new PR. That way we can see if that works better for us, out of the box.

SnakeDoc: I would be more than happy to create a PR using `"tabWidth": 4`, however this will make prettier reformat the entire repo and start enforcing a non-standard tab width going forward. It is also possible such a change could unintentionally break other, existing docs as well.

Lastly, I suspect setting `"tabWidth": 4` might be more of a work around for this specific prettier & mkdocs incompatibility rather than being the objectively correct thing to do.

With all of that said, may I propose we leave the `prettier-ignore` in-place and add a comment/todo referencing this PR as a form of documentation for this specific format exception? If that is undesirable, I could instead explore re-writing that docs section to forgo any embedded lists.

As an outside contributor, I'm unsure which direction would be most preferable from the project's perspective. Waiting on insight/guidance from @rarkins and @HonkingGoose

viceice: we can use the editor config to limit it to markdown files

SnakeDoc: That is a good point @viceice. However, even with only limiting the changes to markdown files, I would be concerned with the risk of breaking other existing docs. The rendering issue here was subtle, caught me off guard, and snuck through a code review, so the concern would be that we wouldn't be able to notice/catch other new rendering issues after the formatting change.

I very well could be overthinking it though! Let me know if the consensus is we should explore this path - I'm happy to put together a new PR.

I suppose if Renovate is exploring a tabWidth change in prettier (even for just markdown files), it may be relevant to mention this ongoing discussion (https://github.com/prettier/prettier/issues/7475) regarding the use of tab chars instead of spaces for accessibility reasons. Personally, I like spaces, but have no other opinion on the subject.

SnakeDoc: OK, PR #30608 explores setting markdown `tabWidth: 4`.

I'll need help with that one, lots of changed files. From what I can see, it mostly changed what you would expect. However, it seems to do some odd things with root-level unordered lists.

Specifically, changing:

```
- List Item A
- List Item B
```

into:

```
-   List Item A
-   List Item B
```
(note the 3 space chars after the hyphen char)

mkdocs seems to render this as you would expect, even though it looks odd. Perhaps this is a readability choice made by prettier?

For embedded unordered lists, such as the `gleam` manager example, it does the correct thing:

```
1. List Item A
    - Example A
```

(note the leading 4 space chars before the hyphen char, followed by a single space char).

SnakeDoc:
While #30608 continues to be worked on as an overall fix for this issue (and general formatting update for all markdown), would it be acceptable to merge this PR in the meantime?

I can add a comment to the ignore directive and point to this issue as a reminder.

There is at least one other doc where this same issue occurs (the [`conan` manager](https://docs.renovatebot.com/modules/manager/conan/)). Perhaps we should close this PR, and open a new PR that addresses each instance of this issue (adding the same `prettier-ignore` + comment to each)?

viceice (approved):

HonkingGoose:
I'm happy to merge this PR, which `@viceice` already did. :upside_down_face: I suspect we'll be stuck on the proper fix for a while.

> I can add a comment to the ignore directive and point to this issue as a reminder.

Sound good, but you should link to the "proper fix PR" instead. :wink: 

> There is at least one other doc where this same issue occurs (the [conan manager](https://docs.renovatebot.com/modules/manager/conan/)).

Do you want to create a PR to apply the same "hack-job fix" to that document?

SnakeDoc:
Yes, I'll work on applying this "hack" to the conan manager, and will do my best to locate any other occurrences while we're at it.

PR #30618, opened 2024-08-06
feat(template): add `lookupArray` and `distinct` helpers

amezin:
## Changes

Add `lookupArray` and `distinct` template helpers.

## Context

I'm using `prBodyDefinitions` to link Renovate pull requests to Jira issues like this:

```js
{
  packageRules: [
    {
      matchPackageNames: ['...'],
      matchNewValue: '...',
      prBodyDefinitions: {
        Issue: 'ABC-123'
      },
      commitMessagePrefix: '{{#each upgrades}}{{{prBodyDefinitions.Issue}}}{{/each}}'
    }
  ]
}
```

But sometimes I get duplicate issue references (for example, when updating the version in multiple locations).

I thought about adding JSONata helper, but JSONata evaluation is async, and handlebars helpers must be sync. So the only option is to add multiple new helpers.


rarkins (commented):
Should both of these have defensive code added in case a non-array is passed?

docs/usage/templates.md
```diff
@@ -116,6 +116,28 @@ Returns `true` if the value is included on the list given.
 
 `{{#if (includes labels 'dependencies')}}Production Dependencies{{else}}Not Production Dependencies{{/if}}`
 
+### lookupArray
```
viceice, line 119: we should sort them alphabetically

amezin:
> Should both of these have defensive code added in case a non-array is passed?

Is `toArray()` not enough?

rarkins:
What's the result if a non array is passed to it?

amezin:
https://github.com/renovatebot/renovate/blob/main/lib/util/array.ts#L25

lib/util/template/index.ts
```diff
@@ -87,6 +88,31 @@ handlebars.registerHelper({
...
+    toArray(obj)
+      // skip elements like #with does
+      .filter((element) => !handlebars.Utils.isEmpty(element))
+      .map((element) => options.lookupProperty(element, key))
+      .filter((value) => value !== undefined)
+  );
+});
+
+handlebars.registerHelper('distinct', (obj) => {
+  const seen = new Set();
```
viceice, line 102:
```suggestion
  const seen = new Set<string>();
```

lib/util/template/index.ts
```diff
@@ -87,6 +88,31 @@ handlebars.registerHelper({
...
+  return (
+    toArray(obj)
+      // skip elements like #with does
+      .filter((element) => !handlebars.Utils.isEmpty(element))
+      .map((element) => options.lookupProperty(element, key))
+      .filter((value) => value !== undefined)
+  );
+});
+
+handlebars.registerHelper('distinct', (obj) => {
```
viceice, line 101:
```suggestion
handlebars.registerHelper('distinct', (obj: unknown) => {
```

lib/util/template/index.ts
```diff
@@ -87,6 +88,31 @@ handlebars.registerHelper({
   },
 });
 
+handlebars.registerHelper('lookupArray', (obj, key, options) => {
+  return (
```
viceice, line 92:
```suggestion
    if (
      !obj ||
      !is.nonEmptyStringAndNotWhitespace(key) ||
      !options ||
      !('lookupProperty' in options)
    ) {
      return [];
    }

  return (
```

validation

lib/util/template/index.ts
```diff
@@ -87,6 +88,31 @@ handlebars.registerHelper({
   },
 });
 
+handlebars.registerHelper('lookupArray', (obj, key, options) => {
```
viceice, line 91:
```suggestion
handlebars.registerHelper(
  'lookupArray',
  (obj: unknown, key: unknown, options: Options) => {
```

lib/util/template/index.ts
```diff
@@ -2,6 +2,7 @@ import is from '@sindresorhus/is';
 import handlebars from 'handlebars';
```
viceice, line 2:
```suggestion
import handlebars, { type HelperOptions } from 'handlebars';
```

lib/util/template/index.ts
```diff
@@ -2,6 +2,7 @@ import is from '@sindresorhus/is';
...
 import { GlobalConfig } from '../../config/global';
 import { logger } from '../../logger';
+import { toArray } from '../array';
 import { getChildEnv } from '../exec/utils';
 import { regEx } from '../regex';
```
viceice, line 7:
```suggestion
import { regEx } from '../regex';

// Missing in handlebars
type Options = HelperOptions & {
  lookupProperty: (element: unknown, key: unknown) => unknown;
};
```

lib/util/template/index.ts
amezin: Handlebars' own `lookup` does none of this: https://github.com/handlebars-lang/handlebars.js/blob/master/lib/handlebars/helpers/lookup.js

https://handlebarsjs.com/api-reference/helpers.html `lookupProperty` isn't described as "optional" - as far as I understand, it's always passed and is part of the API.

Null key doesn't cause any error. Handlebars' `lookup` helper simply passes key to `lookupProperty` as is.

If obj is null, it'll be converted to `[null]` and then filtered out by `!handlebars.Utils.isEmpty()`.

And if I'm passing wrong number of arguments, I'd prefer an error, not just silent empty result.

viceice: add tests

amezin: I already did it: https://github.com/renovatebot/renovate/pull/30618/commits/5dba00c1629abcb546b2ef9079e5d6412bdf3a87

is anything missing?

lib/util/template/index.spec.ts
```diff
@@ -423,4 +423,104 @@ describe('util/template/index', () => {
...
+            },
+          ],
+        },
+        false,
+      );
+
+      expect(output).toBe(' ABC-123');
+    });
+
+    it('handles null key', () => {
```
viceice, line 481: test null object

lib/util/template/index.spec.ts
```diff
@@ -423,4 +423,104 @@ describe('util/template/index', () => {
...
+        },
+        false,
+      );
+
+      expect(output).toBe(' ABC-123');
+    });
+  });
+
+  describe('distinct', () => {
+    it('skips duplicate values', () => {
```
viceice, line 499: test null object

lib/util/template/index.spec.ts
amezin: There is already "handles null input array" test, isn't it doing what you're asking?

And "performs lookup for every array element" test already tests null elements.

viceice: ok

viceice (approved):

PR #30623, opened 2024-08-06
feat: Stats for datasource cache

### LINKED ISSUE #30420: Log datasource hit/miss stats per repo

We need datasource results to return in indicator if a cached result was returned or not, and then at the end of a repo we should log the cached vs non-cached count plus the percentage (in semantic format so that it can be collected/graphed by logging systems).

zharinov:
## Context

- Closes: #30420

### DIFF
--- a/lib/util/stats.ts
+++ b/lib/util/stats.ts
@@ -105,6 +105,128 @@ export class PackageCacheStats {
   }
 }
 
+interface DatasourceCacheDataPoint {
+  datasource: string;
+  registryUrl: string;
+  packageName: string;
+  action: 'hit' | 'miss' | 'set' | 'skip';
+}
+
+export interface DatasourceCacheReport {
+  long: {
+    [datasource in string]: {
+      [registryUrl in string]: {
+        [packageName in string]: {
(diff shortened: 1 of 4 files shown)


lib/util/stats.ts
```diff
@@ -105,6 +105,128 @@ export class PackageCacheStats {
...
+interface DatasourceCacheDataPoint {
+  datasource: string;
+  registryUrl: string;
+  packageName: string;
+  action: 'hit' | 'miss' | 'set' | 'skip';
+}
+
+export interface DatasourceCacheReport {
+  long: {
+    [datasource in string]: {
```
viceice, line 117: looks ugly, can we use `Record`?

zharinov: It's ugly, but it better communicates the name/purpose of the particular level of nesting

rarkins:
Does it log any messages yet?

lib/util/stats.ts
viceice: I still don't like that

viceice (approved):

rarkins (approved):

PR #30639, opened 2024-08-07
feat(preset): add `phpstan/extension-installer` to PHPStan group

RobertMe:
## Changes

Add `phpstan/extension-installer` to PHPStan group

## Context

As PHPStan extensions are part of the group it makes sense to also include the extension installer itself in this group.

### DIFF
--- a/lib/config/presets/internal/group.ts
+++ b/lib/config/presets/internal/group.ts
@@ -354,7 +354,12 @@ const staticGroups = {
       {
         groupName: 'PHPStan packages',
         matchDatasources: ['packagist'],
-        matchPackageNames: ['phpstan/phpstan', '//phpstan-/', '//larastan/', 'phpstan/extension-installer'],
+        matchPackageNames: [
+          'phpstan/phpstan',
+          '//phpstan-/',
+          '//larastan/',
+          'phpstan/extension-installer',
+        ],
       },
     ],
   },
--- a/lib/config/presets/internal/group.ts
+++ b/lib/config/presets/internal/group.ts
@@ -354,7 +354,7 @@ const staticGroups = {
       {
         groupName: 'PHPStan packages',
         matchDatasources: ['packagist'],
-        matchPackageNames: ['phpstan/phpstan', '//phpstan-/', '//larastan/'],
+        matchPackageNames: ['phpstan/phpstan', '//phpstan-/', '//larastan/', 'phpstan/extension-installer'],
       },
     ],
   },


lib/config/presets/internal/group.ts
```diff
@@ -354,7 +354,7 @@ const staticGroups = {
       {
         groupName: 'PHPStan packages',
         matchDatasources: ['packagist'],
-        matchPackageNames: ['phpstan/phpstan', '//phpstan-/', '//larastan/'],
+        matchPackageNames: ['phpstan/phpstan', '//phpstan-/', '//larastan/', 'phpstan/extension-installer'],
```
viceice, line 357:
```suggestion
        matchPackageNames: ['phpstan/**', '//phpstan-/', '//larastan/'],
```

why not group all `phpstan/**` packages?

RobertMe: While possible, it _might_ lead to some undesired effects I think. I.e. at least `phpstan/phpdoc-parser` [is used by other projects as well](https://packagist.org/packages/phpstan/phpdoc-parser/dependents?order_by=downloads) so updating that as part of what seems like an update of PHPStan the "code analyses tool" might / will impact "production code" as well.

In other words, PHPStan and extensions for it would be development dependencies. While (at least) `phpstan/phpdoc-parser` _could_ be an actual dependency (direct or indirect) as well. Or to put it differently: `phpstan/phpdoc-parser` isn't a part of the tool PHPStan (as all the other things in this groups are extensions for the tool etc). It's a "shared library" used by the tool, which happens to be maintained by the developer(s) of the tool, but reaches (far) outside just the tool. And if `phpstan/phpdoc-parser` was maintained by some other person/organization it wouldn't/shouldn't be part of the group either. (But extensions developed by others are matched by the group because of the `/phpstan-` / `/larastan` match, as those are closely related to the tool)

(But at the same time, as Renovate (by default?) does "update including dependencies" it is very likely that an update of PHPStan will also update `phpstan/phpdoc-parser` as well, because PHPStan itself depends on it)

RobertMe: The above actually isn't even totally correct. This as PHPStan (`phpstan/phpstan`) is a "prebuild" package which contains all its dependencies "build in" (in PHP terms, it's a [phar](https://www.php.net/manual/en/intro.phar.php)). So when requiring `phpstan/phpstan` there are no additional dependencies installed. So updating `phpstan/phpstan` will on it's own never need, nor lead to, updates of any other package, not even `phpstan/phpdoc-parser` (as the phar archive contains all dependencies the version needs). So updating `phpstan/phpdoc-parser` as part of updating `phpstan/phpstan` will IMO always be "wrong" as as far as Composer is concerned there is no dependency between both packages (and so there shouldn't be any for Renovate neither).

viceice (approved):

PR #30708, opened 2024-08-12
fix(manager/copier): process update templates outside repository root

PJ-Schulz:
## Changes

I fixed the behavior described in #30684 where the update from a copier dependency takes place in the wrong path. I changed the function where the `copier update` command is created. I split up the path to the `.copier-answers` file give with `packageFileName` at the last separator and use the path to the answer file as destination and the last part as name for the `copy-answers` file

## Context

#30684 

### DIFF
--- a/lib/modules/manager/copier/artifacts.spec.ts
+++ b/lib/modules/manager/copier/artifacts.spec.ts
@@ -120,7 +120,7 @@ describe('modules/manager/copier/artifacts', () => {
           cmd: 'copier update --skip-answered --defaults --answers-file .copier-answers.yml --vcs-ref 1.1.0',
         },
       ]);
-      expect(execSnapshots[0].options?.cwd).toBe('.');
+      expect(execSnapshots[0].options?.cwd).toBe('/tmp/github/some/repo');
     });
 
     it('invokes copier update with nested destination and answer file', async () => {
@@ -138,7 +138,9 @@ describe('modules/manager/copier/artifacts', () => {
           cmd: 'copier update --skip-answered --defaults --answers-file .copier-answers.yml --vcs-ref 1.1.0',
         },
       ]);
-      expect(execSnapshots[0].options?.cwd).toBe('apps/my-app');
+      expect(execSnapshots[0].options?.cwd).toBe(
+        '/tmp/github/some/repo/apps/my-app',
+      );
     });
 
     it.each`
--- a/lib/modules/manager/copier/artifacts.spec.ts
+++ b/lib/modules/manager/copier/artifacts.spec.ts
@@ -118,9 +118,11 @@ describe('modules/manager/copier/artifacts', () => {
       expect(execSnapshots).toMatchObject([
         {
           cmd: 'copier update --skip-answered --defaults --answers-file .copier-answers.yml --vcs-ref 1.1.0',
+          options: {
+            cwd: '/tmp/github/some/repo',
+          },
         },
       ]);
-      expect(execSnapshots[0].options?.cwd).toBe('/tmp/github/some/repo');
     });
 
     it('invokes copier update with nested destination and answer file', async () => {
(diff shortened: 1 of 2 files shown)


lib/modules/manager/copier/artifacts.ts
```diff
@@ -26,12 +26,29 @@ function buildCommand(
...
+  }
   command.push(
     '--answers-file',
-    quote(packageFileName),
+    quote(answersFile),
     '--vcs-ref',
-    quote(newVersion),
+    quote(newVersion)
   );
+  // Only push destination if it is not an empty string
```
HonkingGoose, line 48:
```suggestion
  // Only push to destination if the string has content
```

HonkingGoose (changes requested):

lib/modules/manager/copier/artifacts.spec.ts
```diff
@@ -122,6 +122,23 @@ describe('modules/manager/copier/artifacts', () => {
       ]);
     });
 
+    it('invokes copier update with nested destination and answer file', async () => {
```
HonkingGoose, line 125: Please change the PR title to:

`fix(manager/copier): process update templates outside repository root`

HonkingGoose (changes requested):
Not 100% sure on the correct PR title, but this should be better than what we have now. 😄

lib/modules/manager/copier/artifacts.ts
```diff
@@ -26,12 +26,29 @@ function buildCommand(
...
+  // Check if there is a separator in packageFileName
+  const lastSeparatorIndex = packageFileName.lastIndexOf('/');
+  if (lastSeparatorIndex === -1) {
+    // No separator found
+    destination = "";
+    answersFile = packageFileName;
+  } else {
+    // Separator found, split into destination and answersFile
+    destination = packageFileName.slice(0, lastSeparatorIndex);
+    answersFile = packageFileName.slice(lastSeparatorIndex + 1);
```
viceice, line 40: use `upath` module functions to split path from filename

viceice (changes requested):
you can also pass package file name to exec options, so the exec function, so it will use it as cwd

PJ-Schulz:
I changed it to use upath and put the destination as cwd to the execOptions, but I do not know how to test it now. Do you have a suggestion?

PJ-Schulz:
I've added a test for checking `cmd` on execOptions. Can I test it this way, or should I test it another way?

lib/modules/manager/copier/artifacts.ts
```diff
@@ -72,6 +73,7 @@ export async function updateArtifacts({
 
   const command = buildCommand(config, packageFileName, newVersion);
   const execOptions: ExecOptions = {
+    cwd: dirname(packageFileName),
```
viceice, line 76:
```suggestion
    cwdFile‎: packageFileName,
```
even simpler 😉

lib/modules/manager/copier/artifacts.ts
```diff
@@ -1,4 +1,5 @@
 import { quote } from 'shlex';
+import { basename, dirname } from 'upath';
```
viceice, line 2:
```suggestion
import upath from 'upath';
```
otherwise we'll get esm import issues later

lib/modules/manager/copier/artifacts.spec.ts
```diff
@@ -120,6 +120,27 @@ describe('modules/manager/copier/artifacts', () => {
...
+      });
+
+      expect(execSnapshots).toMatchObject([
+        {
+          cmd: 'copier update --skip-answered --defaults --answers-file .copier-answers.yml --vcs-ref 1.1.0',
+        },
+      ]);
+      expect(execSnapshots[0].options?.cwd).toBe(
+        '/tmp/github/some/repo/apps/my-app',
+      );
```
viceice, line 143:
```suggestion
          options: {
            cwd: '/tmp/github/some/repo/apps/my-app',
          }
        },
      ]);
```

use a single matcher

viceice (approved):

PR #30716, opened 2024-08-12
fix(datasource/pypi): handle non-normalized package names for pypi simple lookup

### LINKED ISSUE #30712: datasource(pypi): simple pypi lookup is not working for packages with normalized names
### Discussed in https://github.com/renovatebot/renovate/discussions/30707

<sup>Originally posted by **Shegox** August 12, 2024</sup>
### What would you like help with?

I think I found a bug

Self-hosted

38.25.1

Renovate, after the v38 update, seems to be no longer able to lookup packages names containing special characters (e.g. `.`) when looking them up in a simple registry (e.g. in a private Artifactory). 

The underlaying problem seems that Renovate is normalizing the name (e.g. replacing `.` with `-`) and only uses the `-`-name to perform the lookup). 

As a concrete example the `ruamel.yaml` package is normalized to `ruamel-yaml` and Renovate then fails to query the datasource.

I think this might be related to https://github.com/renovatebot/renovate/pull/27733

- Reproduction repository (with lookup against `https://www.pypi.org/simple` as this forces Renovate to use the simple lookup method): https://github.com/Shegox/congenial-funicular
- Issue in dependency dashboard: `Renovate failed to look up the following dependencies: Failed to look up pypi package ruamel-yaml.`
- Other non-normalized packages like `requests` work as expected: https://github.com/Shegox/congenial-funicular/pull/4

Notice: Lookups against the standard pypi.org registry use the pypi-API directly and are not affected.

---

I think the problem is within the `extractVersionFromLinkText`-function, which takes the packageName, normalizes it and then uses it for the lookup: 
https://github.com/renovatebot/renovate/blob/6d30cc968033cb62b7e21cc452024ce370c14c0e/lib/modules/datasource/pypi/index.ts#L178-L185

Digging deeper it seems that the `PypiDatasource.normalizeName(text)`-function only normalizes `_` to `-` instead of the additional `.` to `-` as requested by https://packaging.python.org/en/latest/specifications/name-normalization/ : 
https://github.com/renovatebot/renovate/blob/6d30cc968033cb62b7e21cc452024ce370c14c0e/lib/modules/datasource/pypi/index.ts#L85-L87

We do have a correct normalization function in 
https://github.com/renovatebot/renovate/blob/dbd17ed1d78595631b10481e9bf97815fb9acccd/lib/modules/datasource/pypi/common.ts#L9-L12

The problem is however that this will as well normalize a `srcText` like `ruamel.yaml-0.1.tar.gz` to `ruamel-yaml-0-1-tar-gz`. 
We can adjust however our compare logic within the `extractVersionFromLinkText` to do this correctly. 

If there are no objections I will create a PR for this.

### Logs (if relevant)

```

DEBUG: hostRules: no authentication for www.pypi.org (repository=Shegox/congenial-funicular)
DEBUG: Using queue: host=www.pypi.org, concurrency=16 (repository=Shegox/congenial-funicular)
DEBUG: Failed to look up pypi package ruamel-yaml (repository=Shegox/congenial-funicular, packageFile=requirements.txt, dependency=ruamel-yaml)
DEBUG: PackageFiles.add() - Package file saved for base branch (repository=Shegox/congenial-funicular, baseBranch=main)
DEBUG: Package releases lookups complete (repository=Shegox/congenial-funicular, baseBranch=main)
DEBUG: branchifyUpgrades (repository=Shegox/congenial-funicular)
DEBUG: detectSemanticCommits() (repository=Shegox/congenial-funicular)
DEBUG: getCommitMessages (repository=Shegox/congenial-funicular)
DEBUG: semanticCommits: detected "unknown" (repository=Shegox/congenial-funicular)
DEBUG: semanticCommits: disabled (repository=Shegox/congenial-funicular)
DEBUG: 1 flattened updates found: requests (repository=Shegox/congenial-funicular)
DEBUG: Returning 1 branch(es) (repository=Shegox/congenial-funicular)
DEBUG: config.repoIsOnboarded=true (repository=Shegox/congenial-funicular)
DEBUG: packageFiles with updates (repository=Shegox/congenial-funicular, baseBranch=main)
       "config": {
         "pip_requirements": [
           {
             "deps": [
               {
                 "depName": "ruamel.yaml",
                 "packageName": "ruamel-yaml",
                 "currentValue": "==0.18.5",
                 "datasource": "pypi",
                 "currentVersion": "0.18.5",
                 "updates": [],
                 "versioning": "pep440",
                 "warnings": [
                   {
                     "topic": "ruamel-yaml",
                     "message": "Failed to look up pypi package ruamel-yaml"
                   }
                 ]
               },

```

</details>
</div>

Shegox:
## Changes

This PR revises the logic for extracting the version from `extractVersionFromLinkText`. The challenge arises because the HTML tag contains both the non-normalized name and the version, making it difficult to discern where the name ends and the version begins. This issue became prominent after [PR #27733/v38](https://github.com/renovatebot/renovate/pull/27733), where we only have the normalized name available during lookup. Thus, it's necessary to verify if the detected name matches the normalized name.

The approach involves normalizing the tag text and confirming it matches the `packageName`. Once verified, the package name is removed from the non-normalized tag name using the length of the normalized name to isolate the version. It's essential to note that the normalized name can differ in length, and this discrepancy is accounted for.

Additionally, the `normalizeName` function has been removed, and we now rely on the centralized implementation of `normalizePythonDepName`, which adheres to the standard normalization procedure.

We support two formats:
1. Source distribution: `{name}-{version}.tar.gz` ([specification](https://packaging.python.org/en/latest/specifications/source-distribution-format/#source-distribution-file-name))
2. Binary distribution: `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl` ([specification](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-name-convention))

Both `name` and `distribution` correspond to the `packageName` with `-` replaced by `_`. A newer spec replaces `.` with `_`. However, since we convert back to the normalized form and transform all three characters `_-.` to `-`, the comparison remains unaffected.

It’s also worth noting that package names with multiple instances of `.-_` reduced to `-` are rare. Across all of PyPI, there are only 29 such packages, with most being empty:
```
$ curl https://pypi.org/simple/ -sS | grep -P '[\.\-_]{2,3}'
<a href="/simple/0-0/">0-._.-._.-._.-._.-._.-._.-0</a>
<a href="/simple/ahmed-m-gamaleldin/">Ahmed-M.-Gamaleldin</a>
<a href="/simple/al-application-launcher/">AL---Application-Launcher</a>
<a href="/simple/cmc-csci046-data-structures/">cmc-csci046-.data-structures</a>
<a href="/simple/cmc-csci046-yilinli-trees/">cmc-csci046-.yilinli-trees</a>
<a href="/simple/empy-electromagnetic-python/">EMpy----ElectroMagnetic-Python</a>
<a href="/simple/e-s-p-hadouken/">E.S.P.-Hadouken</a>
<a href="/simple/example-pkg-testing-megankuoo/">example-pkg.....testing-megankuoo</a>
<a href="/simple/funniest-test2016/">funniest__test2016</a>
<a href="/simple/h-ello-worl-d/">h__ello__worl__d</a>
<a href="/simple/hydrotools-restclient/">hydrotools.-restclient</a>
<a href="/simple/iaf-interaction-framework/">IAF--Interaction-Framework</a>
<a href="/simple/jungle-py-compiler/">Jungle-.Py-Compiler</a>
<a href="/simple/just-a-try-i/">just-a-try--i</a>
<a href="/simple/jy-2019/">jy-.-2019</a>
<a href="/simple/lh-nester/">lh__nester</a>
<a href="/simple/liuhao-nester/">liuhao__nester</a>
<a href="/simple/micropython-ctypes/">micropython-_ctypes</a>
<a href="/simple/micropython-markupbase/">micropython-_markupbase</a>
<a href="/simple/nester-101/">nester_-101</a>
<a href="/simple/pycopy-ctypes/">pycopy-_ctypes</a>
<a href="/simple/pyhed-python-desktop-framework/">pyHed---Python-desktop-framework</a>
<a href="/simple/quick-torrent-downloader/">Quick-.Torrent-Downloader</a>
<a href="/simple/snowflake-legacy/">snowflake._legacy</a>
<a href="/simple/spyce-python-server-pages/">SPYCE---Python-Server-Pages</a>
<a href="/simple/tango-project-algencan/">TANGO-Project---ALGENCAN</a>
<a href="/simple/tzara-a-personal-assistant/">Tzara---A-Personal-Assistant</a>
<a href="/simple/util-q/">util--q</a>
<a href="/simple/vins-server-messenger/">Vins__server_messenger</a>
```

## Context

Fixes #30712, which caused Renovate to fail to lookup pypi packages with special characters (`.`) in their name.

- [x] Both unit tests + ran on a real repository (https://github.com/asdasfafsgsdfgdsfgdfg/congenial-funicular/pull/1)


Shegox:
//cc @not7cd  if you have some time I would highly appreciate your feedback/review of the pypi lookup logic.

lib/modules/datasource/pypi/__fixtures__/versions-html-with-periods.html
```diff
@@ -9,4 +9,6 @@ <h1>Links for package.with.periods</h1>
     <a href="">package.with.periods-2.0.1-py3-none-any.whl</a><br>
     <a href="">package_with_periods-2.0.2-py3-none-any.whl</a><br>
     <a href="">package.with.periods-2.0.2.tar.gz</a><br>
+    <a href="">package.with.periods-2.0.3.whl</a><br>
+    <a href="">package.with.periods-2.0.4</a><br>
```
not7cd, line 13: This is not obvious to me. Why no version is returned for those entries? Same question applies to: 
https://github.com/renovatebot/renovate/blob/2673867929da63e79468c6a407447e203e9ad733/lib/modules/datasource/pypi/__fixtures__/versions-html-with-non-normalized-name.html#L12-L13

I believe this is coming from this check:
https://github.com/renovatebot/renovate/blob/2673867929da63e79468c6a407447e203e9ad733/lib/modules/datasource/pypi/index.ts#L209

lib/modules/datasource/pypi/index.ts
```diff
@@ -180,42 +176,39 @@ export class PypiDatasource extends Datasource {
...
+    // source distribution format: `{name}-{version}.tar.gz` (https://packaging.python.org/en/latest/specifications/source-distribution-format/#source-distribution-file-name)
+    // binary distribution: `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl` (https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-name-convention)
+    // officially both `name` and `distribution` should be normalized and then the - replaced with _, but in reality this is not the case
+    // We therefore normalize the name we have (replacing `_-.` with -) and then check if the text starts with the normalized name
+
+    if (!normalizedSrcText.startsWith(srcPrefix)) {
+      return null;
+    }
+
+    // strip off the prefix using the prefix length as we have may have normalized the srcPrefix/packageName
```
not7cd, line 192: nit: typo? `have may have`

lib/modules/datasource/pypi/__fixtures__/versions-html-with-periods.html
Shegox: The idea is that according to the spec binary (aka wheel distributions) must look like this `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl`. So must inlcude a `python tag`, `abi tag` and `platform tag` as well as end with `.whl`

So a package like `package.with.periods-2.0.4` (without `.whl`) or a package like `package.with.periods-2.0.3.whl` without a python-tag is not valid. 

The check is not super strict, so something like `package.with.periods-2.0.3-py3.whl` would be valid, but I think we don't need to be too over specific here.

lib/modules/datasource/pypi/index.ts
Shegox: fixed :)

lib/modules/datasource/pypi/__fixtures__/versions-html-with-periods.html
not7cd: Personally, I would add a test case that states this explicitly. That these entries are not valid and must be not present in the output. I simply overlooked that these are not optional in this long name spec.

not7cd: On another note, are such entries present in real index? I'm no expert here.

Shegox: > On another note, are such entries present in real index? I'm no expert here.

Depends on the implementation, on pypi probably not, but I have seen super weird thing on our internal Artifactory. So if we can filter them out I think that makes sense.

lib/modules/datasource/pypi/index.spec.ts
```diff
@@ -453,11 +442,98 @@ describe('modules/datasource/pypi/index', () => {
...
+        { version: '0.7.0' },
+        { version: '0.8.0' },
+        { version: '0.8.1' },
+        { version: '0.9.0' },
+        { version: '0.10.0' },
+        { version: '0.11.0' },
+      ]);
+    });
+
+    it('reply with only valid valid version', async () => {
```
not7cd, line 491: It must be very valid then ;)
I propose this explicit statement:
```suggestion
    it('ignores invalid distribution file name formats', async () => {
```

not7cd (approved):
LGTM

viceice (approved):

PR #28671, opened 2024-04-25
feat(workers/repository): Update package files in the same order they're extracted

mbudnek:
## Changes

For some managers that can have dependencies between input files (pip-compile, for example) it's important to update each input file in the correct order.  This changes getUpdatedPackageFiles to sort its lists of package files into the same order that they were returned by the manager's extract function.

Since package files can be referenced by multiple managers and since each manager could return the package files in a different order I also inverted the update logic from
```
for each package file
    for each manager
        manager.update(package file)
```
to 
```
for each manager with updated package files
    sortPackageFiles
    for each package file
        manager.update(package file)
```

This lets us sort the list of package files only once per manager.

## Context

- #28518 

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

PR #30725, opened 2024-08-12
refactor(workers/repository): Reverse the update logic in getUpdatedPackageFiles

mbudnek:
## Changes

This PR is split off from #28671.  It reverses the logic in `getUpdatedPackageFiles` so that it runs all of the updates for each manager sequentially.

That is, instead of
```
for each package file
    for each manager
        manager.update(package file)
```
it will do
```
for each manager with updated package files
    for each package file
        manager.update(package file)
```

This refactor is the basis for a fix to #28518.  To resolve that issue, the package files need to be re-sorted to be in the same order that the manager extracted them.  Since a package file can be relevant to multiple managers, the list of package files needs to be re-sorted for each manager.  That isn't possible with the existing logic.

## Context

#28518 
#28671 


lib/workers/repository/update/branch/get-updated.ts
```diff
@@ -360,26 +398,36 @@ export async function getUpdatedPackageFiles(
     }
   }
   if (!reuseExistingBranch) {
+    const lockFileMaintenancePackageFiles: FileAddition[] =
+      lockFileMaintenanceFiles.map((name) => ({
+        type: 'addition',
+        path: name,
+        contents: null,
+      }));
```
viceice, line 406: why this is needed? 🤔

mbudnek: It's just to have a common type with the other two update loops so I can use the same `getManagersForPackageFiles` and `getPackageFilesForManager` for all three.  I originally tried to fill in the `contents` field there as well so that I could factor the entire loop out into a common function, but I couldn't get it working with an `await` inside the `map` call.

lib/workers/repository/update/branch/get-updated.ts
```diff
@@ -40,6 +40,36 @@ async function getFileContent(
...
 
+function hasAny(set: Set<string>, targets: Iterable<string>): boolean {
+  for (const target of targets) {
+    if (set.has(target)) {
+      return true;
+    }
+  }
+  return false;
+}
+
```
viceice, line 51:
```suggestion

type FilePath = Pick<FileChange, 'path'>;

```

lib/workers/repository/update/branch/get-updated.ts
```diff
@@ -40,6 +40,36 @@ async function getFileContent(
...
+function getManagersForPackageFiles(
+  packageFiles: FileAddition[],
```
viceice, line 53:
```suggestion
  packageFiles: FilePath[],
```

lib/workers/repository/update/branch/get-updated.ts
```diff
@@ -40,6 +40,36 @@ async function getFileContent(
...
+  const packageFileNames = packageFiles.map((packageFile) => packageFile.path);
+  return new Set(
+    Object.keys(managerPackageFiles).filter((manager) =>
+      hasAny(managerPackageFiles[manager], packageFileNames),
+    ),
+  );
+}
+
+function getPackageFilesForManager(
+  packageFiles: FileAddition[],
```
viceice, line 65:
```suggestion
  packageFiles: FilePath[],
```

lib/workers/repository/update/branch/get-updated.ts
viceice:
```suggestion
    const lockFileMaintenancePackageFiles: FileAddition[] =
      lockFileMaintenanceFiles.map((name) => ({
        path: name,
      }));
```

mbudnek: The `type` and `contents` fields are required in `FileAddition`.

lib/workers/repository/update/branch/get-updated.ts
mbudnek: At that point this may as well take a `string[]`.  I only made it take a `FileAddition[]` so that I wouldn't need to do an extra external transform.

lib/workers/repository/update/branch/get-updated.ts
mbudnek: This returns a filtered version of the list of `FileAddition`s, so it needs to take the whole object.

lib/workers/repository/update/branch/get-updated.ts
```diff
@@ -40,6 +40,36 @@ async function getFileContent(
...
+): Set<string> {
+  const packageFileNames = packageFiles.map((packageFile) => packageFile.path);
+  return new Set(
+    Object.keys(managerPackageFiles).filter((manager) =>
+      hasAny(managerPackageFiles[manager], packageFileNames),
+    ),
+  );
+}
+
+function getPackageFilesForManager(
```
viceice, line 64:
```suggestion
function getPackageFilesForManager<T extends FilePath>(
```

lib/workers/repository/update/branch/get-updated.ts
```diff
@@ -40,6 +40,36 @@ async function getFileContent(
...
+  packageFiles: FileAddition[],
+  managerPackageFiles: Set<string>,
+): FileAddition[] {
```
viceice, line 67:
```suggestion
): T[] {
```

lib/workers/repository/update/branch/get-updated.ts
viceice: make function generic, should work as expected 🤗

viceice (approved):

PR #30734, opened 2024-08-13
fix: gitlab project releases number more than 100

Cluas:
## Changes

fetch all releases from project.
## Context
The current implementation only retrieves 100 pieces of data, but in reality, many projects have more than this number of releases, which results in missing many release notes.


rarkins (commented):
Paginate=true is meant to already handle this at the http layer

Cluas:
> Paginate=true is meant to already handle this at the http layer

got it.

Issue #23772, opened 2023-08-09
Move docs generation to main repository


HonkingGoose:
### Describe the proposed change(s).

@viceice suggested that we use a workflow on the `renovatebot/renovate` repository to push the docs build to GitHub Pages.

### Describe why we need/want these change(s).

I tried the new MkDocs `1.5` link checking feature [^1], with their recommended strictness checks. It reports `460` warnings for links in the docs:

- https://github.com/renovatebot/renovatebot.github.io/pull/305

We should fix the problems, and thus make our links more robust. I think fixing the links is easier when the source files are in the main repository?

[^1]: [MkDocs, validation](https://www.mkdocs.org/user-guide/configuration/#validation)

viceice:
publish needs still be on the docs repo, but all build stuff will be moved here.

HonkingGoose:
@viceice Checking in to see if any of our recent changes affect this issue. We're moving two files into the main repository:

- Homepage for the docs: `index.md` file
- The release notes

viceice:
Need to check what needs to be done to close this

PR #30735, opened 2024-08-13
docs: validate with mkdocs

### LINKED ISSUE #28750: Validate anchors on relative links with docs tests

Additionally to #28561 we should also validate anchor correctness.
Otherwise those fail on docs build step.
Those simple errors often cause later failures which are only seen on our docs repo.

viceice:
## Changes
Validate docs with mkdocs for correct linking

## Context
- part of #23772
- closes #28750


pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[tool.poetry]
```
secustor, line 1: Any specific reason to use poetry instead of PDM? 

The later is nearer to the contributor experience which we have already with PNPM.

secustor (commented):
Else LGTM

pyproject.toml
viceice: it' copy'n'past of docs repo. I don't know pdm yet, so feel free to suggest required changes.

viceice: ok, will checkout pdm now

.pdm-python
```diff
@@ -0,0 +1 @@
+/home/kriese/src/renovate/.venv/bin/python
```
HonkingGoose, line 1: Two things:

- The path points to your computer.
- End of file / end of line problem?

.github/workflows/build.yml
```diff
@@ -583,8 +584,20 @@ jobs:
...
           os: ${{ runner.os }}
 
+      - name: Setup PDM
+        uses: pdm-project/setup-pdm@568ddd69406b30de1774ec0044b73ae06e716aa4 # v4.1
+        with:
+          python-version-file: .python-version
+          version: ${{ env.PDM_VERSION }}
+          cache: true
+
+      - name: Install pdm deps
```
HonkingGoose, line 594:
```suggestion
      - name: Install pdm dependencies
```

tools/mkdocs.ts
```diff
@@ -0,0 +1,43 @@
...
+
+void (async () => {
+  await program.parseAsync();
+  const opts = program.opts();
+  logger.info('validating docs');
+  if (opts.build) {
+    logger.info('* generate docs');
+    await generateDocs('tools/mkdocs', false);
+  } else {
+    logger.info('* using prebuild docs');
```
HonkingGoose, line 27: The message here is not clear to me.

Are you caching a previous build maybe? To speed up a second build on the _same_ machine? Or is this about something else?

tools/mkdocs.ts
```diff
@@ -0,0 +1,43 @@
...
+    cwd: 'tools/mkdocs',
+  });
+  if (res.signal) {
+    logger.error(`Signal received: ${res.signal}`);
+    process.exit(-1);
+  } else if (res.status && res.status !== 0) {
+    logger.error(`Error occured:\n${res.stderr || res.stdout}`);
+    process.exit(res.status);
+  } else {
+    logger.debug(`Build succeeded:\n${res.stdout || res.stderr}`);
```
HonkingGoose, line 41:
```suggestion
    logger.debug(`Build completed:\n${res.stdout || res.stderr}`);
```

HonkingGoose (changes requested):

tools/mkdocs.ts
viceice:
```suggestion
    logger.info('* using prebuild docs from build step');
```
It reusing the build from here
https://github.com/renovatebot/renovate/blob/6ca0b361461edabe343452537f63761852b0f75d/.github/workflows/build.yml#L598

otherwise we get rate limited from github because of the issue search

HonkingGoose: Your suggestion is good! Can you apply it for me?

.github/workflows/build.yml
```diff
@@ -595,6 +608,9 @@ jobs:
           name: docs
           path: tmp/docs/
 
+      - name: Build mkdocs
+        run: pnpm build:mkdocs --build=false
```
HonkingGoose, line 612: Is that `--build=false` bit correct?

It almost looks like you're building something, without actually running a (full) build?

viceice: yes, it's skipping another docs generation which would lead to rate limiting on github api. See other conversation.

HonkingGoose (commented):
What do you want to do with this PR?

- #30285

viceice:
> What do you want to do with this PR?
> 
> - #30285

I think it's obsolete then 🤷‍♂️

.github/workflows/build.yml
```diff
@@ -583,8 +584,20 @@ jobs:
           node-version: ${{ needs.setup-build.outputs.node-version }}
           os: ${{ runner.os }}
 
+      - name: Setup PDM
+        uses: pdm-project/setup-pdm@568ddd69406b30de1774ec0044b73ae06e716aa4 # v4.1
+        with:
+          python-version-file: .python-version
+          version: ${{ env.PDM_VERSION }}
```
secustor, line 591: Shouldn't the action pull the version from the dot file?

viceice: which dot file? we've no dot file for pdm version

secustor: Sry it has been late. I mixed up the python and PDM version.

secustor (approved):

PR #30738, opened 2024-08-13
fix(azure): reduce logging

PhilipAbed:
## Changes
see file list

## Context
While implementing Azure on our MendApp

Those warnings shouldn't appear to users there is nothing they can do about it.
image


viceice (commented):
Log at least a debug message

viceice (approved):

PR #30742, opened 2024-08-13
license: Rename license to LICENSE

Thecode764:


HonkingGoose (changes requested):
The [GitHub Docs, Determining the location of your license](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#determining-the-location-of-your-license) says:

> Most people place their license text in a file named `LICENSE.txt` (or `LICENSE.md` or `LICENSE.rst`) in the root of the repository; [here's an example from Hubot](https://github.com/github/hubot/blob/master/LICENSE.md).

I think the choice is between `LICENSE.txt` or `LICENSE.md`.

Both work, if we use `LICENSE.md` we probably need to configure Prettier to ignore the `LICENSE.md` file.

I'll let the maintainers decide which format they like best. :wink:

PR #30756, opened 2024-08-13
feat(managers): add support for uv universal locking

### LINKED ISSUE #27841: New python package manager: uv
# New package manager questionnaire

Did you read our documentation on adding a package manager?

- [x] I've read the [adding a package manager](adding-a-package-manager.md) documentation.

### What's the name of the package manager?
uv
### What language(s) does this package manager support?
Python
### How popular is this package manager?
This is a new manager, but it feels like it will replace `pip` or `poetry`
### Does this language have other (competing?) package managers?
poetry, pip
- [x] Yes (give names).
- [ ] No.

### What are the big selling points for this package manager?
- performance
- doesn't depend to python
- usage native standard `pep-621`
### Explain how this package manager is different from existing ones
- It was written in a different programming language, which makes it independent of the environment of the Python programming language itself
- It is very performance
- Config file `pyproject.toml` usage native standard `pep-621`

### What kind of package files, and names, does this package manager use?
- `pyproject.toml` - File described dependencies, he usage to generate a lock file. Usage standard `pep-621`
- `requirements.txt` - The manager installs the dependencies described in this file (analog lock file). Created after `uv pip compile -o requirements.txt`
### Which [`fileMatch`](../usage/configuration-options.md#filematch) pattern(s) should Renovate use?
`(^|/)pyproject\\.toml$`

### Do many users need to extend the [`fileMatch`](../usage/configuration-options.md#filematch) pattern for custom file names?

- [ ] Yes, provide details.
- [x] No.

### Is the [`fileMatch`](../usage/configuration-options.md#filematch) pattern going to get many "false hits" for files that have nothing to do with package management?
No.
### Can package files have "local" links to each other that need to be resolved?
No
### Package file parsing method

The package files should be:

- [x] Parsed together (in serial).
- [ ] Parsed independently.

### Which format/syntax does the package file use?

- [ ] JSON
- [x] TOML
- [ ] YAML
- [ ] Custom (explain below)

### How should we parse the package files?

- [X] Off the shelf parser.
- [ ] Using regex.
- [ ] Custom-parsed line by line.
- [ ] Other.

### Does the package file have different "types" of dependencies?
Usage standard `pep-621` and cli `uf pip compile --extra {type}`
- [X] Yes, production and development dependencies. 
- [ ] No, all dependencies are treated the same.

### List all the sources/syntaxes of dependencies that can be extracted
Don't know
### Describe which types of dependencies above are supported and which will be implemented in future
Don't know
### What versioning scheme does the package file(s) use?
classic semantic version - x.y.z
### Does this versioning scheme support range constraints, like `^1.0.0` or `1.x`?

- [x] Supports range constraints (for example: `^1.0.0` or `1.x`), provide details.
- [ ] No.

### Is a new datasource required?

- [ ] Yes, provide details.
- [x] No.

### Will users want (or need to) set a custom host or custom registry for Renovate's lookup?

- [ ] Yes, provide details.
- [x] No.

Where can Renovate find the custom host/registry?

- [ ] No custom host or registry is needed.
- [ ] In the package file(s), provide details.
- [ ] In some other file inside the repository, provide details.
- [x] User needs to configure Renovate where to find the information, provide details.

### Are there any constraints in the package files that Renovate should use in the lookup procedure?
Don't know
- [ ] Yes, there are constraints on the parent language (for example: supports only Python `v3.x`), provide details.
- [ ] Yes, there are constraints on the parent platform (for example: only supports Linux, Windows, etc.), provide details.
- [ ] Yes, some other kind of constraint, provide details.
- [ ] No constraints.

### Will users need the ability to configure language or other constraints using Renovate config?

- [ ] Yes, provide details.
- [x] No.

### Does the package manager use a lock file or checksum file?
`requirements.txt` - lock file
- [x] Yes, uses lock file.
- [ ] Yes, uses checksum file.
- [ ] Yes, uses lock file _and_ checksum file.
- [ ] No lock file or checksum.

### Is the locksum or checksum mandatory?
Don't know
- [ ] Yes, locksum is mandatory.
- [ ] Yes, checksum is mandatory.
- [ ] Yes, lock file _and_ checksum are mandatory.
- [ ] No mandatory locksum or checksum.
- [ ] Package manager does not use locksums or checksums.

### If lockfiles or checksums are used: what tool and exact commands should Renovate use to update one (or more) package versions in a dependency file?
Comand not found
#### Does the package manager use a cache?
https://github.com/astral-sh/uv?tab=readme-ov-file#dependency-caching
- [x] Yes, provide details.
- [ ] No.

#### If the package manager uses a cache, how can Renovate control the cache?
`uv cache clean` - delete .cache directory
- [ ] Package manager does not use a cache.
- [x] Controlled via command line interface, provide details.
- [ ] Controlled via environment variables, provide details.

#### Should Renovate keep a cache?

- [x] Yes, ignore/disable the cache.
- [ ] No.

### Generating a lockfile from scratch

Renovate can perform "lock file maintenance" by getting the package manager to generate a lockfile from scratch.
Can the package manager generate a lockfile from scratch?
`uv pip compile -o requirements.txt`
Must be usage override this command, because `compile` would like execute with parameters `--extra tests --extra linters`, this group dependencies (dev/prod).
- [x] Yes, explain which command Renovate should use to generate the lockfile.
- [ ] No, the package manager does _not_ generate a lockfile from scratch.
- [ ] No, the package manager does not use lockfiles.

### What else should we know about this package manager?

_Originally posted by @staners2 in https://github.com/renovatebot/renovate/discussions/27817#discussioncomment-8731017_

mkniewallner:
## Changes

- Closes #27841.

> [!note]
> If this contribution is accepted, we'll need to install uv in https://github.com/containerbase/base, but before doing that, I'd prefer to know if the contribution will be accepted. I also have a question related to how the dependency should be installed, that could determine how the dependency will be installed in the Docker image.

This PR implements partial support for uv by focusing on supporting the new universal lock file (`uv.lock`).

## Context

[uv](https://docs.astral.sh/uv/) supports 2 ways of managing and locking dependencies:
- in a [pip-tools fashion](https://docs.astral.sh/uv/pip/), where dependencies are locked into a `requirements.txt` file
- in [a universal way](https://docs.astral.sh/uv/concepts/projects/), where dependencies are locked into a `uv.lock` file

In both situations, it is possible to define the dependencies using [PEP 621](https://peps.python.org/pep-0621/), which Renovate already supports.

Although the linked issue mostly discussed supporting the `pip-tools` way of locking dependencies (because at the time, universal locking did not exist), this PR is about implementing `uv.lock` support.

For the implementation, I took a lot of inspiration from the already implemented `pdm` manager, which is really close to uv, since both tools support universal locking and are built on top of PEP 621.

#### Support for `[tool.uv.dev-dependencies]`

Documentation: https://docs.astral.sh/uv/concepts/dependencies/#development-dependencies

We basically have a single list of dependencies (unlike PDM, which allows creating multiple groups of development dependencies)

#### Updating lock file on `pyproject.toml` updates

Documentation: https://docs.astral.sh/uv/reference/cli/#uv-lock

uv supports updating specific packages in the lock file by passing `--upgrade-package` option. Weirdly, the option is not listed in the documentation above, but we can find it in the CLI:
```console
$ uv lock --help | grep -- '--upgrade-package'
  -P, --upgrade-package <UPGRADE_PACKAGE>  Allow upgrades for a specific package, ignoring pinned versions in any existing output file. Implies `--refresh-package`
```

#### Lock file maintenance

Documentation: https://docs.astral.sh/uv/reference/cli/#uv-lock

uv supports refreshing the lock file for transitive dependencies by using `uv lock --upgrade`. Unlike PDM, `--upgrade` is necessary here, because by default, uv will not try to refresh any dependency in the lock  file, unless there are updates in `pyproject.toml` that require to update it. So by passing `--upgrade`, we update all dependencies to the latest available while respecting ranges in `pyproject.toml`.

### Things let aside for now

As mentioned earlier, this PR does not implement the 2nd way of locking dependencies: in a `pip-tools` fashion. We could also support this later though, but I believe we'd first have to define how to differentiate uv from the real `pip-tools`, since both use the same output file.

Another thing I did not add is the extraction of locked versions in `extractLockedVersions` method, which is something implemented in PDM. Although it should not be too complex to handle, I intentionally chose to let it aside for now, because the universal lock file is quite new, and could be subject to changes as projects start switching to it. Case in point, the most recent breaking change to it was last week (https://github.com/astral-sh/uv/pull/5861). Once things stabilise, we can of course add support for reading lock files (which AFAIK is used for detecting dependencies with CVEs and raising PRs to update them?).

Finally, there is also a notion of [workspaces](https://docs.astral.sh/uv/concepts/workspaces/) that is, from all I know, similar to Yarn or Cargo. I don't know how much work it would be to support that in Renovate, but that can probably be handled later, especially if this is subject to change.

I've tested the changes over https://github.com/mkniewallner/showcase-uv-renovate.

This shows:
- an update of a dev dependency in `[tool.uv.dev-dependencies]` (https://github.com/mkniewallner/showcase-uv-renovate/pull/4)
- an update of a regular dependency in `[project.dependencies]` (https://github.com/mkniewallner/showcase-uv-renovate/pull/5)
- an update of an optional dependency in `[project.optional-dependencies]` (https://github.com/mkniewallner/showcase-uv-renovate/pull/8)
- a refresh of the lock file during lock file maintenance (https://github.com/mkniewallner/showcase-uv-renovate/pull/12)


lib/util/exec/containerbase.ts
```diff
@@ -186,6 +186,11 @@ const allToolConfig: Record<string, ToolConfig> = {
     packageName: 'rust',
     versioning: semverVersioningId,
   },
+  uv: {
+    datasource: 'github-releases',
+    packageName: 'astral-sh/uv',
+    versioning: semverVersioningId,
```
mkniewallner, line 192: uv is also available [on PyPI](https://pypi.org/project/uv/). Should we retrieve it from there, or is it better to use GitHub? In this file, PDM is also retrieved from GitHub releases, but Poetry is retrieved from PyPI, although it also has GitHub releases.

viceice: use pypi

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    return deps;
+  }
+
+  extractLockedVersions(
+    project: PyProject,
+    deps: PackageDependency[],
+    packageFile: string,
+  ): Promise<PackageDependency[]> {
+    return Promise.resolve(deps);
```
viceice, line 41: why not reading `uv.lock` and parse versions?

mkniewallner: This is explained in the "Things let aside for now" section of the description (second paragraph). If you believe that we should still add it in this PR, I don't mind adding it (it's not that much code, and I have it on another branch), I'm just really afraid of potential breaking changes in the lock file format.

But if a failure here would only raise a warning/log instead of breaking the whole dependency update logic, that might be ok to add it indeed.

mkniewallner:
Just a note that I intentionally only linked the issue instead of using "Closes" in the PR description because the PR only add support for the universal locking strategy, not the `pip-tools` one. I don't mind closing the original issue though but we should probably create another one to track support for `pip-tools` locking style in that case.

lib/modules/manager/pep621/processors/uv.ts
viceice: ok, let's do it in a follow up PR

secustor (dismissed):

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -0,0 +1,146 @@
...
+      case depTypes.buildSystemRequires:
+        // build requirements are not locked in the lock files, no need to update.
+        break;
+      default: {
+        deps.push(dep.packageName!);
+      }
+    }
+  }
+
+  return `${uvUpdateCMD} ${deps.map((dep) => `--upgrade-package ${dep}`).join(' ')}`;
```
viceice, line 145:
```suggestion
  return `${uvUpdateCMD} ${deps.map((dep) => `--upgrade-package ${quote(dep)}`).join(' ')}`;
```

viceice (approved):

zanieb:
Hi! Is there an example somewhere if what's required to configure Dependenbot to work with the various forms of uv outputs? I'd love to add a guide to our official documentation.

Also, as a note in response to the summary comment, we also support universal locking in `requirements.txt` files with the `uv pip compile --universal` flag. Is that supported here?

mkniewallner:
> Also, as a note in response to the summary comment, we also support universal locking in `requirements.txt` files with the `uv pip compile --universal` flag. Is that supported here?

Not yet, this PR only implements universal locking, but there was some discussion to support `uv pip compile` as well in the linked issue. Since it was closed by this PR, we should probably create a new one specifically for `uv pip compile` support.

> Hi! Is there an example somewhere if what's required to configure Dependenbot to work with the various forms of uv outputs? I'd love to add a guide to our official documentation.

You mean Renovate, not Dependabot, right? 😄 Since Renovate will look for an `uv.lock` file to determine that uv is used, it should not require any specific configuration for users already using Renovate. But we might mention in uv's documentation that Renovate is supported (but only support universal locking, and not `uv pip compile`).

zanieb:
> Not yet, this PR only implements universal locking, but there was some discussion to support uv pip compile as well in the linked issue. Since it was closed by this PR, we should probably create a new one specifically for uv pip compile support.

Thanks for clarifying. That sounds good to me.

> You mean Renovate, not Dependabot, right? 😄 Since Renovate will look for an uv.lock file to determine that uv is used, it should not require any specific configuration for users already using Renovate. But we might mention in uv's documentation that Renovate is supported (but only support universal locking, and not uv pip compile).

Sorry! Yes haha, lost in my pile of notifications. That's great to know — I'll give it a try so I can write a quick guide. Thanks again!

mkniewallner:
If you need an example, there's https://github.com/mkniewallner/showcase-uv-renovate, and if you need more details, happy to help, we can discuss about that in Astral's Discord (same username as in GitHub) in order to not spam maintainers here 🙂

PR #30781, opened 2024-08-14
feat(manager/bundler): improve ruby version detection

knu:
## Changes

This PR adds support for reading ruby version from `.tool-versions` in addition to `.ruby-version`, and implements the fallback to reading the currently used ruby version from `Gemfile.lock` when every method fails.

## Context

The bundler manager currently only supports `Gemfile` and `.ruby-version` as the only sources of the ruby version constraints.

Nowadays, asdf and mise have been increasingly adopted as the ruby version manager, and they use `.tool-versions` instead of `.ruby-version`.

`Gemfile` recently added support for specifying the ruby version via file name (https://bundler.io/guides/gemfile_ruby.html) which eliminates literal version constraints from `Gemfile`.

As a result, if a project has `.tool-versions` and uses it in `Gemfile` (`ruby file: ".tool-versions"`) Renovate cannot detect the ruby version correctly and ends up installing the latest stable version of ruby to update gems, which may cause unwanted updates and build failures.

### DIFF
--- a/lib/modules/manager/bundler/common.ts
+++ b/lib/modules/manager/bundler/common.ts
@@ -34,17 +34,24 @@ export async function getRubyConstraint(
       logger.debug('Using ruby version from gemfile');
       return rubyMatch;
     }
-    const rubyVersionFile = getSiblingFileName(
-      packageFileName,
-      '.ruby-version',
-    );
-    const rubyVersionFileContent = await readLocalFile(rubyVersionFile, 'utf8');
-    if (rubyVersionFileContent) {
-      logger.debug('Using ruby version specified in .ruby-version');
-      return rubyVersionFileContent
-        .replace(regEx(/^ruby-/), '')
-        .replace(regEx(/\n/g), '')
-        .trim();
+    for (const file of ['.ruby-version', '.tool-versions']) {
+      const rubyVersion = (
+        await readLocalFile(getSiblingFileName(packageFileName, file), 'utf8')
+      )?.match(/^(?:ruby(?:-|\s+))?(\d[\d.]*)/m)?.[1];
+      if (rubyVersion) {
+        logger.debug(`Using ruby version specified in ${file}`);
+        return rubyVersion;
+      }
+    }
+    const lockFile = await getLockFilePath(packageFileName);
+    if (lockFile) {
+      const rubyVersion = (await readLocalFile(lockFile, 'utf8'))?.match(
+        /^ {3}ruby (\d[\d.]*)(?:[a-z]|\s|$)/m,
+      )?.[1];
+      if (rubyVersion) {
+        logger.debug(`Using ruby version specified in lock file`);
+        return rubyVersion;
+      }
     }
   }
   return null;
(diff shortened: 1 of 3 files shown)


rarkins (dismissed):

knu:
I'll look into the failed tests.

knu:
I've pushed a commit that hopefully fixes failed tests.

lib/modules/manager/bundler/common.ts
```diff
@@ -34,17 +34,24 @@ export async function getRubyConstraint(
...
     }
-    const rubyVersionFile = getSiblingFileName(
-      packageFileName,
-      '.ruby-version',
-    );
-    const rubyVersionFileContent = await readLocalFile(rubyVersionFile, 'utf8');
-    if (rubyVersionFileContent) {
-      logger.debug('Using ruby version specified in .ruby-version');
-      return rubyVersionFileContent
-        .replace(regEx(/^ruby-/), '')
```
knu, line 45: I don't know if there is (still) `.ruby-version` that contains a ruby version prefixed with `ruby-`, but I kept the support for it anyway (as in line 40 below) not to break any existing projects.

knu:
The `build-docs` job needs to be rerun, which failed with the error `You have exceeded a secondary rate limit.`
https://github.com/renovatebot/renovate/actions/runs/10393865444/job/28798657836?pr=30781#logs

rarkins (dismissed):

lib/modules/manager/bundler/common.ts
```diff
@@ -34,17 +34,24 @@ export async function getRubyConstraint(
...
-        .replace(regEx(/\n/g), '')
-        .trim();
+    for (const file of ['.ruby-version', '.tool-versions']) {
+      const rubyVersion = (
+        await readLocalFile(getSiblingFileName(packageFileName, file), 'utf8')
+      )?.match(/^(?:ruby(?:-|\s+))?(\d[\d.]*)/m)?.[1];
```
viceice, line 40:
```suggestion
      )?.match(regEx(/^(?:ruby(?:-|\s+))?(\d[\d.]*)/m))?.[1];
```

lib/modules/manager/bundler/common.ts
```diff
@@ -34,17 +34,24 @@ export async function getRubyConstraint(
...
+      if (rubyVersion) {
+        logger.debug(`Using ruby version specified in ${file}`);
+        return rubyVersion;
+      }
+    }
+    const lockFile = await getLockFilePath(packageFileName);
+    if (lockFile) {
+      const rubyVersion = (await readLocalFile(lockFile, 'utf8'))?.match(
+        /^ {3}ruby (\d[\d.]*)(?:[a-z]|\s|$)/m,
```
viceice, line 49:
```suggestion
        regEx(/^ {3}ruby (\d[\d.]*)(?:[a-z]|\s|$)/m),
```

viceice (approved):

knu:
@viceice Thank you for the review!

PR #30792, opened 2024-08-14
feat(http): Limit Gradle Plugin repository

zharinov:
## Context

- Ref: #27802

### DIFF
--- a/lib/util/http/rate-limits.ts
+++ b/lib/util/http/rate-limits.ts
@@ -27,6 +27,11 @@ const throttleDefaults: ThrottleLimitRule[] = [
     matchHost: 'https://crates.io/api/',
     throttleMs: 1000,
   },
+  {
+    // https://github.com/renovatebot/renovate/discussions/27018#discussioncomment-10336270
+    matchHost: 'https://plugins.gradle.org',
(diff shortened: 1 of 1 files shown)


lib/util/http/rate-limits.ts
```diff
@@ -27,6 +27,11 @@ const throttleDefaults: ThrottleLimitRule[] = [
     matchHost: 'https://crates.io/api/',
     throttleMs: 1000,
   },
+  {
+    // https://github.com/renovatebot/renovate/discussions/27018#discussioncomment-10336270
+    matchHost: 'https://plugins.gradle.org',
+    throttleMs: 2,
```
rarkins, line 33: Please add a comment referencing how this was calculated. I think for the hosted app it will need to be much higher (less requests)

rarkins (changes requested):

rarkins (approved):

rarkins:
@zharinov is this limit configurable (to less reportable second) through config? @nabeelsaabna will need to lower it because all app hosts share the same public IP and therefore rate limit

viceice (approved):

PR #30799, opened 2024-08-15
feat(manager): Cloud Native Buildpacks project descriptor manager

pbusko:
## Changes

- Adds support for Cloud Native Buildpacks project descriptor manager
- Updates docker references to builder and/or buildpacks in the `project.toml` file

## Context

`project.toml` file is a convenient way to keep Cloud Native Buildpacks configuration, when using [`pack`](https://github.com/buildpacks/pack) CLI to produce OCI images. The project descriptor file contains references to OCI images used during build, which could be updated by Renovate

- [Cloud Native Buildpacks](https://buildpacks.io/)
- [Project Descriptor](https://buildpacks.io/docs/reference/config/project-descriptor/) (and it's [spec](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md))


lib/constants/category.ts
```diff
@@ -3,6 +3,7 @@ export const Categories = [
   'ansible',
   'batect',
   'bazel',
+  'buildpacks',
```
rarkins, line 6: Remove this / it's not needed or helpful if there's only one manager in the category

lib/modules/manager/cnb-project-descriptor/extract.ts
```diff
@@ -0,0 +1,79 @@
...
+const dockerPrefix = /^docker:\/\//;
+
+export function extractPackageFile(
+  content: string,
+  packageFile: string,
+  config: ExtractConfig,
+): PackageFileContent | null {
+  const deps: PackageDependency[] = [];
+  try {
+    const descriptor = parseToml(content) as ProjectDescriptor;
```
rarkins, line 21: Use zod validation

lib/modules/manager/cnb-project-descriptor/index.ts
```diff
@@ -0,0 +1,12 @@
...
+export { extractPackageFile } from './extract';
+
+export const defaultConfig = {
+  commitMessageTopic: 'buildpacks project descriptor {{depName}}',
+  fileMatch: ['(^|/)project\\.toml$'],
+  pinDigests: false,
+};
+
+export const categories: Category[] = ['docker', 'buildpacks'];
+export const supportedDatasources = [DockerDatasource.id];
```
rarkins, line 12: Are there other dependency types which need later implementation?

rarkins (changes requested):

rarkins:
Can we find a simpler name for this manager, maybe even "buildpacks"?

lib/modules/manager/cnb-project-descriptor/index.ts
pbusko: None at the moment. The only other potential source for buildpacks would require brand new datasource (buildpacks registry at https://registry.buildpacks.io/)

pbusko:
> Can we find a simpler name for this manager, maybe even "buildpacks"?

sounds good to me

lib/modules/manager/buildpacks/extract.ts
```diff
@@ -0,0 +1,92 @@
...
+  const deps: PackageDependency[] = [];
+  try {
+    const descriptor = parseProjectToml(content);
+    if (!descriptor) {
+      return null;
+    }
+
+    if (
+      descriptor.io.buildpacks?.builder &&
+      !descriptor.io.buildpacks.builder.startsWith('file://')
```
viceice, line 37:
```suggestion
      descriptor.io.buildpacks.builder.startsWith(dockerPrefix)
```

otherwise it will do wron things when they add new prefixes

lib/modules/manager/buildpacks/extract.ts
```diff
@@ -0,0 +1,92 @@
...
+
+      deps.push(dep);
+    }
+
+    if (
+      descriptor.io.buildpacks?.group &&
+      is.array(descriptor.io.buildpacks.group)
+    ) {
+      for (const group of descriptor.io.buildpacks.group) {
+        if (group.uri && !group.uri.startsWith('file://')) {
```
viceice, line 61:
```suggestion
        if (group.uri && group.uri.startsWith(dockerPrefix)) {
```

See above

lib/modules/manager/buildpacks/index.ts
```diff
@@ -0,0 +1,12 @@
+import type { Category } from '../../../constants';
+import { DockerDatasource } from '../../datasource/docker';
+export { extractPackageFile } from './extract';
+
+export const defaultConfig = {
+  commitMessageTopic: 'buildpacks project descriptor {{depName}}',
+  fileMatch: ['(^|/)project\\.toml$'],
```
viceice, line 7: Julia uses `Project.toml`[^1] but the different case shouldn't cause false positives

[^1]: https://pkgdocs.julialang.org/v1/toml-files/

lib/modules/manager/buildpacks/__fixtures__/project.toml
```diff
@@ -0,0 +1,14 @@
...
+schema-version = "0.2"
+
+[io.buildpacks]
+builder = "registry.corp/builder/noble:1.1.1"
+
+[[io.buildpacks.group]]
+uri = "docker://buildpacks/java:2.2.2"
+
+[[io.buildpacks.group]]
+uri = "buildpacks/nodejs:3.3.3"
```
viceice, line 11: this isn't a docker image

pbusko: it is not a real docker image, however it is a perfectly valid docker image reference. shouldn't it be sufficient for unit tests?

lib/modules/manager/buildpacks/index.ts
pbusko: the `fileMatch` pattern won't match against `Project.toml`

lib/modules/manager/buildpacks/extract.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    });
+
+    it('extracts builder and buildpack images', () => {
+      const res = extractPackageFile(
+        Fixtures.get('project.toml'),
+        'project.toml',
+        {},
+      );
+      expect(res?.deps).toMatchSnapshot();
+      expect(res?.deps).toHaveLength(3);
```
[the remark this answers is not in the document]
pbusko: do you mean that the length assertion is pointless after the `toMatchSnapshot` call?

c0d1ngm0nk3y:
> Can we find a simpler name for this manager, maybe even "buildpacks"?

How about `cnb`? `buildpacks` is maybe confusing - there are several versions of buildpacks. The abbreviation `cnb` (cloud native buildpacks) should be known to anyone `project.toml`.

rarkins:
I'm pretty sure the CNF now have a trademark in buildpacks so this will be the only valid usage of that name

Shegox:
> I'm pretty sure the CNF now have a trademark in buildpacks so this will be the only valid usage of that name

They do: https://www.linuxfoundation.org/legal/trademarks 
> - Buildpacks®
> - Buildpacks.io™
> - Buildpacks.IO and Design (color)

c0d1ngm0nk3y:
> I'm pretty sure the CNF now have a trademark in buildpacks so this will be the only valid usage of that name

I don't think that matters since the v2 buildpacks are in CNCF, aren't they?

But probably the name dosn't really matter anyway. The end user will not have any contact unless a `project.toml` is present and for internal use there is still the `readme.md` if in doubt.

lib/modules/manager/cnb/readme.md
```diff
@@ -0,0 +1,21 @@
+The `cnb` manager updates Cloud Native Buildpacks project descriptor file (`project.toml`), which references builder and/or buildpack images by URIs.
+Updates are performed if the `project.toml` file is found and it conforms to the [project descriptor](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md) specification.
```
HonkingGoose, line 2:
```suggestion
Renovate can update a `project.toml` file if:

- It can find the file
- The file follows the [project descriptor](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md) specifications
```

lib/modules/manager/cnb/readme.md
```diff
@@ -0,0 +1,21 @@
...
+
+__Note__: buildpacks in the `io.buildpacks.group` array must be configured with the docker reference (`uri`) for this manager to work.
```
HonkingGoose, line 4:
```suggestion
If you use buildpacks in the `io.buildpacks.group` array, then you _must_ configure the Docker reference (uri`) for Renovate to work.
```

lib/modules/manager/cnb/readme.md
```diff
@@ -0,0 +1,21 @@
...
+
+```toml
```
HonkingGoose, line 6:
```suggestion
```toml title="Example of a `project.toml` file with Docker reference URIs"
```

Let's add a title to the codeblock. [^1]

[^1]: [Material for MkDocs manual, adding a title to a codeblock](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-a-title)

HonkingGoose (changes requested):

pbusko:
@HonkingGoose thank you, will definitely do it in the future. The [contribution guide states](https://github.com/renovatebot/renovate/blob/main/.github/contributing.md#do-not-force-push-to-your-pull-request-branch):

> PRs will always be squashed by us when we merge your work. Commit as many times as you need in your pull request branch.

So I thought that the number of commits/batching would make no difference

HonkingGoose:
> @HonkingGoose thank you, will definitely do it in the future. The [contribution guide states](https://github.com/renovatebot/renovate/blob/main/.github/contributing.md#do-not-force-push-to-your-pull-request-branch):
> 
> > PRs will always be squashed by us when we merge your work. Commit as many times as you need in your pull request branch.
> 
> So I thought that the number of commits/batching would make no difference

Yeah, that's a partial instruction! We really mean:

- Commit as often as you want
- But batch review suggestions (to limit GitHub notification noise)

I'm improving the instructions with:

- #30829

lib/modules/manager/buildpacks/extract.spec.ts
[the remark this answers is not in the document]
viceice: https://github.com/renovatebot/renovate/blob/main/docs/development/best-practices.md#unit-testing

lib/modules/manager/buildpacks/__fixtures__/project.toml
viceice: I think it needs to have the docker prefix, otherwise it's not a docker image reference

pbusko: no, it does not
missing schema is will be verified by the `name.ParseReference` function from the `go-containerregistry` library, and if it succeeds the URI is treated as a docker reference

lib/modules/manager/cnb/readme.md
```diff
@@ -0,0 +1,25 @@
+The `cnb` manager updates Cloud Native Buildpacks project descriptors in `project.toml` files.
+A `project.toml` file can reference builder / buildpack images by URIs.
+Renovate can update a `project.toml` file if:
+
+- It can find the file
+- The file follows the [project descriptor](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md) specifications
```
HonkingGoose, line 6:
```suggestion
- The file follows the [project descriptor specifications](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md)
```

HonkingGoose (approved):
One small thing, after that the docs look good to me.

I'll let a maintainer check for technical accuracy.

lib/modules/manager/cnb/extract.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+    });
+
+    it('extracts builder and buildpack images', () => {
+      const res = extractPackageFile(
+        Fixtures.get('project.toml'),
+        'project.toml',
+        {},
+      );
+      expect(res?.deps).toHaveLength(3);
+      expect(res?.deps).toMatchSnapshot();
```
viceice, line 27: 👆 still not changed, see open conversation

lib/modules/manager/cnb/extract.ts
```diff
@@ -0,0 +1,95 @@
...
+  ExtractConfig,
+  PackageDependency,
+  PackageFileContent,
+} from '../types';
+import { type ProjectDescriptor, ProjectDescriptorToml } from './schema';
+
+const dockerPrefix = /^docker:\/\//;
+
+function isDockerRef(ref: string): boolean {
+  const schemaMatch = ref.match(/^([a-z0-9]+):\/\//);
```
viceice, line 14:
```suggestion
  const schemaMatch = regEx(/^([a-z0-9]+):\/\//).test(ref);
```

lib/modules/manager/cnb/extract.ts
```diff
@@ -0,0 +1,95 @@
...
+import { logger } from '../../../logger';
+import { getDep } from '../dockerfile/extract';
+import type {
+  ExtractConfig,
+  PackageDependency,
+  PackageFileContent,
+} from '../types';
+import { type ProjectDescriptor, ProjectDescriptorToml } from './schema';
+
+const dockerPrefix = /^docker:\/\//;
```
viceice, line 11:
```suggestion
const dockerPrefix = regEx(/^docker:\/\//);
```

lib/modules/manager/cnb/extract.ts
```diff
@@ -0,0 +1,95 @@
...
+
+function isDockerRef(ref: string): boolean {
+  const schemaMatch = ref.match(/^([a-z0-9]+):\/\//);
+  if (schemaMatch && !schemaMatch[0].match(dockerPrefix)) {
```
viceice, line 15:
```suggestion
  if (schemaMatch && !ref.startsWith('docker://') {
```
string search if faster

lib/modules/manager/buildpacks/__fixtures__/project.toml
viceice: please provide a reference

pbusko: sure thing:

1. https://github.com/buildpacks/pack/blob/3a22a7ffef9e69b33ba32fcbfbe6491a09d5397c/pkg/buildpack/locator_type.go#L56
2. https://github.com/buildpacks/pack/blob/3a22a7ffef9e69b33ba32fcbfbe6491a09d5397c/pkg/buildpack/locator_type.go#L103-L109
3. https://github.com/buildpacks/pack/blob/3a22a7ffef9e69b33ba32fcbfbe6491a09d5397c/pkg/buildpack/locator_type.go#L114-L124

and some tests: https://github.com/buildpacks/pack/blob/3a22a7ffef9e69b33ba32fcbfbe6491a09d5397c/pkg/buildpack/locator_type_test.go#L115-L141

lib/modules/manager/cnb/extract.ts
```diff
@@ -0,0 +1,96 @@
...
+  return true;
+}
+
+function parseProjectToml(content: string): ProjectDescriptor | null {
+  const res = ProjectDescriptorToml.safeParse(content);
+  if (res.success) {
+    return res.data;
+  }
+
+  logger.debug(res.error, 'Failed to parse buildpacks project descriptor TOML');
```
viceice, line 29:
```suggestion
  logger.debug({ err: res.error }, 'Failed to parse buildpacks project descriptor TOML');
```
you should pass the package filename, so it's easier to find the failing file

viceice (commented):
otherwise LGTM

rarkins (changes requested):
Please update from main branch then run `pnpm prettier-fix`

lib/modules/manager/buildpacks/__fixtures__/project.toml
pbusko: the assumption was that the docker datasource would verify validity of the reference passed to `getDep` function.

I've added more checks to ignore unsupported sources.

lib/modules/manager/cnb/extract.ts
```diff
@@ -0,0 +1,108 @@
...
+import { getDep } from '../dockerfile/extract';
+import type {
+  ExtractConfig,
+  PackageDependency,
+  PackageFileContent,
+} from '../types';
+import { type ProjectDescriptor, ProjectDescriptorToml } from './schema';
+
+const dockerPrefix = regEx(/^docker:\/?\//);
+const buildpackRegistryRef = regEx(/^[a-z0-9\-\.]+\/[a-z0-9\-\.]+(?:@.+)?$/);
```
HonkingGoose, line 13: ESLint is complaining about this line:

> Unnecessary escape character: `\..`

I don't see a _literal_ match for `\..` on this line though.

pbusko: fixed

lib/modules/manager/cnb/__fixtures__/empty_project.toml
```diff
@@ -0,0 +1,2 @@
+[_]
+schema-version = "0.2"
```
viceice, line 2: please inline this small fixture with `codeBlock` template helper

lib/modules/manager/cnb/extract.ts
```diff
@@ -0,0 +1,108 @@
...
+import { getDep } from '../dockerfile/extract';
+import type {
+  ExtractConfig,
+  PackageDependency,
+  PackageFileContent,
+} from '../types';
+import { type ProjectDescriptor, ProjectDescriptorToml } from './schema';
+
+const dockerPrefix = regEx(/^docker:\/?\//);
+const buildpackRegistryRef = regEx(/^[a-z0-9\-.]+\/[a-z0-9\-.]+(?:@.+)?$/);
```
viceice, line 13: is this a valid docker image for buildpacks:

```toml
[[io.buildpacks.group]]
uri = "node@sha256:914458d8617650599ec2c05f6754403a1ce08cb471b0b1de1de439c539f8d45f"
[[io.buildpacks.group]]
uri = "node:latest@sha256:914458d8617650599ec2c05f6754403a1ce08cb471b0b1de1de439c539f8d45f"
```

it's a valid docker reference

lib/modules/manager/cnb/readme.md
```diff
@@ -0,0 +1,25 @@
+The `cnb` manager updates Cloud Native Buildpacks project descriptors in `project.toml` files.
+A `project.toml` file can reference builder / buildpack images by URIs.
+Renovate can update a `project.toml` file if:
+
+- It can find the file
+- The file follows the [project descriptor specifications](https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md)
```
HonkingGoose, line 6: You commited [ignore unsupported buildpack sources](https://github.com/renovatebot/renovate/pull/30799/commits/2b174a9b6933f2345809836b9650f35bfb15b21a).

Do we need to update the readme, so it explains the unsupported buildpack sources?

lib/modules/manager/cnb/readme.md
```diff
@@ -0,0 +1,25 @@
+The `cnb` manager updates Cloud Native Buildpacks project descriptors in `project.toml` files.
```
HonkingGoose, line 1: @rarkins do you want to keep calling it `cnb`, or do you want to rename to `buildpack`?

Context:

> > I'm pretty sure the CNF now have a trademark in buildpacks so this will be the only valid usage of that name
> 
> I don't think that matters since the v2 buildpacks are in CNCF, aren't they?
> 
> But probably the name doesn't really matter anyway. The end user will not have any contact unless a `project.toml` is present and for internal use there is still the `readme.md` if in doubt.

Users will still see the `cnb` name in their Renovate config. To me it seems easier to have a readable name like `buildpack`, than an abbreviation `cnb`.

rarkins: I would like it called `buildpacks` with an "s" at the end

lib/modules/manager/cnb/readme.md
pbusko: done

lib/modules/manager/cnb/readme.md
pbusko: the manager has been renamed

lib/modules/manager/buildpacks/index.ts
```diff
@@ -0,0 +1,12 @@
+import type { Category } from '../../../constants';
+import { DockerDatasource } from '../../datasource/docker';
+export { extractPackageFile } from './extract';
+
+export const defaultConfig = {
+  commitMessageTopic: 'buildpacks project descriptor {{depName}}',
```
rarkins, line 6: This is a pretty long string compared to other managers. Are you certain that it needs to be so long to be clear? e.g. we normally use "dependency {{depName}}" or "image {{depName}}", "crate {{depName}}", etc.

rarkins (commented):
I recall you mentioning that buildpacks aren't normally on Docker Hub? But in your examples like these:

```
[[io.buildpacks.group]]
uri = "docker://buildpacks/java:2.2.2"

[[io.buildpacks.group]]
uri = "buildpacks/nodejs:3.3.3"
```

These will both default to Docker Hub unless `registryUrls` are specified. Could you clarify if that's ok?

pbusko:
> I recall you mentioning that buildpacks aren't normally on Docker Hub? But in your examples like these:
> 
> ```
> [[io.buildpacks.group]]
> uri = "docker://buildpacks/java:2.2.2"
> 
> [[io.buildpacks.group]]
> uri = "buildpacks/nodejs:3.3.3"
> ```
> 
> These will both default to Docker Hub unless `registryUrls` are specified. Could you clarify if that's ok?

yes, this is the expected behavior

lib/modules/manager/buildpacks/index.ts
pbusko: > Are you certain that it needs to be so long to be clear?

I agree that it doesn't have to be that long, would it be ok to change to `buildpack {{depName}}`?

edmorley:
> that buildpacks aren't normally on Docker Hub?

The upstream CNB project has a buildpack registry, that acts as an index/mapping of buildpack ID and version to remote image URIs (ie: the registry doesn't host the image content itself, only references to it):
https://registry.buildpacks.io/

Most end users will refer to the buildpack ID and versions on this CNB registry rather than using the resolved `docker://` URIs directly.

They can do this via URIs beginning with `urn:cnb:registry:` or by not adding a prefix and relying on the fallback heuristic described here:
https://buildpacks.io/docs/for-app-developers/how-to/build-inputs/specify-buildpacks/#fallback-behavior

Ideally this new `buildpacks` manager would support both Docker URIs and buildpack registry URIs.

For the buildpack registry URIs, the API to use to determine what versions exist is:
`https://registry.buildpacks.io/api/v1/buildpacks/<buildpack_id>`

For example:
https://registry.buildpacks.io/api/v1/buildpacks/heroku/python

However, if supporting the Buildpack Registry was too much hassle, than a v1 that supported only the Docker URIs would still be of use to some users :-)

viceice:
> > I recall you mentioning that buildpacks aren't normally on Docker Hub? But in your examples like these:
> > 
> > ```
> > [[io.buildpacks.group]]
> > uri = "docker://buildpacks/java:2.2.2"
> > 
> > [[io.buildpacks.group]]
> > uri = "buildpacks/nodejs:3.3.3"
> > ```
> > 
> > These will both default to Docker Hub unless `registryUrls` are specified. Could you clarify if that's ok?
> 
> yes, this is the expected behavior
> 
> > I recall you mentioning that buildpacks aren't normally on Docker Hub? But in your examples like these:
> 
> they are not in the DockerHub's library (any image with no repository, like `nginx:latest` is normalised to `library/nginx:latest`), but they are for sure present in the Docker Hub in general, for example in the [`paketobuildpacks`](https://hub.docker.com/u/paketobuildpacks) repository

please use real (existing on docker hub) docker image samples

lib/modules/manager/buildpacks/__fixtures__/project.toml
```diff
@@ -0,0 +1,35 @@
...
```
viceice, line 11: I see now, this is a docker image, but our docker datasource will never find it, because it needs to be translated via build packs registry to the real image.

https://registry.buildpacks.io//api/v1/buildpacks/heroku/nodejs/3.2.11

so I'm pretty sure we need the buildpacks datasource first.

viceice (dismissed):

lib/modules/manager/buildpacks/__fixtures__/project.toml
modulo11: We'll start working on that soon.

pbusko: > I see now, this is a docker image, but our docker datasource will never find it, because it needs to be translated via build packs registry to the real image.
> 
> https://registry.buildpacks.io//api/v1/buildpacks/heroku/nodejs/3.2.11
> 
> so I'm pretty sure we need the buildpacks datasource first.

No, this example is perfectly valid and will work with the current docker datasource. At the moment, there are several different ways to add a buildpack to the `project.toml` file: by `id` or `uri`. When the buildpack URI is specified and it does not fall under the cases listed [here](https://github.com/sap-contributions/renovate/blob/1b85bedf5f576afde6d2b9acc273b9ac5dae04ea/lib/modules/manager/buildpacks/extract.ts#L31-L35), the URI is considered to be a direct OCI image reference, so no buildpack registry lookup should be performed.

The URIs which must be looked up against the buildpacks registry will be ignored and only direct references are picked up by the manager. We wanted to start working on the buildpacks manager incrementally, first by using already existing datasource for updating docker images. We need a separate `buildpacks registry` (or similar) datasource only when the manager must support the following cases:

- `id` + `version` are set instead of the `uri`
- `uri` starts with `urn:cnb:registry`
- `uri` matches the following regex: https://github.com/buildpacks/pack/blob/3a22a7ffef9e69b33ba32fcbfbe6491a09d5397c/pkg/buildpack/locator_type.go#L39

c0d1ngm0nk3y:
> However, if supporting the Buildpack Registry was too much hassle, than a v1 that supported only the Docker URIs would still be of use to some users :-)

Sounds great. I like the approach. Adding the Buildpack Registry later should be perfectly fine as a separate pr, right?

rarkins:
> > However, if supporting the Buildpack Registry was too much hassle, than a v1 that supported only the Docker URIs would still be of use to some users :-)
> 
> Sounds great. I like the approach. Adding the Buildpack Registry later should be perfectly fine as a separate pr, right?

Yes but please:
- Add a note about this limitation to the README.md now
- Create a "Suggest an idea" discussion with that feature description once this PR is merged, so that it can be converted into an issue and not lost

lib/modules/manager/buildpacks/index.ts
rarkins: Yes that would be better

pbusko: done

c0d1ngm0nk3y:
@viceice , @rarkins I think all your comments have been addressed. Could you have another look? Just let us know if something is missing.

c0d1ngm0nk3y:
> > that buildpacks aren't normally on Docker Hub?
> 
> The upstream CNB project has a buildpack registry, that acts as an index/mapping of buildpack ID and version to remote image URIs (ie: the registry doesn't host the image content itself, only references to it): https://registry.buildpacks.io/
> 
> Most end users will refer to the buildpack ID and versions on this CNB registry rather than using the resolved `docker://` URIs directly.
> 
> They can do this via URIs beginning with `urn:cnb:registry:` or by not adding a prefix and relying on the fallback heuristic described here: https://buildpacks.io/docs/for-app-developers/how-to/build-inputs/specify-buildpacks/#fallback-behavior
> 
> Ideally this new `buildpacks` manager would support both Docker URIs and buildpack registry URIs.
> 
> For the buildpack registry URIs, the API to use to determine what versions exist is: `https://registry.buildpacks.io/api/v1/buildpacks/<buildpack_id>`
> 
> For example: https://registry.buildpacks.io/api/v1/buildpacks/heroku/python
> 
> However, if supporting the Buildpack Registry was too much hassle, than a v1 that supported only the Docker URIs would still be of use to some users :-)

We already added another `Datasource` to support that buildpack registry use case as well. But since this is kind of independent of this pr, we would prefer to get this merged first to keep the review small.

docs/usage/self-hosted-configuration.md
rarkins, line 1: Are these changes related to this PR?

lib/modules/manager/buildpacks/extract.ts
```diff
@@ -0,0 +1,122 @@
...
+    ref.startsWith('urn:cnb') || // buildpacks registry or builder urns
+    ref.startsWith('from=') || // builder reference
+    isBuildpackRegistryRef(ref) || // buildpack registry ID reference
+    (schemaMatch && !ref.startsWith('docker:/')) // unsupported schema
+  ) {
+    return false;
+  }
+
+  return true;
+}
```
rarkins, line 41: I think it would be more robust if this function defaulted to false and returned true when conditions are _met_, rather than defaulting to true and returning false for the four unsupported conditions you know about. Is it possible to reverse this logic for more robustness, i.e. validate that it _is_ a docker ref, not that it _isn't_?

c0d1ngm0nk3y: I am not sure about that. Because that would imply that we have a `regex` for a valid docker ref and that `regex` would be rather complex itself. And nothing "bad" will happen in case we identify something wrongly as docker ref. Nothing would be found. Thoughts?

rarkins: I'm certain that it would be more robust. But if you think it's too hard, then we can probably allow it. Please test what happens if a non-Docker ref passes this isDockerRef() function. e.g. will we attempt to look it up on Docker Hub?

viceice: yes, that would happen. I don't like to pass them blind to docker hub when they aren't docker images as is. (eg need a different registry lookup before)

c0d1ngm0nk3y: From my understanding, we have 2 options here:

1. We handle all known non-docker use cases and the rest will be handled as docker. Which might end up trying to pull an image which is not a valid docker ref, *but* it was none of the other use cases and well and nothing but happens if it is not a docker ref, right? It just does not work.
2. We try the `regex` approach to identify a docker ref. If we miss it, it will be ignored although it would work.

I clearly prefer option 1, but it is not my call. I only think that a `regex` for a valid docker image is error-prone and gets ugly very fast imho.

c0d1ngm0nk3y: > I think it would be more robust if this function defaulted to false...

What aspect of "robustness" are you referring to?

HonkingGoose: > > I think it would be more robust if this function defaulted to false...
> 
> What aspect of "robustness" are you referring to?

A more robust program only accepts valid input, and rejects bad input.

**Current design**

The current design defaults to `True`, and only returns `False` on obviously bad cases. This design relies on us to find, and describe, _all_ the known bad cases. Even worse, there may be bad cases we don't know about yet. :upside_down_face: 

**More robust design**

By defaulting to `False`, we can follow this logic:

1. IF the input is obviously bad, return `False` right way (fail-fast)
2. IF the input matches _all_ the expectations, rules, specifications for a Docker reference: return `True`
3. ELSE return `False` (fail-safe)

This design gives us stronger guarantees about the input.

**Deep dive**

We know of four bad cases already:

- `ref.startsWith('urn:cnb') // buildpacks registry or builder urns`
- `ref.startsWith('from=') // builder reference`
- `isBuildpackRegistryRef(ref) // buildpack registry ID reference`
- `(schemaMatch && !ref.startsWith('docker:/')) // unsupported schema`

But we can't be sure we listed _all_ possible bad cases. :upside_down_face: 

Luckily we know what a valid Docker reference looks like. The hard part is to:

- Figuring out which the conditions must be met for input to be a valid Docker reference
- Put the conditions into TypeScript code

**Proposed control flow**

Once we have the specifications, we can change the control flow to something like this:

```
// Fail fast, check for bad stuff first

// Fail on bad schemas

IF input matches:

- (schemaMatch && !ref.startsWith('docker:/')) // unsupported schema

return false

// Fail on other bad cases

IF input matches ANY of these rules:

  - `ref.startsWith('urn:cnb') || // buildpacks registry or builder urns`
  - `ref.startsWith('from=') || // builder reference`
  - `isBuildpackRegistryRef(ref) || // buildpack registry ID reference`

return false

// Check if input is a valid Docker reference

IF the input matches ALL of these rules:

  - Rule 1 to be a valid Docker reference
  - Rule 2 to be a valid Docker reference
  - Rule 3 to be a valid Docker reference
  - Rule 4 to be a valid Docker reference
  - Keep listing rules until we have the full specification for a valid Docker reference...

return true

// Finally, if none of the above matches, fail-safe

ELSE return false
```

pbusko: > yes, that would happen. I don't like to pass them blind to docker hub when they aren't docker images as is. (eg need a different registry lookup before)

https://github.com/renovatebot/renovate/blob/3a6465ee732c17c581811caac7a8e7f8320f4968/lib/modules/manager/helm-values/extract.ts#L58

@viceice We simply took inspiration from already existing code where no checks take place at all before passing it to the `getDep` function from the docker datasource. We'll try to come up with a valid regex for a positive checks.

modulo11: We flipped the detection logic and use a regex which is loosely based on the [reference implementation.](https://github.com/containers/image/blob/main/docker/reference/regexp.go)

loewenstein-sap:
Any update on this? The failing linter is as far as I see it the result of reversing the defaults as it was requested in the review.

rarkins:
> Any update on this? The failing linter is as far as I see it the result of reversing the defaults as it was requested in the review.

The failing CodeQL references the newly added regex pattern. Are you saying that "it's impossible to do a regex for docker packages without catastrophic backtracking"?

However, as we use re2, I'm not sure if it's possible for it to cause a problem in practice. @viceice any idea?

loewenstein-sap:
> Are you saying that "it's impossible to do a regex for docker packages without catastrophic backtracking"?

That's what my colleagues tried to hint at [here](https://github.com/renovatebot/renovate/pull/30799#discussion_r1761097767) and [here](https://github.com/renovatebot/renovate/pull/30799#discussion_r1776608880). See also [this comment](https://github.com/renovatebot/renovate/pull/30799#discussion_r1799460193) linking the [Docker implementation of the regex](https://github.com/containers/image/blob/main/docker/reference/regexp.go).

rarkins:
> That's what my colleagues tried to hint at [here](https://github.com/renovatebot/renovate/pull/30799#discussion_r1761097767) and [here](https://github.com/renovatebot/renovate/pull/30799#discussion_r1776608880).

I don't think they were hinting at catastrophic backtracking, and even if they were they should have spoken clearly, not hinted.

They wrote:

> that regex would be rather complex itself

Sounds to me like "it's hard", not "it's impossible without catastrophic backtracking"

> We try the regex approach to identify a docker ref. If we miss it, it will be ignored although it would work.

Sounds to me like "we might get it wrong"

loewenstein-sap:
> I don't think they were hinting at catastrophic backtracking, and even if they were they should have spoken clearly, not hinted.

They did not hint at "catastrophic backtracking" per se, but they were clearly speaking out for not going down the road of a complex regex.

Should we revert to the pre-regex state or what should be done on this PR to get it merged?

pbusko:
the regex has been optimized, there should be no catastrophic backtracking issues anymore

lib/modules/manager/buildpacks/extract.ts
```diff
@@ -0,0 +1,103 @@
...
+import type {
+  ExtractConfig,
+  PackageDependency,
+  PackageFileContent,
+} from '../types';
+import { type ProjectDescriptor, ProjectDescriptorToml } from './schema';
+
+const dockerPrefix = regEx(/^docker:\/?\//);
+const dockerRef = regEx(
+  /^((?:[\w]+(?:\.[\w.-]+)*)(?::\d+)?\/)?[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*(?::(\w[\w.-]{0,127})(?:@sha256:[A-Fa-f\d]{32,})?|@sha256:[A-Fa-f\d]{32,})$/,
```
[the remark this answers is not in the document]
pbusko: apparently there are still issues :(

rarkins:
Does https://chatgpt.com/share/e/67349544-aa74-800b-9647-9bac228f5ba4 help?

pbusko:
> Does https://chatgpt.com/share/e/67349544-aa74-800b-9647-9bac228f5ba4 help?

unfortunately I have no access to the link, but the problem should be solved (at least GHAS marked all issues as solved)

rarkins:
> > Does [chatgpt.com/share/e/67349544-aa74-800b-9647-9bac228f5ba4](https://chatgpt.com/share/e/67349544-aa74-800b-9647-9bac228f5ba4) help?
> 
> unfortunately I have no access to the link, but the problem should be solved (at least GHAS marked all issues as solved)

Sorry, I thought "share link" would be public. Maybe not

lib/modules/manager/buildpacks/extract.ts
```diff
@@ -0,0 +1,103 @@
...
+} from '../types';
+import { type ProjectDescriptor, ProjectDescriptorToml } from './schema';
+
+const dockerPrefix = regEx(/^docker:\/?\//);
+const dockerRef = regEx(
+  /^((?:[a-z\d](?:[a-z\d-]{0,61}[a-z\d])?(?:\.[a-z\d](?:[a-z\d-]{0,61}[a-z\d])?)*)(?::\d{2,5}\/)?)?[a-z\d]+((\.|_|__|-+)[a-z\d]+)*(\/[a-z\d]+((\.|_|__|-+)[a-z\d]+)*)*(?::(\w[\w.-]{0,127})(?:@sha256:[A-Fa-f\d]{32,})?|@sha256:[A-Fa-f\d]{32,})$/,
+);
+
+function isDockerRef(ref: string): boolean {
+  if (ref.startsWith('docker:/') || dockerRef.test(ref)) {
```
rarkins, line 18:
```suggestion
  if (ref.startsWith(dockerPrefix) || dockerRef.test(ref)) {
```

or is this somehow wrong?

pbusko: I believe `startsWith` accepts only string as input, while `dockerPrefix` is a regex (to capture both cases, when there's a single and double slashes after `docker:`)

rarkins (approved):

PR #30809, opened 2024-08-15
refactor: Use stage 3 decorators

zharinov:
## Changes

Use modern decorator implementation with stricter types


zharinov:
Seems like coverage tooling doesn't handle decorators well, feel free to suggest the solution better than `istanbul ignore`

lib/modules/datasource/aws-machine-image/index.ts
```diff
@@ -11,8 +11,10 @@ import { Datasource } from '../datasource';
 import type { GetReleasesConfig, ReleaseResult } from '../types';
 import type { AwsClientConfig, ParsedConfig } from './types';
 
+const id = 'aws-machine-image';
```
viceice, line 14: why this change? if it's required then please do in separate PR, otherwise this PR is hard to review because of big changes

lib/modules/datasource/aws-machine-image/index.ts
```diff
@@ -80,11 +82,13 @@ export class AwsMachineImageDataSource extends Datasource {
     return [filters, config];
   }
 
-  @cache({
-    namespace: `datasource-${AwsMachineImageDataSource.id}`,
-    key: (serializedAmiFilter: string) =>
-      `getSortedAwsMachineImages:${serializedAmiFilter}`,
-  })
+  @cache(
+    /* istanbul ignore next */ {
```
viceice, line 86: why do we need to ignore this now? 😕

lib/modules/datasource/aws-machine-image/index.ts
zharinov: Stage 3 decorators seem to be instantiated before the class itself, so static constants are not available

viceice: ok, please split that refactor to a separate PR, so we've less changes.

PR #25983, opened 2023-11-27
feat(cargo): support range=update-lockfile

rami3l:
## Changes

Adds `update-lockfile` rangeStrategy support to the `cargo` manager.

## Context

As requested in https://github.com/renovatebot/renovate/issues/22820.

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

---

Hi there, and thanks a lot for making this tool!
I'm from WG-Rustup of the Rust team and we'd like to have this feature to simplify Rustup's regular maintenance process. 
I'm not very experienced with TypeScript, but I do want to drive #23968 thru the finish line.
Thanks in advance for your help!

PR #30811, opened 2024-08-16
feat(manager/gleam): add lockfile support

SnakeDoc:
## Changes

Adds lock file support to the Gleam manager. Remove restrictions on `"update-lockfile"` `rangeStrategy`.

## Context

The Gleam manager currently does not understand the lock file or support the `"update-lockfile"` `rangeStrategy`. It would be convenient if it did.


SnakeDoc:
the eslint error is very odd and doesn't seem to have anything to do with this PR's changes?

lib/modules/manager/gleam/extract.ts
```diff
@@ -53,7 +58,55 @@ function extractGleamTomlDeps(gleamToml: GleamToml): PackageDependency[] {
...
+
+    packageFileContent.lockFiles = [lockFileName];
+
+    for (const dep of packageFileContent.deps) {
+      const packageName = dep.depName!;
+      const versions = coerceArray(versionsByPackage.get(packageName));
+      const lockedVersion = versioning.getSatisfyingVersion(
+        versions,
+        dep.currentValue!,
+      );
```
viceice, line 101: I think this can throw, so it should contain a try catch?

lib/modules/manager/gleam/range.ts
```diff
@@ -16,18 +16,10 @@ export function getRangeStrategy(config: RangeConfig): RangeStrategy {
...
-  }
-  if (rangeStrategy === 'in-range-only') {
-    logger.warn('Unsupported rangeStrategy in-range-only, defaulting to widen');
-    return 'widen';
-  }
   if (rangeStrategy !== 'auto') {
     return rangeStrategy;
   }
+
+  logger.debug('Defaulting to widen strategy');
```
viceice, line 23: I think this can be removed 
```suggestion
```

lib/modules/manager/gleam/update-locked.ts
```diff
@@ -0,0 +1,25 @@
...
+  logger.debug(
+    `gleam.updateLockedDependency: ${depName}@${currentVersion} -> ${newVersion} [${lockFile}]`,
+  );
+  try {
+    const locked = extractLockFileContentVersions(lockFileContent!);
+    if (locked?.get(depName)?.find((version) => version === newVersion)) {
+      return { status: 'already-updated' };
+    }
+    // we don't currently support directly updating the lock file
+    // so we'll use gleam itself (via updateArtifacts()) to update the lock file instead
```
viceice, line 19: do we need this function at all? isn't the artifacts function already handling it? 🤔

lib/modules/manager/gleam/update-locked.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { UpdateLockedConfig, UpdateLockedResult } from '../types';
+import { extractLockFileContentVersions } from './locked-version';
+
+export function updateLockedDependency(
+  config: UpdateLockedConfig,
+): UpdateLockedResult {
+  const { depName, currentVersion, newVersion, lockFile, lockFileContent } =
+    config;
+  logger.debug(
+    `gleam.updateLockedDependency: ${depName}@${currentVersion} -> ${newVersion} [${lockFile}]`,
```
viceice, line 11: are other managers logging here too? if not reduce to trace level

lib/modules/manager/gleam/extract.ts
SnakeDoc: The Gleam manager is largely based off the Rust Cargo manager, since they share very similar behaviors and formats for lockfiles, although they are not quite the same. In this case, the [Cargo manager does set some precedent](https://github.com/renovatebot/renovate/pull/25983/files/414bf95853196c361edacd963868e4397bbc043e#r1407989471) for allowing this to throw, since we don't expect trouble here if we encounter invalid version numbers. Like Cargo, I have added a test to ensure this condition is visible -  71b2bd0dc57b725792129a864032a3c8b16a8ed3

However, I am open to adding the try/catch if you feel it is needed.

lib/modules/manager/gleam/update-locked.ts
SnakeDoc: It bothered me, having to restrict usage of the `update-lockfile` `rangeStrategy`, especially since it is common in Gleam package files to use broad version ranges, such as `>= 1.0.0 and < 2.0.0`. Even though this function itself does not do anything directly, there seemed to be some [precedent for this in the Cargo manager](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/cargo/update-locked.ts#L18).

From what I can tell, using `update-lockfile` strategy allows us to enter into [get-updated.ts#L131](https://github.com/renovatebot/renovate/blob/main/lib/workers/repository/update/branch/get-updated.ts#L131), and remember the `'unsupported'` update at [get-updated.ts#L147](https://github.com/renovatebot/renovate/blob/main/lib/workers/repository/update/branch/get-updated.ts#L147), so that ultimately we trigger `updateArtifacts()` to be run at [get-updated.ts#L339](https://github.com/renovatebot/renovate/blob/main/lib/workers/repository/update/branch/get-updated.ts#L339). If this understand is correct, this would mean we do need `updateLockedDependency()` even if we do not yet support directly updating the lockfile.

Of note, this also lays the ground work to actually implement direct updates of the lockfile, but there are a few things that must be worked out on the Gleam-side first for that to happen. This should allow users to get most or all of the benefits today.

lib/modules/manager/gleam/update-locked.ts
SnakeDoc: There are some precedents for `debug` being used here instead of `trace`, such as here in [Cargo](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/cargo/update-locked.ts#L10) and here in [Poety](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/poetry/update-locked.ts#L11). I am open to reducing this to `trace` however, if you think that is better.

SnakeDoc:
Following the example set by Cargo, I have also added a note about Gleam manager's support for `update-lockfile` in 385046a285c32c9c94edba876c026c33bad4c811

lib/modules/manager/gleam/update-locked.ts
viceice: yes, reduce, users can use log level remap if they need to troubleshoot something

lib/modules/manager/gleam/update-locked.ts
viceice: if `unsupported` it's doing this
https://github.com/renovatebot/renovate/blob/302718f38e54c6c7590f488d749fb4da8362f36f/lib/workers/repository/update/branch/get-updated.ts#L144-L149

it doing the same if no `updateLockedDependency` 

https://github.com/renovatebot/renovate/blob/302718f38e54c6c7590f488d749fb4da8362f36f/lib/workers/repository/update/branch/get-updated.ts#L172-L181


So i still don't see any benefit to add this dummy function

lib/modules/manager/gleam/extract.ts
viceice: But you're using a diffferent versioning (cargo uses cargo and this uses hex). 

Good to know hey doesn't throw, other versionings are throwing errors on invalid versions.
So no need for a try-catch yet.

lib/modules/manager/gleam/artifacts.ts
```diff
@@ -49,7 +49,7 @@ export async function updateArtifacts(
       ],
     };
 
-    await exec('gleam deps download', execOptions);
+    await exec('gleam deps update', execOptions);
```
viceice, line 52: this will update all transitive deps to their latest version?

viceice (commented):
normally we split those PR's to multiple for easier review:
1. extract locked version
2. update the `updateArtifacts`  enable range strategies
3. implement `updateLocked` if needed

lib/modules/manager/gleam/artifacts.ts
SnakeDoc: yes, respecting compatible versions.

lib/modules/manager/gleam/update-locked.ts
SnakeDoc: thank you, i see this now! removing & testing...

lib/modules/manager/gleam/update-locked.ts
SnakeDoc: since we're likely removing this function, i'll "resolve" this without change for now.

SnakeDoc:
> normally we split those PR's to multiple for easier review:
> 
> 1. extract locked version
> 2. update the `updateArtifacts`  enable range strategies
> 3. implement `updateLocked` if needed

I can see how that would much easier. Would it be preferred if I open new PR's as described above (superseding this PR), or are we already in too deep for that?

SnakeDoc:
This PR has been superseded by #31000 and #31001. Closing in favor of those PR's.

PR #30815, opened 2024-08-16
feat(cache): add a mutex utility and prevent parallel processing with the same cache key

secustor:
## Changes
- Adds a mutex implementation based on our cache key system. 
- This PR also adds the lock usage to the cache decorator 

## Context
This prevents race conditions where we have cache miss and then multiple stacks are getting the same packages. 
Where this can have a big impact are where we are limited on API calls or lookups are expensive until cached. 
e.g. Terraform artifact updates, lookups for the `deb` datasource.

It could be interesting for the bigger installations to extend this to Redis later on so this is also the case if multiple instances share the same cache. 

### DIFF
--- a/lib/util/mutex.ts
+++ b/lib/util/mutex.ts
@@ -1,0 +1,21 @@
+import { Mutex, MutexInterface, withTimeout } from 'async-mutex';
+
+const DEFAULT_NAMESPACE = 'default';
+const mutexes: Record<string, Record<string, MutexInterface>> = {};
+
+export function getMutex(
+  key: string,
+  namespace: string = DEFAULT_NAMESPACE,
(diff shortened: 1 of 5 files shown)


lib/util/mutex.ts
```diff
@@ -0,0 +1,21 @@
+import { Mutex, MutexInterface, withTimeout } from 'async-mutex';
```
viceice, line 1:
```suggestion
import { Mutex, type MutexInterface, withTimeout } from 'async-mutex';
```
isn't that a type only import?

viceice (commented):
otherwise LGTM

viceice (approved):

PR #30818, opened 2024-08-16
fix: rename turbo repo

moltar:
## Changes

Renames Vercel Turbo monorepo.

Turbo was added in #22393.

But since then Vercel must have renamed their repo, because that URL redirects to the one in this PR.

## Context

The side effect of having the wrong URL, is that packages are not grouped correctly.


rarkins:
How long ago was the renaming done? Should we have both for a period?

moltar:
> How long ago was the renaming done?

Don't know, tried to find this out, but didn't find any conclusive information.

I just observed on my repos that all of the packages were merged separately recently, but could have been for a while too, and I just didn't once.

> Should we have both for a period?

That I cannot answer, as I am not too familiar with the Renovate ways.

How would that be done? An array of URLs? Or just under a different name?

moltar:
This PR seems to change the URLs: https://github.com/vercel/turborepo/pull/8965

rarkins:
Then we need to have both repo URLs as part of this rule

moltar:
> Then we need to have both repo URLs as part of this rule

How to do that?

moltar:
> Then we need to have both repo URLs as part of this rule

✅ Fixed

rarkins (approved):

PR #30819, opened 2024-08-16
feat(preset): Add group:ruby-on-railsMonorepo

harm-matthias-harms:
## Changes

* Add `group:ruby-on-railsMonorepo`

## Context

Currently, renovate creates up to 4 PRs for Rails-based projects on each Rails update. Both RubyGems and npm packages are published from the https://github.com/rails/rails monorepo.

image

All packages are published with the same `sourceUrl` and could be grouped using a mono repo config. I built a test repository for this and renovate created https://github.com/harm-matthias-harms/renovate-test/pull/3


rarkins (changes requested):
Please convert it into two package rules

harm-matthias-harms:
Just a question for clarification: Could we use a mono repo group here? Or is it important to divide by the kind of registry?

According to the logs, all packages have the same `sourceUrl`:

```
"registryUrl": "https://registry.npmjs.org",
"sourceUrl": "https://github.com/rails/rails",
...
"registryUrl": "https://rubygems.org",
"sourceUrl": "https://github.com/rails/rails",
```

rarkins:
Source url could also work if all packages are published with it. I suggest you check it with a repo package rule to be sure it works as desired

rarkins (dismissed):

lib/config/presets/internal/group.ts
```diff
@@ -458,7 +458,6 @@ const staticGroups = {
       'group:react',
       'group:remark',
       'group:resilience4j',
-      'group:rubyOnRails',
```
HonkingGoose, line 461: Do we want to setup "config migration" code for this change?

It might be nice if Renovate opens a config migration PR to:

1. Drop `group:rubyOnRails`
1. Add monorepo config for `monorepo:ruby-on-rails`

harm-matthias-harms: Are there existing examples? If so I could give it a try.

HonkingGoose: Let's wait for a maintainer to say if my idea is good/bad, before you start changing things. :wink:

viceice: please leave this group in place, it does nothing by default

viceice (approved):

rarkins (approved):

PR #30842, opened 2024-08-16
fix(github): Sanitize mentions in commit bodies

zharinov:
## Changes

Although GitHub doesn't render Markdown in the commit bodies, the mentions are triggered anyway.

Users can configure Renovate in such a way changelogs are rendered in the commit body, which may cause undesired notifications for GitHub users mentioned in the changelogs.

This PR fixes the problem by adding zero-width space after every occurrence of `@` character.

## Context

- Closes: #30828


lib/modules/platform/github/scm.ts
```diff
@@ -1,8 +1,13 @@
+import is from '@sindresorhus/is';
 import * as git from '../../../util/git';
 import type { CommitFilesConfig, LongCommitSha } from '../../../util/git/types';
 import { DefaultGitScm } from '../default-scm';
 import { commitFiles, isGHApp } from './';
 
+export function sanitizeMentions(input: string): string {
+  return input.replaceAll('@', '@\u{8203}');
```
rarkins, line 8: How does this look if you do e.g. "git log"? Is it invisible?

zharinov: Everything looks correct:

image

image

zharinov: On some terminals, the space is rendered, though

viceice (approved):
🤷‍♂️

lib/modules/platform/github/scm.ts
amezin: Maybe search for `/\b@/` (or even `/\b@\S/`)? Asking as a user of "non-web terminal"

zharinov: The problem isn't with search, the problem is with display. I'm pretty sure it can't be rendered equally nice in both the web and terminal (which could be browser-based or the classic GUI ones). After all, zero-width space and backspace characters are just hacks. Having space characters rendered in terminal is the less evil compared to pinging lots of people mentioned in the changelog.

ceisele-r:
@zharinov @rarkins @viceice could it be that this now causes strange Chinese characters to appear in commit messages (displayed in Github and TortoiseGit)?
I am now facing such commit messages
in TortoiseGit
image
and Github
image
using renovate 38.44.2 (via renovatebot/github-action).

I guess it's because the zero-width space is interpreted as "舃" in these locations.

ceisele-r:
Follow-up discussion: #30911

lib/modules/platform/github/scm.ts
amezin: I mean you could likely reduce the impact - because not every `@` character starts a user mention (although my regex suggestions were wrong)

zharinov: Okay, for now we reverted the change and won't iterate on the underlying problem as far as it's a singular event of some user configuring Renovate in a pretty non-conventional way

PR #30844, opened 2024-08-16
chore: eslint to enforce for typed imports

secustor:
## Changes
Adding this rule so that the implicit understanding that imports should be narrowed to type imports are actually tested


lib/config/presets/internal/replacements.ts
```diff
@@ -1,10 +1,7 @@
 import replacementGroups from '../../../data/replacements.json';
 import type { Preset } from '../types';
-import {
-  PresetTemplate,
-  Replacement,
-  addPresets,
-} from './auto-generate-replacements';
+import type { PresetTemplate, Replacement } from './auto-generate-replacements';
+import { addPresets } from './auto-generate-replacements';
```
viceice, line 4: why separate imports? `type` can be inside the braces before the type name 🤔

secustor: I have not a strong opinion on it as long we can auto fix it, though that is the default and should create less lines overall in the code base 🤷‍♂️ 
This is as we would type then to all occurrences of pure type imports e.g.

https://github.com/renovatebot/renovate/pull/30844/files#diff-aa68cda516aaa0c65459a20c907086b0e6b7e8a930a429f70ca5ed46cdcd5676L29-L36

https://typescript-eslint.io/rules/consistent-type-imports/

viceice (approved):

PR #30861, opened 2024-08-18
refactor(instrumentation): Remove unused span parameter

zharinov:
## Changes

When decorator is well-typed, it will complicate implementation.


viceice (approved):

secustor:
This is there so we can add data to the context, such as events. 

Now that this has been merged upstream we should use that, though this will not make a difference for the decorators. 
https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/experimental/trace/SugaredTracer.ts

PR #30864, opened 2024-08-18
refactor(cache): Remove namespace param from decorator

zharinov:


lib/modules/datasource/docker/index.ts
```diff
@@ -221,12 +220,11 @@ export class DockerDatasource extends Datasource {
   }
 
   @cache({
-    namespace: 'datasource-docker-imageconfig',
     key: (
       registryHost: string,
       dockerRepository: string,
       configDigest: string,
-    ) => `${registryHost}:${dockerRepository}@${configDigest}`,
+    ) => `imageconfig|>${registryHost}:${dockerRepository}@${configDigest}`,
```
viceice, line 227: it's probably a bad idea to use same cache key again 🤔

viceice:
added breaking labels, because cache time override is now only possible on datasource and no longer for functions.was heavily use by the app to reduce docker hub rate limiting

lib/modules/datasource/docker/index.ts
zharinov: Yep, it's better to use new keys, but then we'll nuke a lot of data... Maybe it's better to refactor cache keys first

zharinov: Actually, we could leave long-term cache intact, so it's relatively safe to reset

PR #30880, opened 2024-08-19
refactor(nuget): Rename cache namespace

zharinov:


lib/modules/datasource/nuget/v3.ts
```diff
@@ -24,7 +24,7 @@ import type {
 } from './types';
 
 export class NugetV3Api {
-  static readonly cacheNamespace = 'datasource-nuget';
+  static readonly cacheNamespace = 'datasource-nuget-v3';
```
viceice, line 27: will this collide with the V1 API?

zharinov: Only `v3.ts` uses the package cache

viceice (approved):

PR #30881, opened 2024-08-19
refactor(repology): Rename cache namespace

zharinov:


lib/modules/datasource/repology/index.ts
```diff
@@ -109,7 +109,7 @@ export class RepologyDatasource extends Datasource {
 
   @cache({
     ttlMinutes: 60,
-    namespace: `datasource-${RepologyDatasource.id}-list`,
+    namespace: `datasource-${RepologyDatasource.id}`,
```
viceice, line 112: doesn't this collide with the other cache decorator?

zharinov: There is no other decorator (see how I change the types and nothing breaks)

viceice (approved):

PR #30935, opened 2024-08-21
fix(datasource/helm): use registryStrategy=hunt

syphernl:
## Changes

Ensure that all of the configured Helm registries are checked to resolve dependencies.

## Context

Related to: https://github.com/renovatebot/renovate/discussions/30934

### DIFF
--- a/lib/modules/datasource/helm/index.ts
+++ b/lib/modules/datasource/helm/index.ts
@@ -19,6 +19,8 @@ export class HelmDatasource extends Datasource {
 
   override readonly defaultRegistryUrls = ['https://charts.helm.sh/stable'];
 
+  override readonly registryStrategy = 'hunt';
+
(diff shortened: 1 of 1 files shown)


viceice (changes requested):
you should configure your registry urls more selectively via package rules

syphernl:
> you should configure your registry urls more selectively via package rules

@viceice But that would mean that every time we add a project (that needs a Helm chart) we also need to modify the package rules? That feels a bit counter-productive as Renovate is capable of handling this already?

Last year the [very same change](https://github.com/renovatebot/renovate/pull/25520/files) was accepted for the Bazel data source, why would that be any different for the Helm datasource?

viceice:
because helm works differently.

please provide a reproduction, maybe there's a simpler way while helm detects the registry url automatically

PR #30970, opened 2024-08-22
docs(docker): Specify authType in code example

bmaximuml:
## Changes

Documentation update: While basic auth is mentioned above, the code example doesn't include that. My experimentation has shown that this auth doesn't work without specifying the authType as basic

### DIFF
--- a/docs/usage/docker.md
+++ b/docs/usage/docker.md
@@ -312,6 +312,7 @@ If all your dependencies are on the Google Artifact Registry, you can base64 enc
             "matchHost": "europe-docker.pkg.dev",
             "username": "_json_key_base64",
             "password": "<base64 service account>"
+            "authType": "Basic",
           }
         ]
       }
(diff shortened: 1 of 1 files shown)


rarkins (changes requested):
Renovate defaults to authType=Basic whenever username/password is used, so I don't think this is good to add to the example

bmaximuml:
> Renovate defaults to authType=Basic whenever username/password is used, so I don't think this is good to add to the example

Does that include when it's using `encrypted`?
Found this:
` } else if (options.password !== undefined) {`
https://github.com/renovatebot/renovate/blob/745978f11ffbc2585be20f005fdbe854b19b466c/lib/util/http/auth.ts#L76

Not sure how `options.password` is being populated, ie if it'll get populated by encrypted or if that's just pulling straight from what's set in the config.

Fwiw I'm using google artifact registry as below, and kept seeing an auth error until explicitly setting the auth type to Basic.

```json
    {
      "matchHost": "eu.gcr.io",
      "username": "_json_key_base64",
      "authType": "Basic",
      "encrypted": {
        "password": "<password>"
      }
    }
```

rarkins:
The `encrypted` field is decrypted at the very beginning after config is loaded, so by the time that logic you referenced is executed then there should only be username/password in the host rule

PR #31002, opened 2024-08-25
feat(manager/gleam): enable update-lockfile

SnakeDoc:
## Changes

Enables lockfile updates

~~This should _**not**_ be merged until #31000 is accepted.~~
#31000 has been merged, we can proceed with this PR.

## Context

Supersedes #30811 

The Gleam manager does not allow `rangeStrategy=update-lockfile`


lib/modules/manager/gleam/artifacts.ts
```diff
@@ -49,7 +49,7 @@ export async function updateArtifacts(
       ],
     };
 
-    await exec('gleam deps download', execOptions);
+    await exec('gleam deps update', execOptions);
```
viceice, line 52: does this update all dependencies to their latest allowed version?

SnakeDoc: yes

viceice: then this causes unwanted side effects for normal updates. update should only called when the update is a lockfile update.

@rarkins @secustor WDYT?

rarkins: The problem here is if a user has multiple dependencies which are unpinned in the lock file, and Renovate creates a PR which claims to update one of them, but actually updates all of them.

The gleam manager needs a way to specify one or more dependencies to update, e.g. `gleam deps update foo` to update package `foo`.

SnakeDoc: Adding something like `gleam deps update <package name>` support to the Gleam CLI is probably the best permanent solution.

I don't yet know what is involved in adding this feature to Gleam. It would be great if there was some way to enable this functionality without waiting on a new Gleam feature.

In the past, have other managers worked around similar issues? Any alternative approaches I may not have thought about?

rarkins: Either:
1. Reverse engineer the lock file format (which is a bit risky), or
2. Document that it's not supported and recommend pinning dependencies instead

secustor: This would unexpected behaviour by users and would look ( rightfully ) like a bug to them as `glean` would be the only package manager showing this behavior. 

I see these options to move forward: 
- parse manually the lock file ( as `@rarkins` mentioned  which is very risky and a lot of work to maintain )
- request a scoped update command for the Gleam CLI. This is for example how PDM lock files are updated. 
- Implement as is but make it opt in as an experimental flag. That way people are aware of the limitations before actually using.

rarkins: To reuse some Renovate terminology, if we merged this then it would _look like_ it's meant to update just one dependency, but actually each PR would be a "lock file maintenance". I think the confusion/misleadingness would be too much

viceice: why not stay on `deps download ` like now when it's a normal update? so only necessary changes occur to the manifest

SnakeDoc: Not a bad idea for the interim. I will do some testing and see if the behavior is acceptable.

I have also tagged this issue https://github.com/gleam-lang/gleam/issues/2170, and will work on getting this feature into gleam. That will provide the best permanent solution, but will take unknown time/effort.

viceice (commented):
it seems there is a command merged upstream and this PR now needs an update too.

SnakeDoc:
Yes! Great news, [Gleam v1.6.0 landed this feature](https://gleam.run/news/context-aware-compilation/#precise-dependency-updates) as a result of https://github.com/gleam-lang/gleam/pull/3602. Gleam v1.6.1 (released today) includes a hotfix that stabilizes this new feature - so we should be good to proceed!

I have a staged changeset on my development branch that I will test tonight. If everything looks good, I'll update this PR and ping it for a fresh review.

SnakeDoc:
I never know when to trust GH's notifications! Forgive me if this is already in your queue @viceice

lib/modules/manager/gleam/range.ts
```diff
@@ -16,16 +16,6 @@ export function getRangeStrategy(config: RangeConfig): RangeStrategy {
...
-    logger.warn(
-      'Unsupported rangeStrategy update-lockfile, defaulting to widen',
-    );
-    return 'widen';
-  }
-  if (rangeStrategy === 'in-range-only') {
-    logger.warn('Unsupported rangeStrategy in-range-only, defaulting to widen');
-    return 'widen';
-  }
   if (rangeStrategy !== 'auto') {
```
rarkins, line 19: why remove these lines?

SnakeDoc: The purpose of this PR was to remove these restrictions for the Gleam Manager. Perhaps I overlooked something?

The new `gleam update <list of deps>` command allows us to update only the deps we need in the lockfile, respecting the range defined in the package file. I thought that covers both of these `rangeStrategy`'s.

SnakeDoc:
I'm wondering if there's something wrong with this PR, or if everyone is just very busy this time of year? If something is not up to standards, please let me know so I can address it. 🚀

lib/modules/manager/gleam/range.ts
rarkins: Noted

rarkins (approved):

viceice (commented):
ups, forgot to send my review

lib/modules/manager/gleam/artifacts.ts
```diff
@@ -49,7 +49,13 @@ export async function updateArtifacts(
       ],
     };
 
-    await exec('gleam deps download', execOptions);
+    // `gleam deps update` with no packages rebuilds the lock file
+    const packagesToUpdate = isLockFileMaintenance
+      ? []
+      : updatedDeps.map((dep) => dep.depName).filter(Boolean);
```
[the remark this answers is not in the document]
SnakeDoc: I appreciate the review! This is a good suggested change - if I push into the same branch, I think I'll need to open a new PR though since this one has been closed?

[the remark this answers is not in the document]
viceice: yes, you need a new pr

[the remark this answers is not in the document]
SnakeDoc: Changes applied in #32962

SnakeDoc:
Thanks everyone for your help on this PR, and for your patience. It was a long journey, but we made it!

Super eager to get this out into the wild! :rocket:

PR #31008, opened 2024-08-25
fix(helmfile): url joiner is broken, use upath instead

PhilipAbed:
## Changes

change url-join to upath.

## Context

https://github.com/renovatebot/renovate/discussions/30295


lib/modules/manager/helmfile/extract.ts
```diff
@@ -118,12 +118,12 @@ export async function extractPackageFile(
 
       if (isOCIRegistry(dep.chart)) {
         res.datasource = DockerDatasource.id;
-        res.packageName = joinUrlParts(repoName, depName);
+        res.packageName = join(repoName, depName);
```
viceice, line 121: I don't like that workaround, it will not work on windows

PhilipAbed: should i use `upath.posix.join`?

PhilipAbed: or is there something im not familiar with? please elaborate what should we do here :D

PhilipAbed: i will return it to + '/' + then its best

lib/modules/manager/helmfile/extract.ts
```diff
@@ -118,12 +117,12 @@ export async function extractPackageFile(
 
       if (isOCIRegistry(dep.chart)) {
         res.datasource = DockerDatasource.id;
-        res.packageName = joinUrlParts(repoName, depName);
+        res.packageName = repoName + '/' + depName;
```
secustor, line 120: I'm pretty sure we do this all over the place for docker references 🤔 

Have you checked that?  
In that case this should be new utility and should not be part of manager logic

PhilipAbed: I'm just rolling back to how it was before i changed it and caused it to fail,
If you would like to open an issue to refactor later be my guest.
for now let's make our user happy and solve this.

viceice:
```suggestion
        res.packageName = `${repoName}/${depName}`;
```
use string template

lib/modules/manager/helmfile/extract.ts
```diff
@@ -118,12 +117,12 @@ export async function extractPackageFile(
...
       } else if (registryData[repoName]?.oci) {
         res.datasource = DockerDatasource.id;
         const alias = registryData[repoName]?.url;
         if (alias) {
-          res.packageName = joinUrlParts(alias, depName);
+          res.packageName = alias + '/' + depName;
```
viceice, line 125:
```suggestion
          res.packageName = `${alias}/${depName}`;
```

viceice (approved):

PR #31032, opened 2024-08-26
refactor(cdnjs): Use single cache namespace

zharinov:

### DIFF
--- a/lib/modules/datasource/cdnjs/index.ts
+++ b/lib/modules/datasource/cdnjs/index.ts
@@ -34,7 +34,10 @@ export class CdnjsDatasource extends Datasource {
 
   @cache({
     namespace: `datasource-${CdnjsDatasource.id}`,
-    key: ({ packageName }: GetReleasesConfig) => packageName.split('/')[0],
+    key: ({ packageName }: GetReleasesConfig) => {
+      const library = packageName.split('/')[0];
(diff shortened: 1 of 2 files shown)


lib/modules/datasource/cdnjs/index.ts
```diff
@@ -34,7 +34,10 @@ export class CdnjsDatasource extends Datasource {
 
   @cache({
     namespace: `datasource-${CdnjsDatasource.id}`,
-    key: ({ packageName }: GetReleasesConfig) => packageName.split('/')[0],
+    key: ({ packageName }: GetReleasesConfig) => {
+      const library = packageName.split('/')[0];
```
viceice, line 38: do we really need this extra variable?

zharinov: It's not instantly obvious what does it mean, and looks ugly in the string template literal

viceice: we should later extract it to a static function

viceice (approved):

lib/modules/datasource/cdnjs/index.ts
zharinov: Static functions won't be available in the stage 3 decorators

PR #31054, opened 2024-08-27
refactor(terraform-provider): Use single cache namespace

zharinov:


lib/modules/manager/terraform/lockfile/hash.ts
```diff
@@ -107,8 +107,8 @@ export class TerraformProviderHash {
   }
 
   @cache({
-    namespace: `datasource-terraform-provider-hash`,
```
zharinov, line 110: The idea is that in future we'll infer cache namespace from the class name, by transforming it from `TerraformProviderHash` to `terraform-provider-hash`

rarkins (approved):

PR #31060, opened 2024-08-27
feat(cache): Converter from class name to namespace

zharinov:


zharinov:
Actually I think we don't need it

PR #31079, opened 2024-08-28
refactor(cache): Deprecate namespace parameter in decorators

zharinov:
## Changes

This PR deprecates usage of cache namespace for decorators, using the default one (equals the class name).

We could then remove `namespace` paramter per-datasource. Since the default namespace is different, it leads to cache reset.

The alternative is to apply class name transformation, e.g. `FooBarDatasource` to `datasource-foo-bar`, but maybe it's easier to just gradually move to new namespaces.


lib/util/cache/package/decorator.ts
```diff
@@ -42,6 +44,15 @@ interface CacheParameters {
   ttlMinutes?: number;
 }
 
+function getClassName(instance: unknown): string | undefined {
+  // istanbul ignore if: should never happen
+  if (!is.object(instance)) {
+    return undefined;
+  }
+
+  return instance.constructor.name;
```
viceice, line 53: would it make sense to do `kebab-case` here? would look much better 🤔

viceice (commented):
otherwise LGTM

lib/util/cache/package/decorator.ts
zharinov: Is it worth [introducing and maintaining the new code](https://github.com/renovatebot/renovate/pull/31060/files#diff-3744a7eafe7880efbc56d2f472cf9d5588ddeb724e8d233d10d8f78430a49bdcR3-R44), given that mostly we don't work directly with it in the underlying storage level?

viceice: don't we already have lodash in our prod dependency chain which already has such a function?

zharinov: I still think it's better to just use class name unchanged. As far as we don't want to reuse older cache keys, the less name mangling is better.

zharinov: For example, if the underlying string converter slightly changes its behavior (e.g. for abbreviations), the dependency update will cause unanticipated cache reset. And we may spend time investigating load spike, or something like that.

lib/util/cache/package/decorator.ts
```diff
@@ -42,6 +44,15 @@ interface CacheParameters {
   ttlMinutes?: number;
 }
 
+function getClassName(instance: unknown): string | undefined {
+  // istanbul ignore if: should never happen
+  if (!is.object(instance)) {
+    return undefined;
```
viceice, line 50: we should throw if it's happening. undefined is a bad namespace.

or at least log a stack trace at debug level to see how it happens.

PR #31093, opened 2024-08-29
fix(datasource/repology): throw if package/repo not found

kayman-mk:
## Changes

This PR adds a warning to the log file if the Repology datasource is not able to find the package and/or repository. It was a debug message before.

## Context

In case package/repository are not found it needs to visible in the daily work with Renovate as these packages are never updated due to a wrong setup by the user. Usually the logs are set to `info` or even `warn` as you don't wan't to see `debug` messages on a regular basis.

### DIFF
--- a/lib/modules/datasource/repology/index.ts
+++ b/lib/modules/datasource/repology/index.ts
@@ -177,12 +177,7 @@ export class RepologyDatasource extends Datasource {
       throw err;
     }
 
-    logger.warn(
-      { repoName, pkgName },
-      'Repository or package not found on Repology',
-    );
-
-    return undefined;
+    throw new Error(`Repository or package not found on Repology: ${repoName}/${pkgName}`);
   }
 
   async getReleases({
--- a/lib/modules/datasource/repology/index.ts
+++ b/lib/modules/datasource/repology/index.ts
@@ -177,7 +177,7 @@ export class RepologyDatasource extends Datasource {
       throw err;
     }
 
-    logger.debug(
+    logger.warn(
       { repoName, pkgName },
       'Repository or package not found on Repology',
     );


lib/modules/datasource/repology/index.ts
```diff
@@ -177,7 +177,7 @@ export class RepologyDatasource extends Datasource {
       throw err;
     }
 
-    logger.debug(
```
viceice, line 180: @rarkins I think we should throw the error up instead of catching here

rarkins: To be caught by the datasource index function?

kayman-mk: ok, throwing the Error. Sounds good to mw.

viceice: > To be caught by the datasource index function?

yes

lib/modules/datasource/repology/index.ts
```diff
@@ -177,12 +177,7 @@ export class RepologyDatasource extends Datasource {
...
-      { repoName, pkgName },
-      'Repository or package not found on Repology',
-    );
-
-    return undefined;
+    throw new Error(`Repository or package not found on Repology: ${repoName}/${pkgName}`);
```
rarkins, line 180:
```suggestion
    throw err;
```

kayman-mk: But this does not work as there is no `err`.

`TS2304: Cannot find name err`

rarkins: How is this code reached and for what reasons?

rarkins: I looked and there's very few manually thrown errors in datasources. Ideally we shouldn't invent a new type of error here

kayman-mk: Before there was a ``````logger.debug` which I changed to `logger.warn` to make the problem visible in the logs. See above. May be switch back to the first solution?

rarkins:
Sorry, let's have a discussion about this first, because I don't think this is an obvious change or repology-specific.

`null` is a valid result, Repology probably isn't the only datasource which can return it and shouldn't have unique handling of it

https://github.com/renovatebot/renovate/blob/a35197cd4e038b973bc0969b0a2f44ccf35c5de6/lib/modules/datasource/datasource.ts#L37-L39

PR #31108, opened 2024-08-29
feat(cache): Create `Cacheable` wrapper for values

zharinov:
## Changes

The problem: in general, our ability to tell whether the data is public or private is limited to scope of particular function, i.e. by checking the request/response auth headers.

Using `cacheable` callback before function is too soon for such cases. And by the way, the only parameters we MUST know before the cacheable call is the namespace/key.

This PR provides a wrapper class for configuring the caching parameters, which meant to be returned from the cached functions instead of just pure value.


lib/util/cache/package/cacheable.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+  describe('public data', () => {
+    it('via static method', () => {
+      const res: Cacheable<number> = Cacheable.asPublic(42);
+      expect(res.value).toBe(42);
+      expect(res.isPublic).toBeTrue();
+      expect(res.isPrivate).toBeFalse();
+    });
+
+    it('via method', () => {
+      const res: Cacheable<number> = Cacheable.empty().asPublic(42);
```
viceice, line 69:
```suggestion
      const res = Cacheable.empty().asPublic(42);
```
I think this isn't required

lib/util/cache/package/cacheable.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+      expect(res.value).toBe(42);
+      expect(res.isPublic).toBeTrue();
+      expect(res.isPrivate).toBeFalse();
+    });
+  });
+
+  describe('private data', () => {
+    it('via static method', () => {
+      const res: Cacheable<number> = Cacheable.asPrivate(42);
```
viceice, line 78:
```suggestion
      const res = Cacheable.asPrivate(42);
```

lib/util/cache/package/cacheable.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+      expect(res.value).toBe(42);
+      expect(res.isPublic).toBeFalse();
+      expect(res.isPrivate).toBeTrue();
+    });
+
+    it('via method', () => {
+      const res: Cacheable<number> = Cacheable.empty().asPrivate(42);
```
viceice, line 85:
```suggestion
      const res = Cacheable.empty().asPrivate(42);
```

lib/util/cache/package/cacheable.ts
```diff
@@ -0,0 +1,125 @@
+import { DateTime } from 'luxon';
+
+type NotUndefined<T> = T extends undefined ? never : T;
```
viceice, line 3: please add tsdoc annotations to all types and class members.
especially remark which functions are not mutable (return a new object)

lib/util/cache/package/cacheable.spec.ts
zharinov: Yes, but it ensures the right type

lib/util/cache/package/cacheable.ts
zharinov: Actually, I think it's okay modify and return the same object, instead of constructing the new one.

lib/util/cache/package/cacheable.spec.ts
zharinov: Because `Cacheable.empty()` returns `Cacheable<unknown>`, and we want to make sure `asPublic(42)` makes correct type inference/conversion.

lib/util/cache/package/cacheable.ts
viceice: yes, but the docs should mention what function is mutating and what is returning a new object

viceice (approved):

PR #31112, opened 2024-08-29
feat: add preflight method to ManagerAPI

### LINKED ISSUE #29809: Add a function to allow dynamic configuration of managers based on repository data
### Discussed in https://github.com/renovatebot/renovate/discussions/29792

<sup>Originally posted by **secustor** June 21, 2024</sup>
### Tell us more.

From time to time we have the use case to enable managers instead of another one based on e.g. the presence of a file. 

The most recent case is this https://github.com/renovatebot/renovate/discussions/29786#discussioncomment-9836728

The manager interfaces would get a new optional field `preflight` which provides the ability to add functions/methods which are run before the extraction.

```diff
export interface ManagerApi extends ModuleApi {
  defaultConfig: Record<string, unknown>;

  categories?: Category[];
  supportsLockFileMaintenance?: boolean;
  supersedesManagers?: string[];
  supportedDatasources: string[];

+preflight?(): ((RenovateConfig) => RenovateConfig)[]

  extractPackageFile?(
    content: string,
    packageFile?: string,
    config?: ExtractConfig,
  ): Result<PackageFileContent | null>;
...
``` 

That way we can provide dynamic configuration based on the state of the repository and we can pull logic from the `extractAllPackagesFiles`.

Specific to the use case above. We would check if dot files contain OpenTofu references instead of Terraform and then configure the Terraform manager accordingly. 
</div>

Other possibilities are the split up of managers which handle multiple package manager but should not be run in parallel. e.g. npm or pep621

secustor:
## Changes
Adds a new async preflight function to the ManagerAPI. 

Kept it as single method for simplicity and I can not come up with an use case where this would be helpful. 

## Context
This allows to modify Renovates configuration based on content of the repository or other existing configuration. 

Closes https://github.com/renovatebot/renovate/issues/29809


viceice (commented):
we need to think about config caching behavior.  do the hashes match? I think they are checked before extraction.

 this new  function needs a git sync too.

PR #31122, opened 2024-08-30
fix(buildkite): correctly parse plugins that are quoted

jamietanna:
## Changes

It was noticed that when there are quotes around Buildkite plugin usages,
Renovate would include the quote at the start of the `depName` and the
tail of the `currentValue`.

This makes the regex a little less greedy by excluding either single or
double quotes.

## Context

It was noticed that when there are quotes around Buildkite plugin usages,
Renovate would include the quote at the start of the `depName` and the
tail of the `currentValue`.

### DIFF
--- a/lib/modules/manager/buildkite/extract.spec.ts
+++ b/lib/modules/manager/buildkite/extract.spec.ts
@@ -174,5 +174,38 @@ describe('modules/manager/buildkite/extract', () => {
       };
       expect(res).toEqual([bitbucketDependency, githubDependency]);
     });
+
+    it('extracts plugin tags with quotes', () => {
+      const fileContent = codeBlock`
+        steps:
+          - name: "When single quoted"
+            command: true
+            plugins:
+              - 'test-collector#v1.8.0':
+                  files: junit.xml
+                  format: junit
+
+          - name: "Docker Test %n"
+            command: test.sh
+            parallelism: 25
+            plugins:
+              "docker-compose#v1.3.2":
+                run: app
+      `;
+      const res = extractPackageFile(fileContent)?.deps;
+      const singleQuotesDependency: PackageDependency = {
+        currentValue: 'v1.8.0',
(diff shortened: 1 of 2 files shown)


lib/modules/manager/buildkite/__fixtures__/pipeline10.yml
secustor, line 1: Please inline using a `codeBlock`

secustor (commented):
else LGTM

lib/modules/manager/buildkite/__fixtures__/pipeline10.yml
jamietanna: I'd left it to follow the existing patterns in the file - happy to inline, if we're happy it being inconsistent?

rarkins: We believe it to be more efficient (speed) plus more readable, so we want future tests to inline using codeBlock whenever feasible

viceice: also dependabot doesn't find those fixtures and can't report false alarms.

jamietanna: All good - want me to do a prefactor PR to replace at least this file's fixtures with codeBlock?

(I'm also happy to raise one for everything in the codebase if it feels straightforward)

rarkins: Either would be great!

(I wonder if any of the AI's can do this with a codebase)

jamietanna: https://github.com/renovatebot/renovate/pull/31134 (for just the Buildkite side of things so far)

jamietanna:
Oops, sorry I rebased + force pushed :sweat_smile:

rarkins (approved):

PR #31185, opened 2024-09-03
feat: add support for clustered Redis caches

neoeinstein:
## Changes

Adds a new global configuration option with supporting documentation that allows connections to Redis clusters, enabling their use for caching.

## Context

Implements discussion [#31884](https://github.com/renovatebot/renovate/discussions/31184)

### DIFF
--- a/lib/util/cache/package/redis.ts
+++ b/lib/util/cache/package/redis.ts
@@ -15,6 +15,10 @@ function getKey(namespace: PackageCacheNamespace, key: string): string {
   return `${rprefix}${namespace}-${key}`;
 }
 
+export function normalizeRedisUrl(url: string): string {
+  return url.replace(/^(rediss?)\+cluster:\/\//, '$1://');
+}
+
 export async function end(): Promise<void> {
   try {
     // https://github.com/redis/node-redis#disconnecting
@@ -98,7 +102,7 @@ export async function init(
   rprefix = prefix ?? '';
   logger.debug('Redis cache init');
 
-  const rewrittenUrl = url.replace(/^(redis|rediss)\+cluster:\/\//, '$1://');
+  const rewrittenUrl = normalizeRedisUrl(url);
   // If any replacement was made, it means the regex matched and we are in clustered mode
   const clusteredMode = rewrittenUrl.length !== url.length;
 
--- a/lib/util/cache/package/redis.ts
+++ b/lib/util/cache/package/redis.ts
@@ -91,15 +91,19 @@ export async function set(
 export async function init(
   url: string,
   prefix: string | undefined,
-  mode: 'clustered' | 'standalone' = 'standalone',
 ): Promise<void> {
   if (!url) {
     return;
   }
   rprefix = prefix ?? '';
   logger.debug('Redis cache init');
+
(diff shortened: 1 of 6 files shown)


viceice (changes requested):
I would prefer to encode it into the url, like other tools. so we don't need an extra config option.

eg `redis-cluster://....`

neoeinstein:
Note that this would then require (some minimal) parsing and then re-writing the URLs to set the schema back to `redis` or `rediss` as appropriate in order to accomplish this, and would not be in the format generally expected for Redis URLs. That's probably fine if documented, but I want to make sure that's the explicit decision here.

In addition, if we go this way, I'd recommend `redis+cluster` as that seems to be the general convention used when specifying a "protocol" extension in the schema.

While not included here, using the mode parameter instead would make it easier to accept and map multiple root nodes by allowing `redisUrl` to be a comma-separated list of root nodes without ambiguity.

viceice:
> Note that this would then require (some minimal) parsing and then re-writing the URLs to set the schema back to `redis` or `rediss` as appropriate in order to accomplish this, and would not be in the format generally expected for Redis URLs. That's probably fine if documented, but I want to make sure that's the explicit decision here.
> 
> In addition, if we go this way, I'd recommend `redis+cluster` as that seems to be the general convention used when specifying a "protocol" extension in the schema.
> 
> While not included here, using the mode parameter instead would make it easier to accept and map multiple root nodes by allowing `redisUrl` to be a comma-separated list of root nodes without ambiguity.

I would like to find a unified configuration for the package cache, so we can combine it with the SQLite[^1] cache option.
If not defined, use the current default cacache implementation.

[^1]: https://docs.renovatebot.com/self-hosted-experimental/#renovate_x_sqlite_package_cache

rarkins:
Here's how keyv handles Redis clusters: https://keyv.org/docs/storage-adapters/redis/

@viceice are you hoping for a single URL to define a package cache completely? e.g. `sqlite://` `redis://` `rediss://` `rediss+cluster://` etc?

viceice:
> Here's how keyv handles Redis clusters: https://keyv.org/docs/storage-adapters/redis/
> 
> @viceice are you hoping for a single URL to define a package cache completely? e.g. `sqlite://` `redis://` `rediss://` `rediss+cluster://` etc?

yes, something like that. ❤️

rarkins (changes requested):

lib/util/cache/package/redis.ts
```diff
@@ -94,16 +97,28 @@ export async function init(
   }
   rprefix = prefix ?? '';
   logger.debug('Redis cache init');
-  client = createClient({
-    url,
+
+  const rewrittenUrl = url.replace(/^(redis|rediss)\+cluster:\/\//, '$1://');
```
viceice, line 101:
```suggestion
  const rewrittenUrl = url.replace(regEx(/^rediss?\+cluster:\/\//), '$1://');
```

viceice (commented):
this needs some unit tests

lib/util/cache/package/redis.ts
neoeinstein: Updated to add unit tests around the application, used the `regEx` internal type, and simplified the RegEx.

rarkins (approved):

viceice (approved):

PR #31186, opened 2024-09-03
feat(manager/uv): set registry URLs

mkniewallner:
## Changes

Set registry URLs for `uv` manager based on [`index-url`](https://docs.astral.sh/uv/reference/settings/#index-url) and [`extra-index-url`](https://docs.astral.sh/uv/reference/settings/#extra-index-url). As noted by the documentation, `extra-index-url` have higher priority over `index-url` (whether we use the default one, or set it explicitly). Note that uv does not yet support mapping specific dependencies to specific indexes (this is tracked by https://github.com/astral-sh/uv/issues/171).

There is also an [`index-strategy`](https://docs.astral.sh/uv/reference/settings/#index-strategy) parameter to control a bit the behaviour when dealing with multiple indexes (for instance stop at first match for each registry), but since this happens after the dependency resolution, it's not something we can really act on I believe (or at least not at this level in the codebase).

This PR made me realise that the whole uv-specific configuration can also be set using `uv.toml` instead of `pyproject.toml`, but since it was left out in the initial PR implementing uv, I think it would be best to create a separate discussion/issue for that.

## Context

Closes https://github.com/renovatebot/renovate/discussions/31151.

### DIFF
--- a/lib/modules/manager/pep621/processors/uv.ts
+++ b/lib/modules/manager/pep621/processors/uv.ts
@@ -6,13 +6,14 @@ import { exec } from '../../../../util/exec';
 import type { ExecOptions, ToolConstraint } from '../../../../util/exec/types';
 import { getSiblingFileName, readLocalFile } from '../../../../util/fs';
 import { Result } from '../../../../util/result';
+import { PypiDatasource } from '../../../datasource/pypi';
 import type {
   PackageDependency,
   UpdateArtifact,
   UpdateArtifactsResult,
   Upgrade,
 } from '../../types';
-import { type PyProject, UvLockfileSchema } from '../schema';
+import { type PyProject, type Uv, UvLockfileSchema } from '../schema';
 import { depTypes, parseDependencyList } from '../utils';
 import type { PyProjectProcessor } from './types';
 
@@ -31,6 +32,11 @@ export class UvProcessor implements PyProjectProcessor {
         uv['dev-dependencies'],
       ),
     );
+
+    const registryUrls = extractRegistryUrls(uv);
+    for (const dep of deps) {
+      dep.registryUrls = [...registryUrls];
+    }
 
     return deps;
   }
@@ -140,6 +146,16 @@ export class UvProcessor implements PyProjectProcessor {
       ];
     }
   }
+}
+
+function extractRegistryUrls(uvManifest: Uv): string[] {
+  // Extra indexes have priority over default index: https://docs.astral.sh/uv/reference/settings/#extra-index-url
+  const registryUrls = uvManifest['extra-index-url'] ?? [];
+
+  // If default index URL is not overridden, we need to use default PyPI URL additionally to potential extra indexes.
+  registryUrls.push(uvManifest['index-url'] ?? PypiDatasource.defaultURL);
+
+  return registryUrls;
 }
 
 function generateCMD(updatedDeps: Upgrade[]): string {
--- a/lib/modules/manager/pep621/processors/uv.ts
+++ b/lib/modules/manager/pep621/processors/uv.ts
@@ -34,8 +34,10 @@ export class UvProcessor implements PyProjectProcessor {
     );
 
     const registryUrls = extractRegistryUrls(uv);
-    for (const dep of deps) {
-      dep.registryUrls = [...registryUrls];
+    if (registryUrls.length) {
+      for (const dep of deps) {
+        dep.registryUrls = [...registryUrls];
+      }
     }
 
     return deps;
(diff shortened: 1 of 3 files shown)


lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -32,6 +33,11 @@ export class UvProcessor implements PyProjectProcessor {
       ),
     );
 
+    const registryUrls = extractRegistryUrls(uv);
+    for (const dep of deps) {
+      dep.registryUrls = [...registryUrls];
```
rarkins, line 38: Won't this now always set registry URLs? Isn't it better to leave it blank if none are explicitly set?

mkniewallner: Yep we can, I was hesitant to do it because PDM also always set the registry URLs AFAIK, even if the datasource itself also has the default, but I also prefer to only override if necessary in the manager itself, so updated that, thanks.

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -142,6 +150,20 @@ export class UvProcessor implements PyProjectProcessor {
   }
 }
 
+function extractRegistryUrls(uvManifest: Uv): string[] {
```
rarkins, line 153: Shouldn't this function return an empty array if neither index or extra index are defined?

lib/modules/manager/pep621/processors/uv.ts
rarkins: I'll need to check here with @secustor on whether that's intentional or not. Generally speaking we prefer that managers don't set registryUrls unless they are specified, otherwise we should let the datasource defaults take precedence

lib/modules/manager/pep621/processors/uv.ts
mkniewallner: With the latest change, it should be the case no? If neither `extra-index-url` nor `index-url` is set, we return `[]`. Or would you prefer to return `undefined` in that case?

rarkins: Yes I think you're right, the logic flow is a little complicated for this "default" case.

lib/modules/manager/pep621/processors/uv.ts
secustor: I have set this in PDM on purpose as in case of the manager all registryUrls set in their tool context also applies to "standard" dependencies of `pyproject.toml` 

`registryUrls` should be only set in case of PDM a `tool.pdm.source` table exists
```
[[tool.pdm.source]]
name = "private"
url = "https://private.pypi.org/simple"
verify_ssl = true
```

lib/modules/manager/pep621/processors/uv.ts
mkniewallner: Sorry for the ping, but is there anything you'd like me to refactor based on your last comment? I wasn't sure of that, since the comment is resolved.

rarkins: No it's fine as-is

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -32,6 +33,13 @@ export class UvProcessor implements PyProjectProcessor {
       ),
     );
 
+    const registryUrls = extractRegistryUrls(uv);
+    if (registryUrls.length) {
+      for (const dep of deps) {
+        dep.registryUrls = [...registryUrls];
+      }
```
rarkins, line 40: Normally if registry URLs apply to the entire package file then we want them set once at the packageFile level instead duplicated into every dep.

Also, are these likely to _all_ be pypi datasource now and forever? Or is it possible that there could be a git dependency mixed in there at one point (which would then make the registry URL invalid for that dep)

mkniewallner: > Normally if registry URLs apply to the entire package file then we want them set once at the packageFile level instead duplicated into every dep.

There is a plan to be able to map specific dependencies to specific sources in https://github.com/astral-sh/uv/issues/171, similarly to what PDM does I believe. According to [this comment](https://github.com/astral-sh/uv/pull/6389#issuecomment-2327284874) it should be worked on soon enough. It doesn't mean that we can't implement what you're suggesting, which makes sense, but maybe since this is meant to evolve, setting registries at the dependency level is more conservative?

> Also, are these likely to _all_ be pypi datasource now and forever? Or is it possible that there could be a git dependency mixed in there at one point (which would then make the registry URL invalid for that dep)

That's a really good point, even today it's already possible to set other sources, like [git](https://docs.astral.sh/uv/concepts/dependencies/#git), [URL](https://docs.astral.sh/uv/concepts/dependencies/#url), [path](https://docs.astral.sh/uv/concepts/dependencies/#path), or [workspace](https://docs.astral.sh/uv/concepts/dependencies/#workspace-member) dependencies. We should definitely skip those for now similarly to [what we do in `cargo`](https://docs.astral.sh/uv/concepts/dependencies/#workspace-member), by checking if a source is defined for a dependency (and later on we could support updates for git dependencies for instance).

I can make a separate PR to handle skipping those dependencies, since I believe this is not directly related to what this PR implements, as today we already consider dependencies with different sources as public PyPI ones anyway.

mkniewallner: Created https://github.com/renovatebot/renovate/pull/31270 for the 2nd point, thanks for pointing that out, I should have been implemented from the start.

PR #31222, opened 2024-09-04
docs(datasource): All items in the urls array are wrapped between []() character.

### LINKED ISSUE #30312: Replace bare links in `index.ts` files in the `modules/versioning/*/` pattern with good links

Replace bare links in `index.ts` files in the `modules/versioning/*/` pattern with good links.

Bare links are hard to read, and especially bad for screen reader users. Add a descriptive name to each bare link.

### Todos

- [ ] Write failing test first
- [ ] Update all `urls` constants
- [ ] Open PR

#### Write failing test first

Scope of test: all `urls` constants in the pattern `modules/versioning/*/index.ts`.

| Condition | Test result |
| :--- | :--- |
| Any item in the `urls` array is _not_ wrapped between `[]()` characters. | Fail |
| All items in the `urls` array are wrapped between `[]()` characters. | Pass |

#### Update all `urls` constants

Example of a good link, that passes the test:

```ts
 export const urls = [ 
   '[Microsoft API versioning guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#api-versioning)', 
 ]; 
```

#### Open PR

You are ready to open a pull request when the test passes.

### Related discussion/issue

I asked about fixing the bare links in this discussion:

- #29483

Fixing the bare links is part of the work for:

- #29148

Prtik12:
Fixes #30312

## Changes

Updated the `formatUrls` function to correctly format and escape URLs in the documentation. Fixed bad links  and ensured that links in `const urls` are properly validated. 

## Context

This PR addresses issues with broken links in the documentation and improves URL formatting to prevent build errors and improve the readability of generated documentation.

Replaces #31161


RahulGautamSingh:
Add a test in the `modules/versioning/index.spec.ts`. In the test go through each versioning, take the `urls` field and check if the urls are wrapped in `[]()`.

RahulGautamSingh:
Hey,

Thanks for going the effort to add names to all the urls. 🙌

You will also need to add a test to make sure that all the urls in the versioning & manager module have `[]()`.
For this, we will need a test in the `modules/versioning/index.spec.ts` file (same for managers modules). 

Since, this is your first time contributing it could be tiring to figure out how the tests work so here;s an example for the versioning module. Based on it, you can try to write on for the maangers module too.

Add this to the `modules/versioning/index.spec.ts` file. After [this ](https://github.com/renovatebot/renovate/blob/ff875962708dcaf150ede3fdff724d7a025ab7e5/lib/modules/versioning/index.spec.ts#L59)line:


```ts
const urlRegex = regEx(/^\[.*\]\(.*\)$/);
    for (const [verName, verModule] of Object.entries(loadedVers)) {
      const urls = (verModule as { urls?: string[] }).urls ?? [];
      for (const url of urls) {
        const urlValid = urlRegex.test(url);
        expect({
          versioning: verName,
          urlValid,
        }).toMatchObject({
          versioning: verName,
          urlValid: true,
        });
      }
    }
```

It goes through each versioning module, takes the `urls` field and uses regex matching to check if url is wrapped in `[]()` chars.

Prtik12:
Screenshot 2024-09-07 at 6 06 54 PM

Hi,
I recently ran the test suite and noticed that one of the tests was skipped. I wanted to ask if the skipped test could have any potential impact on the project or if there are any known issues related to it. Should i be concerned about this, or is there a reason why it's skipped that i should be aware of?

Thanks for your assistance!

lib/modules/platform/bitbucket-server/index.spec.ts
```diff
@@ -249,7 +249,7 @@ describe('modules/platform/bitbucket-server/index', () => {
           username,
           password,
         });
-      });
```
RahulGautamSingh, line 252: this change? why?

lib/workers/repository/init/index.spec.ts
```diff
@@ -66,6 +66,6 @@ describe('workers/repository/init/index', () => {
       expect(logger.logger.warn).toHaveBeenCalledWith(
         "Configuration option 'expandCodeOwnersGroups' is not supported on the current platform 'undefined'.",
       );
-    });
```
RahulGautamSingh, line 69: same as last comment.

lib/modules/platform/bitbucket-server/index.spec.ts
Prtik12: Oh, sorry, I forgot to mention this before. I’ve increased the timeout for some tests because the setup for certain tests was exceeding the default 5-second timeout. I’ve set the timeout to 10 seconds to accommodate these longer-running operations.
Screenshot 2024-09-07 at 9 40 30 PM
This was the error I encountered.
 Please let me know if this change aligns with testing practices or if there are any concerns about this adjustment.

RahulGautamSingh: Is that because of tests you added? Else you can revert and add them incase CI tests fail for the same reason.

Prtik12: Hi @RahulGautamSingh ,

Yes, the increased timeout was primarily to address issues with some of the tests I added. If these tests are causing problems, I can revert the timeout change and reintroduce it only if necessary, based on CI test results.

Would you prefer that I revert the timeout change now and monitor if any CI tests fail due to timeouts? If so, I can address any issues that arise with more specific adjustments.

Please let me know how you’d like to proceed.

Thanks!

RahulGautamSingh: Yes, remove than and check the CI results. 
The changes you made should not cause any test issues.

Prtik12: Got it, I'll remove the timeout increase and push the changes. I'll then monitor the CI results to ensure everything runs smoothly.

Thanks for the guidance!

data/kubernetes-api.json5
```diff
@@ -147,8 +147,8 @@
   ImageRepository: ['image.toolkit.fluxcd.io/v1beta2'],
   ImageUpdateAutomation: [
     'image.toolkit.fluxcd.io/v1beta1',
-    'image.toolkit.fluxcd.io/v1beta2'
-    ],
+    'image.toolkit.fluxcd.io/v1beta2',
+  ],
```
viceice, line 151: revert and do in separate PR

lib/modules/manager/index.spec.ts
```diff
@@ -89,6 +90,22 @@ describe('modules/manager/index', () => {
       const mgr = mgrs.get(name)!;
       expect(validate(mgr, name)).toBeTrue();
     }
+
+    const urlRegex = regEx(/^\[.*\]\(.*\)$/);
+
+    for (const [managerName, managerModule] of Object.entries(loadedMgr)) {
+      const urls = (managerModule as { urls?: string[] }).urls ?? [];
```
viceice, line 97:
```suggestion
      const urls = (managerModule as { urls?: string[] })?.urls ?? [];
```
why the cast? 🤔 we've an interface for them, see docs generation inside tools folder

lib/modules/versioning/index.spec.ts
```diff
@@ -57,6 +58,21 @@ describe('modules/versioning/index', () => {
       const ver = vers.get(name)!;
       expect(validate(ver, name)).toBeTrue();
     }
+
+    const urlRegex = regEx(/^\[.*\]\(.*\)$/);
+    for (const [verName, verModule] of Object.entries(loadedVers)) {
+      const urls = (verModule as { urls?: string[] }).urls ?? [];
```
viceice, line 64: see other comments

lib/modules/manager/index.spec.ts
RahulGautamSingh: Its not needed for the manager module because `ManagerApi` extends `ModuleApi` which has the preoperty `urls`.

But, the `VersioningApi` is stand-alone and does not extend the `ModuleApi`, so it doesn't allow the `urls` property. 
Needs to be fixed in diff PR, first.

RahulGautamSingh, line 97:
```suggestion
      const urls = managerModule?.urls ?? [];
```

Issue #30982, opened 2024-08-23
Lazy Maven version validation


rarkins:
### Describe the proposed change(s).

Today we validate each maven datasource version/result just in case: https://github.com/renovatebot/renovate/blob/745978f11ffbc2585be20f005fdbe854b19b466c/lib/modules/datasource/maven/index.ts#L216-L225

The downside of this is that we need to do N checks for N versions of a package, which can introduce a lot of load for registries both public and private.

We should refactor our logic instead like this:
- No longer proactively/eagerly checking _every_ release
- Add a new optional method in datasources called validateRelease() for Maven. It should cache very aggressively
- Instead, once we select a release for upgrading to, we then call validateRelease() to make sure it's valid, and fall back to lower releases in the "bucket" if it fails the check, similar to how we drop versions for reasons like minimum release age

This would be a better alternative to the breaking change proposed in #30415

zharinov:
I almost came up with the working example, and suddenly things got weird: in Maven case, our checks are HEAD requests of the corresponding `.pom` files, but we also use `Last-Modified` header as our `releaseTimestamp`. If we stop those checks, we're losing `releaseTimestamp` fields, and therefore `minimumReleaseAge` checks are skipped.

If we still want to go this way, the extended datasource API will look like this:
```ts
interceptRelease(config: Release): Promise<Release | null>;
```

This could reject the release by returning `null`, or it could enrich its data by attaching `releaseTimestamp` or whatnot. Once non-null value is returned, further confidence level and `minimumReleaseAge` checks are performed.

As far as I can remember, we have couple more datasources lacking `releaseTimestamp`, because it takes N+1 HTTP requests to obtain them. Hopefully, this approach could solve it.

rarkins:
Yes, please go ahead with that approach

PR #31232, opened 2024-09-05
feat(datasource): Add `postprocessRelease` optional method

zharinov:
## Changes

Sometimes it's very convenient to fetch only basic release information such as list of versions.

Once the decision for update is made, we could fetch additional information such as `registryTimestamp`.

Or, if the decision was wrong, we could reject it and select another release having lower version (for example, stability days).

## Context

- This PR is the first step to implement #30982


lib/modules/datasource/types.ts
```diff
@@ -3,6 +3,7 @@ import type {
   CustomDatasourceConfig,
 } from '../../config/types';
 import type { ModuleApi } from '../../types';
+import type { CandidateReleaseConfig } from '../../workers/repository/process/lookup/types';
```
viceice, line 6: I don't like the idea to reference types from worker here 🤔 it's a kind of circular type reference

zharinov: Yes, actually I'll move most of the code to datasource

lib/modules/datasource/postprocess-release.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (!datasource) {
+    return release;
+  }
+
+  const ds = getDatasourceFor(datasource);
+  if (!ds) {
+    return release;
+  }
+
+  if (!ds.interceptRelease) {
```
rarkins, line 20: Why the different name?

lib/modules/datasource/postprocess-release.ts
```diff
@@ -0,0 +1,38 @@
...
+    return release;
+  }
+
+  try {
+    const result = await ds.interceptRelease(
+      {
+        packageName: config.packageName,
+        registryUrl: config.registryUrl,
+        currentValue: config.currentValue,
```
rarkins, line 29: Are we sure currentValue is necessary/relevant?

rarkins (changes requested):
I think it would be good to add the maven implementation as part of this so that we have validated that the API is sufficient and the logic working

zharinov:
> I think it would be good to add the maven implementation as part of this so that we have validated that the API is sufficient and the logic working

There are lots of changed files in Maven due to tests, so probably you won't like it

lib/modules/datasource/postprocess-release.ts
zharinov: Yep, it only causes confusion at this pount, removing it

lib/modules/datasource/postprocess-release.ts
zharinov: Bad "rename all", fixed it

lib/modules/datasource/postprocess-release.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { PostprocessReleaseConfig, Release } from './types';
+
+export async function postprocessRelease(
+  config: PostprocessReleaseConfig,
+  release: Release,
+): Promise<Release | null> {
+  const { datasource } = config;
+
+  if (!datasource) {
+    return release;
```
viceice, line 12: we should warn, should never happen

lib/modules/datasource/postprocess-release.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  const ds = getDatasourceFor(datasource);
+  if (!ds) {
+    return release;
```
viceice, line 17: warn too, should never happen

lib/modules/datasource/types.ts
```diff
@@ -90,6 +90,12 @@ export interface ReleaseResult {
   packageScope?: string;
 }
 
+export interface PostprocessReleaseConfig {
+  datasource?: string;
```
viceice, line 94: we should have two interfaces, a public with datasource and a private without. otherwise someone could expect the datasource property is passed through.

lib/modules/datasource/types.ts
```diff
@@ -90,6 +90,12 @@ export interface ReleaseResult {
...
+  packageName?: string;
+  registryUrl?: string;
```
viceice, line 96: shouldn't they be mandatory here? at least packageName?

lib/workers/repository/process/lookup/filter-checks.ts
```diff
@@ -53,6 +54,16 @@ export async function filterInternalChecks(
       );
       // Apply packageRules in case any apply to updateType
       releaseConfig = applyPackageRules(releaseConfig, 'update-type');
+
+      const updatedCandidateRelease = await postprocessRelease(
+        releaseConfig,
+        candidateRelease,
+      );
+      if (!updatedCandidateRelease) {
+        continue;
```
rarkins, line 63: Add a debug log message here

lib/modules/datasource/postprocess-release.ts
rarkins: Actually this is the most common path - most datasources don't have this function

lib/modules/datasource/postprocess-release.ts
```diff
@@ -0,0 +1,59 @@
...
+      { packageName, registryUrl },
+      release,
+    );
+
+    if (!result) {
+      logger.debug(
+        { packageName, registryUrl, version: release.version },
+        'Rejected release',
+      );
+    }
```
rarkins, line 52: Or should this debug be moved down to the datasource layer where a more meaningful reason could be logged?

lib/modules/datasource/postprocess-release.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    return result;
+  } catch (err) {
+    logger.debug({ err }, `Release interceptor failed for "${datasource}"`);
```
rarkins, line 56: Shouldn't this be a warn?

lib/workers/repository/process/lookup/filter-checks.ts
```diff
@@ -53,6 +54,16 @@ export async function filterInternalChecks(
...
+      }
```
rarkins, line 64: Can we add these to a new/optional list like `rejectedReleleases` so there's visibility in `packageFiles with updates`?

rarkins (changes requested):

lib/modules/datasource/postprocess-release.ts
```diff
@@ -0,0 +1,59 @@
...
+  const { packageName } = config;
+  if (!packageName) {
+    logger.warn(
+      { datasource },
+      'Release postprocessing is not supported for empty `packageName` field',
+    );
+    return release;
+  }
+
+  const registryUrl = config.registryUrl ?? null;
```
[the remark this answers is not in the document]
rarkins: This I disagree with. If there's just a single registry it's better to populate registryUrl on the result level and not duplicate it into every release. The only time we populated registryUrl per-release was when there was a mix of registries

rarkins (dismissed):

zharinov:
Let's :shipit:

rarkins (approved):

PR #31254, opened 2024-09-06
feat(platform/bitbucket-server): support codeowners with email

mikevader:
## Changes

With the flag `reviewersFromCodeOwners` email addresses are supported in the CODEOWNERS file and not just usernames.

## Context

The `.bitbucket/CODEOWNERS` file can contain both usernames and email addresses. They could even contain groups/teams but thats a topic for another time.

Bitbucket itself supports all and will correctly add a default reviewer listed with the email address in the CODEOWNERS file when the PR is created within Bitbucket. Renovate on the other hand only supports users with the correct username. This is because the Bitbucket API only allows reviewers to be added with the username.

This is not ideal because a) it break expectations as emails work with Bitbucket itself and b) because often usernames are hard to read and understand (in our case its just an arbitrary 8-digit number)

The proposed change will check for emails in the reviewer list and tries to resolve them to usernames over a bitbucket API.


mikevader:
I just realised that I created the Commits with my corp git config :(.

It was stated somewhere to not recreate or rebase but I'm not sure I can fix that without rewriting the history.

rarkins:
You can force push to fix it in a case like this! Thanks

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -670,7 +670,13 @@ async function updatePRAndAddReviewers(
     }
 
     // TODO: can `reviewers` be undefined? (#22198)
-    const reviewersSet = new Set([...pr.reviewers!, ...reviewers]);
+    const reviewersSet = new Set(
+      await Promise.all(
```
viceice, line 674:
```suggestion
      await Promise.allSettled(
```
otherwise we've lingering requests when one request fails early

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -731,6 +737,17 @@ async function retry<T extends (...arg0: any[]) => Promise<any>>(
   // eslint-disable-next-line @typescript-eslint/only-throw-error
   throw lastError;
 }
+// https://docs.atlassian.com/bitbucket-server/rest/5.1.0/bitbucket-rest.html#idm45588158982432
+async function resolveEmailToUsername(email: string): Promise<string | null> {
```
viceice, line 741: move to utils

viceice:
- drop snapshot changes
- add context to PR description

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -670,7 +670,13 @@ async function updatePRAndAddReviewers(
     }
 
     // TODO: can `reviewers` be undefined? (#22198)
-    const reviewersSet = new Set([...pr.reviewers!, ...reviewers]);
+    const reviewersSet = new Set(
+      await Promise.allSettled(
```
viceice, line 674: doesn't work, `allSettled` returns an object

lib/modules/platform/bitbucket-server/utils.ts
```diff
@@ -218,3 +218,17 @@ export function getExtraCloneOpts(opts: HostRule): GitOptions {
...
+
+// https://docs.atlassian.com/bitbucket-server/rest/5.1.0/bitbucket-rest.html#idm45588158982432
+export async function resolveEmailToUsername(
+  email: string,
+): Promise<string | null> {
+  // GET /rest/api/1.0/admin/users?filter={filter}
+  const users = (
+    await bitbucketServerHttp.getJson<{ values: { name: string }[] }>(
+      `./rest/api/1.0/admin/users?filter=${email}`,
+    )
```
viceice, line 230: use `zod` schema validation for new requests

lib/modules/platform/bitbucket-server/index.ts
mikevader: fixed

lib/modules/platform/bitbucket-server/utils.ts
mikevader: I use now `getJsonUnchecked`.

lib/modules/platform/bitbucket-server/utils.ts
```diff
@@ -6,9 +6,11 @@ import { logger } from '../../../logger';
 import type { HostRule } from '../../../types';
 import type { GitOptions, GitProtocol } from '../../../types/git';
 import * as git from '../../../util/git';
+import type { BitbucketServerHttp } from '../../../util/http/bitbucket-server';
 import { parseUrl } from '../../../util/url';
 import { getPrBodyStruct } from '../pr-body';
 import type { GitUrlOption } from '../types';
+import type { RestDetailedUser } from './schema';
```
viceice, line 13:
```suggestion
import { RestDetailedUser } from './schema';
```

lib/modules/platform/bitbucket-server/utils.ts
```diff
@@ -148,3 +150,19 @@ export function getExtraCloneOpts(opts: HostRule): GitOptions {
...
+// https://docs.atlassian.com/bitbucket-server/rest/5.1.0/bitbucket-rest.html#idm45588158982432
+export async function resolveEmailToUsername(
+  http: BitbucketServerHttp,
+  email: string,
+): Promise<string | null> {
+  // GET /rest/api/1.0/admin/users?filter={filter}
+  const users = (
+    await http.getJsonUnchecked<RestDetailedUser[]>(
+      `./rest/api/1.0/admin/users?filter=${email}`,
+      { paginate: true },
```
viceice, line 163:
```suggestion
    await http.getJsonUnchecked(
      `./rest/api/1.0/admin/users?filter=${email}`,
      { paginate: true },
      z.array(RestDetailedUser),
```

viceice: you can also precreate the userarray in schema.ts to not need it import zod here

PR #31266, opened 2024-09-07
feat(manager): add inline script metadata (PEP 723) support

mkniewallner:
## Changes

Add support for [Python inline script metadata](https://packaging.python.org/en/latest/specifications/inline-script-metadata/), also known as PEP 723.

## Context

Closes https://github.com/renovatebot/renovate/discussions/27889.

Both the regex and the transformation from commented out TOML to actual TOML have been adapted from the Python reference version. Per the recommendation in the linked discussion, since virtually any Python file can contain such metadata, to avoid slowing down Renovate, `fileMatch` defaults to an empty array.

Just a note that I initially wanted to add a section [similar to `pip-compile`](https://github.com/renovatebot/renovate/blob/e064c5e40da736aeb598eec9897e9f9313dc4477/lib/modules/manager/pip-compile/readme.md?plain=1#L7-L18) to mention that we have an empty `fileMatch` by default, but I then realised there [already is code in place](https://github.com/renovatebot/renovate/blob/e064c5e40da736aeb598eec9897e9f9313dc4477/tools/docs/manager.ts#L122-L124) to generate a default section for this case, as can be seen [here](https://docs.renovatebot.com/modules/manager/pip-compile/#file-matching).

Changes tested in https://github.com/mkniewallner/renovate-pep723, where:
- [Dependency dashboard](https://github.com/mkniewallner/renovate-pep723/issues/1) is able to parse both `requests` and `rich` dependencies
- https://github.com/mkniewallner/renovate-pep723/pull/2 bumping `requests` in the 2 files were it is present was created


lib/modules/manager/pep723/schema.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Toml } from '../../../util/schema-utils';
+import { depTypes, pep508ToPackageDependency } from '../pep621/utils';
+
+const Pep723Dep = z
+  .string()
+  .transform((dep) => pep508ToPackageDependency(depTypes.dependencies, dep));
+
+export const Pep723Schema = Toml.pipe(
+  z
+    .object({
```
mkniewallner, line 11: PEP 723 also supports [setting `requires-python`](https://packaging.python.org/en/latest/specifications/inline-script-metadata/#script-type), and I was wondering if this is necessary to have as a Python constraint. I assumed that this is useful if we actually run a tool to update dependencies in a lock file (like Poetry or uv), but maybe it's also useful without a lock file to update?

lib/modules/manager/pep723/schema.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from 'zod';
+import { Toml } from '../../../util/schema-utils';
+import { depTypes, pep508ToPackageDependency } from '../pep621/utils';
+
+const Pep723Dep = z
+  .string()
+  .transform((dep) => pep508ToPackageDependency(depTypes.dependencies, dep));
```
mkniewallner, line 7: `depType` will be `project.dependencies`, like default dependencies for PEP 621. Should we use a specific one for this manager?

viceice: no, I think it's ok as is

lib/modules/manager/pep723/schema.ts
viceice: yes, extract it, the datasource can use it to filter compatible releases

lib/modules/manager/pep723/extract.ts
```diff
@@ -0,0 +1,41 @@
...
+    return null;
+  }
+
+  // Adapted code from the Python reference implementation: https://packaging.python.org/en/latest/specifications/inline-script-metadata/#reference-implementation
+  const parsedToml = matchedContent
+    .split('\n')
+    .map((line) => line.substring(line.startsWith('# ') ? 2 : 1))
+    .join('\n');
+
+  const { val: deps, err } = Result.parse(parsedToml, Pep723Schema).unwrap();
```
viceice, line 30:
```suggestion
  const { val: deps, err } = Pep723Schema.safeParse(parsedToml);
```
much simpler

lib/modules/manager/pep723/extract.ts
```diff
@@ -0,0 +1,41 @@
+import { logger } from '../../../logger';
+import { regEx } from '../../../util/regex';
+import { Result } from '../../../util/result';
+import type { PackageFileContent } from '../types';
+import { Pep723Schema } from './schema';
+
+// Adapted regex from the Python reference implementation: https://packaging.python.org/en/latest/specifications/inline-script-metadata/#reference-implementation
```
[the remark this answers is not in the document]
mkniewallner: The reference implementation was initially published in [PEP 723](https://peps.python.org/pep-0723/), where a [copyright section](https://peps.python.org/pep-0723/#copyright) specifies:
> This document is placed in the public domain or under the CC0-1.0-Universal license, whichever is more permissive.

Not an expert on licenses, but I think this should be fine.

lib/modules/manager/pep723/extract.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+# /// script
+# requires-python = ">=3.11"
+# ///
+`,
+        'foo.py',
+      );
+
+      expect(res).toMatchObject({
+        deps: [],
+        extractedConstraints: { python: '>=3.11' },
```
viceice, line 83: I think you should return null if empty, like we do inside other managers

lib/modules/manager/pep723/extract.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  if (error) {
+    logger.debug(
+      { packageFile, error },
+      `Error parsing PEP 723 inline script metadata`,
+    );
+    return null;
+  }
+
+  return res;
```
viceice, line 39:
```suggestion
  return res.deps.length ? deps : null;
```

lib/modules/manager/pep723/extract.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+          },
+          {
+            currentValue: '>=13.8.0',
+            datasource: 'pypi',
+            depName: 'rich',
+            depType: 'project.dependencies',
+            packageName: 'rich',
+          },
+        ],
+        extractedConstraints: { python: '>=3.11' },
```
rarkins, line 38: Should this also be a dependency so that it can be updated?

mkniewallner: I'm not entirely sure. [`requires-python`'s purpose](https://packaging.python.org/en/latest/specifications/inline-script-metadata/#script-type) is the same as when used in `pyproject.toml`, it defines the minimum version for which a script is compatible. I believe in some situations (e.g. when building a web application), it would be convenient, but in some others (e.g. building a library that supports multiple Python versions), it might be unwanted.

Currently, PEP 621 manager does not consider `requires-python` as a dependency either (it is just used as a constraint). Regardless of what we chose, we should probably have a consistent implement between the 2.

rarkins: Ok let's leave it as is for now

lib/modules/manager/pep723/extract.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+      const res = extractPackageFile(
+        codeBlock`
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#   "requests==2.32.3",
+#   "rich>=13.8.0",
+# ]
+# ///
+`,
```
viceice, line 16: please indent, codeBlock will remove leading space properly

viceice (approved):

rarkins (approved):

PR #31270, opened 2024-09-08
fix(manager/uv): skip unsupported sources

mkniewallner:
## Changes

Skip dependencies that define [a source](https://docs.astral.sh/uv/concepts/dependencies/#dependency-sources) that is not yet supported by Renovate.

## Context

This emerges from discussion in https://github.com/renovatebot/renovate/pull/31186#discussion_r1747961696. This will avoid having Renovate try to search for dependencies in PyPI if the dependencies are actually mapped to other sources (like a git repository, or a local path).

The end result is really similar to [what is done in Cargo](https://github.com/renovatebot/renovate/blob/3374bd1ce1e13da8129804e119c7643a5eef72bd/lib/modules/manager/cargo/extract.ts#L79-L91), and is defensive against potential new sources that could be added.

### DIFF
--- a/lib/modules/manager/pep621/processors/uv.spec.ts
+++ b/lib/modules/manager/pep621/processors/uv.spec.ts
@@ -71,7 +71,8 @@ describe('modules/manager/pep621/processors/uv', () => {
               dep3: { path: '/local-dep.whl' },
               dep4: { url: 'https://example.com' },
               dep5: { workspace: true },
-              dep6: {},
+              dep6: { workspace: false },
+              dep7: {},
             },
           },
         },
@@ -84,6 +85,7 @@ describe('modules/manager/pep621/processors/uv', () => {
         { depName: 'dep4' },
         { depName: 'dep5' },
         { depName: 'dep6' },
+        { depName: 'dep7' },
       ];
 
       const result = processor.process(pyproject, dependencies);
@@ -111,6 +113,10 @@ describe('modules/manager/pep621/processors/uv', () => {
         },
         {
           depName: 'dep6',
+          skipReason: 'invalid-dependency-specification',
+        },
+        {
+          depName: 'dep7',
           skipReason: 'invalid-dependency-specification',
         },
       ]);
--- a/lib/modules/manager/pep621/processors/uv.spec.ts
+++ b/lib/modules/manager/pep621/processors/uv.spec.ts
@@ -26,7 +26,7 @@ describe('modules/manager/pep621/processors/uv', () => {
   describe('process()', () => {
     it('returns initial dependencies if there is no tool.uv section', () => {
       const pyproject = { tool: {} };
-      const dependencies = [{ packageName: 'dep1' }];
+      const dependencies = [{ depName: 'dep1' }];
 
       const result = processor.process(pyproject, dependencies);
 
@@ -37,12 +37,12 @@ describe('modules/manager/pep621/processors/uv', () => {
       const pyproject = {
         tool: { uv: { 'dev-dependencies': ['dep2==1.2.3', 'dep3==2.3.4'] } },
       };
-      const dependencies = [{ packageName: 'dep1' }];
+      const dependencies = [{ depName: 'dep1' }];
 
       const result = processor.process(pyproject, dependencies);
 
       expect(result).toEqual([
-        { packageName: 'dep1' },
+        { depName: 'dep1' },
         {
           currentValue: '==1.2.3',
           currentVersion: '1.2.3',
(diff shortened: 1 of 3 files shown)


lib/modules/manager/pep621/schema.ts
```diff
@@ -8,6 +8,46 @@ const DependencyRecordSchema = z
...
+    )
+    .optional(),
+});
+
+// https://docs.astral.sh/uv/concepts/dependencies/#dependency-sources
+const UvSource = z.object({
+  git: z.string().optional(),
+  path: z.string().optional(),
+  url: z.string().optional(),
+  workspace: z.boolean(z.literal(true)).optional(),
```
mkniewallner, line 43: Setting `workspace = false` currently errors out in uv:
```console
$ uv lock
Using Python 3.12.1
error: Failed to build: `foo @ file:///<REDACTED>`
  Caused by: Failed to parse entry for: `httpx`
  Caused by: `workspace = false` is not yet supported
```

If this ends up being supported, I'm not sure what this will actually do, so probably best to skip this behaviour as `invalid-dependency-specification`.

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -32,6 +33,38 @@ export class UvProcessor implements PyProjectProcessor {
...
+          } else if (depSource.path) {
+            skipReason = 'path-dependency';
+          } else if (depSource.workspace) {
+            skipReason = 'inherited-dependency';
+          } else {
+            skipReason = 'invalid-dependency-specification';
+          }
+
+          if (skipReason) {
+            dep.currentValue = '';
```
rarkins, line 61: I think we usually have the `currentValue` set even if skipping, because it makes it more easy to understand. Any problem if we do that here too?

lib/modules/manager/pep621/processors/uv.spec.ts
rarkins, line 1: Needs at least one test for the skipReason case, ideally a test file with all of these

rarkins (changes requested):

lib/modules/manager/pep621/processors/uv.spec.ts
mkniewallner: Sorry if I'm misunderstanding your comment, but isn't what https://github.com/renovatebot/renovate/pull/31270/files#diff-4e070e322f5eb5bb28ff9cb6c872a7f1c0dff9021ec92f4ae19971d20507f5a9R65 is about?

rarkins (dismissed):

lib/modules/manager/pep621/schema.ts
viceice: I would allow booleans, otherwise the full fold parsing will fail.

lib/modules/manager/pep621/schema.ts
```diff
@@ -8,6 +8,46 @@ const DependencyRecordSchema = z
   .record(z.string(), z.array(z.string()))
   .optional();
 
+const PdmSchema = z.object({
```
viceice, line 11: I would like to see this refactor in a separate PR. 🤗

mkniewallner: Sure thing, thought it was small enough to not require a separate PR, but since 2 PRs depend on it, might be better anyway :)
https://github.com/renovatebot/renovate/pull/31280

viceice (approved):

lib/modules/manager/pep621/processors/uv.spec.ts
```diff
@@ -37,12 +37,12 @@ describe('modules/manager/pep621/processors/uv', () => {
...
         tool: { uv: { 'dev-dependencies': ['dep2==1.2.3', 'dep3==2.3.4'] } },
       };
-      const dependencies = [{ packageName: 'dep1' }];
+      const dependencies = [{ depName: 'dep1' }];
 
       const result = processor.process(pyproject, dependencies);
 
       expect(result).toEqual([
-        { packageName: 'dep1' },
+        { depName: 'dep1' },
```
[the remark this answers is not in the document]
mkniewallner: For PEP 508 dependencies, we apply [some transformation on `packageName`](https://github.com/mkniewallner/renovate/blob/7a65c10eab15809dac9371f8351553b1eeffee1b/lib/modules/manager/pep621/utils.ts#L66) which means we will not map the dependency name to the correct source in some situations. If we were to use `depName`, I guess we would also need to apply the same kind of transformation on the source.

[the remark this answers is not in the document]
mkniewallner: This makes me wonder though if there's some sort of normalisation on the dependency names and/or the sources in uv. I can quickly check that.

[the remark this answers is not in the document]
mkniewallner: From a few tests, there seems to be normalisation on both the dependency names and the sources.

Using [python-multipart](https://pypi.org/project/python-multipart/) package for testing, this `pyproject.toml`:
```toml
[project]
name = "foo"
version = "0.1.0"
dependencies = ["python_multipart"]

[tool.uv.sources]
python-multipart = { url = "https://files.pythonhosted.org/packages/c0/3e/9fbfd74e7f5b54f653f7ca99d44ceb56e718846920162165061c4c22b71a/python_multipart-0.0.8-py3-none-any.whl" }
```

results in having the following entry in `uv.lock`:
```toml
[package.metadata]
requires-dist = [{ name = "python-multipart", url = "https://files.pythonhosted.org/packages/c0/3e/9fbfd74e7f5b54f653f7ca99d44ceb56e718846920162165061c4c22b71a/python_multipart-0.0.8-py3-none-any.whl" }]

[[package]]
name = "python-multipart"
version = "0.0.8"
source = { url = "https://files.pythonhosted.org/packages/c0/3e/9fbfd74e7f5b54f653f7ca99d44ceb56e718846920162165061c4c22b71a/python_multipart-0.0.8-py3-none-any.whl" }
wheels = [
    { url = "https://files.pythonhosted.org/packages/c0/3e/9fbfd74e7f5b54f653f7ca99d44ceb56e718846920162165061c4c22b71a/python_multipart-0.0.8-py3-none-any.whl", hash = "sha256:999725bf08cf7a071073d157a27cc34f8669af98da0d2435bde1cc1493a50ec3" },
]
```

(also tried the reverse logic, `-` for the dependency name and `_` for the source, and it's the same)

Given that, would it makes sense to you to use `depName` as you suggest, and apply [the same normalisation](https://github.com/renovatebot/renovate/blob/dbd17ed1d78595631b10481e9bf97815fb9acccd/lib/modules/datasource/pypi/common.ts#L10-L12) for the source?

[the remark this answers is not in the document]
mkniewallner: Since the PR got merged, made a follow-up for that here: https://github.com/renovatebot/renovate/pull/31297.

PR #31297, opened 2024-09-09
fix(manager/uv): apply normalization on sources

mkniewallner:
## Changes

Apply normalisation when retrieving uv's sources, as uv itself does that when mapping a dependency to a specific source.

## Context

This was discovered in https://github.com/renovatebot/renovate/pull/31270#discussion_r1750746936, but the PR was merged before this could be resolved.

The test moved from `uv.spec.ts` to `extract.spec.ts` allows testing that the normalisation is applied on sources in the schema, which calling `processor.process` does not, as it happens after the normalisation.


lib/modules/manager/pep621/schema.ts
```diff
@@ -45,7 +46,13 @@ const UvSource = z.object({
 
 const UvSchema = z.object({
   'dev-dependencies': DependencyListSchema,
-  sources: z.record(z.string(), UvSource).optional(),
+  sources: z
+    .record(
+      // uv applies the same normalization as for Python dependencies on sources
+      z.string().transform((source) => normalizePythonDepName(source)),
```
mkniewallner, line 52: Note on this: if we have 2 sources with different casing that would actually be normalised to the same name, Renovate will only keep the 2nd one. This is in line with what uv does today, which will likely be fixed to avoid this case anyway (https://github.com/astral-sh/uv/issues/7229).

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -36,11 +36,12 @@ export class UvProcessor implements PyProjectProcessor {
     // Skip sources that are either not yet handled by Renovate (e.g. git), or do not make sense to handle (e.g. path).
     if (uv.sources) {
       for (const dep of deps) {
-        if (!dep.depName) {
+        // istanbul ignore if
+        if (!dep.packageName) {
```
mkniewallner, line 40: Not a case we can test with the new test location, and adding a test just for that in `uv.spec.ts` seems a bit overkill.

lib/modules/manager/pep621/schema.ts
```diff
@@ -65,7 +66,11 @@ const UvSource = z.union([
 
 const UvSchema = z.object({
   'dev-dependencies': DependencyListSchema,
-  sources: LooseRecord(z.string(), UvSource).optional(),
+  sources: LooseRecord(
+    // uv applies the same normalization as for Python dependencies on sources
+    z.string().transform((source) => normalizePythonDepName(source)),
```
mkniewallner, line 71: Note on this: if we have 2 sources with different casing that would actually be normalised to the same name, Renovate will only keep the 2nd one. In reality, uv prevents having 2 sources that normalise to the same one anyway: https://github.com/astral-sh/uv/issues/7229.

lib/modules/manager/pep621/processors/uv.spec.ts
```diff
@@ -149,7 +86,20 @@ describe('modules/manager/pep621/processors/uv', () => {
...
+        depName: 'dep1',
+        packageName: 'dep1',
+      },
+      {
+        depName: 'dep2',
+        packageName: 'dep2',
+      },
+      {
+        depName: undefined,
+        packageName: undefined,
```
secustor, line 100: In which scenario could this happen 🤔

lib/modules/manager/pep621/extract.spec.ts
```diff
@@ -326,6 +328,68 @@ describe('modules/manager/pep621/extract', () => {
       ]);
     });
 
+    it('should skip dependencies with unsupported uv sources', async () => {
```
viceice, line 331: please don't move test with other changes in same PR. it can easily lead to regression.

lib/modules/manager/pep621/processors/uv.spec.ts
viceice: this should not happen

mkniewallner: Yeah, did that because of code coverage complaining that [this condition was not covered](https://github.com/renovatebot/renovate/blob/2e2e83b4d7316715e4627dc9c7be298f69d248a9/lib/modules/manager/pep621/processors/uv.ts#L49-L51), but maybe it's better to add `// istanbul ignore if` on the condition in that specific case.

lib/modules/manager/pep621/extract.spec.ts
mkniewallner: Sure, sorry, thought it was minimal enough to not deserve a dedicated PR, especially since I initially added this test. Created a dedicated PR for this change: https://github.com/renovatebot/renovate/pull/32195.

viceice (approved):

secustor (approved):

PR #31357, opened 2024-09-12
fix(repo/finalize): prune with different base branches

### LINKED ISSUE #31356: Renovate doesn't prune branches which don't target default branch

Renovate doesn't prune branches, when they target a different base branch because it compares it with default branch for modifications.

https://github.com/renovatebot/renovate/blob/fad3dc889db7b2b3a2086d4dd5390e2e70c6fb4e/lib/workers/repository/finalize/prune.ts#L28-L31

We need to pass the base branch for each renovate branch.

viceice:
## Changes
- pass correct `baseBranch` to PR `targetBranch`
- extract correct `baseBranch` from `branchName`

## Context
- closes #31356

### DIFF
--- a/lib/workers/repository/finalize/prune.spec.ts
+++ b/lib/workers/repository/finalize/prune.spec.ts
@@ -82,6 +82,38 @@ describe('workers/repository/finalize/prune', () => {
       expect(platform.updatePr).toHaveBeenCalledTimes(1);
     });
 
+    it('deletes with base branches', async () => {
+      config.branchList = ['renovate/main-a'];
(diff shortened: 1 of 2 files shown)


rarkins (commented):
Could this be derisked by bypassing all new logic whenever baseBranches is not configured?

viceice:
> Could this be derisked by bypassing all new logic whenever baseBranches is not configured?

what do you mean? the calculate function returns null if no base branches are configured. then the default branch is used as before.

rarkins (approved):

PR #30582, opened 2024-08-05
feat(config): `platformOptions`

RahulGautamSingh:
## Changes
- add `platformOptions` slef-hosted config option
- convert `RENOVATE_X_PLATFORM_VERSION` to a self-hosted config option `platformVersion` within parent `platformOptions`

## Context
Closes: https://github.com/renovatebot/renovate/issues/29716

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository 
- platformVersion: tested locally as `platformVersion` is self-hosted config option

PR #31381, opened 2024-09-13
feat(config): `prOptions`

RahulGautamSingh:
## Changes
- Add new config option: `prOptions`
- Migrate `azureWorkItemId`, `gitLabIgnoreApprovals`, `bbUseDefaultReviewers` 

## Context
Ref: https://github.com/renovatebot/renovate/pull/30582#pullrequestreview-2260380637


docs/usage/configuration-options.md
```diff
@@ -3466,6 +3450,38 @@ This is why we configured an upper limit for how long we wait until creating a P
...
+
+```json
+{
+  "prOptions": {
+    "gitLabIgnoreApprovals": true
+  }
+}
+```
+
+You can use these options:
```
HonkingGoose, line 3467:
```suggestion
You can use these configuration options within `prOptions`:

- `azureWorkItemId`
- `bbUseDefaultReviewers`
- `gitLabIgnoreApprovals`
```

Explain that we mean you can use these options within `prOptions`, and list the options.

Even better would be to autogenerate a table of allowed options within `prOptions`. That way, when we later add new options, the docs are updated automatically. I think we already do this for other bits of the docs.

docs/usage/configuration-options.md
```diff
@@ -3466,6 +3450,38 @@ This is why we configured an upper limit for how long we wait until creating a P
...
+
+### azureWorkItemId
+
+When creating a PR in Azure DevOps, some branches can be protected with branch policies to [check for linked work items](https://docs.microsoft.com/en-us/azure/devops/repos/git/branch-policies?view=azure-devops#check-for-linked-work-items).
+Creating a work item in Azure DevOps is beyond the scope of Renovate, but Renovate can link an already existing work item when creating PRs.
```
HonkingGoose, line 3472:
```suggestion
Renovate will not create a work item in Azure DevOps, this is beyond the scope of Renovate.
But Renovate can _link_ to an existing work item when Renovate creates a PR.
```

Split up sentence into two parts.

Can you configure a `azureWorkItemId` and then forget about it, as it just works? Or do you need to update the `azureWorkItemId` later?

docs/usage/configuration-options.md
```diff
@@ -3466,6 +3450,38 @@ This is why we configured an upper limit for how long we wait until creating a P
...
+
+### bbUseDefaultReviewers
+
+Configuring this to `true` means that Renovate will detect and apply the default reviewers rules to PRs (Bitbucket only).
```
HonkingGoose, line 3476:
```suggestion
If `bbUseDefaultReviewers` is set to `true`, Renovate detects and applies the default reviewers rules to its PRs (Bitbucket only).
```

docs/usage/configuration-options.md
```diff
@@ -3466,6 +3450,38 @@ This is why we configured an upper limit for how long we wait until creating a P
...
+
+### gitLabIgnoreApprovals
+
+Ignore the default project level approval(s), so that Renovate bot can automerge its merge requests, without needing approval(s).
```
HonkingGoose, line 3480:
```suggestion
Ignore the default project level approval(s), to allow Renovate to automerge its Merge Requests, without needing approval(s).
```

docs/usage/configuration-options.md
```diff
@@ -3466,6 +3450,38 @@ This is why we configured an upper limit for how long we wait until creating a P
...
+Under the hood, it creates a MR-level approval rule where `approvals_required` is set to `0`.
```
HonkingGoose, line 3481:
```suggestion
Under the hood, it creates a MR-level approval rule, where `approvals_required` is set to `0`.
```

docs/usage/configuration-options.md
```diff
@@ -3466,6 +3450,38 @@ This is why we configured an upper limit for how long we wait until creating a P
...
+This option works only when `automerge=true` and either `automergeType=pr` or `automergeType=branch`.
```
HonkingGoose, line 3482:
```suggestion
For `gitLabIgnoreApprovals` to work you must set `automerge=true`, and either set `automergeType=pr` _or_ set `automergeType=branch`.
```

docs/usage/configuration-options.md
```diff
@@ -3466,6 +3450,38 @@ This is why we configured an upper limit for how long we wait until creating a P
...
+Also, approval rules overriding should not be [prevented in GitLab settings](https://docs.gitlab.com/ee/user/project/merge_requests/approvals/settings.html#prevent-editing-approval-rules-in-merge-requests).
```
HonkingGoose, line 3483:
```suggestion
Also, approval rules overriding should _not_ be [prevented in GitLab settings](https://docs.gitlab.com/ee/user/project/merge_requests/approvals/settings.html#prevent-editing-approval-rules-in-merge-requests).
```

HonkingGoose (changes requested):

rarkins (changes requested):
I'm not sure this is right for Renovate config. Nesting is normally for a reason, and in this case it's being used purely because "these config options relate to similar things".

RahulGautamSingh:
I agree with yout point, this might be a bit early for this option as there aren't as any pr-specific options.

But, I think it is good have them collected under one parent. The internal logic flow is distinct now that we have a parent, and adding a new pr-specific options will require less effort/time. And, in future if more pr-specific options are added then users will find it easier documented at one place.

rarkins:
I don't find this argument to be compelling. Such change in config approach is significant and wouldn't end here

RahulGautamSingh:
Okay. I will close this in that case.

PR #31387, opened 2024-09-14
improve commit message for Nix lockfile maintenance

monperrus:
## Changes

improve commit message for Nix lockfile maintenance

## Context

Why: because changelogs with generic "build(deps): lock file maintenance" are not self-explanatory


rarkins:
Renovate supports lock file maintenance for many package managers and has a generic name for all, and sometimes combining them in one PR, so I'm not sure of this change just for nix. You should be able to change this yourself through config without needing Renovate to change its defaults for everyone. 

The build(deps) part is optional too - able to be removed by config.

viceice (changes requested):
don't like this change too

monperrus:
Ack, thanks for the feedback

PR #31393, opened 2024-09-15
feat: remediate to greater than or equal versions for github alerts

rarkins:
## Changes

Adds new field `vulnerabilityFixVersion` to use in place of `allowedVersions` for vulnerability rules.

## Context

When we (currently) use `allowedVersions` for vulnerability fixes then it means we override/break any `allowedVersions` rules which users have configured.


rarkins:
Another reason for this refactor was to be able to enable #31395

rarkins:
@Churro I didn't make the changes to OSV alerts yet. This functionality is intended to work side-by-side and allow a later refactoring

lib/workers/repository/init/vulnerability.ts
```diff
@@ -174,10 +174,6 @@ export async function detectVulnerabilityAlerts(
           logger.warn({ err }, 'Error generating vulnerability PR notes');
         }
         // TODO: types (#22198)
-        const allowedVersions =
-          datasource === PypiDatasource.id
-            ? `==${val.firstPatchedVersion!}`
```
viceice, line 179: are you sure to drop this condition for pypi?

rarkins: I think this new approach is superior (it's essentially greater than or equal to). Maybe a feature not a refactor though if results can change

viceice: feature 👍

lib/workers/repository/process/lookup/index.ts
```diff
@@ -372,6 +372,40 @@ export async function lookupUpdates(
       );
       let shrinkedViaVulnerability = false;
       if (config.isVulnerabilityAlert) {
+        if (config.vulnerabilityFixVersion) {
+          res.vulnerabilityFixVersion = config.vulnerabilityFixVersion;
+          if (versioning.isVersion(config.vulnerabilityFixVersion)) {
+            // Filter out versions if the vulnerabilityFixVersion is higher
+            const fixedFilteredReleases = filteredReleases.filter(
+              (r) =>
+                !versioning.isGreaterThan(
```
Churro, line 381: I understand that this works under the assumption `config.vulnerabilityFixVerison` is set as an exact version. For GitHub alerts, `first_patched_version` always includes an exact version. which was defined as a constraint in `allowedVersions` before the regression happened. Otherwise it would be this:
https://github.com/renovatebot/renovate/blob/ec509ab063f56c620093f507c4c30356ef3c8df7/lib/workers/repository/init/vulnerability.ts#L46-L51

For OSV, depending on the concrete advisory there are two kinds of fix versions possible, "first patched" and "last affected", which is why the constraint was either `>= fixVersion` or `> fixVersion`, respectively.

I'm afraid this approach could not distinguish between these two cases. How about making `vulnerabilityFixVersion` a constraint, analogous to `allowedVersions` and applying the same filtering procedure?

lib/workers/repository/init/vulnerability.ts
rarkins: I've rephrased this to feature and will add some extra tests

lib/workers/repository/process/lookup/index.ts
rarkins: I updated the code to allow both version or range

lib/workers/repository/process/lookup/types.ts
```diff
@@ -49,6 +49,7 @@ export interface LookupUpdateConfig
   replacementNameTemplate?: string;
   replacementVersion?: string;
   extractVersion?: string;
+  vulnerabilityFix?: string;
```
Churro, line 52: Shouldn't this be defined in the `PackageRule` interface instead (where `isVulnerabilityAlert` is too)? Same below

Churro (commented):
Since `matchCurrentVersion` also matches version or range, for consistency I'd advocate calling the field `vulnerabilityFixVersion`. With `vulnerabilityFix` I'm somewhat lacking the "what it is"-suffix.

I tested this PR on https://github.com/renovate-demo/31393-github-alerts/pulls and the result of PRs created is basically identical to #29600 and the reproducer in https://github.com/renovate-demo/29600-github-vuln-alerts/pulls.

Churro:
I also tested it with OSV, changing only all occurrences of `allowedVersions` to `vulnerabilityFix`. The PR output is identical to the ones raised via GitHub alerts (https://github.com/renovate-demo/31393-osv-alerts/pulls) but the log included a warning:
```
DEBUG: Vulnerability GHSA-g3vv-g2j5-45f2 affects github.com/ipld/go-codec-dagpb v1.3.0 (repository=renovate-demo/31393-osv-alerts)
DEBUG: Setting allowed version >= 1.3.1 to fix vulnerability GHSA-967g-cjx4-h7j6 in github.com/ipld/go-codec-dagpb v1.3.0 (repository=renovate-demo/31393-osv-alerts)

 WARN: vulnerabilityFix is not valid (repository=renovate-demo/31393-osv-alerts)
       "vulnerabilityFix": ">= 1.3.1",
       "packageName": "github.com/ipld/go-codec-dagpb"
DEBUG: Vulnerability alert found: limiting results to a single release (repository=renovate-demo/31393-osv-alerts)
       "filteredReleases": [{"version": "v1.3.1", "releaseTimestamp": "2022-03-16T09:19:38.000Z"}]
```

The issue is that go uses semver and `semver.isValid(">= 1.3.1")` is false. It worked with `allowedVersions` due to this fallback:
https://github.com/renovatebot/renovate/blob/ffe2b4c1882f6d96b62396bb96c29666cb9875ee/lib/workers/repository/process/lookup/filter.ts#L78-L95

However, this shouldn't be an issue with GitHub alerts, as they always return a single version. Just mentioning it with regards to porting this change also to OSV. If OSV advisories use "last affected", it will inevitably result in a constraint like "> 1.2.3".

lib/workers/repository/process/lookup/types.ts
rarkins: Unfortunately adding it there doesn't remove the need to have it in the two types in this file too

lib/workers/repository/process/lookup/index.ts
```diff
@@ -372,6 +372,46 @@ export async function lookupUpdates(
...
+              ? filteredReleases.filter(
+                  (r) =>
+                    !versioning.isGreaterThan(
+                      config.vulnerabilityFix!,
+                      r.version,
+                    ),
+                )
+              : filteredReleases.filter((r) =>
+                  versioning.matches(r.version, config.vulnerabilityFix!),
+                );
```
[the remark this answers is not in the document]
rarkins: Change accepted/applied manually

rarkins:
@Churro regarding using OSV, can we return vulnerabilityFixVersion as a single version? The warning you describe above is due to Renovate go versioning not having the concept of range syntax plus the lack of semver fallback you already identified. The thing is that in go versioning, ">= 1.3.1" is unnecessary anyway - "1.3.1" is essentially the same thing (although technically it's like ^1.3.1 and does not go past 2.0.0).

Churro:
> @Churro regarding using OSV, can we return vulnerabilityFixVersion as a single version? The warning you describe above is due to Renovate go versioning not having the concept of range syntax plus the lack of semver fallback you already identified. The thing is that in go versioning, ">= 1.3.1" is unnecessary anyway - "1.3.1" is essentially the same thing (although technically it's like ^1.3.1 and does not go past 2.0.0).

With OSV, it depends which fix info is provided in the advisory:
- with explicit fix version -> same `versioning.isGreaterThan()` check will work as for `first_patched_version` in GH alerts
- with only a "last affected" version, there will be a constraint like `> 1.3.1`. Not `>=` but really `>`, which is why specifying "1.3.1" only wouldn't be enough

Churro (approved):

lib/workers/repository/process/lookup/index.ts
```diff
@@ -372,6 +372,45 @@ export async function lookupUpdates(
...
       let shrinkedViaVulnerability = false;
       if (config.isVulnerabilityAlert) {
+        if (config.vulnerabilityFixVersion) {
+          res.vulnerabilityFixVersion = config.vulnerabilityFixVersion;
+          if (versioning.isValid(config.vulnerabilityFixVersion)) {
+            // Filter out versions if the vulnerabilityFixVersion is higher
+            const filterCondition = (release: Release): boolean =>
+              versioning.isVersion(config.vulnerabilityFixVersion)
+                ? !versioning.isGreaterThan(
+                    config.vulnerabilityFixVersion!,
```
viceice, line 382:
```suggestion
                    config.vulnerabilityFixVersion,
```

this is already validated in if above or is typescript not able to follow 🤔

lib/workers/repository/process/lookup/index.ts
```diff
@@ -372,6 +372,45 @@ export async function lookupUpdates(
...
+                    release.version,
+                  )
+                : versioning.matches(
+                    release.version,
+                    config.vulnerabilityFixVersion!,
```
viceice, line 387:
```suggestion
                    config.vulnerabilityFixVersion,
```

same here

lib/workers/repository/process/lookup/index.ts
```diff
@@ -372,6 +372,45 @@ export async function lookupUpdates(
       );
       let shrinkedViaVulnerability = false;
       if (config.isVulnerabilityAlert) {
+        if (config.vulnerabilityFixVersion) {
+          res.vulnerabilityFixVersion = config.vulnerabilityFixVersion;
+          if (versioning.isValid(config.vulnerabilityFixVersion)) {
+            // Filter out versions if the vulnerabilityFixVersion is higher
+            const filterCondition = (release: Release): boolean =>
```
viceice, line 379: this inline function is hard to read, can you move it to module scope and add a tsdoc comment there?

viceice (commented):
otherwise LGTM

lib/workers/repository/process/lookup/index.ts
rarkins: TS doesn't follow it somehow

lib/workers/repository/process/lookup/index.ts
rarkins: Refactored a different way

lib/workers/repository/process/lookup/index.ts
```diff
@@ -372,6 +372,51 @@ export async function lookupUpdates(
...
+              // Retain only releases greater than or equal to the fix version
+              fixedFilteredReleases = filteredReleases.filter(
+                (release) =>
+                  !versioning.isGreaterThan(
+                    config.vulnerabilityFixVersion!,
+                    release.version,
+                  ),
+              );
+            } else {
+              // Retain only releases which max the fix constaraint
```
viceice, line 389:
```suggestion
              // Retain only releases which max the fix constraint
```

viceice (approved):
we should split to smaller functions in a follow up PR

PR #31414, opened 2024-09-16
fix(poetry): poetry versioning should match == with locked version

PhilipAbed:


viceice (commented):
🤔 it seems you need to fix more tests

PhilipAbed:
@viceice 
the tests are failing because poetry.isValid uses the same function for turning poetry into NPM
that means if i add this change ` .replace(/==/, '');` 
it will make all `==` versions poetry valid.

and the test fails because it expects the `==` to not be valid for poetry
from python versioning.

```
function isLessThanRange(version: string, range: string): boolean {
  return poetry.isValid(range)
    ? poetry.isLessThanRange!(version, range)
    : pep440.isLessThanRange!(version, range);
 ```

PhilipAbed:
it's just too dangerous to change, lets avoid this.

PR #31423, opened 2024-09-16
fix(gitlab): correctly detect merged results pipelines

sichapman:
## Changes

A new branch status is now only added if the head_pipeline's sha differs from the known commit sha, in order to prevent unwanted automerges

## Context

See discussion #31375.  The new branch status entry can accidentally trigger automerge if the underlying pipeline it represents is actually a renovate/* pipeline, as it will prevent the subsequent !internalChecksAsSuccess check from working. Since the purpose of the new branch status entry is to represent GitLab's premium "merged results pipeline" feature, the check has been tightened up to only add the status if the head pipeline is one of those.

I have tested against a real repo that there is no more erroneous automerge when the head pipeline is actually a `renovate/stability-days` pipeline. I can't test a merged results pipeline scenario directly, due to lack of access to a premium Gitlab, but have added tests to compensate.


lib/modules/platform/gitlab/index.ts
```diff
@@ -482,10 +482,13 @@ export async function getBranchStatus(
...
-  const mrStatus = (await getBranchPr(branchName))?.headPipelineStatus;
-  if (!is.undefined(mrStatus)) {
+  const mr = await getBranchPr(branchName);
+  if (mr?.sha !== mr?.headPipelineSha) {
+    logger.debug(
+      'Merge request head pipeline has different sha to commit, assuming merged results pipeline',
+    );
     branchStatuses.push({
-      status: mrStatus as BranchState,
+      status: mr?.headPipelineStatus as BranchState,
```
viceice, line 491:
```suggestion
      status: mr.headPipelineStatus as BranchState,
```
should not be undefined here , validate existence

viceice (commented):
otherwise LGTM

lib/modules/platform/gitlab/index.ts
```diff
@@ -482,10 +482,13 @@ export async function getBranchStatus(
   }
   logger.debug(`Got res with ${branchStatuses.length} results`);
 
-  const mrStatus = (await getBranchPr(branchName))?.headPipelineStatus;
-  if (!is.undefined(mrStatus)) {
+  const mr = await getBranchPr(branchName);
+  if (mr?.sha !== mr?.headPipelineSha) {
```
viceice, line 486:
```suggestion
  if (mr && mr.sha !== mr.headPipelineSha && mr.headPipelineStatus) {
```

rarkins (approved):

PR #31430, opened 2024-09-16
refactor(datasource): Simplify merge registry logic

zharinov:


lib/modules/datasource/artifactory/__snapshots__/index.spec.ts.snap
```diff
@@ -5,18 +5,22 @@ exports[`modules/datasource/artifactory/index getReleases parses real data (file
   "registryUrl": "https://jfrog.company.com/artifactory",
   "releases": [
     {
+      "registryUrl": "https://jfrog.company.com/artifactory",
```
rarkins, line 8: Previously we tried to avoid having result.release.registryUrl defined if it was the same as result.registryUrl.

My ideal approach would be:
- If all results come from one registry then set result.registryUrl
- If results come from a mix of registries then unset result.registryUrl and set every result.release.registryUrl

Does that align with what you're trying to do?

viceice (approved):
hopefully it still works as expected 🤞

lib/modules/datasource/index.ts
```diff
@@ -161,53 +162,63 @@ async function mergeRegistries(
...
+      if (!combinedRes) {
         combinedRes = res;
+        continue;
+      }
+
+      if (commonRegistryUrl) {
+        for (const release of coerceArray(combinedRes.releases)) {
+          release.registryUrl ??= combinedRes.registryUrl;
+        }
+        commonRegistryUrl = false;
```
rarkins, line 183: Isn't this a complicated way of checking if it's the first registry or not?

zharinov: The second registry, not the first. It avoids unnecessary iterations in case of 3 and more registries, and communicates the idea of sharing common registry url more clearly.

PR #31467, opened 2024-09-18
feat(crate): Support `releaseTimestamp`

### LINKED ISSUE #17764: Consider release date on crates.io for minimumReleaseAge

The `stabilityDays` configuration does currently not work for rust dependencies. This is caused by the fact that no release date is fetched from crates.io. 

Renovate bot currently relies on the standard registry schema, which does not contain a release date. The crates.io API exposes a release date for all crate versions, though.

We would like to use the `stabilityDays` option with rust dependencies.

Version info for crate.io can be fetched either for all versions via `https://crates.io/api/v1/crates/<crate>/versions` or for individual versions via `https://crates.io/api/v1/crates/<crate>/<version>`.

No

zharinov:
## Context

- Closes: #17764


lib/modules/datasource/crate/index.ts
```diff
@@ -371,4 +377,39 @@ export class CrateDatasource extends Datasource {
...
     return [packageName.slice(0, 2), packageName.slice(2, 4), packageName];
   }
+
+  private static releaseTimestampSchema = z
+    .object({
+      version: z.object({
+        created_at: z.string(),
+      }),
+    })
+    .transform(({ version: { created_at } }) => created_at);
```
rarkins, line 387: Don't we normally put these in a schema.ts?

lib/modules/datasource/crate/index.ts
```diff
@@ -371,4 +377,39 @@ export class CrateDatasource extends Datasource {
...
+
+  @cache({
+    namespace: `datasource-crate`,
+    key: (
+      { registryUrl, packageName }: PostprocessReleaseConfig,
+      { version }: Release,
+    ) => `postprocessRelease:${registryUrl}:${packageName}:${version}`,
+    ttlMinutes: 24 * 60,
```
rarkins, line 395: Are releases immutable on crates.io? If so then we should think about the maximum feasible cache time for this - perhaps a week?

lib/modules/datasource/crate/index.ts
```diff
@@ -371,4 +377,39 @@ export class CrateDatasource extends Datasource {
...
+  ): Promise<Release | null> {
+    if (registryUrl !== 'https://crates.io') {
+      return release;
+    }
+
+    const url = `https://crates.io/api/v1/crates/${packageName}/${release.version}`;
+    const { body: releaseTimestamp } = await this.http.getJson(
+      url,
+      CrateDatasource.releaseTimestampSchema,
+    );
```
rarkins, line 411: Does this need a try/catch for safety?

lib/modules/datasource/crate/index.ts
rarkins: We should also make sure we've considered the scenario where we get a temporary error (including 429) and don't cache that for long

rarkins (changes requested):

lib/modules/datasource/crate/index.ts
zharinov: I think we're okay with standard error handling, which just returns the release object intact

lib/modules/datasource/crate/index.ts
zharinov: It could be yanked, and also there is `updated_at` field, so let's not consider it immutable.

> We should also make sure we've considered the scenario where we get a temporary error (including 429) and don't cache that for long

With current cache decorator implementation, it's hard to set the TTL a posteriori. If the error was thrown, we could either not cache it, or we could catch and return the cacheable null (without distinguishing the reason).

lib/modules/datasource/crate/index.ts
zharinov: Yes, didn't feel like it creating the new file, will do this later

lib/modules/datasource/crate/index.ts
rarkins: If updated_at is only triggered for fields we don't care about, then we can ignore updated_at. And wouldn't the `yanked` status be reflected in our regular crates.io lookup which happens before this API call?

zharinov: Okay, you're right

lib/modules/datasource/crate/index.ts
viceice: at least move it out of the class

0xarvin:
@rarkins @viceice another review plz 🙏

lib/modules/datasource/crate/index.ts
zharinov: Changed schema to be nullable

rarkins (commented):
Do we need any additional rate limiting added?

zharinov:
> Do we need any additional rate limiting added?

I don't think so

rarkins (approved):

PR #31481, opened 2024-09-19
fix(instrumentation/otel): only use strings for span status and record exception

secustor:
## Changes
- Use `recordException` to in case of rejections
- massage error messages to string
- remove usage of deprecated constants

## Context
- https://github.com/renovatebot/renovate/discussions/31459

### DIFF
--- a/lib/instrumentation/utils.ts
+++ b/lib/instrumentation/utils.ts
@@ -1,3 +1,5 @@
+import is from '@sindresorhus/is';
+
 export function isTracingEnabled(): boolean {
   return isTraceDebuggingEnabled() || isTraceSendingEnabled();
 }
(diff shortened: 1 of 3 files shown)


lib/instrumentation/utils.ts
```diff
@@ -9,3 +11,13 @@ export function isTraceDebuggingEnabled(): boolean {
...
 }
+
+export function massageThrowable(e: unknown): string | undefined {
+  if (is.nullOrUndefined(e)) {
+    return undefined;
+  }
+  if (e instanceof Error) {
+    return e.message;
+  }
+  return String(e);
```
viceice, line 22: I don't like this

viceice (approved):

PR #31512, opened 2024-09-21
docs(usage): clarify that `packageRules` must have a selector

jamie-tm:
## Changes
Per [this discussion](https://github.com/renovatebot/renovate/discussions/31509), I updated the documentation to match the behavior and "selector" terminology consistent with `packageRules` validation.

## Context
See the linked discussion.

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2410,12 +2410,11 @@ The order of rules matters, because later rules may override configuration optio
 
 The matching process for a package rule:
 
-- Each package rule must include at least one `match...` matcher to include or exclude dependencies.
-- Each matcher must contain at least one pattern and may contain a mix of positive and negative patterns.
-- A matcher includes a package if it matches at least one positive pattern, or if there are no positive patterns.
-- A matcher exludes a package if it matches any negative pattern.
-- If a package is both included and excluded, the negative pattern wins and the package is excluded.
-- A package rule matches a dependency if it is included by every matcher (they're AND-ed together).
+- Each package rule must include at least one `match...` matcher.
+- If multiple matchers are included in one package rule, all of them must match.
+- Each matcher must contain at least one pattern. Some matchers allow both positive and negative patterns.
+- If a matcher includes any positive patterns, it must match at least one of them.
+- A matcher returns `false` if it matches _any_ negative pattern, even if a positive match also occurred.
 
 For more details on positive and negative pattern syntax see Renovate's [string pattern matching documentation](./string-pattern-matching.md).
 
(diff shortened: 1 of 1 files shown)


rarkins (changes requested):
We no longer have "excludeX" in packageRules so I would prefer to use the terminology "matcher" instead of "selector" everywhere which I think is more intuitive

jftanner:
> We no longer have "excludeX" in packageRules 

I didn't even notice that! I'll switch it back to "matcher" and remove the pre-existing references to `exclude...`

jftanner:
I spend some time reviewing the underlying code to make sure I _really_ understand the actual behavior. It took me a few minutes to figure out properly, but I really appreciate the elegance of it. Very intuititive.

Alas, I'm not a technical writer by trade, but I did the best I could. I started with the intent of changing as little as possible, but wound up largely rewriting the whole list. If you'd prefer it go back to how it was, with only the second bullet updated, I can do that.

I also took out the list at the end of the last bullet. It seems to me like it's a subset of the total list, but that those particular matchers don't behave any differently. So, I removed it in favor of letting the table-of-contents speak for itself. Of course, if I'm misunderstanding something, I can put it back.

docs/usage/configuration-options.md
```diff
@@ -2410,14 +2410,14 @@ The order of rules matters, because later rules may override configuration optio
...
-- Combining multiple matchers will restrict the resulting matches (they're AND-ed together):
-  `matchCurrentVersion`, `matchCurrentValue`, `matchNewValue`, `matchConfidence`, `matchCurrentAge`,
-  `matchManagers`, `matchDatasources`, `matchCategories`, `matchDepTypes`, `matchUpdateTypes`,
-  `matchRepositories`, `matchBaseBranches`, `matchFileNames`
+- Each package rule must include at least one `match...` matcher to include or exclude dependencies.
+- Each matcher must contain at least one pattern and may contain a mix of positive and negative patterns.
+- A matcher includes a package if it matches at least one positive pattern, or if there are no positive patterns.
+- A matcher exludes a package if it matches any negative pattern.
+- If a package is both included and excluded, the negative pattern wins and the package is excluded.
+- A package rule matches a dependency if it is included by every matcher (they're AND-ed together).
```
rarkins, line 2418:
```suggestion
- Each package rule must include at least one `match...` matcher.
- If multiple matchers are included in one package rule, all of them must match.
- Each matcher must contain at least one pattern. Some matchers allow both positive and negative patterns.
- If a matcher includes any positive patterns, it must match at least one of them.
- A matcher returns `false` if it matches _any_ negative pattern, even if a positive match also occurred.
```

rarkins (approved):

jftanner:
I like your wording better. Thanks!

HonkingGoose (commented):
Looks good to me!

PR #31612, opened 2024-09-25
fix(vulnerabilities): set matchCurrentVersion for github alerts

rarkins:
## Changes

Adds `matchCurrentVersion` to package rules generated from github alerts

## Context

My `vulnerabilityFixVersion` exposed an existing problem and made it worse. We've never had `matchCurrentVersion` in our alert rules, and there's an edge case where a dependency exists both as a direct dependency plus a vulnerable indirect dependency (lock file). In such cases the non-vulnerable direct dependency would match the rule intended for the indirect dependency.

### DIFF
--- a/lib/workers/repository/init/vulnerability.ts
+++ b/lib/workers/repository/init/vulnerability.ts
@@ -184,9 +184,15 @@ export async function detectVulnerabilityAlerts(
           matchFileNames,
         };
 
+        let matchCurrentVersion = `< ${val.firstPatchedVersion}`;
+        if (datasource === MavenDatasource.id) {
+          matchCurrentVersion = `(,${val.firstPatchedVersion})`;
+        }
+
         // Remediate only direct dependencies
         matchRule = {
           ...matchRule,
+          matchCurrentVersion,
           vulnerabilityFixVersion: val.firstPatchedVersion,
           prBodyNotes,
           isVulnerabilityAlert: true,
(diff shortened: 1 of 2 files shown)


rarkins:
Using the affected range in alerts is not a quick fix either because:
- GitHub uses a generic but non-standard syntax for all ecosystems which our `versioning` rejects, and
- We combine multiple alerts, which might mean multiple ranges, and we don't have reliable logic for combining them.

viceice (approved):

Churro:
> there's an edge case where a dependency exists both as a direct dependency plus a vulnerable indirect dependency (lock file). In such cases the non-vulnerable direct dependency would match the rule intended for the indirect dependency.

Doesn't that require that indirect deps are also extracted? As far as I'm aware this is only the case for `go` manager and there it's an opt-in setting. In that case, vulnerability fix PRs for indirect deps tend to be a bit short-lived, since the next run of `go mod tidy` will revert changes to `go.sum`. Probably, I'm wrong on this because there is already a check to prevent changes to `go.sum`:
https://github.com/renovatebot/renovate/blob/1c5a6d6a50bcab532c99277268e56dc75c479f31/lib/workers/repository/init/vulnerability.ts#L177-L180

lib/workers/repository/init/vulnerability.ts
```diff
@@ -184,9 +184,15 @@ export async function detectVulnerabilityAlerts(
           matchFileNames,
         };
 
+        let matchCurrentVersion = `< ${val.firstPatchedVersion}`;
+        if (datasource === MavenDatasource.id) {
```
Churro, line 188: This should be extended by `|| datasource === NugetDatasource.id` , as otherwise nuget fix PRs are no longer created / existing ones closed.

PR #31616, opened 2024-09-25
feat: use comments if PR body is too long

darrexx:
## Changes

While creating PRs it is now checked if the maximum body length of the platform fits the whole PR body.

When the body is longer, some parts of the body are moved into comments.

For Onboarding PRs first the PR list is moved to the comments, and if the body is still too large, also the Detected Package Files are moved into comments.

For Update PRs first the changelog is moved to the comments, and if the body is still too large, also the Update Table is moved into comments.

If the body is still too long, it will be truncated, like it is now. Also if the platform limit for comment length is met, the comments are also truncated

## Context

Our Team recently added renovate and we instantly noticed that the PR Body was missing information as the body limit on azure devops is 4000 characters.

Then we found [this discussion](https://github.com/renovatebot/renovate/discussions/17477) with the idea to put some parts of the body into comments.


rarkins (commented):
Previously `smartTruncate()` was exclusively a platform concept and used only in platform modules. With this PR it's now used in many places with lib/workers, which seems undesirable.

Did you try solving this in the platform only, to avoid this undesirable mixing of logic?

darrexx:
I understand your point regarding `smartTruncate()`.
 
After initally looking at the code, it seemed easier to fix this problem for all platforms. While `smartTruncate()` was called by `massageMarkdown()`, it was not possible to do this on the platform independent layer for all platforms, as the body would be truncated before the platform independent layer could check if the limit is read, so we moved it into there.
 
I think it is probably still better to adjust this for all platforms to ensure that no information is lost. In our tests we have had some PRs which were larger than 150000 characters, which would be larger than other length limits as well (e.g. bitbucket 50000)
 
As another Solution, we could maybe create a second function, like `massageMarkdown()`, which is called on the platform with the different contents of body & comments as parameters, to ensure that no text limit is reached.

rarkins:
@darrexx would you mind joining the Renovate Slack channel so we can brainstorm this in a bit more real-time? Please send me an email rhys<AT>arkins.net so I know which email to send to

sport24-rtr:
@darrexx / @rarkins any news on this or is the PR abandoned?
This feature is the missing piece for us to confidently integrate Renovate into our workflow.

sport24-rtr:
That was a shame...

For anyone in the future who might stumble upon this, we ended up patching a fix in an internal fork.
Multiple tests fail as a result of this, and I don't have time to look at that, so I will not post it as a PR.

Here's a patch for Azure Devops that will post in a comment rather than in the PR body to circumvent the 4000 char limit of devops:
```
From 2cd9c3c5192a111a1083e8a9700cae49d7bb6803 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Richard=20Tr=C3=B8ndheim?= <rtr@sport24.dk>
Date: Mon, 24 Mar 2025 10:15:15 +0100
Subject: [PATCH] Add comment instead of description

---
 lib/modules/platform/azure/index.ts | 14 +++++++++++---
 1 file changed, 11 insertions(+), 3 deletions(-)

diff --git a/lib/modules/platform/azure/index.ts b/lib/modules/platform/azure/index.ts
index bd0b0a44f..81c42e094 100644
--- a/lib/modules/platform/azure/index.ts
+++ b/lib/modules/platform/azure/index.ts
@@ -55,7 +55,6 @@ import {
   getRepoByName,
   getStorageExtraCloneOpts,
   mapMergeStrategy,
-  max4000Chars,
 } from './util';
 
 interface Config {
@@ -478,7 +477,7 @@ export async function createPr({
 }: CreatePRConfig): Promise<Pr> {
   const sourceRefName = getNewBranchName(sourceBranch);
   const targetRefName = getNewBranchName(targetBranch);
-  const description = max4000Chars(sanitize(body));
+  const description = 'Dependency update. See comments!';
   const azureApiGit = await azureApi.gitApi();
   const workItemRefs = [
     {
@@ -496,6 +495,11 @@ export async function createPr({
     },
     config.repoId,
   );
+  await ensureComment({
+    number: pr.pullRequestId ?? 0,
+    topic: 'dependency_update',
+    content: sanitize(body),
+  });
   if (platformPrOptions?.usePlatformAutomerge) {
     const mergeStrategy =
       platformPrOptions.automergeStrategy === 'auto'
@@ -572,7 +576,11 @@ export async function updatePr({
   }
 
   if (body) {
-    objToUpdate.description = max4000Chars(sanitize(body));
+    await ensureComment({
+      number: prNo,
+      topic: 'dependency_update',
+      content: sanitize(body),
+    });
   }
 
   if (state === 'open') {
-- 
2.48.1
```

zivillian:
@rarkins we still interested in getting this merged - I've send you an email for the slack invite

PR #31645, opened 2024-09-26
feat(replacements): add replacements:gradle-wrapper-validation-action

mueller-ma:
## Changes

See Readme at https://github.com/gradle/wrapper-validation-action for more information.


rarkins (commented):
We normally don't want to do "replacements"
When the change is breaking. An alternative here is to met the user update to v3 and then migrate them

mueller-ma:
Now the user needs to update to version 3 before the replacement happens.

lib/data/replacements.json
```diff
@@ -669,6 +670,18 @@
       }
     ]
   },
+  "gradle-wrapper-validation-action": {
+    "description": "gradle/wrapper-validation-action got renamed to gradle/actions/wrapper-validation",
+    "packageRules": [
+      {
+        "matchCurrentVersion": "3",
```
HonkingGoose, line 677: Are you sure this rule is correct? It looks like the current rule will only match on precisely version `3`?

What if the user has a version within the `3` version range, like:

- `3.0.0`
- `3.1.0`
- `3.1.1`
- `3.9.9`

mueller-ma: I see, so the rule is incorrect. How should I fix it? Replace it by a regex that matches `3` and `3.x.y`? But what to use as `replacementVersion` then or can I remove it?

HonkingGoose: ## How to improve your preset

> I see, so the rule is incorrect. How should I fix it? Replace it by a regex that matches `3` and `3.x.y`? But what to use as `replacementVersion` then or can I remove it?

What I think you should do:

- The `matchCurrentVersion` on the _old_ package should be a regex that matches `3` and `3.x.y.`. 
- The `replacementVersion` should be the _first_ recommended version of the _new_ dependency. Pin the new version, by setting a specific new version.

## Source for changes

My changes are based on the rules listed in the [Renovate developer docs, Creating/editing Renovate presets](https://github.com/renovatebot/renovate/blob/main/docs/development/creating-editing-renovate-presets.md). Full quote of the section for replacement presets:

> #### Rules for replacement presets
> 
> 1. Replacement PRs should ideally propose a replacement only once the user is on a compatible version, by specifying a compatible `matchCurrentVersion` constraint
> 1. If no compatible replacement upgrade is possible, it's acceptable to propose an incompatible one (e.g. a major version upgrade)
> 1. Replacements should update the user to the first recommended version of the new dependency and not include any new changes - whether breaking or not - if they can be avoided, in short: pin the new version
> 
> If possible, replacement presets should give the user a replacement version that is _functionally identical_ to the _last version_ under the _old_ name.
> We only want a user's tests to fail because of _broken references_ to the old package name, and not because the maintainer(s) changed the _behavior_ of the package.

mueller-ma: Ok, I updated both values.

lib/data/replacements.json
```diff
@@ -669,6 +670,18 @@
       }
     ]
   },
+  "gradle-wrapper-validation-action": {
+    "description": "gradle/wrapper-validation-action got renamed to gradle/actions/wrapper-validation",
+    "packageRules": [
+      {
+        "matchCurrentVersion": "/3(\\.\\d+\\.\\d+)?/",
```
rarkins, line 677: Is the intent to allow either `3` or e.g. `3.2.10`? and not e.g. `3.2`?

HonkingGoose: The code started with:

```
        "matchCurrentVersion": "3",
```

I asked if that would match all possible versions in the `3` range:

- https://github.com/renovatebot/renovate/pull/31645#discussion_r1783061192

Our developers docs list the [rules for replacement presets](https://github.com/renovatebot/renovate/blob/main/docs/development/creating-editing-renovate-presets.md#rules-for-replacement-presets):

> #### Rules for replacement presets
> 
> 1. Replacement PRs should ideally propose a replacement only once the user is on a compatible version, by specifying a compatible `matchCurrentVersion` constraint
> 1. If no compatible replacement upgrade is possible, it's acceptable to propose an incompatible one (e.g. a major version upgrade)
> 1. Replacements should update the user to the first recommended version of the new dependency and not include any new changes - whether breaking or not - if they can be avoided, in short: pin the new version
> 
> If possible, replacement presets should give the user a replacement version that is _functionally identical_ to the _last version_ under the _old_ name.
> We only want a user's tests to fail because of _broken references_ to the old package name, and not because the maintainer(s) changed the _behavior_ of the package.

The key phrase in the docs is:

> Replacement PRs should ideally propose a replacement only once the user is on a compatible version, by specifying a compatible `matchCurrentVersion` constraint

So then the question is, is just being on any version 3 of the program good enough for a user to quickly merge the replacement PR? Or does the user need to be on the last version of the old package?

mueller-ma: > So then the question is, is just being on any version 3 of the program good enough for a user to quickly merge the replacement PR? Or does the user need to be on the last version of the old package?

@HonkingGoose I think it's good replace the dependency when on any version 3.

rarkins: OK but should it be `"matchCurrentVersion": "/^3\\.",` ? Note the ^ to enforce it at the start of the line, and the mandatory `.` (assumes that there will always be some precision to the version and not just `"3" itself, is that safe?)

mueller-ma: The version can be `v3` or `v3.5.0`:
- https://github.com/mueller-ma/Coffee/blob/main/.github/workflows/build.yml#L27
- https://github.com/mueller-ma/PrepaidBalance/blob/992ea54c1be81ba1e3bf2843f6a41cff43b5e153/.github/workflows/build.yml#L21

So the current regex should be fine, correct?

HonkingGoose: > OK but should it be `"matchCurrentVersion": "/^3\\.",` ? Note the ^ to enforce it at the start of the line, and the mandatory `.` (assumes that there will always be some precision to the version and not just `"3" itself, is that safe?)

```suggestion
        "matchCurrentVersion": "/^3\\.",
```

This suggestion is the copy/pasted code from `@rarkins`. I don't know if it's a good idea to apply the suggestion...

I'll let the maintainers decide what to do with the code. 😉

rarkins: The current one allows `3`, `30`, `3d`, etc. Is careful validation necessary to prevent these?

mueller-ma: The repo for the old version has been archived, so there shouldn't be any new versions like `30`: https://github.com/gradle/wrapper-validation-action

rarkins: But are we prevented from having _actually correct_ matching?

mueller-ma: I updated the regex. It should match `3` and `3.*` where `*` stands for anything.

rarkins (dismissed):

lib/data/replacements.json
```diff
@@ -670,6 +671,18 @@
       }
     ]
   },
+  "gradle-wrapper-validation-action": {
+    "description": "gradle/wrapper-validation-action got renamed to gradle/actions/wrapper-validation",
+    "packageRules": [
+      {
+        "matchCurrentVersion": "/^3(\\.)?/",
```
Churro, line 678:
```suggestion
        "matchCurrentVersion": "/^3/",
```
would achieve the same in a more concise form. With `(\\.)?` a dot may or may not follow after `3`. The outcome is the same though.

Churro (approved):

rarkins (approved):

PR #31699, opened 2024-09-30
fix(azure): remove azure prList caching, it is not being updated at all

PhilipAbed:
## Changes

removed azure prList caching, open for improvement later!

## Context

impact:
1. pr limits do not work if the prList is not being updated.
2. Branches info extended is giving wrong information causing prs to not be found.
maybe there is more that i didnt mention, but that is enough for us, because it affects the dashboard.

### DIFF
--- a/lib/modules/platform/azure/index.ts
+++ b/lib/modules/platform/azure/index.ts
@@ -251,27 +251,27 @@ export async function initRepo({
 
 export async function getPrList(): Promise<AzurePr[]> {
   logger.debug('getPrList()');
-  if (!config.prList) {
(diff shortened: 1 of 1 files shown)


rarkins (commented):
Won't this result in the PR list being fetched every single time?

viceice (changes requested):
bad idea. set it back to null when updating PRs and add the new PR on PR like other platforms do

rarkins:
Closing, as removing PR caching altogether is the wrong solution and all changes to this PR should be discarded. Please create a new PR implementing what viceice describes

Issue #29942, opened 2024-06-29
Centralize data to lib/data for easier contributing


rarkins:
### Describe the proposed change(s).

We have a lot of "crowdsourced" data in the repo, spread through different locations.

For example:
- Monorepos: https://github.com/renovatebot/renovate/blob/main/lib/config/presets/internal/monorepos.ts
- Replacements: https://github.com/renovatebot/renovate/blob/main/lib/config/presets/internal/replacements.ts
- Manual metadata for packages: https://github.com/renovatebot/renovate/blob/main/lib/modules/datasource/metadata-manual.ts

We should centralize this into `lib/data` so that it's easier for one-time or occasional contributors to find the right location to edit.

We should keep this folder as "raw" as possible and any wrapper code exists elsewhere. One topic per file. Files should be in .json format.

There should be a readme in the folder which clearly describes what each file/dataset is for.

RahulGautamSingh:
> The data should be in a form which is as easy for users to contribute as possible, instead of matching our internal formats. e.g. instead of forcing packageRules, we should keep them simpler if possible.

I do not understand this, can you share an example?

rarkins:
I was thinking "we shouldn't make people write large packageRules if they don't need to" but now I see that most of our data is simple enough, so I will drop this comment.

rarkins:
I added a requirement that the data be in .json format. @viceice is that still ok for our build process?

viceice:
should work, maybe use jsonc to allow comments?

rarkins:
I'm worried that `jsonc` isn't easily parseable by other ecosystems. For example we may internalize these in Mend systems using a Java backend

HonkingGoose:
When you're moving and renaming files, remember to update the "edit button links/overrides".

The published docs have a edit button which takes you to the file on GitHub. For some files the default path assumption of the edit button tool is wrong, so we have manual overrides. I remember we have overrides for things like the Renovate preset source files, and the `readme.md` files for the Renovate managers, and so on.

RahulGautamSingh:
I don't think the [metatda-manual for the packages](https://github.com/renovatebot/renovate/blob/main/lib/modules/datasource/metadata-manual.ts) have been documented yet.

How should we go about documenting this? Should they be included in the `Included Presets` section like the rest of the presets ?

RahulGautamSingh:
[...]

PR #31714, opened 2024-09-30
test(lib/data): `ReplacementsSchema`

RahulGautamSingh:
## Changes
Add JsonSchema and ZodSchema for the `replacements.json` file.

## Context
It will help users detect errors when adding new replacements instantly + he updated file will get validated against the zod schema during CI, leading to quicker PR merges & less reviews.
Ref: https://github.com/renovatebot/renovate/issues/29942#issuecomment-2241499559

### DIFF
--- a/lib/config/presets/internal/replacements.ts
+++ b/lib/config/presets/internal/replacements.ts
@@ -1,12 +1,12 @@
-import replacementGroups from '../../../data/replacements.json';
+import replacementGroupsJson from '../../../data/replacements.json';
 import type { Preset } from '../types';
 import type { PresetTemplate, Replacement } from './auto-generate-replacements';
 import { addPresets } from './auto-generate-replacements';
 
-const { $schema, ...tempPresets } = replacementGroups;
+const { $schema, ...replacementPresets } = replacementGroupsJson;
 
 /* eslint sort-keys: ["error", "asc", {"caseSensitive": false, "natural": true}] */
-export const presets: Record<string, Preset> = tempPresets;
+export const presets: Record<string, Preset> = replacementPresets;
 
 const muiReplacement: Replacement[] = [
   [['@material-ui/codemod'], '@mui/codemod'],
--- a/lib/config/presets/internal/replacements.ts
+++ b/lib/config/presets/internal/replacements.ts
@@ -3,8 +3,10 @@ import type { Preset } from '../types';
 import type { PresetTemplate, Replacement } from './auto-generate-replacements';
 import { addPresets } from './auto-generate-replacements';
 
+const { $schema, ...tempPresets } = replacementGroups;
+
 /* eslint sort-keys: ["error", "asc", {"caseSensitive": false, "natural": true}] */
-export const presets: Record<string, Preset> = replacementGroups;
+export const presets: Record<string, Preset> = tempPresets;
 
 const muiReplacement: Replacement[] = [
   [['@material-ui/codemod'], '@mui/codemod'],
--- a/tools/schemas/replacements-schema.json
+++ b/tools/schemas/replacements-schema.json
@@ -1,0 +1,61 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "all": {
+      "type": "object",
+      "properties": {
+        "description": { "type": "string" },
+        "extends": {
+          "type": "array",
+          "items": { "type": "string" }
+        },
+        "ignoreDeps": {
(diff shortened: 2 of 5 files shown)


secustor (commented):
Can we generate the schema automatically with [zod-to-json-schema](https://www.npmjs.com/package/zod-to-json-schema)?

tools/schemas/replacements-schema.json
```diff
@@ -0,0 +1,61 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
HonkingGoose, line 2:
```suggestion
  "$schema": "https://json-schema.org/draft-04/schema#",
```

Does the host support HTTPS? If yes, we should default to using HTTPS.

tools/schemas/schema.ts
```diff
@@ -11,4 +11,46 @@ const MonorepoSchema = z.object({
...
+    .refine(
+      (rules) =>
+        rules.some(
+          (rule) =>
+            rule.replacementName !== undefined ||
+            rule.replacementNameTemplate !== undefined,
+        ),
+      {
+        message:
+          'At least one package rule must have either replacementName or replacementNameTemplate',
```
HonkingGoose, line 38:
```suggestion
          'At least one package rule must use either the replacementName config option, or the replacementNameTemplate config option',
```

How about something like this?

lib/config/presets/internal/replacements.ts
```diff
@@ -3,8 +3,10 @@ import type { Preset } from '../types';
 import type { PresetTemplate, Replacement } from './auto-generate-replacements';
 import { addPresets } from './auto-generate-replacements';
 
+const { $schema, ...tempPresets } = replacementGroups;
+
 /* eslint sort-keys: ["error", "asc", {"caseSensitive": false, "natural": true}] */
-export const presets: Record<string, Preset> = replacementGroups;
+export const presets: Record<string, Preset> = tempPresets;
```
viceice, line 9:
```suggestion
export const presets: Record<string, Preset> = replacementGroups;
delete presets['$schema'];
```

what about this? so we don't need an extra variable.

RahulGautamSingh: TS won't like it .. had to do the rpevious shenanigans for TS only

RahulGautamSingh:
> Can we generate the schema automatically with [zod-to-json-schema](https://www.npmjs.com/package/zod-to-json-schema)?

Will it be useful? ie. can we use thus generated schemas for better UX?

secustor:
> > Can we generate the schema automatically with [zod-to-json-schema](https://www.npmjs.com/package/zod-to-json-schema)?
> 
> Will it be useful? ie. can we use thus generated schemas for better UX?

We would not to have to manage a JSON and zod schema at the same time.

RahulGautamSingh:
> We would not to have to manage a JSON and zod schema at the same time.

Got that.

For this we will need to add one step in the build process to save the schemas created using `zod-to-json` in json files, which will be referenced from the json files (`replacements.json`).

I think most users won't run the build process when adding a new replacement since its not necessary. So, maybe we are better to have a json schema. Also, these schemas won't change as much I guess so not a worry about maintaining.

tools/schemas/schema.ts
```diff
@@ -11,4 +11,46 @@ const MonorepoSchema = z.object({
...
+    .refine(
+      (rules) =>
+        rules.some(
+          (rule) =>
+            rule.replacementName !== undefined ||
+            rule.replacementNameTemplate !== undefined,
+        ),
+      {
+        message:
+          'At least one package rule must use either the replacementName config option, or the replacementNameTemplate config option',
```
secustor, line 38: Shouldn't all rules containing a `replacement` name.

secustor (commented):
I, have meant that the json schema file still gets commited. 

Though lets keep it like that. I guess there will be not many changes to the schema anyway.

tools/schemas/schema.ts
RahulGautamSingh: [There is one which does not.](https://github.com/renovatebot/renovate/blob/1562b303449c68e6529ad66a6924db5bf8a36e2a/lib/data/replacements.json#L183-L191) .

I am not sure it is needed here. But, decided to accomodate for it in in the Schema.

secustor (dismissed):

viceice (approved):

HonkingGoose (approved):
I'm happy with the text. I'll let the maintainers check the code and technical working.

PR #31726, opened 2024-10-01
fix(azure): azure pr list caching was not being updated ever

PhilipAbed:
## Changes

added caching in memory for azure PRs also for updatePR and createPR. it wasnt updated

## Context

we are not able to release Azure on our MendApp due to missing functionality.
pr limiting don't work without knowing what changed in the PRs.
pr number was not being set for the dashboard links/ the right sections in Mend App.


lib/modules/platform/azure/index.ts
```diff
@@ -64,7 +65,7 @@ interface Config {
   owner: string;
   repoId: string;
   project: string;
-  prList: AzurePr[];
+  prList?: AzurePr[];
```
viceice, line 68: revert and check, otherwise we get into runtime trouble

viceice:
failing CI checks 😉

lib/modules/platform/azure/index.ts
PhilipAbed: okey

PhilipAbed: how do i invalidate then @viceice ? 
```
  if (!config.prList || !pr || !idx) {
    // if pr doesn't exist in cache then invalidate cache so we get new list from api next time.
    config.prList = undefined;
    return;
  }
  ```

File: lib/modules/platform/azure/index.ts [hunk 2]
```diff
@@ -990,3 +994,44 @@ export async function deleteLabel(
...
+  if (is.undefined(idx)) {
+    config.prList.push(azurePr);
+  } else {
+    config.prList[idx] = azurePr;
+  }
+}
+
+function updateCachedPr(prNo: number, pr?: GitPullRequest): void {
+  const idx = getCachedPrIdx(prNo);
+  if (!config.prList || !pr || !idx) {
```
rarkins, line 1015: What if idx is `0`?

lib/modules/platform/azure/index.ts
PhilipAbed: i thought if i set it for empty array, and i check length later on.. it would work. but i dont think  that works since empty array means there are no PRs on the repo and its valid

lib/modules/platform/azure/index.ts
PhilipAbed: it should be is.undefined sry :D  did the same inside cachePr(

lib/modules/platform/azure/index.ts
PhilipAbed: ill do config.prList ??= [];

PhilipAbed: config.prList = undefined as any;
?? xD

lib/modules/platform/azure/util.ts
```diff
@@ -86,6 +86,25 @@ const stateMap = {
...
+  }
+  if (azurePr.description) {
+    cachedPr.bodyStruct = getPrBodyStruct(azurePr.description);
+  }
+  if (azurePr.targetRefName) {
+    cachedPr.targetBranch = getBranchNameWithoutRefsheadsPrefix(
+      azurePr.targetRefName,
+    );
+  }
+  return cachedPr;
```
rarkins, line 105: What's the point of returning a modified `cachedPr` if you are mutating the input object?

lib/modules/platform/azure/index.ts
```diff
@@ -990,3 +994,42 @@ export async function deleteLabel(
   const azureApiGit = await azureApi.gitApi();
   await azureApiGit.deletePullRequestLabels(config.repoId, prNumber, label);
 }
+
+function cachePr(pr?: GitPullRequest): void {
+  config.prList ??= [];
+  // istanbul ignore if
```
rarkins, line 1000:
```suggestion
```

Don't ignore code which is easy to test. But why not make this param mandatory?

lib/modules/platform/azure/index.ts
```diff
@@ -990,3 +994,42 @@ export async function deleteLabel(
...
+  if (!pr) {
+    return;
+  }
+
+  const azurePr = getRenovatePRFormat(pr);
```
rarkins, line 1005: I find the naming to be confusing. You have `pr` and `azurePr`, which are both non-descriptive. But worse: `azurePr` appears to be a Renovate internal format, so not actually an "azure" PR?

lib/modules/platform/azure/index.ts
```diff
@@ -990,3 +994,42 @@ export async function deleteLabel(
...
+  const idx = getCachedPrIdx(azurePr.number);
+  if (is.undefined(idx)) {
+    config.prList.push(azurePr);
+  } else {
+    config.prList[idx] = azurePr;
+  }
+}
+
+function updateCachedPr(prNo: number, pr?: GitPullRequest): void {
```
rarkins, line 1014:
```suggestion
function updateCachedPr(prNo: number, pr: GitPullRequest): void {
```

lib/modules/platform/azure/index.ts
```diff
@@ -990,3 +994,42 @@ export async function deleteLabel(
...
+  config.prList ??= [];
+  const idx = getCachedPrIdx(prNo);
+  if (!pr || is.undefined(idx)) {
+    // if pr doesn't exist in cache then invalidate cache so we get new list from api next time.
+    config.prList = undefined as any;
+    return;
+  }
+
+  const cachedPr = config.prList[idx];
+  config.prList[idx] = getRenovatePrFormatForCachedPr(cachedPr, pr);
```
rarkins, line 1024:
```suggestion
  getRenovatePrFormatForCachedPr(cachedPr, pr);
```

This function already mutates `cachedPr`

rarkins (changes requested):
I find this logic hard to follow. Variable names are too similar, plus `cachePr` and `updateCachedPr` can _both_ update a PR

lib/modules/platform/azure/index.ts
```diff
@@ -990,3 +994,42 @@ export async function deleteLabel(
...
+    config.prList[idx] = azurePr;
+  }
+}
+
+function updateCachedPr(prNo: number, pr?: GitPullRequest): void {
+  config.prList ??= [];
+  const idx = getCachedPrIdx(prNo);
+  if (!pr || is.undefined(idx)) {
+    // if pr doesn't exist in cache then invalidate cache so we get new list from api next time.
+    config.prList = undefined as any;
```
viceice, line 1019:
```suggestion
    delete config.prList;
```
much simpler, also works if not defined

lib/modules/platform/azure/index.ts
PhilipAbed: I agree. I was confused aswell

lib/modules/platform/azure/index.ts
PhilipAbed: Skill issue on my side xD

lib/modules/platform/azure/index.ts
PhilipAbed: copy paste from github. no idea, ill try

lib/modules/platform/azure/index.ts
PhilipAbed: the pr is supposed to be scmPr and the azurePr is supposed to be our renovate type

lib/modules/platform/azure/index.ts
PhilipAbed: hmm.. im not sending the list though. im sending 
const cachedPr = config.prList[idx];
if i take it out into a new const does it get the same address?
this doesn't happen in java. it just creates a new address for the cachedPr.

lib/modules/platform/azure/index.ts
PhilipAbed: ah, i can't do it. this conflicts with your previous review comment, prList is not optional.
`TS2790: The operand of a delete operator must be optional`

lib/modules/platform/azure/index.ts
PhilipAbed: ill call them scmAzurePr
and renovateAzurePr?

PhilipAbed:
im not sure why the build-docs isnt working. its not my fault i hope :D

lib/modules/platform/azure/index.spec.ts
```diff
@@ -1985,4 +1985,117 @@ describe('modules/platform/azure/index', () => {
...
+              pullRequestId: 1,
+              sourceRefName: 'refs/heads/branch-a',
+              targetRefName: 'refs/heads/branch-b',
+              title: 'branch a pr',
+              status: PullRequestStatus.Active,
+            })),
+            createPullRequestLabel: jest.fn(() => ({})),
+          }) as any,
+      );
+      // create again same pr
```
rarkins, line 2052: This is an invalid test. Azure does not permit the creation of the same PR like this, and even if branch name and title were the same, then the PR ID would not be the same

viceice (commented):
do we still need this? I think @rarkins has already fixed the caching bug in a separate PR. 🤔

PR #31729, opened 2024-10-01
feat(package-rules): add matchReleaseAge

mzglinski:
## Changes

Implement and add `matchReleaseAge` matcher to the packageRules, equivalent to `matchCurrentAge`.

## Context

Discussed here: https://github.com/renovatebot/renovate/discussions/31097


lib/config/options/index.ts
```diff
@@ -1272,6 +1272,18 @@ const options: RenovateOptions[] = [
     cli: false,
     env: false,
   },
+  {
+    name: 'matchReleaseAge',
+    description:
+      'Matches the age of the new / proposed package release derived from its release timestamp. Valid only within a `packageRules` object.',
```
HonkingGoose, line 1278:
```suggestion
      'Matches the age (in days) of a proposed package update. Renovate uses the package release timestamp to decide the package age. Valid only within a `packageRules` object.',
```

The current text is hard to understand. How about something like this, where we split the info into three sentences, and use simpler words?

Also, we should explain to what level of detail you can set the `matchReleaseAge` matcher. Which of these is valid input?

- `1 hour`
- `12 hours`
- `24 hours`
- `1 day`
- `1 day and 12 hours`
- `2 days`
- `3 days`
- `1 week`
- `1 month`

I'm guessing the new feature only allows time to be set based on day-intervals?

HonkingGoose (changes requested):

lib/config/options/index.ts
HonkingGoose: The [Renovate docs, scheduling syntax](https://docs.renovatebot.com/key-concepts/scheduling/#scheduling-syntax) has this warning:

> [!WARNING]
> Renovate does not support scheduled minutes or "at an exact time" granularity. Granularity must be at least one hour.
>
> Renovate uses the [@breejs/later](https://github.com/breejs/later) library to parse the text, so Renovate is limited to that library's syntax support. Read the parser documentation at [breejs.github.io/later/parsers.html#text](https://breejs.github.io/later/parsers.html#text) for more details. The _@breejs/later_ library also handles the concepts of "days", time_before", and "time_after".

Is the "Granularity must be at least one hour." thing relevant to your feature too?

mzglinski: I do not think so. The library you mentioned, only pops up in two files. 

https://github.com/renovatebot/renovate/blob/main/lib/workers/repository/update/branch/schedule.ts

https://github.com/renovatebot/renovate/blob/main/lib/config/migrations/custom/schedule-migration.ts

Both are related strictly to scheduling, not matching, and so is part of the documentation that you referenced.

HonkingGoose: Hmm... I don't know what we should do anymore. :smile:

I'll let a maintainer read the description, and let them decide how to proceed. :wink:

rarkins: I think its fine, although we need to make sure docs are consistent between matchCurrentAge and matchReleaseAge

lib/config/options/index.ts
```diff
@@ -1272,6 +1272,18 @@ const options: RenovateOptions[] = [
     cli: false,
     env: false,
   },
+  {
+    name: 'matchReleaseAge',
+    description:
+      'Matches the age of a proposed package update. Renovate uses the package release timestamp to decide the package age. Valid only within a `packageRules` object.',
```
HonkingGoose, line 1278: The `description` for `matchCurrentAge` says:

> Matches the current age of the package derived from its release timestamp. Valid only within a `packageRules` object.`

We can use a similar format for the new feature:

```suggestion
      'Matches the release age of the package derived from its release timestamp. Valid only within a `packageRules` object.',
```

lib/config/options/index.ts
```diff
@@ -1272,6 +1272,18 @@ const options: RenovateOptions[] = [
     cli: false,
     env: false,
   },
+  {
+    name: 'matchReleaseAge',
```
HonkingGoose, line 1276: We still need to:

- Explain how `matchReleaseAge` is different from `matchCurrentAge`.
- Explain how to decide if you need `matchReleaseAge` or `matchCurrentAge`.
- Give examples of cases where using th.e new feature is helpful.
- Give examples where the new feature is _worse_ than using `matchCurrentAge`.

Please add the explanations to the `docs/usage/configuration-options.md` file.

lib/config/options/index.ts
RahulGautamSingh:
```suggestion
      'Matches the age of a proposed package update derived from its release timestamp. Valid only within a `packageRules` object.',
```

This is better as it specifies that this is the age of the release which will be proposed in an update PR.
Only, saying `release age of the package` is incomplete in this case.

HonkingGoose: Let's go with `@RahulGautamSingh`'s version then. 😉

lib/config/options/index.ts
mzglinski: I think I've covered what you mentioned, but I'm open to any suggestions on how it can be improved.

rarkins:
Heads-up: we may want to implement this using `matchJsonata` instead, which was just merged yesterday

mzglinski:
@rarkins can confirm that I'm able to achieve my desired outcome using `matchJsonata`, with the following config

```json
"matchJsonata": [
      // 10 day old releases
      "$exists(releaseTimestamp) and $toMillis(releaseTimestamp) < $millis() - (10 * 24 * 60 * 60 * 1000)"
]
```
As far as I'm concerned this pull request does not offer any new functionality at this point. Please decide if we should finish it or scrap it.

rarkins:
That's great news! I'll close this, as I'd like to minimize the number of matchers now. FYI once https://github.com/renovatebot/renovate/pull/31818 is merged then you'll be able to have a simplified matchJsonata for this use case

PR #31767, opened 2024-10-03
feat(manager/mix): set versioning for git dependencies

1oglop1:
## Changes

This change should address [Bad Tag name for git dependencies in Elixir project](https://github.com/renovatebot/renovate/discussions/23535#top)

As far as I understand [mix](https://hexdocs.pm/mix/1.12/Mix.Tasks.Deps.html) does not support version ranges for git dependencies.
However for whatever reason when `"datasource": "git-tags", "versioning": "hex"` renovate runs the hex versioning and reaches this line
https://github.com/renovatebot/renovate/blob/8b2d1fa101476880e075f57f4c28773101cefd9e/lib/modules/versioning/hex/index.ts#L104

It produces an invalid tag from `v.0.1` -> `== v0.1` which fails to download the tag and `mix deps.update`

- [] I have updated the documentation, or
- [] No documentation update is required

### DIFF
--- a/lib/modules/manager/mix/__fixtures__/mix.exs
+++ b/lib/modules/manager/mix/__fixtures__/mix.exs
@@ -23,6 +23,7 @@ defmodule MyProject.MixProject do
       {:foo_bar, ">2.1.0 or <=3.0.0"},
       {:cowboy, github: "ninenines/cowboy", tag: "v0.4.1"},
       {:phoenix, git: "https://github.com/phoenixframework/phoenix.git", branch: "main"},
+      {:nonexistentpackage, git: "git@github:some/nonexistentpackage.git", tag: "v0.1.1"},
       {:ecto, github: "elixir-ecto/ecto", ref: "795036d997c7503b21fb64d6bf1a89b83c44f2b5"},
       {:secret, "~> 1.0", organization: "acme"},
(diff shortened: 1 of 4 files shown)


1oglop1:
Please let me know if this is an appropriate solution or if there is a better way to fix this.
I will check/update the documentation once I am sure this is the correct solution.

viceice (commented):
we should check why hex versioning is used for git-tags. versioning should ideally not set by manager. it should use the default from datasource

1oglop1:
@viceice can you point me to the code where this could/should be happening?
I noticed that some managers set versioning eg.

https://github.com/renovatebot/renovate/blob/8b2d1fa101476880e075f57f4c28773101cefd9e/lib/modules/manager/poetry/schema.ts#L133-L135

IIUC the datasource only sets the default versioning and

So in this case it I could set it as well but perhaps the same way as poetry does it. 

@rarkins or @viceice could you please reopen PR or do we need really need a new discussion for this exact same problem?

rarkins:
https://github.com/renovatebot/renovate/discussions/23535#discussioncomment-10831849

1oglop1:
@rarkins added the discussion with reproduction repo 
https://github.com/renovatebot/renovate/discussions/31778

PR #31771, opened 2024-10-03
fix(pypi): lookup simple api first

### LINKED ISSUE #26483: PyPI datasource fall back to simple API if JSON API fails

Similar to how we query rubygems.org, we should use the simple API to get the list / check for changes and then secondarily see if the rich API exists for querying metadata. 

Ie instead of falling back to simple as we do today, we should fall back to API instead, or enrich using it 

Algorithm:
- Query simple API first
- If simple fully fails, try the JSON API and return results
- If simple succeeds, check JSON API cache to see if we have metadata for all releases. If we do, no need to query the JSON API and we can return the combined results
- If the JSON API cache does not have metadata for all, query it, cache the new results, and return the merged data
- If JSON API fails, but simple worked, return simple results

samgiz:
## Changes

This rewrites the getReleases function of pypi datasources:
First, the simple (pypi) api is queried, then the pypijson api is retried to enrich the metadata.

Previously the simple api was only used if pypijson returned 404.

The unit tests stayed mostly the same, but the simple api routes needed to be mocked. Also added testcases to check that both simple and json pypi apis are used, dependency resolution succeeds if json api is unavailable, and dependencies from the two endpoints get merged correctly.

## Context

Building on top of https://github.com/renovatebot/renovate/pull/26505.
Closes https://github.com/renovatebot/renovate/issues/26483.

As pypijson is used additionally, https://github.com/renovatebot/renovate/discussions/26383 is resolved as well.

- [x] Both unit tests + ran on the pytorch pypi repository


lib/modules/datasource/pypi/index.ts
```diff
@@ -43,45 +46,54 @@ export class PypiDatasource extends Datasource {
...
 
-    // not all simple indexes use this identifier, but most do
-    if (hostUrl.endsWith('/simple/') || hostUrl.endsWith('/+simple/')) {
+    const simpleDependencies = await this.getResultsViaSimple(
+      normalizedLookupName,
+      simpleHostUrl,
+    ).catch((err) => {
+      if (err.statusCode !== 404) {
+        throw err;
+      }
```
samgiz, line 61: @rarkins You had a [comment](https://github.com/renovatebot/renovate/pull/26505/files#r1591806441) that this might be too aggressive, what should this be instead? It's similar to current behaviour.

samgiz: It won't throw if a 404 is returned. I can add more error codes / remove this altogether, but I thought that's [undesirable](https://github.com/renovatebot/renovate/pull/26386#issuecomment-1865213796).

In general I think it would be weird to have the json api be available without the simple api also being available, since that's what's used by `pip`.

rarkins: If so, it's a breaking change so should be in a major release, not a fix/patch release. Even if you assume that all registries _should_ support simple, it's a change in Renovate if we now _only_ support that.

In my experience registries in the wild are surprisingly varying. Any time we've assumed "they all surely support X" it's turned out to be wrong.

samgiz: I don't have a preference one way or the other. I've removed the check for now.

lib/modules/datasource/pypi/index.ts
```diff
@@ -43,45 +46,51 @@ export class PypiDatasource extends Datasource {
...
       logger.trace(
-        { packageName, hostUrl },
-        'Looking up pypi simple dependency',
+        { packageName, hostUrl: simpleHostUrl },
+        'Simple api not found. Looking up pypijson api as fallback.',
       );
-      dependency = await this.getSimpleDependency(
-        normalizedLookupName,
-        hostUrl,
+      return null;
```
rarkins, line 63: This appears to return early with `null` and not use pypijson as fallback

lib/modules/datasource/pypi/index.ts
```diff
@@ -43,45 +46,51 @@ export class PypiDatasource extends Datasource {
...
-      registryUrl!.replace('https://pypi.org/simple', 'https://pypi.org/pypi'),
-    );
+    const hostUrl = ensureTrailingSlash(registryUrl!);
+    const simpleHostUrl = hostUrl.replace(jsonPyPiUrl, simplePyPiUrl);
+    const pypiJsonHostUrl = hostUrl.replace(simplePyPiUrl, jsonPyPiUrl);
     const normalizedLookupName = normalizePythonDepName(packageName);
 
-    // not all simple indexes use this identifier, but most do
-    if (hostUrl.endsWith('/simple/') || hostUrl.endsWith('/+simple/')) {
+    const simpleDependencies = await this.getResultsViaSimple(
```
rarkins, line 55:
```suggestion
    const simpleResult = await this.getResultsViaSimple(
```

Isn't this a single package? If so use singular term and not "Dependencies"

lib/modules/datasource/pypi/index.ts
```diff
@@ -43,45 +46,51 @@ export class PypiDatasource extends Datasource {
...
-      dependency = await this.getSimpleDependency(
-        normalizedLookupName,
-        hostUrl,
+      return null;
+    });
+    logger.trace(
+      { packageName, hostUrl: pypiJsonHostUrl },
+      'Querying json api for metadata',
+    );
+    const pypiJsonDependencies = await this.getResultsViaPyPiJson(
```
rarkins, line 69:
```suggestion
    const jsonResult = await this.getResultsViaPyPiJson(
```

lib/modules/datasource/pypi/index.ts
```diff
@@ -43,45 +46,51 @@ export class PypiDatasource extends Datasource {
...
+      normalizedLookupName,
+      pypiJsonHostUrl,
+    ).catch((err) => {
+      if (simpleDependencies === null) {
+        throw err;
```
rarkins, line 74:
```suggestion
        // Both Simple and API lookups failed
        throw err;
```

lib/modules/datasource/pypi/index.ts
```diff
@@ -43,45 +46,51 @@ export class PypiDatasource extends Datasource {
...
-        // error contacting json-style api -- attempt to fallback to a simple-style api
-        logger.trace(
-          { packageName, hostUrl },
-          'Looking up pypi simple dependency via fallback',
-        );
-        dependency = await this.getSimpleDependency(
-          normalizedLookupName,
-          hostUrl,
-        );
+      return null;
```
rarkins, line 80: Won't this return `null` incorrectly? i.e. simple succeeded, API failed, you return `null`

lib/modules/datasource/pypi/index.ts
```diff
@@ -43,45 +46,51 @@ export class PypiDatasource extends Datasource {
...
+    });
+    if (pypiJsonDependencies === null || simpleDependencies === null) {
+      return pypiJsonDependencies ?? simpleDependencies
+    }
+    const dependency = pypiJsonDependencies
+    const added_versions = new Set<string>(dependency.releases.map(release => release.version))
+    for (const release of simpleDependencies.releases) {
+      if (!added_versions.has(release.version)) {
+        dependency.releases.push(release)
```
rarkins, line 89:
```suggestion
        logger.once.warn(`PyPI package ${normalizedName} on registry ${registryUrl} contains releases on the simple API which are missing from the JSON API`);
        dependency.releases.push(release)
```

lib/modules/datasource/pypi/index.ts
```diff
@@ -43,45 +46,51 @@ export class PypiDatasource extends Datasource {
...
-        dependency = await this.getSimpleDependency(
-          normalizedLookupName,
-          hostUrl,
-        );
+      return null;
+    });
+    if (pypiJsonDependencies === null || simpleDependencies === null) {
+      return pypiJsonDependencies ?? simpleDependencies
+    }
+    const dependency = pypiJsonDependencies
```
rarkins, line 85:
```suggestion
    // Default to JSON results because fields should be a superset of Simple results
    const dependency = pypiJsonDependencies
```

rarkins (changes requested):

lib/modules/datasource/pypi/index.spec.ts
```diff
@@ -716,31 +827,95 @@ describe('modules/datasource/pypi/index', () => {
...
+            `);
+        const config = {
+          registryUrls: ['https://custom.pypi.net/foo'],
+        };
+        const result = await getPkgReleases({
+          datasource,
+          ...config,
+          packageName: 'some-package',
+        });
+        console.log("here are the retrieved releases", result!.releases)
```
samgiz, line 893:
```suggestion
```

lib/modules/datasource/pypi/index.ts
samgiz: I think the return here ends up the value of `simpleDependencies`.

lib/modules/datasource/pypi/index.ts
samgiz: I think the return here ends up being the value of `pypiJsonDependencies`.

lib/modules/datasource/pypi/index.ts
```diff
@@ -43,45 +46,54 @@ export class PypiDatasource extends Datasource {
...
+      'Querying json api for metadata',
+    );
+    const pypiJsonDependencies = await this.getResultsViaPyPiJson(
+      normalizedLookupName,
+      pypiJsonHostUrl,
+    ).catch((err) => {
+      if (simpleDependencies === null) {
+        // Both Simple and API lookups failed
+        throw err;
+      }
```
samgiz, line 76: Should this throw an error or return null?

samgiz: I'll delete it for now, since I check for either of the values being null below. Let me know if this was expected behaviour.

rarkins:
I just realized also that this PR unfortunately doesn't follow the approach in the rubygems datasource (see https://github.com/renovatebot/renovate/tree/main/lib/modules/datasource/rubygems) as was suggested in #26483

In short:
- Use the simple index by default
- Any time a new version is found which wasn't seen before, query the JSON API (if present) for metadata, and heavily cache it

In other words, the goal was to avoid hitting the JSON API every time.

samgiz:
> https://github.com/renovatebot/renovate/discussions/26383 needs a reproduction so it can be tested

I think that's covered by this test, it tests a 403 return value:

```
    it.each([404, 403])(
      'process data from simple api with pypijson unavailable',
...
```

rarkins:
I meant a reproduction repo, not a unit test

samgiz:
> I meant a reproduction repo

Something like this? https://github.com/samgiz/renovate-26383-reproducer

rarkins:
Thanks for your patience and helping me understand it better. Based on your last suggestion I have implemented https://github.com/renovatebot/renovate/pull/32024

Please let me know if you think that solves the problem.

BTW I needed to make some config changes to the reproducer to get it to work: https://github.com/renovate-reproductions/26383/blob/master/renovate.json

samgiz:
That does solve my problem, thank you!

samgiz:
Closing now that https://github.com/renovatebot/renovate/pull/32024 has been merged.

PR #31815, opened 2024-10-07
fix(config): set `npmToken` in `npmrc` when not in encrypted

### LINKED ISSUE #31808: Set `npmToken` value in npmrc string even when it is not withtin `encrypted` object
We moved from keeing secrets inside the `encrypted` object to using the Mend Developer portal for users of the Github App. But, missed to make some necessary modififcations to the logic which would use the `npmToken` value from the config and replace `${NPM_TOKEN}` in the `npmrc` string with it.

For example:

```json
{
  "npmToken": "{{ secrets.NPM_TOKEN }}",
  "npmrc": "@fortawesome:registry=https://npm.fontawesome.com\n//npm.fontawesome.com/:_authToken=${NPM_TOKEN}"
}
```

Expected behaviour for the above config is that `npmToken` value is add to the `npmrc` string.

But, actually the `npmToken` value is not added to the `npmrc` string. And, this causes lookup errors down the line.

Example of such cases: https://github.com/renovatebot/renovate/discussions/31707  https://github.com/renovatebot/renovate/discussions/31253

Reason for this:
- `decryptConfig` fn handles the logic to replace `${NPM_TOKEN}` in the `npmrc` string with value of `npmToken`.
- `decryptConfig` only processes items in the `encrypted` object.
- Since `npmToken` does not necessarily have to be in the `encrypted` object anymore, this logic fails for Github App users.

Fix:
Reuse the logic in `decryptConfig` and place it such that the value from `npmToken` replaces the `${NPM_TOKEN}`. Most probably, this logic goes after the secrets have been applied to the config ie. at the end of `mergeRenovateConfig` fn

_Originally posted by @RahulGautamSingh in https://github.com/renovatebot/renovate/discussions/31707#discussioncomment-10814322_

RahulGautamSingh:
## Changes
Add `npmToken` value in the `npmrc` string even when the `npmToken` is not within `encrypted` object

## Context
Closes: #31808 

- [ ] Both unit tests + ran on a real repository (locally, with npmToken within encrypted and without it )


lib/workers/repository/init/merge.ts
```diff
@@ -278,3 +281,37 @@ export async function mergeRenovateConfig(
   }
   return returnConfig;
 }
+
+// needed when using portal secrets for npmToken
```
viceice, line 285: use tsdoc style

lib/workers/repository/init/merge.ts
```diff
@@ -278,3 +281,37 @@ export async function mergeRenovateConfig(
...
+export function setNpmTokenInNpmrc(config: RenovateConfig): void {
+  if (!is.string(config.npmToken)) {
+    return;
+  }
+
+  const token = config.npmToken;
+
+  if (!is.string(config.npmrc)) {
+    logger.debug('Adding npmrc to config');
+    config.npmrc = `//registry.npmjs.org/:_authToken=${token}\n`;
```
viceice, line 295: why scoping to registry here and not below?

lib/workers/repository/init/merge.ts
```diff
@@ -278,3 +281,37 @@ export async function mergeRenovateConfig(
...
+    delete config.npmToken;
+    return;
+  }
+
+  /* eslint-disable no-template-curly-in-string */
+  if (config.npmrc.includes('${NPM_TOKEN}')) {
```
viceice, line 301:
```suggestion
  if (config.npmrc.includes(`\${NPM_TOKEN}`)) {
```
should solve the eslint issue

viceice (commented):
I think the existing token handling should be removed then. 🤔

maybe it should be handled in npm manager only? because there's already some kind of code to handle 🤔

lib/workers/repository/init/merge.ts
RahulGautamSingh: Using the [existing logic](https://github.com/renovatebot/renovate/blob/1562b303449c68e6529ad66a6924db5bf8a36e2a/lib/config/decrypt.ts#L192) in `decryptConfig`

RahulGautamSingh:
> I think the existing token handling should be removed then.

Still needed for self-hosted users.

viceice:
> > I think the existing token handling should be removed then.
> 
> Still needed for self-hosted users.

can't we unify that code? otherwise we risk unexpected behavior in future

RahulGautamSingh:
> can't we unify that code? otherwise we risk unexpected behavior in future

Actually just removing the npmToken & npmrc logic from `decryptConfig` function should work as well. Updating...

lib/workers/repository/init/merge.ts
```diff
@@ -237,6 +239,7 @@ export async function mergeRenovateConfig(
     logger.trace({ config: resolvedConfig }, 'resolved config after migrating');
     resolvedConfig = migrationResult.migratedConfig;
   }
+  setNpmTokenInNpmrc(resolvedConfig);
```
viceice, line 242: why do we need to do it twice instead once at the end?

lib/workers/repository/init/merge.ts
viceice: @rarkins do you remember why we've two different logics? in this case we explicitly add the default registry and below we don't use it.

lib/workers/repository/init/merge.ts
RahulGautamSingh: npmrc is being set at these 2 instances. So its necessary to do so else we will save `npmrc` string in the `.npmrc` file which does not contain the token.

lib/workers/repository/init/merge.ts
rarkins: I'm not sure I understand the question. IIRC the logic is like this:
- If no npmrc, assume npmToken is for the public registry (create a .npmrc)
- If npmrc, and there's a placeholder for NPM_TOKEN already, then substitute it
- If npm, and no placeholder, append it (with no scope?)

viceice: Yes, that's the current logic, but i don't understand why we assume different scope for point 1 and 3.

So if no `npmrc`we scope to public registry but if there is a `npmrc` config and no `placeholder then we don't scope to public registry. That's confusing behavior.

RahulGautamSingh: >  i don't understand why we assume different scope for point 1 and 3.

My understanding, In 1 we do not have the npmrc string so we create the full string, whereas in 3 we have a npmrc string present, most possibly with the scope hence we append the auth token to the end.

viceice: but why scope on point 1 then?  it doesn't need to be scoped then too.

RahulGautamSingh: The code was added in 2018, maybe it was okay back then. Currently, any docs I can find it does mention to scope the token to the registry rather than adding it in a new line.

viceice: @rarkins I think the best is to always scope to default registry. If a user needs something different, then they should use the `{NPM_TOKEN}` replacement option instead.

It's probably a breaking behavior, doing in separate PR targeting next major?

rarkins: Sure, that's fine

viceice: so @RahulGautamSingh leave as is and fix in a follow up pr for next major, which hopefully comes this month

viceice (commented):
LGTM, but needs real repo testing

RahulGautamSingh:
> LGTM, but needs real repo testing

Done ✅

viceice (approved):

PR #31818, opened 2024-10-07
feat: currentVersionAgeInDays, newVersionAgeInDays

rarkins:
## Changes

Adds `currentVersionAgeInDays` dependencies if `currentReleaseTimestamp` is known. Also adds `newVersionAgeInDays` to an update if the update has a `releaseTimestamp`.

## Context

For use in `matchJsonata`


lib/workers/repository/process/lookup/index.ts
```diff
@@ -312,9 +312,16 @@ export async function lookupUpdates(
 
       if (is.nonEmptyString(currentVersionTimestamp)) {
         res.currentVersionTimestamp = currentVersionTimestamp;
+        // Calculate age in days
+        const currentVersionTimestampDate = new Date(currentVersionTimestamp);
+        const now = new Date();
+        const diffInMs = now.getTime() - currentVersionTimestampDate.getTime();
+        const currentVersionAge = Math.ceil(diffInMs / (1000 * 60 * 60 * 24));
```
viceice, line 319: use luxon, will simplify the code and is much more safe for parsing errors

lib/util/date.ts
```diff
@@ -1,12 +1,13 @@
 import { DateTime } from 'luxon';
 
 const ONE_MINUTE_MS = 60 * 1000;
-const ONE_DAY_MS = 24 * 60 * ONE_MINUTE_MS;
 
 export function getElapsedDays(timestamp: string): number {
```
rarkins, line 5: This function was previously unused

rarkins:
I think these fields could enable a rule like this:

```
{
  "description": "Don't update renovate more than weekly",
  "matchJsonata": "packageName = 'renovate/renovate' and newVersionAgeInDays - currentVersionAgeInDays < 7",
  "enabled": false
}
```

viceice (commented):
otherwise LGTM

lib/util/date.ts
```diff
@@ -1,12 +1,13 @@
...
-  return Math.floor(
-    (new Date().getTime() - new Date(timestamp).getTime()) / ONE_DAY_MS,
-  );
+  const currentVersionTimestampDate = DateTime.fromISO(timestamp);
```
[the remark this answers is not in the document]
rarkins: In theory they're all validated before they reach this point, so I think we can leave this as-is unless a problem is found

secustor (approved):

PR #31821, opened 2024-10-07
feat(packageRules): matchJsonata

### LINKED ISSUE #31820: packageRules.matchJsonata

Add new config option `matchJsonata` with parent `packageRules`, to sit alongside `matchPackageNames`, `matchCategories`, etc.

`matchJsonata` should accept a string.

When used, it should use JSONata (npm package `jsonata`) to query the string against the full config object which it is passed, and return `true` or `false`.

rarkins:
Fixes #31820

Add `matchJsonata` configuration option to `packageRules`.

* Add new option `matchJsonata` of type string with parent `packageRules` in `lib/config/options/index.ts`.
* Create `JsonataMatcher` class in `lib/util/package-rules/jsonata.ts` extending `Matcher` and implement `matches` method using JSONata.
* Import `JsonataMatcher` in `lib/util/package-rules/matchers.ts` and add it to the `matchers` array.
* Add tests for `JsonataMatcher` class in `lib/util/package-rules/jsonata.spec.ts`.
* Update types to include `matchJsonata` under `packageRules` in `lib/config/types.ts`.

### DIFF
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -1513,6 +1513,17 @@ const options: RenovateOptions[] = [
     cli: false,
     env: false,
   },
+  {
+    name: 'matchJsonata',
+    description:
+      'A JSONata expression to match against the full config object. Valid only within a `packageRules` object.',
+    type: 'string',
+    stage: 'package',
+    parents: ['packageRules'],
+    mergeable: true,
+    cli: false,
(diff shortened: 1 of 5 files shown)


rarkins:
Replaced by https://github.com/renovatebot/renovate/pull/31826

PR #31826, opened 2024-10-07
feat(packageRules): matchJsonata

### LINKED ISSUE #31820: packageRules.matchJsonata

rarkins:
## Changes

Add `matchJsonata` configuration option to `packageRules`.

## Context

Closes #31820


lib/config/validation.ts
```diff
@@ -442,6 +442,7 @@ export async function validateConfig(
               'matchCurrentAge',
               'matchRepositories',
               'matchNewValue',
+              'matchJsonata',
```
astellingwerf, line 445: Would it make sense to change this hardcoded array into something like?:
```
getOptions().filter(o => o.parents?.includes('packageRules') && o.name.startsWith('match')).map(o => o.name);
```

rarkins: Yeah, we could do that now

viceice: but out of scope for this pr

lib/util/package-rules/base.ts
```diff
@@ -11,5 +11,5 @@ export abstract class Matcher implements MatcherApi {
   abstract matches(
     inputConfig: PackageRuleInputConfig,
     packageRule: PackageRule,
-  ): boolean | null;
+  ): boolean | null | Promise<boolean | null>;
```
viceice, line 14: unrelated change

viceice (commented):
LGTM besides the unrelated changes

lib/util/package-rules/jsonata.ts
```diff
@@ -0,0 +1,30 @@
...
+export class JsonataMatcher extends Matcher {
+  override async matches(
+    inputConfig: PackageRuleInputConfig,
+    { matchJsonata }: PackageRule,
+  ): Promise<boolean | null> {
+    if (!matchJsonata) {
+      return null;
+    }
+
+    let expression;
```
viceice, line 15: add type

lib/config/validation.ts
astellingwerf: It would have prevented the validation error on the first attempt at this change set...

docs/usage/configuration-options.md
```diff
@@ -2856,6 +2856,21 @@ The following example matches any file in directories starting with `app/`:
...
+See [the JSONata docs](https://docs.jsonata.org/) for more details on JSONata syntax.
+
+Here are some example `matchJsonata` strings for inspiration:
+
+```
+$exists(deprecationMessage)
+$exists(vulnerabilityFixVersion)
+manager = 'dockerfile' and depType = 'final'
+```
+
```
secustor, line 2873: We should create an example on https://try.jsonata.org/ with the full so people have not to poke around in the dark.

lib/config/types.ts
```diff
@@ -381,6 +381,7 @@ export interface PackageRule
   matchRepositories?: string[];
   matchSourceUrls?: string[];
   matchUpdateTypes?: UpdateType[];
+  matchJsonata?: string;
```
secustor, line 384: IMO an array would be benefitial here by allowing user to break the logic in separat elements.

rarkins: Do you mean instead of using "and"s?

secustor: I mean OR chains

rarkins: Doesn't this then make it harder to debug eg in the playground?

Maybe we can start this way with single string and see if its really a problem

secustor: That way it is possible to test each element independently.  

But fine for me, it would be easy to migrate configs.

docs/usage/configuration-options.md
rarkins: Let's have it as a separate issue though

viceice (dismissed):

lib/config/validation.ts
```diff
@@ -835,6 +837,18 @@ export async function validateConfig(
         }
       }
     }
+
+    if (key === 'matchJsonata' && is.array(val)) {
+      for (const expression of val as string[]) {
```
secustor, line 842:
```suggestion
    if (key === 'matchJsonata' && is.array(val, is.string)) {
      for (const expression of val) {
```

secustor (approved):

PR #31832, opened 2024-10-07
feat(jsonata): validation, caching, better logging

rarkins:
## Changes

Adds caching to jsonata expression compiling. Also elevated logging to warn

## Context

Important for #31826 to use afterwards


lib/util/jsonata.ts
```diff
@@ -0,0 +1,18 @@
+import jsonata from 'jsonata';
+import * as memCache from './cache/memory';
+
+export function getExpression(input: string): jsonata.Expression | Error {
+  const cacheKey = `jsonata:${input}`;
```
secustor, line 5: I'm on my phone, do we hash the value afterwards? 

This transformations can become quite big.

lib/util/jsonata.ts
```diff
@@ -0,0 +1,18 @@
+import jsonata from 'jsonata';
+import * as memCache from './cache/memory';
+
+export function getExpression(input: string): jsonata.Expression | Error {
```
viceice, line 4: I think you need tests here in utils, otherwise CI will fail with coverage issues

lib/modules/datasource/custom/index.ts
```diff
@@ -46,13 +46,22 @@ export class CustomDatasource extends Datasource {
...
+          { errorMessage: expression.message },
+          `Error while compiling JSONata expression: ${transformTemplate}`,
+        );
+        return null;
+      }
+
       try {
-        const expression = jsonata(transformTemplate);
         data = await expression.evaluate(data);
       } catch (err) {
```
viceice, line 61: this should be checked in utils to not repeat the pattern

lib/util/jsonata.ts
rarkins: memCache is a simple `Record<string, any>`. We need a way to avoid repeated compilations if they're expensive, particularly for packageRules.matchJsonata which will be evaluated many times

lib/modules/datasource/custom/index.ts
rarkins: What's the benefit if it's one line?

viceice: never mind, it's now gone so this comment is obsolete

viceice (dismissed):

lib/modules/datasource/custom/index.ts
rarkins: Actually I brought it back

rarkins: I made your suggested change but didn't like it, prefer this way unless I'm missing something

lib/util/jsonata.ts
secustor: I'm not questioning that caching is useful here, but rather I'm concerned of the potentional key sizes

rarkins: Gotcha, I guess we could hash it

rarkins: Now hashed

lib/util/jsonata.ts
viceice, line 1: We should add tests to `./jsonata.spec.ts`, so utils still has 100% coverage when only those tests are run.

viceice (dismissed):

lib/util/jsonata.ts
rarkins: added. FYI lib/util doesn't have full coverage already though - it's missing some pieces

viceice (dismissed):

secustor (approved):

PR #31840, opened 2024-10-07
feat: export UV_EXTRA_INDEX_URL before doing 'uv lock'

torarvid:
This commit solves the challenge when a repo uses 'uv' to manage Python packages, and the repo also includes any dependencies in private repositories.

Current configuration options lets the repo provide secrets that allow Renovate to look for new versions of these dependencies, but it would fail on the step that updates the uv.lock lockfile, since the 'uv lock' subcommand has no knowledge of the Renovate configuration. The 'uv lock' command supports setting UV_EXTRA_INDEX_URL (a space-separated value of URLs to check before the default pypi index).

This commit then solves the challenge by using a combination of having Renovate configuration like so:

```json5
  "packageRules": [
    {
      matchDatasources: ["pypi"],
      registryUrls: ["https://europe-python.pkg.dev/project-123/python/simple/"],
    }
  ],
  hostRules: [
    {
      matchHost: "https://europe-python.pkg.dev",
      username: "_json_key_64",
      password: "{{ secrets.ARTIFACT_REGISTRY_JSON_KEY }}",
    },
  ]
```

The 'uv' processor will now, when it can match a `hostRule` with a `registryUrl` inject whatever `username` and `password` from the `hostRule` into the `registryUrl`. The URLs will be exported as `UV_EXTRA_INDEX_URL` before calling 'uv lock'.


torarvid:
Also created [a discussion](https://github.com/renovatebot/renovate/discussions/31841), in case that's a more appropriate place to discuss "the topic", as opposed to "the implementation".

I'll happily add unit tests to both prove the solution and fix the coverage, but would be great to know if people think this is a decent way to solve it first 😄

rarkins (changes requested):

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -191,3 +198,34 @@ function generateCMD(updatedDeps: Upgrade[]): string {
 
   return `${uvUpdateCMD} ${deps.map((dep) => `--upgrade-package ${quote(dep)}`).join(' ')}`;
 }
+
+function getMatchingHostRule(url: string | undefined): HostRule {
+  const scopedMatch = find({ hostType: PypiDatasource.id, url });
+  return is.nonEmptyObject(scopedMatch) ? scopedMatch : find({ url });
+}
+
+function getUvExtraIndexUrl(updatedDeps: Upgrade[]): NodeJS.ProcessEnv {
```
rarkins, line 207: It might be possible that there are index URLs _not_ in one of the Upgrades which uv needs, but we can wait for that edge case and add/fix it later

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -191,3 +198,34 @@ function generateCMD(updatedDeps: Upgrade[]): string {
...
+  const registryUrls = updatedDeps.map((dep) => dep.registryUrls).flat();
+  const extraIndexUrls: string[] = [];
+
+  for (const registryUrl of registryUrls) {
+    if (!registryUrl) {
+      continue;
+    }
+    const parsedUrl = new URL(registryUrl);
```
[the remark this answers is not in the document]
torarvid: Nice one, thanks! 👍🏼

lib/modules/manager/pep621/processors/uv.ts
torarvid: Rrrright, the idea I had in my head in the beginning was to just pry out all of the `registryUrl` things from `packageRules` for `pypi`, but not knowing the code very well, I didn't figure out how to do that 😅

lib/modules/manager/pep621/processors/uv.ts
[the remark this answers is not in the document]
torarvid: Should be fixed now

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -191,3 +199,34 @@ function generateCMD(updatedDeps: Upgrade[]): string {
 
   return `${uvUpdateCMD} ${deps.map((dep) => `--upgrade-package ${quote(dep)}`).join(' ')}`;
 }
+
+function getMatchingHostRule(url: string | undefined): HostRule {
+  const scopedMatch = find({ hostType: PypiDatasource.id, url });
+  return is.nonEmptyObject(scopedMatch) ? scopedMatch : find({ url });
```
viceice, line 205: are you sure to need this fallback? I thought the find is already doing that.

@rarkins do you know more?

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -191,3 +199,34 @@ function generateCMD(updatedDeps: Upgrade[]): string {
...
+}
+
+function getUvExtraIndexUrl(updatedDeps: Upgrade[]): NodeJS.ProcessEnv {
+  const registryUrls = updatedDeps.map((dep) => dep.registryUrls).flat();
```
viceice, line 209: use a `Set` to dedupe.

lib/modules/manager/pep621/processors/uv.ts
torarvid: I copy-pasted this from the poetry processor 😇 (which I guess doesn't prove anything... 😊)

lib/modules/manager/pep621/processors/uv.ts
torarvid: Good input. Will do 👍🏼

torarvid: Done 😄

lib/modules/manager/pep621/processors/uv.ts
rarkins: I think it's unnecessary because find() already does a fallback

rarkins (approved):

secustor (approved):

PR #31858, opened 2024-10-08
feat(manager/maven): provide warning log if pom contains windows line ending

sebtiem:
## Changes

As discussed in [19090](https://github.com/renovatebot/renovate/discussions/19090), I added a warning log message if windows line endings are present when extracting the contents of a maven pom.xml.

## Context

See discussion [19090](https://github.com/renovatebot/renovate/discussions/19090).

- [] Newly added/modified unit tests, or


lib/modules/manager/maven/__fixtures__/simple.pom.crlf.xml
rarkins, line 1: Reduce this to the minimum possible and inline it in the test file instead of as a fixture

lib/modules/manager/maven/__fixtures__/.gitattributes
rarkins, line 1: Is this file necessary?

rarkins (changes requested):

lib/modules/manager/maven/__fixtures__/.gitattributes
sebtiem: We need to have the information to tell github not to switch to lf-endings. There already is a .gitattributes file at root  folder but to me, it feels more suitable to put this setting here as it is only for this specific file.

rarkins: Do your tests fail if this file is removed?

rarkins: Can this test file be inlined instead, e.g. using `\r\n` as part of the string?

sebtiem: Sounds good.

rarkins (approved):

PR #31859, opened 2024-10-08
feat(versioning): isBreaking

### LINKED ISSUE #24513: Add first class support for breaking/non-breaking concept

Because Semver is so widespread, the terms "breaking" and "major" are sometimes considered synonymous, when that's not the case. Most relevant examples include:
- An update from v0.1.0 to v0.2.0 may include breaking changes, even though it's a minor release
- Python language minor releases (e.g. 3.10 to 3.11) are considered like "breaking" changes to some, or at least akin to a major update in some other languages

We've had requests in the past for Renovate to do things like "Treat pre-1.0.0 releases as major releases" but that would be a mistake. We should not misuse the term "major" which is clearly defined. Instead, we should introduce the parallel concept of _breaking_, and let people define grouping and rules based on that.

As a simple example, in semver the following types of updates would all be potentiall breaking ones:
- Where the from version is <1.0.0
- Where the from version is a pre-release
- Any other major update

I suspect that the best way to do this is to include a new, optional function in `versioning` called `isBreakingUpdate(fromVersion, toVersion)` which returns true/false. If this function is absent then we assume major updates are breaking, otherwise the update is non-breaking. Otherwise the versioning scheme can use whatever logic it wishes in order to return a response.

Then we can supplement this with a way to match in packageRules

rarkins:
## Changes

Adds new concept `isBreaking`

TODO:
- [x] docs, including matchJsonata example
- [x] branch.isBreaking logic
- [x] Cargo versioning
- [x] Decide on unstable
- [ ] Handle python minor=breaking out of the box somehow?

## Context

Should be ready to be used with `matchJsonata` as an alternative to `matchUpdateTypes=major`.

- Closes #24513


lib/workers/repository/process/lookup/generate.ts
```diff
@@ -80,6 +80,13 @@ export async function generateUpdate(
   update.updateType =
     update.updateType ??
     getUpdateType(config, versioning, currentVersion, newVersion);
+  if (versioning.isBreaking) {
+    // This versioning scheme has breaking awareness
+    update.isBreaking = versioning.isBreaking(currentVersion, newVersion);
+  } else {
+    // This versioning scheme does not have breaking awareness - assume only major updates are breaking
```
rarkins, line 87:
```suggestion
    // This versioning scheme does not have breaking awareness - assume only major updates are breaking
    // Updates from or to unstable releases should be treated as breaking too, but we should not add that as default behavior until we stop treating non-LTS 
```

lib/workers/repository/process/lookup/generate.ts
```diff
@@ -80,6 +80,14 @@ export async function generateUpdate(
   update.updateType =
     update.updateType ??
     getUpdateType(config, versioningApi, currentVersion, newVersion);
+  if (versioningApi.isBreaking) {
+    // This versioning scheme has breaking awareness
+    update.isBreaking = versioningApi.isBreaking(currentVersion, newVersion);
+  } else {
+    // This versioning scheme does not have breaking awareness - assume only major updates are breaking
+    // Updates from or to unstable releases should be treated as breaking too, but we should not add that as default behavior until we stop treating non-LTS
```
HonkingGoose, line 88:
```suggestion
    // Updates from, or to, unstable releases should be treated as breaking too.
    // But we should not add that as default behavior until we stop treating non-LTS versions ...INSERT EXPLANATION HERE...
```

Let's split this into two sentences, and add some commas to help read the first sentence.

I think the last sentence is missing information about what needs to change to `non-LTS`, before you can "treat updates from, or to unstable releases as breaking". So I've added some placeholder text for you to fill out. :wink:

HonkingGoose (changes requested):

lib/workers/repository/process/lookup/generate.ts
```diff
@@ -82,6 +82,15 @@ export async function generateUpdate(
   update.updateType =
     update.updateType ??
     getUpdateType(config, versioningApi, currentVersion, newVersion);
+  if (versioningApi.isBreaking) {
+    // This versioning scheme has breaking awareness
+    update.isBreaking = versioningApi.isBreaking(currentVersion, newVersion);
+  } else {
+    // This versioning scheme does not have breaking awareness - assume only major updates are breaking
+    // Updates from, or to, unstable releases should be treated as breaking too.
+    // But we should not add that as default behavior until we stop treating non-LTS versions ...INSERT EXPLANATION HERE...
```
HonkingGoose, line 91: Placeholder for the maintainer to fill out. 🙂

secustor (dismissed):
LGTM, but we should add docs what the functions are for on versioning. Just like we have for managers.

lib/modules/versioning/cargo/index.ts
```diff
@@ -1,9 +1,16 @@
+import semver from 'semver';
+import stable from 'semver-stable';
+
```
viceice, line 3:
```suggestion
```

lib/modules/versioning/cargo/index.ts
```diff
@@ -1,9 +1,16 @@
...
 import { logger } from '../../../logger';
 import type { RangeStrategy } from '../../../types/versioning';
 import { regEx } from '../../../util/regex';
 import { api as npm } from '../npm';
 import type { NewValueConfig, VersioningApi } from '../types';
 
+const { is: isStable } = stable;
+
+const { major: getMajor, minor: getMinor } = semver;
```
viceice, line 12: why not doing this as part of the import?

lib/modules/versioning/cargo/index.ts
```diff
@@ -149,9 +156,28 @@ function getNewValue({
   return newCargo;
 }
 
+function isBreaking(current: string, version: string): boolean {
+  // The change may be breaking if either version is unstable
+  if (!isStable(version) || !isStable(current)) {
```
viceice, line 161: can this throw somehow? do we validate single version somewhere before?

lib/modules/versioning/semver-coerced/index.ts
```diff
@@ -133,6 +134,16 @@ function getNewValue({
   return newVersion;
 }
 
+function isBreaking(version: string, current: string): boolean {
+  const coercedVersion = semver.coerce(version)?.toString();
+  const coercedCurrent = semver.coerce(current)?.toString();
```
viceice, line 139:
```suggestion
  const coercedVersion = semver.coerce(version);
  const coercedCurrent = semver.coerce(current);
```
why the `toString`?

lib/workers/repository/process/lookup/generate.ts
viceice: ☝️

lib/modules/versioning/cargo/index.ts
rarkins: Yes, we only call this if we have a valid update, which implies it was valid versions

lib/modules/versioning/cargo/index.ts
rarkins: Done

lib/workers/repository/process/lookup/generate.ts
rarkins, line 91:
```suggestion
    // But we should not add that as default behavior until we stop treating non-LTS versions as unstable first
```

lib/modules/versioning/semver-coerced/index.ts
rarkins: Because it returns type `Semver | null`

viceice (dismissed):

secustor (dismissed):

docs/usage/modules/versioning/index.md
```diff
@@ -57,6 +68,34 @@ The configuration below overrides Renovate's default `docker` versioning for the
 }
 ```
 
+## Breaking Changes
+
+In most ecosystems, especially SemVer, major upgrades are synonymous with breaking changes.
```
HonkingGoose, line 73:
```suggestion
In most ecosystems, especially SemVer, major upgrades are treated as breaking changes.
```

docs/usage/modules/versioning/index.md
```diff
@@ -57,6 +68,34 @@ The configuration below overrides Renovate's default `docker` versioning for the
...
+However, there are other cases too:
```
HonkingGoose, line 74:
```suggestion
But other updates may have breaking changes too, for example:
```

docs/usage/modules/versioning/index.md
```diff
@@ -57,6 +68,34 @@ The configuration below overrides Renovate's default `docker` versioning for the
...
+The definitions of major and minor should not be redefined and broken in order to shoehorn in the related concept of "breaking change".
+Instead, Renovate has the concept of `isBreaking`, which can be decided independently of `updateType`.
+
+Here's an example of grouping all non-breaking updates together:
+
+```json
+{
+  "packageRules": [
+    {
+      "description": "Group together non-breaking updates",
```
HonkingGoose, line 90:
```suggestion
      "description": "Group non-breaking updates",
```

HonkingGoose (changes requested):
Some docs improvements. 😉

viceice (dismissed):
LGTM, but would love to include the docs improvements

viceice (approved):

PR #31921, opened 2024-10-11
feat(nix): rework to work with all flake inputs

SuperSandro2000:
## Changes

Before nix flakes where only updated if a github:nixos/nixpkgs input was used. Now all flake inputs are treated like a dependency and updated.


MarcelCoding (approved):

lib/modules/manager/nix/types.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export interface NixInput {
+  readonly inputs?: Record<string, string>;
+  readonly locked?: LockedInput;
+  readonly original?: OriginalInput;
+}
+
+export interface LockedInput {
+  readonly lastModified: number;
```
SuperSandro2000, line 14:
```suggestion
  readonly host?: string;
  readonly lastModified: number;
```

lib/modules/manager/nix/types.ts
```diff
@@ -0,0 +1,40 @@
+export interface NixFlakeLock {
```
viceice, line 1: use zod schema

lib/modules/manager/nix/__fixtures__/flake.4.lock
```diff
@@ -0,0 +1,44 @@
...
+        "lastModified": 1718457448,
+        "narHash": "sha256-FSoxTcRZMGHNJh8dNtKOkcUtjhmhU6yQXcZZfUPLhQM=",
+        "ref": "refs/heads/master",
+        "rev": "a0f54334df36770b335c051e540ba40afcbf8378",
+        "revCount": 844,
+        "type": "git",
+        "url": "https://github.com/NixOS/patchelf.git"
+      },
+      "original": {
+        "type": "git",
```
Mic92, line 32: This should than also work with submodules and shallow clones, right?

lib/modules/manager/nix/__fixtures__/flake.5.lock
```diff
@@ -0,0 +1,27 @@
...
+        "narHash": "sha256-Ho/sAhEUeSug52JALgjrKVUPCBe8+PovbJj/lniKxp8=",
+        "owner": "~gpanders",
+        "repo": "ijq",
+        "rev": "88f0d9ae98942bf49cba302c42b2a0f6e05f9b58",
+        "type": "sourcehut"
+      },
+      "original": {
+        "owner": "~gpanders",
+        "repo": "ijq",
+        "type": "sourcehut"
```
Mic92, line 16: Could we in theory also support "type": "tarball"?

Mic92: Would be acceptable if this is too hard to do in this PR.

lib/modules/manager/nix/types.ts
SuperSandro2000: that's complicated to parse a json file 💀

SuperSandro2000: Got it working with the help of a friend 😅

SuperSandro2000:
I am not sure if this is really breaking. Before updating flake locks just didn't really work and wasn't doing what people would expect.

SuperSandro2000:
> ``❌ Error: More files have been added to '__fixtures__' directories than deleted.``

I don't really have an idea how to test the lockfiles other than using fixtures.

lib/modules/manager/nix/__fixtures__/flake.5.lock
SuperSandro2000: see https://github.com/NixOS/nix/blob/master/doc/manual/src/protocols/tarball-fetcher.md#gitea-and-forgejo-support 

For the format. We can probably easily regex that and if it doesn't match just skip it.

rarkins:
> > `❌ Error: More files have been added to '__fixtures__' directories than deleted.`
> 
> I don't really have an idea how to test the lockfiles other than using fixtures.

This is a valid use when lockfiles are long, we can ignore this test and merge anyway.

lib/modules/manager/nix/__fixtures__/flake.5.lock
SuperSandro2000: Just pushed a commit with that as well

SuperSandro2000:
I sadly didn't find a way to check the coverage locally without literally running all tests but should be good now.

What's next? I am pretty sure my typescript code is mediocre at best 😂

viceice (commented):
can we reduce and inline fixtures to spec files?

lib/modules/manager/nix/index.ts
```diff
@@ -6,8 +6,8 @@ export { updateArtifacts } from './artifacts';
 export const supportsLockFileMaintenance = true;
 
 export const defaultConfig = {
-  fileMatch: ['(^|/)flake\\.nix$'],
-  commitMessageTopic: 'nixpkgs',
+  fileMatch: ['(^|/)flake\\.lock'],
+  commitMessageTopic: 'nix',
```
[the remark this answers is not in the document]
MarcelCoding: nixpkgs is one of many flakes - so it would only be correct if nixpkgs gets updated

lib/modules/manager/nix/index.ts
```diff
@@ -6,8 +6,8 @@ export { updateArtifacts } from './artifacts';
 export const supportsLockFileMaintenance = true;
 
 export const defaultConfig = {
-  fileMatch: ['(^|/)flake\\.nix$'],
-  commitMessageTopic: 'nixpkgs',
+  fileMatch: ['(^|/)flake\\.lock'],
```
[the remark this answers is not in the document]
SuperSandro2000: Not everyone is commit their flake.lock and there is no point in updating the flake.lock if it isn't commited.

lib/modules/manager/nix/index.ts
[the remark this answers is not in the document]
SuperSandro2000: Nix is the package manager and there are other flakes like eg home-manager where using nixpkgs in the commit message would be wrong.

SuperSandro2000:
> can we reduce and inline fixtures to spec files?

I generated them all with a single flake input. I could remove the peer dependencies but then they would be invalid.

lib/modules/manager/nix/index.ts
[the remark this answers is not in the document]
viceice: this won't decided by file filter on renovate. this should be decided in update artifacts. we do the same for others managers.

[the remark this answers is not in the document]
SuperSandro2000: okay, I reverted it to the old value

SuperSandro2000:
FYI an example PR created with this branch https://github.com/NuschtOS/search.nuschtos.de/pull/20

Is there existing functionality in renovate to add github compare links? I only know that renovate can attach the changelog from releases which often contain such a link and I would lik to "emulate" them to make reviewing those updates super convenient.

Mfazer (approved):

lib/modules/manager/nix/artifacts.spec.ts
```diff
@@ -36,17 +36,14 @@ process.env.CONTAINERBASE = 'true';
 const config: UpdateArtifactsConfig = {};
 const lockMaintenanceConfig = { ...config, isLockFileMaintenance: true };
 const updateInputCmd = `nix \
-    --extra-experimental-features nix-command \
-    --extra-experimental-features flakes \
+--extra-experimental-features 'nix-command flakes' \
```
viceice, line 39: please do this argument refactor in a separate PR to be merged first, so we've less changes here

lib/modules/manager/nix/schema.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const LockedInput = z.object({
+  host: z.string().optional(),
+  lastModified: z.number(),
+  narHash: z.string(),
+  owner: z.string().optional(),
+  repo: z.string().optional(),
+  ref: z.string().optional(),
+  rev: z.string(),
+  revCount: z.number().optional(),
```
viceice, line 21: remove properties we don't use

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,129 @@
...
-  if (match?.groups) {
-    const { ref } = match.groups;
-    deps.push({
-      depName: 'nixpkgs',
-      currentValue: ref,
-      datasource: GitRefsDatasource.id,
-      packageName: 'https://github.com/NixOS/nixpkgs',
-      versioning: nixpkgsVersioning,
-    });
+  const flakeLockParsed = NixFlakeLock.safeParse(content);
```
viceice, line 23: you need to search for lock file first, as this is the `flake.nix` content.

we've a getSilbling fs util function

lib/modules/manager/nix/artifacts.spec.ts
```diff
@@ -66,7 +63,7 @@ describe('modules/manager/nix/artifacts', () => {
   it('returns if no flake.lock found', async () => {
     const execSnapshots = mockExecAll();
     const res = await updateArtifacts({
-      packageFileName: 'flake.nix',
+      packageFileName: 'flake.lock',
```
viceice, line 66: wrong, you pass the `flake.nix` and corresponding content instead

viceice (changes requested):

lib/modules/manager/nix/artifacts.spec.ts
SuperSandro2000: #32289

lib/modules/manager/nix/artifacts.spec.ts
SuperSandro2000: should be resolved

lib/modules/manager/nix/schema.ts
SuperSandro2000: removed the unused ones

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,132 @@
...
 import type { PackageDependency, PackageFileContent } from '../types';
+import { NixFlakeLock } from './schema';
 
-const nixpkgsRegex = regEx(/"github:nixos\/nixpkgs(\/(?<ref>[a-z0-9-.]+))?"/i);
+// TODO: add support to update nixpkgs branches in flakes.nix using nixpkgsVersioning
+
+// as documented upstream
+// https://github.com/NixOS/nix/blob/master/doc/manual/source/protocols/tarball-fetcher.md#gitea-and-forgejo-support
+const lockableHTTPTarballProtocol = regEx(
+  '^https://(.+)/(.+)/(.+)/archive/(.+).tar.gz$',
```
viceice, line 13:
```suggestion
  '^https://([^/]+)/([^/]+)/([^/]+)/archive/(.+).tar.gz$',
```
you don't want to match slashes in those groups.

are you using them later? then please use named capture groups.

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,132 @@
...
+  const flakeLock = flakeLockParsed.data;
+
+  for (const depName of Object.keys(flakeLock.nodes)) {
+    // the root input is a magic string for the entrypoint and only references other flake inputs
+    if (depName === 'root') {
+      continue;
+    }
+
+    // skip all locked nodes which are not in the flake.nix and cannot be updated
+    // istanbul ignore next: a valid flake.lock file will never run into this
```
viceice, line 44:
```suggestion
    // istanbul ignore if: a valid flake.lock file will never run into this
```

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,132 @@
...
+    if (!(depName in (flakeLock.nodes['root'].inputs ?? []))) {
+      continue;
```
viceice, line 46: log a warning for invalid lockfile

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,132 @@
...
+
+    const flakeInput = flakeLock.nodes[depName];
+    const flakeLocked = flakeInput.locked;
+    const flakeOriginal = flakeInput.original;
+
+    // istanbul ignore if: if we are not in a root node then original and locked always exist which cannot be easily expressed in the type
+    if (flakeLocked === undefined || flakeOriginal === undefined) {
+      logger.debug(
+        { packageLockFile },
+        `Found empty flake input '${JSON.stringify(flakeInput)}', skipping`,
```
viceice, line 57: @rarkins I think that json should be logged as metadata?

lib/modules/manager/nix/schema.ts
```diff
@@ -0,0 +1,41 @@
+import { z } from 'zod';
+import { Json } from '../../../util/schema-utils';
+
+export const InputType = z.enum([
```
viceice, line 4:
```suggestion
const InputType = z.enum([
```
don't export if not used

lib/modules/manager/nix/schema.ts
```diff
@@ -0,0 +1,41 @@
...
+  'git',
+  'github',
+  'gitlab',
+  'indirect',
+  'sourcehut',
+  'tarball',
+]);
+
+export const LockedInput = z.object({
```
viceice, line 13:
```suggestion
const LockedInput = z.object({
```

lib/modules/manager/nix/schema.ts
```diff
@@ -0,0 +1,41 @@
...
+  ref: z.string().optional(),
+  rev: z.string(),
+  type: InputType,
+});
+
+export const OriginalInput = z.object({
```
viceice, line 19:
```suggestion
const OriginalInput = z.object({
```

lib/modules/manager/nix/extract.ts
SuperSandro2000: How would I do that? I am not familiar at all with the internal structure of renovate 😅

lib/modules/manager/nix/extract.ts
SuperSandro2000: that seems to not work and reduce the coverage

https://app.codecov.io/gh/renovatebot/renovate/pull/31921?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=renovatebot#52e1fb44112718801520ba62dc1c7a7a-R45

blitz:
Thanks @SuperSandro2000. This is a super valuable feature! :pray:

lib/modules/manager/nix/extract.ts
rarkins: I thought it should produce the same result

SuperSandro2000: I think I am ignoring the `??` here instead of the if.

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,138 @@
...
+  for (const depName of Object.keys(flakeLock.nodes)) {
+    // the root input is a magic string for the entrypoint and only references other flake inputs
+    if (depName === 'root') {
+      continue;
+    }
+
+    // skip all locked nodes which are not in the flake.nix and cannot be updated
+    // istanbul ignore next: a valid flake.lock file will never run into this
+    if (!(depName in (flakeLock.nodes['root'].inputs ?? []))) {
+      logger.debug(
```
SuperSandro2000, line 46: so we could do this, not sure if that makes any difference 

```suggestion
    if (!(depName in (flakeLock.nodes['root'].inputs ?? /* istanbul ignore next */ '' []))) {
      logger.debug(
```

SuperSandro2000:
@rarkins @viceice How can I move this forward? I am not really familiar with renovate 👀

viceice (commented):
see comment

mfenniak:
It would be wonderful to see this merged -- is there anything that an outsider like myself can volunteer to help with to get it to thee finish line? :-)

rarkins:
@mfenniak Yes - please create a public (github.com) repo and demonstrate this new code working as intended (e.g. creating Pull Requests which don't happen on the current stable/released Renovate)

SuperSandro2000:
Nothing easier than that :) https://github.com/NuschtOS/nixos-modules/issues/130 https://github.com/NuschtOS/nixos-modules/pull/166 https://github.com/NuschtOS/search/issues/97
https://github.com/NuschtOS/search/pull/114

rarkins:
Happy to merge this once @viceice approves

lib/modules/manager/nix/extract.ts
mfenniak: Although this comment is resolved, it doesn't seem to be implemented correctly -- the file path of the sibling is stored in the `packageLockFile` variable, but the contents that are parsed here are from `flake.nix` and not `flake.lock`.

The unit tests covering this in `extract.spec.ts` don't match the actual implementation as the packageFile is `flake.nix` and not `flake.lock`, which those unit tests are spec'd for.

I think that when https://github.com/renovatebot/renovate/pull/31921/commits/4bcf310a8d30e3f2eb7b8d12b886e6c6308249ba was reverted, all of this broke.

mfenniak:
I've tried to pull the branch here and test it out end-to-end, and it seems to be broken in the current commit -- https://github.com/renovatebot/renovate/pull/31921/files#r1871652963.  This seems to be a pretty small problem easily fixed, but then the tests need updating.

I've patched this problem out in my local env...
```
  const packageLockFile = getSiblingFileName(packageFile, 'flake.lock');
  const lockContents = await readLocalFile(packageLockFile);
  // ..
  const flakeLockParsed = NixFlakeLock.safeParse(lockContents);
```
but now I'm experiencing a parse failure in my flake.lock:
```
DEBUG: invalid flake.lock file (repository=testtrim/testtrim)
       "packageLockFile": "flake.lock",
       "error": {
         "message": "Schema error",
         "stack": "ZodError: Schema error\n    at Object.get error [as error] (/run/host/home/mfenniak/Dev/renovate/node_modules/zod/lib/types.js:55:31)\n    at Object.extractPackageFile (/run/host/home/mfenniak/Dev/renovate/lib/modules/manager/nix/extract.ts:33:49)\n    at async getManagerPackageFiles (/run/host/home/mfenniak/Dev/renovate/lib/workers/repository/extract/manager-files.ts:58:19)\n    (snip)",
         "issues": "Expected string, received object"
       }
```
I'll poke this a bit more and see if I can figure out what is going on.  If I can resolve these issues I'll share a branch that @SuperSandro2000 can pull into this PR today.

mfenniak:
The `Schema error` was just due to passing a buffer to `safeParse`, rather than a string.  These tweaks to the branch look to be functioning in a limited test against one repo: https://github.com/SuperSandro2000/renovate/pull/1

mfenniak:
I've rebuilt and redeployed my renovate install on the latest commit (70912bf17f9106b2a23cff99eefe508faa0aaa12) and it's looking good to me.  I don't have a GitHub repo configured with this right now, but I do have a public Forgejo repo:
- Onboarding PR that includes proposed flake updates: https://forgejo.kainnef.com/mfenniak/pixelperfectpi/pulls/3
- Successful PR with update request created: https://forgejo.kainnef.com/mfenniak/pixelperfectpi/pulls/5

At a bare minimum this PR has two hands-on testers now, and this tester says LGTM. :-)

(BTW, didn't mention earlier, but thanks for this work @SuperSandro2000 🎉)

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,139 @@
...
+  const flakeLock = flakeLockParsed.data;
+
+  for (const depName of Object.keys(flakeLock.nodes)) {
+    // the root input is a magic string for the entrypoint and only references other flake inputs
+    if (depName === 'root') {
+      continue;
+    }
+
+    // skip all locked nodes which are not in the flake.nix and cannot be updated
+    const rootInputs = flakeLock.nodes['root'].inputs;
```
viceice, line 45: move outside the loop, it's constant, so no need to fetch in every iteration

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,139 @@
...
+    logger.debug(
+      { packageLockFile, error: flakeLockParsed.error },
+      `invalid flake.lock file`,
+    );
+    return null;
+  }
+
+  const flakeLock = flakeLockParsed.data;
+
+  for (const depName of Object.keys(flakeLock.nodes)) {
```
viceice, line 38:
```suggestion
  for (const [depName, flakeInput] of Object.entries(flakeLock.nodes)) {
```
you can have key and value at once, so you don't need to fetch later.

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,139 @@
...
+    const rootInputs = flakeLock.nodes['root'].inputs;
+    if (!rootInputs || !(depName in rootInputs)) {
+      logger.debug(
+        { packageLockFile, error: flakeLockParsed.error },
+        `flake.lock file cannot find "root" node or reference to ${depName}`,
+      );
+      continue;
+    }
+
+    const flakeInput = flakeLock.nodes[depName];
```
viceice, line 54:
```suggestion
```
see other comment

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,139 @@
...
+      case 'tarball':
+        deps.push({
+          depName,
+          currentValue: flakeLocked.ref,
+          currentDigest: flakeLocked.rev,
+          replaceString: flakeLocked.rev,
+          datasource: GitRefsDatasource.id,
+          // type tarball always contains this link
+          packageName: (
+            flakeOriginal.url ?? /* istanbul ignore next */ ''
```
viceice, line 122: why the fallback? isn't the url always defined?
```suggestion
            flakeOriginal.url!
```

lib/modules/manager/nix/index.ts
```diff
@@ -8,8 +8,8 @@ export const supportsLockFileMaintenance = true;
 export const url = 'https://nix.dev';
 
 export const defaultConfig = {
-  fileMatch: ['(^|/)flake\\.nix$'],
-  commitMessageTopic: 'nixpkgs',
+  fileMatch: ['(^|/)flake\\.lock'],
```
viceice, line 11: didn't we told to revert this?

lib/modules/manager/nix/extract.spec.ts
```diff
@@ -1,76 +1,508 @@
...
+        "repo": "default",
+        "type": "github"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}`;
+const flake11Lock = `{
+  "nodes": {
```
viceice, line 307: please remove unused properties from fixtures

lib/modules/manager/nix/extract.spec.ts
```diff
@@ -1,76 +1,508 @@
...
+describe('modules/manager/nix/extract', () => {
+  it('returns null when no inputs', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(flake1Lock);
+    expect(await extractPackageFile('', 'flake.nix')).toBeNull();
+  });
 
-    expect(res?.deps).toEqual([
+  it('returns nixpkgs input', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(flake2Lock);
+    expect((await extractPackageFile('', 'flake.nix'))?.deps).toEqual([
```
viceice, line 402:
```suggestion
    expect((await extractPackageFile('', 'flake.nix'))).toEqual([
```
please match full object

lib/modules/manager/nix/extract.spec.ts
```diff
@@ -1,76 +1,508 @@
...
+  });
+
+  it('includes nixpkgs with ref and shallow arguments', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(flake8Lock);
+    expect((await extractPackageFile('', 'flake.nix'))?.deps).toMatchObject([
+      {
+        currentDigest: '5633bcff0c6162b9e4b5f1264264611e950c8ec7',
+        datasource: 'git-refs',
         depName: 'nixpkgs',
-        currentValue: 'nixos-21.11',
```
viceice, line 47: is this still supported to update in the flake file?

SuperSandro2000: Nope but I would argue that this is the least useful feature for the nix package manager. For major updates especially on NixOS people should read the changelog as not every change can be notified via a warning/error and it contains extra information for special cases. Also the showed warnings should be fixed as they are turned into errors the next release or might be removed.

lib/modules/manager/nix/extract.spec.ts
SuperSandro2000: I don't think it would be a very good decision to do that as the fixtures would then be artificial and not longer real world. Also in future refactors they maybe would need to be redone as relevant information would no longer be there.

lib/modules/manager/nix/index.ts
SuperSandro2000: Oh, yeah. With the fixes from @mfenniak we no longer need that.

lib/modules/manager/nix/extract.ts
SuperSandro2000: Yeah, unless people create artificial broken lock files.

lib/modules/manager/nix/extract.ts
SuperSandro2000: not a js/ts developer, thanks

rarkins (dismissed):

mfenniak (approved):
I've performed end-to-end testing on the latest commit (d2c3b5edf4772c883cb832f523cacf0483db0a6c) and have verified that it works without any errors. ✅

I don't yet have any PRs generated from this latest commit to reference, but I'll update when one comes in.

A few behavior notes:

Note 1: The new log entry `flake.lock file cannot find "root" node or reference to ...` is routinely output whenever you have a flake input "follow" another input.  It's correct that the reference isn't updated, but maybe the log is a little pointlessly noisey.

Note 2: For some nix inputs that change frequently (eg. `rust-overlay` is updated daily), I wanted my Renovate configuration to pop these over to a weekly schedule and was surprised to find that I had to match the package name on the GitHub URL (eg. `https://github.com/oxalica/rust-overlay`), not the more obvious package name of `rust-overlay`.  (config reference: https://codeberg.org/testtrim/testtrim/src/commit/c76c3f6ef462d045605c53945cf8fe4e5136898a/renovate.json5#L16-L22).  I later realized that `matchDepNames` would have had the expected behavior.

I can see that it was what was originally implemented before this PR, so it isn't a change per-say -- but it has been extended in its implementation here.  I've proposed an addition to this PR to add documentation describing this, but it could be considered out-of-scope.  https://github.com/SuperSandro2000/renovate/pull/2

lib/modules/manager/nix/extract.spec.ts
mfenniak: I think there's an argument to be made that updating `flake.nix` references to the latest nixpkgs release stream would be good behavior to have.  Of course, this type of update isn't expected to be a quick non-breaking change, but Renovate's scope isn't limited to just updating things that are non-breaking.

I think that the scope of doing this correctly is extremely narrow... which I guess is what was present before this PR. 🙂

As a user, if I had to choose between "getting all my flake inputs updated regularly" (this PR) versus "getting my nixpkgs release reference updated twice a year", I would definitely chose this PR.  The cases where I want to use Renovate are going to be references to `unstable` (development environments), not my desktop/server where I want stable packages.

But I suppose it's a fair question -- does it have to be a choice, and why both can't be?

SuperSandro2000: > does it have to be a choice, and why both can't be?

my typescript knowledge to develop that is not good enough 😅

SuperSandro2000:
> `https://github.com/oxalica/rust-overlay`

Specifically for that you can use https://github.com/oxalica/rust-overlay/tree/stable to reduce the noise but it does not solve the general problem.

> Note 1: The new log entry flake.lock file cannot find "root" node or reference to ... is routinely output whenever you have a flake input "follow" another input. It's correct that the reference isn't updated, but maybe the log is a little pointlessly noisey.

Reduced in [`7e76bfd` (#31921)](https://github.com/renovatebot/renovate/pull/31921/commits/7e76bfd6e3100af350270f45e6e6bc5ca8708cac)

and merged your PR

mfenniak:
@viceice Sorry to be a bother, but I believe this work is blocked on your review and has been changed significantly since that review -- could I ask you to please take a look when you have time?

I'm an interested party and will help resolve any outstanding issues, as this change is key to using Renovate effectively with Nix -- running a branch-build for the past few weeks and would love to get it mainlined. :-)

h0nIg:
@viceice any update?

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,140 @@
...
+    }
+
+    const flakeLocked = flakeInput.locked;
+    const flakeOriginal = flakeInput.original;
+
+    // istanbul ignore if: if we are not in a root node then original and locked always exist which cannot be easily expressed in the type
+    if (flakeLocked === undefined || flakeOriginal === undefined) {
+      logger.debug(
+        { packageLockFile },
+        `Found empty flake input '${JSON.stringify(flakeInput)}', skipping`,
```
viceice, line 65:
```suggestion
        { packageLockFile, flakeInput },
        `Found empty flake input, skipping`,
```

lib/modules/manager/nix/extract.ts
viceice: ☝️ ❓

lib/modules/manager/nix/extract.spec.ts
viceice: can you at least move them to the relevant test when only used once?

lib/modules/manager/nix/extract.spec.ts
viceice: If we drop support for something, then we need to mark this PR as breaking and it needs to wait for next major release.

h0nIg: @viceice nixos 21.11 is equal to 2021.11 release, which hast tons of CVE and is certainly Not used anymore, this ist rather a cleanup. Current release is 2024.11

mfenniak: I'm not sure that this is actually a breaking change -- can someone follow and see if they reach the same conclusion?

The previous implementation:
- detected the presence of nixpkgs via regex (in `extract.ts`)
- would either issue `nix flake lock --update-input nixpkgs`, or `nix flake update`, depending on whether it was performing lock file maintenance or not:
https://github.com/renovatebot/renovate/blob/ff99265bb35191b07791c7ceaab95b69157e462c/lib/modules/manager/nix/artifacts.ts#L38-L47
- would never update `flake.nix` -- if flake.nix had `nixpkgs.url = "github:nixos/nixpkgs/nixos-21.11";`, `artifacts.ts` has no capability in it to update flake.nix -- it would remain a reference to nixos-21.11 forever.

The new implementation:
- detects the presence of nixpkgs (multiple) by parsing `flake.lock`, including other lock references
- issues the same commands to update the dependencies
- also never updates `flake.nix`

> is this still supported to update in the flake file?

If I'm interpreting the question correctly -- this was never implemented? -- the `flake.nix` file doesn't ever seem to be updated if it has an out-of-date reference to nixpkgs.

I don't believe this is a breaking change after all, but would love a second opinion if I'm missing something.

lib/modules/manager/nix/extract.ts
mfenniak: The case being skipped is a locked transitive dependency, it isn't an invalid lockfile.  The code comment here may be misleading -- this isn't `cannot be updated` due to a situation that needs attention -- it is `cannot be updated because it would be incorrect`.  I don't believe a warning is appropriate in this case.

lib/modules/manager/nix/extract.spec.ts
SuperSandro2000: > If we drop support for something, then we need to mark this PR as breaking and it needs to wait for next major release.

The previous nix flake support was very broken and barely ever issued any updates. It basically only ever issued a PR after a major update which by upstream recommendation *highly* recommends to read the changelog and fix deprecations in there. It never refreshed the lock file afterwards which is absolutely required as nixpkgs is not using release tags but advancing branches which get important security updates. It also only worked with nixpkgs which is the most comment flake input but definitely not the only one that needs to be updated.

So I don't think we loose an important feature, only one which is broken by design.

Also since this PR was opened, there was at least one major update but since this went so stale, it didn't get into there.

> would never update flake.nix -- if flake.nix had nixpkgs.url = "github:nixos/nixpkgs/nixos-21.11";, artifacts.ts has no capability in it to update flake.nix -- it would remain a reference to nixos-21.11 forever.

I think it previously updated the flake.nix on a major version update and then refreshed the lock file once. From a major NixOS contributor perspective this is doing things wrong.

lib/modules/manager/nix/extract.ts
SuperSandro2000: Yep. We cannot touch the transitive dependencies as that is not possible via the cli and manually changing those dependencies results in unexpected lock files. Building this relation in zod is probably very complicated, so we just skip them here for the type check.

If people want to update those deps, they add the flake input as their input and use follows.

SuperSandro2000:
Can we finally get this merged?

The current flake support in renovate is practically none existing and doesn't align with any standard nix flake workflows.

h0nIg:
@viceice please get this merged, thank you

lib/modules/manager/nix/extract.spec.ts
viceice: so this no longer works?
https://github.com/renovate-reproductions/31921-nix-rework/pull/3/files

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,140 @@
...
+          datasource: GitRefsDatasource.id,
+          packageName: flakeOriginal.url,
+        });
+        break;
+      case 'sourcehut':
+        deps.push({
+          depName,
+          currentValue: flakeOriginal.ref,
+          currentDigest: flakeLocked.rev,
+          replaceString: flakeLocked.rev,
```
viceice, line 111: this is wrong and causes autoreplace to fail, drop it
```suggestion
```

```
 INFO: Cannot find replaceString in current file content. Was it already updated? (repository=renovate-reproductions/31921-nix-rework, packageFile=flake.nix, branch=renovate-test/nixpkgs-208.x)
       "depName": "nixpkgs",
       "existingContent": "{\n  description = \"A simple script\";\n\n  inputs = {\n    nixpkgs.url = \"github:NixOS/nixpkgs/nixos-21.11\";\n  };\n\n  outputs = { self, nixpkgs }: {\n    defaultPackage.x86_64-linux = self.packages.x86_64-linux.my-script;\n\n    packages.x86_64-linux.my-script =\n      let\n        pkgs = import nixpkgs { system = \"x86_64-linux\"; };\n      in\n      pkgs.writeShellScriptBin \"my-script\" ''\n        DATE=\"$(${pkgs.ddate}/bin/ddate +'the %e of %B%, %Y')\"\n        ${pkgs.cowsay}/bin/cowsay Hello, world! Today is $DATE.\n      '';\n  };\n}\n",
       "replaceString": "eabc38219184cc3e04a974fe31857d8e0eac098d"
DEBUG: No content changed (repository=renovate-reproductions/31921-nix-rework, packageFile=flake.nix, branch=renovate-test/nixpkgs-208.x)
       "depName": "nixpkgs"
```

viceice:
Please verify on forks of these repos that the new implementation works as intended:
- https://github.com/renovate-reproductions/31921-nix-rework
- https://github.com/Simon-Boyer/nixos-config
- https://github.com/genebean/flakeypi

The current implementation fails on the first repo.

viceice:
you can easily find more repos for tersting https://github.com/search?q=nixpkgs.url+path%3Aflake.nix&type=code

SuperSandro2000:
> The current implementation fails on the first repo.

It no longer bumps nixpkgs-21.11 to nixpkgs-24.11 as I already described multiple times before. I lack the typescript knowledge to implement support for both updating this and general lock file updates. After bumping that manual once https://github.com/SuperSandro2000/31921-nix-rework/commit/6b394afa3f014d745a82dc1b662478051fcacc9b this PR is created which is the intended feature https://github.com/SuperSandro2000/31921-nix-rework/pull/1

lib/modules/manager/nix/extract.ts
```diff
@@ -1,23 +1,140 @@
...
+
+    switch (flakeLocked.type) {
+      case 'github':
+        deps.push({
+          depName,
+          currentValue: flakeOriginal.ref,
+          currentDigest: flakeLocked.rev,
+          replaceString: flakeLocked.rev,
+          datasource: GitRefsDatasource.id,
+          packageName: `https://${flakeOriginal.host ?? 'github.com'}/${flakeOriginal.owner}/${flakeOriginal.repo}`,
```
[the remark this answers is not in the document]
SuperSandro2000: Those tags mark the fork off point for that release. Updating to them is wrong. Nixpkgs and NixOS use rolling branches and only those matter. People are expected to update their flake.lock files somewhat often to get the latest commits on that branch.

lib/modules/manager/nix/extract.ts
SuperSandro2000: Good spot. Removed that and created https://github.com/SuperSandro2000/31921-nix-rework/pull/2 with it

lib/modules/manager/nix/extract.spec.ts
SuperSandro2000: Yep. That is the minor feature that got removed. It was limited to nixpkgs only and didn't support all flake inputs. Also it only ever updated the nixpkgs release once for every 6 month release but flake inputs should be updated more often especially within the release to receive security fixes.

ojsef39:
> you can easily find more repos for tersting https://github.com/search?q=nixpkgs.url+path%3Aflake.nix&type=code

Im adding mine to the list since my config is bit weird and gets all its inputs from a base config in not expecting it to work but would be cool:

https://github.com/ojsef39/nix-personal

SuperSandro2000:
Below tests where run with [`5b12fed` (#31921)](https://github.com/renovatebot/renovate/pull/31921/commits/5b12fed1a048640408ac8895da584bce584d7364)

> Im adding mine to the list since my config is bit weird and gets all its inputs from a base config im not expecting it to work but would be cool:
> 
> [ojsef39/nix-personal](https://github.com/ojsef39/nix-personal?rgh-link-date=2025-01-25T22%3A25%3A45Z)

https://github.com/SuperSandro2000/nix-personal/pull/1

> [renovate-reproductions/31921-nix-rework](https://github.com/renovate-reproductions/31921-nix-rework?rgh-link-date=2025-01-24T11%3A08%3A37Z)

https://github.com/SuperSandro2000/31921-nix-rework/pull/1/files

> [Simon-Boyer/nixos-config](https://github.com/Simon-Boyer/nixos-config?rgh-link-date=2025-01-24T11%3A08%3A37Z)

That repo has no renovate config.

> you can easily find more repos for tersting [github.com/search?q=nixpkgs.url+path%3Aflake.nix&type=code](https://github.com/search?q=nixpkgs.url+path%3Aflake.nix&type=code&rgh-link-date=2025-01-24T11%3A09%3A52Z)

I have testing repos myself (https://github.com/NuschtOS/nixos-modules/issues/130 https://github.com/NuschtOS/search/issues/97) where I created PRs in the past and I have the latest version of this branched deployed to my renovate instance.

elikoga:
> > The current implementation fails on the first repo.
> 
> It no longer bumps nixpkgs-21.11 to nixpkgs-24.11 as I already described multiple times before. I lack the typescript knowledge to implement support for both updating this and general lock file updates. After bumping that manual once [SuperSandro2000/31921-nix-rework@6b394af](https://github.com/SuperSandro2000/31921-nix-rework/commit/6b394afa3f014d745a82dc1b662478051fcacc9b) this PR is created which is the intended feature [SuperSandro2000/31921-nix-rework#1](https://github.com/SuperSandro2000/31921-nix-rework/pull/1)

I think there's a bit of confusion here. The existing manager is obviously a nixpkgs manager that has domain-knowledge of `github:nixos/nixpkgs` release channels and makes the breaking change for you. Additionally (as a feature) it updates the flake lockfile so that this is a complete integration.

We additionally need a nix integration that updates flake.lock in general, which this PR tackles as far as I can see.

I believe that we need to preserve both behaviors and distinguish that these are two different managers.

@SuperSandro2000 is my assessment correct or am I missing something

elikoga:
See this discussion and line change.

https://github.com/renovatebot/renovate/pull/31921/files#diff-ba688fc899a85e87e072cb5408f08fa6a579a14f608d75da4f711fc02505fa8aL12-R12

I think there is one obvious change. Change the docs to reflect this. Renovate has a nixpkgs manager, for the nix ecosystem. Renovate does not have a flake.lock manager. I think I'll try a PR sometime.

SuperSandro2000:
> I believe that we need to preserve both behaviors and distinguish that these are **two different managers**.

In the current form the PR does both. I quickly added the old behaviour back after realizing that I can just append the list with the old logic.

I don't really want to split them in two managers at this point 😅

h0nIg:
i don't see a need to split the manager as well. With the old and new behaviour, the "breaking" label can get removed?

elikoga:
> In the current form the PR does both. I quickly added the old behaviour back after realizing that I can just append the list with the old logic.

Seems like I missed that. I'm looking forward to when it's merged then! 🤞

viceice (approved):
thankyou for being patient and react on my concerns. so now we've a non breaking feature 🎉

SuperSandro2000:
I think we overlooked something before merging....

```
DEBUG: currentDigest not found in string to replace (repository=NuschtOS/ixx, branch=renovate/flake-inputs)
       "stringToReplace": "nixos-unstable",
       "currentDigest": "5df43628fdf08d642be8ba5b3625a6c70731c19c"
DEBUG: Starting search at index 55 (repository=NuschtOS/ixx, packageFile=flake.nix, branch=renovate/flake-inputs)
       "depName": "nixpkgs"
DEBUG: Found match at index 55 (repository=NuschtOS/ixx, packageFile=flake.nix, branch=renovate/flake-inputs)
       "depName": "nixpkgs"
DEBUG: Digest is not updated (repository=NuschtOS/ixx, packageFile=flake.nix, branch=renovate/flake-inputs)
       "depName": "nixpkgs",
       "manager": "nix",
       "expectedValue": "9d3ae807ebd2981d593cddd0080856873139aa40",
       "foundValue": "5df43628fdf08d642be8ba5b3625a6c70731c19c"
 WARN: Error updating branch: update failure (repository=NuschtOS/ixx, branch=renovate/flake-inputs)
DEBUG: syncBranchState() (repository=NuschtOS/ixx, branch=renovate/crates)
```

which results in a problem: stringToReplace is not looking into the flake.lock and also nix is never run ... 🫠

SuperSandro2000:
Should be fixed in #33991

PR #31927, opened 2024-10-12
fix: activate repo on automerge branch

EpicPlayerA10:
## Changes

Right now, repo is activated only when any renovate PR has been merged. But what if the user set up the renovate with an option `automergeType: "branch"`? Then there isn't any PR created and the repo can't be activated in any way.

This PR fixes it by additionally checking if renovate has made any commit to the base branches. If so, then the repo should be activated.

## Context

https://github.com/renovatebot/renovate/discussions/31875


rarkins (changes requested):
This change isn't ok because it will result in git cloning every run.

EpicPlayerA10:
Hmm, you are right. My first idea to this was to directly modify the `repoIsActivated` on automerge, but i don't know if `repoIsActivated` option will persist between runs.

So I need to do the `hasCommittedTo` method per platform?

rarkins:
Closing this PR, as the code can't be used. Please return to discussing in the Discussion

PR #31934, opened 2024-10-13
feat(submodules): use env vars from getGitEnvironmentVariables

tim-thacker-nullify:
## Changes

This change uses the existing `getGitEnvironmentVariables` function during the cloning of git submodules.

This function has been used to include support for updating private git ssh submodules on github.com, but it was not added in the function triggered when the `cloneSubmodules` config file option is set to true.

see: https://github.com/renovatebot/renovate/pull/22061

## Context

https://github.com/renovatebot/renovate/discussions/30942

Private submodules on the github.com app cannot currently be cloned with the `cloneSubmodules` option when the `.gitmodules` file is configured with ssh.

This is because the GitHub App token does not have SSH permissions, it can only auth via HTTP.

My use case is private Go submodules on github.com, for the `"postUpdateOptions": ["gomodTidy", "gomodUpdateImportPaths"],` options to work, I need my private git submodules cloned.


lib/util/git/index.ts
```diff
@@ -349,13 +350,20 @@ export async function cloneSubmodules(shouldClone: boolean): Promise<void> {
     return;
   }
   submodulesInitizialized = true;
+  const gitSubmoduleAuthEnvironmentVariables = getGitEnvironmentVariables();
+  const gitEnv = {
+    // pass all existing env variables
+    ...process.env,
```
rarkins, line 356: Was `git` already passing all env variables before this? Because normally we'd want this controlled by `exposeAllEnv`.

tim-thacker-nullify: not sure what you mean exactly, but right now git is not configured with the ENV var that forces submodules to be cloned with https rather than ssh.

tim-thacker-nullify: this is the discussion about the problem https://github.com/renovatebot/renovate/discussions/30942

rarkins: The question I'm asking is: does the `simple-git` library pass the entire env through to `git` already today?

Renovate contains controls for admins to decide if they want to pass all env (including secrets) through to child processes. I just want to make sure that we're not opening up any security holes here.

lib/util/git/index.ts
```diff
@@ -24,7 +24,6 @@ import type { GitProtocol } from '../../types/git';
 import { incLimitedValue } from '../../workers/global/limits';
 import { getCache } from '../cache/repository';
 import { newlineRegex, regEx } from '../regex';
-import { matchRegexOrGlobList } from '../string-match';
```
rsahrakorpi-cambri, line 27: this import is still needed

lib/util/git/index.ts
```diff
@@ -353,6 +352,13 @@ export async function cloneSubmodules(
     return;
   }
   submodulesInitizialized = true;
+  const gitSubmoduleAuthEnvironmentVariables = getGitEnvironmentVariables();
```
rsahrakorpi-cambri, line 355: need to add the import for this

lib/util/git/index.ts
```diff
@@ -366,7 +372,7 @@ export async function cloneSubmodules(
     try {
       logger.debug(`Cloning git submodule at ${submodule}`);
       await gitRetry(() =>
-        git.submoduleUpdate(['--init', '--recursive', submodule]),
+        git.env(gitEnv).submoduleUpdate(['--init', '--recursive', submodule]),
```
rsahrakorpi-cambri, line 375:
```suggestion
        git.env(getGitEnvironmentVariables()).submoduleUpdate(['--init', '--recursive', submodule]),
```

is sufficient to get cloning of private submodules to work. Passing process.env doesn't seem necessary

rsahrakorpi-cambri (commented):
Would be good to get some traction on this PR. I did some quick testing and the solution works for cloning submodules (once you fix the imports)

lib/util/git/index.ts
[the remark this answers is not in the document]
rsahrakorpi-cambri: yes, because that is how it is configured on L241 https://github.com/renovatebot/renovate/blob/main/lib/util/git/index.ts#L241

lib/util/git/index.ts
```diff
@@ -24,7 +24,6 @@ import type { GitProtocol } from '../../types/git';
 import { incLimitedValue } from '../../workers/global/limits';
 import { getCache } from '../cache/repository';
 import { newlineRegex, regEx } from '../regex';
-import { matchRegexOrGlobList } from '../string-match';
 import { parseGitAuthor } from './author';
```
rsahrakorpi-cambri, line 27:
```suggestion
import { matchRegexOrGlobList } from '../string-match';
import { getGitEnvironmentVariables } from './auth';
import { parseGitAuthor } from './author';
```

stazz:
I know this is old-ish PR but it would be nice to get it merged, I think @rsahrakorpi-cambri 's reviews had a good points. There probably shouldn't be any major blockers anymore for this?

The reason I'm asking is because this bug currently blocks us from using Mend-hosted Renovate (which would also have a nice benefit of auto-approve working out of the box), and even with self-hosted one we need to use [workaround with magical env vars](https://zenn.dev/sijiaoh/articles/b91d741e52f18a).

lib/util/git/index.ts
viceice: > The question I'm asking is: does the `simple-git` library pass the entire env through to `git` already today?
> 
> Renovate contains controls for admins to decide if they want to pass all env (including secrets) through to child processes. I just want to make sure that we're not opening up any security holes here.

yes, it's passing through to to `child_process` module, which uses `process.env` if no env is passed by user.

rsahrakorpi-cambri:
ping @rarkins any thoughts on this? Can we proceed with it?

rarkins:
If the original creator is no longer updating this PR then feel free to copy this branch into your own fork and resubmit it, replacing this one.

RGunning:
I forked and resubmitted this PR with the fix to the imports https://github.com/renovatebot/renovate/pull/35180

RahulGautamSingh:
Replacement exists

PR #31958, opened 2024-10-15
docs(schedule): improve description, deprecate Later syntax

rarkins:

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -3780,35 +3780,32 @@ We recommend you do this selectively with `packageRules` and not globally.
 
 ## schedule
 
-The `schedule` option allows you to define times of week or month for Renovate updates.
-Running Renovate around the clock can be too "noisy" for some projects.
-To reduce the noise you can use the `schedule` config option to limit the time frame in which Renovate will perform actions on your repository.
-You can use the standard [Cron syntax](https://crontab.guru/crontab.5.html) and [Later syntax](https://github.com/breejs/later) to define your schedule.
+The `schedule` option allows you to define times of the day, week or month when you are willing to allow Renovate to create branches.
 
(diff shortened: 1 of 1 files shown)


docs/usage/configuration-options.md
```diff
@@ -3780,77 +3780,49 @@ We recommend you do this selectively with `packageRules` and not globally.
...
 
-The default value for `schedule` is "at any time", which is functionally the same as declaring a `null` schedule.
-i.e. Renovate will run on the repository around the clock.
+Setting a `schedule` does not itself cause or trigger Renovate to run. It's like putting a sign on your office which says "DHL deliveries only accepted between 9-11am". Such a sign won't _cause_ DHL to come to your office only at 9-11am, instead it simply means that if they come at any other time of the day then they'll honor the sign and skip you. It also means that if they rarely attempt between 9-11am then you'll often get no deliveries in a day. Similarly, if you set too restrictive of a Renovate `schedule` and the chance of Renovate running on your repo during those hours is low, then you might find your dependency updates regularly skipped. For this reason we recommend you usually allow a time window of at least 3-4 hours in any `schedule` unless your instance of Renovate is expected to run more frequently than that.
+
+Renovate supports the standard [Cron syntax](https://crontab.guru/crontab.5.html) as well as deprecated support for a subset of [Later syntax](https://github.com/breejs/later).
```
viceice, line 3787: I would simply drop later from docs and add a debug message to the code when a deprecated later string was found.

rarkins: I don't want to warn users yet - there's a lot of use in the wild. We should instead migrate as many as possible automatically before falling back to user action required

secustor (approved):

docs/usage/configuration-options.md
viceice: don't warn, just a debug log 🤗

Issue #19056, opened 2022-11-23
Change Flux tests to use inline fixtures


onedr0p:
### Describe the proposed change(s).

Discussion was brought up here: https://github.com/renovatebot/renovate/pull/19017/files/a34a061b34bafd8d8cb41d1504e7f7164b4246a0#diff-47684b7d836903a8e38ac22ae0b985ad7919d0e8347ce5182a6a669236e41148

PR #31962, opened 2024-10-15
feat(manager/flux): support chartRef in HelmRelease resources

ssams:
## Changes

Adds support for `chartRef` fields within `HelmRelease` resources in the _flux_ manager, as described in #31945.

## Context

- #31945


lib/modules/manager/flux/extract.ts
```diff
@@ -235,6 +301,27 @@ function resolveResourceManifest(
...
+function collectHelmReposAndCharts(
+  manifests: FluxManifest[],
+): [HelmRepository[], HelmChart[]] {
+  const helmRepositories: HelmRepository[] = [];
+  const helmCharts: HelmChart[] = [];
+
+  for (const manifest of manifests) {
+    if (manifest.kind === 'resource') {
+      for (const resource of manifest.resources) {
+        if (resource.kind === 'HelmRepository') {
```
viceice, line 313: use switch case

lib/modules/manager/flux/extract.ts
```diff
@@ -235,6 +301,27 @@ function resolveResourceManifest(
   return deps;
 }
 
+function collectHelmReposAndCharts(
```
viceice, line 304: move to `./common.ts`

lib/modules/manager/flux/extract.ts
```diff
@@ -117,29 +122,92 @@ function resolveSystemManifest(
   ];
 }
 
+function resolveHelmReleaseManifest(
```
viceice, line 125: move to `./common.ts`

lib/modules/manager/flux/__fixtures__/helmChart.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: source.toolkit.fluxcd.io/v1
```
viceice, line 1: can you join both fixtures and inline them to test? or will it cause code coverage issues?

ssams: could be also inlined in _most_ tests, but they are also used in one test for `extractAllPackageFiles`, where file paths are currently expected in the results (https://github.com/renovatebot/renovate/pull/31962/files#diff-e4fcec4565b58234272751cf6a22e089c6e13b651bb297b4db19834442b124e1R960-R961). 

quick search shows that this seems to be related to #19056, similar situation as in the referenced discussion/comments there?

ssams: so, how do we proceed here?

rarkins: It's ok to leave some fixtures if it's impossible or painful to inline them. We can ignore the automated "no new fixtures" test and merge anyway

lib/modules/manager/flux/extract.ts
```diff
@@ -120,26 +124,29 @@ function resolveSystemManifest(
...
     switch (resource.kind) {
       case 'HelmRelease': {
-        const dep: PackageDependency = {
-          depName: resource.spec.chart.spec.chart,
-          currentValue: resource.spec.chart.spec.version,
-          datasource: HelmDatasource.id,
-        };
+        if (resource.spec.chartRef?.kind === 'OCIRepository') {
+          // Can be skipped, version is handled via OCIRepository resource directly
+          break;
```
viceice, line 136: why break instead of continue?
```suggestion
          continue;
```

log a trace message

lib/modules/manager/flux/extract.ts
```diff
@@ -120,26 +124,29 @@ function resolveSystemManifest(
...
-              (resource.spec.chart.spec.sourceRef.namespace ??
-                resource.metadata?.namespace),
+        const resolvedRelease = resolveHelmReleaseManifest(
+          resource,
+          helmRepositories,
+          helmCharts,
         );
+        if (!resolvedRelease) {
+          // invalid or incomplete HelmRelease spec, thus failed to extract the dependency itself
+          break;
```
viceice, line 146: why break instead of continue?
```suggestion
          continue;
```

lib/modules/manager/flux/common.ts
```diff
@@ -8,3 +12,87 @@ export const systemManifestHeaderRegex =
...
+export function resolveHelmReleaseManifest(
+  resource: HelmRelease,
+  helmRepositories: HelmRepository[],
+  helmCharts: HelmChart[],
+): {
+  name: string;
+  dep: PackageDependency;
+  matchingRepositories: HelmRepository[];
+} | null {
+  if (resource.spec.chartRef) {
```
viceice, line 25: invert condition to reduce nesting.

lib/modules/manager/flux/extract.spec.ts
```diff
@@ -297,6 +342,68 @@ describe('modules/manager/flux/extract', () => {
...
+        Fixtures.get('helmChartRefRelease.yaml'),
+        'test.yaml',
+      );
+      expect(result).toBeNull();
+    });
+
+    it('extracts HelmRelease resources using a chartRef without a valid HelmRepository', () => {
+      const result = extractPackageFile(
+        codeBlock`
+          ${Fixtures.get('helmChartRefRelease.yaml')}
```
viceice, line 356: move to module const if used multiple times (perf)

lib/modules/manager/flux/common.ts
ssams: not sure what you mean here? I can do a `if (!resource.spec.chartRef) { return null; }`, but note that there's an `else if (resource.spec.chartRef) {` later, so I don't think the condition here can be dropped completely. so the nesting level would remain unchanged anyway? unless I'm missing something on how you suggest to update the condition?

viceice:
conflicted

lib/modules/manager/flux/common.ts
```diff
@@ -8,3 +12,87 @@ export const systemManifestHeaderRegex =
...
+          (chartRef.namespace ?? resource.metadata?.namespace),
+    );
+    if (!helmChart) {
+      return null;
+    }
+    const chartName = helmChart.spec.chart;
+
+    const dep: PackageDependency = {
+      depName: chartName,
+      currentValue: helmChart.spec.version,
```
viceice, line 41: wait, this won't work as intended when the `HelmChart` is in a different file than the `HelmRelease` because renovate tries to replace the version in the wrong file.

instead of resolving the chartRef of helm release, you should collect the helm chart as separate dependency.

viceice (changes requested):

lib/modules/manager/flux/common.ts
ssams: right, I missed this and tested it only in cases where it was located in the same file, where it worked. on second look, it's even apparent in the test case where the expectation for the `packageFile` references the wrong name due to that.

I'll update/rework the MR accordingly

ssams:
unfortunately a lot of commits were introduced accidentally on this branch when merging with `main` somehow, but as far as I understand that shouldn't be a problem as you'll squash this anyway when merging? at least wanted to avoid force committing to remove them for now as described in the contrib guide, let me know if you need any changes to clean that up. the diff itself is clean though.

lib/modules/manager/flux/common.ts
```diff
@@ -8,3 +10,19 @@ export const systemManifestHeaderRegex =
 export function isSystemManifest(file: string): boolean {
   return regEx(systemManifestFileNameRegex).test(file);
 }
+
+export function collectHelmRepos(manifests: FluxManifest[]): HelmRepository[] {
```
viceice, line 14: please create a separate refactor pr for this

lib/modules/manager/flux/extract.ts
```diff
@@ -102,6 +106,39 @@ function resolveGitRepositoryPerSourceTag(
   }
 }
 
+function resolveHelmRepository(
```
viceice, line 109: same, extract to refactor pr. can combined with comment above

ssams: sorry, but I cannot really follow this. the extracted code in this function is _literally_ identical to what was previously within the switch statement, and I need the _exact_ same logic in the new case (as it's just about resolving the details of the referenced repository, properly supporting OCI repos etc). If I don't extract this into a function, I'd have to duplicate the whole block, copy+paste it down into the new case. which causes test coverage to decrease, and forcing the addition (and duplication) of test cases just to test the same code again in the other context - and to remove all of these additional tests again right afterwards in yet another PR too? that doesn't make sense.

I've removed the other refactoring as requested though, as that was in a part that is now mostly unchanged again after working on the various topics. but I really don't see how the second one here could be avoided without creating more questionable quality issues.

rarkins: You're being asked to split this PR into two:
- A PR containing only refactorings
- A second PR containing the new functionality

ssams: argh, you seriously want another one to be merged _first_ then? given the previous experience on this one so far I'm not sure how much more time I'm able to invest on this, I only wanted to contribute a seemingly simple addition which now starts to take hours due to these nits and indirections (the dev setup and tooling itself is good though, compliments on that part).
But well, I'll try to find the time, otherwise somebody else may need to take over these changes if there's interest in it...

rarkins (dismissed):

viceice (dismissed):

viceice (commented):
sorry conflicted

viceice (dismissed):

ssams:
do you expect me to keep it updated now that it failed the merge checks since the tests have been modified on `main` again? missing some feedback on how to proceed here, but I suppose despite the latest comment about not updating the branch this is one of the mentioned "specific reason to do so"...

PR #32011, opened 2024-10-18
feat(manager/nuget): add support for "disabledPackageSources" in nuget.config

Otiel:
## Changes

This PR adds support for the `disabledPackageSources` element that can be present in the `nuget.config` file. See the Microsoft documentation about [disabledPackageSources](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#disabledpackagesources).

## Context

Closes https://github.com/renovatebot/renovate/discussions/31379

## Documentation

## How I've tested my work


Otiel:
@rarkins I've noticed you disabled auto-merge, was it due to a misclick when enabling auto-merge? Or is there some action still required from my part in order for you to review this PR?

rarkins:
Enabling automerge was a misclick at the time

lib/modules/manager/nuget/util.ts
```diff
@@ -52,13 +52,17 @@ export async function getConfiguredRegistries(
   }
 
   const packageSources = nuGetConfig.childNamed('packageSources');
-  if (!packageSources) {
+  const disabledPackageSources = nuGetConfig.childNamed(
+    'disabledPackageSources',
+  );
+
+  if (!packageSources && !disabledPackageSources) {
```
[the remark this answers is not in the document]
Otiel: `disabledPackageSources` should be parsed in order to check if one of the default registries (there is only "nuget.org" currently) should be disabled.

I've added two tests to cover this case.

lib/modules/manager/nuget/util.spec.ts
```diff
@@ -106,6 +106,156 @@ describe('modules/manager/nuget/util', () => {
         'NuGet.Common',
       ]);
     });
+
+    it('reads nuget config file with default registry disabled', async () => {
+      fs.findUpLocal.mockReturnValue(
+        Promise.resolve<string | null>('NuGet.config'),
+      );
```
viceice, line 113: use mockResolvedValue

lib/modules/manager/nuget/util.spec.ts
```diff
@@ -106,6 +106,156 @@ describe('modules/manager/nuget/util', () => {
...
+
+      const registries = await getConfiguredRegistries('NuGet.config');
+      expect(registries?.length).toBe(0);
+    });
+
+    it('reads nuget config file with default registry disabled and added sources', async () => {
+      fs.findUpLocal.mockReturnValue(
+        Promise.resolve<string | null>('NuGet.config'),
+      );
```
viceice, line 130: same here

lib/modules/manager/nuget/util.spec.ts
```diff
@@ -106,6 +106,156 @@ describe('modules/manager/nuget/util', () => {
...
+              <add key="nuget.org" value="true" />
+            </disabledPackageSources>
+          </configuration>`,
+      );
+
+      const registries = await getConfiguredRegistries('NuGet.config');
+      expect(registries?.length).toBe(1);
+      expect(registries![0].name).toBe('contoso.com');
+      expect(registries![0].url).toBe('https://contoso.com/packages/');
+    });
```
viceice, line 147: use single toEqual

lib/modules/manager/nuget/util.ts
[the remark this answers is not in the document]
rarkins: What should be Renovate's behavior if the only default registry is disabled _and_ there's no other registries specified? What's nuget's behavior in such cases, e.g. if you run it with no package sources plus nuget disabled?

lib/modules/manager/nuget/util.spec.ts
Otiel: I'm not sure I understand this comment. We do `fs.readLocalFile.mockResolvedValueOnce(` line 114, is that not correct (sorry, I don't have any typescript knownledge)?

lib/modules/manager/nuget/util.ts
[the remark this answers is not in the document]
Otiel: Nuget's behavior in such case is that it will simply fail to restore the nuget packages (as if all the sources were unreachable).

I think Renovate should fail as well, but that is really an edge case. If "nuget.org" is disabled and no other registry is specified, I would expect developers to encounter issues on the CI far before calling Renovate.

But in fact, the current implementation in Renovate is that it will fallback to the default registry ("nuget.org") if no registry is retrieved calling `getConfiguredRegistries`:

lib\modules\manager\nuget\artifacts.ts:34
```
async function createCachedNuGetConfigFile(
...
    (await getConfiguredRegistries(packageFileName)) ?? getDefaultRegistries();
```

lib/modules/manager/nuget/util.spec.ts
Otiel: Sorry, TypeScript newbie 😞  . I can't find any "single" method in Typescript doc or in Jest, thus I don't understand what you mean.

Can you show me the entire line modified please?

lib/modules/manager/nuget/util.ts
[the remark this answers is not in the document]
rarkins: Yes, Renovate by default will use the default registry for a datasource, i.e. trying nuget.org as you describe. Although it's usually best that we match the way each packager manager works, I don't see much use in attempting to replicate its failure in this case.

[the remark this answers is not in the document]
Otiel: > I don't see much use in attempting to replicate its failure in this case.

I agree. 👍 

So we're fine as it is? I'll let you resolve this conversation if that's OK for you.

Or you want me to modify the `getConfiguredRegistries` function in order to return the default registry in case there is no other `packageSource`?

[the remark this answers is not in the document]
rarkins: I'm n such case we should skip returning any registryUrls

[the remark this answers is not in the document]
Otiel: So the code is correct if the following test passes?

```
  it('reads nuget config file without packageSources and with default registry disabled', async () => {
    fs.findUpLocal.mockReturnValue(
      Promise.resolve<string | null>('NuGet.config'),
    );
    fs.readLocalFile.mockResolvedValueOnce(
      codeBlock`
        <configuration>
          <disabledPackageSources>
                <add key="nuget.org" value="true" />
          </disabledPackageSources>
        </configuration>`,
    );

    const registries = await getConfiguredRegistries('NuGet.config');
    expect(registries?.length).toBe(0);
  });
```

lib/modules/manager/nuget/util.spec.ts
Otiel: @viceice Any help please? 😁

lib/modules/manager/nuget/util.spec.ts
Otiel: @viceice Let me know if the latest changes correspond to what you're expecting (pun intended).

lib/modules/manager/nuget/util.spec.ts
Otiel: @viceice Let me know if the latest changes correspond to what you're expecting.

lib/modules/manager/nuget/util.spec.ts
Otiel: Fixed.

lib/modules/manager/nuget/util.ts
rarkins, line 59:
```suggestion
  if (!packageSources) {
```

Shouldn't we return here if there are no packageSources? If there's no packageSources, there's nothing to disable. This would allow the if() later down to be reverted

rarkins (changes requested):

lib/modules/manager/nuget/util.ts
Otiel: >  If there's no packageSources, there's nothing to disable

That's not true, there is always the default registry that could be disabled. That's what we discussed already [here](https://github.com/renovatebot/renovate/pull/32011#discussion_r1814422778).

Otiel: @rarkins Do you agree?

rarkins: I'm not sure I do.

If there are no packageSources, BUT there are disabledPackageSources, what do you expect/recommend the resulting behavior to be?

Otiel: In any case, **I would expect Renovate to respect exactly the content of the _nuget.config_ file in order to match the behavior of NuGet** when NuGet is called outside of Renovate context.

That means:
* If there are no packageSources, BUT there are disabledPackageSources, AND the default source (nuget.org) is NOT among the disabledPackageSources, use nuget.org.
* If there are no packageSources, BUT there are disabledPackageSources, AND the default source (nuget.org) is among the disabledPackageSources, disable nuget.org.
In that case, the user will end up with a configuration with no source, thus NuGet will fail to restore/update the packages. I agree it kind of make no sense to end up with a configuration with no source (or, at least, I don't see a valid use case), but that is what the user will get locally in his repository or on his CI using dotnet / NuGet, so it _does_ make sense to get the same behavior with Renovate.

rarkins: If you leave `registryUrls` blank then Renovate would go back to defaulting to the datasource's default registry (i.e. nuget.org).

So instead you'd need to disable all dependencies by setting a `skipReason`, which can't be overridden. Are you sure you we should try to hard to intentionally fail here, rather than allowing a nuget lookup?

Otiel: Ok, I see what you mean. 👌 

I've pushed a commit with the related changes.

rarkins:
@Otiel please don't merge from main unless it's to resolve conflicts. It resets the CI approval each time

lib/modules/manager/nuget/util.ts
```diff
@@ -52,13 +52,22 @@ export async function getConfiguredRegistries(
   }
 
   const packageSources = nuGetConfig.childNamed('packageSources');
+  const disabledPackageSources = nuGetConfig.childNamed(
+    'disabledPackageSources',
+  );
```
rarkins, line 57: Move this lower closer to where it's used

rarkins (changes requested):

lib/modules/manager/nuget/util.ts
Otiel: 👍 Done

lib/modules/manager/nuget/util.ts
```diff
@@ -52,13 +52,19 @@ export async function getConfiguredRegistries(
   }
 
   const packageSources = nuGetConfig.childNamed('packageSources');
+
   if (!packageSources) {
+    // If there are no packageSources, don't even look for any
+    // disabledPackageSources
+    // Even if NuGet default source (nuget.org) was among the
+    // disabledPackageSources, Renovate will default to the default source
```
viceice, line 60: should we issue a warning if default source is disabled?

Otiel: @viceice Do you mean if default source is disabled and there are no packageSources, or if default source is disabled and there are other packageSources?

Otiel: @viceice It is a perfectly valid case to disable the default source. For instance, in a corporate environment, you might not want to query directly nuget.org, but a proxy for security reasons.

I don't think we should issue a warning. Feel free to reopen if you don't agree as I'm not sure if your comment was really a question or more a suggestion (in that case, I'll implement the warning).

Otiel:
@rarkins Bump for review again please. 😊

rarkins (dismissed):

rarkins (approved):

rarkins (dismissed):

Otiel:
@rarkins Bump for review again please. 😊

lib/modules/manager/nuget/util.spec.ts
```diff
@@ -59,22 +57,20 @@ describe('modules/manager/nuget/util', () => {
...
+      expect(registries![0]).toEqual({
+        name: 'nuget.org',
+        url: 'https://api.nuget.org/v3/index.json',
+        sourceMappedPackagePatterns: ['*'],
+      });
+      expect(registries![1]).toEqual({
+        name: 'contoso.com',
+        url: 'https://contoso.com/packages/',
+        sourceMappedPackagePatterns: ['Contoso.*', 'NuGet.Common'],
+      });
```
viceice, line 69: don't change existing tests if not needed, otherwise combine this to a single `toEqual` or `toMatchObject`

lib/modules/manager/nuget/util.spec.ts
```diff
@@ -95,17 +91,133 @@ describe('modules/manager/nuget/util', () => {
...
+    });
```
viceice, line 104: same here

lib/modules/manager/nuget/util.spec.ts
```diff
@@ -95,17 +91,133 @@ describe('modules/manager/nuget/util', () => {
...
+      const registries = await getConfiguredRegistries('NuGet.config');
+      expect(registries?.length).toBe(2);
+      expect(registries![0]).toEqual({
+        name: 'nuget.org',
+        url: 'https://api.nuget.org/v3/index.json',
+      });
+      expect(registries![1]).toEqual({
+        name: 'contoso.com',
+        url: 'https://contoso.com/packages/',
+      });
```
viceice, line 176: combine this to a single `toEqual` or `toMatchObject`

lib/modules/manager/nuget/util.spec.ts
```diff
@@ -95,17 +91,133 @@ describe('modules/manager/nuget/util', () => {
...
```
viceice, line 204: same

lib/modules/manager/nuget/util.ts
```diff
@@ -108,9 +114,27 @@ export async function getConfiguredRegistries(
           );
         }
       }
-      // child.name === 'remove' not supported
```
viceice, line 111: why removed?

viceice:
use GitHub re-request review button

lib/modules/manager/nuget/util.spec.ts
Otiel: I didn't plan to change existing tests first, but [this job](https://github.com/renovatebot/renovate/actions/runs/11907985528/job/33182852346) forced me to do so. I don't know why it didn't report an error on the previous commits. 🤷

Otiel:
@viceice Sorry to bump again, it's been two weeks since I've requested a review via GitHub, so I'm wondering if you've got the notification?

rarkins (approved):

viceice (approved):

Issue #16884, opened 2022-08-01
Pass `automergeStrategy` via `platformOptions` to `createPr`


viceice:
### What would you like Renovate to be able to do?

We should pass `automergeStrategy` via `platformOptions` to `createPr`, so native automerge feature can use the configured merge method.

### If you have any ideas on how this should be implemented, please tell us here.

yes, add `strategy` to `PlatformPrOptions` interface and pass it here:

https://github.com/renovatebot/renovate/blob/f8c789efb44311d74320865c8cb85abc015b73c4/lib/workers/repository/update/pr/index.ts#L36-L52

### Is this a feature you are interested in implementing yourself?

Maybe

PhilipAbed:
can you point us to which task this is related to?
cause seems to me like half a job with the purpose of `native automerge feature can use the configured merge method.` but not sure where/what how that is happening, any links to other discussions/issues?

viceice:
@joegoldman2 is this really fixed for all platforms?

joegoldman2:
I don't really know but I closed this one because I thought this issue is just about passing `automergeStrategy` via `platformOptions`. If it's also about using it in all supported platforms, we can reopen it and I will try to have a look to see what platforms are missing.

PR #32016, opened 2024-10-18
fix(github)!: change automerge strategy priority order to allow platform signed commits

pballandras:
## Changes

This PR introduces a change in the way Renovate selects an automerge strategy/method for GitHub repositories. Previously it would test which merge strategy the repo allows and choose the first one available in this order in an `if` statement:

1. Rebase
2. Squash
3. Merge

This PR changes this to:

1. Squash
2. Merge
3. Rebase

## Context

Discussion:

- https://github.com/renovatebot/renovate/discussions/32062

### Why is the order important

The reason why the order is important lies in a limitation GitHub has when it comes to verifying/signing commits when merging PRs. More specifically:

> When using the Rebase and Merge option on a pull request, it's important to note that the commits in the head branch are added to the base branch without commit signature verification.

https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#signature-verification-for-rebase-and-merge

Since Renovate uses this method if there's no other configuration to explicitly change it in the repository, by default, it means the PRs Renovate merges won't be signed. Hence, this PR.

Although I don't believe there are changes needed to the doc, if you feel like additions could be made somewhere I didn't think of, I'll gladly contribute.

To test if the reordering would work I disabled some merge strategies to force Renovate into other branches of the `if` statement. With no surprises, the current order is respected and the interesting part is when Renovate squashes or uses merge commits (the merge strategy), the resulting commits are signed. Which is what is we want.


rarkins:
@pballandras thanks for this PR, but please create a Discussion first so we can discuss. I'll mark this PR as draft in the meantime. Please be sure to describe exactly which use case/scenario you find not be working today.

pballandras:
I see you use semantic release and in the discussion you said you wanted this to be in a major release. I believe you can add it in the commit footer during merge, but do you need anything from me at this point? Thanks!

lib/modules/platform/github/index.ts
```diff
@@ -1910,25 +1911,25 @@ export async function mergePr({
...
         try {
-          options.body.merge_method = 'merge';
+          options.body.merge_method = 'rebase';
           logger.debug({ options, url }, `mergePr`);
           automergeResult = await githubApi.putJson(url, options);
         } catch (err3) {
-          logger.debug({ err: err3 }, `Failed to merge commit PR`);
+          logger.debug({ err: err3 }, `Failed to rebase merge PR`);
           logger.info({ pr: prNo }, 'All merge attempts failed');
           return false;
```
pballandras, line 1934: I took the liberty of changing this bit too when I realized the comment was no longer in sync with the code. It now follow the same order (squash -> merge -> rebase)

lib/modules/platform/github/index.ts
```diff
@@ -1910,25 +1911,25 @@ export async function mergePr({
     }
   }
   if (!automerged) {
-    // We need to guess the merge method and try squash -> rebase -> merge
-    options.body.merge_method = 'rebase';
+    // We need to guess the merge method and try squash -> merge -> rebase
```
viceice, line 1914: @rarkins we should use this breaking change to only try the defined/ configured merge method and don't retry other methods. we don't do it on other platforms.

lib/modules/platform/github/index.ts
```diff
@@ -553,12 +553,12 @@ export async function initRepo({
     // Base branch may be configured but defaultBranch is always fixed
     logger.debug(`${repository} default branch = ${config.defaultBranch}`);
     // GitHub allows administrators to block certain types of merge, so we need to check it
-    if (repo.rebaseMergeAllowed) {
-      config.mergeMethod = 'rebase';
-    } else if (repo.squashMergeAllowed) {
+    if (repo.squashMergeAllowed) {
```
viceice, line 556: we should support user configured merge method in same PR, like we do on other platforms. so we unify logic

lib/modules/platform/github/index.ts
rarkins: Are we able to reliably determine it?

pballandras: As far as I understand, yes. This logic is already done during the `initRepo` when we check `if (repo.squashMergeAllowed)` etc etc. The query uses the GraphQL API which has the information on allowed merging methods on the repo (I just checked the [query](https://github.com/renovatebot/renovate/blob/6c1bd70bfd4c528f8fc50b5d24c19a69b4b9a0ac/lib/modules/platform/github/graphql.ts#L2) and the [doc](https://docs.github.com/en/graphql/reference/objects#repository))

lib/modules/platform/github/index.ts
pballandras: I just want to confirm that you are asking me to implement this: https://github.com/renovatebot/renovate/issues/16884

Not against it at all, just don't want to create more noise for you guys if you meant something else.

rarkins: I think it should be done separately / after

lib/modules/platform/github/index.ts
rarkins: BTW this code path which is changed is only executed if config.merge_method is unknown or the merge failed. But then you mention here that it can be reliably determined

pballandras: Yeah, that's on me. You will reliably determine a merge method during the `initRepo`, but only if the token _has the right to know_ (there is a comment mentioning non-admins won't be able to read the merge methods allowed for a repo). Which means, indeed, we can't reliably know.

If this block of code is to be removed, then it would mean Renovate would have to require the set of permissions (which I do not explicitly know sadly) required to know the merge methods allowed. From my experience testing yesterday, the official app already has the rights to this information, but then self-hosted is another thing.

rarkins: Gotcha. In this case we need to keep trying multiple methods for the same reason

lib/modules/platform/github/index.ts
```diff
@@ -1851,6 +1851,7 @@ export async function reattemptPlatformAutomerge({
 export async function mergePr({
   branchName,
   id: prNo,
+  strategy,
```
rarkins, line 1854:
```suggestion
```

This is unused

rarkins (changes requested):
Remove unused param

rarkins:
@pballandras can you accept the change today? We want to release v39 soon.

rarkins (approved):

PR #32024, opened 2024-10-19
feat(datasource/pypi): always fallback to try simple endpoints

### LINKED ISSUE #26483: PyPI datasource fall back to simple API if JSON API fails

rarkins:
## Changes

Modifies pypi datasource so that it _always_ attempts to fall back to a simple result if the JSON API throws an error.

## Context

Built on top of #26386 (thanks @Kakadus) and #31771 (thanks @samgiz).

Closes #26483
Closes #26383

Closes #26483
Closes #26383


samgiz:
Do you need any help from me to get this merged?

lib/modules/datasource/pypi/index.ts
```diff
@@ -66,10 +66,6 @@ export class PypiDatasource extends Datasource {
         // we need to resolve early here so we can catch any 404s and fallback to a simple lookup
         dependency = await this.getDependency(normalizedLookupName, hostUrl);
       } catch (err) {
-        if (err.statusCode !== 404) {
-          throw err;
-        }
-
         // error contacting json-style api -- attempt to fallback to a simple-style api
         logger.trace(
           { packageName, hostUrl },
```
viceice, line 71: we should log the error on trace for troubleshooting

samgiz:
Is anything stopping this from being merged?

viceice (approved):

Kakadus (approved):
nice to see this getting addressed after some time :tada: 

Sadly, I were not able to proceed my initial PR as priorities changed and time ran short. Thank you for yours and for renovate!

PR #32046, opened 2024-10-20
refactor(github): Use matching refs search instead of HEAD requests

zharinov:
## Context

- It could help with #30644

### DIFF
--- a/lib/modules/platform/github/branch.ts
+++ b/lib/modules/platform/github/branch.ts
@@ -1,34 +1,33 @@
-import { HttpError } from '../../../util/http';
-import { Result } from '../../../util/result';
+import { z } from 'zod';
+import { regEx } from '../../../util/regex';
 import { githubApi } from './common';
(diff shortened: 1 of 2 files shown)


rarkins (approved):

lib/modules/platform/github/branch.ts
```diff
@@ -1,34 +1,33 @@
...
-    githubApi.headJson(`/repos/${repo}/git/refs/heads/${branchName}`, {
-      memCache: false,
-    }),
-  )
-    .transform(() => true)
-    .catch((err) => {
-      if (err instanceof HttpError && err.response?.statusCode === 404) {
-        return Result.ok(false);
-      }
+const MatchingRef = z
```
viceice, line 5: we should move it to schema.ts, like we do elsewhere.

viceice (commented):
otherwise LGTM

PR #32069, opened 2024-10-22
feat: calculate `semanticCommitType` priority

### LINKED ISSUE #31979: Add semanticCommitType awareness to generateBranchConfig()

Renovate is not aware of the meaning of values of `semanticCommitType` - they're all just strings. Currently, when a branch has mixed values for `semanticCommitType` then whichever is the first update in the branch is used as the value for branch `semanticCommitType`.

Instead, we could:
- Add awareness of the "importance" of `semanticCommitType` values (e.g. `fix` is more important than `chore`)
- Implement this awareness in https://github.com/renovatebot/renovate/blob/f9ab074bd49de5c3ad25054ed8787c1d4f6a31b9/lib/workers/repository/updates/generate.ts#L162 so that the "most important" `semanticCommitType` value is chosen from the branch, instead of the first.

breadadams:
## Changes

- Introduce logic to `lib/workers/repository/updates/generate.ts` for getting the common `semanticCommitType` based on priority.
- Add unit tests.

## Context

Closes #31979


breadadams:
I've tested this is working by running it on this repo: https://github.com/breadadams/renovate-semantic-commit-type-priority/pulls

If you compare against [the original repro repo](https://github.com/breadadams/renovate-semantic-commit-repro/pulls), you'll see that the new one opens both PRs with the `fix` prefix (as expected).

image

lib/workers/repository/updates/generate.ts
```diff
@@ -355,6 +363,27 @@ export function generateBranchConfig(
     releaseTimestamp: releaseTimestamp!,
   }; // TODO: fixme (#9666)
 
+  let highestPrioritySemanticCommitType = '';
+
+  for (const upgrade of config.upgrades) {
+    if (
+      upgrade.semanticCommitType &&
```
rarkins, line 370:
```suggestion
      upgrade.semanticCommits &&
      upgrade.semanticCommitType &&
```

lib/workers/repository/updates/generate.ts
```diff
@@ -355,6 +363,27 @@ export function generateBranchConfig(
...
+      if (priority > highestPriority) {
+        highestPrioritySemanticCommitType = upgrade.semanticCommitType;
+      }
+    }
+  }
+
+  if (highestPrioritySemanticCommitType) {
+    config.semanticCommitType = highestPrioritySemanticCommitType;
+  }
+
```
rarkins, line 386:
```suggestion
config.semanticCommits = config.upgrades.some((upgrade) => upgrade.semanticCommits));

```

rarkins (changes requested):

lib/workers/repository/updates/generate.ts
```diff
@@ -355,6 +363,29 @@ export function generateBranchConfig(
...
+      if (priority > highestPriority) {
+        highestPrioritySemanticCommitType = upgrade.semanticCommitType;
+      }
+    }
+  }
+
+  if (highestPrioritySemanticCommitType) {
+    config.semanticCommitType = highestPrioritySemanticCommitType;
+  }
+config.semanticCommits = config.upgrades.some((upgrade) => upgrade.semanticCommits));
```
breadadams, line 387:
```suggestion

  config.semanticCommits = config.upgrades.some((upgrade) => upgrade.semanticCommits));
```

lib/workers/repository/updates/generate.ts
breadadams: Hey @rarkins, I applied this suggestion however later realised that `semanticCommits` is a union of strings, not a boolean:

Screenshot 2024-10-22 at 10 22 05

What would you suggest we do here, check if some of them are `=== 'enabled' || === 'auto'`, and if so set it to `enabled`?

rarkins: Yeah, I forgot about that sorry. In short, while you're doing this, you should also make sure that "whether or not to use semantic commits in the first place" is also correctly handled if there's mixed values in upgrades in the same branch.

I'm not certain but this logic might be "Set config.semanticCommits to `enabled` if any upgrade has it set to `enabled`". I'm not sure that `auto` is still present at this point, needs to be checked.

breadadams: Right, from what I see here (if I've followed the code around correctly):

https://github.com/renovatebot/renovate/blob/c7d66d15af867623f6aee42c2a198085bfe59a16/lib/workers/repository/index.ts#L70-L79

The "top-level" auto can still be present at this point, since `detectSemanticCommits()` is called _after_ `extractDependencies()`. However now I'm not sure whether `generateBranchConfig()` should be taking into consideration these top-level config properties, or just the package/branch based options (?).

I'm wondering if the solution should be something like the following (pseudocode):

```
foreach config.upgrades

  if upgrade object contains "semanticCommits"

    if some of the upgrade objects contain "semanticCommits: enabled"
       set config.sematicCommits = "enabled"
       break;
    endif

    if some of the upgrade objects contain "semanticCommits: auto"
       set config.sematicCommits = await detectSemanticCommits()
    endif

  endif

end foreach
```

rarkins: It think you can assume that upgrade.semanticCommits is already set to `enabled` or `disabled`

breadadams: Ok, are you saying no further logic is necessary?

Or that we just need to check for `.some(u => u.semanticCommits === 'enabled')`?

rarkins: Yes, the latter. if any contains 'enabled' then make the branch enabled, otherwise disabled (which it should already be).

breadadams: Should be good now. I've added that logic plus a couple of minor code comments.

breadadams:
Hi @rarkins, is there anything missing here? Sorry for the ping, just wondering if maybe this slipped by. Thanks

rarkins:
I have this saved for follow-up, just unfortunately have had little time recently

lib/workers/repository/updates/generate.ts
```diff
@@ -355,6 +363,35 @@ export function generateBranchConfig(
...
+  if (highestPrioritySemanticCommitType) {
+    config.semanticCommitType = highestPrioritySemanticCommitType;
+  }
+
+  // Enable `semanticCommits` if one of the branches has it enabled
+  if (
+    config.upgrades.some((upgrade) => upgrade.semanticCommits === 'enabled')
+  ) {
+    config.semanticCommits = 'enabled';
+  }
```
rarkins, line 393: You could maybe put this logic first and then only do the other logic if this one is enabled.

rarkins (changes requested):

rarkins (changes requested):

rarkins (approved):

PR #32115, opened 2024-10-24
fix(gerrit): improve commit message vs pr title workaround

felipecrs:
## Changes

This improves the workaround for the Gerrit change title mismatch issue.

The previous approach had big downsides:

1. It would create new patchsets every time the commit message title did not match with the prTitle
   - Creating new patchsets means triggerring unecessary CI jobs, which depending on the project, might be expensive
   - Increases the list of iterations of a change. Some Gerrit deployments will limit the amount of iterations a change can have and if the limit is reached, the change is doomed and needs to be recreated, which requires manual intervention.

   Image:

   description

2. If Renovate for whatever reason fails before updating the commit message to match the PR title, it would mean that the change would be forgot by Renovate forever, and Renovate would attempt to create a new one next time. Also, if this happens and someone closes the change (to ignore future updates), Renovate will also never realize that the change was closed and will keep trying to create a new one.
   - Although it seems rare, it has happened several times with my Renovate deployment. I spent several hours to figure out this was the root cause.

The new approach will always ensure the commit is created or recreated with the proper title. In this case, none of the downsides above will happen. Also, it cleans the code a little, removing gimmicks no longer necessary.

I understand this fix touches a bit of code outside of the Gerrit platform, but I believe it is better enough to at least justify it being merged (when compared to the current approach). If a better approach is found in the future, we can always refactor it.

With this fix, I also believe that the user doesn't even need to be aware of how this is handled. Previously, maybe the user would notice the **second patchset** and thus the documentation would explain this behavior. Now, the user will never notice this behavior to begin with. Obviously, I'm open to suggestions.

## Context

See above.

### DIFF
--- a/lib/modules/platform/gerrit/client.spec.ts
+++ b/lib/modules/platform/gerrit/client.spec.ts
@@ -208,39 +208,6 @@ describe('modules/platform/gerrit/client', () => {
     });
   });
 
-  describe('setCommitMessage()', () => {
-    it('setCommitMessage', async () => {
-      const change = partial<GerritChange>({});
-      httpMock
-        .scope(gerritEndpointUrl)
-        .put('/a/changes/123456/message', { message: 'new message' })
-        .reply(200, gerritRestResponse(change), jsonResultHeader);
-      await expect(client.setCommitMessage(123456, 'new message')).toResolve();
-    });
-  });
-
-  describe('updateChangeSubject', () => {
-    it('updateChangeSubject - success', async () => {
-      const change = partial<GerritChange>({});
-      const newSubject = 'new subject';
-      const previousSubject = 'some subject';
-      const previousBody = `some body\n\nChange-Id: some-change-id\n`;
-      httpMock
-        .scope(gerritEndpointUrl)
-        .put('/a/changes/123456/message', {
-          message: `${newSubject}\n\n${previousBody}`,
-        })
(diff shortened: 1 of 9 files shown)


felipecrs:
@NiasSt90 you may want to take a look at this.

NiasSt90:
> @NiasSt90 you may want to take a look at this.

the changes are looking okay for me (not tried, perhaps next Tuesday i can try it).
I think it's good to get rid of these unnecessary patches. The new workaround is better then my original one.

As long as renovate needs the distinction between prTitle and the first line of the commit message, the only option is such extra handling within the Gerrit platform.

rarkins:
> As long as renovate needs the distinction between prTitle and the first line of the commit message, the only option is such extra handling within the Gerrit platform.

This is a requirement we can probably loosen. Technically you can probably loosen it today, although we say that editing `prTitle` is deprecated. Would it be something to look at before, or part of, or after this PR?

felipecrs:
@rarkins I believe such change can (and I'd prefer it to) be done after this PR.

So long the `prTitle` is important, this PR is still relevant. It retains the current requirements but improves **how**.

To be honest, it doesn't seem anything simple to change such requirement (at least to me).

felipecrs:
I have been hosting Renovate with this PR for three weeks by now. My self-hosted Renovate manages hundreds of changes daily. I can confidently say everything is working as expected.

rarkins (approved):

PR #32148, opened 2024-10-25
feat(pep621): add support for `dependency-groups` (PEP 735)

mkniewallner:
Closes #32147.

## Changes

Add support for Python `dependency-groups`, which recently got accepted as part of [PEP 735](https://peps.python.org/pep-0735/).

## Context

As mentioned in the linked discussion, not all tools that support PEP 621 for defining project dependencies support `dependency-groups` (so far, only uv does AFAIK), but since the feature would only be used by users of tools that support it, it should not be problematic to assume that all tools support it already.

But if you feel like we should take a more defensive approach and only parse dependencies for tools that explicitly support it, happy to update the PR.

Tested over https://github.com/mkniewallner/renovate-uv-dependency-groups.

### DIFF
--- a/lib/modules/manager/pep621/schema.ts
+++ b/lib/modules/manager/pep621/schema.ts
@@ -1,4 +1,3 @@
-import is from '@sindresorhus/is';
 import { z } from 'zod';
 import { LooseArray, LooseRecord, Toml } from '../../../util/schema-utils';
 
--- a/lib/modules/manager/pep621/schema.ts
+++ b/lib/modules/manager/pep621/schema.ts
@@ -88,7 +88,7 @@ export const PyProjectSchema = z.object({
     .record(
       z.string(),
       // Skip non-string entries, like `{include-group = "typing"}`, as they are not dependencies.
-      LooseArray(z.string())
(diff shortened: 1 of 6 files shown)


rarkins (dismissed):

lib/modules/manager/pep621/schema.ts
```diff
@@ -83,6 +84,15 @@ export const PyProjectSchema = z.object({
       'build-backend': z.string().optional(),
     })
     .optional(),
+  'dependency-groups': z
+    .record(
+      z.string(),
+      z
+        .array(z.union([z.string(), z.object({})]))
+        // Skip non-string entries, like `{include-group = "typing"}`, as they are not dependencies.
+        .transform((deps) => deps.filter(is.string)),
```
secustor, line 93: Rather use the `LooseArray` utility

```suggestion
      // Skip non-string entries, like `{include-group = "typing"}`, as they are not dependencies.
      LooseArray(z.string())
```

secustor (changes requested):

secustor (approved):

PR #32151, opened 2024-10-26
test(lib/data): schemas for changelog and source urls

### LINKED ISSUE #29942: Centralize data to lib/data for easier contributing

RahulGautamSingh:
## Changes
- Add json schema and ZodSchema for the changelog-urls.json file
- Add json schema and ZodSchema for the source-urls.json file

## Context
Closes: https://github.com/renovatebot/renovate/issues/29942

### DIFF
--- a/tools/schemas/changelog-urls-schema.json
+++ b/tools/schemas/changelog-urls-schema.json
@@ -5,7 +5,7 @@
     "^[a-zA-Z]+$": {
       "type": "object",
       "patternProperties": {
-        "^[a-zA-Z0-9. -/:]+$": {
+        "^[a-zA-Z0-9. -/:@]+$": {
           "type": "string",
           "format": "uri"
         }
--- a/tools/schemas/changelog-urls-schema.json
+++ b/tools/schemas/changelog-urls-schema.json
@@ -2,7 +2,7 @@
   "$schema": "https://json-schema.org/draft-04/schema#",
   "type": "object",
   "patternProperties": {
-    "^[a-zA-Z]+$": {
+    "^[a-zA-Z-]+$": {
       "type": "object",
       "patternProperties": {
         "^[a-zA-Z0-9. -/:@]+$": {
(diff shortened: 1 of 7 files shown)


tools/schemas/changelog-urls-schema.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "https://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "patternProperties": {
+    "^[a-zA-Z]+$": {
```
viceice, line 5: shouldn't we allow more, depNames can have `-/@` too, eg npm

RahulGautamSingh: This regex is for the datasource. There's another below which is used for the depNames. I have updated it to include @.

viceice: we need `-` for datasource 😁

tools/schemas/source-urls-schema.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "https://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "patternProperties": {
+    "^[a-zA-Z]+$": {
```
viceice, line 5:
```suggestion
    "^[a-zA-Z-]+$": {
```

tools/schemas/changelog-urls-schema.json
RahulGautamSingh, line 5:
```suggestion
    "^[a-zA-Z-]+$": {
```

secustor (approved):

PR #32165, opened 2024-10-27
feat(bitbucket): main branch status for mend app dashboard

setchy:
## Changes

Adds a new optional configuration, `bbMendAppDashboardCheck` which provides a convenient way to quick-launch the Renovate Dashboard for a repository for users of the [Mend App on Bitbucket](https://docs.renovatebot.com/mend-hosted/hosted-apps-config/).

When set, a successful branch status check will be recorded on the repository's main branch, hyperlinking to that repository's Mend Renovate Dashboard page

## Context

Screenshot 2024-10-27 at 3 05 41 PM
Screenshot 2024-10-27 at 3 06 04 PM


setchy:
Of course, a Bitbucket Forge App that adds Renovate as a repository left-hand navigation menu would be preferable, but this is may be a sufficient solution in the near-term

setchy:
I note that the [marketplace listing](https://marketplace.atlassian.com/apps/1232072/mend?tab=overview&hosting=cloud) calls this Mend App.  Will happily rename the above config option and status name/description to best fit

rarkins (commented):
Which of these would you expect?
- Turn it on by default, almost nobody would want it off
- Turn it on by default, but in some scenarios users will want it off
- Turn it off by default, users will need to opt-in

setchy:
> Which of these would you expect?
> - Turn it on by default, almost nobody would want it off
> - Turn it on by default, but in some scenarios users will want it off
> - Turn it off by default, users will need to opt-in

Turn it on by default for Mend App users

lib/modules/platform/bitbucket/index.ts
```diff
@@ -223,6 +224,17 @@ export async function initRepo({
 
     config.defaultBranch = mainBranch;
 
+    if (bbMendAppDashboardStatus) {
+      logger.debug('Creating branch status for Mend App Repository Dashboard');
+      await setBranchStatus({
```
rarkins, line 229: This will set the status every single run. Will it result in duplicates or will the duplicates be ignored server-side by Bitbucket?

And because we also call getStatus() after, it means 2 additional API calls per run even if nothing has changed.

At the very minimum I would like to have repository cache used here to store the last SHA which was set so that we can skip if it's unchanged.

rarkins (changes requested):

lib/modules/platform/bitbucket/index.ts
setchy: It is handled server side, checks are updated in place based on context key. 

I think I'll pivot from this approach that has a bunch of unneeded getStatus and getSha calls and instead simply setStatus with sha=HEAD.

setchy: Updated to reduce unnecessary API calls since we're always setting HEAD branch status

rarkins (changes requested):
We would want to cache the SHA between runs to avoid setting it unnecessarily on future runs. Renovate already ties the repo extract results to defaultBranchSha, so it would be similar to that.

Also, this is in repo config but is done in initRepo() which occurs _before_ repo config is read

setchy:
I'm not 100% following the ask.  I've refactored it to use `HEAD` so that there isn't a need to resolve or keep track of a SHA.

rarkins:
Your approach will still use one API call per repo per run, even if the head SHA is unchanged. I would like to leverage the repo cache so that we use an API call only when the default branch SHA has changed.

lib/modules/platform/bitbucket/index.ts
```diff
@@ -223,6 +224,29 @@ export async function initRepo({
 
     config.defaultBranch = mainBranch;
 
+    if (bbMendAppDashboardStatus) {
+      const statusName = 'Renovate Dashboard';
```
setchy, line 228: Renovate Dashboard or Mend Dashboard?

rarkins: "Mend.io Dashboard" maybe?

lib/modules/platform/bitbucket/index.ts
```diff
@@ -223,6 +224,29 @@ export async function initRepo({
...
+
+      const mendAppDashboardStatus = await getBranchStatusCheck(
```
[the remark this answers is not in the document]
rarkins: Will it cache across runs? I didn't think it did

lib/modules/platform/bitbucket/index.ts
setchy: Updated status name

lib/modules/platform/bitbucket/index.ts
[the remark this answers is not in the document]
setchy: I'm not 100% certain, but i think it does...  testing locally I was using `pnpm cache-clean` in order to test first-time behavior

rarkins:
To clarify, is this the goal?
- If default branch SHA has changed since previous run, set a Renovate Dependency Dashboard check
- If default branch SHA hasn't changed, do no API calls (rely on cache to know)

setchy:
Correct, yes.

rarkins:
Adding @zharinov to help get the caching part added so we can merge

setchy:
closing this PR as it's superseded by https://github.com/renovatebot/renovate/pull/38085

Issue #32040, opened 2024-10-20
Ensure getPrList() results integrity mid-run


rarkins:
### Describe the proposed change(s).

It's important that `platform.getPrList()` returns an accurate list, particularly after PRs are created.

Check each platform to make sure that createPr() and updatePr() both mutate/update any saved PR lists for each platform.

rarkins:
While looking into this, please also check on whether the PR list is cached reliably too.

RahulGautamSingh:
Related Info:

| Platform | getPrList() info stored in | createPr updates stored info | updatePr updates stored info  |
|-|-|-|-|
| Azure | local config | ✅ | ✅ | 
| Bitbucket Cloud | repo cache | ✅ | ❌ | 
| Bitbucket Server | local config | ✅ | ❌ | 
| CodeCommit | local config | ❌  | ❌ | 
| Gerrit | function not used | ❌  | ❌ | 
| Gitea | repo cache | ✅ | ✅ | 
| Github | local config & repo cache | ✅ | ✅ | 
| Gitlab | loca config | ✅ | ❌ | 

Tasks:
- [ ] Add the logic in platforms Bitbucket Cloud & Server, Gitlab such that `updatePr()` function, updates the prList in config or cache. Should we implement the same for codecommit/gerrit they need it but I doubt they are priority?

_Test the following cases for all the platforms:_
- [ ] After pr creation log the prList to check if created pr is available
- [ ] On next run, the created pr should be available in the result of `getPrList()`
- [ ] After pr is updated "by bot" the updated info should be available in the result of `getPrList()`
- [ ] On next run the updated pr should be returned from `getPrList()` 
- [ ] Modify any pr from platform UI. Then run the bot ... now `getPrList()` should return updated info for that modified PR ie. validate cache reconciliation which happens the first time `getPrList()` is called during the run and cache is initialzed

rarkins:
Code Commit can be skipped as it's deprecated. Gerritt also unless it's really quick. Ideally all platforms would update the PR cache correctly too

rarkins:
@RahulGautamSingh it's important to close the gap for Bitbucket Cloud and Azure

RahulGautamSingh:
Should we close this as the main objective is completed? And move the task of adding cache to the Azure, BitbucketServer, Gitlab platforms to another issue.

rarkins:
Please create separate issues then close this one

PR #32192, opened 2024-10-29
refactor(bitbucket-server): update runtime pr list

RahulGautamSingh:
## Changes
Update the pr list stored at runtime with new/updated info when a pr is updated.

## Context
Ref: https://github.com/renovatebot/renovate/issues/32040


lib/modules/platform/bitbucket-server/index.spec.ts
```diff
@@ -1694,16 +1694,22 @@ describe('modules/platform/bitbucket-server/index', () => {
...
-              targetBranch: 'new_base',
-            }),
-          ).toResolve();
+          await bitbucket.updatePr({
+            number: 5,
+            prTitle: 'title',
+            prBody: 'body',
+            targetBranch: 'new_base',
+          });
+          expect(1).toBe(1);
```
viceice, line 1712: why this change? I would prefer the previous one

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1011,22 +1011,25 @@ export async function updatePr({
       };
     }
 
-    const { body: updatedPr } = await bitbucketServerHttp.putJson<{
-      version: number;
-      state: string;
-    }>(
+    const { body: updatedPrRes } = await bitbucketServerHttp.putJson<
+      BbsRestPr & {
```
viceice, line 1015: doesn't this interface already have version and state? 🤔

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1011,22 +1011,25 @@ export async function updatePr({
       };
     }
 
-    const { body: updatedPr } = await bitbucketServerHttp.putJson<{
-      version: number;
-      state: string;
-    }>(
+    const { body: updatedPrRes } = await bitbucketServerHttp.putJson<
```
viceice, line 1014: can you avoid the remaining in this pr?

lib/modules/platform/bitbucket-server/index.spec.ts
RahulGautamSingh: Sorry, got left :)

lib/modules/platform/bitbucket-server/index.ts
RahulGautamSingh: Didn't get it

viceice:
```suggestion
    const { body: updatedPr } = await bitbucketServerHttp.putJson<BbsRestPr>(
```

viceice (commented):
see comment

lib/modules/platform/bitbucket-server/index.ts
RahulGautamSingh: Why though?

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1011,22 +1011,20 @@ export async function updatePr({
       };
     }
 
-    const { body: updatedPr } = await bitbucketServerHttp.putJson<{
-      version: number;
-      state: string;
-    }>(
+    const { body: updatedPr } = await bitbucketServerHttp.putJson<BbsRestPr>(
```
viceice, line 1014:
```suggestion
    const { body: updatedPr } = await bitbucketServerHttp.putJson<BbsRestPr & {
        version: number;
      }>(
```

wonder why `version` is optional. 🤔  

we should switch to zod schema as soon as possible

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1011,22 +1011,20 @@ export async function updatePr({
...
       `./rest/api/1.0/projects/${config.projectKey}/repos/${config.repositorySlug}/pull-requests/${prNo}`,
       { body },
     );
 
-    updatePrVersion(prNo, updatedPr.version);
+    updatePrVersion(prNo, updatedPr.version!);
```
viceice, line 1019:
```suggestion
    updatePrVersion(prNo, updatedPr.version);
```

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1011,22 +1011,20 @@ export async function updatePr({
...
 
     const currentState = updatedPr.state;
     // TODO #22198
     const newState = {
       ['open']: 'OPEN',
       ['closed']: 'DECLINED',
     }[state!];
+    let finalState: string = currentState;
```
viceice, line 1027: why do we need this `string` cast?

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1040,8 +1038,23 @@ export async function updatePr({
...
     }
+
+    if (config.prList) {
+      const bbsPr = utils.prInfo(updatedPr);
+      const existingIndex = config.prList.findIndex(
+        (item) => item.number === prNo,
+      );
+      // istanbul ignore if: should never happen
+      if (existingIndex === -1) {
+        config.prList.push({ ...bbsPr, state: finalState });
```
viceice, line 1053: log an warn / error so someone sees if it happens on runtime

lib/modules/platform/bitbucket-server/index.ts
RahulGautamSingh: its optional in the `BbsRestPr` interface..hence :) Should be changed, is it okay to do in this pr?

lib/modules/platform/bitbucket-server/index.ts
RahulGautamSingh: Nice catch! Thanks... It was a wrong cast and wrong value too.

viceice (approved):

PR #32220, opened 2024-10-30
fix(presets): apply NodeJS docker workaround for private registries

secustor:
## Changes
Changes NodeJS docker image workaround to match any docker image too which ends on `/node` 

### DIFF
--- a/lib/config/presets/internal/workarounds.ts
+++ b/lib/config/presets/internal/workarounds.ts
@@ -235,7 +235,7 @@ export const presets: Record<string, Preset> = {
     packageRules: [
       {
         matchDatasources: ['docker'],
-        matchDepNames: ['node'],
+        matchDepNames: ['/(^|\\/)node$/'],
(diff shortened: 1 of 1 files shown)


viceice (commented):
I remember there is a docker image we need to exclude because it's not nodejs. 🤔

lib/config/presets/internal/workarounds.ts
```diff
@@ -235,7 +235,7 @@ export const presets: Record<string, Preset> = {
     packageRules: [
       {
         matchDatasources: ['docker'],
-        matchDepNames: ['node'],
+        matchDepNames: ['/(^|\\/)node$/'],
```
viceice, line 238: exclude some packages 

https://github.com/renovatebot/renovate/blob/a471762e137619c06e73a678d6b63ca984da7dba/lib/config/presets/internal/group.ts#L352-L355

viceice (changes requested):

viceice (approved):

PR #32222, opened 2024-10-30
feat(hex): Add protobuf definitions for registry v2 support

zharinov:
## Context

- Ref: #29756


lib/modules/datasource/hex/v2/package.proto
```diff
@@ -0,0 +1,55 @@
+// https://github.com/hexpm/specifications/blob/main/registry/package.proto
```
viceice, line 1: should we use a perma link? so we know from when copied?

viceice (dismissed):

rarkins (approved):

PR #32255, opened 2024-10-31
fix(gomod): GOPROXY with slash doesnt work

PhilipAbed:
## Changes

trimm slash while parsing GOPROXY

## Context

some customer run failed with this error
`DEBUG: Goproxy error: trying next URL provided with GOPROXY`

the URL in the error had 2 slashes // after go.
`"url": "https://{XYZ.ARTIFACTORY}/go//x/y/z/v3/@v/list",`

when i checked code for this i found the slash is hardcoded into this

https://github.com/renovatebot/renovate/blob/4bea95b2d005b2d661405a3e4f3bca7861266fc5/lib/modules/datasource/go/releases-goproxy.ts#L132

the simple solution is to trim the slash from GOPROXY in the parser before it, as it is not necessary.

let me know if you think otherwise.

### DIFF
--- a/lib/modules/datasource/go/goproxy-parser.spec.ts
+++ b/lib/modules/datasource/go/goproxy-parser.spec.ts
@@ -12,8 +12,23 @@ describe('modules/datasource/go/goproxy-parser', () => {
       expect(result).toMatchObject([{ url: 'foo' }]);
     });
 
+    it('parses url with trailing slash', () => {
+      const result = parseGoproxy('foo/');
+      expect(result).toMatchObject([{ url: 'foo' }]);
+    });
+
     it('parses multiple urls', () => {
       const result = parseGoproxy('foo,bar|baz,qux');
+      expect(result).toMatchObject([
+        { url: 'foo', fallback: ',' },
+        { url: 'bar', fallback: '|' },
+        { url: 'baz', fallback: ',' },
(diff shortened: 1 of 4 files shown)


rarkins (commented):
Why is a trailing slash a problem if parsed with it? It's valid in URLs obviously

rarkins (changes requested):
I think that this approach to fixing it is not robust. It's done in a separate _file_ to where the problem occurs. We should either use a better way of joining URLs, or do the trailing slash trimming at the point of static joining

PhilipAbed:
so you are saying the issue is with 
 const url = `${baseUrl}/${this.encodeCase(packageName)}/@v/list`; 
 its kind of hardcoded, we should use some URL joiner here

viceice:
> so you are saying the issue is with 
>  const url = `${baseUrl}/${this.encodeCase(packageName)}/@v/list`; 
>  its kind of hardcoded, we should use some URL joiner here

yes, that's better

PhilipAbed:
im afraid to use url joiner since the url.parser and joiner are broken for protocols
https://github.com/renovatebot/renovate/pull/31008/files
https://github.com/renovatebot/renovate/discussions/30295

rarkins (approved):

viceice (approved):

PR #32289, opened 2024-11-02
fix(nix): simplify --extra-experimental-features argument

SuperSandro2000:

### DIFF
--- a/lib/modules/manager/nix/artifacts.ts
+++ b/lib/modules/manager/nix/artifacts.ts
@@ -22,9 +22,7 @@ export async function updateArtifacts({
     return null;
   }
 
-  let cmd = `nix \
-    --extra-experimental-features nix-command \
-    --extra-experimental-features flakes `;
+  let cmd = `nix --extra-experimental-features 'nix-command flakes' `;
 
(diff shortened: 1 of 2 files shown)


rarkins (approved):

lib/modules/manager/nix/artifacts.ts
```diff
@@ -22,9 +22,7 @@ export async function updateArtifacts({
     return null;
   }
 
-  let cmd = `nix \
-    --extra-experimental-features nix-command \
-    --extra-experimental-features flakes `;
+  let cmd = `nix --extra-experimental-features 'nix-command flakes' `;
```
viceice, line 25: I'm not fully sure, but single quotes are failing on windows. you should use `quote` from `shlex` instead

viceice (changes requested):

lib/modules/manager/nix/artifacts.ts
SuperSandro2000: I don't think that matters much right now as nix' windows support is pretty bare bones and only compiles with some unmerged changes.

PR #32339, opened 2024-11-05
feat(bitbucket): Add more logging to the PR cache

### LINKED ISSUE #32329: Improve Bitbucket getPrList() verbosity

Lower this to trace as it's called often: https://github.com/renovatebot/renovate/blob/85267b9861cffc680a5efc4ae8266791232bffa7/lib/modules/platform/bitbucket/index.ts#L286

Add debug logs to https://github.com/renovatebot/renovate/blob/85267b9861cffc680a5efc4ae8266791232bffa7/lib/modules/platform/bitbucket/pr-cache.ts#L59-L66 and subfunctions for any thing which is run _once_ (no need to repeat if getPrList() is called repeatedly in a run)

zharinov:
## Context

- Closes #32329

### DIFF
--- a/lib/modules/platform/bitbucket/pr-cache.ts
+++ b/lib/modules/platform/bitbucket/pr-cache.ts
@@ -23,8 +23,15 @@ export class BitbucketPrCache {
     let pullRequestCache = repoCache.platform.bitbucket.pullRequestsCache as
       | BitbucketPrCacheData
       | undefined;
-    if (!pullRequestCache || pullRequestCache.author !== author) {
+    if (!pullRequestCache) {
       logger.debug('Initializing new PR cache at repository cache');
+      pullRequestCache = {
+        items: {},
+        updated_on: null,
+        author,
+      };
+    } else if (pullRequestCache.author !== author) {
+      logger.debug('Resetting PR cache because authors do not match');
       pullRequestCache = {
         items: {},
         updated_on: null,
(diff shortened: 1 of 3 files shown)


lib/modules/platform/bitbucket/pr-cache.ts
```diff
@@ -24,6 +24,7 @@ export class BitbucketPrCache {
       | BitbucketPrCacheData
       | undefined;
     if (!pullRequestCache || pullRequestCache.author !== author) {
+      logger.debug('Initializing new PR cache at repository cache');
```
rarkins, line 27: Let's log these two separate cases

lib/modules/platform/bitbucket/pr-cache.ts
```diff
@@ -135,7 +137,13 @@ export class BitbucketPrCache {
       cacheProvider: repoCacheProvider,
     };
     const res = await http.getJson<PagedResult<PrResponse>>(url, opts);
-    this.reconcile(res.body.values);
+
+    const items = res.body.values;
+    logger.debug(`Fetched ${items.length} PRs to sync with cache`);
+
+    this.reconcile(items);
```
rarkins, line 144: If we have a faulty cache (somehow), and we are not self-healing through the reconcile command, would this logging be enough to catch it?

rarkins (changes requested):

lib/modules/platform/bitbucket/pr-cache.ts
```diff
@@ -135,7 +145,22 @@ export class BitbucketPrCache {
...
+
+    const oldCache = clone(this.cache.items);
```
viceice, line 154: needs to be before reconcile, otherwise it's the already updated cache.

viceice (approved):

rarkins (approved):

PR #32356, opened 2024-11-06
docs: document hostRule needed for some self-hosted Gitea instances

Xe138:
## Changes

Document the need for an additional hostRule for some self-hosted Gitea configurations.

## Context

- #32311 

### DIFF
--- a/lib/modules/platform/gitea/readme.md
+++ b/lib/modules/platform/gitea/readme.md
@@ -28,6 +28,21 @@ The PAT should have these permissions:
 
 If you use Gitea packages, add the `read:packages` scope.
 
+For Gitea instances configured with a base URL that differs from the Renovate `endpoint` (for example, where Gitea is configured with an externally accessible base URL, but Renovate is configured to access the Gitea instance locally.), Renovate may attempt to access repositories using the base URL.  In this case it may be necessary to configure a `hostRule` to use the PAT configured above for host authentication. (see [hostRules](https://docs.renovatebot.com/configuration-options/#hostrules))
+
+An example of a hostRule for an external domain with a token:
+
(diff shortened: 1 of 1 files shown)


lib/modules/platform/gitea/readme.md
```diff
@@ -28,6 +28,21 @@ The PAT should have these permissions:
 
 If you use Gitea packages, add the `read:packages` scope.
 
+For Gitea instances configured with a base URL that differs from the Renovate `endpoint` (for example, where Gitea is configured with an externally accessible base URL, but Renovate is configured to access the Gitea instance locally.), Renovate may attempt to access repositories using the base URL.  In this case it may be necessary to configure a `hostRule` to use the PAT configured above for host authentication. (see [hostRules](https://docs.renovatebot.com/configuration-options/#hostrules))
```
rarkins, line 31: The first statement makes it sound like Renovate's fault/problem, but it's not. Gitea is responding with Link header pointing to a different host, and Renovate is following it.

If Renovate can access that other URL, why not use it as the `endpoint` to save this complexity?

Xe138: It's expected behavior that Gitea is responding using the externally accessible URL.  It's not clear to me why Renovate would follow that link after making a successful request using its configured endpoint.  There may be a good reason, but that's why this documentation is helpful - to explain a potentially unintuitive behavior.

The reason for not using the externally accessible endpoint for Renovate is it adds an external network dependency to the deployment of Gitea + Renovate that would otherwise not be required.

rarkins: PR won't be merged unless this is fixed

PR #32376, opened 2024-11-06
fix: Avoid rejecting releases when `extractVersion` is used

zharinov:
## Changes

We don't preserve original versions, so it's impossible to provide them as the additional  information to the postprocess phase.

There is two options:
- preserve such information and provide it to the downstream code
- don't be so strict about rejected versions

This PR implements the latter approach since it's the simplest one.


lib/modules/datasource/postprocess-release.ts
```diff
@@ -49,6 +49,20 @@ export async function postprocessRelease(
     );
 
     if (result === 'reject') {
+      if (config.extractVersion) {
+        logger.debug(
```
rarkins, line 53: Trace maybe?

zharinov: Don't know, I think it's rare combination, so maybe it's good to have it at the debug level

rarkins (approved):

PR #32384, opened 2024-11-07
fix(composer): resolve the `lockFileMaintenance` not working

kayw-geek:
## Changes

Fixed an issue where the `lockFileMaintenance` configuration was not working for the Composer manager. 
Currently, triggering `LockFileMaintenance` via webhook does not perform any actions.


rarkins:
Do you have a reproduction repo to demonstrate this before/after?

kayw-geek:
Hi @rarkins 
This repository can reproduce the current behavior of the Composer manager. 
The expected behavior is that composer’s `lockFileMaintenance` should execute `composer update`, after which the `composer.lock` file should change, and the version of `guzzlehttp/guzzle` should update from `7.9.0` to the latest `7.9.2`.

viceice (changes requested):
usally both ways should work. most managers are using the `updateType`

https://github.com/renovatebot/renovate/blob/dc7b8d9d12be52ebeae1f8fe8d9d88b250f99be3/lib/workers/repository/updates/flatten.ts#L150-L151

@rarkins should we check all managers to be consistent?

viceice:
maybe related to the grouping?
https://github.com/kayw-geek/24958/blob/6a7804b2030da6dcdc53bb71f39aa31378bfa2a5/renovate.json#L17

kayw-geek:
> maybe related to the grouping? https://github.com/kayw-geek/24958/blob/6a7804b2030da6dcdc53bb71f39aa31378bfa2a5/renovate.json#L17

Should I delete this groupName?

rarkins:
> @rarkins should we check all managers to be consistent?

Yes, I suspect there may be some inconsistency here. And we should pay attention to the case where people group lock file maintenance with regular updates. It's not something which we recommend, and it will have different behavior per manager, but many people do it

viceice:
> > @rarkins should we check all managers to be consistent?
> 
> Yes, I suspect there may be some inconsistency here. And we should pay attention to the case where people group lock file maintenance with regular updates. It's not something which we recommend, and it will have different behavior per manager, but many people do it

We should issue a warning if lockfile maintenance is grouped with other updates, so users are aware that it's not always working as intended.

I think i know why it's not working here:
First update is a regular update, so `isLockFileMaintenance` is not copied to branch config, but somehow the `updateType` is copied. 🤔

kayw-geek:
Hi @rarkins @viceice ,

Just wanted to check if there’s any update on the PR, Please let me know if you need anything from my side to move it forward.

kayw-geek:
Hi @rarkins @viceice, 

I would like to confirm if this is a more appropriate approach for composer update command. 
Currently, we use the `composer update packageName:packageVersion` format to update dependencies. However, if a package version is locked in `composer.json` (e.g., `foo:1.0.0`), when **Renovate** detects version `1.1.0`, it attempts to execute the following command to upgrade: `composer update foo:1.1.0 --with-dependencies`. In reality, this command will fail and produce the following error:

```
The temporary constraint "1.1.0" for "foo" must be a subset of the constraint in your composer.json (1.0.0)
Run `composer require foo` or `composer require foo:1.1.0` instead to replace the constraint
```

If my conclusion is correct, should we update the logic to use `composer require` instead, and depending on the `depType`, decide whether to add the `--dev` flag?

kayw-geek:
hi @rarkins  @viceice,
I would like to confirm if this is a more appropriate approach for Composer update command. 

Currently, we use the `composer update packageName:packageVersion` format to update dependencies. However, if a package version is locked in `composer.json` (e.g., `foo:1.0.0`), when **Renovate** detects version `1.1.0`, it attempts to execute the following command to upgrade: `composer update foo:1.0.0 --with-dependencies`. In reality, this command will fail and produce the following error:

```
The temporary constraint "1.1.0" for "foo" must be a subset of the constraint in your composer.json (1.0.0)
Run `composer require foo` or `composer require foo:1.1.0` instead to replace the constraint
```

This situation may also happen when updating the major version.

If my conclusion is correct, should we update the logic to use `composer require` instead, and depending on the `depType`, decide whether to add the `--dev` flag?

viceice:
> hi @rarkins  @viceice,
> I would like to confirm if this is a more appropriate approach for Composer update command. 
> 
> Currently, we use the `composer update packageName:packageVersion` format to update dependencies. However, if a package version is locked in `composer.json` (e.g., `foo:1.0.0`), when **Renovate** detects version `1.1.0`, it attempts to execute the following command to upgrade: `composer update foo:1.0.0 --with-dependencies`. In reality, this command will fail and produce the following error:
> 
> ```
> The temporary constraint "1.1.0" for "foo" must be a subset of the constraint in your composer.json (1.0.0)
> Run `composer require foo` or `composer require foo:1.1.0` instead to replace the constraint
> ```
> 
> This situation may also happen when updating the major version.
> 
> If my conclusion is correct, should we update the logic to use `composer require` instead, and depending on the `depType`, decide whether to add the `--dev` flag? 
> 
> Alternatively, should we consider modifying the version constraint in composer.json before running the update command to avoid this issue altogether?
> 

this seems to be a change in recent composer versions, can you pinpoint the version which breaks renovate?

I think we can change the `composer.json` before running composer, like we do for other managers.

viceice (commented):
see other comment.

kayw-geek:
> see other comment.

The feature introduces seems like on this [commit](https://github.com/composer/composer/commit/436a112651a83045bca8c396ee219b83dc7f2122)

rarkins (changes requested):
I think this will result in a change of behavior for when lock file maintenance is grouped together with other updates.

Before, `config.isLockFileMaintenance` will be `true` if ANY of the updates have updateType=lockFileMaintenance.

Now with these changes, config.updateType will equal `lockFileMaintenance` only if the first update in a branch has that.

Also, using `config.updateType` is not semantically correct because branches do not have updateTypes. Branches have one or more updates, which each have updateType fields.

lib/modules/manager/gradle/artifacts.ts
```diff
@@ -205,7 +206,7 @@ export async function updateArtifacts({
         .join(' ')}`;
 
       if (
-        config.isLockFileMaintenance === true ||
+        config.updateType === 'lockFileMaintenance' ||
```
rarkins, line 209: Should this have an `|| config.isLockFileMaintenance` ?

lib/modules/manager/nuget/artifacts.ts
```diff
@@ -138,7 +138,11 @@ export async function updateArtifacts({
   }
 
   try {
-    if (updatedDeps.length === 0 && config.isLockFileMaintenance !== true) {
+    if (
+      updatedDeps.length === 0 &&
+      (config.updateType !== 'lockFileMaintenance' ||
+        config.isLockFileMaintenance)
```
rarkins, line 144:
```suggestion
      (config.updateType !== 'lockFileMaintenance' &&
        config.isLockFileMaintenance !== true)
```

Shouldn't it be this?

rarkins (changes requested):

rarkins (dismissed):

kayw-geek:
Hi @rarkins,
Thank you for your review! I have fixed the lint job failure

rarkins (dismissed):

lib/modules/manager/bundler/artifacts.ts
```diff
@@ -93,7 +93,10 @@ export async function updateArtifacts(
 
     const commands: string[] = [];
 
-    if (config.isLockFileMaintenance) {
+    if (
+      config.updateType === 'lockFileMaintenance' ||
```
viceice, line 97: as told, this doesn't work for groups because this field only reflects first update in group which isn't always the lockfile maintenance 


```suggestion
      updatedDeps.some(dep => dep.updateType === 'lockFileMaintenance') ||
```
you need to check the updated deps

viceice (changes requested):

rarkins (approved):

kayw-geek:
Hi @viceice, Do you have time to help approve it?

lib/modules/manager/bundler/artifacts.ts
```diff
@@ -93,7 +93,10 @@ export async function updateArtifacts(
 
     const commands: string[] = [];
 
-    if (config.isLockFileMaintenance) {
+    if (
+      updatedDeps.some((dep) => dep.updateType === 'lockFileMaintenance') ||
+      config.isLockFileMaintenance
+    ) {
```
viceice, line 99: @rarkins i think i'll fix it by properly fill `isLockFileMaintenance` on grouping, which is the only case where this can be false because of the first update in group isn't a lockfile update

viceice: - #33971

PR #32428, opened 2024-11-09
refactor(platform/gitlab): fix api reponse type in `createPr` fn

RahulGautamSingh:
## Changes
Use correct return type in the `createPr` function. And fix the tests + snaps accordingly.

## Context
Instead of using `Pr & { iid: number}` which works but is wrong cause many of the corresponding field in the actual repsonse from api has different names. Eg. `source_branch` in response but `sourceBranch` in type `Pr`. 

This also means that we have been saving the pr info in wrong/incorrect form in the runtime config list as well as the snaps are based on this wrong structure which is not good.


lib/modules/platform/gitlab/index.spec.ts
```diff
@@ -3110,7 +3166,6 @@ describe('modules/platform/gitlab/index', () => {
           number: 1,
           prTitle: 'title',
           prBody: 'body',
-          state: 'closed',
```
secustor, line 3113: This shouldn't be missing

secustor: Where is the switch from `closed` and `open` coming from?

viceice (commented):
please split test and code refactoring to separate PR so the risk for regression is much lower

lib/modules/platform/gitlab/index.spec.ts
RahulGautamSingh: It should not be closed the test description says add/remove labels. No, logic to add/remove labels & close a pr so I changed it to `open`.

RahulGautamSingh:
> please split test and code refactoring to separate PR so the risk for regression is much lower

For `createPr` function the types used are wrong as they do not match the response we get from the api, hence the test and snaps are wrong too. Cannot correct one without disturbing the other.

I have split the `getPr` and `fetchPrList` refactoring in [another PR](https://github.com/renovatebot/renovate/pull/32566) which do not affect any tests. But `createPr` will have to get both code and test refactor together.

viceice (commented):
blocked

viceice (approved):

PR #32432, opened 2024-11-10
feat(manager/rust-toolchain): Add support for rust toolchain versioning

kyle-leonhard:
## Changes

This PR adds a new manager that manages Rust toolchain versions in `rust-toolchain.toml` files based on releases in the [rust-lang/rust](https://github.com/rust-lang/rust) github repository. The relevant Rust docs are here: https://rust-lang.github.io/rustup/overrides.html

## Context

Renovate supports updating Cargo crates but not the Rust toolchain itself. This has been discussed here, for example: https://github.com/renovatebot/renovate/discussions/24762.

Rust allows for the toolchain version (e.g. 1.82.0) to be specified as an override in the `channel` key in a `rust-toolchain.toml` file ([docs](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file)). The version may be specified either as a fully qualified version (e.g. 1.82.0) a major.minor version (e.g. 1.82) or as a named release (e.g. beta, nightly, stable, nightly-2014-11-15). This PR ignores named releases.

https://github.com/rust-lang/rust releases are the datasource.


docs/usage/rust.md
```diff
@@ -52,3 +52,18 @@ module.exports = {
...
+## How it works
+
+1. Renovate searches in each repository for any `rust-toolchain.toml` files that apply [overrides](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file)
+1. Renovate extracts existing the version from the `channel` key/value, ignoring any `nightly`, `beta` or `stable` release references (e.g. "stable", "nightly-2014-12-18")
+1. Renovate resolves the Rust toolchain version using the [rust-lang/rust](https://github.com/rust-lang/rust) Github repositories releases.
+1. If Renovate finds an update, Renovate will update the version in `rust-toolchain.toml`
+
+## Enabling Rust Modules Updating
+
+TODO: I'm not sure whether it's ok to enable a new manager by default.
```
rarkins, line 69: Is this something different than Cargo modules?

lib/modules/manager/api.ts
rarkins, line 1: I think it might be better to rename this `rust-toolchain`

lib/modules/versioning/rust/index.ts
```diff
@@ -0,0 +1,24 @@
...
+export const id = 'rust';
+export const displayName = 'Rust';
+export const urls = [
+  'https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file',
+];
+export const supportsRanges = false;
+
+export class RustVersioningApi extends RegExpVersioningApi {
+  // Format described in https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file
+  // Examples: 1.82.1, 1.82
```
rarkins, line 13: Does 1.82 mean 1.82.0? i.e. it's traditional to drop the patch .0?

lib/modules/manager/rust/readme.md
```diff
@@ -0,0 +1 @@
+This will maintain `rust-toolchain.toml` files. Available versions will be determined from the official Rust Github repository.
```
rarkins, line 1:
```suggestion
This will maintain `rust-toolchain.toml` files.
Available versions will be determined from the official Rust Github repository.
```

One line per sentence

lib/modules/versioning/rust/index.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    ({ version, major, minor, patch }) => {
+      expect(versioning.getMajor(version)).toBe(major);
+      expect(versioning.getMinor(version)).toBe(minor);
+      expect(versioning.getPatch(version)).toBe(patch);
+    },
+  );
+
+  it.each`
+    version      | other        | expected
+    ${'1.82.0'}  | ${'1.82.0'}  | ${true}
```
rarkins, line 51:
```suggestion
    ${'1.82'}    | ${'1.82.0'}  | ${true}
    ${'1.82.0'}  | ${'1.82.0'}  | ${true}
```

lib/modules/versioning/rust/index.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+  `(
+    'matches("$version", "$other") === $expected',
+    ({ version, other, expected }) => {
+      expect(versioning.matches(version, other)).toBe(expected);
+    },
+  );
+
+  it.each`
+    version     | other       | expected
+    ${'1.82.0'} | ${'1.83.0'} | ${false}
```
rarkins, line 81:
```suggestion
    ${'1.82.1'} | ${'1.82'}   | ${true}
    ${'1.81.1'} | ${'1.82'}   | ${false}
    ${'1.82.0'} | ${'1.83.0'} | ${false}
```

lib/modules/manager/rust/index.ts
```diff
@@ -0,0 +1,14 @@
+import type { Category } from '../../../constants';
+import { GithubReleasesDatasource } from '../../datasource/github-releases';
+import * as rustVersioning from '../../versioning/rust';
+
+export { extractPackageFile } from './extract';
+
+export const supportedDatasources = [GithubReleasesDatasource.id];
+
+export const defaultConfig = {
+  fileMatch: ['(^|/)rust-toolchain\\.toml?$'],
```
rarkins, line 10:
```suggestion
  fileMatch: ['(^|/)rust-toolchain(\\.toml)?$'],
```

Isn't it the entire .toml extension which should be optional like this?

lib/modules/manager/rust/extract.ts
```diff
@@ -0,0 +1,19 @@
...
+    depName: 'rust',
+    packageName: 'rust-lang/rust',
+    currentValue: content.trim(),
+    datasource: GithubReleasesDatasource.id,
+  };
+  return {
+    deps: [dep],
+    // Format described in https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file
+    // Examples: 'channel = "1.82.1"', 'channel = "1.82"'
+    matchStrings: ['channel\\s*=\\s*"(?<currentValue>\\d+\\.\\d+(\\.\\d+)?)"'],
```
rarkins, line 17:
```suggestion
```

lib/modules/manager/rust/extract.ts
```diff
@@ -0,0 +1,19 @@
+import { logger } from '../../../logger';
+import { GithubReleasesDatasource } from '../../datasource/github-releases';
+import type { PackageDependency, PackageFileContent } from '../types';
+
+export function extractPackageFile(content: string): PackageFileContent {
+  logger.trace('rust.extractPackageFile()');
+  const dep: PackageDependency = {
+    depName: 'rust',
+    packageName: 'rust-lang/rust',
+    currentValue: content.trim(),
```
rarkins, line 10: This will include the entire contents of the (toml) file as currentValue, which I think is incorrect

lib/modules/manager/rust/extract.ts
```diff
@@ -0,0 +1,19 @@
...
+    datasource: GithubReleasesDatasource.id,
```
rarkins, line 11: Does the rust toolchain install directly from github releases? e.g. a second after a new release is published on github.com, it's possible to install it using rust toolchain?

rarkins (changes requested):

docs/usage/rust.md
```diff
@@ -52,3 +52,18 @@ module.exports = {
...
+## How it works
+
+1. Renovate searches in each repository for any `rust-toolchain.toml` files that apply [overrides](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file)
+1. Renovate extracts existing the version from the `channel` key/value, ignoring any `nightly`, `beta` or `stable` release references (e.g. "stable", "nightly-2014-12-18")
+1. Renovate resolves the Rust toolchain version using the [rust-lang/rust](https://github.com/rust-lang/rust) Github repositories releases.
+1. If Renovate finds an update, Renovate will update the version in `rust-toolchain.toml`
+
+## Enabling Rust toolchain updating
+
+TODO: I'm not sure whether it's ok to enable a new manager by default.
```
rarkins, line 69: It's OK for it be default if it's simple enough to be considered "reliable" from day one

lib/modules/versioning/rust/index.ts
rarkins: This means that it's to be treated as a range and not as a version

lib/modules/versioning/rust/index.ts
```diff
@@ -0,0 +1,201 @@
...
+export const supportsRanges = true;
+export const supportedRangeStrategies: RangeStrategy[] = [
+  'bump',
+  'pin',
+  'replace',
+];
+
+// Format described in https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file
+// Versions may be fully qualified: <major>.<minor>.<patch> (e.g. 1.82.1), indicating an exact version.
+// Or version may be inexact: <major>.<minor> (e.g. 1.82), allowing patch level changes.
```
rarkins, line 22: Where did you find the inexact definition? On that page I see the definition as:

```
<channel>[-<date>]

<channel>       = stable|beta|nightly|<major.minor.patch>
<date>          = YYYY-MM-DD
```

This doesn't indicate patch as optional

lib/modules/manager/rust/extract.ts
```diff
@@ -0,0 +1,26 @@
+import { regEx } from '../../../util/regex';
+import { GithubReleasesDatasource } from '../../datasource/github-releases';
+import { regexMatchAll } from '../custom/regex/utils';
+import type { PackageDependency, PackageFileContent } from '../types';
+
+const channelRegex = regEx(
+  /channel\s*=\s*"(?<currentValue>\d+\.\d+(\.\d+)?)"/,
+  'g',
+);
```
rarkins, line 9: My preference: always capture the channel= line, no matter what the value is. Then let versioning take care of whether it's something we can upgrade or not. At least users will know that we found it, and it's not upgradeable

rarkins (changes requested):

lib/modules/manager/rust/extract.ts
rami3l: @kyle-leonhard As far as rustup is concerned, we use https://static.rust-lang.org/dist/rust-1.80-beta-aarch64-apple-darwin.tar.gz.sha256 as the single source of truth (replace `1.80-beta` and `aarch64` according to your situation.

rami3l: @kyle-leonhard As a matter of fact, we do GitHub tagging after other steps (https://github.com/rust-lang/promote-release/blob/110ad45f13ac5f0aa660acc218c090c6b24f14c8/src/main.rs#L246-L263), so I guess it's a reasonable reference to be used here as well.

lib/modules/manager/rust-toolchain/extract.ts
```diff
@@ -0,0 +1,23 @@
...
+  const deps: PackageDependency[] = [];
+  for (const match of regexMatchAll(channelRegex, content)) {
+    const groups = match.groups;
+    if (groups) {
+      const { currentValue } = groups;
+      deps.push({
+        depName: 'rust',
+        packageName: 'rust-lang/rust',
+        currentValue,
+        datasource: GithubReleasesDatasource.id,
```
viceice, line 18: are all binaries available when they publish the GitHub releases? or should we better depend on official docker images?

lib/modules/manager/rust-toolchain/extract.spec.ts
```diff
@@ -0,0 +1,117 @@
+import { extractPackageFile } from '.';
+
+describe('modules/manager/rust-toolchain/extract', () => {
+  describe('extractPackageFile()', () => {
+    it('returns result when fully qualified version is set', () => {
+      const res = extractPackageFile(`
```
viceice, line 6: use `codeBlock` template helper for proper indention

lib/modules/manager/rust-toolchain/extract.ts
```diff
@@ -0,0 +1,23 @@
+import { regEx } from '../../../util/regex';
+import { GithubReleasesDatasource } from '../../datasource/github-releases';
+import { regexMatchAll } from '../custom/regex/utils';
+import type { PackageDependency, PackageFileContent } from '../types';
+
+const channelRegex = regEx(/channel\s*=\s*"(?<currentValue>.+)"/, 'g');
+
+export function extractPackageFile(content: string): PackageFileContent {
+  const deps: PackageDependency[] = [];
+  for (const match of regexMatchAll(channelRegex, content)) {
```
viceice, line 10: theres only one occurrence, so need of match all

I would prefer to use our toml parser and zod schema validation. so we don't have false deps when they extend the format

withinfocus:
@kyle-leonhard do you need any assistance completing the requested changes? Bitwarden could benefit from using this so we'd like to help getting it merged.

rarkins:
Closing PR, as it's no longer being worked on. Others are welcome to take it up

PR #32437, opened 2024-11-10
fix(presets): switch kiota monorepo to group

melotic:
## Changes

Added other kiota repositories to the `kiota` monorepo preset.

## Context

#32164 added the `kiota` repository, but it also spans multiple repos for each language it supports.

### DIFF
--- a/lib/config/presets/internal/group.ts
+++ b/lib/config/presets/internal/group.ts
@@ -293,6 +293,46 @@ const staticGroups = {
       },
     ],
   },
+  kiota: {
+    description: 'Group Kiota packages together.',
+    packageRules: [
+      {
+        description: 'Groups Kiota Node.JS packages together.',
+        groupName: 'Kiota',
+        groupSlug: 'kiota-node',
+        matchDatasources: ['npm'],
+        matchPackageNames: ['@microsoft/kiota**'],
+      },
+      {
+        description: 'Groups Kiota .NET packages together.',
+        groupName: 'Kiota',
+        groupSlug: 'kiota-dotnet',
+        matchDatasources: ['nuget'],
+        matchPackageNames: ['Microsoft.Kiota**', 'Microsoft.OpenApi.Kiota**'],
+      },
+      {
+        description: 'Groups Kiota Java packages together.',
+        groupName: 'Kiota',
+        groupSlug: 'kiota-java',
+        matchDatasources: ['maven'],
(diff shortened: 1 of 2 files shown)


rarkins (changes requested):
Are the versions between repos kept in-sync? If not then this is probably not right for a monorepo grouping and instead might be part of group:recommended, or just a standalone grouping of its own

melotic:
> Are the versions between repos kept in-sync? If not then this is probably not right for a monorepo grouping and instead might be part of group:recommended, or just a standalone grouping of its own

They are not, only the version in each repo are in sync. I'll port this over to `group:recommended`.

viceice (changes requested):
I don't like this new default grouping.

those packages should be grouped by their monorepo when the repo published them with the same version.

the other groups should be fully optional and opt-in, because they create immortal PRs.

melotic:
> I don't like this new default grouping.
> 
> those packages should be grouped by their monorepo when the repo published them with the same version.
> 
> the other groups should be fully optional and opt-in, because they create immortal PRs.

Is what you're saying define multiple monorepos for each Kiota repo and an optional group to group all the monorepos? That makes sense. 

For context, each kiota repo publishes multiple packages, versioned together, but the versions are not the same across repos. (_i.e_., the .NET packages are at `1.14.0`, while the NPM ones are at `1.0.0-preview.75`)

viceice:
i think most user are only using one tech stack, eg npm or dotnet on a single repo

melotic:
Hard to say exactly, Kiota is fiarly new. I would guesstimate that 50% of users have a dependency on the Kiota library itself (`microsoft/kiota`) to generate OpenAPI schemas & bindings, and then those bindings have a dependency on say `microosft/kiota-typescript` or `microsoft/kiota-dotnet`. The other half are just consumers without a need to generate bindings.

Most would only have one language like @viceice mentioned for the bindings. In my case, it's a C# backend with a TypeScript frontend.

There's an argument that any updates to the generation code (`microsoft/kiota`) should also cause an update to the code generated `microsoft/kiota-typscript`. 

But I as a user would be okay with the default behavior for 2 PRs, one for the actual generation kiota libraries, and another for the bindings.

astellingwerf:
Doesn't this change technically demand for a migration step? Folks might reference "monorepo:kiota", but it won't resolve any longer.

PR #32480, opened 2024-11-12
feat(versioning): Add a NodeJS Lambda Versioning module

Garethp:
## Changes

This PR adds a versioning module for NodeJS Lambda Runtimes provided by AWS

## Context

As discussed in [this discussion](https://github.com/renovatebot/renovate/discussions/32313), I want to keep versions of Node in with the release schedule of AWS Lambdas. AWS NodeJS Lambdas release within a month or two after a NodeJS version goes LTS so there is a period of time where a version of Node is LTS but not available as a Lambda Runtime. To keep a project's node dependency on the same version as the Runtime it will run on, I need to be able to set those dependencies (nvmrc, docker etc...) to follow the Lambda Runtime release schedule rather than the NodeJS release schedule.

Since this is meant to be a drop-in replacement for any packages that follow `node` versioning, I've re-used most of the functions provided by the `node` versioning module and only changed the `isStable` function.


viceice (commented):
needs code coverage 

https://github.com/renovatebot/renovate/actions/runs/11800328119/job/33317532023#step:6:22

Garethp:
@viceice, I've removed the unused code and changed how I'm mocking the data. I've also mocked out `dataFiles` so that I can test some of the if conditions that are based on unstable data from the datasource. Let me know if this is the appropriate way of testing that instability.

Garethp:
@rarkins, @viceice, could I please get a re-review on this and an approval to run the tests for this MR?

viceice:
you should use the re-request review button

tools/static-data/generate-lambda-node-schedule.mjs
```diff
@@ -0,0 +1,44 @@
...
+ * @property {string} cycle - The ID of the Runtime.
+ * @property {boolean|string} support - Either `true` if in support or a string denoting when support for this Runtime
+ *                                      will end. 0.10.x is a sole exception which has `false` and will be filtered out.
+ */
+
+const lambdaDataUrl = 'https://endoflife.date/api/aws-lambda.json';
+
+await (async () => {
+  console.log('Generating node schedule');
+  const { body } = await got(lambdaDataUrl);
```
viceice, line 16: use builtin fetch, see

https://github.com/renovatebot/renovate/blob/6dc2b37be1c3a3f8924ecb45caa850a877cc5631/tools/static-data/generate-node-schedule.mjs#L8

tools/static-data/generate-lambda-node-schedule.mjs
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * @type Array<RuntimeDefinition>
+   */
+  const lambdas = JSON.parse(body);
+  const nodeRuntimes = lambdas
+    // Filter Runtimes down to only NodeJS Runtimes
+    .filter((lambda) => lambda.cycle.startsWith('nodejs'))
+    // The only Runtime where support is not either `true` or a Date as a string is `0.10.x`, which we don't need
+    .filter((lambda) => lambda.support !== false)
+    .reduce((schedule, lambda) => {
```
viceice, line 27: use for-of loop

package.json
```diff
@@ -58,7 +58,8 @@
     "type-check": "run-s 'generate:*' 'tsc --noEmit {@}' --",
     "update-static-data": "run-s 'update-static-data:*'",
     "update-static-data:distro-info": "node tools/static-data/generate-distro-info.mjs",
-    "update-static-data:node-schedule": "node tools/static-data/generate-node-schedule.mjs"
+    "update-static-data:node-schedule": "node tools/static-data/generate-node-schedule.mjs",
+    "update-static-data:lambda-node-schedule": "node tools/static-data/generate-lambda-node-schedule.mjs"
```
viceice, line 62: sort alphanumeric

lib/modules/versioning/node/index.ts
```diff
@@ -53,7 +53,7 @@ export function isStable(version: string): boolean {
     const schedule = findScheduleForVersion(version);
     if (schedule?.lts) {
       // TODO: use the exact release that started LTS (#9716)
-      return DateTime.local() > DateTime.fromISO(schedule.lts);
+      return DateTime.now() > DateTime.fromISO(schedule.lts);
```
viceice, line 56: why this change? should be in a separate PR

lib/modules/versioning/node/index.spec.ts
```diff
@@ -53,8 +43,8 @@ describe('modules/versioning/node/index', () => {
     ${'10.0.0a'}  | ${t1} | ${false}
     ${'9.0.0'}    | ${t1} | ${false}
   `('isStable("$version") === $expected', ({ version, time, expected }) => {
-    DateTime.local = (...args: any[]) =>
-      args.length ? dtLocal.apply(DateTime, args) : time;
+    jest.spyOn(Settings, 'now').mockReturnValue(time.valueOf());
```
viceice, line 46: use vitest fake timer instead

lib/modules/versioning/lambda-node/index.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    ...lambdaSchedule['20'],
+    support: true,
+  },
+  '22': undefined,
+});
+
+const mockDataFiles = new Map(dataFiles);
+mockDataFiles.set('data/lambda-node-js-schedule.json', scheduleWithSupportTrue);
+
+jest.spyOn(dataFiles, 'get').mockImplementation((fileName) => {
```
viceice, line 25: we're now using vitest

viceice (changes requested):

lib/modules/versioning/lambda-node/index.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+describe('modules/versioning/lambda-node/index', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(DateTime.fromISO('2021-03-20').valueOf());
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
```
viceice, line 47: drop it

lib/modules/versioning/lambda-node/index.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+  );
+
+  return {
+    default: mockDataFiles,
+  };
+});
+
+describe('modules/versioning/lambda-node/index', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
```
viceice, line 40: move to before all

lib/modules/versioning/lambda-node/schedule.ts
```diff
@@ -0,0 +1,30 @@
...
+export type LambdaData = Record<string, LambdaSchedule>;
+
+const lambdaSchedule: LambdaData = JSON.parse(
+  dataFiles.get('data/lambda-node-js-schedule.json')!,
+);
+
+export function findLambdaScheduleForVersion(
+  version: string,
+): LambdaSchedule | null {
+  const majorVersionMatch = /^v?([0-9]+)\./.exec(version);
```
viceice, line 23:
```suggestion
  const majorVersionMatch = regEx(/^v?([0-9]+)\./).exec(version);
```
move below condition (perf)

lib/modules/versioning/lambda-node/schedule.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (!isStable(version)) {
+    return null;
+  }
+
+  return lambdaSchedule[majorVersionMatch![1]] ?? null;
```
viceice, line 29: are you sure it always matched when coming here?

tools/static-data/generate-lambda-node-schedule.mjs
```diff
@@ -0,0 +1,50 @@
...
+const lambdaDataUrl = 'https://endoflife.date/api/aws-lambda.json';
+
+await (async () => {
+  console.log('Generating node schedule');
+
+  /**
+   * @type Array<RuntimeDefinition>
+   */
+  const lambdas = await fetch(lambdaDataUrl).then((response) =>
+    response.json(),
```
viceice, line 20:
```suggestion
    if (!response.ok) {
      console.error(`Failed to fetch ${lambdaDataUrl}`, response);
      process.exit(1);
    }
    response.json(),
```

catch download errors

lib/modules/versioning/lambda-node/index.spec.ts
```diff
@@ -0,0 +1,154 @@
+import { DateTime } from 'luxon';
+import type { DataFile } from '../../../data-files.generated';
+import type { LambdaData } from './schedule';
+import { api as lambdaVer } from '.';
+
+vi.mock('../../../data-files.generated', async (importOriginal) => {
+  const dataFiles = (await importOriginal<{ default: Map<DataFile, string> }>())
```
viceice, line 7:
```suggestion
  const dataFiles = (await importOriginal<typeof import('../../../data-files.generated')>())
```

lib/modules/versioning/lambda-node/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export function isStable(version: string): boolean {
+  const schedule = findLambdaScheduleForVersion(version);
+
+  if (schedule === null) {
+    return false;
+  }
+
+  if (typeof schedule.support === 'string') {
+    return DateTime.now() < DateTime.fromISO(schedule.support);
```
viceice, line 26:
```suggestion
    return DateTime.local() < DateTime.fromISO(schedule.support);
```

use local time for comparison like we do in node versioning

lib/modules/versioning/lambda-node/schedule.ts
Garethp: If it's not matched, the `isStable(version)` check will fail first and return null. But good point, that's non-obvious and not within this function. I've  added a check to see if it was matched and verified that there's already a test `isStable("Iron")` that hits that return and is testing it

lib/modules/versioning/node/index.ts
```diff
@@ -18,7 +18,7 @@ function normalizeValue(value: string): string {
   return value;
 }
 
-function getNewValue({
+export function getNewValue({
```
viceice, line 21: why not using it from the `api` export?

tools/static-data/generate-lambda-node-schedule.mjs
```diff
@@ -0,0 +1,55 @@
+import { updateJsonFile } from './utils.mjs';
+
+/**
+ * @typedef RuntimeDefinition
+ * @type {object}
+ * @property {string} cycle - The ID of the Runtime.
+ * @property {boolean|string} support - Either `true` if in support or a string denoting when support for this Runtime
+ *                                      will end. 0.10.x is a sole exception which has `false` and will be filtered out.
```
viceice, line 8: maybe create a zod schema instead? 🤔

tools/static-data/generate-lambda-node-schedule.mjs
```diff
@@ -0,0 +1,53 @@
...
+  const lambdas = await fetch(lambdaDataUrl).then(async (response) => {
+    if (!response.ok) {
+      console.error(`Failed to fetch ${lambdaDataUrl}`, response);
+      process.exit(1);
+    }
+
+    return RuntimesSchema.parseAsync(await response.json());
+  });
+
+  const nodeRuntimes = z.record(z.string(), RuntimesSchema.element).parse({});
```
Garethp, line 29: I had to do a `.parse()` here because typescript is still inferring a type, even in JS, and `nodeRuntimes[versionMatch[1]] = lambda;` wanted `nodeRuntimes` to be better typed. Before I was using the JSDocs parameters, but moving to zod I couldn't find a better public interface for getting the output type. This lets typescript infer the correct typing of this record

Garethp:
@viceice, are there any other changes you think need to be made before this is ready to be merged in?

tools/static-data/generate-lambda-node-schedule.mjs
viceice: what about this?
```suggestion
  /** @type {Record<string,RuntimesSchema>} */
  const nodeRuntimes = {};
```

viceice (commented):
otherwise lgtm

tools/static-data/generate-lambda-node-schedule.mjs
Garethp: That was the first thing I tried and it doesn't work. `nodeRuntimes[versionMatch[1]] = lambda;` ends up complaining because it expects `lambda` to match the `ZodObject` typing, which contains extra properties. I couldn't figure out a way to get it to type-hint against the original shape that zod is wrapping instead of the `ZodObject` type. Here's the type error when I try it:

```
error TS2740: Type '{ support: string | boolean; cycle: string; }' is missing the following properties from type 'ZodObject<{ cycle: ZodString; support: ZodUnion<[ZodBoolean, ZodString]>; }, "strip", ZodTypeAny, { support: string | boolean; cycle: string; }, { ...; }>': _cached, _getCached, _parse, shape, and 52 more.
```

tools/static-data/generate-lambda-node-schedule.mjs
```diff
@@ -0,0 +1,53 @@
...
+const RuntimesSchema = z.array(
+  z.object({
+    cycle: z.string().describe('The ID of the Runtime'),
+    support: z
+      .union([z.boolean(), z.string()])
+      .describe(
+        'Either `true` if in support, or a string denoting when support for this Runtime will end. 0.10.x is a sole exception which has a value of `false` and will be filtered out',
+      ),
+  }),
+);
```
viceice, line 13:
```suggestion
const RuntimesSchema =
  z.object({
    cycle: z.string().describe('The ID of the Runtime'),
    support: z
      .union([z.boolean(), z.string()])
      .describe(
        'Either `true` if in support, or a string denoting when support for this Runtime will end. 0.10.x is a sole exception which has a value of `false` and will be filtered out',
      ),
  });
const RuntimesArraySchema = z.array(RuntimesSchema);
```

tools/static-data/generate-lambda-node-schedule.mjs
```diff
@@ -0,0 +1,53 @@
...
+await (async () => {
+  console.log('Generating node schedule');
+
+  const lambdas = await fetch(lambdaDataUrl).then(async (response) => {
+    if (!response.ok) {
+      console.error(`Failed to fetch ${lambdaDataUrl}`, response);
+      process.exit(1);
+    }
+
+    return RuntimesSchema.parseAsync(await response.json());
```
viceice, line 26:
```suggestion
    return RuntimesArraySchema.parseAsync(await response.json());
```

tools/static-data/generate-lambda-node-schedule.mjs
viceice: and what about this? (needs other change above)
```suggestion
  /** @type {Record<string, z.infer<RuntimesSchema>>} */
  const nodeRuntimes = {};
```

viceice (approved):

rarkins (approved):

PR #32499, opened 2024-11-13
feat(logger): prefix log env settings with `RENOVATE_`

viceice:
## Changes
Prefix all logger relevant settings with `RENOVATE_`. Fallback to env vars without prefix.


docs/usage/examples/self-hosting.md
```diff
@@ -383,12 +383,12 @@ spec:
 
 ## Logging
 
-If you're ingesting/parsing logs into another system then we recommend you set `LOG_LEVEL=debug` and `LOG_FORMAT=json` in your environment variables.
+If you're ingesting/parsing logs into another system then we recommend you set `RENOVATE_LEVEL=debug` and `RENOVATE_LOG_FORMAT=json` in your environment variables.
```
rarkins, line 386:
```suggestion
If you're ingesting/parsing logs into another system then we recommend you set `RENOVATE_LOG_LEVEL=debug` and `RENOVATE_LOG_FORMAT=json` in your environment variables.
```

rarkins (changes requested):

lib/logger/index.ts
```diff
@@ -123,15 +128,16 @@ loggerLevels.forEach((loggerLevel) => {
 });
 
 // istanbul ignore if: not easily testable
```
viceice, line 130:
```suggestion
```

lib/logger/index.ts
```diff
@@ -123,15 +128,16 @@ loggerLevels.forEach((loggerLevel) => {
...
-if (is.string(process.env.LOG_FILE)) {
+const logFile = getEnv('LOG_FILE');
+if (is.string(logFile)) {
```
viceice, line 132:
```suggestion
// istanbul ignore if: not easily testable
if (is.string(logFile)) {
```

viceice:
I've left some dev settings as `LOG_LEVEL`, so containerbase can benefit from it too

.github/label-actions.yml
```diff
@@ -145,7 +145,7 @@
     A maintainer wants you to find and share a specific log message (`packageFiles with updates`) from your repository or environment.
 
 
-    If you self-host Renovate: make sure you run Renovate with `LOG_LEVEL=debug` to get the debug log messages! Next, open the debug-level logs and search for `packageFiles with updates`. This text marks the start of a structured log message that shows every package file, dependency, and update that Renovate found.
+    If you self-host Renovate: make sure you run Renovate with `RENOVATE_LOG_LEVEL=debug` to get the debug log messages! Next, open the debug-level logs and search for `packageFiles with updates`. This text marks the start of a structured log message that shows every package file, dependency, and update that Renovate found.
```
rarkins, line 148: Do we need to add a note that LOG_LEVEL may be needed for older versions? Or just hope there isn't too much confusion during the changeover?

rarkins: True, we can leave the old way, as it's still supported

viceice (commented):
@rarkins should we defer all docs changes to not confuse with older versions?

rarkins (approved):

PR #32506, opened 2024-11-13
fix(go): Remove deprecated '-d' flag when doing 'go get'

echarrod:
## Changes

- Remove deprecated `-d` flag from go get. 

## Context

- This flag was changed to have no effect in [Go 1.18](https://go.dev/doc/go-get-install-deprecation)
- Go's [release policy](https://go.dev/doc/devel/release) only officially support the latest two major versions, i.e. 1.23 and 1.22 at the moment

### DIFF
--- a/lib/modules/manager/dockerfile/extract.spec.ts
+++ b/lib/modules/manager/dockerfile/extract.spec.ts
@@ -463,7 +463,7 @@ describe('modules/manager/dockerfile/extract', () => {
 
     it('detects ["stage"] and ["final"] deps of docker multi-stage build.', () => {
       const res = extractPackageFile(
-        'FROM node:8.15.1-alpine as skippedfrom\nFROM golang:1.7.3 as builder\n\n# comment\nWORKDIR /go/src/github.com/alexellis/href-counter/\nRUN go get -v golang.org/x/net/html  \nCOPY app.go    .\nRUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .\n\nFROM alpine:latest  \nRUN apk --no-cache add ca-certificates\nWORKDIR /root/\nCOPY --from=builder /go/src/github.com/alexellis/href-counter/app .\nCMD ["./app"]\n',
+        'FROM node:8.15.1-alpine as skippedfrom\nFROM golang:1.7.3 as builder\n\n# comment\nWORKDIR /go/src/github.com/alexellis/href-counter/\nRUN go get -d -v golang.org/x/net/html  \nCOPY app.go    .\nRUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .\n\nFROM alpine:latest  \nRUN apk --no-cache add ca-certificates\nWORKDIR /root/\nCOPY --from=builder /go/src/github.com/alexellis/href-counter/app .\nCMD ["./app"]\n',
         '',
         {},
       )?.deps;
--- a/lib/modules/manager/dockerfile/extract.spec.ts
+++ b/lib/modules/manager/dockerfile/extract.spec.ts
@@ -463,7 +463,7 @@ describe('modules/manager/dockerfile/extract', () => {
 
     it('detects ["stage"] and ["final"] deps of docker multi-stage build.', () => {
       const res = extractPackageFile(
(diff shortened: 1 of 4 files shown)


echarrod:
Ah I see we're using older versions of Go elsewhere, maybe we should update those at the same time 🤔

rarkins:
Usually Renovate will use the latest Go version, but users may pin to an older version. Therefore we ideally would look at the go.mod directive and decide on -d based on the version in use

lib/modules/manager/dockerfile/extract.spec.ts
```diff
@@ -463,7 +463,7 @@ describe('modules/manager/dockerfile/extract', () => {
 
     it('detects ["stage"] and ["final"] deps of docker multi-stage build.', () => {
       const res = extractPackageFile(
-        'FROM node:8.15.1-alpine as skippedfrom\nFROM golang:1.7.3 as builder\n\n# comment\nWORKDIR /go/src/github.com/alexellis/href-counter/\nRUN go get -d -v golang.org/x/net/html  \nCOPY app.go    .\nRUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .\n\nFROM alpine:latest  \nRUN apk --no-cache add ca-certificates\nWORKDIR /root/\nCOPY --from=builder /go/src/github.com/alexellis/href-counter/app .\nCMD ["./app"]\n',
+        'FROM node:8.15.1-alpine as skippedfrom\nFROM golang:1.7.3 as builder\n\n# comment\nWORKDIR /go/src/github.com/alexellis/href-counter/\nRUN go get -v golang.org/x/net/html  \nCOPY app.go    .\nRUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .\n\nFROM alpine:latest  \nRUN apk --no-cache add ca-certificates\nWORKDIR /root/\nCOPY --from=builder /go/src/github.com/alexellis/href-counter/app .\nCMD ["./app"]\n',
```
viceice, line 466: revert, unrelated

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -243,7 +243,7 @@ export async function updateArtifacts({
       }
     }
 
-    let args = `get -d -t ${goGetDirs ?? './...'}`;
+    let args = `get -t ${goGetDirs ?? './...'}`;
```
viceice, line 246: extract go version from `go` and / or `toolchain` directive. drop `-d` only if it is `>=1.18`.

viceice (changes requested):

lib/modules/manager/gomod/artifacts.ts
echarrod: Sorry was just about to comment: I made an update to update Go version after I realised it was currently older: https://github.com/renovatebot/renovate/pull/32514

echarrod:
> Usually Renovate will use the latest Go version, but users may pin to an older version. Therefore we ideally would look at the go.mod directive and decide on -d based on the version in use

Thanks for the response, I wouldn't have guessed anyone is on an older version, given Go's [release policy](https://go.dev/doc/devel/release) of only officially support the latest two major versions, i.e. 1.23 and 1.22 at the moment. 

It also means you can barely upgrade to any libraries at all if you're on < 1.18, so I wouldn't have thought it would be feasible to do this _and_ run Renovate!

viceice:
🤔  go v1.17 was realesed ~3 years ago. https://go.dev/blog/go1.17

https://github.com/search?q=%22go+1.17%22++NOT+is%3Aarchived+path%3Ago.mod&type=code

It seems there are some repos with go v1.17.

@secustor @rarkins Can we really assume someone uses go <1.18 and also is using renovate?

I think @echarrod is right and we can assume any update needs a newer go version now.

rarkins:
Go is a little bit unique though because even if a repo has a go mod directive of 1.17, this means "Use Go 1.17 compatibility" and not "Use Go 1.17". i.e. it's not like a .nvmrc file. So Renovate intentionally runs the latest 1.x Go no matter what the Go mod directive says, because Go is backwards compatible. The only way removing this flag could cause problems is if a user _pinned_ to an older exact version using `constraints`.

viceice:
> Can we detect if someone has pinned their go constraints, or should we assume they won't?

we can detect

echarrod:
Hey all 👋 
Is this the blocker for this MR:

 - if:
   - a user has pinned their go constraint
   - and it's < 1.18
 - still use `-d` in `go get`?

viceice:
yes, as told before

rarkins:
Converted to issue due to inactivity: https://github.com/renovatebot/renovate/issues/34097

PR #32514, opened 2024-11-13
chore(manager/gomod): Update go version to 1.23.3 in tests/examples

echarrod:
## Changes

- Update Go version to 1.23

## Context

- Go's [release policy](https://go.dev/doc/devel/release) only officially support the latest two major versions, i.e. 1.23 and 1.22 at the moment
- Originally was removing redundant flag in https://github.com/renovatebot/renovate/pull/32506, but wanted to ensure this wouldn't have any adverse effects for Go versions < 1.18


rarkins (commented):
None of these changes makes any difference to how Renovate behaves at runtime

echarrod:
> None of these changes makes any difference to how Renovate behaves at runtime

I just thought it would be more familiar for users to see an up to date Go version in the docs, that they would be using, and thought we might as well run the tests against a later Go version specified

rarkins:
OK, I've updated this from `feat` to `chore`

rarkins (approved):

lib/modules/manager/gomod/__fixtures__/3/go-mod
```diff
@@ -133,4 +133,3 @@ replace (
 )
 
 toolchain go1.22.1
```
viceice, line 135: this should be removed, ho would remove it. should never lower than go directive

viceice (changes requested):

lib/modules/manager/gomod/__fixtures__/3/go-mod
echarrod: Updated in https://github.com/renovatebot/renovate/pull/32546

PR #32529, opened 2024-11-14
test: add rustup fixture for parsing

rarkins:


lib/modules/manager/cargo/extract.spec.ts
```diff
@@ -557,6 +560,16 @@ replace-with = "mcorbin"
...
     });
 
+    it('extracts rustup with nested versions', async () => {
+      const res = await extractPackageFile(
+        rustupCargotoml,
+        'Cargo.toml',
+        config,
+      );
+      mockReadLocalFile({ 'Cargo.lock': rustupCargolock });
+      expect(res?.deps).toHaveLength(25);
```
rarkins, line 570: I'm not sure exactly what this number should be

viceice:
we should try to reduce it to minimal

viceice:
Obsolete
- #32533

PR #32530, opened 2024-11-14
fix(manager/cargo): skip non-string `package.version`

### LINKED ISSUE #32528: Error parsing Cargo.toml file due to ZoD schema validation error
### Discussed in https://github.com/renovatebot/renovate/discussions/32523

<sup>Originally posted by **rami3l** November 14, 2024</sup>

A Mend.io-hosted app

_No response_

Coming from Rustup, we are using the following configuration WRT `cargo` dependencies:

```json
        {
            "matchManagers": [
                "cargo"
            ],
            "matchUpdateTypes": [
                "patch"
            ],
            "enabled": false
        },
        {
            "matchManagers": [
                "cargo"
            ],
            "matchUpdateTypes": [
                "minor"
            ],
            "matchCurrentVersion": "!/^0/",
            "enabled": false
        },
```
_https://github.com/rust-lang/rustup/blob/c8eddcb4c1dc9aea8067fa73179ce1791aa24d1d/.github/renovate.json#L24-L41_

... and we set the lock file maintenance to be executed every Monday. It **used to be working alright**, at least till the point of 2024/10/28 (https://github.com/rust-lang/rustup/pull/4066), but after that I can no longer see `cargo` on the dependency dashboard (https://github.com/rust-lang/rustup/issues/3142, you can refer to the editing history of that post as well). The logs say the same:

```json
INFO: Dependency extraction complete
{
  "baseBranch": "master"
  "stats": {
    "managers": {
      "dockerfile": {
        "fileCount": 23,
        "depCount": 23
      },
      "github-actions": {
        "fileCount": 10,
        "depCount": 107
      }
    },
    "total": {
      "fileCount": 33,
      "depCount": 130
    }
  }
}
```

Many thanks in advance!
</div>

mkniewallner:
## Changes

Parsing fails because `package.version` is not necessarily a string, and can also reference the current workspace package, for instance for `rustup`:

```toml
[package]
name = "rustup"
version.workspace = true
```

Before switching to `zod`, I think this was also not fully working as expected, because we [either expected a string or `undefined`](https://github.com/renovatebot/renovate/blob/89ebb35d7a303b2e2dcac83bcda0937a25971a6f/lib/modules/manager/cargo/extract.ts#L248-L252), but it did not hard fail. We might want to parse the version from `workspace.package` later on, but since this was not supported before introducing `zod`, probably best to just skip this case for now.

Cherry-picked the test case added in https://github.com/renovatebot/renovate/pull/32529, but I don't know if we want to use it or if this was just for reproducing. I can write a dedicated test case that does not use snapshots if you prefer, not sure what was desirable since it's a regression test.

## Context

- Closes #32528.


lib/modules/manager/cargo/schema.ts
```diff
@@ -105,7 +105,13 @@ const CargoTarget = z.record(z.string(), CargoSection);
 
 export const CargoManifestSchema = Toml.pipe(
   CargoSection.extend({
-    package: z.object({ version: z.string().optional() }).optional(),
+    package: z
+      .object({
+        version: z
+          .union([z.string(), z.object({}).transform(() => undefined)])
```
viceice, line 111: working on a better fix

viceice (changes requested):

PR #32543, opened 2024-11-14
fix(git): fix gpg commit signing (#32483)

bpfoster:
## Changes

After upgrading to Renovate 39, committing using GPG keys began failing with the error message `cannot open '/dev/tty': No such device or address`.

This change adds the `--no-tty` and `--batch` commands to the `gpg` import of the private key.
These flags are described in the manpage as:
> --no-tty
>   Make sure that the TTY (terminal) is never used for any output. This option is needed in some cases because GnuPG sometimes prints warnings to the TTY even if --batch is used. 
> --batch
>   Use batch mode. Never ask, do not allow interactive commands. --no-batch disables this option. Note that even with a filename given on the command line, gpg might still need to read from STDIN (in particular if gpg figures that the input is a detached signature and no data file has been specified). Thus if you do not want to feed data via STDIN, you should connect STDIN to /dev/null. It is highly recommended to use this option along with the options --status-fd and --with-colons for any unattended use of gpg. Should not be used in an option file.

I am not sure that `--batch` is strictly necessary to fix this issue, but it does seem to be recommended in this usage.

## Context

This references discussion #32483.

I have not been able to build renovate here to test these exact changes.  I did however override the `gpg` command in the docker image to insert the 2 flags into the command, and this did solve the issue.

### DIFF
--- a/lib/util/git/private-key.ts
+++ b/lib/util/git/private-key.ts
@@ -64,7 +64,9 @@ class GPGKey extends PrivateKey {
   protected async importKey(): Promise<string | undefined> {
     const keyFileName = upath.join(os.tmpdir() + '/git-private-gpg.key');
     await fs.outputFile(keyFileName, this.key);
-    const { stdout, stderr } = await exec(`gpg --batch --no-tty --import ${keyFileName}`);
+    const { stdout, stderr } = await exec(
+      `gpg --batch --no-tty --import ${keyFileName}`,
+    );
     logger.debug({ stdout, stderr }, 'Private key import result');
     await fs.remove(keyFileName);
     return `${stdout}${stderr}`
--- a/lib/util/git/private-key.ts
+++ b/lib/util/git/private-key.ts
@@ -64,7 +64,7 @@ class GPGKey extends PrivateKey {
   protected async importKey(): Promise<string | undefined> {
     const keyFileName = upath.join(os.tmpdir() + '/git-private-gpg.key');
     await fs.outputFile(keyFileName, this.key);
-    const { stdout, stderr } = await exec(`gpg --import ${keyFileName}`);
+    const { stdout, stderr } = await exec(`gpg --batch --no-tty --import ${keyFileName}`);
     logger.debug({ stdout, stderr }, 'Private key import result');
     await fs.remove(keyFileName);
     return `${stdout}${stderr}`
--- a/lib/util/git/private-key.ts
+++ b/lib/util/git/private-key.ts
@@ -65,6 +65,7 @@ class GPGKey extends PrivateKey {
     const keyFileName = upath.join(os.tmpdir() + '/git-private-gpg.key');
     await fs.outputFile(keyFileName, this.key);
     const { stdout, stderr } = await exec(
+      // --batch --no-tty flags allow Renovate to skip warnings about unsupported algorithms in the key
       `gpg --batch --no-tty --import ${keyFileName}`,
     );
     logger.debug({ stdout, stderr }, 'Private key import result');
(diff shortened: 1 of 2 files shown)


rarkins (commented):
Do you have a link to any docs describing why this is necessary? Eg was it a change of GPG version from Ubuntu 20 to 24?

bpfoster:
Hi @rarkins.

GPG was upgraded from 2.2.19 to 2.4.4.  After digging more into this to answer your question, I have a reason why this is happening.

The overlooked bit in the error message was "WARNING: key xxx contains preferences for unavailable algorithms".

Whether a change in gpg itself or in Ubuntu's configured algorithms I'm not sure, but there is a difference in supported algorithms between the 2 renovate images:
```
gpg (GnuPG) 2.2.19
libgcrypt 1.8.5
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Home: /home/ubuntu/.gnupg
Supported algorithms:
Pubkey: RSA, ELG, DSA, ECDH, ECDSA, EDDSA
Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
        CAMELLIA128, CAMELLIA192, CAMELLIA256
Hash: SHA1, RIPEMD160, SHA256, SHA384, SHA512, SHA224
Compression: Uncompressed, ZIP, ZLIB, BZIP2
```
vs
```
gpg (GnuPG) 2.4.4
libgcrypt 1.10.3
Copyright (C) 2024 g10 Code GmbH
License GNU GPL-3.0-or-later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Home: /home/ubuntu/.gnupg
Supported algorithms:
Pubkey: RSA, ECDH, ECDSA, EDDSA
Cipher: AES, AES192, AES256
Hash: SHA1, SHA256, SHA384, SHA512, SHA224
Compression: Uncompressed, ZIP, ZLIB, BZIP2
```

It would appear that my gpg key contains some now-unsupported algorithms (IDEA, 3DES, CAST5 specifically) in its preferences.  This appears to be only a warning, since if they key _is_ imported (via adding these 2 flags), it is still able to be used to sign (as it also contains preferences for still-supported ciphers).

If I try and import this key within a tty, gpg will automatically enter into an interactive, preferences update mode:
```
Set preference list to:
     Cipher: 3DES
     AEAD: 
     Digest: SHA1
     Compression: ZIP, Uncompressed
     Features: Keyserver no-modify
Really update the preferences? (y/N)
```

**This** is what's causing the TTY error when invoked from renovate.

It is in fact the `--batch` flag that allows gpg to proceed through this warning and not attempt to prompt for key update.  Just `--no-tty` gives an error:
```
Sorry, no terminal at all requested - can't get input
```

So, to wrap up...
It appears that part of this is on me - I _should_ (and will) update my GPG key, and doing so will resolve the renovate error.  I do think you may wish to consider adding these flags to Renovate regardless though, as their usage seems to be indicated in this scenario (when invoking from a headless, non-interactive script).

rarkins (dismissed):

lib/util/git/private-key.ts
```diff
@@ -64,7 +64,9 @@ class GPGKey extends PrivateKey {
   protected async importKey(): Promise<string | undefined> {
     const keyFileName = upath.join(os.tmpdir() + '/git-private-gpg.key');
     await fs.outputFile(keyFileName, this.key);
-    const { stdout, stderr } = await exec(`gpg --import ${keyFileName}`);
+    const { stdout, stderr } = await exec(
+      `gpg --batch --no-tty --import ${keyFileName}`,
```
rarkins, line 68:
```suggestion
      // --batch --no-tty flags allow Renovate to skip warnings about unsupported algorithms in the key
      `gpg --batch --no-tty --import ${keyFileName}`,
```

rarkins (approved):

Issue #14149, opened 2022-02-10
Support replacement PRs on all managers


HonkingGoose:
### What would you like Renovate to be able to do?

Right now we only raise replacement PRs for `npm` dependencies.

I think we want to support all managers eventually.
This is a big issue to track support for all managers:

- [ ] `ansible`
- [ ] `docker-compose`
- [ ] `dockerfile`
- [ ] `droneci`
- [ ] `gitlabci`
- [ ] `kubernetes`
- [ ] `cake`
- [ ] `nuget`
- [ ] `mix`
- [ ] `gomod`
- [ ] `gradle`
- [ ] `maven`
- [ ] `meteor`
- [x] `npm`
- [ ] `nodenv`
- [ ] `nvm`
- [ ] `travis`
- [ ] `composer`
- [ ] `pip-compile`
- [ ] `pip_requirements`
- [ ] `pip_setup`
- [ ] `pipenv`
- [ ] `poetry`
- [ ] `pyenv`
- [ ] `setup-cfg`
- [ ] `bundler`
- [ ] `ruby-version`
- [ ] `cargo`
- [ ] `ansible-galaxy`
- [ ] `argocd`
- [ ] `azure-pipelines`
- [ ] `batect`
- [ ] `batect-wrapper`
- [ ] `bazel`
- [ ] `bitbucket-pipelines`
- [ ] `buildkite`
- [ ] `cdnurl`
- [ ] `circleci`
- [ ] `cloudbuild`
- [ ] `cocoapods`
- [ ] `deps-edn`
- [ ] `flux`
- [ ] `git-submodules`
- [ ] `github-actions`
- [ ] `gitlabci-include`
- [ ] `gradle-wrapper`
- [ ] `helm-requirements`
- [ ] `helm-values`
- [ ] `helmfile`
- [ ] `helmsman`
- [ ] `helmv3`
- [ ] `homebrew`
- [ ] `html`
- [ ] `jenkins`
- [ ] `jsonnet-bundler`
- [ ] `kustomize`
- [ ] `leiningen`
- [ ] `pre-commit`
- [ ] `pub`
- [ ] `regex`
- [ ] `sbt`
- [ ] `swift`
- [ ] `terraform`
- [ ] `terraform-version`
- [ ] `terragrunt`
- [ ] `terragrunt-version`

### If you have any ideas on how this should be implemented, please tell us here.

Nope.

### Is this a feature you are interested in implementing yourself?

No

[...]

PR #32550, opened 2024-11-14
feat(maven): Add maven relocation support

### LINKED ISSUE #5667: Detect and follow Maven group and artifact moves
Some maven artifacts change group and artifact ids on major releases.
Many projects include a link forward to the new location, for example the "Note: This artifact was moved to" sections of:
* https://mvnrepository.com/artifact/commons-collections/commons-collections
* https://mvnrepository.com/artifact/org.apache.commons/commons-io
* https://mvnrepository.com/artifact/org.apache.commons/commons-math

When group and artifacts have their next major version linked in maven central, allow renovate updates to upgrade to the new version at the new group/artifact location.

N/A

**Additional context**

NOTE: this applies to all maven based package managers, including gradle and sbt

jonasrutishauser:
## Changes

Use maven relocation information to detect package name changes and allow replacement updates for maven.

## Context

Closes #5667.

The issue #14149 should be updated after merging this.


rarkins:
@jonasrutishauser thanks for this interesting feature. Could you share an example repo you've tested against?

viceice (commented):
please split this PR into three parts
- worker
- datasource 
- manager

you can use this PR for one of them and the order probably doesn't matter.

lib/workers/repository/process/lookup/index.ts
```diff
@@ -569,6 +569,12 @@ export async function lookupUpdates(
 
     if (isReplacementRulesConfigured(config)) {
       addReplacementUpdateIfValid(res.updates, config);
+    } else if (dependency?.replacementName || dependency?.replacementVersion) {
```
astellingwerf, line 572: There's code in datasource which assumes that both properties must be set together. Wouldn't it be safer to do the same thing here, so turn the `||` into `&&`?

jonasrutishauser: Yes this makes more sense.

astellingwerf (approved):

astellingwerf:
@jonasrutishauser, do we understand why on your test repo, [there's no PR for Ant 1.7.0](https://github.com/jonasrutishauser/renovate-32550/pulls)? Is that just a matter of throttling, or is there a reason why the replacement does not get noticed?

jonasrutishauser:
> @jonasrutishauser, do we understand why on your test repo, [there's no PR for Ant 1.7.0](https://github.com/jonasrutishauser/renovate-32550/pulls)? Is that just a matter of throttling, or is there a reason why the replacement does not get noticed?

I think the reason is, that the version 1.7.0 has no timestamp in the [html](https://repo1.maven.org/maven2/ant/ant/) and is therefore ignored/not seen. In the `maven-metadata.xml` this version is missing too.

lib/workers/repository/process/lookup/index.spec.ts
```diff
@@ -4947,6 +4947,29 @@ describe('workers/repository/process/lookup/index', () => {
...
+      config.datasource = MavenDatasource.id;
+      getMavenReleases.mockResolvedValueOnce({
+        releases: [{ version: '2.0.0' }],
+        replacementName: 'foo:bar',
+        replacementVersion: '2.0.0',
+      });
+
+      const { updates } = await Result.wrap(
+        lookup.lookupUpdates(config),
+      ).unwrapOrThrow();
```
viceice, line 4962:
```suggestion
      const { updates } = await lookup.lookupUpdates(config);
```
no need to wrap here

jonasrutishauser: With your suggestion I receive `Property 'updates' does not exist on type 'Result<UpdateResult, Error>'.ts(2339)`.

The only working code would be the following:
```ts
      const { updates } = (await lookup.lookupUpdates(config)).unwrapOrThrow();
```
As all other tests don't use this variant, I think it shouldn't be changed.

lib/workers/repository/process/lookup/index.ts
```diff
@@ -571,6 +571,12 @@ export async function lookupUpdates(
 
     if (isReplacementRulesConfigured(config)) {
       addReplacementUpdateIfValid(res.updates, config);
+    } else if (dependency?.replacementName && dependency?.replacementVersion) {
```
viceice, line 574:
```suggestion
    } else if (dependency?.replacementName && dependency.replacementVersion) {
```

viceice (approved):

PR #32556, opened 2024-11-15
feat(config): allow multiple branch/pr limits

### LINKED ISSUE #10546: Allow multiple branch/PR limit groups

It would be great to be able to set a prConcurrentLimit per manager. It looks like there was a bit of discussion in renovatebot/config-help: https://github.com/renovatebot/config-help/issues/563, which was closed without a solution.

eg. Ideally we could set a prConcurrentLimit like the following to allow up to 5 PRs each for both the `gradle` and `npm` manager.

```
  packageRules: [
        {
            matchManagers: ["gradle"],
            addLabels: ["java"],
            prConcurrentLimit: 5
        },
        {
            matchManagers: ["npm"],
            addLabels: ["javascript"],
            prConcurrentLimit: 5
        },
    ],
```

### LINKED ISSUE #32133: Evaluate pr/branch limits per-branch

Current, limit are calculated prior to branch execution: https://github.com/renovatebot/renovate/blob/ed7b17acc9d03bc56bedccb7d39cff5b9c5c844d/lib/workers/repository/process/write.ts#L130-L138

And evaluated within branch execution such as https://github.com/renovatebot/renovate/blob/ed7b17acc9d03bc56bedccb7d39cff5b9c5c844d/lib/workers/repository/update/branch/index.ts#L212-L224

Much of the logic is also inside https://github.com/renovatebot/renovate/blob/main/lib/workers/repository/process/limits.ts

Instead of having global limits as today, we should instead allow separate limits per-branch, and evaluate _those_ limits per-branch instead. This means removing these global limit settings:
- https://github.com/renovatebot/renovate/blob/ed7b17acc9d03bc56bedccb7d39cff5b9c5c844d/lib/workers/repository/process/write.ts#L132
- https://github.com/renovatebot/renovate/blob/ed7b17acc9d03bc56bedccb7d39cff5b9c5c844d/lib/workers/repository/process/write.ts#L138

And instead passing the relevant limit like `prBranchLimit` and `prConcurrentLimit` through to the "is limit reached?" functions.

Ideally the point in our workers where these limits are evaluated is kept the same as today so that there is "less changes" and risk of breaking things. And so the change is that the limits are passed through rather than used globally.

Note: commits will still be global

RahulGautamSingh:
## Changes
Allows concurrent and hourly limits to be set in package rules. (implemented everything mentioned in [this ](https://github.com/renovatebot/renovate/issues/10546#issuecomment-2461734472)comment)

## Context
- Closes: https://github.com/renovatebot/renovate/issues/10546
- Closes: https://github.com/renovatebot/renovate/issues/32133

- [ ] I have updated the documentation, or (required will add after review of code)

- [x] Both unit tests + ran on a real repository (Tested both old and new behaviour)

Old Behaviour: [follows hourly limits](https://github.com/Rahul-renovate-testing/normal-repo), [1](https://github.com/Rahul-renovate-testing/1056-1), [2](https://github.com/Rahul-renovate-testing/10546-2), [3](https://github.com/Rahul-renovate-testing/10546-2)
New Behaviour [1](https://github.com/Rahul-renovate-testing/10546_4), [2](https://github.com/Rahul-renovate-testing/10546-5), [3](https://github.com/Rahul-renovate-testing/10546-6)

_Note: Test descriptions present in readme of each test repo_


lib/workers/repository/update/branch/index.ts
```diff
@@ -209,9 +209,15 @@ export async function processBranch(
         };
       }
     }
+
+    logger.debug({
```
rarkins, line 213: Are you sure you want this printed once per branch, even if no limits are used? Maybe use a more compact form without metadata

lib/workers/repository/process/limits.ts
```diff
@@ -7,135 +7,77 @@ import { scm } from '../../../modules/platform/scm';
...
 }
 
-export async function getConcurrentPrsRemaining(
+export async function getConcurrentPrsCount(
   config: RenovateConfig,
   branches: BranchConfig[],
 ): Promise<number> {
-  if (config.prConcurrentLimit) {
-    logger.debug(`Calculating prConcurrentLimit (${config.prConcurrentLimit})`);
+  const openPrs: Pr[] = [];
```
rarkins, line 43:
```suggestion
  const openPrCount = 0;
```

Simplify below. Same for branch count

rarkins (changes requested):

lib/workers/repository/update/branch/index.ts
RahulGautamSingh: > Before executing branches, we should calculate the hourly and concurrent existing and log these values (i.e. log how many are existing, not how many left)

I misunderstod this and thought we need to log before each branch.
I will move it above in the flow.

lib/workers/global/limits.spec.ts
```diff
@@ -60,4 +67,242 @@ describe('workers/global/limits', () => {
...
+      expect(calcLimit(upgrades, 'prHourlyLimit')).toBe(10);
+      expect(calcLimit(upgrades, 'branchConcurrentLimit')).toBe(11);
+      expect(calcLimit(upgrades, 'prConcurrentLimit')).toBe(12);
+    });
+
+    it('inherits prConcurrentLimit if branchConcurrentLimit is null', () => {
+      const upgrades = partial<BranchUpgradeConfig>([
+        {
+          prHourlyLimit: 10,
+          branchConcurrentLimit: null as never,
```
viceice, line 90: use `@ts-expect-error` comment instead of `never` cast or maybe allow `null`?

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,140 @@ export function isLimitReached(key: Limit): boolean {
...
+  logger.debug(
+    {
+      limits: upgrades.map((upg) => {
+        return { depName: upg.depName, [limitName]: upg[limitName] };
+      }),
+    },
+    `${limitName} of the upgrades present in this branch`,
+  );
+
+  let lowestLimit = Number.MAX_SAFE_INTEGER;
```
viceice, line 112: move after the condition, it's only used afterwards

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,140 @@ export function isLimitReached(key: Limit): boolean {
   const { max, current } = limit;
   return max - current <= 0;
 }
+
+export type CountName = 'PullRequests' | 'HourlyPullRequests' | 'Branches';
+interface CountValue {
```
viceice, line 43: Count sounds strange, can't we use `Limit` instead? 🤔

lib/workers/repository/process/limits.ts
```diff
@@ -1,141 +1,79 @@
...
-    } catch (err) {
-      // istanbul ignore if
-      if (err instanceof ExternalHostError) {
-        throw err;
-      }
-      logger.error({ err }, 'Error checking PRs created per hour');
-      return config.prHourlyLimit;
+  try {
+    const prList = await platform.getPrList();
+    const currentHourStart = DateTime.local().startOf('hour');
```
secustor, line 14: I think we should rather use UTC time here, so we are using internally UTC and user facing local time ( e.g. schedules )

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,140 @@ export function isLimitReached(key: Limit): boolean {
...
+      limits: upgrades.map((upg) => {
+        return { depName: upg.depName, [limitName]: upg[limitName] };
+      }),
+    },
+    `${limitName} of the upgrades present in this branch`,
+  );
+
+  let lowestLimit = Number.MAX_SAFE_INTEGER;
+  if (hasMultipleLimits(upgrades, limitName)) {
+    logger.warn(
```
secustor, line 114:
```suggestion
    logger.once.debug(
```
IMO should be debug.

lib/workers/global/limits.ts
```diff
@@ -27,7 +28,7 @@ export function incLimitedValue(key: Limit, incBy = 1): void {
   });
 }
 
-export function isLimitReached(key: Limit): boolean {
+function handleCommitsLimit(key: Limit): boolean {
```
rarkins, line 31: It makes no sense to rename this `handleCommitsLimit()` but still require `key` (which can only be `Commits`). Either keep it the same name but with one possible key (in case we have new keys in future), or do the rename like this but drop the `key` param.

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,164 @@ export function isLimitReached(key: Limit): boolean {
   const { max, current } = limit;
   return max - current <= 0;
 }
+
+export type CountName = 'PullRequests' | 'HourlyPullRequests' | 'Branches';
```
rarkins, line 42:
```suggestion
export type CountName = 'ConcurrentPRs' | 'HourlyPRs' | 'Branches';
```

Make consistent, and shorter

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,164 @@ export function isLimitReached(key: Limit): boolean {
...
+interface CountValue {
+  current: number;
+}
```
rarkins, line 45: Why this abstraction/complexity? instead of it being a number?

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,164 @@ export function isLimitReached(key: Limit): boolean {
...
+type BranchLimitName =
+  | 'prHourlyLimit'
+  | 'branchConcurrentLimit'
+  | 'prConcurrentLimit';
```
rarkins, line 49:
```suggestion
  | 'branchConcurrentLimit'
  | 'prConcurrentLimit'
  | 'prHourlyLimit';
```

lib/workers/repository/update/branch/index.ts
```diff
@@ -209,9 +209,14 @@ export async function processBranch(
         };
       }
     }
+
+    logger.debug(
+      `Open PR Count: ${counts.get('PullRequests')?.current}, Existing Branch Count: ${counts.get('Branches')?.current}, Hourly PR Count: ${counts.get('HourlyPullRequests')?.current}`,
+    );
+
```
rarkins, line 216: Is this logged once per branch, regardless of whether any limits are set?

rarkins (changes requested):

lib/workers/repository/update/branch/index.ts
RahulGautamSingh: Yes

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,159 @@ export function isLimitReached(key: Limit): boolean {
...
+  counts.set(key, count);
+  logger.debug(`${key} count = ${count}`);
+}
+
+export function incCountValue(key: CountName, incBy = 1): void {
+  const count = counts.get(key) ?? 0;
+  counts.set(key, count + incBy);
+}
+
+function handleOtherLimits(
```
secustor, line 62:
```suggestion
function handleConcurrentLimits(
```
IMO this is a more fitting naming

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,159 @@ export function isLimitReached(key: Limit): boolean {
...
+    key === 'Branches' ? 'branchConcurrentLimit' : 'prConcurrentLimit';
+
+  // calculate the limits for this branch
+  const hourlyLimit = calcLimit(config.upgrades, 'prHourlyLimit');
+  const limitValue = calcLimit(config.upgrades, limitKey);
+
+  const hourlyPrCount =
+    counts.get('HourlyPRs') ??
+    // istanbul ignore next: should not happen
+    0;
```
secustor, line 76: If this is unexpected behavior, we should print an warn log message. Same with other occurences

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,159 @@ export function isLimitReached(key: Limit): boolean {
...
+  // assuming hourly limits will be lesser than concurrent ones as it's only logical
+  if (hourlyLimit === 0 || hourlyPrCount < hourlyLimit) {
+    if (limitValue === 0 || currentCount < limitValue) {
+      return false;
+    }
+
+    return true;
+  }
+
+  return true;
```
secustor, line 91:
```suggestion
// if a limit is defined ( >0 ) and limit reached return false
  if ( hourlyLimit && hourlyPrCount >= hourlyLimit {
    return false;
  }
  
  if ( limitValue && currentCount >= limitValue {
    return false;
  }  
  
  return true;
```
Better readability

lib/workers/global/limits.ts
RahulGautamSingh: It never happens as we set the count for hourlyPRs, concurrent PRs and concurrent branches before we start branch execution. But, we have to handle that case as native `Map` has return type `<T>|undefined`.

Should we still warn?

lib/workers/global/limits.ts
RahulGautamSingh: return values will be reversed

RahulGautamSingh: return values should be reversed

lib/workers/global/limits.ts
secustor: Yes, so we are seeing this in logs and have pointer to what is happening.

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,163 @@ export function isLimitReached(key: Limit): boolean {
...
+}
+
+export function setCount(key: CountName, val: number): void {
+  const count = val;
+  counts.set(key, count);
+  logger.debug(`${key} count = ${count}`);
+}
+
+export function incCountValue(key: CountName, incBy = 1): void {
+  const count = getCount(key) ?? 0;
```
RahulGautamSingh, line 68:
```suggestion
  const count = getCount(key);
```

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,163 @@ export function isLimitReached(key: Limit): boolean {
...
+  const currentCount = getCount(key);
+
+  // if a limit is defined ( >0 ) and limit reached return true ie. limit has been reached
+  if (hourlyLimit && hourlyPrCount >= hourlyLimit) {
+    return true;
+  }
+
+  if (limitValue && currentCount >= limitValue) {
+    return true;
+  }
```
viceice, line 93:
```suggestion
  const hourlyLimit = calcLimit(config.upgrades, 'prHourlyLimit');
  const hourlyPrCount = getCount('HourlyPRs');

  // if a limit is defined ( >0 ) and limit reached return true ie. limit has been reached
  if (hourlyLimit && hourlyPrCount >= hourlyLimit) {
    return true;
  }

  const limitValue = calcLimit(config.upgrades, limitKey);
  const currentCount = getCount(key);

  if (limitValue && currentCount >= limitValue) {
    return true;
  }
```

perf only compute second limit if first isn't reached

lib/workers/global/limits.ts
```diff
@@ -37,3 +38,163 @@ export function isLimitReached(key: Limit): boolean {
...
+  for (const upgrade of upgrades) {
+    let limitValue = upgrade[limitName];
+
+    // inherit prConcurrentLimit value incase branchConcurrentLimit is null
+    if (limitName === 'branchConcurrentLimit' && !is.number(limitValue)) {
+      limitValue = upgrade.prConcurrentLimit;
+    }
+
+    // istanbul ignore if: should not happen as the limits are of type number
+    if (is.null_(limitValue)) {
```
viceice, line 167:
```suggestion
    if (limitValue === null) {
```

i think this looks better

lib/workers/repository/process/limits.spec.ts
```diff
@@ -13,13 +13,13 @@ beforeEach(() => {
     onboardingBranch: 'bar/configure',
     prHourlyLimit: 2,
     prConcurrentLimit: 10,
-    branchConcurrentLimit: null,
+    branchConcurrentLimit: null as never,
```
viceice, line 16: is this realy required? i see `null` on the type 🤔

lib/workers/repository/process/limits.ts
```diff
@@ -1,141 +1,79 @@
...
-      if (err instanceof ExternalHostError) {
-        throw err;
-      }
-      logger.error({ err }, 'Error checking PRs created per hour');
-      return config.prHourlyLimit;
+  try {
+    const prList = await platform.getPrList();
+    const currentHourStart = DateTime.local().setZone('utc').startOf('hour');
+    logger.debug(
+      `Calculating PRs created so far in this hour currentHourStart=${String(currentHourStart)}`,
```
viceice, line 16:
```suggestion
      `Calculating PRs created so far in this hour currentHourStart=${currentHourStart}`,
```
number should be ok

@rarkins is the app parsing this log message for something?

viceice (commented):
It seems a big refactor is included in this PR and it would be easier to review if it would be split out.
Before we calculated and logged the remaining values, now we calculate the current count only

rarkins:
> It seems a big refactor is included in this PR and it would be easier to review if it would be split out.
Before we calculated and logged the remaining values, now we calculate the current count only

You seem to be describing the intended _feature_ / behavior change of this PR. It's not a refactor..

lib/workers/repository/process/limits.ts
rarkins: I don't think so

rarkins:
Merge after holidays

rarkins (changes requested):
Merge after holidays

MPV:
Happy new year! 🎊 
Fingers crossed for being able to try this out soon. ❤️ 🙏

rarkins (approved):

viceice (approved):

rarkins:
https://docs.renovatebot.com/mend-hosted/hosted-apps-config/#which-version-is-the-mend-renovate-app-using

PR #32565, opened 2024-11-15
fix(terraform): correct hostname regex and add tests

Jellyfrog:
## Changes
Corrects the hostname regex for terraform, by starting to match hyphens, and stop matching underscores.

## Context
The previous regex did not match hyphens in source urls, such as `git.my-server.com/` was matched as: `git.my`


Jellyfrog:
note this also fixes the regex for terragrunt which has the same problem (maybe they should share the regex somewhere?)

rarkins (dismissed):

rarkins (approved):

PR #32604, opened 2024-11-18
fix(gradle): Strip `/.` prefix when searching for `gradlew`

candrews:
## Changes

`gradleWrapperFileName` returns `./gradlew` when not running on Windows. `findUpLocal` returns unexpected results due to the `./` prefix, therefore strip the `./` prefix when calling `findUpLocal`.

## Context

This change fixes an issue where Renovate debug logs [`Found Gradle dependency lockfiles but no gradlew - aborting update`](https://github.com/renovatebot/renovate/blob/46194117afc160863d70ec0430b25e167f341bd1/lib/modules/manager/gradle/artifacts.ts#L158) even when `gradlew` exists, resulting in gradle lockfiles not being updated as expected.

### DIFF
--- a/lib/modules/manager/gradle/artifacts.ts
+++ b/lib/modules/manager/gradle/artifacts.ts
@@ -153,7 +153,7 @@ export async function updateArtifacts({
 
   const gradlewName = gradleWrapperFileName();
   const gradlewFile = await findUpLocal(
-    gradlewName.replace(new regEx(`^\\./`), ''),
+    gradlewName.replace(regEx(`^\\./`), ''),
     dirname(packageFileName),
   );
   if (!gradlewFile) {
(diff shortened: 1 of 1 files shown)


lib/modules/manager/gradle/artifacts.ts
```diff
@@ -152,7 +152,10 @@ export async function updateArtifacts({
   }
 
   const gradlewName = gradleWrapperFileName();
-  const gradlewFile = await findUpLocal(gradlewName, dirname(packageFileName));
+  const gradlewFile = await findUpLocal(
+    gradlewName.replace(new RegExp(`^\\./`), ''),
```
viceice, line 156:
```suggestion
    gradlewName.replace(new regEx(`^\\./`), ''),
```
use our regex helper

viceice (commented):
needs a test

candrews:
This was not actually a problem, sorry for the noise!

PR #32621, opened 2024-11-19
docs(managers): add missing url and displayName

risu729:
## Changes

Updates the docs of managers. Mainly adds missing URLs of the tools.

Also, it adds `displayName` if the default (title case split by `-`) isn't appropriate.

My apologies for reordering some codes. Please let me know if it's better to preserve the order.

This PR updates the codes in `lib/modules/managers/` but all changes should affect only docs.

### DIFF
--- a/lib/modules/manager/bitbucket-pipelines/index.ts
+++ b/lib/modules/manager/bitbucket-pipelines/index.ts
@@ -4,14 +4,15 @@ import { extractPackageFile } from './extract';
 
 export { extractPackageFile };
 
+export const url =
+  'https://support.atlassian.com/bitbucket-cloud/docs/get-started-with-bitbucket-pipelines/';
+export const categories: Category[] = ['ci'];
+export const urls = [
+  'https://support.atlassian.com/bitbucket-cloud/docs/bitbucket-pipelines-configuration-reference/',
+];
+
 export const defaultConfig = {
   fileMatch: ['(^|/)\\.?bitbucket-pipelines\\.ya?ml$'],
 };
 
-export const categories: Category[] = ['ci'];
-
 export const supportedDatasources = [DockerDatasource.id];
-
-export const urls = [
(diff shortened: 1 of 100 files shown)


lib/modules/manager/bitbucket-pipelines/index.ts
```diff
@@ -4,14 +4,15 @@ import { extractPackageFile } from './extract';
 
 export { extractPackageFile };
 
+export const url =
+  'https://support.atlassian.com/bitbucket-cloud/docs/get-started-with-bitbucket-pipelines/';
+export const categories: Category[] = ['ci'];
+export const urls = [
```
viceice, line 10: do we support url and urls? 🤔

lib/modules/manager/maven/index.ts
```diff
@@ -5,6 +5,9 @@ import * as mavenVersioning from '../../versioning/maven';
 export { extractAllPackageFiles } from './extract';
 export { bumpPackageVersion, updateDependency } from './update';
 
+export const url = 'https://maven.apache.org/index.html';
```
viceice, line 8:
```suggestion
export const url = 'https://maven.apache.org';
```
I would drop the index file

lib/modules/manager/meteor/index.ts
```diff
@@ -3,10 +3,11 @@ import { NpmDatasource } from '../../datasource/npm';
 
 export { extractPackageFile } from './extract';
 
+export const url = 'https://docs.meteor.com/';
```
viceice, line 6: maybe we should also drop trailing slash

lib/modules/manager/bitbucket-pipelines/index.ts
risu729: `url` is for the name, and `urls` is for the references.

`url`
image
`urls`
image

viceice (approved):

Issue #8804, opened 2021-02-22
featureRequest: preUpgradeTasks configuration option.


r-plus:
**What would you like Renovate to be able to do?**

We try to create PR for CocoaPods managed dependency, but renovate display error like this.

```
{
  "artifactErrors": [
    {
      "lockFile": "Podfile.lock",
      "stderr": "Analyzing dependencies\n[!] Unable to find the Xcode project `/mnt/renovate/gh/username/repo/repo.xcodeproj` for the target `Pods-repo`.\n"
    }
  ]
}
```

yes, we are not putting xcodeproj file in git repository.
instead, using `xcodegen` https://github.com/yonaskolb/XcodeGen to create xcodeproj file from cli before we build the project.

**Did you already have any implementation ideas?**

like the `postUpgradeTasks` https://docs.renovatebot.com/configuration-options/#postupgradetasks configuration, we want to execute arbitrary commands by `preUpgradeTasks` to prepare for renovate.

rarkins:
We don't have any timeline for supporting custom tasks in the hosted app, so you'd need to self-host once such a feature existed.

Regarding your specific need, do you know if this can even be run on Linux? From the project's install instructions:

![image](https://user-images.githubusercontent.com/6311784/108692035-6cbf7780-74fc-11eb-9751-9a6b10b96993.png)

r-plus:
thanks for your reply.

I didn't test it on Linux.
But project unit testing will execute on ubuntu-latest when every commit by github CI.
https://github.com/yonaskolb/XcodeGen/runs/1906702739

So maybe.

rarkins:
Back to the generic concept of `preUpgradeTasks`:

Ideally they are configurable per-upgrade.

We also need to consider how, where and when they are run:
- Do they need to be run before we edit the package file, or just before artifacts updating?
- Do they need to be run within the same container as the artifacts updating (e.g. to install a necessary tool inside the container) or run independently - meaning that the result of such a task needs to be visible in the repo

Leksat:
A use case from https://github.com/renovatebot/renovate/discussions/8984: Drupal project updates.

Drupal configuration is stored in Git, but updates from Drupal core or modules can modify it. To achieve completely automated updates we need:
- Pre: Install Drupal using existing configuration.
- Update dependencies.
- Post: Run Drupal updates and export the updated config.

Leksat:
For the above Drupal config use case:

[...]

PR #32657, opened 2024-11-21
feat: rename allowedPostUpgradeCommands/allowPostUpgradeCommandTemplating

rarkins:
## Changes

Rename allowedPostUpgradeCommands to allowedCommands

## Context
- #8804


viceice (commented):
Don't we need a rename migration? Otherwise LGTM

rarkins:
Yes, needs some testing on a real repo too

lib/workers/global/config/parse/env.ts
```diff
@@ -41,6 +41,8 @@ const renameKeys = {
   gitLabAutomerge: 'platformAutomerge', // migrate: gitLabAutomerge
   mergeConfidenceApiBaseUrl: 'mergeConfidenceEndpoint',
   mergeConfidenceSupportedDatasources: 'mergeConfidenceDatasources',
+  allowPostUpgradeCommandTemplating: 'allowCommandTemplating',
+  allowedPostUpgradeCommands: 'allowedCommands',
```
RahulGautamSingh, line 45: Needs tests

RahulGautamSingh (commented):
Locally tested. Cli and Env migrations work fine.

Test Runs:
[Old names](https://github.com/Rahul-renovate-testing/repro-32657-old/)
[Old names but new code](https://github.com/Rahul-renovate-testing/repro-32657-new-with-migration) ie. migrated
[New names](https://github.com/Rahul-renovate-testing/repro-32657-new-no-migration/pull)

lib/workers/global/config/parse/env.ts
rarkins: I don't think we need to test every single alias if at least one is tested

RahulGautamSingh: Better to test the migrations work via tests than not.
Thw test file already has a describe block for testing migrations.

rarkins: It's wasteful and pointless to test every single pair in this object. We should focus on testing functionality, and this is not new functionality. If the lines above work, these will work too.

rarkins:
Merge after holidays

viceice:
what's the state of this pr now?

rarkins:
Ready for review

viceice (approved):

PR #32662, opened 2024-11-21
feat(maven): Remove unnecessary HTML page fetches

zharinov:
## Changes

It doesn't seem necessary now when we're able to obtain all necessary information by querying the corresponding POM file of a particular package version.

## Context

- Ref: #27768


lib/modules/datasource/maven/index.ts
```diff
@@ -210,9 +139,7 @@ export class MavenDatasource extends Datasource {
 
     logger.debug(`Looking up ${dependency.display} in repository ${repoUrl}`);
 
-    let releaseMap = await this.fetchReleasesFromMetadata(dependency, repoUrl);
-    releaseMap = await this.addReleasesFromIndexPage(
-      releaseMap,
+    const releaseMap = await this.fetchReleasesFromMetadata(
```
Churro, line 142: I assume this could have an impact for those packages on Maven Central which have no `maven-metadata.xml`? See https://maven.apache.org/repository/central-metadata.html

To alleviate such cases, would it work to `addReleasesFromIndexPage` only if `releaseMap` is still empty after  `fetchReleasesFromMetadata` instead of doing it always? It would also mean one `index.html` query for packages that don't exist on Maven Central.

Churro (commented):
> Currently, for every package we're fetching:
> 
>     * `maven-metadata.xml`
>     * GET the latest POM file (for `homepage`/`sourceUrl`)
>     * HEAD the POM file of the selected/filtered release (often matches with GET of the latest one)
>     * [Maven Central only] HTML index page
>
> If first 3 are done correctly, seems like we don't need the latter one.

If I'm not mistaken "GET the latest POM file (for `homepage`/`sourceUrl`)" is done after populating the list of releases, i.e.., any release that is unknown at this point won't be pulled. Hence, the "[Maven Central only] HTML index page" step is actually second, not last. Do you agree?

lib/modules/datasource/maven/index.ts
zharinov: And order is different, just like you described, yes

rarkins: @zharinov do you agree with @Churro's suggestion on falling back to HTML? or is there no reason to treat Maven Central differently?

zharinov: Is there any example of an actively used and maintained package without `maven-metadata.xml`? Wouldn't it break any other tooling anyways?

We haven't noticed the lack of HTML fetch functionality when the `index.html` files were queried. I think at some point, Maven Central reconfigured their servers to not redirect `.../index.html` to `../` paths anymore, so we had 404s for some time. Or it never worked, and we shipped it like that, which I don't necessary believe we did. 

What I suggest is to check out how many keys containing `datasource-maven:index-html-releases` substring do exist uin Renovate cache right now, and drop this html functionality if possible.

Churro: I think missing `maven-metadata.xml` is not a widespread problem and mostly occurs during cross-publishing as it's the case with SBT plugins or deps that are imported from repos like sonatype:
* https://github.com/spring-cloud/spring-cloud-build/issues/223
* https://github.com/spring-projects/spring-security/issues/9988

Three weeks ago, PR #32295 has correct `index.html` fetching, so the current implementation should be functional.  If it's not needed, I'd also suggest removing it. In any case, the suggestion to check for valid cache keys sounds very reasonable to me get to more data on this.

viceice: I agree to drop this code.

has the metadata all the timestamps we currently fetch from the html?

viceice (commented):
otherwise LGTM

Churro (approved):

rarkins (approved):

PR #32677, opened 2024-11-22
feat(datasource/cpan): populate latest tag

RahulGautamSingh:
## Changes
Use the [module/{module} ](https://github.com/metacpan/metacpan-api/blob/master/docs/API-docs.md#modulemodule)endpoint to get the latest version and populate `tags[latest]` field

## Context
Ref: https://github.com/renovatebot/renovate/discussions/32670#discussioncomment-11348627

- [x] Both unit tests + ran on a real repository [Before](https://github.com/wwuck/renovate-cpanfile-itch/pull/1)  ..... [After](https://github.com/Rahul-renovate-testing/renovate-cpanfile-itch) (no pr created as we are on the latest version)

### DIFF
--- a/lib/modules/datasource/cpan/__fixtures__/Plack.json
+++ b/lib/modules/datasource/cpan/__fixtures__/Plack.json
@@ -24,6 +24,7 @@
                      "version" : "1.0048"
                   }
                ],
+               "status": "latest",
                "date" : "2020-11-30T00:21:36",
                "maturity" : "released",
                "distribution" : "Plack",
@@ -47,6 +48,7 @@
                      "name" : "Plack"
                   }
                ],
+               "status": "cpan",
                "date" : "2018-02-10T09:25:30",
                "maturity" : "released"
             },
@@ -63,6 +65,7 @@
                      "name" : "Plack"
                   }
                ],
+               "status": "cpan",
                "date" : "2018-02-10T07:52:31",
                "deprecated" : false
             },
@@ -84,6 +87,7 @@
                      "name" : "Plack"
                   }
                ],
+               "status": "cpan",
                "date" : "2017-12-31T20:42:50",
                "distribution" : "Plack",
                "download_url" : "https://cpan.metacpan.org/authors/id/M/MI/MIYAGAWA/Plack-1.0045.tar.gz"
@@ -107,6 +111,7 @@
                      "version" : "1.0044"
                   }
                ],
+               "status": "cpan",
                "date" : "2017-04-27T17:48:20",
                "maturity" : "released",
                "deprecated" : false
@@ -134,6 +139,7 @@
                      "version" : "1.0043"
                   }
                ],
+               "status": "cpan",
                "date" : "2017-02-22T03:02:05",
                "maturity" : "released",
                "distribution" : "Plack",
@@ -158,6 +164,7 @@
                      "version" : "1.0042"
                   }
                ],
+               "status": "cpan",
                "date" : "2016-09-29T05:38:42",
                "deprecated" : false
             }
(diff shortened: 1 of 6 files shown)


lib/modules/datasource/cpan/index.ts
```diff
@@ -107,8 +107,24 @@ export class CpanDatasource extends Datasource {
...
+        ).body;
+        const latestVersion = latestVersionRes.module[0].version;
+
+        if (latestVersion) {
+          result.tags ??= {};
+          result.tags.latest ??= latestVersion;
+        }
       }
     }
+
```
RahulGautamSingh, line 127:
```suggestion
```

lib/modules/datasource/cpan/index.ts
```diff
@@ -107,6 +107,21 @@ export class CpanDatasource extends Datasource {
...
+        const latestVersionUrl = joinUrlParts(
+          registryUrl,
+          'v1/module/',
+          packageName,
+        );
+        const latestVersionRes = (
+          await this.http.getJson<MetaCpanApiFile>(latestVersionUrl)
+        ).body;
+        const latestVersion = latestVersionRes.module[0].version;
```
rarkins, line 119: This needs to be more "defensive" in case data is not as we expect

viceice: it should use zod from beginning. also this causes an additional API request.

viceice (commented):
see comment

viceice (commented):
missing zod schema , as requested

lib/modules/datasource/cpan/index.ts
rarkins: It's necessary, unless it can be done with the new URL only?

viceice: should be done for every **new** request. we can leave existing as is

rarkins: I meant that getting the latest tag from the new URL is necessary. Which means two requests per package is necessary unless the new URL data can obsolete the existing URL data

lib/modules/datasource/cpan/__fixtures__/Plack.json
```diff
@@ -227,6 +237,29 @@
...
             ]
+         },
+         {
+            "_type" : "file",
+            "_source" : {
+               "deprecated" : false,
+               "maturity" : "released",
+               "module" : [
+                  {
+                     "version" : "1.0039",
```
RahulGautamSingh, line 248:
```suggestion
                     "version" : "1.0038",
```

lib/modules/datasource/cpan/index.ts
```diff
@@ -60,55 +60,47 @@ export class CpanDatasource extends Datasource {
...
-          if (!latestDistribution) {
-            latestDistribution = distribution;
-          }
+    let latestVersion: string | null = null;
+    if (releases) {
+      for (const release of releases) {
+        if (!latestDistribution) {
+          latestDistribution = release.distribution;
+        }
+        if (release.isLatest) {
```
RahulGautamSingh, line 86:
```suggestion
        if (!latestVersion && release.isLatest) {
```

lib/modules/datasource/cpan/index.ts
```diff
@@ -38,7 +38,7 @@ export class CpanDatasource extends Datasource {
     let result: ReleaseResult | null = null;
     const searchUrl = joinUrlParts(registryUrl, 'v1/file/_search');
 
-    let hits: MetaCpanApiFile[] | null = null;
+    let releases: (Release & { distribution: string })[] | null = null;
```
viceice, line 41: use new interface

viceice (approved):

PR #32690, opened 2024-11-23
feat: update atlaskit package preset for changelog retrieval

setchy:
## Changes

Update atlaskit package rule to include `sourceUrl` and `sourceDirectory` for changelog retrieval 


viceice (changes requested):
please update PR description to include reasons why this needs to be done.

this fundamentally changes the rule behavior

setchy:
> please update PR description to include reasons why this needs to be done.

updated description now that https://github.com/renovatebot/renovate/pull/32701 landed

lib/config/presets/internal/packages.ts
```diff
@@ -19,6 +19,8 @@ export const presets: Record<string, Preset> = {
   atlaskit: {
     description: 'All @atlaskit packages published by Atlassian.',
     matchPackageNames: ['@atlaskit/**'],
+    sourceDirectory: "design-system/{{ lookup (split packageName '/') 1 }}",
+    sourceUrl: 'https://bitbucket.org/atlassian/atlassian-frontend-mirror',
```
rarkins, line 23: I don't think we _set_ anything in any of these packages: presets. Until now they're purely "matching" rules which only work if extended

setchy: Thanks for the feedback

Issue #5743, opened 2020-03-18
GitLab: Two pipelines per MR


CommonGuy:
**What Renovate type are you using?**

We are using the Docker version (`renovate/renovate:19`) on https://gitlab.com. It is scheduled to run all 2 hours.

**Describe the bug**

We configured renovate with `stabilityDays: 3`. In GitLab, this creates a new job, which makes the MR wait for the specified time. So far, all good.

However, sporadically the `renovate/stability-days` job will be placed on a new pipeline instead of the existing pipeline. Note that two pipelines exist instead of one. The newer pipeline with only one job is the `renovate/stability-days` job.
![image](https://user-images.githubusercontent.com/8035999/76976197-ef76a480-6933-11ea-982a-ae5cd05cff36.png)

Usually, the `renovate/stability-days` job will be placed correctly in the existing pipeline (the job on the far right in this screenshot):
![image](https://user-images.githubusercontent.com/8035999/76976767-b5f26900-6934-11ea-80c6-3ab6e81c1bce.png)

Creating two pipelines has a huge impact: In the merge request, only the most recent pipeline is checked. So if our testing-pipeline fails, the merge request can still be merged (because the `renovate/stability-days` job passes).

**Did you see anything helpful in debug logs?**

Debug-Logs of a case:
```
DEBUG: generateBranchConfig(1) (repository=(redacted), branch=renovate/webpack-bundle-analyzer-3.x)
DEBUG: hasGroupName: false (repository=(redacted), branch=renovate/webpack-bundle-analyzer-3.x)
DEBUG: groupEligible: false (repository=(redacted), branch=renovate/webpack-bundle-analyzer-3.x)
DEBUG: useGroupSettings: false (repository=(redacted), branch=renovate/webpack-bundle-analyzer-3.x)
DEBUG: generateBranchConfig(1) (repository=(redacted), branch=renovate/major-workbox-monorepo)
DEBUG: hasGroupName: true (repository=(redacted), branch=renovate/major-workbox-monorepo)
DEBUG: groupEligible: false (repository=(redacted), branch=renovate/major-workbox-monorepo)
DEBUG: useGroupSettings: false (repository=(redacted), branch=renovate/major-workbox-monorepo)
DEBUG: config.repoIsOnboarded=true (repository=(redacted))
DEBUG: Processing 13 branches: (redacted...) renovate/webpack-bundle-analyzer-3.x (repository=(redacted))
[...]

PR #32692, opened 2024-11-23
feat(gitlab): add branch status check attemps

philippe-granet:
## Changes

* Add branch status check attemps before trying to add status check
* Allow overriding the branch status check attempts via experimental env `RENOVATE_X_GITLAB_BRANCH_STATUS_CHECK_ATTEMPS`.

## Context

* https://github.com/renovatebot/renovate/discussions/27111
* https://github.com/renovatebot/renovate/pull/23239
* https://github.com/renovatebot/renovate/issues/5743


docs/usage/self-hosted-experimental.md
```diff
@@ -50,6 +50,13 @@ The formula for the delay between attempts is `RENOVATE_X_GITLAB_MERGE_REQUEST_D
 
 Default value: `5` (attempts results in max. 13.75 seconds timeout).
 
+## `RENOVATE_X_GITLAB_BRANCH_STATUS_CHECK_ATTEMPS`
```
mikaello, line 53:
```suggestion
## `RENOVATE_X_GITLAB_BRANCH_STATUS_CHECK_ATTEMPTS`
```

Remember to fix all places

philippe-granet: Done

rarkins (changes requested):

lib/modules/platform/gitlab/index.ts
```diff
@@ -1005,21 +1008,29 @@ export async function setBranchStatus({
     options.target_url = targetUrl;
   }
 
-  if (branchSha) {
+  const retryTimes = parseInteger(
+    process.env.RENOVATE_X_GITLAB_BRANCH_STATUS_CHECK_ATTEMPTS,
+    2,
+  );
+
+  for (let attempt = 1; attempt <= retryTimes; attempt += 1) {
```
viceice, line 1016: the whole loop should have a try catch and early return on failures from gitlab API

philippe-granet:
> Could this instead be default functionality, with a large number of attempts, e.g. 10?

This will penalize projects that do not have defined pipelines, this will systematically wait 10s

philippe-granet:
> > > Could this instead be default functionality, with a large number of attempts, e.g. 10?
> > 
> > 
> > This will penalize projects that do not have defined pipelines, this will systematically wait 10s and slow down the renovate analyzes
> 
> Thanks for the info. I think I missed this from the various conversations before. Could make sure this info of when this is setting is applicable and when it's not is captured in the markdown doc in this PR? I will then decide if it needs to be a first-class option or brainstorm if we have a clever way to detect this. If this is a common problem we prefer that users don't need to set obscure experimental options.

Done in this commit :[`5a3f6d3` (#32692)](https://github.com/renovatebot/renovate/pull/32692/commits/5a3f6d304a017351f478dbcc41707c6add33b4f1)

lib/modules/platform/gitlab/index.ts
```diff
@@ -1005,21 +1008,34 @@ export async function setBranchStatus({
...
+    process.env.RENOVATE_X_GITLAB_BRANCH_STATUS_CHECK_ATTEMPTS,
+    2,
+  );
 
   try {
-    // give gitlab some time to create pipelines for the sha
-    await setTimeout(
-      parseInteger(process.env.RENOVATE_X_GITLAB_BRANCH_STATUS_DELAY, 1000),
-    );
+    for (let attempt = 1; attempt <= retryTimes; attempt += 1) {
```
rarkins, line 1017: Is this intended to skip the for loop if the value is unset (value 0)?

It's kind of a confusing way to do this, I'd prefer it is more explicit/obvious for future maintainability

docs/usage/self-hosted-experimental.md
```diff
@@ -61,6 +61,15 @@ The formula for the delay between attempts is `RENOVATE_X_GITLAB_MERGE_REQUEST_D
...
 Default value: `5` (attempts results in max. 13.75 seconds timeout).
 
+## `RENOVATE_X_GITLAB_BRANCH_STATUS_CHECK_ATTEMPTS`
+
+If set to a positive integer, Renovate will use this as the number of attempts to check branch status before trying to add a status check.
+The delay between attempts is `RENOVATE_X_GITLAB_BRANCH_STATUS_DELAY` milliseconds.
+
+Default value: `2` (attempts results in maximum 2 seconds timeout).
+
+!!! warning Increasing this value too much penalizes projects that do not have defined pipelines, Renovate will systematically wait `RENOVATE_X_GITLAB_BRANCH_STATUS_CHECK_ATTEMPTS * RENOVATE_X_GITLAB_BRANCH_STATUS_DELAY` milliseconds on these projects and slow down the Renovate analyzes.
```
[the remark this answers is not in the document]
philippe-granet: "do not have defined pipelines" means projects without `.gitlab-ci.yml` or with a Gitlab pipeline where all jobs are desactivated, for example with [rules](https://docs.gitlab.com/ee/ci/jobs/job_rules.html) that desactivate Gitlab pipeline on Renovate branch.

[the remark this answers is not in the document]
rarkins: Neither of those sound like common or best practice. Shouldn't we optimize for the normal way?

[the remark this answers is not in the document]
philippe-granet: For normal projects, with a pipeline/tests, before this MR, Renovate waits by default 1s before checking branch status, and most of time, it fails retrieving status because pipeline is often launched 1s after branch created.
With this MR, Renovate retry to check branch status 2 more times, which on my Gitlab On premise instance is sufficient. And on a very slow Gitlab instance, you can increase numbers of retry. I ran this MR on more than 100 projects for weeks on our Gitlab instance, and it works very well.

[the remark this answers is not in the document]
rarkins: So in your opinion this is already a reasonable compromise?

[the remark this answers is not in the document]
philippe-granet: Yes, on projects with no pipelines/tests (a very bad practice), theses retries will only add 2 more seconds for each Renovate branch. For example, with 20 projects and 5 dependencies to update, it adds 20*5*2=200s , which I think is reasonable.

lib/modules/platform/gitlab/index.ts
```diff
@@ -1013,21 +1016,34 @@ export async function setBranchStatus({
...
+      const commitUrl = `projects/${config.repository}/repository/commits/${branchSha}`;
+      await gitlabApi
+        .getJsonSafe(commitUrl, { memCache: false }, LastPipelineId)
+        .onValue((pipelineId) => {
+          options.pipeline_id = pipelineId;
+        });
+      if (options.pipeline_id !== undefined) {
+        break;
+      }
+      logger.debug(`Pipeline not yet created. Retrying ${attempt}`);
```
rarkins, line 1035: Does this print the first time "Retrying 0"?

lib/modules/platform/gitlab/index.spec.ts
```diff
@@ -1072,6 +1118,14 @@ describe('modules/platform/gitlab/index', () => {
...
           '/api/v4/projects/some%2Frepo/repository/commits/0d9c7726c3d628b7e28af234595cfd20febdbf8e',
         )
+        .reply(200, [])
+        .get(
+          '/api/v4/projects/some%2Frepo/repository/commits/0d9c7726c3d628b7e28af234595cfd20febdbf8e',
+        )
+        .reply(200, [])
+        .get(
+          '/api/v4/projects/some%2Frepo/repository/commits/0d9c7726c3d628b7e28af234595cfd20febdbf8e',
+        )
```
viceice, line 1128:
```suggestion
        .times(3)
```

simplify 😉

lib/modules/platform/gitlab/index.spec.ts
```diff
@@ -1099,6 +1153,14 @@ describe('modules/platform/gitlab/index', () => {
...
```
viceice, line 1163:
```suggestion
        .times(3)
```

lib/modules/platform/gitlab/index.spec.ts
```diff
@@ -1131,6 +1193,10 @@ describe('modules/platform/gitlab/index', () => {
         .get(
           '/api/v4/projects/some%2Frepo/repository/commits/0d9c7726c3d628b7e28af234595cfd20febdbf8e',
         )
+        .reply(200, [])
+        .get(
+          '/api/v4/projects/some%2Frepo/repository/commits/0d9c7726c3d628b7e28af234595cfd20febdbf8e',
+        )
         .reply(200, { last_pipeline: { id: 123 } });
```
viceice, line 1200: wait, you should not return an array if it expects an object. please fix test accordingly to return right type

viceice (approved):

rarkins (approved):

PR #32743, opened 2024-11-26
feat(manager/dockerfile): add support for Dockerfile `RUN --mount=from`

bschoenmaeckers:
## Changes

This add support for updating a docker image mount during a RUN step in a Dockerfile. I used the handling for `COPY --from` as an example.

## Context

See for all syntax supported by docker: https://docs.docker.com/reference/dockerfile/#run---mount


viceice:
please don't force-push, we do squash merge so only do it when really required for conflict resolution

viceice (approved):

PR #32746, opened 2024-11-26
fix(manager/kustomize): fix parsing kustomize resource URLs with additional query parameters

roosmaa:
## Changes

Fix extracting of resource version when there's additional query parameters present, eg:

```
https://github.com/argoproj/argo-cd/manifests/cluster-install?ref=v2.13.0&timeout=90s
```

Would currently generate the following change:

| Package | Type | Update | Change |
|---|---|---|---|
| [argoproj/argo-cd](https://redirect.github.com/argoproj/argo-cd) | Kustomization | patch | `v2.13.0&timeout=90s` -> `v2.13.1` |

Which would remove the configured timeout of 90s by accident.

## Context

Kustomize allows the following query string parameter in it's [remote resource URLs](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md):

- ref
- version
- timeout
- submodules

This PR makes renovate aware of these options and allows to correctly extract the current version in case more parameters are being used by the end-user.

### DIFF
--- a/lib/modules/manager/kustomize/extract.ts
+++ b/lib/modules/manager/kustomize/extract.ts
@@ -1,4 +1,4 @@
-import querystring from 'querystring';
+import querystring from 'node:querystring';
 import is from '@sindresorhus/is';
 import { logger } from '../../../logger';
 import { coerceArray } from '../../../util/array';
--- a/lib/modules/manager/kustomize/extract.ts
+++ b/lib/modules/manager/kustomize/extract.ts
@@ -1,3 +1,4 @@
+import querystring from 'querystring';
 import is from '@sindresorhus/is';
 import { logger } from '../../../logger';
 import { coerceArray } from '../../../util/array';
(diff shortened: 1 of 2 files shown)


lib/modules/manager/kustomize/extract.ts
```diff
@@ -50,10 +51,17 @@ export function extractResource(base: string): PackageDependency | null {
...
   }
 
-  const { path } = match.groups;
+  const { path, queryString } = match.groups;
+  const params = querystring.parse(queryString);
+  const refParam = Array.isArray(params.ref) ? params.ref[0] : params.ref;
+  const versionParam = Array.isArray(params.version)
+    ? params.version[0]
+    : params.version;
+  const currentValue = refParam ?? versionParam;
```
roosmaa, line 60: Kustomize docs allow usage of `version` query parameter as well (`ref` trumps it, though). So added it as a fallback in case `ref` parameter is missing. Since it's not related to the actual fix, let me know if you'd prefer not to support `version` parameter (for the time being).

lib/modules/manager/kustomize/extract.ts
```diff
@@ -1,3 +1,4 @@
+import querystring from 'querystring';
```
rarkins, line 1:
```suggestion
import querystring from 'node:querystring';
```

Should this have the `node:` prefix?

I'm also surprised to find that this would be our first use of this module.

lib/modules/manager/kustomize/extract.ts
```diff
@@ -50,10 +51,17 @@ export function extractResource(base: string): PackageDependency | null {
     return null;
   }
 
-  const { path } = match.groups;
+  const { path, queryString } = match.groups;
+  const params = querystring.parse(queryString);
```
rarkins, line 55: Looking at the docs, I'm pretty sure this doesn't ever throw, right?

rarkins (changes requested):

lib/modules/manager/kustomize/extract.ts
rarkins: What's the behavior now vs before if there's a query string present but it contains neither ref or version params?

rarkins (approved):

roosmaa:
Thanks for reviewing @rarkins :bow:

lib/modules/manager/kustomize/extract.ts
```diff
@@ -50,10 +51,20 @@ export function extractResource(base: string): PackageDependency | null {
...
+  const refParam = Array.isArray(params.ref) ? params.ref[0] : params.ref;
+  const versionParam = Array.isArray(params.version)
+    ? params.version[0]
+    : params.version;
```
viceice, line 59: I think we should log a warning for array case, because it's pretty sure an invalid url for kustomize.

roosmaa: Just tested with Kustomize, using the following url:

```
https://github.com/argoproj/argo-cd/manifests/cluster-install?ref=v2.13.1&timeout=90s&ref=v1
```

It didn't bat an eye, and used the `v2.13.1` for fetching the resource. I'd say the behavior matches this implementation, take the first value and ignore the rest.

PR #32772, opened 2024-11-27
feat(preset): Add axis2 monorepo group

felipecrs:
## Changes

Add axis2 monorepo preset. They all follow the exact same version btw.

https://github.com/apache/axis-axis2-java-core/tree/master/modules

### DIFF
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -571,6 +571,7 @@
     "apache-camel": "/^org.apache.camel:/",
     "aws-java-sdk": "/^com.amazonaws:aws-java-sdk-/",
     "aws-java-sdk-v2": "/^software.amazon.awssdk:/",
+    "axis2": "/^org.apache.axis2:/",
     "babel6": "/^babel6$/",
     "clarity": ["/^@cds//", "/^@clr//"],
     "embroider": "/^@embroider//",
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -55,6 +55,10 @@
       "https://github.com/awslabs/aws-sdk-rust"
     ],
     "awsappsync": "https://github.com/awslabs/aws-mobile-appsync-sdk-js",
+    "axis2": [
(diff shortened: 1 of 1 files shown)


lib/data/monorepo.json
```diff
@@ -571,6 +571,7 @@
     "apache-camel": "/^org.apache.camel:/",
     "aws-java-sdk": "/^com.amazonaws:aws-java-sdk-/",
     "aws-java-sdk-v2": "/^software.amazon.awssdk:/",
+    "axis2": "/^org.apache.axis2:/",
```
viceice, line 574: do they publish the source url? so they can be matched via source url instead of names?

viceice: yes, they do

https://github.com/apache/axis-axis2-java-core/blob/master/modules%2Fadb-codegen%2Fpom.xml#L41

felipecrs: @viceice I initially thought about it, but also thought this would be worse:

1. They could change the source URL in each individual pom.xml to point to the subdirectory of the module, this would break the preset
2. They could change it to GitHub.

I believe the preset is less likely to break if we rely on the group id.

But I can make the changes, you decide. Please let me know.

rarkins: It looks like a source URL prefix of `https://gitbox.apache.org/repos/asf` would be fairly safe, but I think you're right about them possibly changing to github in future. In that case I'm ok with it being based on the prefix

rarkins (dismissed):

lib/data/monorepo.json
viceice: we're doing it for almost all others when source url is available

viceice (commented):
see new comments

lib/data/monorepo.json
rarkins: Good point

felipecrs: Please check again.

viceice (approved):

PR #32777, opened 2024-11-28
docs(versioning:regex): mention numeric capture group requirement

mbrodala:
## Context

See https://github.com/renovatebot/renovate/discussions/13052#discussioncomment-11403472

### DIFF
--- a/lib/modules/versioning/regex/readme.md
+++ b/lib/modules/versioning/regex/readme.md
@@ -1,6 +1,7 @@
 Regular Expression Versioning is designed to be a flexible fallback versioning approach if Renovate's other versioning schemes don't do the job.
 
-The `regex` scheme makes use of Regular Express capture groups.
+The `regex` scheme makes use of Regular Expression capture groups.
+
 The valid capture groups for `regex` versioning are:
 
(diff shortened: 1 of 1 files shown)


lib/modules/versioning/regex/readme.md
```diff
@@ -11,6 +12,8 @@ The valid capture groups for `regex` versioning are:
 
 The compatibility concept was originally introduced for Docker versioning but sometimes package authors may use/misuse suffixes to mean compatibility in other versioning schemes.
 
+**Important: all capture groups must contain only purely numeric values.** So even if there is a string prefix which is identical in all available versions, it must not be part of the capture group. For example a `build` capture group containing `r4` cannot be evaluated as number; Renovate cannot compare the `build` in this case. The capture group must be `4` instead.
```
rarkins, line 15:
```suggestion
**Important: all capture groups must contain only purely numeric values.**
Even if there is a string prefix which is identical in all available versions, it must not be part of the capture group. For example a `build` capture group containing `r4` cannot be evaluated as number; Renovate cannot compare the `build` in this case. The capture group must be `4` instead.
```

mbrodala:
Notice that it probably makes sense to adjust the `regex` version to fail with an error in case `parseInt()` calls yield `NaN`. This way errors like this can be avoided. But that's a separate change.

rarkins (approved):

PR #17077, opened 2022-08-09
feat: Implement json-jsonata manager that can extract dependencies json configuration files

dariocc:
## Changes

Implements a generic json-jsonata manager (https://github.com/renovatebot/renovate/issues/15193) that is able to extract dependencies from json files.

```
{
  "jsonataManagers": [
    {
      "fileMatch": ["<file match pattern>"],
      "matchQueries": ['<query>'],
      ...
    }
  ]
}
```

Example configuration file:

```json
{
  "production": [
    {
      "version": "1.2.3",
      "package": "foo"
    }
  ],
  "development": [
    {
      "version": "4.5.6",
      "package": "bar"
    }
  ]
}
```

Query to extract dependencies:

```
production.{ "depName": package, "currentValue": version }
```

## Context

Covered in https://github.com/renovatebot/renovate/issues/15193.

Note to reviewers:

* This work is based on the modifications we keep at work from our own fork of Renovate. There it was important to minimize conflicts when syncing with upstream and duplication is a lesser problem. As consequence there is some obvious code duplication because regex managers are  _special_ and handling is spread across different places.
* It may be possible to generalize the notion of a _custom manager_ to cover _any sort of "custom" manager. This way renovate would be able to provide a _regex manager_, a _jsonata manager_, a _jq manager_, _jsonpath manger ..._  you name it... and perhaps keep manager validation and specific types within their corresponding manager module. You'll notice that some types such as `CustomExtractConfig` do not make lot of sense because now there are different _custom configs_.
* The notion of _strategy_ is not useful for this manager because the `JSONata` language is complete enough to be able to handle any json document into an array of dependencies. The readability of the expression is however dependent on each particular case and it feels that sometimes it would be useful to apply a final regex transformation to the extracted dependency fields. But I didn't want to over-complicate this work.
* As for support for applying this to yaml it would be possible to parse configuration files using different deserializers (json & yaml), chosen based on config file extension or(optionally) enumeration option under `jsonataManagers` entries. While this is simple to implement, it can be done in a separate PR if this one is accepted.
* When I did the core of this work, I considered `jq` and `JSONPath`. For `jq` (which would've been my preference due to it's ubiquity) I didn't find any maintained library that I could use. For `JSONPath` I believe the reason was that it was great _selecting_ but not so much _transforming_. `JSONata` was maintained, stable, documented, had no external dependency and looked well thought and implemented. But you'll notice that there are just 2 calls to this library code, so if you find a better fit go for it :).

## Documentation (please check one with an [x])

[...]

Issue #15193, opened 2022-04-20
Generic manager that understands json and yaml structure.


dariocc:
### What would you like Renovate to be able to do?

We are using Renovate at work and we make use of the power of the regex manager to support custom configuration formats where we store versioning information of pinned dependencies.

We've realized however that many of our configuration formats are JSON or Yaml and that we'd benefit of having language to extract dependencies that understands the structure of those documents, as regex get tricky to reason about once written and maintained.

We'd like to have a manager that is similar in concept and configuration to the regex manager but where instead of using regex to match dependencies it uses expressions that can navigate objects resulting from loading json and yaml.

### If you have any ideas on how this should be implemented, please tell us here.

We have already implemented a solution to this problem that we call _JSON manager_. Usage & configuration is analogous to `regexManager` but we replace the `matchStrings` with a `matchQueries`:

```
{
  "jsonManagers": [
    {
      "fileMatch": ["<file match pattern>"],
      "matchQueries": ['<query>'],
      ... all regex manager configuration options are supported
    }
  ]
}
```

The `matchQueries` is a collection of expressions whose goal is to transform input files into a JSON object with the following fields:

```json
{ 
    "depName": "<dep-name>",
    "packageName": "<package-name>",
    "currentValue": "<current-value>",
    "currentDigest": "<current-digest>",
    "datasource": "<data-source>",
    "versioning": "<versioning>",
    "extractVersion": "<extract-version>",
    "registryUrl": "<registry-url>",
    "depType": "<dep-type>"
}
```

We chose [JSONata expressions](https://jsonata.org/) for the query language after evaluating a number of options. We wouldn't mind a different query language if that is preferred by the Renovate team.

### Example

For the following file:

```json
{
  "production": [
    {
      "version": "1.2.3",
      "package": "foo"
    }
  ],
  "development": [
    {
      "version": "4.5.6",
      "package": "bar"
    }
  ]
}
```

We could easily update production dependencies with the following query:

```
production.{ "depName": package, "currentValue": version }
```

### Are we interested in implementing the feature ourselves?

Partially yes: We have already implemented the feature and we'd be happy to contribute it back to Renovate but only if there is some initial interest which what we are trying to asses by creating this issue. 

[...]

PR #32784, opened 2024-11-28
feat(managers/custom): generic manager for json files

RahulGautamSingh:
## Changes
- Implement a new custom manager which uses `jsonata` queries to extract the deps info from a package file. Details are similar to regex manager implementation, except that we do not need to explicitly handle the recursive matching as that is already handled by the `jsonata` query language. 
- Added tests
- Added docs
 

## Context
- Json Part of: https://github.com/renovatebot/renovate/issues/15193
- Carrying forward work done by @dariocc in https://github.com/renovatebot/renovate/pull/17077 

- [x] Both unit tests + ran on a real repository https://github.com/Rahul-renovate-testing/repro-15193


docs/usage/configuration-options.md
```diff
@@ -797,6 +797,20 @@ It will be compiled using Handlebars and the regex `groups` result.
 If `extractVersion` cannot be captured with a named capture group in `matchString` then it can be defined manually using this field.
 It will be compiled using Handlebars and the regex `groups` result.
 
+### matchQueries
+
+Each `matchQueries` must be a valid regular expression, optionally with named capture groups.
```
rarkins, line 802:
```suggestion
Each `matchQueries` must be a valid jsonata query (escaped).
```

And does it need or support more than one?

lib/config/options/index.ts
```diff
@@ -2749,6 +2749,27 @@ const options: RenovateOptions[] = [
...
+    name: 'matchStrings',
+    description:
+      'Regex capture rule to use. Valid only within a `customManagers` object.',
+    type: 'array',
+    subType: 'string',
+    format: 'regex',
+    parents: ['customManagers'],
+    cli: false,
+    env: false,
+  },
```
rarkins, line 2772: Was this missing before?

rarkins (commented):
I definitely want to see an example of this working on a real repo

docs/usage/configuration-options.md
RahulGautamSingh: Forgot to push this change 😅

RahulGautamSingh: Yes, it supports more than one

lib/config/options/index.ts
RahulGautamSingh: No, I dupicated it accidentally and was confused all this while when the docs test was failing. Thanks

RahulGautamSingh: We should have a test to catch such duplicates in the `options` object. 🤔

RahulGautamSingh:
> I definitely want to see an example of this working on a real repo

Test run on renovate repo, targeting the `package.json` and extracting `dependencies`, `devDependencies`, `optionalDependencies`
https://github.com/Rahul-renovate-testing/renovate/issues/2

docs/usage/configuration-options.md
```diff
@@ -797,6 +797,20 @@ It will be compiled using Handlebars and the regex `groups` result.
 If `extractVersion` cannot be captured with a named capture group in `matchString` then it can be defined manually using this field.
 It will be compiled using Handlebars and the regex `groups` result.
 
+### matchQueries
+
+Each `matchQueries` must be a valid regular expression(escaped).
```
viceice, line 802: regular expression? isn't it jsonata?

lib/modules/manager/custom/jsonata/index.ts
```diff
@@ -0,0 +1,52 @@
...
+export const displayName = 'Jsonata';
+
+export async function extractPackageFile(
+  content: string,
+  packageFile: string,
+  config: JsonataExtractConfig,
+): Promise<PackageFileContent | null> {
+  let json;
+  try {
+    json = JSON.parse(content);
```
viceice, line 22: use our json utils

lib/modules/manager/custom/jsonata/index.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { JSONataManagerTemplates, JsonataExtractConfig } from './types';
+import { handleMatching, validMatchFields } from './utils';
+
+export const categories: Category[] = ['custom'];
+
+export const defaultConfig = {
+  pinDigests: false,
+};
+export const supportedDatasources = ['*'];
+export const displayName = 'Jsonata';
```
viceice, line 13:
```suggestion
export const displayName = 'JSONata';
```

docs/usage/configuration-options.md
```diff
@@ -797,6 +797,20 @@ It will be compiled using Handlebars and the regex `groups` result.
 If `extractVersion` cannot be captured with a named capture group in `matchString` then it can be defined manually using this field.
 It will be compiled using Handlebars and the regex `groups` result.
 
+### matchQueries
```
viceice, line 800: can't we reuse `matchStrings`?

RahulGautamSingh: Better to not have confusion

rarkins: I'm not sure about this. I think I prefer `matchStrings` too, including because these two fields would be mutually exclusive

docs/usage/configuration-options.md
```diff
@@ -799,18 +799,29 @@ It will be compiled using Handlebars and the regex `groups` result.
 
 ### matchStrings
 
-Each `matchStrings` must be a valid regular expression, optionally with named capture groups.
+Each `matchStrings` must be one of the two:
+
+1. a valid regular expression, optionally with named capture groups
+2. a valid [JSONata](https://docs.jsonata.org/overview.html) query (escaped)
```
rarkins, line 805:
```suggestion
Each `matchStrings` must be one of the two:

1. a valid regular expression, optionally with named capture groups (if using `customType=regex`)
2. a valid, escaped [JSONata](https://docs.jsonata.org/overview.html) query (if using customType=json)
```

rarkins (changes requested):

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  function updateDependency(field: ValidMatchFields, value: string): void {
+    switch (field) {
+      case 'registryUrl':
+        // check if URL is valid and pack inside an array
+        try {
+          const url = new URL(value).toString();
+          dependency.registryUrls = [url];
+        } catch {
+          logger.warn({ value }, 'Invalid json manager registryUrl');
```
secustor, line 74:
```suggestion
          logger.warn({ value }, 'Invalid JSONata manager registryUrl');
```
This is a good example why we should look into allowing people to name custom managers again. 

If somebody is getting this warning in a bigger setup, the user will have a hard time to identify the correct manager especially with presets.

lib/modules/manager/custom/jsonata/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  packageFile: string,
+  config: JsonataExtractConfig,
+): Promise<PackageFileContent | null> {
+  let json;
+  try {
+    json = parseJson(content, packageFile);
+  } catch (err) {
+    logger.warn(
+      { err, fileName: packageFile },
+      `Error parsing '${packageFile}'`,
```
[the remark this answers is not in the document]
rarkins: Ideally we have no string interpolation in WARN or ERROR messages

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+): Promise<PackageDependency[]> {
+  // Pre-compile all JSONata expressions once
+  const compiledExpressions = config.matchStrings
+    .map((query) => {
+      try {
+        return jsonata(query);
+      } catch (err) {
+        logger.warn(
+          { err },
+          `Failed to compile JSONata query: ${query}. Excluding it from queries.`,
```
rarkins, line 36:
```suggestion
          { query, err },
          `Failed to compile JSONata query`,
```

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  function updateDependency(field: ValidMatchFields, value: string): void {
+    switch (field) {
+      case 'registryUrl':
+        // check if URL is valid and pack inside an array
+        try {
+          const url = new URL(value).toString();
+          dependency.registryUrls = [url];
+        } catch {
+          logger.warn({ value }, 'Invalid JSONata manager registryUrl');
```
rarkins, line 74:
```suggestion
          logger.warn({ url: value }, 'Invalid JSONata manager registryUrl');
```

rarkins (changes requested):
Improve warning messages (and avoid string interpolation)

lib/modules/manager/custom/jsonata/index.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+      res?.deps.filter(
+        (dep) => dep.extractVersion === 'custom-extract-version',
+      ),
+    ).toHaveLength(1);
+    expect(
+      res?.deps.filter((dep) =>
+        dep.registryUrls?.includes('https://registry.npmjs.org/'),
+      ),
+    ).toHaveLength(1);
+    expect(res?.deps.filter((dep) => dep.depType === 'dev')).toHaveLength(1);
```
viceice, line 73: Use a single `toEqual` or `toMatchObject`. will simplify this all

lib/modules/manager/custom/jsonata/index.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+      extractVersionTemplate:
+        '{{#if extractVersion}}{{extractVersion}}{{else}}default-extract-version{{/if}}',
+      registryUrlTemplate:
+        '{{#if registryUrl}}{{registryUrl}}{{else}}https://default.registry.url{{/if}}',
+      depTypeTemplate:
+        '{{#if depType}}{{depType}}{{else}}default-dep-type{{/if}}',
+    };
+    const res = await extractPackageFile(json, 'unused', config);
+
+    expect(res?.deps).toHaveLength(2);
```
viceice, line 129: see above

lib/modules/manager/custom/jsonata/index.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+export const supportedDatasources = ['*'];
+export const displayName = 'JSONata';
+
+export async function extractPackageFile(
+  content: string,
+  packageFile: string,
+  config: JsonataExtractConfig,
+): Promise<PackageFileContent | null> {
+  let json;
```
viceice, line 22:
```suggestion
  let json: unknown;
```

lib/modules/manager/custom/types.ts
```diff
@@ -1,8 +1,11 @@
+import type { JSONataManagerConfig } from './jsonata/types';
 import type { RegexManagerConfig } from './regex/types';
 
-export interface CustomExtractConfig extends Partial<RegexManagerConfig> {}
+export interface CustomExtractConfig
+  extends Partial<RegexManagerConfig>,
+    Partial<JSONataManagerConfig> {}
 
-export type CustomManagerName = 'regex';
+export type CustomManagerName = 'jsonata' | 'regex';
```
viceice, line 8:
```suggestion
export type CustomManagerName = 'json' | 'regex';
```
see other comment

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+  config: JsonataExtractConfig,
+): PackageDependency | null {
+  const dependency: PackageDependency = {};
+
+  function updateDependency(field: ValidMatchFields, value: string): void {
+    switch (field) {
+      case 'registryUrl':
+        // check if URL is valid and pack inside an array
+        try {
+          const url = new URL(value).toString();
```
viceice, line 68: use `parseUrl` helper

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+    .filter((dep) => dep !== null);
+}
+
+export function createDependency(
+  queryResult: Record<string, string>,
+  config: JsonataExtractConfig,
+): PackageDependency | null {
+  const dependency: PackageDependency = {};
+
+  function updateDependency(field: ValidMatchFields, value: string): void {
```
viceice, line 63: please no nested function

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -0,0 +1,195 @@
+With `customManagers` using `JSONata` queries you can configure Renovate so it finds dependencies in JSON files, that are not detected by its other built-in package managers.
```
viceice, line 1: Docs generation needs an update too, otherwise this page will not included.

lib/config/options/index.ts
```diff
@@ -2733,18 +2733,17 @@ const options: RenovateOptions[] = [
     description:
       'Custom manager to use. Valid only within a `customManagers` object.',
     type: 'string',
-    allowedValues: ['regex'],
+    allowedValues: ['jsonata', 'regex'],
```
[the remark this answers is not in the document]
rarkins: agree

lib/modules/manager/custom/jsonata/index.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+            "versioning": versioning,
+            "extractVersion": extract_version,
+            "registryUrl": registry_url,
+            "depType": dep_type
+        }`,
+      ],
+    };
+    const res = await extractPackageFile(json, 'unused', config);
+
+    expect(res?.deps).toHaveLength(1);
```
secustor, line 46: Why not simply use a matchObject instead?

lib/modules/manager/custom/jsonata/index.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+
+describe('modules/manager/custom/jsonata/index', () => {
+  it('has default config', () => {
+    expect(defaultConfig).toEqual({
+      pinDigests: false,
+    });
+  });
+
+  it('extracts data when no templates are used', async () => {
+    const json = `
```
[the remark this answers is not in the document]
secustor: Why has been this resolved? 
```suggestion
    const json = codeBlock`
```

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -0,0 +1,195 @@
...
+}
+```
+
+Where `<query>` is a [JSONata](https://docs.jsonata.org/overview.html) query that transform the contents into a JSON object with the following schema:
+
+```json5
+{
+  depName: '<dep-name>',
+  packageName: '<package-name>', // fallback to depName
+  currentValue: '<current-value>',
```
secustor, line 70: Shouldn't be everything outside of `currentValue` be optional?

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,99 @@
...
+        return jsonata(query);
+      } catch (err) {
+        logger.warn({ query, err }, 'Failed to compile JSONata query');
+        return null;
+      }
+    })
+    .filter((expr) => expr !== null);
+
+  // Execute all expressions in parallel
+  const results = await Promise.all(
```
secustor, line 41: This is implicitly any, parse this object with a Zod schema like we [do for the `custom` datasource](https://github.com/renovatebot/renovate/blob/2f805f4b7d746ffbe0c6509be592ae0df24979f4/lib/modules/datasource/custom/index.ts#L71)

lib/config/options/index.ts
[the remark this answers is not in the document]
secustor: I agree with @RahulGautamSingh

lib/modules/manager/custom/jsonata/readme.md
RahulGautamSingh: We need depName and datasource

secustor: How does this than relate to `depNameTemplate` and `datasourceTemplate`? 🤔

RahulGautamSingh: This codeblock is confusing, as this shows the query result .. where nothing is mandatory as al fields can be set directly using templates from the custom manager config.

lib/modules/manager/custom/jsonata/readme.md
RahulGautamSingh: Custom manager logic is already handled in `tools/docs/maangers.ts`

rarkins (dismissed):

lib/config/options/index.ts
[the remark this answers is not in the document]
ysangkok: Really looking forward to YAML support, I think it would make it possible to write a custom manager for [Hpack](https://github.com/sol/hpack?tab=readme-ov-file#dependencies).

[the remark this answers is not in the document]
rarkins: So what did we agree here? Is the customer manager type based on:
- The file format it parses (e.g. json, yaml), or
- The language/framework which is used to transform/manipulate the parsed result (e.g. jsonata, regex)?

If in future we have a file named like `filetoparse`, how would a user direct Renovate to do the first level parsing as JSON or YAML?

ysangkok:
You give the example JSONata expression `"packages.{ \"depName\": package, \"currentValue\": version }"`.

Let's say we then have two dependencies with the same version, but only one has an update:

 ```yaml
packages:
- depName: has-no-update
   version: "1.0" # we'd want the is to stay unchanged!
- depName: has-update
   version: "1.0" # we'd want this to update
```

From reading the code, it seems like this PR relies on the user to provide an `autoReplaceStringTemplate`. But how would I do that? I don't think this is obvious, it would be nice if the documentation covered this.

It seems to me that really it would be nice if there was an autoReplace variant that worked on tree-like data like JSON/YAML instead of strings. 

I'd be willing to accept if this is simply not supported yet. But I'd like to know whether this is a known limitation or not?

Apologies if you consider this off-topic. It just seems like there is an inherent parallel between changing from regex matching to tree matching, and changing from regex patching to tree patching.

File: lib/modules/manager/custom/jsonata/utils.ts [hunk 24]
```diff
@@ -0,0 +1,119 @@
...
+        // can either be a single object, an array of objects or undefined (no match)
+        let result = (await expr.evaluate(json)) ?? [];
+        if (is.emptyObject(result) || is.emptyArray(result)) {
+          return [];
+        }
+
+        result = is.array(result) ? result : [result];
+
+        QueryResultZodSchema.parse(result);
+        return structuredClone(result);
```
secustor, line 54:
```suggestion
        const parsed = QueryResultZodSchema.safeParse(result);
        if ( parsed.success ) {
          return structuredClone(parsed.data);
        }
        logger.warn({ err: parsed.error, result }, 'Returned value does not meet minimum required fields');
        return []
```

lib/modules/manager/custom/jsonata/schema.ts
```diff
@@ -0,0 +1,16 @@
+import { z } from 'zod';
+
+const DepObjectSchema = z.object({
+  currentValue: z.string().optional(),
```
secustor, line 4:
```suggestion
  currentValue: z.string(),
```
Shouldn't be `currentValue` be required based on the above?

lib/modules/manager/custom/jsonata/index.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+            "versioning": versioning,
+            "extractVersion": extract_version,
+            "registryUrl": registry_url,
+            "depType": dep_type
+        }`,
+      ],
+    };
+    const res = await extractPackageFile(json, 'unused', config);
+
+    expect(res?.deps).toHaveLength(1);
```
viceice, line 47:
```suggestion
```

drop, matcher below will do it

lib/modules/manager/custom/jsonata/index.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+      extractVersionTemplate:
+        '{{#if extractVersion}}{{extractVersion}}{{else}}default-extract-version{{/if}}',
+      registryUrlTemplate:
+        '{{#if registryUrl}}{{registryUrl}}{{else}}https://default.registry.url{{/if}}',
+      depTypeTemplate:
+        '{{#if depType}}{{depType}}{{else}}default-dep-type{{/if}}',
+    };
+    const res = await extractPackageFile(json, 'unused', config);
+
+    expect(res?.deps).toHaveLength(2);
```
viceice, line 119:
```suggestion
```
same

lib/modules/manager/custom/jsonata/index.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+      'Invalid JSONata manager registryUrl',
+    );
+  });
+
+  it('extracts multiple dependencies with multiple matchStrings', async () => {
+    const config = {
+      matchStrings: [`{"depName": "foo"}`, `{"depName": "bar"}`],
+    };
+    const res = await extractPackageFile('{}', 'unused', config);
+    expect(res?.deps).toHaveLength(2);
```
viceice, line 247: use to equal or match object

lib/modules/manager/custom/jsonata/index.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+  });
+
+  it('excludes and warns if invalid jsonata query found', async () => {
+    const config = {
+      matchStrings: ['{', `{"depName": "foo"}`, `{"depName": "bar"}`],
+    };
+    const res = await extractPackageFile('{}', 'unused', config);
+    expect(res?.deps).toHaveLength(2);
```
viceice, line 255: same

lib/modules/manager/custom/jsonata/schema.ts
```diff
@@ -0,0 +1,16 @@
...
+  datasource: z.string().optional(),
+  depName: z.string().optional(),
```
viceice, line 6: I think depName is also required?

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,119 @@
+import is from '@sindresorhus/is';
+import jsonata from 'jsonata';
+import { logger } from '../../../../logger';
+import * as template from '../../../../util/template';
+import { parseUrl } from '../../../../util/url';
+import type { PackageDependency } from '../../types';
+import { QueryResultZodSchema } from './schema';
+import type { JSONataManagerTemplates, JsonataExtractConfig } from './types';
+
+export const validMatchFields = [
```
viceice, line 10: are those the same as in regex manager? then we should share them.

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,119 @@
...
+  const compiledExpressions = config.matchStrings
+    .map((query) => {
+      try {
+        return jsonata(query);
+      } catch (err) {
+        logger.warn({ query, err }, 'Failed to compile JSONata query');
+        return null;
+      }
+    })
+    .filter((expr) => expr !== null);
```
viceice, line 39: we shouldn't skip errors and break with failure instead

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,119 @@
...
+        return [];
+      }
+    }),
+  );
+
+  // Flatten results and create dependencies
+  return results
+    .flat()
+    .map((queryResult) => {
+      return createDependency(queryResult, config);
```
viceice, line 66: use lamda and remove return and braces

lib/modules/manager/custom/jsonata/schema.ts
RahulGautamSingh: It doesn't need to be there, as we can also configure a `currentValueTemplate`

lib/modules/manager/custom/jsonata/schema.ts
RahulGautamSingh: It doesn't need to be there, as we can also configure a `depNameTemplate`

lib/modules/manager/custom/jsonata/schema.ts
secustor: What is the use case for `currentValueTemplate`?

RahulGautamSingh: In regex manager's context it is currently being used to provide the template for modifying the captured curretValue or to directly provide one.

RahulGautamSingh: It is possible/reasonable to make `currentValue`, `depName` and `datasource` manadatory for the jsonata manager..because if user wants a single value then its possible to do so directly using the query unlike regex manager.

For example:

_If we want a constant datasource_
```
query = packages.{"depName": depNameVar, "datasource": "npm"}
```
Same for `depName` & `currentValue`. And we can also modify the "depNameVar" using the functionality provided by jsonata queries.

lib/config/options/index.ts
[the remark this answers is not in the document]
rarkins: What is the answer?

rarkins (commented):
Need clarity on how yaml and others will be supported later

lib/config/options/index.ts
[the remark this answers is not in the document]
RahulGautamSingh: We agreed to base it on the language/framework which is used to transform/manipulate the parsed result. 

If a user has a `filetoparse` then if it's a json/yaml we will convert it internally and parse it to json. But lets say if its another  filetype say, xml and the user wants  to use `jsonata` then they will need to add the support for converting the xml content to json like we do for yaml. 

`jsonata` will be limited imcomparison to regex because it's primarliy made for json structure.

RahulGautamSingh:
1. If you want to do an `autoReplace` of the dep `has-no-update` only then you need to match the whole string `- package: has-no-update\n version: "1.0"` or use the `depNameTemplate="has-no-update"` so the bot can target only that.

2. `autoReplaceStringTemplate` doesn't make sense for the `jsonata` manager. I will remove it from the docs in this PR. Thanks for catching this. Whether, we would like to add support for a similar variant will depend on future demands and is good to continue in a discussion once this PR is merged.

docs/usage/configuration-options.md
```diff
@@ -777,6 +777,19 @@ Example:
...
+  "customManagers": [
+    {
+      "customType": "jsonata",
+      "matchStrings": [
+        "packages.{ \"depName\": package, \"currentValue\": version }"
+      ]
+    }
+  ]
+}
+```
```
rarkins, line 791: Isn't this - and the example above it which you copy/pasted from - an incomplete `customManagers` example? e.g. missing `fileMatch`

lib/modules/manager/custom/jsonata/utils.ts
RahulGautamSingh: What if there are multiple queries. Won't it be better t process the valid ones and suggest updates for them and log a warning for the queries that fail?

lib/modules/manager/custom/jsonata/utils.ts
RahulGautamSingh: We cannot perform the check for minimum required fields using the Zod Schema here, because we still need to process the templates. I think having a zod schema here does not makes much sense.

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -0,0 +1,182 @@
...
+#### Optional fields you can include in the resulting structure
+
+You may use any of these items:
+
+- `depType`, or a use `depTypeTemplate` config field
+- `versioning`, or a use `versioningTemplate` config field. If neither are present, Renovate defaults to `semver-coerced`
+- `extractVersion`, or use an `extractVersionTemplate` config field
+- `currentDigest`
+- `registryUrl`, or a use `registryUrlTemplate` config field. If it's a valid URL, it will be converted to the `registryUrls` field as a single-length array
+- `indentation`. It must be either empty, or whitespace only (otherwise `indentation` will be reset to an empty string)
```
ysangkok, line 45: newlines and carriage returns are whitespace too. Wouldn't it be better to restrict it tabs and spaces?

RahulGautamSingh: not sure :)

lib/modules/manager/custom/jsonata/utils.ts
secustor: As discussed in the 1:1 chat, this should be still there with a different log message to safeguard against false types.

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -0,0 +1,182 @@
...
+
+Query:
+
+```
+packages.{ "depName": package, "currentValue": version }
+```
+
+```json title="The name of the dependency and the version are in the same string"
+{
+  "packages": ["foo@1.2.3", "bar@4.5.6"]
```
ysangkok, line 160: I don't understand how this expression matches this JSON? The expression talks about an object inside `packages` but the sample has an array of strings, and no objects at all.

ysangkok: Oh, I get it, the JSON is for the *next* query. Would be better to make the sections such that this is unambiguous.

RahulGautamSingh:
Tried to make the matching logic in `handleMatching` simpler.

docs/usage/configuration-options.md
```diff
@@ -706,22 +706,22 @@ You can define custom managers to handle:
 - Proprietary file formats or conventions
 - Popular file formats not yet supported as a manager by Renovate
 
-Currently we only have one custom manager.
+Currently we only have two custom managers.
```
rarkins, line 709:
```suggestion
Currently we have two custom managers.
```

docs/usage/configuration-options.md
```diff
@@ -806,20 +825,53 @@ It will be compiled using Handlebars and the regex `groups` result.
 If `extractVersion` cannot be captured with a named capture group in `matchString` then it can be defined manually using this field.
 It will be compiled using Handlebars and the regex `groups` result.
 
+### fileFormat
+
+It specifies the syntax of the package file being managed by the custom JSONata manager.
+This setting helps the system correctly parse and interpret the configuration file's contents.
+
+Currently, only the `json` format is supported. We intend to support more formats in the future.
```
rarkins, line 833:
```suggestion
Currently, only the `json` format is supported.
```

lib/modules/manager/custom/utils.ts
rarkins, line 1: Remove all additions to this file

rarkins (dismissed):

docs/usage/configuration-options.md
```diff
@@ -806,20 +825,53 @@ It will be compiled using Handlebars and the regex `groups` result.
...
 It will be compiled using Handlebars and the regex `groups` result.
 
+### fileFormat
+
+It specifies the syntax of the package file being managed by the custom JSONata manager.
+This setting helps the system correctly parse and interpret the configuration file's contents.
+
+Currently, only the `json` format is supported.
+
+```json
```
secustor, line 835:
```suggestion
```json title="Parsing a JSON file with a custom manager"
```

docs/usage/configuration-options.md
```diff
@@ -786,6 +790,21 @@ Example:
 }
 ```
 
+```json
```
secustor, line 793:
```suggestion
```json title="Parsing a JSON file with a custom manager"
```

lib/config/options/index.ts
```diff
@@ -2734,18 +2734,27 @@ const options: RenovateOptions[] = [
...
+    allowedValues: ['json'],
     parents: ['customManagers'],
     cli: false,
     env: false,
   },
   {
     name: 'matchStrings',
     description:
-      'Regex capture rule to use. Valid only within a `customManagers` object.',
+      'Regex pattern or JSONata query to use. Valid only within a `customManagers` object.',
```
secustor, line 2755:
```suggestion
      'Queries to use. Valid only within a `customManagers` object. See `customType` docs.',
```
Maybe keep it technology neutral.

lib/modules/manager/custom/jsonata/index.spec.ts
secustor: Why you have resolved this? 🤔

lib/modules/manager/custom/jsonata/index.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  const deps = await handleMatching(json, packageFile, config);
+  if (!deps.length) {
+    return null;
+  }
+
+  return {
+    deps,
+    matchStrings: config.matchStrings,
```
secustor, line 48: Why are you returning `matchStrings`?

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    if (is.emptyObject(queryResult) || is.emptyArray(queryResult)) {
+      logger.warn(
+        {
+          jsonataQuery: query,
+          packageFile,
+        },
+        'The jsonata query returned no matches. Possible error, please check your query',
+      );
+      continue;
```
secustor, line 33: I rather think we should abort if something unexpected happens, so users trying to fix their setup do not create accidentally PRs.

lib/modules/manager/custom/jsonata/index.ts
RahulGautamSingh: `autoReplacetemplate` won't work for jsonata manager. Hence the other fields not needed.

I forgot to remove `matchStrings` too :)

lib/config/validation.ts
```diff
@@ -924,6 +930,59 @@ function validateRegexManagerFields(
   }
 }
 
+function validateJSONataManagerFields(
```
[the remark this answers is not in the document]
RahulGautamSingh: Will refactor in a separate PR

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -0,0 +1,182 @@
+With `customManagers` using `JSONata` queries you can configure Renovate so it finds dependencies in JSON files, that are not detected by its other built-in package managers.
+
+Renovate uses the `jsonata` package to process the `json` file content. Read about the [jsonata query language](https://docs.jsonata.org/overview.html) in their readme.
+
+The JSONata manager is unique in Renovate in because:
+
+- It is configurable via [JSONata](https://jsonata.org/) queries
+- It can extract any `datasource`
```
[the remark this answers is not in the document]
RahulGautamSingh: Copied regex maanger :)

lib/modules/manager/custom/types.ts
```diff
@@ -1,10 +1,15 @@
...
 
-export interface CustomManager extends Partial<RegexManagerConfig> {
+export interface CustomManager
+  extends Partial<RegexManagerConfig>,
+    Partial<JSONataManagerConfig> {
```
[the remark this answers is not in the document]
RahulGautamSingh: fileFormat & matchString is not

lib/config/validation.ts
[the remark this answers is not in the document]
RahulGautamSingh: https://github.com/renovatebot/renovate/pull/33206

lib/config/options/index.ts
```diff
@@ -2734,18 +2734,27 @@ const options: RenovateOptions[] = [
...
+    allowedValues: ['json'],
     parents: ['customManagers'],
     cli: false,
     env: false,
   },
   {
     name: 'matchStrings',
     description:
-      'Regex capture rule to use. Valid only within a `customManagers` object.',
+      'Queries to use. Valid only within a `customManagers` object. See `customType` docs.',
```
viceice, line 2755: I think the see part should be moved to the markdown file with a proper link

viceice (commented):
waiting for the refactor PR to be merged and this PR update onto

lib/modules/manager/custom/jsonata/readme.md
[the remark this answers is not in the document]
viceice: @HonkingGoose Do you have suggestions for better wording?
```suggestion
- It can be used with any `datasource`
```

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+  packageFile: string,
+  config: JsonataExtractConfig,
+): Promise<PackageDependency[]> {
+  let results: Record<string, string>[] = [];
+  const { matchStrings: jsonataQueries } = config;
+  for (const query of jsonataQueries) {
+    // won't fail as this is verified during config validation
+    const jsonataExpression = jsonata(query);
+    // this does not throw error, just returns undefined if no matches
+    let queryResult = (await jsonataExpression.evaluate(json)) ?? [];
```
viceice, line 23:
```suggestion
    let queryResult = (await jsonataExpression.evaluate(json));
```

not needed

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    if (is.emptyObject(queryResult) || is.emptyArray(queryResult)) {
```
viceice, line 25:
```suggestion
    if (!queryResult || is.emptyObject(queryResult) || is.emptyArray(queryResult)) {
```

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+      );
+      return [];
+    }
+
+    queryResult = is.array(queryResult) ? queryResult : [queryResult];
+    const parsed = QueryResultZodSchema.safeParse(queryResult);
+    if (parsed.success) {
+      results = results.concat(parsed.data);
+    } else {
+      logger.warn(
```
viceice, line 41: should this really be a warning? we debug log such things in other managers.

lib/modules/manager/custom/jsonata/index.ts
```diff
@@ -0,0 +1,49 @@
...
+): Promise<PackageFileContent | null> {
+  let json: unknown;
+  try {
+    switch (config.fileFormat) {
+      case 'json':
+        json = parseJson(content, packageFile);
+        break;
+    }
+  } catch (err) {
+    logger.warn(
```
viceice, line 30: i think we should only use debug like in other managers? 🤔 

@rarkins WDYT?

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  for (const field of validMatchFields) {
+    const fieldTemplate = `${field}Template` as keyof JSONataManagerTemplates;
+    const tmpl = config[fieldTemplate];
+    if (tmpl) {
+      try {
+        const compiled = template.compile(tmpl, queryResult, false);
+        updateDependency(field, compiled, dependency);
+      } catch {
+        logger.warn(
```
viceice, line 70: debug?

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+function updateDependency(
+  field: ValidMatchFields,
+  value: string,
+  dependency: PackageDependency,
+): PackageDependency {
+  switch (field) {
+    case 'registryUrl': {
+      const url = parseUrl(value)?.toString();
+      if (!url) {
+        logger.warn({ url: value }, 'Invalid JSONata manager registryUrl');
```
viceice, line 92:
```suggestion
        logger.debug({ url: value }, 'Invalid JSONata manager registryUrl');
```

viceice (commented):
otherwise LGTM

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -0,0 +1,182 @@
...
+      "matchStrings": ['<query>'],
+      ...
+    }
+  ]
+}
+```
+
+Where `<query>` is a [JSONata](https://docs.jsonata.org/overview.html) query that transform the contents into a JSON object with the following schema:
+
+To be effective with the JSONata manager, you should understand jsonata queries. But enough examples may compensate for lack of experience.
```
HonkingGoose, line 67:
```suggestion
Creating your Renovate JSONata manager config is easier if you understand JSONata queries.
We recommend you follow these steps:

1. Read the official JSONata query language docs
2. Check our example queries below
3. You're ready to make your own config

Alternatively you can "try and error" to a working config, by adjusting our examples.
```

How about something like this? I think we want to nudge newcomers to read the JSONata docs, then read our examples, and only then make their own config.

I expect most users prefer to "try and error their way into a working config" instead of reading the docs.

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+Query:
+
+```
+*.{ "depName": package, "currentValue": version }
+```
+
+```json title="The dependency name is in a JSON node name and the version is in a child leaf to that node"
```
HonkingGoose, line 120:
```suggestion
```json title="The dependency name is in a JSON node name, and the version is in a child leaf to that node"
```

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+Query:
+
+```
+$each(function($v, $n) { { "depName": $n, "currentValue": $v.version } })
+```
+
+```json title="The name of the dependency and the version are both value nodes of the same parent node"
```
HonkingGoose, line 137:
```suggestion
```json title="The dependency name and its version are both value nodes of the same parent node"
```

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+Query:
+
+```
+packages.{ "depName": package, "currentValue": version }
+```
+
+```json title="The name of the dependency and the version are in the same string"
```
HonkingGoose, line 158:
```suggestion
```json title="The dependency name and version are part of the same string"
```

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+Query:
+
+```
+$map($map(packages, function ($v) { $split($v, "@") }), function ($v) { { "depName": $v[0], "currentVersion": $v[1] } })
+```
+
+```json title="JSONata manager config to extract deps from package.json file in the renovate repository"
```
HonkingGoose, line 170:
```suggestion
```json title="JSONata manager config to extract deps from a `package.json` file in the Renovate repository"
```

HonkingGoose (changes requested):

docs/usage/configuration-options.md
```diff
@@ -706,22 +706,22 @@ You can define custom managers to handle:
...
+Currently we have two custom managers.
 The `regex` manager which is based on using Regular Expression named capture groups.
+The `jsonata` manager which is based on using JSONata queries.
 
-You must have a named capture group matching (e.g. `(?<depName>.*)`) _or_ configure its corresponding template (e.g. `depNameTemplate`) for these fields:
+You must capture/extract the following three fields _or_ configure its corresponding template (e.g. `depNameTemplate`) for these fields:
 
 - `datasource`
 - `depName` and / or `packageName`
 - `currentValue`
```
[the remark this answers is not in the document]
RahulGautamSingh: I think the existing one is better.

docs/usage/configuration-options.md
```diff
@@ -706,22 +706,22 @@ You can define custom managers to handle:
 - Proprietary file formats or conventions
 - Popular file formats not yet supported as a manager by Renovate
 
-Currently we only have one custom manager.
+Currently we have two custom managers.
 The `regex` manager which is based on using Regular Expression named capture groups.
+The `jsonata` manager which is based on using JSONata queries.
```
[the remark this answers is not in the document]
RahulGautamSingh: Matching Engine sounds good.

docs/usage/configuration-options.md
```diff
@@ -706,22 +706,22 @@ You can define custom managers to handle:
...
 
-You must have a named capture group matching (e.g. `(?<depName>.*)`) _or_ configure its corresponding template (e.g. `depNameTemplate`) for these fields:
+You must capture/extract the following three fields _or_ configure its corresponding template (e.g. `depNameTemplate`) for these fields:
 
 - `datasource`
 - `depName` and / or `packageName`
 - `currentValue`
 
-Use named capture group matching _or_ set a corresponding template.
 We recommend you use only _one_ of these methods, or you'll get confused.
```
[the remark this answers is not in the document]
RahulGautamSingh: > For example, if you use a template for the first field, then use a template for the other fields too.

Sometimes one might have to combine both methods, so a recommendation should be fine here.

lib/config/options/index.ts
```diff
@@ -2734,18 +2734,26 @@ const options: RenovateOptions[] = [
...
     parents: ['customManagers'],
     cli: false,
     env: false,
   },
   {
-    name: 'matchStrings',
+    name: 'fileFormat',
     description:
-      'Regex capture rule to use. Valid only within a `customManagers` object.',
+      'It specifies the syntax of the package file being managed by the custom JSONata manager.',
```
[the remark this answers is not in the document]
RahulGautamSingh: I prefer the older one here.

lib/config/options/index.ts
```diff
@@ -2734,18 +2734,26 @@ const options: RenovateOptions[] = [
...
+    type: 'string',
+    allowedValues: ['json'],
+    parents: ['customManagers'],
+    cli: false,
+    env: false,
+  },
+  {
+    name: 'matchStrings',
+    description: 'Queries to use. Valid only within a `customManagers` object.',
```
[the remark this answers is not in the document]
RahulGautamSingh: We use valid term quite often in this file, so it will be more regular with the rest of the descriptions to use `valid`

docs/usage/configuration-options.md
```diff
@@ -706,22 +706,25 @@ You can define custom managers to handle:
...
 
-We recommend that you also tell Renovate what `versioning` to use.
+We also recommend that you also tell Renovate what `versioning` to use.
 If the `versioning` field is missing, then Renovate defaults to using `semver` versioning.
```
viceice, line 725: this isn't fully true. the datasource can set a different versioning (most data sources doing it). we should suggest this only if the datasource default versioning isn't suitable

lib/modules/manager/custom/jsonata/index.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+        '{{#if extractVersion}}{{extractVersion}}{{else}}default-extract-version{{/if}}',
+      registryUrlTemplate:
+        '{{#if registryUrl}}{{registryUrl}}{{else}}https://default.registry.url{{/if}}',
+      depTypeTemplate:
+        '{{#if depType}}{{depType}}{{else}}default-dep-type{{/if}}',
+    };
+    const res = await extractPackageFile(json, 'unused', config);
+
+    expect(res).toMatchObject({
+      deps: [
```
viceice, line 141: add templates here because they should returned too?

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -0,0 +1,203 @@
...
+- Capture the `currentValue` of the dependency _or_ use the `currentValueTemplate` template field
+- Capture the `depName` or `packageName`. _Or_ use a template field: `depNameTemplate` and `packageNameTemplate`
+- Capture the `datasource`, _or_ use the `datasourceTemplate` template field
+
+#### Optional fields you can include in the resulting structure
+
+You may use any of these items:
+
+- `depType`, _or_ use the `depTypeTemplate` template field
+- `versioning`, _or_ the use `versioningTemplate` template field. If neither are present, Renovate defaults to `semver-coerced`
```
viceice, line 43: defaults to datasource default

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -0,0 +1,108 @@
...
+  const { matchStrings: jsonataQueries } = config;
+  for (const query of jsonataQueries) {
+    // won't fail as this is verified during config validation
+    const jsonataExpression = jsonata(query);
+    // this does not throw error, just returns undefined if no matches
+    let queryResult = await jsonataExpression.evaluate(json);
+
+    if (
+      !queryResult ||
+      is.emptyObject(queryResult) ||
```
viceice, line 27: do we really need to check it here? wouldn't it fail later anyway 🤔

docs/usage/configuration-options.md
HonkingGoose, line 725: Copying `@viceice`'s comment:

> this isn't fully true. the datasource can set a different versioning (most data sources doing it). we should suggest this only if the datasource default versioning isn't suitable

What do you think of this text?

```suggestion
Renovate defaults to `semver` versioning if _both_ condition are met:

- The `versioning` field is missing in the custom manager config
- The Renovate datasource does _not_ set its own default versioning
```

lib/modules/manager/custom/jsonata/index.spec.ts
RahulGautamSingh: We do not return the templates in this manager as autoreplace possible. 

And, the templates where being copied over in the regex manager and returned as they were needed for autoreplace.

docs/usage/configuration-options.md
viceice: I need to check, but it probably uses `semver-coerced` as fallback 🤔

secustor: Yes, it is `semver-coerced`.

HonkingGoose: I updated my suggestion so it says `semver-coerced`. 😉

lib/modules/manager/custom/jsonata/utils.ts
RahulGautamSingh: An early return with a some logs gives better visibility of this issue, hence I added it.

RahulGautamSingh: But, it should allow empty objects to pass, as someone could've configured templates for all the fields ie. currentValue, datasource and packageName.

viceice (approved):

viceice:
@RahulGautamSingh can you add yaml support right after this is merged? otherwise i would do it because i need yaml support 😉

RahulGautamSingh:
> @RahulGautamSingh can you add yaml support right after this is merged? otherwise i would do it because i need yaml support 😉

Sure

dariocc:
Thanks for taking over the work and finishing the feature @RahulGautamSingh!

PR #32813, opened 2024-11-29
refactor(maven): Unified result type for http fetch

zharinov:


lib/modules/datasource/maven/util.ts
```diff
@@ -312,25 +333,21 @@ export async function downloadMavenXml(
...
-        xml: new XmlDocument(body),
-        isCacheable: !res.authorization,
-      };
-    }
+    const rawResult = await downloadHttpProtocol(http, pkgUrl);
+    const xmlResult = rawResult.transform(({ isCacheable, data }): MavenXml => {
+      const xml = new XmlDocument(data);
+      return { isCacheable, xml };
+    });
+    return xmlResult.unwrapOr({});
```
viceice, line 341: does it no longer throw if invalid XML?

viceice: can you please log a message for invalid XML. eg if a user configured something wrong for the registry url?

viceice (commented):
see comment

lib/modules/datasource/maven/util.ts
zharinov: I'll have better handling of it it in further PR

zharinov: Could we live about a week without XML handling?

viceice (approved):
please add XML parse error logging as soon as possible

PR #32814, opened 2024-11-29
refactor(maven): Unified result type for s3 fetch

zharinov:


lib/modules/datasource/maven/util.ts
```diff
@@ -142,42 +144,97 @@ function isS3NotFound(err: Error): boolean {
...
+
+      return Result.err({ type: 'unknown', err });
+    })
+    .onError((err) => {
+      const failedUrl = pkgUrl.toString();
+
+      if (err.type === 'credentials-error') {
+        logger.debug(
+          { failedUrl },
+          'Maven S3 lookup error: credentials provider error, check "AWS_ACCESS_KEY_ID" and "AWS_SECRET_ACCESS_KEY" variables',
```
viceice, line 212: why not directly log this on the `catch` callback above? would make the code simpler.

rarkins (dismissed):

viceice (commented):
otherwise LGTM

lib/modules/datasource/maven/util.ts
zharinov: It's better to have side effects in `.onError` handler, though yeah may be sometimes it's too much

viceice: I usually agree, but in this case it makes conde more complicate to maintain

zharinov: Actually I agree. Logging is a reasonable exception from "separate transforms and side effects" rule, otherwise it often will be a duplicate code structure just for logging statements.

lib/modules/datasource/maven/util.ts
```diff
@@ -334,10 +379,12 @@ export async function downloadMavenXml(
...
-    const res = await downloadS3Protocol(pkgUrl);
-    if (res) {
-      return { xml: new XmlDocument(res) };
-    }
+    const rawResult = await downloadS3Protocol(pkgUrl);
+    const xmlResult = rawResult.transform(({ isCacheable, data }): MavenXml => {
+      const xml = new XmlDocument(data);
+      return { xml };
+    });
+    return xmlResult.unwrapOr({});
```
viceice, line 387: can we catch and log XML parse errors?

zharinov: Will be done in future, when some function will combine both http and s3

viceice (approved):

PR #32817, opened 2024-11-29
feat(gerrit): Reduce email notifications

felipecrs:
## Changes

This PR will reduce the number of email notifications sent as result of some Renovate operation when using the Gerrit platform.

Here are the operations that will no longer send email notifications:

- When rebasing a change (i.e. pushing a new patchset)
- When auto-approving (not auto-merging) a change (i.e. voting with Code-Review +2)
- When posting a comment/message (e.g. when posting pr body)

Here are the operations that will still send email notifications:

- When adding someone as a reviewer of a change

  However this operation was refactored to avoid calling one Gerrit API for each reviewer, instead it calls the API once with all reviewers. This also means that the email notification will be sent only once, instead of one email per reviewer.

Unrelated, but this PR also adds a comment about an API that was removed by Gerrit 3.8 (`/assignee`). It doesn't fix the problem but at least makes us aware of it.

### DIFF
--- a/lib/modules/platform/gerrit/client.ts
+++ b/lib/modules/platform/gerrit/client.ts
@@ -114,7 +114,7 @@ class GerritClient {
     const message = this.normalizeMessage(fullMessage);
     await this.gerritHttp.postJson(
       `a/changes/${changeNumber}/revisions/current/review`,
-      { body: { message, tag } },
+      { body: { message, tag, notify: 'NONE' } },
     );
   }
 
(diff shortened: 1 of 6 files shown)


rarkins (approved):

lib/modules/platform/gerrit/client.ts
```diff
@@ -149,18 +149,25 @@ class GerritClient {
...
     );
   }
 
-  async addReviewer(changeNumber: number, reviewer: string): Promise<void> {
-    await this.gerritHttp.postJson(`a/changes/${changeNumber}/reviewers`, {
-      body: { reviewer },
-    });
+  async addReviewers(changeNumber: number, reviewers: string[]): Promise<void> {
+    await this.gerritHttp.postJson(
+      `a/changes/${changeNumber}/revisions/current/review`,
```
viceice, line 158: this breaks older Gerrit versions. since when this API exists? 😨

viceice (changes requested):

lib/modules/platform/gerrit/client.ts
felipecrs: I use Gerrit 3.4, of which this API works.

image

What is the version of Gerrit you are using?

IIRC this endpoint is pretty old. Maybe the adding reviewers functionality isn't.

felipecrs: @viceice, installed Gerrit 3.0 with `docker run --rm -p 8080:8080 gerritcodereview/gerrit:3.0.16` and its REST API docs still says it's supported:

image

But I surely did not mean to break this functionality for anyone. And it is working here, I tested for real with Gerrit 3.4 several times.

Can you please watch your Renovate logs to see if it prints anything when it fails?

viceice: ok, the readme says we support all 3.x versions, so please ensure compatibility if possible.

otherwise update readme and raise the minimum supported version.

felipecrs: @viceice, maybe I did not make myself clear:

This API **is supported** by all Gerrit 3.x versions.

If you have noticed that it broke, it is unexpected, and I would like to debug it.

viceice: This PR is ok, i don't use Gerrit. But we like to ensure to not break users if possible (even this platform is alpha grade).

Just make sure do add some extended descriptions to the PR for changed API and validate the mentioned supported versions are not broken. If it needs to break, then also update the readme to reflect it. also add a description why we need to break it.

Hopefully this is more clear now 🤗

felipecrs: @viceice, understood.

Be sure I would have made it explicit in case of a breaking change.

I am positive this API change did not break any existing user of Renovate.

viceice (approved):

PR #32818, opened 2024-11-29
fix(gerrit): not auto-merging if Code-Review changed from +2 to +1

felipecrs:
## Changes

The `checkIfApproved` function was not working as expected when someone already had published a Code-Review +1 vote. It was wrongly assuming that the change was approved in such case, while the change should only be approved if the change had a Code-Review +2 vote.

Note that when `o=DETAILED_LABELS` is set, the change response from Gerrit API no longer contains the content from `o=LABELS`, meaning it should only ever be used in this case (when looking for the approvals), otherwise it would break existing code.

This fix also ensures that the change will be auto-approved again during the auto-merge process, in case the Code-Review +2 was changed to +1 by some CI check using the same account as Renovate, like SonarQube (which posts Code-Review +1).

Finally, this fix will improve the approval logic for when the change is initially created or updated. It will avoid two API calls in each case.

## Context

I self-host Renovate and I noticed that auto-merge was not working as expected. I realized it was caused by this issue, because for a given project we have here, the same user/robot also votes the change with a Code-Review +1 vote when SonarQube analysis is successful (which is a very known pattern).


rarkins (dismissed):

viceice (commented):
conflicted

felipecrs:
I just realized this function is also implemented by the GitLab provider, and there it doesn't check for whether it's approved already or not, meaning it will probably issue an error, which is not desired.

I'll think about it.

felipecrs:
Please note I've refactored the other two platforms (Azure and GitLab) that also use `autoApprove` to account for these improvements as well:

1. As last resort, the PR will be attempted to be auto-approved before auto-merge, in case the approval got dismissed after it was initially published for whatever reason.
2. It will no longer attempt to auto-approve if the PR is already approved

Please not I do not use Azure or GitLab. I coded these functions based on their documentation. But I believe there is no big risk involved, given it will still attempt to auto-approve in case the _is approved_ condition doesn't match.

felipecrs:
I've retitled the PR to reflect the broader impact of it.

lib/workers/repository/update/pr/automerge.ts
```diff
@@ -69,6 +69,12 @@ export async function checkAutoMerge(
       prAutomergeBlockReason: 'PlatformNotReady',
     };
   }
+  // Usually the PR will already be approved, this is a last resort in case the
+  // approval was lost for some reason
+  if (config.autoApprove && platform.approvePr) {
+    logger.debug('Auto-approving PR if needed before automerge');
+    await platform.approvePr(pr.number);
+  }
```
rarkins, line 77: Will this cause each PR to be approved on each run? i.e. one extra API call per PR per run?

felipecrs: > Will this cause each PR to be approved on each run?

Only if they are not yet approved.

> i.e. one extra API call per PR per run?

Per auto-merge attempt, there will be one extra API call to verify whether the PR is not yet approved, and one more API call to approve it in case it's not yet approved.

Just remember:

1. this is only applicable for azure, gitlab, and gerrit
2. and for people who have `autoApprove: true`
3. and all other conditions in `automerge.ts` already passed (except `branchStatus`, which depends on the approval to be green)

felipecrs: But if you believe it's not a good idea, I can find another way.

rarkins (changes requested):

felipecrs:
@rarkins, it's presumptuous from my part that this would be of any help to the existing platforms. I will revert the changes to Azure and GitLab, and focus on Gerrit, which is what I use and know this would make sense for.

lib/modules/platform/gerrit/client.ts
```diff
@@ -72,10 +72,13 @@ class GerritClient {
...
-  async getChange(changeNumber: number): Promise<GerritChange> {
+  async getChange(
+    changeNumber: number,
+    extraOptions?: string[],
+  ): Promise<GerritChange> {
+    const options = [...this.requestDetails, ...(extraOptions ?? [])];
     const changes = await this.gerritHttp.getJson<GerritChange>(
-      `a/changes/${changeNumber}?` +
-        this.requestDetails.map((det) => `o=${det}`).join('&'),
+      `a/changes/${changeNumber}?` + options.map((det) => `o=${det}`).join('&'),
```
viceice, line 81: use `getQueryString` for proper quoting

https://github.com/renovatebot/renovate/blob/2f49607fecb80757a1d2a3a5ee26dd0204c14cd3/lib/util/url.ts#L72

lib/modules/platform/gerrit/client.ts
```diff
@@ -196,15 +199,11 @@ class GerritClient {
...
-  }
-
-  wasApprovedBy(change: GerritChange, username: string): boolean | undefined {
+    const change = await client.getChange(changeId, ['DETAILED_LABELS']);
+    const reviewLabel = change?.labels?.['Code-Review'];
     return (
-      change.labels?.['Code-Review'].approved &&
-      change.labels['Code-Review'].approved.username === username
+      reviewLabel === undefined ||
+      Boolean(reviewLabel.all?.some((label) => label.value === 2))
```
viceice, line 206:
```suggestion
      reviewLabel.all?.some((label) => label.value === 2) === true
```

lib/modules/platform/gerrit/index.spec.ts
```diff
@@ -187,28 +187,6 @@ describe('modules/platform/gerrit/index', () => {
       gerrit.writeToConfig({ labels: {} });
     });
 
-    it('updatePr() - auto approve enabled', async () => {
```
viceice, line 190: why so many removed tests? I see missing coverage

lib/workers/repository/update/pr/automerge.spec.ts
```diff
@@ -124,6 +124,19 @@ describe('workers/repository/update/pr/automerge', () => {
       expect(platform.mergePr).toHaveBeenCalledTimes(0);
     });
 
+    it('should attempt to auto-approve if supported', async () => {
+      config.autoApprove = true;
+      config.automerge = true;
+      config.pruneBranchAfterAutomerge = true;
+      platform.approvePrForAutomerge.mockResolvedValueOnce();
```
viceice, line 131:
```suggestion
```

not needed, default for a mock

lib/modules/platform/gerrit/scm.ts
```diff
@@ -129,19 +129,16 @@ export class GerritScm extends DefaultGitScm {
         hasChanges = await git.hasDiff('HEAD', 'FETCH_HEAD'); //avoid empty patchsets
       }
       if (hasChanges || commit.force) {
+        const pushOptions = ['notify=NONE'];
+        if (commit.autoApprove) {
+          pushOptions.push('l=Code-Review+2');
```
viceice, line 134: is this a faster ways for `client.approveChange` ?

lib/modules/platform/gerrit/index.spec.ts
felipecrs: Because `approveChange` no longer runs in `updatePr` function, this test is now obsolete. Also, `wasApprovedBy` function no longer exists because it is no longer needed, thus I removed tests of it.

lib/modules/platform/gerrit/scm.ts
felipecrs: Not only faster, but it also saves 2 API calls to Gerrit on each run.

lib/modules/platform/gerrit/index.spec.ts
felipecrs: I will fix coverage. I had not realized it had decreased. Thank you.

felipecrs:
@viceice, sorry for taking so long. I thought it would be harder to address your comments, but they were in fact straightforward and helpful. Thank you very much.

felipecrs:
@viceice, this is a kind ping. :)

viceice (commented):
conflicted

lib/workers/repository/update/pr/automerge.ts
```diff
@@ -69,6 +69,10 @@ export async function checkAutoMerge(
       prAutomergeBlockReason: 'PlatformNotReady',
     };
   }
+  if (config.autoApprove && platform.approvePrForAutomerge) {
+    logger.debug('Auto-approving PR for automerge');
+    await platform.approvePrForAutomerge(pr.number);
```
viceice, line 74: it this a noop if it's already approved? see other open conversation

viceice (commented):
otherwise LGTM

lib/workers/repository/update/pr/automerge.ts
felipecrs: But the PR could have been un-approved after initial approve. See the PR body's _Context_:

> I self-host Renovate and I noticed that auto-merge was not working as expected. I realized it was caused by this issue, because for a given project we have here, the same user/robot also votes the change with a Code-Review +1 vote when SonarQube analysis is successful (which is a very known pattern).

SonarQube analysis in such case happens during CI, after Renovate has initially approved the PR upon creation or update.

This extra-check will account for cases where the PR had its vote downgraded from +2 to +1.

felipecrs: I believe this is a special case that can only happen with Gerrit. On GitHub and Gitlab, there's no such concept of approval level. It's either approved or not, by a given user.

However, in Gerrit, "approvals" can either be +2 (highest, needed to merge), or +1 (not enough to merge). And a user, can have its vote decreased from +2 to +1.

lib/workers/repository/update/pr/automerge.ts
felipecrs: The implementation was changed so that the approving logic for when the PR is created or updated will no longer do an extra API call to check on whether it's approved already or not (it's surely not yet approved in such case).

In fact, it will even prevent a second _approve_ API call, as now the approval request is passed during `git push` as a push option.

lib/workers/repository/update/pr/automerge.ts
felipecrs: @viceice, I believe this was the last concern on this PR.

Have you had a chance to re-think it?

I can probably break this PR into two PRs by the way. One for the saving API call improvement, and another one fixing the issue reported.

felipecrs:
I no longer think this PR is a good idea.

It works around a real problem (the SonarQube example at PR description), but at the cost of 1 extra API call per PR. Or even 2, if the problem is found.

Instead, this should be fixed in the source: SonarQube should not vote in the same label as Renovate (`Code-Review`), it should be configured to vote in its own label (e.g. `Sonar-Verified`).

That said, I have extracted the good bits of this PR and included it into https://github.com/renovatebot/renovate/pull/35315.

PR #32844, opened 2024-12-02
feat(replacements): framer-motion to motion

yutakobayashidev:
## Changes

Created a replacement rule for `framer-motion` with its continuation package `motion`.

## Context

The package has been published as `motion` since version 11.11.12 following the merger between Framer Motion and Motion One.

Proof:
* Package documentation: https://www.npmjs.com/package/motion
* Official announcement: https://motion.dev/blog/framer-motion-is-now-independent-introducing-motion

### DIFF
--- a/lib/data/replacements.json
+++ b/lib/data/replacements.json
@@ -16,6 +16,7 @@
       "replacements:eslint-plugin-vitest-to-scoped",
       "replacements:fakerjs-to-scoped",
       "replacements:fastify-to-scoped",
+      "replacements:framer-motion-to-motion",
       "replacements:hapi-to-scoped",
       "replacements:jade-to-pug",
       "replacements:joi-to-scoped",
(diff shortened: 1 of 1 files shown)


lib/data/replacements.json
```diff
@@ -670,6 +671,18 @@
...
   },
+  "framer-motion-to-motion": {
+    "description": "`framer-motion` and Motion One have merged under the package name `motion`, though `framer-motion` is still being published. `motion` is now the recommended name.",
+    "packageRules": [
+      {
+        "matchCurrentVersion": ">=11.11.12",
+        "matchDatasources": ["npm"],
+        "matchPackageNames": ["framer-motion"],
+        "replacementName": "motion",
+        "replacementVersion": "11.11.12"
```
rarkins, line 682: This is not ideal because it could result in a _rollback_ of versions.

I think that while @mattgperry is releasing both packages in parallel, maybe Renovate shouldn't proactively replace `framer-motion` with `motion`. If/once `framer-motion` is stopped then we no longer have a moving target for the replacement

rarkins (changes requested):

lib/data/replacements.json
viceice: I think we should support replace name only. so we can simply leave the replacement version undefined and retain the current version for such use cases

rarkins: OK so requires a modification to the replacement logic to accept empty replacement version?

viceice: I think so. @zharinov do you know?

zharinov: Seems like so

rarkins:
Blocked by https://github.com/renovatebot/renovate/issues/33033

PR #32845, opened 2024-12-02
feat: apply `ignorePresets` when resolving the `globalExtends` array

Gabriel-Ladzaretti:
## Changes
Respect self hosted `ignorePresets` config option when resolving the `globalExtends` array.

## Context
- https://github.com/renovatebot/renovate/discussions/32596#discussioncomment-11318670

### DIFF
--- a/lib/workers/global/index.ts
+++ b/lib/workers/global/index.ts
@@ -96,10 +96,11 @@ export async function validatePresets(config: AllConfig): Promise<void> {
 
 export async function resolveGlobalExtends(
   globalExtends: string[],
+  ignorePresets?: string[],
 ): Promise<AllConfig> {
   try {
     // Make a "fake" config to pass to resolveConfigPresets and resolve globalPresets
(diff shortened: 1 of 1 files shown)


lib/workers/global/index.ts
```diff
@@ -133,10 +134,13 @@ export async function start(): Promise<number> {
     await instrument('config', async () => {
       // read global config from file, env and cli args
       config = await getGlobalConfig();
-      if (config?.globalExtends) {
+      if (is.nonEmptyArray(config?.globalExtends)) {
```
viceice, line 137: why this change?

Gabriel-Ladzaretti: because for some reason it is an empty array even when not explicitly defined, this if condition is basically always truthy.
this has been my experience when debugging this section at least.

viceice (approved):

rarkins (approved):

PR #32870, opened 2024-12-03
refactor(bundler): refactor extraction regexes

Djiit:
## Changes:

Break the line regex into mutliple one. Instead of capturing all the line at once, we ensure it's a `gem` line, then apply other regexes sequentially. A few benefits:

- Readability
- Support randomly ordered arguments (e.g. `gem 'foo', source: baz, require: 'bar'` and `gem 'foo', require: 'bar', source: baz` are now parsed the same way

Current tests are still passing.

## How I've tested my work (please tick one)

- [x] Newly added unit tests, or
- [ ] Unit tests + ran on a real repository

### DIFF
--- a/lib/modules/manager/bundler/__snapshots__/extract.spec.ts.snap
+++ b/lib/modules/manager/bundler/__snapshots__/extract.spec.ts.snap
@@ -542,12 +542,15 @@ exports[`modules/manager/bundler/extract extractPackageFile() parse mastodon Gem
       },
     },
     {
+      "currentDigest": "54b17ba8c7d8d20a16dfc65d1775241833219cf2",
       "currentValue": "'~> 0.6'",
-      "datasource": "rubygems",
+      "datasource": "git-refs",
       "depName": "http_parser.rb",
       "managerData": {
         "lineNumber": 57,
       },
+      "packageName": "https://github.com/tmm1/http_parser.rb",
+      "sourceUrl": "https://github.com/tmm1/http_parser.rb",
     },
     {
       "currentValue": "'~> 1.3'",
@@ -1606,7 +1609,7 @@ exports[`modules/manager/bundler/extract extractPackageFile() parses rails Gemfi
       },
     },
     {
-      "currentValue": "">= 3.0.5", "< 3.2"",
+      "currentValue": "">= 3.0.5"",
       "datasource": "rubygems",
       "depName": "listen",
       "lockedVersion": "3.1.5",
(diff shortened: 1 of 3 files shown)


lib/modules/manager/bundler/__snapshots__/extract.spec.ts.snap
```diff
@@ -542,12 +542,15 @@ exports[`modules/manager/bundler/extract extractPackageFile() parse mastodon Gem
       },
     },
     {
+      "currentDigest": "54b17ba8c7d8d20a16dfc65d1775241833219cf2",
       "currentValue": "'~> 0.6'",
-      "datasource": "rubygems",
```
Djiit, line 546: Why this change? We were missing this one because of an un-handled line order. This refactor allow finer extraction, and allow for such fixes.

lib/modules/manager/bundler/extract.spec.ts
```diff
@@ -171,14 +171,21 @@ describe('modules/manager/bundler/extract', () => {
   it('parses inline source in Gemfile', async () => {
     const sourceInlineGemfile = codeBlock`
       baz = 'https://gems.baz.com'
+      gem 'inline_gem'
       gem "inline_source_gem", source: 'https://gems.foo.com'
       gem 'inline_source_gem_with_version', "~> 1", source: 'https://gems.bar.com'
       gem 'inline_source_gem_with_variable_source', source: baz
+      gem "inline_source_gem_with_require_after", source: 'https://gems.foo.com', require: %w[inline_source_gem]
```
Djiit, line 178: Added a few combination to ensure futures changes won't break any known ordering.

lib/modules/manager/bundler/__snapshots__/extract.spec.ts.snap
```diff
@@ -1606,7 +1609,7 @@ exports[`modules/manager/bundler/extract extractPackageFile() parses rails Gemfi
       },
     },
     {
-      "currentValue": "">= 3.0.5", "< 3.2"",
+      "currentValue": "">= 3.0.5"",
```
rarkins, line 1612: This seems like a regression?

Djiit: Good catch. I updated the regex and added a dedicated test.

rarkins (approved):

secustor (approved):

PR #32903, opened 2024-12-04
fix(schedule): use and logic to handle dow+dom

RahulGautamSingh:
## Changes
Use the `legacyMode:false` option so the `Cron` job is scheduled when both DOM and DOW match and not when either matches.

## Context
- Ref: https://github.com/renovatebot/renovate/discussions/32879


docs/usage/configuration-options.md
```diff
@@ -3852,6 +3852,13 @@ You could then configure a schedule like this at the repository level:
 This would mean that Renovate can run for 7 hours each night, plus all the time on weekends.
 Note how the above example makes use of the "OR" logic of combining multiple schedules in the array.
 
+<!-- prettier-ignore -->
+!!! note
+    When both day of week and day of month are restricted in the schedule.
+    The behavior is to run when both the day of the month and day of week match
+    Example:
+    `* * 1-7 * 4` this schedule will be triggered on a Thursday which falls in the first week of the month.
```
HonkingGoose, line 3860:
```suggestion
    If both the day of the week _and_ the day of the month are restricted in the schedule, then Renovate only runs when both the day of the month _and_ day of the week match!
    For example: `* * 1-7 * 4` means Renovate only runs on the _first_ Thursday of the month.
```

How about this?

lib/workers/repository/update/branch/schedule.spec.ts
```diff
@@ -299,15 +299,17 @@ describe('workers/repository/update/branch/schedule', () => {
       });
     });
 
-    describe('complex cron schedules', () => {
+    describe('handles schedule with DOM and DOW using AND logic', () => {
```
HonkingGoose, line 302:
```suggestion
    describe('handles schedule with Day Of Month and Day Of Week using AND logic', () => {
```

Expanding the abbreviation is easier to read.

HonkingGoose (changes requested):

rarkins (approved):

PR #32930, opened 2024-12-05
refactor: Flatten package lookup queue

### LINKED ISSUE #6575: Refactor: separate lookup from filter+decide stages
Currently Renovate does lookup+filter+decide logic per-dependency. Instead, I'd prefer to:
- Flatten extracted dependencies into a single-dimensioned array
- Look up (via datasources) all dependencies simultaneously (but with rate limits)
- Once lookups are complete, perform the filter+decide step

Separating like this makes more logical sense, and will also make it easier in future in case we want to optimize such as:
- Caching the entire lookup result and re-using it if running multiple times in a row, or
- Querying a lookup cache with a single query (list of all packages) instead of one cache query per package

zharinov:
## Changes

Reorganize task fetching to be performed using a single array of `Promise` objects

## Context

- Closes: #6575
- Similar to #27369, but prevents #30906 by using `WeakMap` for detecting exact `PackageFile` object to update (because it's possible to have multiple `PackageFile`s with single name thanks to regex manager)


lib/workers/repository/process/fetch.ts
```diff
@@ -17,27 +17,43 @@ import { Result } from '../../../util/result';
...
+type LookupResult = Result<PackageDependency, Error>;
+
+interface LookupTaskResult {
+  packageFile: PackageFile;
+  result: LookupResult;
+}
+
+type LookupTask = Promise<LookupTaskResult>;
+
+async function lookup(
```
viceice, line 31: can you split these `Result` refactor to a separate PR? so this gets smaller and easier to review.

zharinov: This part is intact from the latest review, I just reverted the revert. The signature of function has changed, so I'll need modify upstream code just to match new signature, and then we'll remove it anyways. Let's not split this one.

viceice (approved):
hopefully oh won't break again 🫣🤞

florianmutter:
Seems to have broken renovate for us. See https://github.com/renovatebot/renovate/discussions/30906#discussioncomment-11594604

PR #32963, opened 2024-12-07
fix(hex): Relax case in typing of schema

bernardo-martinez:
## Changes

In hex datasource schema, change the meta.links to accept either `Github` or `GitHub`

## Context
https://github.com/renovatebot/renovate/discussions/29993#discussioncomment-9944037

### DIFF
--- a/lib/modules/datasource/hex/schema.ts
+++ b/lib/modules/datasource/hex/schema.ts
@@ -8,11 +8,21 @@ export const HexRelease = z
     html_url: z.string().optional(),
     meta: z
       .object({
-        links: z.record(z.unknown()).transform(links => Object.fromEntries(
-          Object.entries(links).map(([key, value]) => [key.toLowerCase(), value]),
-        )).pipe(z.object({
-          github: z.string(),
-        })),
+        links: z
+          .record(z.unknown())
+          .transform((links) =>
+            Object.fromEntries(
+              Object.entries(links).map(([key, value]) => [
+                key.toLowerCase(),
+                value,
+              ]),
+            ),
+          )
+          .pipe(
+            z.object({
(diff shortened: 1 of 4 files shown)


zharinov (changes requested):

zharinov (approved):

bernardo-martinez:
dunno why this was not auto merged if it has been approved 🤔

lib/modules/datasource/hex/schema.ts
```diff
@@ -8,9 +8,21 @@ export const HexRelease = z
     html_url: z.string().optional(),
     meta: z
       .object({
-        links: z.object({
-          Github: z.string(),
-        }),
+        links: z
+          .record(z.unknown())
```
viceice, line 12: why use unknown here? is there anything else besides strings allowed? 🤔

bernardo-martinez: you are right, values are expected to be strings [ref](https://github.com/hexpm/hexpm/blob/47893d95cfe6024c85bf08007d6b6a41ee6bbfe1/lib/hexpm/repository/package_metadata.ex#L9), i will make the change

zharinov: Because it's irrelevant at the first stage of the validation and internally I think it'll skip unnecessary type checks for values for all keys we don't care about anyways

bernardo-martinez: ok. Anyway i assume that the type I added does not affect at all and probably gives more clarity to the future reader?

zharinov: Yeah, I don't mind this change

viceice (approved):

PR #32966, opened 2024-12-07
feat(datasource/azure-pipelines-tasks): Azure DevOps API based datasource

bdovaz:
## Changes

Initially as I do not have much experience of the Renovate API, I tried (looking at other code examples in this repository) that only changes the behavior for the case that you are running it from `platform: azure` because this way there is no need to create any new variable and it can use the values of `RENOVATE_PLATFORM`, `RENOVATE_ENDPOINT` and `RENOVATE_TOKEN`.

This PR does not solve the case that platform is a value different to `azure` (in that case it makes fallback to what it was already doing) because for that case I suppose that new variables will be needed and I consider that this can go in another future PR because it requires more analysis.

## Context

The need comes from the fact that we currently take as a source of data the static files that are generated from this repository: https://github.com/renovatebot/azure-devops-marketplace

The problem is that task updates are not propagated at the same speed to all Azure Devops organizations so there are times when renovate mistakenly suggests updating to versions of tasks that do not exist in the organization and if you do not know why it happens it is very confusing.

There is an open discussion: https://github.com/renovatebot/renovate/discussions/24820


lib/modules/datasource/azure-pipelines-tasks/index.ts
```diff
@@ -22,13 +25,70 @@ export class AzurePipelinesTasksDatasource extends Datasource {
...
-      (await this.getTasks(MARKETPLACE_TASKS_URL))[packageName.toLowerCase()];
+    const platform = GlobalConfig.get('platform');
+    const endpoint = GlobalConfig.get('endpoint');
 
-    if (versions) {
-      const releases = versions.map((version) => ({ version }));
-      return { releases };
+    if (platform === 'azure' && endpoint) {
+      const auth = Buffer.from(
+        `renovate:${process.env.RENOVATE_TOKEN}`,
```
rarkins, line 33: It's not certain that RENOVATE_TOKEN is defined. Instead, use `hostRules` to get the token

rarkins (changes requested):
Are you sure that this can replace the existing functionality in a backwards-compatible way?

How efficient is this approach? e.g. how many API calls and how many MB should it require each run?

Any caching possible?

lib/modules/datasource/azure-pipelines-tasks/index.ts
bdovaz: I have used the GlobalConfig and hostRules based on other data sources but I don't know if I have done it right.

I don't see any API documentation, I've only seen the pages you recommend here: https://github.com/renovatebot/renovate/blob/main/.github/contributing.md#code

bdovaz:
> Are you sure that this can replace the existing functionality in a backwards-compatible way?
> 
> How efficient is this approach? e.g. how many API calls and how many MB should it require each run?
> 
> Any caching possible?

Doesn't the caching already do it with the code that was there? 

https://github.com/bdovaz/renovate/blob/620fa90803e995e404e29c478b1cfcd32ad028f8/lib/modules/datasource/azure-pipelines-tasks/index.ts#L81

And the backwards-compatible thing is for when these conditions are not met that it at least does what it did before:

https://github.com/bdovaz/renovate/blob/620fa90803e995e404e29c478b1cfcd32ad028f8/lib/modules/datasource/azure-pipelines-tasks/index.ts#L36

rarkins:
So it caches once per 24 hours, _if_ the user has persistent datasource caching? (note: most people running Renovate in pipelines do not have such persistence)

I don't understand your answer to my backwards-compatibility question. Is it replacing existing logic, or adding to it? If it's replacing it, is it fully backwards compatible?

How many API calls and how many MB should it require?

bdovaz:
> So it caches once per 24 hours, _if_ the user has persistent datasource caching? (note: most people running Renovate in pipelines do not have such persistence)

@rarkins is there any documentation on how to implement persistence? Without documentation or examples I will hardly be able to do it since it is the first time I do a PR. As I said, the persistence there was already implemented, I haven't touched anything in this PR.

> I don't understand your answer to my backwards-compatibility question. Is it replacing existing logic, or adding to it? If it's replacing it, is it fully backwards compatible?

If these conditions are met: https://github.com/renovatebot/renovate/pull/32966/files#diff-2ef2acf656366f4dd2036d71f106b4d8d0b41cd546e1349942a4e40efffde808R36

It will use the new logic, otherwise it will do what it did before. The difference is that if you work against Azure DevOps and provide a PAT (personal access token), it will resolve the task versions according to your organization and not to a static file as now (https://github.com/renovatebot/azure-devops-marketplace/blob/main/azure-pipelines-builtin-tasks.json). **Not all organizations propagate task updates at the same time** and **with the current implementation the problem is that it suggests updates that have not yet been propagated in your organization** (= do not exist yet) and it is very confusing.

In the discussion of where this PR comes from that I mention in my first post of this PR https://github.com/renovatebot/renovate/discussions/24820 puts you in context of what I am trying to solve. One of the keys is here: https://github.com/renovatebot/azure-devops-marketplace/issues/42#issuecomment-2242327738

> How many API calls and how many MB should it require?

The API (https://dev.azure.com/{org}/_apis/distributedtask/tasks/) resolves which are the latest versions of each task so that this call can be cached and used equally for all subsequent requests. In the tests that I add you can see an example of response.

The MB depends on each organization, because in my organization I can have 5 third party extensions and another organization 50. This API lists the latest versions of the official/built-in tasks (same for all organizations) + those of third parties/marketplace (these are the ones that can not be measured, it depends on each organization).

Example API result in my organization (built-in + third party installed): 1.93 MB

lib/modules/datasource/azure-pipelines-tasks/index.ts
```diff
@@ -22,13 +26,53 @@ export class AzurePipelinesTasksDatasource extends Datasource {
   async getReleases({
     packageName,
   }: GetReleasesConfig): Promise<ReleaseResult | null> {
-    const versions =
-      (await this.getTasks(BUILT_IN_TASKS_URL))[packageName.toLowerCase()] ??
-      (await this.getTasks(MARKETPLACE_TASKS_URL))[packageName.toLowerCase()];
+    const platform = GlobalConfig.get('platform');
+    const endpoint = GlobalConfig.get('endpoint');
+    const { token } = hostRules.find({
+      hostType: AzurePipelinesTasksDatasource.id,
```
rarkins, line 32: I'm not sure this is set, you  might want to drop this line

lib/modules/datasource/azure-pipelines-tasks/index.ts
```diff
@@ -22,13 +26,53 @@ export class AzurePipelinesTasksDatasource extends Datasource {
...
+      url: endpoint,
+    });
 
-    if (versions) {
-      const releases = versions.map((version) => ({ version }));
-      return { releases };
+    if (platform === 'azure' && endpoint && token) {
+      const auth = Buffer.from(`renovate:${token}`).toString('base64');
```
rarkins, line 37: username doesn't matter for this endpoint?

lib/modules/datasource/azure-pipelines-tasks/types.ts
rarkins, line 1: Best to use zod parsing to ensure the structure matches what we expect

rarkins (changes requested):

lib/modules/datasource/azure-pipelines-tasks/index.ts
bdovaz: Do you mean the line to search by “hostType”? If I remove it I think it won't work, if you see the implementation of that find both the “hostType” and “url” fields have to be defined, unless I'm wrong, as I say, I'm new to TS: https://github.com/renovatebot/renovate/blob/98693e2e6797053b06b409a94f26400348e19631/lib/util/host-rules.ts#L133

lib/modules/datasource/azure-pipelines-tasks/types.ts
rarkins: @zharinov could you guide? Or even PR?

bdovaz: @rarkins done! I've figured out what the whole zod thing is about.

lib/modules/datasource/azure-pipelines-tasks/index.ts
bdovaz: @rarkins can you answer please?

lib/modules/datasource/azure-pipelines-tasks/index.ts
bdovaz: @rarkins can you answer please?

lib/modules/datasource/azure-pipelines-tasks/index.ts
bdovaz: @rarkins can you answer please?

lib/modules/datasource/azure-pipelines-tasks/index.ts
```diff
@@ -39,8 +85,49 @@ export class AzurePipelinesTasksDatasource extends Datasource {
...
     return body;
   }
+
+  @cache({
+    namespace: `datasource-${AzurePipelinesTasksDatasource.id}`,
+    key: (url: string) => url,
+    ttlMinutes: 24 * 60,
+  })
+  async getFallbackTasks<ResT>(url: string): Promise<ResT> {
+    const { body } = await this.http.getJson<ResT>(url);
```
zharinov, line 103: Our http functions accept schema object, so internally it will check against it and return the proper type, so you don't have to specify generic

bdovaz: I was going to do it but I don't know how to convert Record<string, string[]> to a schema.

That code you mention is not from this PR, we are talking about the fallback that already existed: https://github.com/renovatebot/azure-devops-marketplace/blob/main/azure-pipelines-builtin-tasks.json

I have simply split the methods in 2 so that “the new” uses schemas.

zharinov: Use `z.record(z.string().array())`

bdovaz: Done, thanks!

bdovaz:
@rarkins @zharinov I think it's all done, I just need you to answer some doubts you had to see if any more changes are needed or not.

Besides, I don't understand this workflow, it fails by adding files to the tests? https://github.com/renovatebot/renovate/actions/runs/12457890306/job/34782564100

rarkins:
We prefer inline test fixtures instead of dedicated file fixtures, so the test checks for that. It's possible to override and still merge

bdovaz:
> We prefer inline test fixtures instead of dedicated file fixtures, so the test checks for that. It's possible to override and still merge

I suppose that in this case, as it stands, it is correct and the inline is only for when there is little content.

lib/modules/datasource/azure-pipelines-tasks/index.ts
rarkins: My reading of the linked article is that username is used and significant. I believe the `${0}` is username: `string.Format("{0}:{1}", "", personalaccesstoken))));`

bdovaz: If you notice the example is literal putting those double quotes is not a placeholder to replace it with the username.

I have a lot of experience with the use of Azure DevOps APIs and I can tell you that the user is not used.

Think that the PAT (personal access token) carries that information implicitly it is not the user's password which in that case of course it would need the username.

rarkins (approved):

lib/modules/datasource/azure-pipelines-tasks/index.ts
rarkins: RENOVATE_TOKEN is not mandatory. Better to do a hostRules search as the token will have been added to hostRules

rarkins (changes requested):
Needs RENOVATE_TOKEN fix

bdovaz:
> Needs RENOVATE_TOKEN fix

I just replied to you in the comment.

rarkins (approved):

bdovaz:
@rarkins I see that it has been removed from the “merge queue” because of the “Undesirable Code” check but as we already commented, although ideally it is better that the input of the tests is inline, in this case it is a large json file and so I guess it will be justified but there is no way to pass that test, I guess you will have to ignore it?

PR #32974, opened 2024-12-08
refactor: Rearrange const in the `lookupUpdates` function

zharinov:

### DIFF
--- a/lib/workers/repository/process/lookup/index.ts
+++ b/lib/workers/repository/process/lookup/index.ts
@@ -754,8 +754,6 @@ export async function lookupUpdates(
         rollbackPrs: config.rollbackPrs,
         isVulnerabilityAlert: config.isVulnerabilityAlert,
         updatePinnedDependencies: config.updatePinnedDependencies,
-        unconstrainedValue:
-          !!config.lockedVersion && is.undefined(config.currentValue),
         err,
       },
       'lookupUpdates error',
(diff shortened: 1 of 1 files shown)


rarkins (dismissed):

lib/workers/repository/process/lookup/index.ts
```diff
@@ -752,7 +754,8 @@ export async function lookupUpdates(
         rollbackPrs: config.rollbackPrs,
         isVulnerabilityAlert: config.isVulnerabilityAlert,
         updatePinnedDependencies: config.updatePinnedDependencies,
-        unconstrainedValue,
+        unconstrainedValue:
+          !!config.lockedVersion && is.undefined(config.currentValue),
```
viceice, line 758: is the const out of scope now?

zharinov: Yes

rarkins: What's the benefit of this change? We're now left with non-DRY code. Will that be fixed later?

zharinov: Just removed it completely for simplicity. If anyone ever care about this logged field, this could be inferred from two others that still exist.

viceice (approved):

rarkins (approved):

PR #32984, opened 2024-12-09
feat(mix): handle hex package aliases

### LINKED ISSUE #32972: Custom package name for Hex dependency
### Discussed in https://github.com/renovatebot/renovate/discussions/32834

<sup>Originally posted by **bernardo-martinez** December  1, 2024</sup>
### Tell us more.

In Mix in elixir you can setup a dependency with a different name than the default one fetched from hex.pm, see: https://hex.pm/docs/usage#options

For example:
```
{:gun, "~> 2.0.0", hex: :grpc_gun},
```

This will fetch the `grpc_gun` dependency from hex.pm located here: https://hex.pm/packages/grpc_gun but will treat it locally as `gun`

**If you have any ideas on how this should be implemented, please tell us here.**

I think the mix manager should have an additional regex to search for such `hex:` option and use it to set `PackageDependency.sourceUrl` accordingly, which then would be used in the hex versioning?

**Is this a feature you are interested in implementing yourself?**
Yes, just need some discussion and guidance first
</div>

Reproduction: https://github.com/renovate-reproductions/32834/pull/1

bernardo-martinez:
## Changes

Enables mix manager to handle the `hex` option described [here](https://hex.pm/docs/usage#options)

## Context

- closes #32972 


lib/modules/manager/mix/extract.ts
```diff
@@ -44,6 +45,9 @@ export async function extractPackageFile(
         const ref = refRegexp.exec(opts)?.groups?.value;
         const branchOrTag = branchOrTagRegexp.exec(opts)?.groups?.value;
         const organization = organizationRegexp.exec(opts)?.groups?.value;
+        const hex =
+          hexRegexp.exec(opts)?.groups?.strValue ??
+          hexRegexp.exec(opts)?.groups?.atomValue;
```
rarkins, line 50: Both of the tests added seem like the same syntax to me. What's strValue vs atomValue? Is there an example of that?

bernardo-martinez: I did the example [here](https://github.com/bernardo-martinez/renovate-reproduction-32834/blob/main/mix.exs#L19). As you can see the option accepts either a string or an atom (atoms in elixir being _interned strings_)

viceice, line 50:
```suggestion
        const hexGroups = hexRegexp.exec(opts)?.groups;
        const hex =
          hexGroups?.strValue ??
          hexGroups?.atomValue;
```
optimize

bernardo-martinez: good catch thanks!

bernardo-martinez:
hey guys, something else you see that is missing here?

rarkins (approved):

viceice (approved):

viceice:
please avoid merging from main if not conflictrd

PR #32991, opened 2024-12-09
fix(platform/azure): Add current project id as sourceRepositoryId to azure getPrList

jessebluemr:
... to avoid that the bot needs read permissions to all PR source repositories

## Changes

This PR adds a second `searchCriteria` to the Azure getPrList call. The added `searchCriteria` is the `sourceRepositoryId` with the id of the current project. This has the effect that only pull requests are returned, which have their sources in branches of the current project or in other words no pull requests from forks are returned.

## Context

This PR tries to solve the problem that the user account associated with "Renovate-Git-Token" requires read access to any source repository from which pull requests are created. 
Without the read access currently the getPrList call fails and an exception like this is reported:

```plain
"err": {
             "message": "Cannot read properties of null (reading 'length')",
             "stack": "TypeError: Cannot read properties of null (reading 'length')\n    at Proxy.getPrList (/home/AzDevOps/work/1/s/node_modules/renovate/lib/modules/platform/azure/index.ts:270:25)\n    at processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at finalizeRepo (/home/AzDevOps/work/1/s/node_modules/renovate/lib/workers/repository/finalize/index.ts:25:18)\n    at Object.renovateRepository (/home/AzDevOps/work/1/s/node_modules/renovate/lib/workers/repository/index.ts:113:7)\n    at attributes.repository (/home/AzDevOps/work/1/s/node_modules/renovate/lib/workers/global/index.ts:206:11)\n    at start (/home/AzDevOps/work/1/s/node_modules/renovate/lib/workers/global/index.ts:191:7)\n    at /home/AzDevOps/work/1/s/node_modules/renovate/lib/renovate.ts:19:22"
           },
```

I'm not sure if the restriction to PRs only from the current repository has unwanted sideeffects.
Maybe the PR can be changed to be a solution strategie if the fetching of all PRs fails?
Is there a requirement for renovate to see all available PRs including forks?

### DIFF
--- a/lib/modules/platform/azure/index.ts
+++ b/lib/modules/platform/azure/index.ts
@@ -259,7 +259,11 @@ export async function getPrList(): Promise<AzurePr[]> {
     do {
       fetchedPrs = await azureApiGit.getPullRequests(
         config.repoId,
-        { status: 4 },
+        {
+          status: 4,
+          // fetch only prs directly created on the repo and not by forks
+          sourceRepositoryId: config.project,
+        },
         config.project,
         0,
         skip,


rarkins:
This seems reasonable because I don't think Renovate can create PRs from forks on Azure anyway. We would need someone to run this past a real repository and vouch that it works before merging though.

rarkins:
@jessebluemr can you verify this works on a real repo when run locally?

jessebluemr:
@rarkins  Thats what I did before opening this pull request. Do you need some screenshot or the like, as a confirmation?

rarkins (approved):

PR #32995, opened 2024-12-09
fix(presets/packages): include @types/react-dom in react preset

secustor:
## Context
The `packages:react` preset does not currently match the `@types/react-dom` package
https://github.com/secustor/backstage-plugins/pull/499

### DIFF
--- a/lib/config/presets/internal/packages.ts
+++ b/lib/config/presets/internal/packages.ts
@@ -135,7 +135,7 @@ export const presets: Record<string, Preset> = {
   },
   react: {
     description: 'All React packages.',
-    matchPackageNames: ['@types/react', 'react**'],
+    matchPackageNames: ['@types/react**', 'react**'],
   },
   stylelint: {
     description: 'All Stylelint packages.',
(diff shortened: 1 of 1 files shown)


rarkins:
Do the other packages in there already have synchronized versioning? Adding an unsynchronized package can make PRs immortal

secustor:
The preset would already match packages like `react-router-dom`, which are out of sync with react itself.

lib/config/presets/internal/packages.ts
```diff
@@ -135,7 +135,7 @@ export const presets: Record<string, Preset> = {
   },
   react: {
     description: 'All React packages.',
-    matchPackageNames: ['@types/react', 'react**'],
+    matchPackageNames: ['@types/react**', 'react**'],
```
viceice, line 138: this probably lead to false positives when a package starts with `react`. this rule is not limited to npm datasource

viceice (approved):
ok, didn't found a false positive match on npmjs for `react**` on first 12 pages.

PR #32999, opened 2024-12-09
feat(maven): Unify fetching utilities

zharinov:
## Changes

Replace `checkResource` and `checkS3Resource` with updated `downloadMaven` function which supports everything we need: timestamps, private registry detection and, well, the content itself


lib/modules/datasource/maven/index.spec.ts
```diff
@@ -719,13 +712,22 @@ describe('modules/datasource/maven/index', () => {
         s3mock.reset();
       });
 
+      function body(input: string): never {
```
viceice, line 715: isn't there a better way than this never? eg use `Buffer.from`?

viceice (commented):
can we split this into smaller parts? it's pretty hard to understand all changes.

and I also think this is a feature because you drop the head requests. that doesn't look like a refactor to me. it's at least a fix.

refactor PRs are not causing releases, so they accumulate

package.json
```diff
@@ -269,6 +269,7 @@
     "@openpgp/web-stream-tools": "0.1.3",
     "@renovate/eslint-plugin": "file:tools/eslint",
     "@semantic-release/exec": "6.0.3",
+    "@smithy/util-stream": "3.3.2",
```
zharinov, line 272: https://github.com/m-radzikowski/aws-sdk-client-mock?tab=readme-ov-file#s3-getobjectcommand

lib/modules/datasource/maven/index.spec.ts
```diff
@@ -719,13 +713,22 @@ describe('modules/datasource/maven/index', () => {
         s3mock.reset();
       });
 
+      function body(input: string) {
```
viceice, line 716: doesn't typescript or eslint complain bout missing return type?

viceice (dismissed):

lib/modules/datasource/maven/index.spec.ts
rarkins: I think TS can imply return types now?

zharinov: In the tests I believe it doesn't

viceice (dismissed):

viceice (commented):
please add a test or a comment to the coverage ignore

zharinov:
- Blocks #34949
- Blocks #34948

rarkins (dismissed):

lib/modules/datasource/maven/util.ts
```diff
@@ -369,44 +279,51 @@ export function getMavenUrl(
...
-      return { isCacheable, xml };
-    });
-    return xmlResult.unwrapOr({});
+  url: URL,
+): Promise<MavenFetchResult> {
+  const protocol = url.protocol.slice(0, -1);
+
+  let result: MavenFetchResult = Result.err({ type: 'unsupported-protocol' });
+
+  if (protocol === 'http' || protocol === 'https') {
```
viceice, line 290: use our is http url helper?

lib/modules/datasource/maven/util.ts
```diff
@@ -369,44 +279,51 @@ export function getMavenUrl(
...
-  if (protocol === 'http:' || protocol === 'https:') {
-    const rawResult = await downloadHttpProtocol(http, pkgUrl);
-    const xmlResult = rawResult.transform(({ isCacheable, data }): MavenXml => {
-      const xml = new XmlDocument(data);
-      return { isCacheable, xml };
-    });
-    return xmlResult.unwrapOr({});
+  url: URL,
+): Promise<MavenFetchResult> {
+  const protocol = url.protocol.slice(0, -1);
```
viceice, line 286: I wouldn't slice here (perf)

lib/modules/datasource/maven/util.ts
```diff
@@ -458,13 +375,14 @@ async function getSnapshotFullVersion(
...
-  /* v8 ignore next 3 -- hard to test */
-  if (!mavenMetadata) {
-    return null;
-  }
+  const metadataXmlResult = await downloadMavenXml(http, metadataUrl);
 
-  return extractSnapshotVersion(mavenMetadata);
+  return metadataXmlResult
+    .transform(({ data }) => {
+      const nullErr = { type: 'snapshot-extract-error' };
```
viceice, line 382: inline variable?

lib/modules/datasource/maven/util.ts
zharinov: I'm pretty sure this is just an integer decrement under the hood, but okay

viceice (approved):

PR #33058, opened 2024-12-12
refactor(logger): improve readability and modularity

Gabriel-Ladzaretti:
## Context
This improves modularity and usability in other Renovate wrapper projects.


Gabriel-Ladzaretti:
hmm, i most certainly didnt add any snapshots :thinking: 
https://github.com/renovatebot/renovate/actions/runs/12293666908/job/34306816832?pr=33058

lib/logger/once.spec.ts
```diff
@@ -55,7 +56,7 @@ describe('logger/once', () => {
 
   describe('logger', () => {
     it('logs once per function call', () => {
-      const debug = jest.spyOn(logger, 'debug');
+      const debug = jest.spyOn((logger as RenovateLogger).logger, 'debug');
```
viceice, line 59: why do we need this cast?

lib/logger/index.ts
```diff
@@ -17,75 +17,89 @@ import {
...
 const problems = new ProblemStream();
-
 let stdoutLevel = validateLogLevel(getEnv('LOG_LEVEL'), 'info');
-const stdout: bunyan.Stream = {
-  name: 'stdout',
-  level: stdoutLevel,
-  stream: process.stdout,
-};
+
+export function getProblems(): BunyanRecord[] {
```
viceice, line 23: can you do these code moves in a seperate pr? so we've less changes here?

lib/logger/index.ts
```diff
@@ -94,11 +108,21 @@ const logFactory = (
...
+function getMessage(
+  p1: string | Record<string, any>,
+  p2?: string,
+): string | undefined {
+  return is.string(p1) ? p1 : p2;
+}
+
+function toMeta(p1: string | Record<string, any>): Record<string, unknown> {
+  return is.object(p1) ? p1 : {};
+}
```
viceice, line 125: can you move these util functions to a separate file?

lib/logger/index.ts
```diff
@@ -109,69 +133,145 @@ const loggerLevels: bunyan.LogLevelString[] = [
...
+    config: configSerializer,
+    migratedConfig: configSerializer,
+    originalConfig: configSerializer,
+    presetConfig: configSerializer,
+    oldConfig: configSerializer,
+    newConfig: configSerializer,
+    err: errSerializer,
+  },
+  streams: defaultStreams.map(withSanitizer),
+});
```
viceice, line 241: can't these be also moved to the logger class? and the constructor more parameterized? that way the logger would be much easier testable (when the class is in a seperate file too.

lib/logger/index.ts
```diff
@@ -17,75 +17,89 @@ import {
...
+  const stdout: bunyan.Stream = {
+    name: 'stdout',
+    level: stdoutLevel,
+    stream: process.stdout,
+  };
+
+  // istanbul ignore if: not testable
+  if (getEnv('LOG_FORMAT') !== 'json') {
+    // TODO: typings (#9615)
+    const prettyStdOut = new RenovateStream() as any;
```
secustor, line 49:
```suggestion
    const prettyStdOut = new RenovateStream();
```
If we are refactoring we should get rid of the uses of `any`.

Gabriel-Ladzaretti: this was just moved as is and is not part of this refactor.
i suppose it can be addressed in another pr, as it is out of scope for this one :smiley: 

its part of `pretty-stdout.ts`, and this class extends `node:stream`, but it is later expected to be `NodeJS.WritableStream | bunyan.WriteFn | undefined`, hence the `any`

lib/logger/index.ts
Gabriel-Ladzaretti: we currently inject the `bunyan` logger instance via the constructor. This approach maximizes flexibility for the user, allowing them to configure the logger as needed. It also keeps the logger class minimal, as it doesnt need to manage details like the `problem` stream or other specifics of the injected logger.


for example, if we move this to the logger class, it would need to expose additional methods, such as `getProblems`, etc.

i prefer to keep this approach, as it is the most generic and flexible for the caller, if thats ok with you :smile:

Gabriel-Ladzaretti: whats more, for the `problem` stream as an example, we could either pass it in the constructor or create it ourselves inside the class.
but what if the caller doesnt need a `problem` stream at all?

i truly believe that leaving all these details outside the logger class is much cleaner while still providing all the flexibility to the user.

lib/logger/renovate-logger.ts
```diff
@@ -0,0 +1,137 @@
...
+      logFn(p1, p2);
+    }
+  }
+}
+
+function logFactory(
+  bunyanLogger: bunyan,
+  _level: bunyan.LogLevelString,
+  curMeta: Record<string, unknown>,
+  logContext: string,
```
Gabriel-Ladzaretti, line 114: this is passed as a value. therefore it wont be possible to change the log context via the setter. either this function should be moved inside the logger class, or the context should be passed using the `String` wrapper type.

Gabriel-Ladzaretti:
there was a bug with the `logFactory` and the way the logging attributes were handled. 
it has been fixed, and I also improved some of the unit tests that didnt catch it initially.

can we please review it again? thanks a lot.

cc @viceice @secustor

viceice:
please split the fix from this refactoring

Gabriel-Ladzaretti:
> please split the fix from this refactoring

but thats impossible, its a bug i introduced in this pr :sweat_smile: . 
maybe i should have improved my phrasing in the commend i made.

lib/logger/index.spec.ts
```diff
@@ -17,16 +19,40 @@ import {
...
 jest.unmock('.');
+jest.mock('nanoid', () => ({
+  nanoid: () => 'initial_context',
+}));
+
+const bunyanDebugSpy = jest.spyOn(bunyan.prototype, 'debug');
 
 describe('logger/index', () => {
+  beforeEach(() => {});
+
```
viceice, line 33:
```suggestion
```

😏

lib/logger/renovate-logger.ts
```diff
@@ -0,0 +1,128 @@
...
+  'info',
+  'warn',
+  'error',
+  'fatal',
+];
+
+type loggerFunction = (p1: string | Record<string, any>, p2?: string) => void;
+
+export class RenovateLogger implements Logger {
+  logger: Logger = { once: { reset: onceReset } } as any;
```
viceice, line 20:
```suggestion
  readonly logger: Logger = { once: { reset: onceReset } } as any;
```

why this needs to be public?

lib/logger/renovate-logger.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  trace = this.log.bind(this, 'trace');
+  debug = this.log.bind(this, 'debug');
+  info = this.log.bind(this, 'info');
+  warn = this.log.bind(this, 'warn');
+  error = this.log.bind(this, 'error');
+  fatal = this.log.bind(this, 'fatal');
+
+  once = this.logger.once;
```
viceice, line 52:
```suggestion
  readonly trace = this.log.bind(this, 'trace');
  readonly debug = this.log.bind(this, 'debug');
  readonly info = this.log.bind(this, 'info');
  readonly warn = this.log.bind(this, 'warn');
  readonly error = this.log.bind(this, 'error');
  readonly fatal = this.log.bind(this, 'fatal');

  readonly once = this.logger.once;
```

I don't like those, i would prefer to use wrapper functions.

lib/logger/renovate-logger.ts
```diff
@@ -0,0 +1,128 @@
...
+        logContext: this.context,
+        ...this.meta,
+        ...toMeta(p1),
+      };
+      const msg = getMessage(p1, p2);
+      let level = _level;
+
+      if (is.string(msg)) {
+        const remappedLevel = getRemappedLevel(msg);
+        // istanbul ignore if: not testable
```
viceice, line 90:
```suggestion
        // istanbul ignore if: not easily testable
```

lib/logger/renovate-logger.ts
```diff
@@ -0,0 +1,128 @@
...
+const loggerLevels: bunyan.LogLevelString[] = [
+  'trace',
+  'debug',
+  'info',
+  'warn',
+  'error',
+  'fatal',
+];
+
+type loggerFunction = (p1: string | Record<string, any>, p2?: string) => void;
```
viceice, line 17:
```suggestion
type LoggerFunction = (p1: string | Record<string, unknown>, p2?: string) => void;
```

Types should start with UpperCase

lib/logger/index.ts
```diff
@@ -17,161 +16,120 @@ import {
...
 
-  addStream({
+  return {
     name: 'logfile',
     path: logFile,
     level: validateLogLevel(getEnv('LOG_FILE_LEVEL'), 'debug'),
+  };
+}
+
+export function serializedSanitizedLogger(streams: bunyan.Stream[]): bunyan {
```
viceice, line 76: can we move these to a different file too? or not make them public?

lib/logger/renovate-logger.ts
Gabriel-Ladzaretti: i love golang too much i guess :)

lib/logger/index.ts
```diff
@@ -17,161 +16,120 @@ import {
...
 }
 
+const defaultStreams = createDefaultStreams(
+  stdoutLevel,
+  problems,
+  getEnv('LOG_FILE'),
+);
+
+const bunyanLogger = serializedSanitizedLogger(defaultStreams);
+const logContext: string = getEnv('LOG_CONTEXT') ?? nanoid();
```
viceice, line 101:
```suggestion
const logContext = getEnv('LOG_CONTEXT') ?? nanoid();
```
looks like we don't need this type annotation

lib/logger/renovate-logger.ts
```diff
@@ -0,0 +1,157 @@
...
+    this.log('error', p1, p2);
+  }
+
+  fatal(p1: string): void;
+  fatal(p1: Record<string, any>, p2?: string): void;
+  fatal(p1: string | Record<string, any>, p2?: string): void {
+    this.log('fatal', p1, p2);
+  }
+
+  once = this.logger.once;
```
viceice, line 69:
```suggestion
  readonly once = this.logger.once;
```

viceice (commented):
let's improve further in next pr

viceice (approved):

PR #33066, opened 2024-12-12
feat(manager/gomod): option to always run go mod vendor

omercnet:
## Changes

Adding an option to always run `go mod vendor` regardless of vendor dir

## Context

- https://github.com/renovatebot/renovate/discussions/33063


lib/modules/manager/gomod/artifacts.ts
```diff
@@ -134,9 +134,9 @@ export async function updateArtifacts({
   const vendorDir = upath.join(goModDir, 'vendor/');
   const vendorModulesFileName = upath.join(vendorDir, 'modules.txt');
```
viceice, line 135: so in your case this file doesn't exist?

omercnet: that's right

rarkins: Is there an alternative file which exists which renovate should have been able to detect?

omercnet: Nope
This use case is for when vendored files aren't committed

rarkins: So your use case is to _add_ vendored files, using Renovate? And then presumably once each repo has added them, this feature isn't really needed?

omercnet: No
I don't want to add them to the commit, I wanna use the vendored files in a post update task

Running go mod vendor in post update is missing the host rules so I need it in the context of the update where those vars are available 

See the related discussion for full details of the use case

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -134,9 +134,9 @@ export async function updateArtifacts({
...
   const useVendor =
-    !config.postUpdateOptions?.includes('gomodSkipVendor') &&
-    (await readLocalFile(vendorModulesFileName)) !== null;
-
+    (!config.postUpdateOptions?.includes('gomodSkipVendor') &&
+      (await readLocalFile(vendorModulesFileName)) !== null) ||
+    config.postUpdateOptions?.includes('gomodVendor');
```
rarkins, line 139:
```suggestion
    config.postUpdateOptions?.includes('gomodVendor') ||
    ((await readLocalFile(vendorModulesFileName)) !== null &&
      !config.postUpdateOptions?.includes('gomodSkipVendor'));
```

I think this ordering is easier to understand/maintain. I'm not sure if prettier will want the outer () there or not.

docs/usage/configuration-options.md
```diff
@@ -3348,6 +3348,7 @@ Table with options:
 | `gomodTidyE`                 | Run `go mod tidy -e` after Go module updates.                                                                                                              |
 | `gomodUpdateImportPaths`     | Update source import paths on major module updates, using [mod](https://github.com/marwan-at-work/mod).                                                    |
 | `gomodSkipVendor`            | Never run `go mod vendor` after Go module updates.                                                                                                         |
+| `gomodVendor`                | Always run `go mod vendor` after Go module updates.                                                                                                        |
```
rarkins, line 3351:
```suggestion
| `gomodVendor`                | Always run `go mod vendor` after Go module updates even if vendor files aren't detected.                                                                                                        |
```

rarkins (changes requested):

lib/modules/manager/gomod/artifacts.ts
omercnet: That was my initial approach but the linter shut me down for using OR on a nullish value

omercnet:
I wanna add another test to ensure go work vendor also applies if needed so don't merge yet

lib/modules/manager/gomod/artifacts.ts
rarkins: How about now? Changed || to ??

omercnet: For some reason the ?? operator failed tests but will try to figure it out later today

omercnet: ok figured the issue, now that the condition is before just needed to remove the mock that reads the vendor file

omercnet: will add ignoreConditionalTests to allow `??` in conditions

.eslintrc.js
```diff
@@ -114,7 +114,10 @@ module.exports = {
       },
     ],
     '@typescript-eslint/prefer-optional-chain': 2,
-    '@typescript-eslint/prefer-nullish-coalescing': 2,
+    '@typescript-eslint/prefer-nullish-coalescing': [
+      2,
+      { ignoreMixedLogicalExpressions: true },
```
rarkins, line 119: I prefer not to add this generally

rarkins (changes requested):

.eslintrc.js
omercnet: Fair enough
But somethings gotta give

Add a nolint comment in the code then?
It won't pass eslint otherwise

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -134,7 +134,7 @@ export async function updateArtifacts({
   const vendorDir = upath.join(goModDir, 'vendor/');
   const vendorModulesFileName = upath.join(vendorDir, 'modules.txt');
   const useVendor =
-    config.postUpdateOptions?.includes('gomodVendor') ||
+    !!config.postUpdateOptions?.includes('gomodVendor') ||
```
omercnet, line 137: ah, nice :)

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -134,9 +134,9 @@ export async function updateArtifacts({
   const vendorDir = upath.join(goModDir, 'vendor/');
   const vendorModulesFileName = upath.join(vendorDir, 'modules.txt');
   const useVendor =
-    !config.postUpdateOptions?.includes('gomodSkipVendor') &&
-    (await readLocalFile(vendorModulesFileName)) !== null;
-
+    !!config.postUpdateOptions?.includes('gomodVendor') ||
+    ((await readLocalFile(vendorModulesFileName)) !== null &&
+      !config.postUpdateOptions?.includes('gomodSkipVendor'));
```
rarkins, line 139:
```suggestion
    !!config.postUpdateOptions?.includes('gomodVendor') ||
    (!config.postUpdateOptions?.includes('gomodSkipVendor') && 
      (await readLocalFile(vendorModulesFileName)) !== null);
```

Restore original ordering

omercnet: sure thing, will fix the tests too

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -134,9 +134,9 @@ export async function updateArtifacts({
   const vendorDir = upath.join(goModDir, 'vendor/');
   const vendorModulesFileName = upath.join(vendorDir, 'modules.txt');
   const useVendor =
-    !config.postUpdateOptions?.includes('gomodSkipVendor') &&
-    (await readLocalFile(vendorModulesFileName)) !== null;
-
+    !!config.postUpdateOptions?.includes('gomodVendor') ||
+    (!config.postUpdateOptions?.includes('gomodSkipVendor') &&
```
rarkins, line 138:
```suggestion
    (!config.postUpdateOptions?.includes('gomodSkipVendor') &&
```

rarkins (approved):

omercnet:
@rarkins @viceice thanks for the review, is anyone able to merge this?

omercnet:
@rarkins ?

PR #33099, opened 2024-12-13
fix(datasource/gitlab-packages): prefer checking for conan_package_name if it exists

FFace32:
## Changes

Checks if `conan_package_name` exists and matches the filtered package name for the gitlab-packages datasource.

## Context

When querying for uploaded Conan packages on GitLab, the `name` value contains the version and the channel as well, and this doesn't pass the filter in `lib/modules/datasource/gitlab-packages/index.ts`.

`conan_package_name`, if it exists, contains only the package name.

This can also be seen in the GitLab documentation as an example here: https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/api/packages.md?plain=1#L69-70


rarkins (approved):

rarkins (changes requested):
Please add a test

lib/modules/datasource/gitlab-packages/index.spec.ts
```diff
@@ -8,7 +8,7 @@ describe('modules/datasource/gitlab-packages/index', () => {
     it('returns package from custom registry', async () => {
       const body = [
         {
-          version: '1.0.0',
+          version: 'v1.0.0',
```
FFace32, line 11: Added this `v` to match the others below.

rarkins: Are you sure it wasn't present to test absence of v prefix?

FFace32: Fair point, reverting this.

rarkins (dismissed):

FFace32:
Should I update the `index.spec.ts.snap` snapshot?

It seems like that's the reason the pipeline fails.

FFace32:
> Should I update the `index.spec.ts.snap` snapshot?
> 
> It seems like that's the reason the pipeline fails.

Sorry, didn't see the thumbs up until now.

FFace32:
Is there something I can do about the `Snapshots have been added in this PR. Use toMatch instead.` error I can do?

lib/modules/datasource/gitlab-packages/index.spec.ts
```diff
@@ -45,6 +45,50 @@ describe('modules/datasource/gitlab-packages/index', () => {
...
+          per_page: '100',
+        })
+        .reply(200, body);
+      const res = await getPkgReleases({
+        datasource,
+        registryUrls: ['https://gitlab.com'],
+        packageName: 'user/project1:myconanpkg',
+      });
+      expect(res).toMatchSnapshot();
+      expect(res?.releases).toHaveLength(3);
```
viceice, line 89: use `toEqual` here

rarkins (approved):

RahulGautamSingh:
Thanks for the PR! 

It has been inactive for some months, and the work is mostly completed.Therefore, I have created a [PR ](https://github.com/renovatebot/renovate/pull/36954 )replacing this.

PR #33108, opened 2024-12-14
feat: extend skipStage to extract, lookup

rarkins:
## Changes

Extends existing `skipStage` to include `extract` and `lookup` stages.
Adds extra logic to remove `skipReason`/`skipStage` from extract cache if skipReason != extract

## Context

Discovered as part of #32612


lib/workers/repository/process/extract-update.ts
```diff
@@ -128,6 +128,10 @@ export async function extract(
         for (const file of files) {
           for (const dep of file.deps) {
             delete dep.updates;
+            if (dep.skipStage && dep.skipStage !== 'extract') {
```
viceice, line 131: so currently it can only be lookup?

rarkins: Or any time package rules apply (was existing)

viceice (approved):

PR #33115, opened 2024-12-15
feat(git): cloneSubmodulesFilter

rarkins:
## Changes

Adds new option `cloneSubmodulesFilter` to allow selective cloning of submodules.

## Context

#33027

### DIFF
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -2857,6 +2857,14 @@ const options: RenovateOptions[] = [
     default: false,
   },
   {
+    name: 'cloneSubmodulesFilter',
+    description:
+      'List of submodules names or patterns to clone when cloneSubmodules=true.',
+    type: 'array',
+    subType: 'string',
+    default: ['*'],
+  },
+  {
     name: 'ignorePrAuthor',
     description:
       'Set to `true` to fetch the entire list of PRs instead of only those authored by the Renovate user.',
--- a/lib/util/git/index.ts
+++ b/lib/util/git/index.ts
@@ -347,7 +347,7 @@ export async function getSubmodules(): Promise<string[]> {
 
 export async function cloneSubmodules(
   shouldClone: boolean,
-  cloneSubmodulesFilter: string[] = ['*'],
+  cloneSubmodulesFilter: string[] | undefined,
 ): Promise<void> {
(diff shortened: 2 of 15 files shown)


rarkins:
Needs testing on at least one real repo

rarkins:
I thought about converting `cloneSubmodules` from boolean to string list but decided it was too complex/risky. e.g. also need to change CLI, env, etc.

lib/config/options/index.ts
```diff
@@ -2856,6 +2856,14 @@ const options: RenovateOptions[] = [
     type: 'boolean',
     default: false,
   },
+  {
+    name: 'cloneSubmodulesFilter',
+    description:
+      'List of submodules names or patterns to clone when cloneSubmodules=true.',
+    type: 'array',
+    subType: 'string',
+    default: ['*'],
```
viceice, line 2865: `*`  only matches `something` but not `some/sub/module`
```suggestion
    default: ['**'],
```

lib/util/git/index.ts
```diff
@@ -344,14 +345,24 @@ export async function getSubmodules(): Promise<string[]> {
   }
 }
 
-export async function cloneSubmodules(shouldClone: boolean): Promise<void> {
+export async function cloneSubmodules(
+  shouldClone: boolean,
+  cloneSubmodulesFilter: string[] = ['*'],
```
viceice, line 350:
```suggestion
  cloneSubmodulesFilter: string[] | undefined,
```

we shouldn't set a default value here, but require this arg to be passed, even when `undefined`

lib/util/git/index.ts
```diff
@@ -344,14 +345,24 @@ export async function getSubmodules(): Promise<string[]> {
...
+): Promise<void> {
   if (!shouldClone || submodulesInitizialized) {
     return;
   }
   submodulesInitizialized = true;
   await syncGit();
   const submodules = await getSubmodules();
   for (const submodule of submodules) {
+    if (!matchRegexOrGlobList(submodule, cloneSubmodulesFilter)) {
```
viceice, line 359:
```suggestion
    if (cloneSubmodulesFilter && !matchRegexOrGlobList(submodule, cloneSubmodulesFilter)) {
```

viceice (approved):

PR #26876, opened 2024-01-26
fix(gitea): Use correct URL in PR Pagination

ASDFGamer:
## Changes

This PR fixes a Bug that occures on the gitea plattform.  

Previously it would use the url that was returned by gitea for the second page without checking if the url should be changed according to the gitUrl option.

This change removes the host part of the Url, like it is done for the first page of PRs and lets renovate choose the appropiate host and scheme.

## Context

The bug occures only if the gitea api returns more than one page of PRs and the gitUrl option is set to endpoint or ssh.
I have no minimal reproduction of this bug becuse it needs a gitea instance and i don't have a public instance. 

## Documentation (please check one with an [x])

- [X] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [X] Both unit tests + ran on a real repository

PR #33121, opened 2024-12-15
fix(gitea): use configured endpoint in PR cache

javiertury:
## Changes

Ensure that gitea PR cache uses the endpoint configured by the user

## Context

As detailed in issue https://github.com/renovatebot/renovate/discussions/32825, renovate paginates using the url returned by gitea, which may change the endpoint. This causes authentication failures since tokens are not attached to the request or a network error if the returned url is not accesible from within the network. Also a gitea instance configured with a `http` public url would remove transport encryption even if renovate was configured with a `https` endpoint.

### DIFF
--- a/lib/modules/platform/gitea/pr-cache.ts
+++ b/lib/modules/platform/gitea/pr-cache.ts
@@ -115,13 +115,16 @@ export class GiteaPrCache {
   }
 
   private async sync(http: GiteaHttp): Promise<GiteaPrCache> {
-    const query = getQueryString({
+    const urlPath = `${API_PATH}/repos/${this.repo}/pulls`;
+    const queryParams = {
       state: 'all',
       sort: 'recentupdate',
-    });
+    };
 
+    let page: number = 1;
+    const query = getQueryString(queryParams);
     let url: string | undefined =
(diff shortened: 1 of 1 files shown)


viceice (changes requested):
implement `RENOVATE_X_REBASE_PAGINATION_LINKS` like GitHub 

https://github.com/renovatebot/renovate/blob/d0c0aa7333b0c00582dfc8cc2dbebe7e4f80d0e7/lib/util/http/github.ts#L338

javiertury:
What is the advantage of `RENOVATE_X_REBASE_PAGINATION_LINKS` over the proposed solution? We can address the root of the problem if we eliminate the need to rebase in the first place.

According to gitea docs, the [pulls endpoint](https://demo.gitea.com/api/swagger#/repository/repoListPullRequests) works with limit-offset pagination and `next.url` just increases the page counter so both methods are equivalent. It is not the case that `next.url` uses stateless cursor pagination or something more performant, and even if it was the case my guess is that gitea would provide a well-designed API such that we can pass the cursor through query parameters.

viceice:
we did the decision some time ago, there where more discussions about it. so please do it like we suggest.

viceice:
- #26876

PR #33169, opened 2024-12-17
fix(go): Request only info for versions with the same major

zharinov:


zharinov:
image

Seems like some registries/proxies could return all versions no matter which major prefix we're querying. It leads us to increment major and fetch them over and over.

zharinov:
It's confirmed that artifactory returns all versions in any major, so this is the fix

rarkins (approved):

rarkins:
@zharinov please test on public goproxy too, to reduce chances of regression

PR #33190, opened 2024-12-18
fix(yaml): less aggressive jinja2 stripping

rarkins:
## Context

https://github.com/renovatebot/renovate/discussions/32967
https://regexr.com/89otc

### DIFF
--- a/lib/util/yaml.ts
+++ b/lib/util/yaml.ts
@@ -143,7 +143,7 @@ export function dump(obj: any, opts?: DumpOptions): string {
 function massageContent(content: string, options?: YamlOptions): string {
   if (options?.removeTemplates) {
     return content
-      .replace(regEx(/\s*{{.+?}}(:.*)?/gs), '')
+      .replace(regEx(/\n\s*{{.+?}}:\s*\n/gs), '')
       .replace(regEx(/{{`.+?`}}/gs), '')
       .replace(regEx(/{{.+?}}/gs), '')
       .replace(regEx(/{%`.+?`%}/gs), '')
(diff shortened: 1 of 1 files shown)


lib/util/yaml.ts
```diff
@@ -143,7 +143,7 @@ export function dump(obj: any, opts?: DumpOptions): string {
 function massageContent(content: string, options?: YamlOptions): string {
   if (options?.removeTemplates) {
     return content
-      .replace(regEx(/\s+{{.+?}}:.+/gs), '')
+      .replace(regEx(/\n\s*{{.+?}}:\s*\n/gs), '')
```
viceice, line 146:
```suggestion
      .replace(regEx(/^\s*{{.+?}}:\s*\n/gs), '')
```
otherwise you'll remove one new line to much

rarkins: Will that ^ match "start of line" or "start of file"?

Sculas: If I fork the regexr reproducer and use this ^ regex, it no longer matches the template. The documentation mentions:

> Matches the beginning of the string, or the beginning of a line if the multiline flag (m) is enabled.

So, to answer the question above, it matches the "start of file" when the multiline flag is disabled (by default).
Enabling the multiline flag makes the regex work as expected ([regexr link](https://regexr.com/8a79p)):
```suggestion
      .replace(regEx(/^\s*{{.+?}}:\s*\n/gms), '')
```

viceice: single line flag is also required, so `.` doesn't match newline

Sculas: If you don't want `.` to match newlines, you should remove the dotall flag:
```suggestion
      .replace(regEx(/^\s*{{.+?}}:\s*\n/gm), '')
```

rarkins: I think we're stuck again? @viceice please propose a fully working change

PR #33310, opened 2024-12-27
feat(mix): add depType support

sheerlox:
## Changes

Adds support for `depType` based on the `only` Mix dependency option.

## Context

I'm using Conventional Commits & Semantic Release for my Elixir projects. Not having `depType` available means I have to choose between always releasing (`build(deps):`) or never releasing (`chore(deps):`) dependency updates.


rarkins (commented):
It seems you're reusing the npm terms "dependencies" and "devDependencies" but that's not necessary. Probably more intuitive for mix users if they are like "prod" and "dev"?

sheerlox:
This PR revealed a bug in the `hex` versioning, changing to draft while I push the fix PR to avoid unwanted behavior on user's end.

rarkins (dismissed):

lib/modules/manager/mix/extract.ts
```diff
@@ -20,6 +20,8 @@ const lockedVersionRegExp = regEx(
   /^\s+"(?<app>\w+)".*?"(?<lockedVersion>\d+\.\d+\.\d+)"/,
 );
 const hexRegexp = regEx(/hex:\s*(?:"(?<strValue>[^"]+)"|:(?<atomValue>\w+))/);
+const onlyValueRegexp = /only:\s*(?<only>\[[^\]]*\]|:\w+)/;
+const onlyEnvironmentsRegexp = /:(\w+)/gm;
```
viceice, line 24:
```suggestion
const onlyValueRegexp = regEx(/only:\s*(?<only>\[[^\]]*\]|:\w+)/);
const onlyEnvironmentsRegexp = regEx(/:(\w+)/gm);
```

rarkins (approved):

PR #33325, opened 2024-12-29
fix(hex): fix Elixir's special operator in list

sheerlox:
## Changes

Changes the `~` to the correct value `~>` in `hex` versioning's operators list.

## Context

https://hexdocs.pm/elixir/Version.html#module-requirements

Detected while making tests on a branch which contained the changes from #33322:

- https://github.com/sheerlox-repros/renovate-pr-33326/commit/61d99928ce17e7e589b7358cdb2e494183efc071

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [X] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [X] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

PR #33326, opened 2024-12-29
feat(mix): implement lockFileMaintenance

sheerlox:
## Changes

Adds support for the `lockFileMaintenance` option.

Refactors the lock file detection to differentiate umbrella projects from classic ones, as this implementation won't work on the former because they share a single `mix.lock` file for multiple `mix.exs` files.

Cleans up the `getPkgReleases` mocks by only returning the latest version, since these tests aren't about the tool version resolution.

## Context

Resulting commit: https://github.com/sheerlox-repros/renovate-pr-33326/commit/7ec9c870e7022e2d9936ec6322dd310530360ce4


sheerlox:
Hello :wave: Is there anything I can do to help move this PR forward? No worries if it's a time issue, I totally get it. All the best!

rarkins (dismissed):

sheerlox:
Could one of you please explain why we should choose of of the following over the others, from a testing point of view?

I searched for a while when I did the PR and couldn't find an answer (this is why I initially left the `removeDanglingContainers` mocks as I didn't know what I was working with).

```js
GlobalConfig.set({ ...adminConfig, binarySource: 'install' });
```

```js
GlobalConfig.set({
  ...adminConfig,
  binarySource: 'docker',
  dockerSidecarImage: 'ghcr.io/containerbase/sidecar',
});
```

sheerlox:
https://github.com/sheerlox/nodelix/pull/28

Well, that's unexpected! The `update-lockfile` strategy wasn't working until right now, so I'm not 100% sure what caused this PR to be raised, but I'd say it's a (positive) side effect of https://github.com/renovatebot/renovate/pull/33325.

lib/modules/manager/mix/artifacts.spec.ts
```diff
@@ -260,9 +285,9 @@ describe('modules/manager/mix/artifacts', () => {
     // elixir
     getPkgReleases.mockResolvedValueOnce({
       releases: [
-        { version: 'v1.8.2' },
-        { version: 'v1.13.3' },
-        { version: 'v1.13.4' },
+        { version: '1.8.2' },
+        { version: '1.13.3' },
+        { version: '1.13.4' },
```
viceice, line 290: please revert, those versions will be `v` prefixed

https://github.com/renovatebot/renovate/blob/9b4e5150248a6b0026acf9c73bb00a295cba4c48/lib/util/exec/containerbase.ts#L61-L65

https://github.com/elixir-lang/elixir/releases/tag/v1.18.1

rarkins (dismissed):

sheerlox:
Hey there :wave: is this ready for merging or can I do anything else?

lib/modules/manager/mix/artifacts.ts
```diff
@@ -26,15 +28,35 @@ export async function updateArtifacts({
...
+  if (!existingLockFileContent) {
+    lockFileName =
+      (await findLocalSiblingOrParent(packageFileName, 'mix.lock')) ??
+      'mix.lock';
+    existingLockFileContent = await readLocalFile(lockFileName, 'utf8');
+    isUmbrella = !!existingLockFileContent;
+  }
+
+  if (isLockFileMaintenance && isUmbrella) {
+    logger.debug('Cannot use lockFileMaintenance in an umbrella project');
```
viceice, line 51: we should log a url to a description why it's not possible.
```suggestion
    logger.debug('Cannot use lockFileMaintenance in an umbrella project, see https://docs.renovatebot.com/modules/manager/mix/#lockFileMaintenance');
```

lib/modules/manager/mix/artifacts.ts
```diff
@@ -26,15 +28,35 @@ export async function updateArtifacts({
...
+  if (is.emptyArray(updatedDeps) && !isLockFileMaintenance) {
+    logger.debug('No updated mix deps');
+    return null;
+  }
+
+  let lockFileName = getSiblingFileName(packageFileName, 'mix.lock');
+  let isUmbrella = false;
+
+  let existingLockFileContent = await readLocalFile(lockFileName, 'utf8');
+  if (!existingLockFileContent) {
```
viceice, line 42: test `if (await localPathExists(lockfileName))` then return error while reading lockfile.

lib/modules/manager/mix/artifacts.ts
```diff
@@ -113,14 +134,20 @@ export async function updateArtifacts({
...
-  const command = [
-    'mix',
-    'deps.update',
-    ...updatedDeps
-      .map((dep) => dep.depName)
-      .filter(is.string)
-      .map((dep) => quote(dep)),
-  ].join(' ');
+
+  let command;
```
viceice, line 138:
```suggestion
  let command: string;
```

lib/modules/manager/mix/artifacts.ts
sheerlox: Can you be more specific here? I'm having trouble to understand.

viceice: you currently simply fallback to another umbrella check. the read local file function eats any errors and returns null. so you need to check at least if the file exists.

if it exists and content is null, the error out instead of silently doing the umbrella fallback inside the condition

viceice (commented):
still open conversation

lib/modules/manager/mix/artifacts.spec.ts
```diff
@@ -292,27 +317,155 @@ describe('modules/manager/mix/artifacts', () => {
...
+    fs.readLocalFile.mockResolvedValueOnce('New mix.lock');
+
+    // erlang
+    getPkgReleases.mockResolvedValueOnce({
+      releases: [{ version: '25.0.0.0' }],
+    });
+    // elixir
+    getPkgReleases.mockResolvedValueOnce({
+      releases: [{ version: 'v1.13.4' }],
+    });
```
viceice, line 420: if you like to simplify new tests, then you can simply pass the versions via `contraints` config object

lib/modules/manager/mix/artifacts.ts
```diff
@@ -161,3 +205,20 @@ export async function updateArtifacts({
     },
   ];
 }
+
+async function lockFileReadError(
+  lockFileName: string,
+): Promise<UpdateArtifactsResult[] | null> {
+  if (await localPathExists(lockFileName)) {
+    logger.warn(`Could not read ${lockFileName}`);
```
viceice, line 213:
```suggestion
```

artifact error will be logged from caller

lib/modules/manager/mix/artifacts.ts
```diff
@@ -26,15 +29,51 @@ export async function updateArtifacts({
...
+      if (lockFileError) {
+        return lockFileError;
+      }
+    }
+  }
+
+  if (isLockFileMaintenance && isUmbrella) {
+    logger.debug(
+      'Cannot use lockFileMaintenance in an umbrella project, see https://docs.renovatebot.com/modules/manager/mix/#lockFileMaintenance',
+    );
```
viceice, line 68: @rarkins should we use a different log level here? i'm unsure.

rarkins: I don't mind too much about debug logs

sheerlox:
Hello, can I do anything more to get this through?

P.S. crazy how I seem to have a 2-week stale PR reminder hardwired in my brain :sweat_smile:

lib/modules/manager/mix/artifacts.ts
```diff
@@ -26,15 +29,51 @@ export async function updateArtifacts({
   config,
 }: UpdateArtifact): Promise<UpdateArtifactsResult[] | null> {
   logger.debug(`mix.getArtifacts(${packageFileName})`);
-  if (updatedDeps.length < 1) {
-    logger.debug('No updated mix deps - returning null');
+  const isLockFileMaintenance = config.updateType === 'lockFileMaintenance';
```
rarkins, line 32: @viceice is this your recommended way?

viceice: - https://github.com/renovatebot/renovate/pull/33971

lib/modules/manager/mix/artifacts.ts
```diff
@@ -26,15 +29,51 @@ export async function updateArtifacts({
...
+  let existingLockFileContent = await readLocalFile(lockFileName, 'utf8');
+  if (!existingLockFileContent) {
+    const lockFileError = await lockFileReadError(lockFileName);
+    if (lockFileError) {
+      return lockFileError;
+    }
+
+    const parentLockFileName =
+      (await findLocalSiblingOrParent(packageFileName, 'mix.lock')) ??
+      'mix.lock';
```
viceice, line 51:
```suggestion
    const parentLockFileName = await findLocalSiblingOrParent(packageFileName, 'mix.lock');
    if (!parentLockFileName) {
      return null;
    }
```

return artifact error if lockfile is mandatory for mix

viceice (commented):
needs deconflicting

lib/modules/manager/mix/artifacts.ts
sheerlox: `mix` can work without committing the lockfile to version control, although it is recommended to use it. Some users might want to use Renovate in a repository without `mix.lock`.

rarkins (approved):

sheerlox:
:eyes:

viceice (approved):

PR #33339, opened 2024-12-30
fix(bundler): authentication using only env vars

lbarros-asml:
## Changes

Change the bundle authentication to use ENV vars for hostnames with hyphen. Discussion #33205

I might need help with the unit tests because I'm not a js/typescript developer.

## Context

#33205


lib/modules/manager/bundler/artifacts.ts
```diff
@@ -149,47 +150,12 @@ export async function updateArtifacts(
...
-      [],
-    );
-
     const bundler = getBundlerConstraint(
       updateArtifact,
       existingLockFileContent,
     );
     const preCommands = ['ruby --version'];
 
-    // Bundler < 2 has a different config option syntax than >= 2
```
rarkins, line 173: Is this PR dropping support for Bundler v1 too? I think it should be a separate PR if so.

rarkins (changes requested):
Needs tests/test updating too

lib/modules/manager/bundler/artifacts.ts
lbarros-asml: As far as I can tell, the bundler v1 distinction is only needed here because the bundler authentication command in v1 differs from v2, but when using env vars the command is not needed anymore and works the same for both.

lbarros-asml:
> Needs tests/test updating too

I'll try to update the tests, but a help would be welcome. As I  mentioned, typescript is not my thing.

lib/modules/manager/bundler/artifacts.spec.ts
```diff
@@ -681,19 +470,18 @@ describe('modules/manager/bundler/artifacts', () => {
...
+              '-e BUNDLE_GEMS___PRIVATE__COM ' +
               '-e GEM_HOME ' +
               '-e CONTAINERBASE_CACHE_DIR ' +
               '-w "/tmp/github/some/repo" ' +
               'ghcr.io/containerbase/sidecar' +
               ' bash -l -c "' +
               'install-tool ruby 1.2.0' +
               ' && ' +
-              'install-tool bundler 1.3.0' +
+              'install-tool bundler 2.3.5' +
```
rarkins, line 481: What caused this line to change in the test?

lbarros-asml: This is GitHub diff pranking us. I actually didn't remove these lines:
https://github.com/renovatebot/renovate/pull/33339/files#diff-5421e221c5a44026d537b2f1168b1ce07db041069cba49bf2ba6462848e6c5cfL438-L489

So, because they're too similar to the lines I removed, it is showing a bogus visualization.

Happy new year! :tada:

lib/modules/manager/bundler/artifacts.ts
```diff
@@ -32,14 +30,16 @@ import {
 const hostConfigVariablePrefix = 'BUNDLE_';
 
 function buildBundleHostVariable(hostRule: HostRule): Record<string, string> {
-  if (!hostRule.resolvedHost || hostRule.resolvedHost.includes('-')) {
+  if (!hostRule.resolvedHost) {
     return {};
```
lbarros-asml, line 34: @rarkins , this is the line failing the test coverage. I wonder if checking the  "hostRule.resolvedHost" is even necessary. If not, I can remove the if statement. Otherwise, I'd appreciate any help creating a proper test.

Jenibabi1984 (approved):

lib/modules/manager/bundler/artifacts.ts
```diff
@@ -32,14 +30,16 @@ import {
 const hostConfigVariablePrefix = 'BUNDLE_';
 
 function buildBundleHostVariable(hostRule: HostRule): Record<string, string> {
-  if (!hostRule.resolvedHost || hostRule.resolvedHost.includes('-')) {
+  if (!hostRule.resolvedHost) {
```
rarkins, line 33:
```suggestion
  // istanbul ignore if
  if (!hostRule.resolvedHost) {
```

lib/modules/manager/bundler/artifacts.ts
rarkins: I've added an istanbul ignore

rarkins (approved):

Issue #33302, opened 2024-12-27
feat: New config option staticRepoConfig


nabeelsaabna:
### Discussed in https://github.com/renovatebot/renovate/discussions/33280

<div type='discussions-op-text'>

<sup>Originally posted by **nabeelsaabna** December 25, 2024</sup>
### Tell us more.

Mend needs a new config option called `RENOVATE_STATIC_REPO_CONFIG`
This config option is part of the “self-hoted config” (can’t be used from the repository `renovate.json`)
Renovate will evaluate this new config for each repository immediately before reading/processing the repository config 

Current config order:

1. Default config
2. Global config
3. File config
4. Environment config
5. CLI config
6. Inherited config
7. Resolved presets referenced in config
8. Repository config

The new option should be applied and merged after `#6` and before `#7`

better naming options than `staticRepoConfig` are welcomed, preferably a name that is not in use already to reduce confusion.  </div>

PR #33350, opened 2024-12-31
refactor(config): reusable env getConfig function

Gabriel-Ladzaretti:
## Changes
Makes the env parsing function reusable for future use.
It is intended to use this function for parsing and processing the new config option
specified in:
- https://github.com/renovatebot/renovate/issues/33302

### DIFF
--- a/lib/workers/global/config/parse/env.ts
+++ b/lib/workers/global/config/parse/env.ts
@@ -244,7 +244,10 @@ async function parseAndValidateOrExit(
   }
 
   try {
-    const config = parseJson(env[configEnvKey], `${configEnvKey}.env.json5`) as AllConfig;
+    const config = parseJson(
+      env[configEnvKey],
+      `${configEnvKey}.env.json5`,
+    ) as AllConfig;
     logger.debug({ config }, `Detected config in env ${configEnvKey}`);
 
     return await migrateAndValidateConfig(config, `${configEnvKey}`);
--- a/lib/workers/global/config/parse/env.ts
+++ b/lib/workers/global/config/parse/env.ts
@@ -118,9 +118,9 @@ function massageConvertedExperimentalVars(
 
 export async function getConfig(
   inputEnv: NodeJS.ProcessEnv,
+  configEnvKey = 'RENOVATE_CONFIG',
 ): Promise<AllConfig> {
-  let env = inputEnv;
-  env = normalizePrefixes(inputEnv, inputEnv.ENV_PREFIX);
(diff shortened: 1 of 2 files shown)


lib/workers/global/config/parse/env.ts
```diff
@@ -118,102 +118,92 @@ function massageConvertedExperimentalVars(
...
+    if (option.env === false) {
+      continue;
     }
-  }
 
-  config.hostRules ??= [];
+    const envName = getEnvName(option);
+    const envVal = env[envName];
+    if (!envVal) {
+      continue;
```
Gabriel-Ladzaretti, line 143: the diff is missleading,
the only thing changed apart of the `parseAndValidateOrExit` function is converting foreach to for ... of.
\+ flip if statements so that we can early exit (continue here), so that we can have less indentation.

lib/workers/global/config/parse/env.ts
```diff
@@ -237,7 +227,28 @@ export async function getConfig(
...
+async function parseAndValidateOrExit(
+  env: NodeJS.ProcessEnv,
+  configEnvKey: string,
+): Promise<AllConfig> {
+  if (!env[configEnvKey]) {
+    return {};
+  }
+
+  try {
+    const config = JSON5.parse(env[configEnvKey]);
```
viceice, line 246: use our json utils

Gabriel-Ladzaretti: i dont think we can, because the util func catches errors, and here we rely on it to throw when failing to parse it.

lib/workers/global/config/parse/env.ts
```diff
@@ -237,7 +228,28 @@ export async function getConfig(
...
+async function parseAndValidateOrExit(
+  env: NodeJS.ProcessEnv,
+  configEnvKey: string,
+): Promise<AllConfig> {
+  if (!env[configEnvKey]) {
+    return {};
+  }
+
+  try {
+    const config = parseJson(env[configEnvKey], '.json5') as AllConfig;
```
viceice, line 247:
```suggestion
    const config = parseJson(env[configEnvKey], `${configEnvKey}.env.json5`) as AllConfig;
```
should be more descriptive

Gabriel-Ladzaretti: given that this is will be taken to the json5 trinery branch, the filename will never be printed or used. so actually i think `.json5` is cleaner. but i dont mind either :)

viceice (approved):

PR #33376, opened 2025-01-02
feat(managers/npm): support pnpm catalogs

### LINKED ISSUE #30079: Support pnpm Catalogs

Pnpm v9.5[^1] introduced a new way to share dependencies inside workspaces, they call it catalogs[^2].

This is related to #25335.

[^1]: https://github.com/pnpm/pnpm/releases/tag/v9.5.0
[^2]: https://pnpm.io/catalogs

fpapado:
## Changes

This PR adds support for pnpm catalogs, extracting dependencies from / updating dependencies in `pnpm-workspace.yaml` files.

I took @rarkins' suggestion in https://github.com/renovatebot/renovate/issues/30079#issuecomment-2565733221, and added this as part of the `npm` manager. I tried to keep the functionality contained in `pnpm` files inside the relevant modules, to make it easier to split out in the future, if needed.

I took care of the following things:
- `pnpm-workspace.yaml` files are picked up by default; their dependencies are extracted, and any updates are applied
- Updates should preserve formatting as much as possible (tests back this up, but YAML is complicated)
- `pnpm-workspace.yaml` files can be renamed and specified in `fileMatch`. The `npm` manager tries to parse a workspace YAML file first, before falling back to pakcage.json-like parsing.
- The `pnpmShrinkwrap` file gets picked up, to use for updating. I believe I implemented that correctly, but I might have missed something.
- Locked catalog dependencies are resolved from the lockfile. This should allow range replacements and other more advanced use-cases.

---

_These questions have been resolved; I leave the text here for posterity_

### Open question 1: ways of picking up `pnpm-workspace.yaml` for extraction

I have placed `pnpm-workspace.yaml` in the `fileMatch` defaultConfig of the npm manager, and added an extra branch to `extractAllPackageFiles`. 

An alternative would be to do this work in `postExtract`, by looking up sibling/parent `pnpm-workspace.yaml` file, and doing similar processing. This avoids the need for changing the config, but might have downsides that I cannot think of. What do you think?

### Open question 2: `depType` and allowing granular updates to named catalogs

In the current version, I gave catalog-related upgrades a `depType` of `pnpm.catalog`, just for namespacing. However, I am wondering whether we could go more granular with this, such as `pnpm.catalog.<catalogName>`

[One of the pnpm examples for catalogs, is allowing incremental updates, by defining named catalogs](https://pnpm.io/catalogs#named-catalogs):

```yaml
catalogs:
  # Can be referenced through "catalog:react17"
  react17:
    react: ^17.0.2
    react-dom: ^17.0.2

  # Can be referenced through "catalog:react18"
  react18:
    react: ^18.2.0
    react-dom: ^18.2.0
```

In such a scenario, I imagine users wanting to keep react 17 and 18 each in their range (allowing for bugfixes), without bumping the major. To do that, we would need to expose the catalog name to the user for matching, such as `{matchDepTypes: ['pnpm.catalog.react17']}`.

Have you had similar scenarios in the past? I noticed that `pnpm.overrides` _kind of_ works like this, when using the `pkg-a>pkg-b` syntax ([reference](https://pnpm.io/package_json#pnpmoverrides:~:text=You%20may%20specify%20the%20package%20the%20overridden%20dependency%20belongs%20to%20by%20separating%20the%20package%20selector%20from%20the%20dependency%20selector%20with%20a%20%22%3E%22%2C), [example](https://github.com/fpapado/renovate-testing-catalogs/pull/3)), but based on the depName. This could probably be done in a follow-up PR, but it's good to think about 💭 

## Context

Closes #30079 

[pnpm catalogs are described in the pnpm docs](https://pnpm.io/catalogs)

This is my first Renovate contribution, I'm open to any and all ideas 😊 

- [] No documentation update is required

I made a small documentation addition for the readme.md of the npm manager, in order to mention `pnpm.catalogs.<name>` as a `depType`.

I added unit tests for the extraction and updates, the latter being a similar set to the `package.json` update tests. I will double-check the code coverage on CI, I think I missed something when running coverage locally.

I did a fair bit of repository testing at https://github.com/fpapado/renovate-testing-catalogs/pulls, to understand how the npm extraction/update phases work for `package.json` files.


lib/modules/manager/npm/update/dependency/pnpm.ts
```diff
@@ -0,0 +1,104 @@
...
+  logger.debug(
+    `npm.updatePnpmCatalogDependency(): ${depType}:${managerData?.catalogName}.${depName} = ${newValue}`,
+  );
+
+  let document;
+  let parsedContents;
+
+  try {
+    document = parseSingleYamlDocument(fileContent);
+    parsedContents = pnpmCatalogsSchema.parse(document.toJS());
```
fpapado, line 38: I opted to do the YAML replacements with a document representation, instead of string replacements, because it seemed simpler as a start. I particularly did not want to drill down into the catalogs, accounting for the different YAML Map representations, string quoting etc. 

Now that the tests are relatively set, I could change the implementation to do string replacement, if you think that is best.

lib/util/yaml.ts
```diff
@@ -117,6 +125,29 @@ export function parseSingleYaml<ResT = unknown>(
...
+ *
+ * Only a single document is supported.
+ *
+ * @param content
+ * @param options
+ */
+export function parseSingleYamlDocument(
+  content: string,
+  options?: YamlParseDocumentOptions,
+): Document {
```
fpapado, line 150: I added this just to keep the parsing centralised, even if there are no other use-cases for YAML Document transforms at the moment.

fpapado:
Ah, I noticed now in my test repo that the lockfile does not get updated after a standalone catalog dependency upgrade. I missed that codepath when trying to keep the functionality minimal. I'll add it to my todo 🗒️

fpapado:
@secustor and @viceice, do you have any pointers for the open questions here? If you think the PR is not in a reviewable state, I'm happy to implement with my gut feeling and re-request a review later down the line. I'm asking because I don't want to waste your time if something ends up being unworkable or unidiomatic, but I also don't want this to be stuck in limbo 😌

viceice:
we usually don't review draft PRs, will try to have a look tomorrow

lib/modules/manager/npm/extract/pnpm.ts
```diff
@@ -222,3 +232,108 @@ function getLockedDependencyVersions(
 
   return res;
 }
+
+/**
+ * A pnpm catalog is either the default catalog (catalog:, catlog:default), or a
+ * named one (under catalogs:)
+ */
+type PnpmCatalog = { name: string; dependencies: NpmPackageDependency };
```
secustor, line 240: This should be moved to `./types.ts`

lib/modules/manager/npm/extract/pnpm.ts
```diff
@@ -222,3 +232,108 @@ function getLockedDependencyVersions(
...
+
+export function extractPnpmWorkspaceFile(
+  content: string,
+  packageFile: string,
+): PackageFile | null {
+  logger.trace(`pnpm.extractPnpmWorkspaceFile(${packageFile})`);
+
+  let pnpmCatalogs: Array<PnpmCatalog>;
+  try {
+    pnpmCatalogs = parsePnpmCatalogs(content);
```
secustor, line 250:
```suggestion
  const pnpmCatalogs: PnpmCatalog[] = [];
  try {
    pnpmCatalogs.push(...parsePnpmCatalogs(content));
```
consistency

lib/modules/manager/npm/extract/pnpm.ts
```diff
@@ -222,3 +232,108 @@ function getLockedDependencyVersions(
...
+  } catch {
+    logger.debug({ packageFile }, `Invalid pnpm workspace YAML.`);
+    return null;
+  }
+
+  const extracted = extractPnpmCatalogDeps(pnpmCatalogs);
+
+  if (!extracted) {
+    logger.debug({ packageFile }, 'No dependencies found');
```
secustor, line 259:
```suggestion
```
This information is already logged outside of the manager

lib/modules/manager/npm/extract/pnpm.ts
```diff
@@ -222,3 +232,108 @@ function getLockedDependencyVersions(
...
+    return null;
+  }
+
+  logger.debug(extracted, 'Extracted catalog dependencies.');
```
secustor, line 263:
```suggestion
```
Same as above

lib/modules/manager/npm/extract/pnpm.ts
```diff
@@ -222,3 +232,108 @@ function getLockedDependencyVersions(
...
+
+  return {
+    ...extracted,
+    packageFile,
+  };
+}
+
+function extractPnpmCatalogDeps(
+  catalogs: Array<PnpmCatalog>,
```
secustor, line 272:
```suggestion
  catalogs: PnpmCatalog[],
```

lib/modules/manager/npm/extract/pnpm.ts
```diff
@@ -222,3 +232,108 @@ function getLockedDependencyVersions(
...
+      deps.push(dep);
+    }
+  }
+
+  return {
+    deps,
+  };
+}
+
+export const pnpmCatalogsSchema = z.object({
```
secustor, line 320: Should be moved to the top or `./schema.ts`

lib/modules/manager/npm/extract/pnpm.ts
```diff
@@ -222,3 +232,108 @@ function getLockedDependencyVersions(
...
+  catalog: z.optional(z.record(z.string())),
+  catalogs: z.optional(z.record(z.record(z.string()))),
+});
+
+function parsePnpmCatalogs(content: string): Array<PnpmCatalog> {
```
secustor, line 325:
```suggestion
function parsePnpmCatalogs(content: string): PnpmCatalog[] {
```

lib/modules/manager/npm/extract/pnpm.ts
```diff
@@ -222,3 +232,108 @@ function getLockedDependencyVersions(
...
+  return [
+    {
+      name: 'default',
+      dependencies: defaultCatalogDeps ?? {},
+    },
+    ...Object.entries(namedCatalogs ?? {}).map(([name, dependencies]) => ({
+      name,
+      dependencies,
+    })),
+  ];
```
secustor, line 338:
```suggestion
  const result = [
    {
      name: 'default',
      dependencies: defaultCatalogDeps ?? {},
    }
  ]

  if (!namedCatalogs) {
    return result;
  }

  for (const [name, dependencies] of Object.entries(namedCatalogs)) {
    result.push({
      name,
      dependencies,
    });
  }

  return result
```
Better readability.
https://github.com/renovatebot/renovate/blob/main/docs/development/best-practices.md#write-understandable-code

lib/modules/manager/npm/update/dependency/pnpm.spec.ts
```diff
@@ -0,0 +1,336 @@
...
+ * Per the YAML spec, a document ends with a newline. The 'yaml' library always
+ * uses that when serialising, but `codeBlock` strips the last indentation. This
+ * helper makes assertions simpler.
+ */
+function yamlCodeBlock(
+  literals: TemplateStringsArray,
+  ...placeholders: any[]
+): string {
+  return codeBlock(literals, placeholders) + '\n';
+}
```
secustor, line 14: If this is the case it should be moved to `lib/util/yaml`

lib/modules/manager/npm/update/dependency/pnpm.spec.ts
```diff
@@ -0,0 +1,336 @@
...
+        - pkg-a
+
+      catalog:
+        react: 19.0.0
+    `;
+    const testContent = npmUpdater.updateDependency({
+      fileContent: pnpmWorkspaceYaml,
+      upgrade,
+    });
+    expect(testContent).toEqual(expected);
```
secustor, line 44:
```suggestion
    const testContent = npmUpdater.updateDependency({
      fileContent: pnpmWorkspaceYaml,
      upgrade,
    });
    expect(testContent).toEqual(yamlCodeBlock`
      packages:
        - pkg-a

      catalog:
        react: 19.0.0
    `;);
```
Inline expected for better readability. Same goes for the other tests

lib/modules/manager/npm/update/dependency/pnpm.ts
```diff
@@ -0,0 +1,104 @@
...
+    );
+    return null;
+  }
+
+  let { newValue } = upgrade;
+
+  newValue = getNewGitValue(upgrade) ?? newValue;
+  newValue = getNewNpmAliasValue(newValue, upgrade) ?? newValue;
+
+  logger.debug(
```
secustor, line 29:
```suggestion
  logger.trace(
```

lib/modules/manager/npm/update/dependency/pnpm.ts
secustor: Does your solution keep whitespaces and comments?  

If yes, you can keep it like it is.

lib/modules/manager/npm/update/dependency/common.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+}
+
+export function getNewNpmAliasValue(
+  value: string | undefined,
+  upgrade: Upgrade,
+): string | undefined {
+  if (!upgrade.npmPackageAlias) {
+    return;
+  }
```
secustor, line 31:
```suggestion
): string | null {
  if (!upgrade.npmPackageAlias) {
    return null;
  }
```
Code base convention is to never return actively `undefined`

lib/modules/manager/npm/update/dependency/common.ts
```diff
@@ -0,0 +1,33 @@
+import { logger } from '../../../../../logger';
+import type { Upgrade } from '../../../types';
+
+export function getNewGitValue(upgrade: Upgrade): string | undefined {
+  if (!upgrade.currentRawValue) {
+    return;
+  }
```
secustor, line 7:
```suggestion
export function getNewGitValue(upgrade: Upgrade): string | null {
  if (!upgrade.currentRawValue) {
    return null;
  }
```

secustor (commented):
I had it on my plate, but haven't come around to finish it.

> Open question 1: ways of picking up pnpm-workspace.yaml for extraction

`pnpm-workspace.yaml` should be part of fileMatch. 
The expectation is that users will have use cases which deviating filenames e.g. because they are overwriting this name in a config. 
Ideally we detect the fact that it is a workspace file via schema or other marker and [are not relying on it](https://github.com/renovatebot/renovate/pull/33376/files#diff-43bee6aaab05445cea15363ac769893671dc75dcc55db59b50801e1a6c97e8f5R235). That way we are not bound to the filename and it "magically" works. 

>Open question 2: depType and allowing granular updates to named catalogs

Having `pnpm.catalog.foo` is preferable as with [string pattern matching](https://docs.renovatebot.com/string-pattern-matching/) users can match all catalogs if needed.
Can you add an example of this in the `readme.md` that way it will show up in the manager docs.

rarkins:
@secuster can't we discover the pnpm-workspace.yaml file automatically instead of needing it in fileMatch?

secustor:
> @secuster can't we discover the pnpm-workspace.yaml file automatically instead of needing it in fileMatch?

If we are fine with only supporting this file name, we can drop it out of fileMatch.
[There are requests regarding the possiblitly to change the file name and workarounds](https://github.com/pnpm/pnpm/issues/1291#issuecomment-637685956)

lib/modules/manager/npm/extract/pnpm.ts
viceice: and use interface instead of type

lib/modules/manager/npm/extract/types.ts
```diff
@@ -36,6 +36,8 @@ export interface LockFile {
 
 export interface PnpmWorkspaceFile {
   packages: string[];
+  catalog?: Partial<Record<string, string>>;
+  catalogs?: Partial<Record<string, Partial<Record<string, string>>>>;
```
viceice, line 40: partial makes no sense here, drop it

lib/modules/manager/npm/update/dependency/pnpm.spec.ts
viceice: no, it's only used for tests, so should be somewhere in `test/...`

wonder why it's needed at all 🤔 we use code block in other tests with yaml

lib/modules/manager/npm/post-update/index.ts
```diff
@@ -242,7 +242,12 @@ export async function writeUpdatedPackageFiles(
       await writeLocalFile(packageFile.path, packageFile.contents!);
       continue;
     }
-    if (!packageFile.path.endsWith('package.json')) {
+    if (
+      !(
+        packageFile.path.endsWith('package.json') ||
+        packageFile.path.endsWith('pnpm-workspace.yaml')
```
fpapado, line 248: The hardcoding here tripped me up, I kept not seeing the lockfiles changing 😅 Does this not beat some of the point of allowing `pnpm-workspace.yaml` (and `package.json`) being renamed via `fileMatch`? I wonder what the original context for the check was.

fpapado:
Thank you all kindly for your earler reviews! They helped clarify the direction quite a bit. I've made relevant changes now, tidied up the tests, implemented lockfile updating and locked version detection, and the coverage should be ok too. 

I'll put this up for a review now. Let me know if you spot anything that can be improved, and I'll get to it 🔧

lib/modules/manager/npm/extract/pnpm.ts
```diff
@@ -222,3 +250,98 @@ function getLockedDependencyVersions(
 
   return res;
 }
+
+export function isPnpmWorkspaceYaml(content: string): boolean {
+  try {
+    parseSingleYaml(content, { customSchema: PnpmWorkspaceFileSchema });
```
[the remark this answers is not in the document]
fpapado: Yeah, I agree, it is quite wasteful to parse twice for the happy path (once to check, once to extract).

What if instead of an `isPnpmWorkspaceYaml` function, we just used the parsing function directly, keeping the result on success? We could then pass the parsed object directly to the extraction function, meaning that we only need to parse once. It would make the code a bit less neat, but I think it makes sense as a performance optimisation.

I'll sketch the above in a commit 💭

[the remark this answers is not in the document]
fpapado: Made the relevant changes at 917dd6aa4f36f09a26f88e503db5d0e5d499af88. The extraction call-site is not as nice now perhaps, but I think the tradeoff is worth it. Please let me know what you think 😌

lib/modules/manager/npm/extract/pnpm.spec.ts
```diff
@@ -264,10 +266,73 @@ describe('modules/manager/npm/extract/pnpm', () => {
       expect(Object.keys(res.lockedVersionsWithPath!)).toHaveLength(1);
     });
 
+    it('extracts version from catalogs', async () => {
+      const plocktest1Lock = Fixtures.get(
+        'lockfile-parsing/pnpm-lock-v9.yaml',
+        '..',
+      );
+      jest.spyOn(fs, 'readLocalFile').mockResolvedValueOnce(plocktest1Lock);
```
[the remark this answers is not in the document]
fpapado: This was a slightly larger change than I anticipated. The rest of the tests are using `spyOn` for specific functions, and letting other calls of the `fs` module fall-through to the real implementation. 

It seemed confusing to have both strategies in the same file, especially given that the `fs` test utils assume that the whole module is mocked. So I changed the other test cases here to use the test utils, since it also seemed an outlier among other tests files. The relevant commits are in bdc4d03d2433a7669f26830386bab620b4956de3 and 7d1189dd43bef2a4010f7b23ad6b2037e601e00a. Let me know if this looks out of scope, I could split it off to another PR after (or parallel to) this one 🗒️

lib/modules/manager/npm/post-update/index.ts
secustor: K, I think this will lead to too much changes in this PR. :/

Let's support `pnpm-workspace.yaml` and then support more names as needed.

lib/modules/manager/npm/extract/pnpm.spec.ts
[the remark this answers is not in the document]
secustor: Yes, please move it to a separate PR. You can tag me in there to get merged fast.

secustor (commented):
Sry, for the back and forth.

lib/modules/manager/npm/extract/pnpm.spec.ts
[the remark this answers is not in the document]
fpapado: Thank you! I opened https://github.com/renovatebot/renovate/pull/33756 now. I will rebase this PR after merging it, just to tidy up a bit 🧹

lib/modules/manager/npm/post-update/index.ts
fpapado: Yeah, it turned out to be a bit complicated indeed 😅  Would you rather I keep `pnpm-workspace.yaml` in `fileMatch` still? It probably does not hurt to keep it there, just to avoid any more changes, but I'm happy to remove it as well.

secustor (commented):
LGTM, but there is a conflict with `main`

fpapado:
> LGTM, but there is a conflict with `main`

Oh right, there was a conflict in the npm README around `depTypes`. Rebased and resolved it now ✅

secustor (approved):
Please also no force pushes, your branch will automatically get rebased and squashed anyway

Thanks for your contribution!

gluxon:
@fpapado @secustor I love this! Thank you both so much for adding this feature. ❤️

PR #33382, opened 2025-01-03
fix(config): encrypted field validation

RahulGautamSingh:
## Changes
- Throw error if `encrypted` field found without `privateKey` & `RENOVATE_X_ENCRYPTED_STRICT=true` resulting in a config issue being raised

## Context
- Ref: https://github.com/renovatebot/renovate/discussions/33224

1. Self Hosted: 
- [Encrypted Config in the root config](https://github.com/Rahul-renovate-testing/repro-33224-2)
- [Encrypted Config within hostRules](https://github.com/Rahul-renovate-testing/repro-33224-1)
- [Encrypted Config in preset](https://github.com/Rahul-renovate-testing/repro-33224-3)

2. [Mend Hosted (with `MEND_HOSTED=true` in env)](https://github.com/Rahul-renovate-testing/repro-33224-mend)

### DIFF
--- a/lib/config/decrypt.ts
+++ b/lib/config/decrypt.ts
@@ -130,12 +130,15 @@ export async function decryptConfig(
   logger.trace({ config }, 'decryptConfig()');
   const decryptedConfig = { ...config };
   const privateKey = GlobalConfig.get('privateKey');
+  // eslint-disable-next-line
+  console.log(GlobalConfig.get());
   const privateKeyOld = GlobalConfig.get('privateKeyOld');
   for (const [key, val] of Object.entries(config)) {
     if (key === 'encrypted' && is.object(val)) {
       logger.debug({ config: val }, 'Found encrypted config');
 
       const encryptedWarning = GlobalConfig.get('encryptedWarning');
+      // istanbul ignore if: todo
       if (is.string(encryptedWarning)) {
         logger.once.warn(encryptedWarning);
       }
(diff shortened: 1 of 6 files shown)


viceice (commented):
this is a breaking change, we should do it in a major version bump

lib/config/validation.ts
```diff
@@ -634,6 +634,26 @@ export async function validateConfig(
...
+              if (res !== true) {
+                errors.push({
+                  topic: 'Configuration Error',
+                  message: `Invalid \`${currentPath}.${key}.${res}\` configuration: value is not a string`,
+                });
+              }
+
+              const privateKey =
+                configType === 'global'
+                  ? (config.privateKey ?? config.privateKeyOld)
```
rarkins, line 648: When decrypting, does renovate try privateKeyOld if no privateKey is found?

lib/config/validation.ts
```diff
@@ -634,6 +634,26 @@ export async function validateConfig(
...
+                  : (GlobalConfig.get('privateKey') ??
+                    GlobalConfig.get('privateKeyOld'));
+              if (!is.nonEmptyString(privateKey)) {
+                errors.push({
+                  topic: 'Configuration Error',
+                  message: `No privateKey found in hosted config. \`encrypted\` object cannot be used without a privateKey.`,
```
rarkins, line 654: This error needs to be different when MEND_HOSTED="true". See discussion

rarkins (changes requested):

lib/config/validation.ts
viceice: can we use `config.strings` for it instead of adding a mend specific env? so self hosted users can define custom message.

lib/config/validation.ts
viceice: private key should be required if private key old is defined.

rarkins: We need to model the current behavior

justo-mend:
@RahulGautamSingh Some tests are failing. Are you able to get them to pass?

justo-mend:
Thanks @RahulGautamSingh

lib/config/decrypt.ts
```diff
@@ -179,6 +179,12 @@ export async function decryptConfig(
           error.validationSource = 'config';
           error.validationError = 'Encrypted config unsupported';
           error.validationMessage = `This config contains an encrypted object at location \`$.${key}\` but no privateKey is configured. To support encrypted config, the Renovate administrator must configure a \`privateKey\` in Global Configuration.`;
+          if (process.env.MEND_HOSTED === 'true') {
+            error.validationMessage = `Mend-hosted Renovate Apps no longer support the use of encrypted secrets in Renovate file config (ie. renovate.json).
```
rarkins, line 183:
```suggestion
            error.validationMessage = `Mend-hosted Renovate Apps no longer support the use of encrypted secrets in Renovate file config (e.g. renovate.json).
```

rarkins (changes requested):

rarkins (dismissed):
Tested on real repo?

RahulGautamSingh:
> Tested on real repo?

Yes, linked in description

RahulGautamSingh:
Also, don't think it's breaking anymore.

rarkins (approved):

justo-mend:
Thanks @RahulGautamSingh !

PR #33416, opened 2025-01-05
feat(nuget): Support for variables

zharinov:


lib/modules/manager/nuget/__snapshots__/extract.spec.ts.snap
```diff
@@ -92,17 +92,12 @@ exports[`modules/manager/nuget/extract extractPackageFile() extracts all depende
     "depName": "Microsoft.AspNetCore.Hosting",
     "depType": "nuget",
   },
-  {
-    "currentValue": "4.1.0",
-    "datasource": "nuget",
-    "depName": "Autofac.Extensions.DependencyInjection",
-    "depType": "nuget",
-  },
```
rarkins, line 100: Can this be retained but skipped?

rarkins (approved):

rarkins (changes requested):
Needs deconflicting

lib/modules/manager/nuget/__fixtures__/sample.csproj
```diff
@@ -3,15 +3,16 @@
...
   </PropertyGroup>
 
   <ItemGroup>
     <Folder Include="Controllers\" />
     <Folder Include="wwwroot\" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Autofac" Version="4.5.0" />
-    <PackageReference Include="Autofac.Extensions.DependencyInjection" Version="4.1.0" />
+    <PackageReference Include="Autofac" Version="$(AutofacVersion)" />
```
AArnott, line 14: Should we have a test where the msbuild property is used by _multiple_ packages (which matches my usual case)?
I'm curious whether renovate will treat that as a 'group' by the name of the variable automatically, or somehow misbehave if package1 has a newer version available than package2, and somehow the data source tries to update the property to match the highest version of the two rather than the highest *common* version of the two.

zharinov: Yes, but let's keep this out of scope of this PR, because if it's handled at all, it's outside of Nuget.

rarkins: I think it's relevant to show in the extract testing that two dependencies can share one variable

rarkins (approved):

PR #33453, opened 2025-01-07
feat(gradle): bump strategy

ldt-fweichert:
## Changes

- added support for the bump strategy for gradle versioning using the maven style range syntax
- added support for maven style range syntax to the gradle manager
- [repro repo](https://github.com/ldt-fweichert/renotave-gradle-repro)

## Todos
- [x] Sign the CLA
- [x] Fix last failling test
- [x] Do code //TODO 's
- [ ] Get feedback on the implementation
- [ ] Add docs


lib/modules/manager/gradle/parser.ts
```diff
@@ -127,11 +127,12 @@ export function parseProps(
 ): { vars: PackageVariables; deps: PackageDependency<GradleManagerData>[] } {
   let offset = 0;
   const vars: PackageVariables = {};
-  const deps: PackageDependency[] = [];
+  const deps: PackageDependency<GradleManagerData>[] = [];
   for (const line of input.split(newlineRegex)) {
     const lineMatch = propRegex.exec(line);
     if (lineMatch?.groups) {
       const { key, value, leftPart } = lineMatch.groups;
+      // TODO: we dont need to check isDependency, the parseDependency does this by itself
```
Churro, line 135: That's correct but please keep your changes strictly limited to what's relevant for the functionality you'd like to add. Refactoring can be done separately.

lib/modules/manager/gradle/utils.ts
```diff
@@ -44,6 +53,11 @@ export function isDependencyString(input: string): boolean {
     return false;
   }
 
+  const mavenStyleRange = parseMavenStyleRange(tempVersionPart);
+  if (mavenStyleRange) {
+    return true;
```
Churro, line 58: This is wrong as it bypasses all subsequent checks

lib/modules/manager/gradle/utils.ts
```diff
@@ -44,6 +53,11 @@ export function isDependencyString(input: string): boolean {
     return false;
   }
 
+  const mavenStyleRange = parseMavenStyleRange(tempVersionPart);
```
Churro, line 56: I see no need for this here. Should be enough to have `versionLikeSubstring` match ranges, which it already can except for those containing `!` and spaces.

lib/modules/manager/gradle/utils.ts
```diff
@@ -1,22 +1,20 @@
...
-} from './types';
+import type { GradleManagerData, PackageVariables, VariableRegistry } from './types';
 
 const artifactRegex = regEx(
-  '^[a-zA-Z][-_a-zA-Z0-9]*(?:\\.[a-zA-Z0-9][-_a-zA-Z0-9]*?)*$',
+  '^[a-zA-Z][-_a-zA-Z0-9]*(?:\\.[a-zA-Z0-9][-_a-zA-Z0-9]*?)*$'
 );
 
-const versionLikeRegex = regEx('^(?<version>[-_.\\[\\](),a-zA-Z0-9+]+)');
+const versionLikeRegex = regEx('^(?<version>[-_.\\[\\](),a-zA-Z0-9+! ]*[-_.\\[\\](),a-zA-Z0-9+])');
```
Churro, line 10: What's the intention here? Is it an attempt to add `!` and a space to the list of allowed characters? If yes, this can be done simpler

lib/modules/versioning/gradle/compare.ts
```diff
@@ -253,6 +253,84 @@ export function parsePrefixRange(input: string): PrefixRange | null {
...
+
+export type MavenStyleRangeBound = {
+  version: string | null,
+  type: 'inclusive' | 'exclusive',
+  bracket: string,
+}
+
+const mavenStyleRangeRegex = regEx(/^(?<openingBracket>[[\](])(?<lowerBound>[-._+a-zA-Z0-9]+)?(?<seperator>\s*,\s*)(?<upperBound>[-._+a-zA-Z0-9]+)?(?<closingBracket>[[\])])(?:!!(?<preferredVersion>[-._+a-zA-Z0-9]+))?$/)
+
+export function parseMavenStyleRange(input: string): MavenStyleRange | null {
```
Churro, line 271: There already is `mavenBasedRangeRegex` and `parseMavenBasedRange`, which does quite the same, except that it lacks support for preferred versions. What speaks against extending the existing implementation instead of reinventing maven style range parsing as a whole?

lib/modules/versioning/gradle/index.ts
```diff
@@ -128,30 +131,31 @@ const matches = (a: string, b: string): boolean => {
     return equals(x, y);
   }
 
-  const mavenBasedRange = parseMavenBasedRange(b);
```
Churro, line 131: This looks wrong to me. It seems you're trying to replace `parseMavenBasedRange()` with your implementation, while it's still used e.g. in `isValid()`.

lib/modules/versioning/gradle/index.ts
```diff
@@ -128,30 +131,31 @@ const matches = (a: string, b: string): boolean => {
...
-  if (!mavenBasedRange) {
+  const mavenStyleRange = parseMavenStyleRange(b);
+  // we can not update an exclusive lower bound as we don't know
```
Churro, line 135: Do you mean `(1.0,)` for example? If so, this should be reflected in a test too

lib/modules/manager/gradle/utils.ts
```diff
@@ -70,23 +84,33 @@ export function isDependencyString(input: string): boolean {
...
     return null;
   }
-  const [groupId, artifactId, FullValue] = input.split(':');
-  if (FullValue === versionLikeSubstring(FullValue)) {
+  const [groupId, artifactId, fullValue] = input.split(':');
+  if (fullValue === versionLikeSubstring(fullValue)) {
+
+    const mavenStyleRange = parseMavenStyleRange(fullValue);
+    if (mavenStyleRange) {
+      const updatablePart = mavenStyleRange.upperBound ?? mavenStyleRange.lowerBound;
```
Churro, line 97: I guess `?? mavenStyleRange.lowerBound` can never hapepns since `parseMavenStyleRange` returns null if not both lower and upper are provided?!

Churro (changes requested):
Thanks for the contribution. Looks promising but can benefit from some refinement:
* Please add a reproduction repo so that your changes can be tested against, incl a dep with regular maven based range and one with a range + strict version enforcement.
* Run `prettier` and undo changes unrelated to the bump strategy.
* Keep dep extraction to the minimum, it's not the place to decide what part to bump

**Most importantly:** To reduce the complexity of your changes, it would be great if you could limit this PR to the already supported Maven range syntax, for now ignoring gradle's strict version notation. This should allow you to reuse the existing maven range parsing, it should work without changes to the parser and, unless I missed something, only require a few adaptations in the gradle versioning

ldt-fweichert:
Thank you for the quick and precise feedback.

> **Most importantly:** To reduce the complexity of your changes, it would be great if you could limit this PR to the already supported Maven range syntax, for now ignoring gradle's strict version notation. This should allow you to reuse the existing maven range parsing, it should work without changes to the parser and, unless I missed something, only require a few adaptations in the gradle versioning

The primary driver for this PR was the need for Gradle's strict version notation at my company. It allows us to specify: *yes, you can use a newer dependency if you want, but if there are options to choose from, prefer the one specified with the '!!'* Unfortunately, this functionality isn’t achievable without adding Gradle’s syntax.

Would your suggested path forward be to first implement Maven-only ranges (using the existing Maven module, which seems feasible), and then add Gradle syntax in a follow-up? If so, is there a way to minimize complexity and avoid lengthy delays? I work part-time (twice a week) and want to ensure this progresses smoothly without causing blockers.

lib/modules/manager/gradle/utils.ts
ldt-fweichert: The intention was to allow `!` and ` ` as chars, but only if they are not the last chars in the sequence. There were some false positives before i did this change

lib/modules/versioning/gradle/index.ts
ldt-fweichert: Yes i've `meant (1.0,)`, will add it to the tests and fix the comment

lib/modules/manager/gradle/utils.ts
ldt-fweichert: Only when both are null the result is null, a single bound is also ok: `if (!lowerBound && !upperBound) {return null;}`

lib/modules/manager/gradle/utils.ts
ldt-fweichert: same here

lib/modules/manager/gradle/utils.ts
Churro: Correct, it is more of a fuzzy search and no strict version input validation.

lib/modules/manager/gradle/utils.ts
Churro: These chars need to be added to match gradle strict versions but the pattern you suggested doesn't protect against implausible versions. For ex., it still matches ` ! ! !-4`. The current pattern doesn't protect against invalid versions like `--2` either but as said this is not what is intended to be asserted here.

lib/modules/manager/gradle/utils.ts
```diff
@@ -70,23 +84,33 @@ export function isDependencyString(input: string): boolean {
...
+      return {
+        depName: `${groupId}:${artifactId}`,
+        currentValue: updatablePart,
```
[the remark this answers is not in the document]
ldt-fweichert: right, is removed

lib/modules/versioning/gradle/compare.ts
ldt-fweichert: It was more of a learning strategy for me so disregarading the old implementation was not my intention. In the upcoming pull-request for gradles maven-style version syntax i will mostly stick to the original implementation.

ldt-fweichert: This pull-request now only adds the already supported maven range syntax so i guess the change was irrelevant from the start: removed

lib/modules/versioning/gradle/index.ts
ldt-fweichert: Yeah, that was not my intention. The change has also been removed for the reason above

lib/modules/versioning/gradle/index.ts
ldt-fweichert: As this is no longer in the scope of this pull request, these changes have also been removed regardless

lib/modules/manager/gradle/utils.ts
ldt-fweichert: As this is no longer in the scope of this pull request, these changes have also been removed regardless

ldt-fweichert:
I've adjusted this merge-request to match the new scope.
> Please add a reproduction repo so that your changes can be tested against, incl a dep with regular maven based range and one with a range + strict version enforcement.

Is there a guide on how to do that or a pull request that had the same requirement where I could take a look at?

lib/modules/manager/gradle/__snapshots__/parser.spec.ts.snap
```diff
@@ -51,51 +51,59 @@ exports[`modules/manager/gradle/parser calculations parses fixture from "gradle"
       "packageFile": "build.gradle",
     },
   },
+  {
+    "currentValue": "[3.3.0, ]",
+    "depName": "org.springframework.boot:spring-boot-starter-jetty",
```
Churro, line 56: This test isn't needed here as there are no changes to finding `fileReplacePosition`. Plus you have newly added tests in `parser.spec.ts` that check the same

Churro (changes requested):
Regarding a minimal reproduction, take a look here -> https://github.com/renovatebot/renovate/blob/main/docs/development/minimal-reproductions.md

Please also rebase your changes on the current main branch.

lib/modules/manager/gradle/__snapshots__/parser.spec.ts.snap
ldt-fweichert: Done!

lib/modules/manager/gradle/utils.spec.ts
```diff
@@ -23,7 +23,6 @@ describe('modules/manager/gradle/utils', () => {
       expect(versionLikeSubstring(`${input}'`)).toEqual(input);
       expect(versionLikeSubstring(`${input}"`)).toEqual(input);
       expect(versionLikeSubstring(`${input}\n`)).toEqual(input);
-      expect(versionLikeSubstring(`${input}  `)).toEqual(input);
```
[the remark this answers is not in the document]
ldt-fweichert: Done!

ldt-fweichert:
I did the changes you've suggest, the repro will have to wait until my next workday this week. Thanks for the link!

Churro:
You may also want to add `bump` here:
https://github.com/renovatebot/renovate/blob/a6151daabf0f02ee0c06d24916d4c3c5815613ed/lib/modules/versioning/gradle/index.ts#L21

ldt-fweichert:
I am trying to get the repro running right now.
[The repro repo is here](https://github.com/ldt-fweichert/renotave-gradle-repro) \
To verify that my test setup is working correctly i wanted to try the already implemented pin strategy first. The weird thing that even that doesn't even seem to work. The version update fails due to a wrongly positioned `fileReplacePosition`:
grafik
Have i misconfigured something?
I am running renovate locally using `pnpm debug ldt-fweichert/renotave-gradle-repro`, nothing unusual there i guess.

Churro:
> Have i misconfigured something? I am running renovate locally using `pnpm debug ldt-fweichert/renotave-gradle-repro`, nothing unusual there i guess.

Your [build.gradle.kts](https://github.com/ldt-fweichert/renotave-gradle-repro/blob/main/app/build.gradle.kts) uses Unix line endings, so `fileReplacePosition` 814 is correct for `io.ktor:ktor-server-netty:3.0.0`. See https://github.com/renovatebot/renovate/blob/main/docs/development/best-practices.md#windows on how to resolve your CRLF issue.

Churro (approved):
Thanks for the updates. The [documentation](https://docs.renovatebot.com/modules/versioning/gradle/) for gradle-versioning will be auto-updated and include `supportedRangeStrategies: RangeStrategy[] = ['pin', 'bump'];`

rarkins (dismissed):

lib/modules/manager/gradle/utils.ts
```diff
@@ -11,7 +11,7 @@ const artifactRegex = regEx(
   '^[a-zA-Z][-_a-zA-Z0-9]*(?:\\.[a-zA-Z0-9][-_a-zA-Z0-9]*?)*$',
 );
 
-const versionLikeRegex = regEx('^(?<version>[-_.\\[\\](),a-zA-Z0-9+]+)');
+const versionLikeRegex = regEx('^(?<version>[-_.\\[\\](),a-zA-Z0-9+ ]+)');
```
viceice, line 14: can we make it more strict, so it doesn't allow `[1.6.0, ]  ,  abc` which is invalid obviously.
Let's only allow spaces inside braces 🤔 

This change should be in a separate PR, because it's totally unrelated to the `bump` range strategy

viceice (changes requested):

lib/modules/manager/gradle/utils.ts
Churro: How about leaving the version extraction as-is but adding a call to `gradleVersioning.isValid(...)` in `isDependencyString()`?

lib/modules/manager/gradle/utils.ts
```diff
@@ -1,5 +1,6 @@
 import upath from 'upath';
 import { regEx } from '../../../util/regex';
+import { api as gradleVersioning } from '../../versioning/gradle/index';
```
viceice, line 3:
```suggestion
import { api as gradleVersioning } from '../../versioning/gradle';
```

viceice (dismissed):

viceice:
needs prettier run

viceice (approved):

rarkins:
Test failures appear to have been a temporary problem

Issue #31891, opened 2024-10-10
Strongly type timestamps


rarkins:
### Describe the proposed change(s).

Similar to what was done for `LongCommitSha`, we should add a strongly typed `Timestamp` type, which is a validated Timestamp. We should also consistently use `luxon` `DateTime` capabilities throughout if we're not already.

```
export type Timestamp = string & { __timestamp: never };
```

Our datasource logic should be updated to only return `releaseTimestamp` if it's of type `Timestamp`. Individual datasource implementations could possibly return invalid strings, but they should be filtered before being returned to the worker layer.

PR #33512, opened 2025-01-10
refactor: Typed timestamp utility

zharinov:
## Context

- Ref: #31891


lib/util/timestamp.ts
```diff
@@ -0,0 +1,75 @@
...
+
+function isValid(date: DateTime): boolean {
+  const tomorrow = DateTime.now().toMillis() + tomorrowOffset; // 24 * 60 * 60 * 1000;
+  const ts = date.toMillis();
+  return date.isValid && ts > millenium && ts < tomorrow;
+}
+
+export function asTimestamp(input: unknown): Timestamp | null {
+  if (input instanceof Date) {
+    const date = DateTime.fromJSDate(input, { zone: 'UTC' });
```
viceice, line 18: don't we change the timezone to a probably wrong value here? or is datetime correctly converting it.

lib/util/timestamp.ts
```diff
@@ -0,0 +1,75 @@
...
+    return null;
+  }
+
+  if (typeof input === 'number') {
+    const millisDate = DateTime.fromMillis(input, { zone: 'UTC' });
+    if (isValid(millisDate)) {
+      return millisDate.toISO() as Timestamp;
+    }
+
+    const secondsDate = DateTime.fromSeconds(input, { zone: 'UTC' });
```
viceice, line 32: do we need this?

lib/util/timestamp.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export type Timestamp = string & { __timestamp: never };
+
+const timezoneOffset = new Date().getTimezoneOffset() * 60000;
+
+const millenium = 946684800000; // 2000-01-01T00:00:00.000Z
+const tomorrowOffset = 86400000; // 24 * 60 * 60 * 1000;
+
+function isValid(date: DateTime): boolean {
+  const tomorrow = DateTime.now().toMillis() + tomorrowOffset; // 24 * 60 * 60 * 1000;
```
viceice, line 11: I would check the valid property first and early return, so we skip  some computation.

lib/util/timestamp.ts
zharinov: If some datasource is capable of providing timestamp in the millisecond format, I could imagine it's done in the seconds format as well

lib/util/timestamp.ts
zharinov: It's converted from whatever timezone the original value is, to the UTC

zharinov: BTW this branch is unlikely to be used at all

viceice (approved):

PR #33517, opened 2025-01-10
feat(presets): add smithy-typescript to aws-sdk-js monorepo

TheHolyWaffle:
## Changes

Add `https://github.com/smithy-lang/smithy-typescript` to the `aws-sdk-js-v3` monorepo group

## Context

Ref: https://github.com/renovatebot/renovate/discussions/33507

I followed the same strategy as for the `smithy-rs` repository.

### DIFF
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -47,7 +47,10 @@
     "aws-sdk-client-mock": "https://github.com/m-radzikowski/aws-sdk-client-mock",
     "aws-sdk-go": "https://github.com/aws/aws-sdk-go",
     "aws-sdk-go-v2": "https://github.com/aws/aws-sdk-go-v2",
-    "aws-sdk-js-v3": "https://github.com/aws/aws-sdk-js-v3",
+    "aws-sdk-js-v3": [
+      "https://github.com/aws/aws-sdk-js-v3",
+      "https://github.com/smithy-lang/smithy-typescript"
+    ],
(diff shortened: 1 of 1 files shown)


lib/data/monorepo.json
```diff
@@ -47,7 +47,10 @@
     "aws-sdk-client-mock": "https://github.com/m-radzikowski/aws-sdk-client-mock",
     "aws-sdk-go": "https://github.com/aws/aws-sdk-go",
     "aws-sdk-go-v2": "https://github.com/aws/aws-sdk-go-v2",
-    "aws-sdk-js-v3": "https://github.com/aws/aws-sdk-js-v3",
+    "aws-sdk-js-v3": [
+      "https://github.com/aws/aws-sdk-js-v3",
+      "https://github.com/smithy-lang/smithy-typescript"
```
rarkins, line 52: This is not right

TheHolyWaffle: What's the suggested approach?

I followed the same strategy as with `aws-sdk-rust` monorepo:

```
    "aws-sdk-rust": [
      "https://github.com/smithy-lang/smithy-rs",
      "https://github.com/awslabs/smithy-rs",
      "https://github.com/awslabs/aws-sdk-rust"
    ],
```

viceice: we probably should check all our monorepos and validate if they share same version or absolutely need to be grouped. we should remove all which don't share same version and don't need to be merged together.

PR #33523, opened 2025-01-10
refactor(result): Use star import to simplify Zod types usage

zharinov:

### DIFF
--- a/lib/util/result.ts
+++ b/lib/util/result.ts
@@ -1,5 +1,4 @@
-import type { SafeParseReturnType, ZodType, ZodTypeDef } from 'zod';
-import { ZodError, z } from 'zod';
+import * as z from 'zod';
 import { logger } from '../logger';
 
 type Val = NonNullable<unknown>;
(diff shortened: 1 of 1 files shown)


lib/util/result.ts
```diff
@@ -1,5 +1,4 @@
-import type { SafeParseReturnType, ZodType, ZodTypeDef } from 'zod';
-import { ZodError, z } from 'zod';
+import * as z from 'zod';
```
viceice, line 1: I don't like star imports. they block unused code stripping when bundling

viceice (commented):
I don't see a benefit here. looks more like a step back

lib/util/result.ts
zharinov: I don't see any difference for the code we generate:

image

image

Do we apply any dead code stripping for `node_modules`?

I created this PR to make the code less messy before further refactoring, i.e. to make the zod-specific types be easily detectable in the code.

viceice: we plan to bundle renovate some time in future and it would be great if dead code would be dropped. I think star imports don't allow it.

PR #33526, opened 2025-01-10
fix(gradle): remove redundant checks and parser statements

Churro:
## Changes

- `parser.ts`:
  - Remove `isDependencyString()` since `parseDependencyString()` does it too
- `common.ts`:
  - 3 parsing clauses had the same handler redundantly -> unified
- `utils.ts`:
  - `isDependencyString()`: shortened a check to `versionPart.includes('@')` because whenever it is true, `versionLikeSubstring(versionPart)` is false (since it doesn't match `@`). Saves a very frequently executed regex query.
  - `parseDependencyString()`: removed the call to `versionLikeSubstring()` since we already know after `isDependencyString()` that a given dependency string is valid.

... and some other minor refactorings.

### DIFF
--- a/lib/modules/manager/gradle/parser.ts
+++ b/lib/modules/manager/gradle/parser.ts
@@ -18,7 +18,7 @@ import type {
   PackageVariables,
   ParseGradleResult,
 } from './types';
-import { isDependencyString, parseDependencyString } from './utils';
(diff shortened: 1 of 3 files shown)


viceice (approved):
looks more like a fix to me

PR #33577, opened 2025-01-11
feat(helm): Use schema for datasource

zharinov:


lib/modules/datasource/helm/index.ts
```diff
@@ -34,63 +32,22 @@ export class HelmDatasource extends Datasource {
...
   })
-  async getRepositoryData(
-    helmRepository: string,
-  ): Promise<HelmRepositoryData | null> {
-    let res: HttpResponse<string>;
-    try {
-      res = await this.http.get('index.yaml', {
+  async getRepositoryData(helmRepository: string): Promise<HelmRepositoryData> {
+    const { val, err } = await Result.wrap(
+      this.http.get('index.yaml', {
```
viceice, line 39: we should add a `Http.getYaml` function similar to the `getJson` , so we can simply it further

lib/modules/datasource/helm/schema.ts
```diff
@@ -0,0 +1,68 @@
...
+import type { Release } from '../types';
+import { githubRelease, isPossibleChartRepo } from './common';
+
+const HelmReleaseSchema = z.object({
+  version: z.string(),
+  created: z.string().nullable().catch(null),
+  digest: z.string().optional().catch(undefined),
+  home: z.string().optional(),
+  sources: z.array(z.string()).optional(),
+  urls: z.array(z.string()),
```
viceice, line 12:
```suggestion
  urls: z.array(z.string()).optional (),
```
it's not a mandatory helm field

lib/modules/datasource/helm/schema.ts
```diff
@@ -0,0 +1,68 @@
...
+});
+type HelmRelease = z.infer<typeof HelmReleaseSchema>;
+
+export function findSourceUrl(release: HelmRelease): string | null {
```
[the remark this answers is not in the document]
zharinov: It's the part of data extraction/transform, it's not used anywhere else

lib/modules/datasource/helm/schema.ts
```diff
@@ -0,0 +1,82 @@
...
+import { LooseRecord } from '../../../util/schema-utils';
+import type { Release } from '../types';
+
+const HelmReleaseSchema = z.object({
+  version: z.string(),
+  created: z.string().nullable().catch(null),
+  digest: z.string().optional().catch(undefined),
+  home: z.string().optional().catch(undefined),
+  sources: z.array(z.string()).catch([]),
+  urls: z.array(z.string()).catch([]),
```
viceice, line 14: so catch ensures it still works if one of those is missing?

viceice (commented):
otherwise LGTM

lib/modules/datasource/helm/schema.ts
zharinov: Yes, exactly

viceice (approved):

PR #33579, opened 2025-01-11
feat(batect): Use Zod schema for extract code

zharinov:


lib/modules/manager/batect/schema.ts
```diff
@@ -0,0 +1,63 @@
...
+    ).catch([]),
+  }),
+).transform(({ containers, include }) => {
+  const imageDependencies = containers.map((image) => ({
+    ...getDep(image),
+    versioning: dockerVersioning,
+  }));
+
+  const gitIncludes = [];
+  const fileIncludes = [];
```
viceice, line 38: add types, they otherwise any

viceice (commented):
otherwise LGTM

viceice:
we should use feat to force a release, so users can find the version more easily if it causes regressions

viceice (approved):

PR #33584, opened 2025-01-12
feat(github-actions): Use schema for dependency extraction

zharinov:


viceice (dismissed):

lib/modules/manager/github-actions/extract.ts
```diff
@@ -198,34 +172,31 @@ function extractWithYAMLParser(
...
-    // TODO: use schema (#9610)
-    pkg = parseSingleYaml(content);
-  } catch (err) {
-    logger.debug(
-      { packageFile, err },
-      'Failed to parse GitHub Actions Workflow YAML',
-    );
-    return [];
-  }
+  const jobs = WorkflowJobsSchema.parse(content);
```
viceice, line 175: please add some error logging, so we know if parsing fails (probably of some github changes)

viceice (approved):
we should move the action extract to schema too instead of continuing to use regex parsing

PR #33624, opened 2025-01-15
feat(preset): adding Azure and AzureAD monorepos (SDK and auth)

klippx:
## Changes

Added a couple of known monorepos related to Azure and AzureAD


lib/data/monorepo.json
```diff
@@ -56,10 +56,26 @@
...
+    "azure-azure-sdk-for-cpp": "https://github.com/Azure/azure-sdk-for-cpp",
+    "azure-azure-sdk-for-go": "https://github.com/Azure/azure-sdk-for-go",
+    "azure-azure-sdk-for-ios": "https://github.com/Azure/azure-sdk-for-ios",
+    "azure-azure-sdk-for-java": "https://github.com/Azure/azure-sdk-for-java",
+    "azure-azure-sdk-for-js": "https://github.com/Azure/azure-sdk-for-js",
+    "azure-azure-sdk-for-net": "https://github.com/Azure/azure-sdk-for-net",
+    "azure-azure-sdk-for-python": "https://github.com/Azure/azure-sdk-for-python",
+    "azure-azure-sdk-for-rust": "https://github.com/Azure/azure-sdk-for-rust",
+    "azure-azure-storage-net": "https://github.com/Azure/azure-storage-net",
     "azure-functions-dotnet-worker": "https://github.com/Azure/azure-functions-dotnet-worker",
```
klippx, line 71: If we want I could also change this

```suggestion
    "azure-azure-functions-dotnet-worker": "https://github.com/Azure/azure-functions-dotnet-worker",
```

rarkins: Do you think we really need the double azure-azure in the new ones? Is there any confusion or risk if a single azure- is used?

klippx: Taking a sample: `https://github.com/Azure/azure-sdk-for-rust`

I was thinking `${org}.${repoName}` so first I tried `Azure.azure-sdk-for-rust` but `.` were not allowed, as well as everything has to be lower case. So I had to replace `Azure` with `azure` and `.` with `-` that's how it was conceived.

In theory another org `MuchBetterAzure` can fork `azure-sdk-for-rust` and then how do we keep them apart?

Not likely this would happen but naming collisions might happen especially for more generically named repos such as `autocomplete` which today looks like this:

```
"algoliasearch-autocomplete": "https://github.com/algolia/autocomplete",
```

klippx:
@rarkins Should I add new entries to `orgGroups` ?

```json
  "orgGroups": {
    "azure": "https://github.com/Azure/",
    "azure-ad": "https://github.com/AzureAD/",
  }
```

I notice there are relatively few entries there, not sure what it is used for, interested to learn.

lib/data/monorepo.json
rarkins: We don't have a fixed rule. We need the name to be long enough that it won't be confused for other, similarly named repos

klippx: Got it, removed unnecessary double `azure-` ✅  but still kept the prefix so they are appearing together

klippx: @rarkins fixed comments

rarkins:
> @rarkins Should I add new entries to `orgGroups` ?
> 
> ```json
>   "orgGroups": {
>     ...
>     "azure": "https://github.com/Azure/",
>     "azure-ad": "https://github.com/AzureAD/",
>     ...
>   }
> ```

Only if every single package published within those orgs should be upgraded together, which sounds unlikely to me

lib/data/monorepo.json
```diff
@@ -12,9 +12,9 @@
     "algoliasearch-autocomplete": "https://github.com/algolia/autocomplete",
     "algoliasearch-client-javascript": "https://github.com/algolia/algoliasearch-client-javascript",
     "analog": "https://github.com/analogjs/analog",
-    "angular": "https://github.com/angular/angular",
```
viceice, line 15: this change seems wrong

lib/data/monorepo.json
```diff
@@ -45,8 +45,8 @@
       "https://github.com/aws-powertools/powertools-lambda-typescript"
     ],
     "aws-sdk-client-mock": "https://github.com/m-radzikowski/aws-sdk-client-mock",
-    "aws-sdk-go": "https://github.com/aws/aws-sdk-go",
```
viceice, line 48: same

lib/data/monorepo.json
```diff
@@ -56,10 +56,26 @@
...
+    "azure-ad-microsoft-authentication-library-for-dotnet": "https://github.com/AzureAD/microsoft-authentication-library-for-dotnet",
+    "azure-ad-microsoft-authentication-library-for-go": "https://github.com/AzureAD/microsoft-authentication-library-for-go",
+    "azure-ad-microsoft-authentication-library-for-java": "https://github.com/AzureAD/microsoft-authentication-library-for-java",
+    "azure-ad-microsoft-authentication-library-for-js": "https://github.com/AzureAD/microsoft-authentication-library-for-js",
+    "azure-ad-microsoft-authentication-library-for-objc": "https://github.com/AzureAD/microsoft-authentication-library-for-objc",
+    "azure-ad-microsoft-authentication-library-for-python": "https://github.com/AzureAD/microsoft-authentication-library-for-python",
     "azure-functions-dotnet-worker": "https://github.com/Azure/azure-functions-dotnet-worker",
-    "azure azure-libraries-for-net": "https://github.com/Azure/azure-libraries-for-net",
-    "azure azure-sdk-for-net": "https://github.com/Azure/azure-sdk-for-net",
-    "azure azure-storage-net": "https://github.com/Azure/azure-storage-net",
```
viceice, line 62: this is a breaking change

viceice (changes requested):

lib/data/monorepo.json
klippx: Reverted

lib/data/monorepo.json
klippx: reverted

lib/data/monorepo.json
klippx: reverted

klippx:
@viceice No longer breaking; I want a smooth merge experience

viceice (approved):

PR #33626, opened 2025-01-15
feat(platform/azure): Allow to bypass certain policies for auto-merge

janicmikes:
## Changes

This allows to use bypass functionality of Azure DevOps to provide a seamless auto-merge even if some policies are not met.

Some Policy Types one might want to bypass because they prevent auto-merge:

| UUID                                 | Policy Type Name         |
| ------------------------------------ | ------------------------ |
| fd2167ab-b0be-447a-8ec8-39368250530e | RequiredReviewers        |
| fa4e907d-c16b-4a4c-9dfa-4916e5d171ab | RequireAMergeStrategy    |
| fa4e907d-c16b-4a4c-9dfa-4906e5d171dd | MinimumNumberOfReviewers |
| 0609b952-1397-4640-95ec-e00a01b2c241 | Build                    |
| 40e92b44-2fe1-4dd6-b3d8-74a9c21d0c6e | WorkItemLinking          |

Additionally the required bypass reason can be customized if needed.

## Context

Automatic Merges for Azure DevOps might not be done (using merge) if some policies prevent the pr from being merged.
This can be bypassed by using the bypass policy option when completing the pull-request.

The thing is, that any other policy, not to be bypassed will prevent the bypass from being used.


lib/modules/platform/azure/index.spec.ts
```diff
@@ -1898,6 +1902,151 @@ describe('modules/platform/azure/index', () => {
...
+      azureApi.gitApi.mockImplementationOnce(
+        () =>
+          ({
+            getPullRequestById: jest.fn(() => ({
+              lastMergeSourceCommit: lastMergeSourceCommitMock,
+              targetRefName: 'refs/heads/ding',
+              title: 'title',
+            })),
+            updatePullRequest: updatePullRequestMock,
+          }) as any,
```
viceice, line 1925: no any, use partial helper function

viceice (commented):
@rarkins @secustor the platform specific PR options are getting worse. we should move them to a sub config object for easier pass through.

janicmikes:
> @rarkins @secustor the platform specific PR options are getting worse. we should move them to a sub config object for easier pass through.

In order to keep the options to a minimum I could limit them to the "generic" option because.

I wanted to make the coverage check to pass but didn't find the time to add the required unit tests yet. Sorry to have this stale for so long.

The other open question is probably if the platform configuration should be reorganized prior or after this.

janicmikes:
@viceice I had to migrate jest to vi as of #34475 , #34499 and #34500.

lib/config/options/index.ts
```diff
@@ -1246,6 +1246,42 @@ const options: RenovateOptions[] = [
...
+    default: [],
+    supportedPlatforms: ['azure'],
+  },
+  {
+    name: 'azureBypassPolicyReason',
+    description: 'The reason to set when bypassing policies.',
+    type: 'boolean',
+    default: false,
+    supportedPlatforms: ['azure'],
+  },
```
viceice, line 1301: @rarkins @secustor I don't like to have all those simple boolean options. Should we use a new `platformFlags` (sample name) field where you can add those flags?

viceice: We can also internally comvert them to bool and pass to exisiting `plarformOptions`

viceice (changes requested):

lib/config/options/index.ts
secustor: I think a user facing list  ( `azureBypassPolicies` ) of allowed values should be the best way here. 

Though I do not think we have this option type right now 🤔

janicmikes: I added (at least some of these policies) here: lib/modules/platform/azure/types.ts

The thing is, if you use external tools or custom policies, this list is endless.

What if we just keep the "azureBypassPolicyTypeUuids" option and add the list of known UUIDs to its documentation?

This would also clean up the ignoreTest option which is basically an alias of Build validation.

janicmikes: I refactored the individual options into a value restricted string array.

For the sake of simplicity I opted out from allowing arbitrary UUIDs, and instead allow a Named version for the known types.

Those Names match the official policy type names from Azure DevOps and should make maintaining the config more readable.

lib/modules/platform/azure/types.ts
```diff
@@ -11,3 +11,23 @@ export const AzurePrVote = {
...
   Approved: 10,
 } as const;
+
+export const AzurePolicyTypes = {
+  RequiredReviewers: 'fd2167ab-b0be-447a-8ec8-39368250530e',
+  RequireAMergeStrategy: 'fa4e907d-c16b-4a4c-9dfa-4916e5d171ab',
+  MinimumNumberOfReviewers: 'fa4e907d-c16b-4a4c-9dfa-4906e5d171dd',
+  Build: '0609b952-1397-4640-95ec-e00a01b2c241',
+  WorkItemLinking: '40e92b44-2fe1-4dd6-b3d8-74a9c21d0c6e',
+} as const;
```
viceice: we usually store those constants in constants when they're use outside of the platform, so we don't need to deep import platform stuff

https://github.com/renovatebot/renovate/blob/db9eef8dc62e30a48afb769f74c55ea370d39455/lib/constants/platforms.ts#L1

lib/util/json-writer/__fixtures__/.global_editorconfig
```diff
@@ -1,3 +1,5 @@
+root = true
```
viceice: revert unrelated changes

lib/util/json-writer/__fixtures__/.json_editorconfig
```diff
@@ -1,3 +1,5 @@
+root = true
```
viceice: same

lib/util/json-writer/__fixtures__/.non_json_editorconfig
```diff
@@ -1,3 +1,5 @@
+root = true
```
viceice: same

lib/util/json-writer/editor-config.spec.ts
```diff
@@ -29,7 +29,7 @@ describe('util/json-writer/editor-config', () => {
   });
 
   it('should handle empty .editorconfig file', async () => {
-    await fs.writeFile(`${tmpDir!.path}/.editorconfig`, '');
+    await fs.writeFile(`${tmpDir!.path}/.editorconfig`, 'root = true');
```
viceice: why? please do those stuff in a separate pr

lib/workers/repository/onboarding/branch/config.spec.ts
```diff
@@ -34,7 +34,7 @@ describe('workers/repository/onboarding/branch/config', () => {
       mockedPresets.getPreset.mockResolvedValueOnce({ enabled: true });
       const contents = await getOnboardingConfigContents(config, '');
       expect(mockedPresets.getPreset).toHaveBeenCalledTimes(1);
-      expect(contents).toEqual(
+      expect(contents).toEqualIgnoringWhitespace(
```
viceice: same, unrelated

lib/workers/repository/update/pr/automerge.ts
```diff
@@ -11,6 +11,7 @@ import { scm } from '../../../../modules/platform/scm';
 import type { BranchConfig } from '../../../types';
 import { isScheduledNow } from '../branch/schedule';
 import { resolveBranchStatus } from '../branch/status-checks';
+import { getPlatformPrOptions } from './index';
```
viceice:
```suggestion
import { getPlatformPrOptions } from '.';
```

lib/workers/repository/update/pr/index.ts
```diff
@@ -62,6 +62,8 @@ export function getPlatformPrOptions(
     bbUseDefaultReviewers: !!config.bbUseDefaultReviewers,
     gitLabIgnoreApprovals: !!config.gitLabIgnoreApprovals,
     forkModeDisallowMaintainerEdits: !!config.forkModeDisallowMaintainerEdits,
+    azureBypassPolicyTypes: config.azureBypassPolicyTypes ?? [],
```
viceice:
```suggestion
    azureBypassPolicyTypes: config.azureBypassPolicyTypes,
```

they're optional on pr options too

lib/util/json-writer/editor-config.spec.ts
janicmikes: I ran the tests locally and run into these "issues". 
Basically if there is an ".editorconfig" in any of the parent directories where this repository was cloned, these parent settings will apply. The root = true prevents searching the parent directories. (I will revert and open another issue, PR)

lib/modules/platform/azure/types.ts
```diff
@@ -11,3 +12,8 @@ export const AzurePrVote = {
   ApprovedWithSuggestions: 5,
   Approved: 10,
 } as const;
+
+export const getAzurePolicyTypes = (): string[] => [
```
viceice: also move this to the contants

viceice: or inline to config options because it's only used once there

janicmikes: I put them inline in the config options.

lib/constants/platforms.ts
```diff
@@ -58,3 +58,18 @@ export const BITBUCKET_SERVER_API_USING_HOST_TYPES = [
   'bitbucket-server-changelog',
   'bitbucket-server-tags',
 ];
+
+export const AZURE_POLICY_TYPES = {
```
RahulGautamSingh: Please add a reference link here:

lib/modules/platform/azure/index.spec.ts
RahulGautamSingh, line 1: Checking my understanding.

We wait to see if the policy status is still being processed and then regardless of result bypass it if the policy is present in `azureBypassPolicyTypes`?

I think so cause in the two tests you added:

1. policy status is `rejected` & it is bypassed
2. policy status is `queued` & it is not bypassed 

If we are going to bypass the policy regardless of result when the user has configured it in the `azureBypassPolicyTypes`, why different behaviour for `queued`

janicmikes: Let me check on the tests again. Since there is a lot of mocking going on lets verify its close to a real situation.

It should only use the bypass if the policy blocks the pr from being merged, this could be any state but succeeded.

In general if it could become green on its own we should rather wait for this to happen I would say.

The initial driver for this are mainly the two policies related with workItem linking and required Reviewers.

Both can be worked around with their own downsides. (Arbitrary WorkItem, a second service account to approve)

Bypassing is more like the "sudo" way of doing it, the downside is that renovate will require and effectively have more privileges on your repo.

lib/constants/platforms.ts
janicmikes: Thats actually a funny thing, at first I thought these would be random since they at least look like uuid v4 values.

Some appear in the docs: https://learn.microsoft.com/en-us/rest/api/azure/devops/policy/configurations/list?view=azure-devops-rest-7.1&tabs=HTTP

Others are off by just one character which proves their not-so randomness.

In the end the truth lies in the response of this endpoint: https://learn.microsoft.com/en-us/rest/api/azure/devops/policy/types/list?view=azure-devops-rest-5.0


If you follow the discussion I was almost going for a string only option, and leave a hint to those in the docs (as examples) but in the end I think this is much more user friendly if you know what you bypass when reading your config.

secustor: Can we than document the most convenient way to retrieve this?

lib/modules/platform/azure/index.ts
```diff
@@ -784,23 +786,61 @@ export async function mergePr({
...
+
+    // only use bypass if all required policies are in approved state
+    bypassPolicy = policyEvaluations
+      .filter(
+        (policy) =>
+          policy.configuration?.isEnabled &&
+          policy.configuration.isBlocking &&
+          !bypassPolicyTypes.has(policy.configuration.type?.id ?? ''),
+      )
+      .every((policy) => policy.status === PolicyEvaluationStatus.Approved);
```
secustor: use for ... of

lib/modules/platform/azure/index.ts
```diff
@@ -784,23 +786,61 @@ export async function mergePr({
...
+  const azurePolicyApi = await azureApi.policyApi();
 
   let pr = await azureApiGit.getPullRequestById(pullRequestId, config.project);
 
   const mergeStrategy =
     strategy === 'auto'
       ? await getMergeStrategy(pr.targetRefName!)
       : mapMergeStrategy(strategy);
+
+  let bypassPolicy: boolean | undefined;
```
secustor:
```suggestion
  let hasBypassPolicy: boolean  = false;
```

lib/modules/platform/azure/index.ts
```diff
@@ -812,7 +852,7 @@ export async function mergePr({
       pr.lastMergeSourceCommit?.commitId
     } using mergeStrategy ${mergeStrategy} (${
       GitPullRequestMergeStrategy[mergeStrategy]
-    })`,
+    })${bypassPolicy ? ' and bypassPolicies' : ''}`,
```
secustor: Log the bypass policies as a struct field.

docs/usage/self-hosted-configuration.md
```diff
@@ -219,6 +219,36 @@ For example:
...
+
+By default the reason for bypassing policies is set to "Auto-merge by Renovate".
+
+Define a custom reason when policies are being bypassed to perform auto-merge.
+
+## azureBypassPolicyTypes
+
+Add a list of policy UUIDs which might be failing but should not block auto-merge.
+
+> For this to work the account that executes renovate must be allowed the `Bypass policies when completing pull requests` permission on the respective repository.
```
RahulGautamSingh: Lets convert it to a !!!note so it's more visible and move

lib/modules/platform/azure/index.spec.ts
```diff
@@ -1904,6 +1909,254 @@ describe('modules/platform/azure/index', () => {
...
+            bypassReason: 'Auto-merge by Renovate',
+          },
+        },
+        '1',
+        pullRequestIdMock,
+      );
+      expect(res).toBeTrue();
+    });
+
+    it('should complete pr but not bypass policies if all policies are approved, not applicable, disabled or not blocking', async () => {
```
RahulGautamSingh: Please re-phrase  description. It's hard to read/understand,

lib/modules/platform/azure/index.spec.ts
```diff
@@ -1904,6 +1909,254 @@ describe('modules/platform/azure/index', () => {
...
+            mergeCommitMessage: 'title',
+          },
+        },
+        '1',
+        pullRequestIdMock,
+      );
+      expect(res).toBeTrue();
+    });
+
+    it('should not complete pr should not bypass policies if not all non approved policies are bypassed', async () => {
```
RahulGautamSingh: Please re-phrase description. It's hard to read/understand,

lib/modules/platform/azure/index.ts
```diff
@@ -784,35 +787,87 @@ export async function mergePr({
...
+        continue;
+      }
+
+      if (bypassPolicyTypes.has(policyEvaluation.configuration.type.id)) {
+        bypassPolicies.push(policyEvaluation.configuration);
+        bypassCompletionOptions = {
+          bypassPolicy: true,
+          bypassReason:
+            platformOptions?.azureBypassPolicyReason ??
+            'Auto-merge by Renovate',
```
RahulGautamSingh: Default values are automatically populated.
```suggestion
```

lib/modules/platform/azure/index.ts
```diff
@@ -784,35 +787,87 @@ export async function mergePr({
...
+    },
     `Updating PR ${pullRequestId} to status ${PullRequestStatus.Completed} (${
       PullRequestStatus[PullRequestStatus.Completed]
     }) with lastMergeSourceCommit ${
       // TODO: types (#22198)
       pr.lastMergeSourceCommit?.commitId
     } using mergeStrategy ${mergeStrategy} (${
       GitPullRequestMergeStrategy[mergeStrategy]
-    })`,
+    })${bypassCompletionOptions.bypassPolicy ? ' and bypassPolicies' : ''}`,
```
RahulGautamSingh: Not sure but maybe works.
```suggestion
    })${bypassCompletionOptions.bypassPolicy &&  ' and bypassPolicies'}`,
```

RahulGautamSingh (changes requested):

lib/modules/platform/azure/index.ts
janicmikes: I added the number of bypassed policies to the trace-logs.

docs/usage/self-hosted-configuration.md
janicmikes: Where do you want this to be moved?

lib/modules/platform/azure/index.ts
```diff
@@ -867,7 +865,7 @@ export async function mergePr({
       pr.lastMergeSourceCommit?.commitId
     } using mergeStrategy ${mergeStrategy} (${
       GitPullRequestMergeStrategy[mergeStrategy]
-    })${bypassCompletionOptions.bypassPolicy ? ' and bypassPolicies' : ''}`,
+    }) and bypassed ${bypassPolicies.length} policies.`,
```
RahulGautamSingh: ? typo

janicmikes: You mean "Bypassed"?

janicmikes: I changed it to "and bypassing {n} policies"

RahulGautamSingh: My bad, I misread.

RahulGautamSingh (commented):
LGTM, did you confirm this by testing manually?

janicmikes:
> LGTM, did you confirm this by testing manually?

Give me a couple of days, I will test with own self hosted repository to confirm.

The logic (as it was added) is currently running in a separate script prior to renovate (because otherwise the running build would always block merging)

Depending on how one does configure the grouping (branch) it is crucial to have the schedule properly set up for auto merge to work properly.

I would generally never allow anything not yet passed the integration tests to me merged.

RahulGautamSingh:
@janicmikes Closing for now. Feel free to re-open when you get time, for it.

PR #33648, opened 2025-01-16
feat: Switch to `Timestamp` type

### LINKED ISSUE #31891: Strongly type timestamps

zharinov:
## Context

- Closes: #31891


lib/modules/datasource/custom/schema.ts
```diff
@@ -1,12 +1,13 @@
...
+import { TimestampSchema } from '../../../util/timestamp';
 
 export const ReleaseResultZodSchema = z.object({
   releases: z.array(
     z
       .object({
         version: z.string(),
         isDeprecated: z.boolean().optional(),
-        releaseTimestamp: z.string().optional(),
+        releaseTimestamp: TimestampSchema.nullable().catch(null),
```
viceice, line 10: shouldn't we create a shortcut for this? So we've less duplicate code?

```ts
export type TimestampSchemaSafe = TimestampSchema.nullable().catch(null);
export type NullableTimestampSchema = TimestampSchema.nullable().catch(null);
```

or even make `TimestampSchema` nullable and catch to null?

zharinov: Yes, I think just make it nullable is reasonable and is consistent with `asTimetamp` return type

lib/modules/datasource/go/releases-direct.spec.ts
```diff
@@ -129,19 +129,19 @@ describe('modules/datasource/go/releases-direct', () => {
           {
             gitRef: 'v0.1.0',
             newDigest: 'd73d815ec22c421e7192a414594ac798c73c89e5',
-            releaseTimestamp: '2022-05-15T16:29:42Z',
+            releaseTimestamp: '2022-05-15T16:29:42.000Z',
```
viceice, line 132: why this change? 🤔

zharinov: We're normalizing everything to the milliseconds now

viceice (approved):

PR #33672, opened 2025-01-17
fix: remove colons from depNameSanitized

rarkins:

### DIFF
--- a/lib/workers/repository/updates/flatten.ts
+++ b/lib/workers/repository/updates/flatten.ts
@@ -17,12 +17,13 @@ import { generateBranchName } from './branch-name';
 const upper = (str: string): string =>
   str.charAt(0).toUpperCase() + str.substring(1);
 
-function sanitizeDepName(depName: string): string {
+export function sanitizeDepName(depName: string): string {
   return depName
     .replace('@types/', '')
     .replace('@', '')
     .replace(regEx(/\//g), '-')
     .replace(regEx(/\s+/g), '-')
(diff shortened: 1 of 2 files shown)


lib/workers/repository/updates/flatten.ts
```diff
@@ -17,12 +17,13 @@ import { generateBranchName } from './branch-name';
...
   str.charAt(0).toUpperCase() + str.substring(1);
 
-function sanitizeDepName(depName: string): string {
+export function sanitizeDepName(depName: string): string {
   return depName
     .replace('@types/', '')
     .replace('@', '')
     .replace(regEx(/\//g), '-')
     .replace(regEx(/\s+/g), '-')
+    .replace(regEx(/:/g), '-')
```
reitzig, line 26: 💡 Maybe something like `/[^-.a-z0-9]+/g` would be neater?

rarkins: It's much more risky. For example, I'm guessing you left off A-Z by accident? I'd prefer in this case to work by excluding/replacing instead of including

viceice (approved):
this can close some exising pr's

lib/workers/repository/updates/flatten.ts
reitzig: Yes and no. Adding bad chars one by one is awkward, but messing up valid strings is bad as well. One would basically want to match what Git does, right? Didn't check.

Anyway, your call. Just thought I'd mention it.

PR #33673, opened 2025-01-17
feat(nuget): respect `global.json` for artifact updates

### LINKED ISSUE #33644: dotnet: check global.json for sdk version
### Discussed in https://github.com/renovatebot/renovate/discussions/33636

<sup>Originally posted by **konradmalik** January 16, 2025</sup>

The scenario is as follows:
- a new major dotnet version comes out (like 9.0.102)
- our projects still use dotnet 8 (like 8.0.405)
- our dotnet version is defined in a global.json file like below:
```json
{
    "sdk": {
        "version": "8.0.405",
        "rollForward": "latestMinor",
        "allowPrerelease": false
    }
}
```
- renovate creates an MR with package update, (like `chore(deps): update dependency coverlet.collector to 6.0.3`)
- our pipeline for MR fails, because renovate could not update `packages.lock.json` files after the generated update (we use lockfiles and restore fails)
- after a quick check, renovate failed to do update those files because it's using `dotnet 9.0.102` instead of `8.0.405` defined in the global.json, see logs below (renovate's comment on MR)

Solution: As part of Artifacts updating, Renovate should look for a `global.json` in the repo (starting in project directory and looking "up"), and then use the sdk.version value if present to pass to the tool installation

viceice:
## Changes
Respect dotnet sdk version specified in `global.json` if not set via renovate constraints.
Use new zod schema for extraction too.

## Context
- closes #33644

### DIFF
--- a/lib/modules/manager/nuget/artifacts.ts
+++ b/lib/modules/manager/nuget/artifacts.ts
@@ -25,7 +25,11 @@ import {
   NUGET_CENTRAL_FILE,
   getDependentPackageFiles,
 } from './package-tree';
-import { getConfiguredRegistries, getDefaultRegistries } from './util';
+import {
+  findGlobalJson,
+  getConfiguredRegistries,
+  getDefaultRegistries,
+} from './util';
 
 async function createCachedNuGetConfigFile(
   nugetCacheDir: string,
@@ -55,6 +59,9 @@ async function runDotnetRestore(
     packageFileName,
   );
 
+  const dotnetVersion =
+    config.constraints?.dotnet ??
+    (await findGlobalJson(packageFileName))?.sdk?.version;
   const execOptions: ExecOptions = {
     docker: {},
     userConfiguredEnv: config.env,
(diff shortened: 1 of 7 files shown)


lib/modules/manager/nuget/artifacts.ts
```diff
@@ -55,16 +59,17 @@ async function runDotnetRestore(
     packageFileName,
   );
 
+  const dotnetVersion =
+    config.constraints?.dotnet ??
+    (await findGlobalJson(packageFileName))?.sdk?.version;
```
rarkins, line 64: The schema only enforces that it's a string. Should we worry about there being some value like `"foo"` or `"lts"` or even just `"6"` which isn't a valid constraint?

viceice: I think the docs only mention a full version, so I think everything else is an error and should not happen 🤷‍♂️

rarkins (approved):

PR #33680, opened 2025-01-17
feat(datasource/github-runners): Add `ubuntu-22.04-arm` and `ubuntu-24.04-arm` support

edgarrmondragon:
## Changes

Adds support for public preview GitHub Runners `ubuntu-22.04-arm` and `ubuntu-24.04-arm`.

## Context

- https://github.com/community/maintainers/discussions/480
- https://github.com/renovatebot/renovate/discussions/33656.

### DIFF
--- a/lib/modules/datasource/github-runners/index.ts
+++ b/lib/modules/datasource/github-runners/index.ts
@@ -23,6 +23,10 @@ export class GithubRunnersDatasource extends Datasource {
       { version: '18.04', isDeprecated: true },
       { version: '16.04', isDeprecated: true },
     ],
+    ubuntu_arm64: [
+      { version: '24.04', isStable: false },
+      { version: '22.04', isStable: false },
+    ],
     macos: [
       { version: '15', isStable: false },
       { version: '15-large', isStable: false },
@@ -46,6 +50,13 @@ export class GithubRunnersDatasource extends Datasource {
     ],
   };
 
+  private static readonly supportsLatest: Record<string, boolean> = {
+    ubuntu: true,
+    ubuntu_arm64: false,
+    macos: true,
+    windows: true,
+  };
+
   public static isValidRunner(
     runnerName: string,
     runnerVersion: string,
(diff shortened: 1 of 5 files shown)


lib/modules/datasource/github-runners/index.ts
```diff
@@ -46,6 +50,13 @@ export class GithubRunnersDatasource extends Datasource {
     ],
   };
 
+  private static readonly supportsLatest: Record<string, boolean> = {
+    ubuntu: true,
+    ubuntu_arm64: false,
```
edgarrmondragon, line 55: https://github.com/community/maintainers/discussions/480#discussioncomment-11860123

lib/modules/datasource/github-runners/index.ts
```diff
@@ -23,6 +23,10 @@ export class GithubRunnersDatasource extends Datasource {
       { version: '18.04', isDeprecated: true },
       { version: '16.04', isDeprecated: true },
     ],
+    ubuntu_arm64: [
+      { version: '24.04', isStable: false },
+      { version: '22.04', isStable: false },
```
rarkins, line 28: Are you marking them as unstable because GitHub has them in "public preview"? I would lean towards treating them both as stable seeing as ubuntu is stable

edgarrmondragon: That's correct, though I don't have a strong opinion so I'm happy to mark them as stable 🙂

edgarrmondragon: 226d36b73

viceice (commented):
I would prefer to simply add them as additional versions to the datasource and use the suffix as version compatibility.

@rarkins @secustor wdyt?

edgarrmondragon:
> I would prefer to simply add them as additional versions to the datasource and use the suffix as version compatibility.
> 
> @rarkins @secustor wdyt?

Thanks for the feedback @viceice! I'm getting my feet wet in the codebase, I see the github-runners datasource uses `docker` as the default versioning scheme. Does that mean that `ubuntu-24.04` and `ubuntu-24.04-arm` will be deemed incompatible (and thus never updated from one to the other) similar to Docker images `python:3.13` and `python:3.13-slim`?

viceice:
yes

PR #33718, opened 2025-01-21
docs: use issueType field for docs generation

### LINKED ISSUE #33691: Update docs generation to use native "type" field

Our docs generation today uses "type:bug" and "type:feature" labels, but we need to migrate these to support the "native" Task field in GitHub issues.

Currently all Issues have both a "type" label and Type category, so a full switch over can be done

RahulGautamSingh:
## Changes
- Use `gh` command to fetch issues based on `issueType` 

## Context
- Closes: #33691 

- [x] No unit tests but ran on a real repository, or (generated the docs locally)


lib/modules/platform/github/graphql.ts
```diff
@@ -85,3 +85,36 @@ mutation EnablePullRequestAutoMerge(
   }
 }
 `;
+
+export const getIssuesByIssueTypeQuery = `
```
rarkins, line 89: Not sure we want this in this file, as it's not part of the application itself. Please move to under tools/docs

RahulGautamSingh:
WIP

openitems.json
rarkins, line 1: Was this meant to be added?

lib/modules/platform/github/graphql.ts
rarkins: Could we use the `gh` command, or some more native github library, instead of all this graphql code?

rarkins (changes requested):

openitems.json
RahulGautamSingh: No

tools/docs/github-query-items.ts
```diff
@@ -46,6 +40,39 @@ export interface Items {
   features: ItemsEntity[];
 }
 
+async function getIssuesByIssueType(
+  issueType: 'Bug' | 'Feature',
+): Promise<ItemsEntity[]> {
+  const queryString = `type:${issueType}, repo:renovatebot/renovate, state:open`;
+  const res = (await githubApi.querySearchField<unknown>(
```
viceice, line 47: we need to skip it if no token provided because graphql needs authentication!

viceice (changes requested):

lib/modules/platform/github/graphql.ts
RahulGautamSingh: We can use the `gh` command. Will need to parse the string o/p for required info.

rarkins (dismissed):

tools/docs/github-query-items.ts
```diff
@@ -46,6 +28,25 @@ export interface Items {
...
+async function getIssuesByIssueType(
+  issueType: 'Bug' | 'Feature',
+): Promise<ItemsEntity[]> {
+  const command = `gh issue list --json "title,number,url,labels" --search "type:${issueType}" --limit 1000`;
+  const res = await exec(command);
+  const issues = JSON.parse(res.stdout) as GhIssueOutput[];
+
+  return issues.map((issue) => {
+    return { ...issue, html_url: issue.url, issueType };
+  });
```
viceice, line 47: drop the mapping and fix code generator to use url field

tools/docs/github-query-items.ts
```diff
@@ -46,6 +28,25 @@ export interface Items {
...
+  url: string;
+  labels: { name: string; id: string; description: string }[];
+}
+
+async function getIssuesByIssueType(
+  issueType: 'Bug' | 'Feature',
+): Promise<ItemsEntity[]> {
+  const command = `gh issue list --json "title,number,url,labels" --search "type:${issueType}" --limit 1000`;
+  const res = await exec(command);
+  const issues = JSON.parse(res.stdout) as GhIssueOutput[];
```
viceice, line 43: we should use zod schema to catch changes

tools/docs/github-query-items.ts
```diff
@@ -59,18 +60,17 @@ export async function getOpenGitHubItems(): Promise<RenovateOpenItems> {
...
   }
 
-  const q = `repo:renovatebot/renovate type:issue is:open`;
-  const per_page = 100;
+  if (!process.env.GITHUB_TOKEN) {
+    logger.warn(
+      'No GITHUB_TOKEN found in env, cannot fetch Github issues. Skipping...',
+    );
+    return result;
+  }
```
viceice, line 68: there is an existing skip check for CI, please move this to same place

tools/docs/github-query-items.ts
```diff
@@ -59,18 +60,17 @@ export async function getOpenGitHubItems(): Promise<RenovateOpenItems> {
     return result;
   }
 
-  const q = `repo:renovatebot/renovate type:issue is:open`;
-  const per_page = 100;
+  if (!process.env.GITHUB_TOKEN) {
+    logger.warn(
```
viceice, line 64: throw if `process.env.CI` so we don't build without on main

rarkins: Is this a new logic, or did it get dropped during this change somehow?

tools/docs/github-query-items.ts
```diff
@@ -46,6 +29,34 @@ export interface Items {
...
+      }),
+    ),
+    number: z.number(),
+  }),
+);
+
+async function getIssuesByIssueType(
+  issueType: 'Bug' | 'Feature',
+): Promise<ItemsEntity[]> {
+  const command = `gh issue list --json "title,number,url,labels" --search "type:${issueType}" --limit 1000`;
```
viceice, line 48: can we search for open issues only here?

tools/docs/github-query-items.ts
```diff
@@ -46,6 +29,34 @@ export interface Items {
...
+  const execRes = await exec(command);
+  const res = GhOutputSchema.safeParse(JSON.parse(execRes.stdout));
+  if (res.error) {
+    throw res.error;
+  }
+
+  return res.data.map((issue) => {
+    return { ...issue, issueType };
```
viceice, line 56: sad that it's not a property on result type 😔

tools/docs/github-query-items.ts
```diff
@@ -59,18 +70,21 @@ export async function getOpenGitHubItems(): Promise<RenovateOpenItems> {
...
+
   try {
-    const query = getQueryString({ q, per_page });
-    const res = await githubApi.getJsonUnchecked<GithubApiQueryResponse>(
-      gitHubApiUrl + query,
-      {
-        paginationField: 'items',
-        paginate: true,
-      },
+    const rawItems: ItemsEntity[] = (await getIssuesByIssueType('Bug')).concat(
```
viceice, line 85:
```suggestion
    const rawItems = (await getIssuesByIssueType('Bug')).concat(
```
type should not needed here

tools/docs/github-query-items.ts
RahulGautamSingh: that is the default behavior

tools/docs/github-query-items.ts
RahulGautamSingh: Yes, they are still working on the APIs

tools/docs/github-query-items.ts
viceice, line 48: why the limit to 1000? we didn't had it before 🤔

can we change page sizes to 100?

RahulGautamSingh: I made the limit 1000 cause we have 600+ issues currently, so we can get the json all the issues in one command execution instead of making 7 executions.

viceice, line 48: one bad side effect of this change is that a developer needs `gh` cli with proper token to build docs locally 🤔 

this needs at least documented!

RahulGautamSingh: Yes, I will add.

viceice (dismissed):

docs/development/local-development.md
```diff
@@ -181,6 +181,8 @@ You can run `pnpm build:docs` to generate the docs.
 Then use `pnpm mkdocs serve` to preview the documentation locally.
 The docs will update automatically when you run `pnpm build:docs` again, no need to stop the `pnpm mkdocs serve` command.
 
+Also, make sure to set the `GITHUB_TOKEN` in your environment as we use `gh` to fetch the issues. If you wish to skip fetching the issues, then set `SKIP_GITHUB_ISSUES` to `true` before building the docs.
```
rarkins, line 184:
```suggestion
Also, make sure to set the `GITHUB_TOKEN` in your environment as we use `gh` to fetch the issues.
If you wish to skip fetching the issues, then set `SKIP_GITHUB_ISSUES` to `true` before building the docs.
```

rarkins (dismissed):

docs/development/local-development.md
```diff
@@ -181,6 +181,9 @@ You can run `pnpm build:docs` to generate the docs.
 Then use `pnpm mkdocs serve` to preview the documentation locally.
 The docs will update automatically when you run `pnpm build:docs` again, no need to stop the `pnpm mkdocs serve` command.
 
+Also, make sure to set the `GITHUB_TOKEN` in your environment as we use `gh` to fetch the issues.
```
viceice, line 184:
```suggestion
Also, make sure to set the `GITHUB_TOKEN` in your environment as we use `gh` cli to fetch the issues.
```
should we add a link to gh cli docs?

viceice (commented):
otherwise LGTM

docs/development/local-development.md
rarkins: Sure

viceice (approved):

PR #33732, opened 2025-01-21
feat(azure): add support for rebase-merge

basilfx:
## Changes

Azure DevOps has support for semi-linear rebase (rebase with merge commit, see [1]). Our branch policies enforce this option, which is incompatible with any of the `automergeStrategy` options currently available.

In fact, when this option is set to auto, Renovate Bot will actually select semi-linear rebasing if that is the only option. But if there are other (allowed) options, then there is nothing to enforce it.

This adds the `rebase-merge` option, which is different from using `rebase` (again, see [1]).

Based on closer inspection, I see that Gitea will do a rebase with merge when the option is `rebase`. I did not adapt this, because that would break current configurations. Also, I only have the option to test this with Azure DevOps.

[1]: https://devblogs.microsoft.com/devops/pull-requests-with-rebase/

## Context

I could not get automerge to work, until I found out that the `rebase` option does not map to semi-linear rebase within Azure DevOps. This silently fails when a PR gets created.

Given that rebase-merge is rebase + merge combined, this option makes sense.


docs/usage/configuration-options.md
```diff
@@ -263,6 +263,7 @@ You may choose from these values:
 - `fast-forward`, "fast-forwarding" the main branch reference, no new commits in the resultant tree
 - `merge-commit`, create a new merge commit
 - `rebase`, rewrite history as part of the merge, but usually keep the individual commits
+- `rebase-merge`, create a new merge commit, but rebase the commits prior merging
```
rarkins, line 266: Note here that it's Azure-only

rarkins (changes requested):

viceice (changes requested):
we need to check gitea and bitbucket. at least gitea also supports both strategies too.

basilfx:
@viceice Thank you for the review. Happy to make the requested changes, but any advice on how to deal with Gitea?

Like mentioned in the PR and commit, I saw that Gitea will use rebase-merge if the `rebase` option is selected. I could change that, but that would be a breaking change. 

So technically, it is not an Azure DevOps specific feature, as requested.

docs/usage/configuration-options.md
rarkins, line 266:
```suggestion
- `rebase-merge`, create a new merge commit, but rebase the commits prior merging (azure-only)
```

rarkins (approved):

rarkins:
@viceice I would like to merge this as-is

viceice (approved):

PR #33756, opened 2025-01-22
test(npm): use fs test utils in extract/pnpm.spec.ts

fpapado:
## Changes

This PR is split off from https://github.com/renovatebot/renovate/pull/33376, specifically the discussion at https://github.com/renovatebot/renovate/pull/33376#discussion_r1920385993. It changes tests in lib/modules/manager/npm/extract/pnpm.spec.ts to use the shared `fs` test utilities. cc @secustor ✉️ 

The `fs` test utilities are used by most test suites. This suite was an outlier, instead using `spyOn` as a way to implement partial mocks, letting other calls of the `fs` module fall through to the real implementation.

The test utilities assume that the whole `fs` module is mocked, so I made relevant changes to ensure that. A couple of tests were incidentally relying on the "real" implementation (primarily for path manipulation). For those, I explicitly passed the real version of the relevant functions to them. I think this still has some of the pitfalls of partial mocking, but it is more explicit about what is going on. The blast radius is also isolated to those couple of tests that are relying on this behaviour, instead of the whole suite.

## Context

Making these tests more consistent with other test suites, and paving the way for using the `fs` test utilities in https://github.com/renovatebot/renovate/pull/33376.

### DIFF
--- a/lib/modules/manager/npm/extract/pnpm.spec.ts
+++ b/lib/modules/manager/npm/extract/pnpm.spec.ts
@@ -97,6 +97,10 @@ describe('modules/manager/npm/extract/pnpm', () => {
         realFs.findLocalSiblingOrParent,
       );
       fs.getSiblingFileName.mockImplementation(realFs.getSiblingFileName);
+
+      // Falls through to reading from the fixture path defined in GlobalConfig
+      // at the top of this file
+      fs.readLocalFile.mockImplementation(realFs.readLocalFile);
     });
 
     it('uses pnpm workspaces', async () => {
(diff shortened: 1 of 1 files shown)


secustor (commented):
This change is reducing the coverage. 
I have retried the pipeline just to be sure it is not a hiccup

fpapado:
> This change is reducing the coverage. I have retried the pipeline just to be sure it is not a hiccup

Oh, you are right, I remember running into this issue in the other branch. I made a similar fix now in 4ccb0cf66b6b2740114a66c7d1b247939b71f2ef. The gist is that the same tests were relying on reading the workspace file from the fixtures directory. While the tests do not fail if they do not read the file, they do not get to cover the same validations that they would have otherwise.

secustor (approved):

PR #33768, opened 2025-01-22
feat(cloudbuild): Parse config using Zod schema

zharinov:


viceice (approved):

lib/modules/manager/cloudbuild/schema.ts
```diff
@@ -0,0 +1,13 @@
+import { z } from 'zod';
+import { LooseArray, Yaml } from '../../../util/schema-utils';
+import { getDep } from '../dockerfile/extract';
+
+export const CloudbuildSteps = Yaml.pipe(
+  z
+    .object({
+      steps: LooseArray(
+        z.object({ name: z.string() }).transform(({ name }) => getDep(name)),
```
viceice, line 9: how would we pass `registryAliases` to `getDep` now?

zharinov: It wasn't used in the original code, so we don't need it here. But if we do, the modification is straightforward: remove transform and do it in the `.map` on the final result.

viceice (changes requested):
see comment

viceice (approved):

PR #33800, opened 2025-01-23
fix(manager/custom): Support range strategy with custom managers

mpkorstanje:
## Changes

Ensures `getRangeStrategy` does not return null when a custom manager is used. Thus enabling the use of range strategies with custom managers. 

Note: custom manger do not support setting a range strategy, but the range strategy can be set through package rules for a specific manager. For example:

```json
{
  "description": "Parse Ruby Gemspec files",
  "extends": [
    "config:recommended"
  ],
  "customManagers": [
    {
      "customType": "regex",
      "fileMatch": [
        "^.*\\.gemspec"
      ],
      "matchStrings": [
        ".*\\.add_(:?(?<depType>.*?)_)?dependency\\s*(['\"])(?<depName>[^'\"]+)(['\"])(\\s*,\\s*(?<currentValue>(['\"])[^'\"]+['\"](\\s*,\\s*['\"][^'\"]+['\"])?))?"
      ],
      "datasourceTemplate": "rubygems",
      "versioningTemplate": "ruby"
    }
  ],
  "packageRules": [
    {
      "description": "Applies \":preserveSemverRanges\" to the custom above manager. See: https://github.com/renovatebot/renovate/discussions/28090",
      "matchDatasources": [
        "rubygems"
      ],
      "matchManagers": [
        "custom.regex"
      ],
      "rangeStrategy": "replace"
    }
  ]
}
```

## Context

- Fixes: https://github.com/renovatebot/renovate/discussions/28090


lib/modules/manager/index.ts
```diff
@@ -78,10 +78,13 @@ export function extractPackageFile(
 
 export function getRangeStrategy(config: RangeConfig): RangeStrategy | null {
   const { manager, rangeStrategy } = config;
-  if (!manager || !managers.has(manager)) {
+  if (!manager) {
+    return null;
+  }
+  const m = managers.get(manager)! ?? customManagers.get(manager)!;
```
viceice, line 84:
```suggestion
  const m = managers.get(manager) ?? customManagers.get(manager);
```
non-null cast not needed here, you've a null check below.

viceice (commented):
otherwise LGTM

ysangkok:
It sounds good that getRangeStrategy would be prevented from returning `null`. But it seems from the PR that there is nothing preventing a manager from *returning* `auto`, which, if I understand correctly, would then reach the versioning code. This isn't supposed to happen: https://github.com/renovatebot/renovate/pull/33142#discussion_r1907301682.

If Renovate has custom managers *returning* auto, and custom managers then only work for versioning that *accepts* `auto`, custom managers would, of course, only work for versioning that actually implements `auto`, which isn't all of them (I know PVP doesn't implement it). So to make it work for all, we then have to *duplicate* the conversion logic (from `auto` into the actual non-auto strategy) in versioning *and* in managers (which usually decide the range strategy). This isn't ideal.

So, what are the ways to effectively avoid requiring all versioning modules to implement `auto` conversion logic (in addition to having it in managers)?

- This PR seems like a good opportunity to ensure that a custom manager never fails to commit to non-auto versioning strategy, like all other managers do. There could be an error in the case that no non-auto range strategy is provided by the manager. This is what I propose is added to this PR.
- Another option would be to remove `getRangeStrategy` from managers entirely, but that seems like a much larger chance. The logic would only be in versioning, and this PR could stay unchanged.
- Maybe I have misunderstood something, and this PR does prevent `auto` from reaching versioning somehow? I'd like to hear how.

CCing @rarkins since he also seems to think that `auto` should never be sent to versioning's `getNewValue` (in the thread linked above)

lib/modules/manager/index.spec.ts
```diff
@@ -180,6 +180,25 @@ describe('modules/manager/index', () => {
...
+        supportedDatasources: [],
+      });
+      expect(manager.getRangeStrategy({ rangeStrategy: 'auto' })).toBeNull();
+    });
+
+    it('handles custom managers', () => {
+      customManager.getCustomManagers().set('dummy', {
+        defaultConfig: {},
+        supportedDatasources: [],
+        getRangeStrategy: () => 'bump',
```
ysangkok, line 195: Custom managers are configured using JSON, right? And I don't see a way to set the range strategy [in the docs for regex managers](https://docs.renovatebot.com/modules/manager/regex/). So how would a situation like this ever actually arise? If it's not triggerable by the user, it might make sense to note that with a comment?

RahulGautamSingh:
I think then ,we need to implement the `getRangeStrategy` function for the `bundler` manager and move the corresponding logic out of `ruby` versioning.

lib/modules/manager/index.spec.ts
RahulGautamSingh: packageRules

rarkins (approved):

ysangkok (approved):

viceice (approved):

rarkins:
I am reverting this PR due to multiple problems raised by users: https://github.com/renovatebot/renovate/pull/34083

PR #33840, opened 2025-01-25
feat(bazel): set sourceUrl for Bazel modules

r0bobo:
## Changes

This updates the Bazel datasource to read the [homepage attribute](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/abseil-cpp/metadata.json#L2) from the Bazel central registry and assign that to `sourceUrl`.

## Context

I've made this change to get automatic changelogs for Bazel module PR:s.

### DIFF
--- a/lib/modules/datasource/bazel/__fixtures__/metadata-no-yanked-versions.json
+++ b/lib/modules/datasource/bazel/__fixtures__/metadata-no-yanked-versions.json
@@ -1,4 +1,5 @@
 {
+    "homepage": "https://github.com/foo/bar",
     "versions": [
         "0.14.8",
         "0.14.9",
--- a/lib/modules/datasource/bazel/index.spec.ts
+++ b/lib/modules/datasource/bazel/index.spec.ts
@@ -58,6 +58,7 @@ describe('modules/datasource/bazel/index', () => {
           { version: '0.15.0' },
           { version: '0.16.0' },
         ],
+        sourceUrl: 'https://github.com/foo/bar',
       });
     });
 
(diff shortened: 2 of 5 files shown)


lib/modules/datasource/bazel/index.spec.ts
```diff
@@ -76,6 +77,7 @@ describe('modules/datasource/bazel/index', () => {
           { version: '0.15.0', isDeprecated: true },
           { version: '0.16.0' },
         ],
+        sourceUrl: 'https://github.com/foo/bar',
```
viceice, line 80: only change one test, so the other verifies it works without

lib/modules/datasource/bazel/index.ts
```diff
@@ -57,6 +57,7 @@ export class BazelDatasource extends Datasource {
           }
           return release;
         });
+      result.sourceUrl = metadata.homepage;
```
viceice, line 60:
```suggestion
      if (metadata.homepage) result.sourceUrl = metadata.homepage;
```

lib/modules/datasource/bazel/__fixtures__/metadata-no-yanked-versions.json
```diff
@@ -1,4 +1,5 @@
 {
+    "homepage": "https://github.com/foo/bar",
```
viceice, line 2: can it be null too?

rarkins:
Homepage is a different concept to source url. Homepage should be used as source url only if:
- source url isn't otherwise defined ✅ 
- the homepage is from a recognized source host 🟥 

I think renovate's datasource wrapper code will handle this logic if basel uses "homepage" instead of "sourceUrl"

rarkins (approved):

viceice (approved):

PR #33848, opened 2025-01-26
feat(cache): Make http cache providers respect `cache-control` header

zharinov:
## Changes

This PR adds support for skipping cache when `cache-control` response header miss the `public` value.

In package-level caching, it's important to separate private packages, so this check is enabled by default.


lib/util/http/cache/abstract-http-cache-provider.ts
```diff
@@ -43,17 +46,35 @@ export abstract class AbstractHttpCacheProvider implements HttpCacheProvider {
...
 
+  private preventCaching<T>(resp: HttpResponse<T>): boolean {
+    if (this.checkCacheControlPublic === false) {
+      return false;
+    }
+
+    const isPublic = resp.headers?.['cache-control']
+      ?.toLocaleLowerCase()
+      ?.split(regEx(/\s*,\s*/))
+      ?.includes('public');
```
viceice, line 57:
```suggestion
      .split(regEx(/\s*,\s*/))
      .includes('public');
```
I think it's safe without

lib/util/http/cache/abstract-http-cache-provider.ts
```diff
@@ -43,17 +46,35 @@ export abstract class AbstractHttpCacheProvider implements HttpCacheProvider {
...
+        return resp;
+      }
+
       const etag = resp.headers?.['etag'];
       const lastModified = resp.headers?.['last-modified'];
 
       HttpCacheStats.incRemoteMisses(url);
 
       const httpResponse = copyResponse(resp, true);
+
```
viceice, line 77: unrelated

viceice (commented):
will it cause less caching/ more requests?

can you please expand the PR description why this is needed?

zharinov:
> will it cause less caching/ more requests?

No, current functionality will work as ususal

lib/util/http/cache/abstract-http-cache-provider.ts
```diff
@@ -43,11 +46,28 @@ export abstract class AbstractHttpCacheProvider implements HttpCacheProvider {
...
+    if (this.checkCacheControlHeader === false) {
+      return false;
+    }
+
+    const isPublic = resp.headers?.['cache-control']
+      ?.toLocaleLowerCase()
+      .split(regEx(/\s*,\s*/))
+      .includes('public');
+
+    return !isPublic;
```
viceice, line 59: I think we should allow caching if the header is missing 🤔

we should first add some logging and track some stats in the hosted app. so we're sure most ecosystems behave properly. and if not we need config options to workaround.

rarkins (approved):

lib/util/http/cache/abstract-http-cache-provider.ts
zharinov: The only place it's being used is NPM. And it's likely to be so for long time, because usually we tend to use cache decorators.

zharinov:
Please merge it

viceice (approved):

PR #33850, opened 2025-01-26
feat(manager/pre-commit): parse 'ssh://git@' in pre-commit repo

Alicipy:
It is possible to define a private hook repository with leading ssh:// protocol in pre-commit (as git:// does not necessarily use SSH), which is handled by pre-commit but not at renovate, this currently gives an invalid url.

This change adds a proper regex so renovate properly parses it.

## Changes

It is possible to define a private hook repository with leading ssh:// protocol in pre-commit (as git:// does not necessarily use SSH), which is handled by pre-commit but not at renovate, this currently gives an invalid url.

This change adds a proper regex so renovate properly parses it, extracting the hostname and depName from it

## Context

See discussion https://github.com/renovatebot/renovate/discussions/33849, I am currently explicitly stating the protocol 'ssh' for my private hook repo, which is not detected correctly.

### DIFF
--- a/lib/modules/manager/pre-commit/extract.ts
+++ b/lib/modules/manager/pre-commit/extract.ts
@@ -96,6 +96,8 @@ function extractDependency(
     regEx('^git@(?<hostname>[^:]+):(?<depName>\\S*)'),
     // This split "git://github.com/pre-commit/pre-commit-hooks" -> "github.com" "pre-commit/pre-commit-hooks"
     regEx(/^git:\/\/(?<hostname>[^/]+)\/(?<depName>\S*)/),
+    // This splits "ssh://git@github.com/pre-commit/pre-commit-hooks" -> "github.com" "pre-commit/pre-commit-hooks"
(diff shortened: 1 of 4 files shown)


rarkins (approved):

viceice (approved):

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -96,6 +96,8 @@ function extractDependency(
     regEx('^git@(?<hostname>[^:]+):(?<depName>\\S*)'),
     // This split "git://github.com/pre-commit/pre-commit-hooks" -> "github.com" "pre-commit/pre-commit-hooks"
     regEx(/^git:\/\/(?<hostname>[^/]+)\/(?<depName>\S*)/),
+    // This splits "ssh://git@github.com/pre-commit/pre-commit-hooks" -> "github.com" "pre-commit/pre-commit-hooks"
+    regEx(/^ssh:\/\/git@(?<hostname>[^/]+)\/(?<depName>\S*)/),
```
viceice, line 100: could also easily be merged to the regex above

PR #33852, opened 2025-01-26
feat(http): Return stale cache on revalidation errors

zharinov:
## Changes

We could return cached result without hitting the server in 2 cases:
- The "soft TTL" is still valid
- The "softTTL" is expired, we hit the server for data fetch/revalidation, but it responded with error. Well, at least we could serve stale data from our cache.

This PR implements the second case.


viceice (commented):
LGTM, but please provide some context

zharinov:
Is current PR description enough?

viceice (approved):

viceice:
> Is current PR description enough?

yes, thanks 🙏

PR #33853, opened 2025-01-26
refactor(npm): Switch to package http cache provider

zharinov:


lib/modules/datasource/npm/__snapshots__/index.spec.ts.snap
```diff
@@ -30,7 +30,6 @@ exports[`modules/datasource/npm/index > should fetch package info from custom re
 
 exports[`modules/datasource/npm/index > should fetch package info from npm 1`] = `
 {
-  "isPrivate": false,
```
zharinov, line 33: Now it's only set for `true`

lib/modules/datasource/npm/get.spec.ts
```diff
@@ -563,69 +565,101 @@ describe('modules/datasource/npm/get', () => {
...
-    setNpmrc('registry=https://test.org\n_authToken=XXX');
-
-    packageCache.get.mockResolvedValueOnce({
-      some: 'result',
-      cacheData: { softExpireAt: '2099' },
-    });
+  describe('cache', () => {
+    const mockTime = (time: string) => {
+      const value = DateTime.fromISO(time).valueOf();
+      Settings.now = () => value;
```
viceice, line 571: use vitest time mocks. it will still work when we drop luxon. In near future the buildin date api will be expanded so we don't need libs like luxon.

viceice (commented):
@rarkins This will flush all npm data from caches

@zharinov only the date mock, otherwise lgtm

viceice (approved):

PR #33892, opened 2025-01-28
feat(presets): add `react-spectrum` monorepo

amitdahan:
## Changes

Adds a `monorepo:react-spectrum` preset sourced from https://github.com/adobe/react-spectrum.

## Context

While using `group:monorepos`, we still get 2 separate bumps for:
- `Update dependency react-aria to v3.37.0`
- `Update dependency react-aria-components to v1.6.0`

Even though they have shared monorepo dependencies, and should ideally be bumped together.

### DIFF
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -485,6 +485,7 @@
     "react": "https://github.com/facebook/react",
     "react-admin": "https://github.com/marmelab/react-admin",
     "react-apollo": "https://github.com/apollographql/react-apollo",
+    "react-aria": "https://github.com/adobe/react-spectrum",
     "react-dnd": "https://github.com/react-dnd/react-dnd",
     "react-navigation": "https://github.com/react-navigation/react-navigation",
     "react-page": "https://github.com/react-page/react-page",
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -485,7 +485,6 @@
     "react": "https://github.com/facebook/react",
     "react-admin": "https://github.com/marmelab/react-admin",
     "react-apollo": "https://github.com/apollographql/react-apollo",
-    "react-aria": "https://github.com/adobe/react-spectrum",
     "react-dnd": "https://github.com/react-dnd/react-dnd",
     "react-navigation": "https://github.com/react-navigation/react-navigation",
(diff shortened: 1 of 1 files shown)


lib/data/monorepo.json
```diff
@@ -485,6 +485,7 @@
     "react": "https://github.com/facebook/react",
     "react-admin": "https://github.com/marmelab/react-admin",
     "react-apollo": "https://github.com/apollographql/react-apollo",
+    "react-aria": "https://github.com/adobe/react-spectrum",
```
rarkins, line 488: Won't this rule capture all packages published from that repo? There seems to be a name/repo mismatch here where this "Update react-aria packages" monorepo rule would group more than just Aria

amitdahan: Fair enough, that repo contains the following:

[React Spectrum](https://react-spectrum.adobe.com/react-spectrum/index.html):
- `@adobe/react-spectrum`
- `@react-spectrum/*`
- `@spectrum-icons/*`

[React Aria](https://react-spectrum.adobe.com/react-aria/index.html):
- `react-aria`
- `react-aria-components`
- `tailwindcss-react-aria-components`
- `@react-aria/*`

[React Stately](https://react-spectrum.adobe.com/react-stately/index.html):
- `react-stately`
- `@react-stately/*`

Some utilities shared between all/most of the above:
- `@react-types/*`
- `@internationalized/*`

Devs might use React Spectrum, or React Aria, or React Stately, or a mixture.

Since they all share a lot of common dependencies (that devs might also install themselves) it's extra important to bump them together.

I guess I can see why `react-aria` isn't a good enough fit.

I guess `react-spectrum` is at least fitting enough as all of the above packages live in the `react-spectrum` repo, even though many of its users might not be using React Spectrum itself.

I'll rename to `react-spectrum`.

rarkins (approved):

viceice:
they're versioned differently, so PRs are now immortal 🫣

@rarkins @secustor we should create a new docs policy to clarify monorepos. we ideally should only group them by default if they really need to be grouped.

this PR is just an optimistic group. I don't like to have immortal PRs by default. especially in the recommended preset.

we should make those monorepos a separate preset.

PR #33901, opened 2025-01-28
feat(http): Switch to memory cache provider

zharinov:
## Changes

Make caching of GET/HEAD HTTP requests disabled by default.
When it's needed, the `memCacheProvider` is be used.


zharinov:
- Blocked by #33900

rarkins (commented):
I will test first

lib/modules/platform/github/branch.ts
```diff
@@ -13,7 +13,6 @@ async function matchingBranches(
 ): Promise<string[]> {
   const { body: branches } = await githubApi.getJson(
     `/repos/${repo}/git/matching-refs/heads/${branchName}`,
-    { memCache: false },
```
rarkins, line 16: Can you explain why the memCache=false parts are no longer needed?

zharinov: And yes, no memory cache is the default now

rarkins: How can we be confident that we haven't _missed_ anything?

zharinov: Generally speaking, we can't. What we can is analyze HTTP stats and watch for repeating GET requests (sometimes, though, it's intended).

rarkins (commented):
Can you summarize all the cases where http memcache is still being explicitly enabled in this PR (seems to be within platform code) and we evaluate whether we're better to add platform-level caching (e.g. of PR lists) instead of relying on http-level? Maybe you can do this through inline PR comments

lib/modules/platform/bitbucket/index.ts
```diff
@@ -301,10 +301,11 @@ export async function findPr({
   logger.debug(`findPr(${branchName}, ${prTitle}, ${state})`);
 
   if (includeOtherAuthors) {
-    // PR might have been created by anyone, so don't use the cached Renovate PR list
     const prs = (
       await bitbucketHttp.getJsonUnchecked<PagedResult<PrResponse>>(
         `/2.0/repositories/${config.repository}/pullrequests?q=source.branch.name="${branchName}"&state=open`,
+        // PR might have been created by anyone, so don't use the short-term memory cache only
```
rarkins, line 307: What does this mean?

zharinov: Reformulated

rarkins (approved):
This should have a _significant_ reduction in the number of http requests which we cache in memory. This should apply especially to datasource requests

secustor (commented):
Code wise this LGTM, but what is the reason behind this? Accumulating memory if Renovate is run against big number of repositories?  

I'm not sure if we are trading here speed ( because we running only a limited number concurrent requests ) with memory and that should be at least an informed decision.

rarkins:
Memory cache is cleared per repo. So this influences peak memory use within a single repo job. In the Mend hosted app we're seeing multi GB RAM for some repos, and exceeding our assigned limits at times. Packages like Renovate on npmjs return over 100MB, for example

rarkins:
Btw this isn't meant to be a speed trade off. Requests which need men caching should have it, but those are rare.

secustor:
K, let's try it then. I'm mostly worried about our concurrency limits in combination with API rate limits of 3rd parties.

rarkins (changes requested):
I would like to take an additional step before this to reduce the chances we need to "test in production" to find caching we've missed.

Conceptually, this PR changes us from http memcache by default (opt out) to opt-in http memcache. The theory is that we already do enough caching at platform and datasource layers that the additional default caching at http layer is wasteful. The reality is that there may be some we _miss_, and that instead of "efficiently" returning cached http results, we're doing to increase http requests externally, increase runtime, etc.

So I propose keeping _both_ types of caches, and logging counts whenever the existing/simple http cache returns a result. In other words, keep the existing cache as a fallback so that we can be more confident that there's no "missed" caching needs before we then turn it off.

rarkins:
@zharinov can we _log_ when caching fails/falls through to the lowest level?

lib/util/http/index.ts
```diff
@@ -198,11 +196,25 @@ export class Http<Opts extends HttpOptions = HttpOptions> {
...
+
     let resPromise: Promise<HttpResponse<T>> | null = null;
 
     // Cache GET requests unless memCache=false
     if (memCacheKey) {
       resPromise = memCache.get(memCacheKey);
+
+      // istanbul ignore next
+      if (resPromise && options.cacheProvider !== memCacheProvider) {
+        logger.debug({ url }, 'Cache hit on the obsolete memCache option');
```
rarkins, line 216: I switched this to logger.error and ran on containerbase/base with dryRun=lookup and got many errors. They're all either `/v2/` queries or `/token?` ones from docker datasource.

Maybe this datasource should do its _own_ caching of these, with its own TTLs?

rarkins (changes requested):
Also needs deconflicting

lib/util/http/index.ts
rarkins: I think I'd prefer to accumulate these URLs into a Set and then at the end print a debug message about number of URLs using legacy caching. Then we can capture and graph that in our hosted app and know when we reach 0

lib/util/http/http.ts
```diff
@@ -176,11 +172,26 @@ export abstract class HttpBase<
...
+
     let resPromise: Promise<HttpResponse<unknown>> | null = null;
 
     // Cache GET requests unless memCache=false
     if (memCacheKey) {
       resPromise = memCache.get(memCacheKey);
+
+      /* v8 ignore start: temporary code */
+      if (resPromise && options.cacheProvider !== memCacheProvider) {
+        logger.debug({ url }, 'Cache hit on the obsolete memCache option');
```
rarkins, line 192: Add these URLs to a set, and then log the full list at the end of a run (instead of one debug at a time)

rarkins (changes requested):
Needs a refactor of docker datasource too

lib/util/stats.ts
```diff
@@ -488,3 +488,61 @@ export class HttpCacheStats {
...
+      const callsite = this.getCallsite();
+      data[url] = { callsite, count: 0 };
+    }
+    data[url].count++;
+    memCache.set('obsolete-cache-stats', data);
+  }
+
+  static report(): void {
+    const hits = this.getData();
+    logger.debug({ count: Object.keys(hits).length, hits }, 'Cache hit URLs');
```
rarkins, line 545:
```suggestion
    logger.debug({ count: Object.keys(hits).length, hits }, 'Cache fallback URLs');
```

Bruno-bot-stack (approved):
Leave a review

rarkins (changes requested):
When I run on a fork of the renovate repo, and there are PRs needing changelogs, we see some more cache fallbacks:

```
DEBUG: Cache hit URLs (repository=renovate-reproductions/renovate2)
       "count": 7,
       "hits": {
         "https://api.github.com/repos/yarnpkg/berry": {
           "callsite": "async Object.getReleaseNotesMd (/Users/rhys/src/renovatebot/renovate/lib/workers/repository/update/pr/changelog/github/index.ts)",
           "count": 1
         },
         "https://api.github.com/repos/yarnpkg/berry/git/trees/master?recursive=1": {
           "callsite": "async Object.getReleaseNotesMd (/Users/rhys/src/renovatebot/renovate/lib/workers/repository/update/pr/changelog/github/index.ts)",
           "count": 1
         },
         "https://api.github.com/repos/yarnpkg/berry/git/blobs/3e96b54ffe13ba43b83ee1844a3ba37aeb4f77f8": {
           "callsite": "async Object.getReleaseNotesMd (/Users/rhys/src/renovatebot/renovate/lib/workers/repository/update/pr/changelog/github/index.ts)",
           "count": 1
         },
         "https://api.github.com/repos/aws/aws-sdk-js-v3": {
           "callsite": "async Object.getReleaseNotesMd (/Users/rhys/src/renovatebot/renovate/lib/workers/repository/update/pr/changelog/github/index.ts)",
           "count": 5
         },
         "https://api.github.com/repos/aws/aws-sdk-js-v3/git/trees/main?recursive=1": {
           "callsite": "async Object.getReleaseNotesMd (/Users/rhys/src/renovatebot/renovate/lib/workers/repository/update/pr/changelog/github/index.ts)",
           "count": 5
         },
         "https://api.github.com/repos/eslint/eslint": {
           "callsite": "async Object.getReleaseNotesMd (/Users/rhys/src/renovatebot/renovate/lib/workers/repository/update/pr/changelog/github/index.ts)",
           "count": 1
         },
         "https://api.github.com/repos/eslint/eslint/git/blobs/7d070fd0c3cbfeb53f719f85ee0e6677b70a0f5b": {
           "callsite": "async Object.getReleaseNotesMd (/Users/rhys/src/renovatebot/renovate/lib/workers/repository/update/pr/changelog/github/index.ts)",
           "count": 1
         }
       }
```

rarkins (changes requested):
When running again on `renovate-reproductions/renovate2`, I still get this:

```
DEBUG: Cache fallback URLs (repository=renovate-reproductions/renovate2)
       "count": 4,
       "hits": {
         "https://api.github.com/repos/yarnpkg/berry": {
           "callsite": "async Object.getReleaseNotesMd (/home/rhys/src/github/renovatebot/renovate/lib/workers/repository/update/pr/changelog/github/index.ts)",
           "count": 1
         },
         "https://api.github.com/repos/yarnpkg/berry/git/blobs/3e96b54ffe13ba43b83ee1844a3ba37aeb4f77f8": {
           "callsite": "async Object.getReleaseNotesMd (/home/rhys/src/github/renovatebot/renovate/lib/workers/repository/update/pr/changelog/github/index.ts)",
           "count": 1
         },
         "https://api.github.com/repos/renovate-reproductions/renovate2/commits/renovate/main-yarn-monorepo/check-runs?per_page=100": {
           "callsite": "async Proxy.getBranchStatus (/home/rhys/src/github/renovatebot/renovate/lib/modules/platform/github/index.ts)",
           "count": 1
         },
         "https://api.github.com/repos/renovate-reproductions/renovate2/commits/renovate/main-mkdocs-material-9.x/check-runs?per_page=100": {
           "callsite": "async Proxy.getBranchStatus (/home/rhys/src/github/renovatebot/renovate/lib/modules/platform/github/index.ts)",
           "count": 1
         }
       }
```

rarkins (dismissed):

secustor (dismissed):

secustor:
LGTM, though needs a conflict resolution

rarkins (approved):

felipecrs:
@zharinov, @rarkins this change broke Gerrit platform in several ways. Some examples:

1. This error now always throws:
   https://github.com/renovatebot/renovate/blob/847c9c60f2de5f142024ddf4315bdd59ac5dfd2d/lib/modules/platform/gerrit/index.ts#L177-L192

2. PR limits (concurrent/hourly) are never refreshed, which makes Renovate create PRs limitless.

I believe the only explanation for this is that the previous `httpOptions.memCache: true` was not working/not doing what it was intended to.

And now that the memory cache implementation is working, we are catching these issues.

But it's also weird because in the first point above, `refreshCache` is already being set to true, and the response still seems to be returning a cached version.

felipecrs:
I wonder if the other platforms that were also using `memCache` are not affected by this change, like BitBucket.

I recommend disabling memory caching unconditionally for Gerrit platform now and re-evaluate where it is safe to re-enable in some places later.

I can push a PR for it, if welcome.

zharinov:
> I recommend disabling memory caching unconditionally for Gerrit platform now and re-evaluate where it is safe. And re-enable in some places later.

Agreed, I think for this particular platform it's not critical to make couple more requests

viceice (approved):

felipecrs:
I pushed a fix: https://github.com/renovatebot/renovate/pull/35056

PR #33904, opened 2025-01-28
feat(http): Implement soft and hard TTL for all cache providers

zharinov:
## Changes

- Move TTL logic to abstract provider
- For repository cache, set soft ttl to zero, and hart TTL to 365 days
- For package cache, nothing changes
- Don't require cache to have at least one of `etag` or `lastModified`, now they're optional (useful for future memory cache implementation)


zharinov:
I think, it actually is not necessary

PR #33907, opened 2025-01-29
fix(http): allow overwriting default Renovate headers

samgiz:
User specified headers should be able to overwrite the default headers.

## Changes

The headers set by Renovate by default (`Accept` and `User-Agent` are the only ones I have observed) can now be overwritten if specified in `renovate.json` and allowed by `allowedHeaders`.

## Context

I encountered this issue when trying to retrieve the raw contents of a file from GitHub and needed to set the `"Accept": "application/vnd.github.v3.raw"` header for that purpose.

### DIFF
--- a/lib/util/http/host-rules.spec.ts
+++ b/lib/util/http/host-rules.spec.ts
@@ -543,4 +543,24 @@ describe('util/http/host-rules', () => {
       },
     });
   });
+
+  it('should replace existing headers with host rule headers', () => {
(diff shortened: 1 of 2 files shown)


rarkins (changes requested):
Please add a test so that this doesn't regress (test should fail on main branch, pass on this branch)

rarkins (dismissed):

rarkins (approved):

PR #33911, opened 2025-01-29
fix(manager): retain `depName` when using `registryAliases`

viceice:
## Changes
Retain `depName` when using `registryAliases`, so only replace `packageName` for lookup.
`registryAliases` where never intended to be shown in commit / PR.

Technically a breaking change because of the depName change which causes title and branch name changes.

## Context
- https://github.com/renovate-reproductions/33911-docker-reg-aliases/pull/6


lib/modules/platform/gitea/schema.ts
rarkins, line 1: Is this unrelated?

rarkins (changes requested):
Needs testing on a real repo

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -176,14 +177,19 @@ export function getDep(
   // Resolve registry aliases first so that we don't need special casing later on:
   for (const [name, value] of Object.entries(registryAliases ?? {})) {
     const escapedName = escapeRegExp(name);
-    const groups = regEx(`(?<prefix>${escapedName})/(?<depName>.+)`).exec(
+    const groups = regEx(`${escapedName}/(?<depName>.+)`).exec(
```
viceice, line 180: technically we can use a registy alias like this: `domain.test=>docker.io`

this will replace like this `some.domain.test/some/image` to `docker.io/some/image`

I'm not sure if this is intended behavior 🤔

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -141,6 +141,7 @@ export function splitImageParts(currentFrom: string): PackageDependency {
 
   const dep: PackageDependency = {
     depName,
+    packageName: depName,
```
viceice, line 144: always set `packageName`

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -176,14 +177,19 @@ export function getDep(
...
       currentFrom,
     )?.groups;
     if (groups) {
       const dep = {
-        ...getDep(`${value}/${groups.depName}`),
+        ...getDep(`${value}/${groups.depName}`, false),
         replaceString: currentFrom,
       };
+      // retain depName, not sure if condition is necessary
+      if (dep.depName?.startsWith(value)) {
```
viceice, line 189: I'm not sure if this is ever false 🤔

lib/modules/platform/gitea/schema.ts
viceice: yes, sorry

secustor (commented):
Needs test fixes, but else LGTM

viceice:
> Needs testing on a real repo

test pr linked

secustor (approved):

rarkins (approved):

PR #33919, opened 2025-01-29
fix(batec): don't override versioning

viceice:
## Changes
Use default versioning of datasource. it allows passing special handling for debian and ubuntu.
Also add docker to supported datasources

### DIFF
--- a/lib/modules/manager/batect/schema.ts
+++ b/lib/modules/manager/batect/schema.ts
@@ -1,7 +1,6 @@
 import { z } from 'zod';
 import { LooseArray, LooseRecord, Yaml } from '../../../util/schema-utils';
 import { GitTagsDatasource } from '../../datasource/git-tags';
-import { id as dockerVersioning } from '../../versioning/docker';
 import { id as semverVersioning } from '../../versioning/semver';
 import { getDep } from '../dockerfile/extract';
 import type { PackageDependency } from '../types';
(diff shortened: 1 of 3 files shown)


lib/modules/manager/batect/schema.ts
```diff
@@ -30,10 +29,8 @@ export const BatectConfigSchema = Yaml.pipe(
     ).catch([]),
   }),
 ).transform(({ containers, include }) => {
-  const imageDependencies = containers.map((image) => ({
-    ...getDep(image),
-    versioning: dockerVersioning,
-  }));
+  // TODO: @zharinov How to pass `registryAliases` to `getDep`?
+  const imageDependencies = containers.map((image) => getDep(image));
```
zharinov, line 33:
```suggestion
```

lib/modules/manager/batect/schema.ts
```diff
@@ -30,10 +29,8 @@ export const BatectConfigSchema = Yaml.pipe(
     ).catch([]),
   }),
 ).transform(({ containers, include }) => {
```
zharinov, line 31:
```suggestion
).transform(({ containers: imageDependencies, include }) => {
```

zharinov (changes requested):
Skip doing `.map()` in the schema and do the transform in the scope where your alias map is available.

zharinov:
Another option is to move transform to the scope where alias is available

rarkins (approved):

viceice:
ok. please pay attention for it when creating more schema. i'll enable registry aliases for all docker images usages tonight or tomorrow

PR #33922, opened 2025-01-29
fix: only replace prefix with registry alias

viceice:
## Changes
Currently it would replace `proxy.com/repo/docker.io/bitnami/kubectl` to `public.ecr.aws/bitnami/kubectl` if we define an alias like `docker.io/bitnami>public.ecr.aws/bitnami`.
This isn't an expected behavior.
Also all samples are only showing prefix handling.


lib/modules/manager/dockerfile/extract.ts
```diff
@@ -168,24 +168,26 @@ export function getDep(
   specifyReplaceString = true,
   registryAliases?: Record<string, string>,
 ): PackageDependency {
-  if (!is.string(currentFrom) || is.emptyStringOrWhitespace(currentFrom)) {
+  if (
+    !is.string(currentFrom) ||
+    !is.nonEmptyStringAndNotWhitespace(currentFrom)
```
secustor, line 173: I guess the `is.string` is because of the type assertion of `nonEmptyStringAndNotWhitespace` does not narrow this to `string`?

viceice: yes, before it narrowed the type here to string, so after this condition the type was never. 🤷‍♂️

secustor (dismissed):

rarkins (approved):

PR #33949, opened 2025-01-30
feat: Add `mjs` options to prettier

alexrecuenco:
See here: https://prettier.io/docs/en/configuration.html

Now ESM modules are allowed for prettier config files

At the same time, it might make sense to use [prettier.resolveConfig](https://prettier.io/docs/en/api.html#prettierresolveconfigfileurlorpath--options), which returns `null` if no config is found. That way you don't need to keep matching prettier's interface ad-infinitum 

## Changes

Prettier allows mjs files.

## Context

Prettier now allows ESM configuration

### DIFF
--- a/lib/workers/repository/config-migration/branch/migrated-data.ts
+++ b/lib/workers/repository/config-migration/branch/migrated-data.ts
@@ -32,8 +32,10 @@ const prettierConfigFilenames = new Set([
   '.prettierrc.json5',
   '.prettierrc.js',
   '.prettierrc.cjs',
+  '.prettierrc.mjs',
   'prettier.config.js',
   'prettier.config.cjs',
+  'prettier.config.mjs',
   '.prettierrc.toml',
 ]);
 


viceice (approved):
can you propose a PR to use the resolve config function?

alexrecuenco:
I am not sure I get what "Snapshots have been added in this PR. Use toMatch instead." means

alexrecuenco:
> can you propose a PR to use the resolve config function?

If I have enough time, I'll try, I am unfortunately quite busy lately

rarkins:
Doesn't prettier.resolveConfig also _execute_ any JS found? Ie a massive attack vector if renovate were to allow users to execute arbitrary code?

alexrecuenco:
> Doesn't prettier.resolveConfig also _execute_ any JS found? Ie a massive attack vector if renovate were to allow users to execute arbitrary code?

If you know that to be the case, then ignore the suggestion,

alexrecuenco:
I am getting in the checks some pnpm checks failing... but I got the impression it might be unrelated to the change here, any idea?

PR #33969, opened 2025-01-31
feat(presets): add payloadcms monorepo

dsm23:
## Changes

Add payloadcms to the list of monorepos

## Context

Payloadcms is a relatively new CMS solution and at the time of writing is far from the most successful CMS product however it requires its npmjs modules to all be using the same version and it has a double-digit number of official packages.

I guess the renovate docs of monorepo presets would need to be updated but I don't know how that is done.

I have not run any unit tests

### DIFF
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -463,6 +463,7 @@
     "orleans": "https://github.com/dotnet/orleans",
     "panda-css": "https://github.com/chakra-ui/panda",
     "parcel": "https://github.com/parcel-bundler/parcel",
+    "payloadcms": "https://github.com/payloadcms/payload",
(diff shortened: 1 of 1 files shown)


lib/data/monorepo.json
```diff
@@ -471,6 +471,7 @@
     ],
     "playwright": "https://github.com/microsoft/playwright",
     "playwright-dotnet": "https://github.com/microsoft/playwright-dotnet",
+    "payloadcms": "https://github.com/payloadcms/payload",
```
MindTooth, line 474: That is not alphabetical.  `payloadcms` should be under `parcel`.

viceice (approved):

PR #33971, opened 2025-01-31
fix(lockfilemaintenance): ensure `isLockFileMaintenance` on grouping

viceice:
## Changes
ensure `isLockFileMaintenance` is `true` if grouping lockefile maintenance with other updates even it's not really supported.

## Context
- closes #32384

### DIFF
--- a/lib/modules/manager/gleam/artifacts.ts
+++ b/lib/modules/manager/gleam/artifacts.ts
@@ -17,7 +17,7 @@ export async function updateArtifacts(
   const { packageFileName, updatedDeps, newPackageFileContent, config } =
     updateArtifact;
   logger.debug(`gleam.updateArtifacts(${packageFileName})`);
-  const isLockFileMaintenance = config.updateType === 'lockFileMaintenance';
+  const { isLockFileMaintenance } = config;
 
   if (is.emptyArray(updatedDeps) && !isLockFileMaintenance) {
     logger.debug('No updated gleam deps - returning null');
--- a/lib/workers/repository/updates/generate.ts
+++ b/lib/workers/repository/updates/generate.ts
@@ -436,9 +436,23 @@ export function generateBranchConfig(
         .reduce((a, b) => a.concat(b), []),
     ),
   ];
+
   if (config.upgrades.some((upgrade) => upgrade.updateType === 'major')) {
     config.updateType = 'major';
   }
+
+  // explicit set `isLockFileMaintenance` for the branch for groups
+  if (config.upgrades.some((upgrade) => upgrade.isLockFileMaintenance)) {
+    config.isLockFileMaintenance = true;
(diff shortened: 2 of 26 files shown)


lib/modules/manager/gleam/artifacts.ts
```diff
@@ -17,7 +17,7 @@ export async function updateArtifacts(
   const { packageFileName, updatedDeps, newPackageFileContent, config } =
     updateArtifact;
   logger.debug(`gleam.updateArtifacts(${packageFileName})`);
-  const isLockFileMaintenance = config.updateType === 'lockFileMaintenance';
+  const { isLockFileMaintenance } = config;
```
rarkins, line 20: When I see old code which references `config.updateType`, it implies to me that `config` just be an "update" config and not a "branch" config.

lib/workers/repository/updates/generate.ts
```diff
@@ -436,9 +436,23 @@ export function generateBranchConfig(
         .reduce((a, b) => a.concat(b), []),
     ),
   ];
+
   if (config.upgrades.some((upgrade) => upgrade.updateType === 'major')) {
     config.updateType = 'major';
   }
+
+  // explicit set `isLockFileMaintenance` for the branch for groups
+  if (config.upgrades.some((upgrade) => upgrade.isLockFileMaintenance)) {
```
rarkins, line 445: Here, config = branch config. So you're setting branchConfig.isLockFileMaintenance=true if any update in the branch  has isLockFileMaintenance set. This seems logically correct. But I think `updateArtifacts()` is no called on a full branch, so the refactorings within manager code seem unrelated or mismatched with this change?

lib/modules/manager/gleam/artifacts.ts
viceice: no. it's the branch config, so it simply assumes the proper update type was copied to branch config

lib/workers/repository/updates/generate.ts
viceice: No, it's passing the full branch config. `patchConfigForArtifactsUpdate` is only patching the `lockFiles` property and `config` is of `BranchConfig` type.

https://github.com/renovatebot/renovate/blob/f182708232b5aa623b77f0f3dc7d5e2e002b6567/lib/workers/repository/update/branch/get-updated.ts#L334-L344
https://github.com/renovatebot/renovate/blob/f182708232b5aa623b77f0f3dc7d5e2e002b6567/lib/workers/repository/update/branch/get-updated.ts#L375-L385
https://github.com/renovatebot/renovate/blob/f182708232b5aa623b77f0f3dc7d5e2e002b6567/lib/workers/repository/update/branch/get-updated.ts#L421-L429

secustor (approved):
I would test this at least once on a real repo.

PR #33992, opened 2025-02-02
fix(manager/jsonata): populate extract result with manager config fields

RahulGautamSingh:
## Changes
- Populate the manager config fields necessary for extraction in the extract result

## Context
- These fields (especially `matchStrings` & `fileFormat` ) are necessary in the `confirmIfDepUpdated` function where we re-extract after updating the content in `doAutoReplace`. 

The test runs on the renovate repo passed without these fields after which I had removed them. But when I tried some other variations it failed. I think it was working before because the `fileFormat` had a default value `json`, which prevented the errors.

 https://github.com/Rahul-renovate-testing/migrate-cstm-preset-1
 https://github.com/Rahul-renovate-testing/renovate/pulls

### DIFF
--- a/lib/modules/manager/custom/jsonata/index.ts
+++ b/lib/modules/manager/custom/jsonata/index.ts
@@ -43,28 +43,25 @@ export async function extractPackageFile(
     return null;
   }
 
-  let deps = await handleMatching(json, packageFile, config);
-
-  // filter all null values
-  deps = deps.filter(is.truthy);
-  if (deps.length) {
-    const res: PackageFileContent & JSONataManagerTemplates = {
-      deps,
-      matchStrings: config.matchStrings,
-      fileFormat: config.fileFormat,
-    };
-
-    // copy over templates for autoreplace
-    for (const field of validMatchFields.map(
-      (f) => `${f}Template` as keyof JSONataManagerTemplates,
-    )) {
-      if (config[field]) {
-        res[field] = config[field];
-      }
-    }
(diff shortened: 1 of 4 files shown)


lib/modules/manager/custom/jsonata/index.ts
```diff
@@ -42,12 +43,28 @@ export async function extractPackageFile(
...
   }
 
-  const deps = await handleMatching(json, packageFile, config);
-  if (!deps.length) {
-    return null;
+  let deps = await handleMatching(json, packageFile, config);
+
+  // filter all null values
+  deps = deps.filter(is.truthy);
+  if (deps.length) {
```
viceice, line 50: invert to reduce nesting

lib/modules/manager/custom/jsonata/index.ts
```diff
@@ -42,12 +43,28 @@ export async function extractPackageFile(
     return null;
   }
 
-  const deps = await handleMatching(json, packageFile, config);
-  if (!deps.length) {
-    return null;
+  let deps = await handleMatching(json, packageFile, config);
+
+  // filter all null values
+  deps = deps.filter(is.truthy);
```
viceice, line 49: isn't that already done in the handle function? 🤔

RahulGautamSingh: True

viceice (approved):

PR #34007, opened 2025-02-03
build(docker): use pnpm instead of corepack

viceice:
## Changes
install pnpm instead of using corepack to no longer fail ci (only for build step).

## Context
Corepack is failing because of some pnpm packaging changes, so simply use pnpm directly.


viceice:
ignore devcontainer, it's fixed separately https://github.com/containerbase/devcontainer/blob/3b03c6cd3b6796ec06811790ca94dbdd1fa8f2f8/Dockerfile#L62

rarkins (approved):

PR #34022, opened 2025-02-04
fix(gitea): use endpoint for pr cache pagination

viceice:
## Changes
Use configured endpoint in pr cache. The cache uses own pagination.

## Context
- closes #33121


viceice:
@zharinov i think we should create / update the http class to return an async iterator when paginate is true, so we can iterate and cancel fetching. That way we don't need another pagination just for pr cache. WDYT?

rarkins (commented):
So will this always rewrite, and ignore the host in the link header?

viceice:
i can also change it to simple pagination, like we do in `GiteaHttp` and Github pr cache 🤷‍♂️

rarkins (approved):

PR #33952, opened 2025-01-30
feat(pip-compile): support `uv pip compile`

maxbrunet:
## Changes

This adds the basics to support `uv pip compile` inside the `pip-compile` manager.

## Context

Closes #30909

Relevant docs:

* [uv | Commands - `uv pip compile`](https://docs.astral.sh/uv/reference/cli/#uv-pip-compile)
* [uv | Compatibility with pip - `pip compile` defaults](https://docs.astral.sh/uv/pip/compatibility/#pip-compile-defaults)

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

PR #34029, opened 2025-02-04
feat(pip-compile): support tool version constraints with uv

maxbrunet:
## Changes

* Extract Python version constraint from uv's `--python-version` flag
* Support `uv` tool constraint from config

## Context

Follow up to https://github.com/renovatebot/renovate/pull/33952#discussion_r1940231210


lib/modules/manager/pip-compile/artifacts.ts
```diff
@@ -115,11 +115,12 @@ export async function updateArtifacts({
...
-      const pythonVersion = extractPythonVersion(
-        compileArgs.commandType,
-        existingOutput,
-        outputFileName,
-      );
+      let pythonVersion: string | undefined;
+      if (compileArgs.commandType === 'uv') {
+        pythonVersion = compileArgs.pythonVersion;
+      } else {
+        pythonVersion = extractPythonVersion(existingOutput, outputFileName);
```
rarkins, line 122: Shouldn't you still fall back to this if python version is missing from the args?

maxbrunet: That would not be useful since `uv` does not write the Python version to the header. If the user does not specify the version with `--python-version`, `uv`'s behavior is to use the latest version available in the environment

zanieb: @viceice can you clarify what you mean by that?

maxbrunet: I want to say, yes, but I'm not clear on what you mean either

secustor: He is referring to what does take priority? The python version extracted from the package file or user configuration in `renovate.json`

maxbrunet: If so, yes, it prefers the value in `renovate.json` over the extracted value, I have not changed this part of the code:

https://github.com/renovatebot/renovate/blob/4e48b5f44299b5bf33ba4fc8e870674c49373d40/lib/modules/manager/pip-compile/common.ts#L18-L36

maxbrunet:
Anything needs to be addressed in this PR?

viceice (approved):

PR #34030, opened 2025-02-04
docs(managers/pip-compile): update list of source files and managers

maxbrunet:
## Changes

As in title

## Context

Follow up to https://github.com/renovatebot/renovate/pull/33952#issuecomment-2632353684

This has been in the code for a while, but the doc was not updated:

https://github.com/renovatebot/renovate/blob/4cfc794fd9cb150599d508eb3e6c67f1aa8dca65/lib/modules/manager/pip-compile/common.ts#L350-L365

### DIFF
--- a/lib/modules/manager/pip-compile/readme.md
+++ b/lib/modules/manager/pip-compile/readme.md
@@ -38,7 +38,7 @@ The following files are currently supported:
 |       `setup.py` | `pip_setup`        |
 |      `setup.cfg` | `setup-cfg`        |
 | `pyproject.toml` | `pep621`           |
-|           `*.in` | `pip_requirements` |
+| `*.in` / `*.txt` | `pip_requirements` |
 
 Example header:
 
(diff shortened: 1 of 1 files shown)


secustor (approved):

ShepleySound:
I can open a new issue if this is in the wrong spot, but we updated a project in response to this doc change and it seems like there might be more work to do. `matchManager` does indeed match these files, but `extractPackageFile` has no support for `pep621` or `setup-cfg`. https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/pip-compile/extract.ts#L22

So when Renovate runs against a `requirements.txt` that references `pyproject.toml`, we get the following message:

```
 WARN: pip-compile: support for manager is not yet implemented (repository=..., packageFile=pyproject.toml)
       "manager": "pep621"
 WARN: pip-compile: failed to find dependencies in source file (repository=..., packageFile=pyproject.toml)
```

maxbrunet:
Thank you @ShepleySound for reporting it, that was overlooked and the doc change has been reverted (#34317)

PR #34092, opened 2025-02-07
chore(eslint): test unused-directives in CI

rarkins:


rarkins:
@viceice @zharinov any idea why this wouldn't be now causing test failures?

It's being run here: https://github.com/renovatebot/renovate/actions/runs/13199018093/job/36846619944?pr=34092#step:5:23

But maybe the syntax is somehow wrong?

rarkins:
(To clarify, I had expected the tests on this PR to _fail_)

rarkins:
Actually I see that the eslint-ci takes only 1s to run, so either it's not running, or the cache is bypassing this change

viceice:
we probably should not retain the cache

viceice (approved):

PR #34102, opened 2025-02-07
feat(platform/gitlab): use Notes API for automerge to support merge trains

### LINKED ISSUE #5573: Support GitLab "merge trains"

Support GitLab's "merge train" concept for dependencies, if it's feasible.

Ideally Renovate MRs could be layered on top of each other to reduce the amount of rebasing required after each merge.

Today, rebasing can be an O(n) task.

**Additional context**

https://docs.gitlab.com/ee/ci/merge_request_pipelines/pipelines_for_merged_results/merge_trains/
https://about.gitlab.com/blog/2020/01/30/all-aboard-merge-trains/

Preisschild:
closes https://github.com/renovatebot/renovate/issues/5573

Gitlab's /merge API does not support merge trains, but sending "/merge" over the MR comment (notes) API does work:
https://gitlab.com/gitlab-org/gitlab/-/issues/32665

## Changes

When using automerge with Gitlab, renovate will now add MRs to merge trains. If not it will be merged as previously.

## Context

- https://github.com/renovatebot/renovate/issues/5573

### DIFF
--- a/lib/modules/platform/gitlab/index.ts
+++ b/lib/modules/platform/gitlab/index.ts
@@ -658,6 +658,8 @@ async function tryPrAutomerge(
         250,
       );
 
+      let diffHeadSha = '';
+
       // Check for correct merge request status before setting `merge_when_pipeline_succeeds` to  `true`.
       for (let attempt = 1; attempt <= retryTimes; attempt += 1) {
         const { body } = await gitlabApi.getJsonUnchecked<{
@@ -666,9 +668,13 @@ async function tryPrAutomerge(
           pipeline: {
             status: string;
           };
+          sha: string;
         }>(`projects/${config.repository}/merge_requests/${pr}`, {
           memCache: false,
         });
+
+        diffHeadSha = body.sha;
+
         // detailed_merge_status is available with Gitlab >=15.6.0
         const use_detailed_merge_status = !!body.detailed_merge_status;
         const detailed_merge_status_check =
@@ -690,28 +696,15 @@ async function tryPrAutomerge(
         await setTimeout(mergeDelay * attempt ** 2); // exponential backoff
       }
 
-      // Even if Gitlab returns a "merge-able" merge request status, enabling auto-merge sometimes
-      // returns a 405 Method Not Allowed. It seems to be a timing issue within Gitlab.
-      for (let attempt = 1; attempt <= retryTimes; attempt += 1) {
-        try {
-          await gitlabApi.putJson(
-            `projects/${config.repository}/merge_requests/${pr}/merge`,
-            {
-              body: {
-                should_remove_source_branch: true,
-                merge_when_pipeline_succeeds: true,
-              },
-            },
-          );
-          break;
-        } catch (err) {
-          logger.debug(
-            { err },
-            `Automerge on PR creation failed. Retrying ${attempt}`,
-          );
-        }
-        await setTimeout(mergeDelay * attempt ** 2); // exponential backoff
-      }
+      await gitlabApi.postJson(
+        `projects/${config.repository}/merge_requests/${pr}/notes`,
+        {
+          body: {
+            body: '/merge',
+            merge_request_diff_head_sha: diffHeadSha,
+          },
+        },
+      );
     }
   } catch (err) /* istanbul ignore next */ {
     logger.debug({ err }, 'Automerge on PR creation failed');
--- a/lib/modules/platform/gitlab/index.ts
+++ b/lib/modules/platform/gitlab/index.ts
@@ -696,6 +696,11 @@ async function tryPrAutomerge(
         await setTimeout(mergeDelay * attempt ** 2); // exponential backoff
       }
 
+      // Use the /merge "quick action" to automerge, which will work for
+      // both merge trains or non-train merges
+      // Associated docs:
+      //  - https://docs.gitlab.com/api/notes/#create-new-merge-request-note
+      //  - https://docs.gitlab.com/user/project/quick_actions/
       await gitlabApi.postJson(
         `projects/${config.repository}/merge_requests/${pr}/notes`,
         {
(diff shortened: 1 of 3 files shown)


Boomtokn (approved):

rarkins (commented):
Merge trains are only supported for premium GitLab customers. Would these changes mean that Renovate automerge no longer works for free users?

pallost:
We've had some issues in our project that when a merge request is created, then in the first seconds if someone presses the "Merge" button then the merge request gets merged immediately, without actually running the pipeline. This is a known gitlab bug. I hope that using the `/merge` quick action immediately when an MR is created wouldn't trigger this same bug. Would it make sense to put a few seconds of wait between creating the MR and triggering the `/merge` quick action to stay on the safe side?

rarkins:
If this is certain to work perfectly for everyone then it's OK to merge. But if there's unknowns, or a chance that it might not work for all, then it should be configurable.

rarkins:
OK, so one way forward is to merge this as-is but revert quickly if we find any problems, and change it to be opt-in - or at least opt-out.

lib/modules/platform/gitlab/index.ts
```diff
@@ -690,28 +696,15 @@ async function tryPrAutomerge(
...
-        } catch (err) {
-          logger.debug(
-            { err },
-            `Automerge on PR creation failed. Retrying ${attempt}`,
-          );
-        }
-        await setTimeout(mergeDelay * attempt ** 2); // exponential backoff
-      }
+      await gitlabApi.postJson(
+        `projects/${config.repository}/merge_requests/${pr}/notes`,
```
viceice, line 700: which is the minimum gitlab version to support this API?

does this API never return 405? like the merge API?

rarkins: Even if `/notes` has been there for years, how about the inner `/merge` logic - also years?

rarkins: image
Yes it seems very old

lib/modules/platform/gitlab/index.ts
```diff
@@ -690,28 +696,15 @@ async function tryPrAutomerge(
...
-          break;
-        } catch (err) {
-          logger.debug(
-            { err },
-            `Automerge on PR creation failed. Retrying ${attempt}`,
-          );
-        }
-        await setTimeout(mergeDelay * attempt ** 2); // exponential backoff
-      }
+      await gitlabApi.postJson(
```
rarkins, line 699:
```suggestion
      // Use the /merge "quick action" to automerge, which will work for
      // both merge trains or non-train merges
      // Associated docs:
      //  - https://docs.gitlab.com/api/notes/#create-new-merge-request-note
      //  - https://docs.gitlab.com/user/project/quick_actions/
      await gitlabApi.postJson(
```

rarkins (approved):

secustor (approved):

simonoscr:
I am not sure if this is the cause of my issue, but since this version, Renovate has been merging merge requests in GitLab without waiting for the pipeline to succeed.

If I understand this correctly, because Renovate now uses the `/notes` API endpoint and no longer uses the `/merge` API endpoint, it can no longer set `merge_when_pipeline_succeeds`?
And thus it no longer does something like this:

`Renovate enabled an automatic merge when the pipeline for [c45d8433] succeeds` 

in merge requests. 

This behaviour has to be configured now directly in all of my GitLab repositories?

rarkins:
This change will be reverted ASAP

rarkins:
I promise to bookmark this PR and refer to it every time in future a user tries assuring us that something is totally tested and safe and doesn't need a feature flag

rarkins:
A new release with this change removed is being built right now

simonoscr:
As far as I understand it, the `/merge` quick action does NOT enable the `merge_when_pipeline_succeeds` option. It states clearly that it is `Depending on the project setting`. So in my gitlab projects, `Merge changes when the pipeline succeeds` is not enabled in the project settings.

With the `/merge` API endpoint, Renovate sends this in the API call body:
```
body: {
  should_remove_source_branch: true,
  merge_when_pipeline_succeeds: true,
},
```

With your change,  Renovate bot no longer sets and can set `merge_when_pipeline_succeeds: true`. Instead, renovate now depends entirely on the gitlab project setting.

The problem for me is, that this change moves the control over the `merge_when_pipeline_succeeds` option from renovate itself, to gitlab.

Now, I am forced to enable `merge_when_pipeline_succeeds` for all pipelines in my project. But what if I only want to prevent Renovate from auto-merging MRs that don’t have a green pipeline?


Currently, I have no merge checks enabled in my project settings. I fully relied on Renovate setting `merge_when_pipeline_succeeds` by itself. So I never had to worry about it.

simonoscr:
yes

simonoscr:
sounds like I am the odd one here? I always thought this was a built-in feature/failsafe to prevent Renovate from accidentally merging an MR without a successful pipeline?

pallost:
Don't give up on this feature yet! I was crying with joy that finally all the automerge MRs are getting merged without me needing to retry them over and over again :D would it make sense to hide it behind a feature flag so that different projects could test it extensively and send feedback on specific corner cases (like the one mentioned above)?

rarkins:
It will need a new config option called something like gitLabNotesMerge to go alongside gitLabIgnoreApprovals

echarrod:
> With your change, Renovate bot no longer sets and can set merge_when_pipeline_succeeds: true. Instead, renovate now depends entirely on the gitlab project setting.

Renovate depends entirely on the gitlab project setting for merging when merge trains _aren't_ enabled. 

I guess since it might catch people off-guard, like @simonoscr above, I agree with putting it behind a feature flag 👍

rarkins:
BTW it's also possible to:
- Use notes automatically if merge trains are _detected_
- Use old approach if no merge trains, unless the new option is selected

pallost:
FYI, we kept on using 39.176.0 version that uses the Notes API for merging but yesterday night it automerged a change without actually running the pipeline. So I assume this has the same issue as with the gitlab UI: if you press the merge button too fast it merges without running the pipeline.

pallost:
I'm sorry false alarm, apparently "pipelines must succeed" wasn't enabled even though I thought it was 😅. However the gitlab UI still made us wait for the pipelines to run and succeed before merging, so I don't know what difference that setting makes 🤔

Issue #33290, opened 2024-12-26
Implement PR cache for all platforms


RahulGautamSingh:
### Describe the proposed change(s).

PR cache has been implemented for Github, Gitea and Bitbucket Cloud. Based on these implementations also cover the remaining platforms.

felipecrs:
FYI, I unintentionally implemented PR cache for Gerrit:

- https://github.com/renovatebot/renovate/pull/35676

But I was not aware of this initiative, I did it my own way (which I should probably revisit).

Issue #3903, opened 2019-06-07
Support changelogs from bitbucket-server


k2snowman69:
**What would you like Renovate to be able to do?**
Similar to how renovate pulls from github.com to retrieve existing changelogs, it would be great if it did similar for on-premise versions of bitbucket-server

**Describe the solution you'd like**
Input - A bitbucket-server repository with a changelog.md at the root
Output - When a PR is created, it uses that changelog to improve the PR body created

**Describe alternatives you've considered**
I thought about other locations where we maybe able to get the changelog from such as unpkg or artifactory if it was part of the npm package but I feel that it adds more configuration changes and dependencies than less

**Additional context**
N/A

k2snowman69:
Haven't done much investigation into this but I saw the one for gitlab and maven so I figured we should open one for bitbucket-server as well

rarkins:
No, it would be a separate task because the APIs are quite different

Zenedith:
Should [this](https://github.com/renovatebot/renovate/pull/7000) help?

viceice:
@Zenedith no, as that is only for shared presets

before we can start working on this, the changelog retrival code must be refctored, so we can add more platofrms more easy

jaysiee:
Any chance someone less experienced could contribute to the refactoring? If there is enough interest, I'd be happy to help.

k2snowman69:
Since no one else chimed in yet... I'd say it depends on what you mean by "less experienced" but as the old adage goes, "you never know until you try".

From my personal experience creating my first PR a few months back, I'd say the biggest hurdle is just getting this project running in a way that you can debug it in a way that's comfortable for you. Everyone has their own coding style so getting this project to run with your personal style along with in a way that doesn't screw up a project in actual use.

After that though, I found it pretty easy to probe around the code base and figure out what's going on where.

jaysiee:
Thanks for your feedback. I'm fairly opimistic that I'll get the project up and running. The part I that threw me off was the required refactoring and whether this is realistically something that can be done by someone new to the project (not knowing what the maintainers have in mind). 

I've had a quick look and currently the release-notes part is branching at multiple points based on whether it should work for github or gitlab. Since it is not intended to simply add another conditional branch for bitbucket-server, maybe we could sketch out the code structure that is required for a more general approach to fetching changelogs and then someone less involved could give it a shot. 

[...]

PR #34114, opened 2025-02-09
feat(bitbucket-server): add paginate http option

Churro:
## Changes

Adds a `paginate` http option for Bitbucket Server, analogous to what exists for other supported platforms. 
It doesn't change any existing request behavior. 

In a follow-up PR, I intend to replace all [utils.accumulateValues()](https://github.com/renovatebot/renovate/blob/dbabbdc5b3ee230eff277916c72f7c6bc75559c0/lib/modules/platform/bitbucket-server/utils.ts#L75) calls and then remove it.

I verified that it does what it should with tests on real repos and ensured it adheres to the [docs on paged APIs](https://docs.atlassian.com/bitbucket-server/rest/7.21.0/bitbucket-rest.html#paging-params).

## Context

- https://github.com/renovatebot/renovate/pull/22135
- https://github.com/renovatebot/renovate/issues/33290
- https://github.com/renovatebot/renovate/issues/3903

### DIFF
--- a/lib/util/http/bitbucket-server.ts
+++ b/lib/util/http/bitbucket-server.ts
@@ -53,7 +53,7 @@ export class BitbucketServerHttp extends Http<BitbucketServerHttpOptions> {
     );
 
     if (opts.paginate && isPagedResult(result.body)) {
-      const collectedValues: T[] = [...result.body.values];
+      const collectedValues = [...result.body.values];
       let nextPageStart = result.body.nextPageStart;
 
       while (nextPageStart) {
--- a/lib/util/http/bitbucket-server.ts
+++ b/lib/util/http/bitbucket-server.ts
@@ -1,30 +1,79 @@
-import { resolveBaseUrl } from '../url';
+import is from '@sindresorhus/is';
+import { logger } from '../../logger';
+import { parseUrl, resolveBaseUrl } from '../url';
 import type { HttpOptions, HttpResponse, InternalHttpOptions } from './types';
 import { Http } from '.';
+
+const MAX_LIMIT = 100;
 
 let baseUrl: string;
 export const setBaseUrl = (url: string): void => {
   baseUrl = url;
 };
 
-export class BitbucketServerHttp extends Http {
+export interface BitbucketServerHttpOptions extends HttpOptions {
+  paginate?: boolean;
+  limit?: number;
+}
+
+interface PagedResult<T = unknown> {
+  nextPageStart?: number;
+  values: T[];
+}
+
+export class BitbucketServerHttp extends Http<BitbucketServerHttpOptions> {
   constructor(options?: HttpOptions) {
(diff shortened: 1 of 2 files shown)


lib/util/http/bitbucket-server.ts
```diff
@@ -1,30 +1,79 @@
...
+
+    const resolvedUrl = parseUrl(resolveBaseUrl(baseUrl, path));
+    if (!resolvedUrl) {
+      logger.error({ path }, 'Bitbucket Server: cannot parse path');
+      throw new Error(`Bitbucket Server: cannot parse path ${path}`);
+    }
+
+    if (opts.paginate) {
+      const limit = opts.limit ?? MAX_LIMIT;
+      resolvedUrl.searchParams.set('limit', limit.toString());
```
rarkins, line 47: is that a page limit or results count limit?

lib/util/http/bitbucket-server.ts
```diff
@@ -1,30 +1,79 @@
...
-    return super.request<T>(url, opts);
+
+    const resolvedUrl = parseUrl(resolveBaseUrl(baseUrl, path));
+    if (!resolvedUrl) {
+      logger.error({ path }, 'Bitbucket Server: cannot parse path');
+      throw new Error(`Bitbucket Server: cannot parse path ${path}`);
+    }
+
+    if (opts.paginate) {
+      const limit = opts.limit ?? MAX_LIMIT;
```
rarkins, line 46: Is `MAX_LIMIT` newly added? I don't see it when searching for it in existing source code.

lib/util/http/bitbucket-server.ts
rarkins: Thanks I'm fine with this `limit` approach here

lib/util/http/bitbucket-server.ts
Churro: The existing implementation defines it here:
https://github.com/renovatebot/renovate/blob/858f26b58670e732609367e6b23ebeab3b1080c6/lib/modules/platform/bitbucket-server/utils.ts#L41

lib/util/http/bitbucket-server.ts
```diff
@@ -1,30 +1,79 @@
...
+      resolvedUrl.searchParams.set('limit', limit.toString());
+    }
+
+    const result = await super.request<T | PagedResult<T>>(
+      resolvedUrl.toString(),
+      opts,
+    );
+
+    if (opts.paginate && isPagedResult(result.body)) {
+      const collectedValues: T[] = [...result.body.values];
```
viceice, line 56:
```suggestion
      const collectedValues = [...result.body.values];
```
I think we don't need this extra type

viceice (commented):
otherwise LGTM

viceice (approved):

PR #34137, opened 2025-02-10
feat(config/self-hosted): `processEnv`

### LINKED ISSUE #34132: New global config option "processEnv"

Add a new global config option `processEnv`, which is an object accepting key/value strings. It is to be used in file config only, so no need to support it for env or CLI parsing.

Its purpose is to provide a way of supplying env variables to Renovate without putting them in actual env. When combined with `deleteConfigFile`, this is more secure than using actual env.

After file parsing, keys/values in `processEnv` should be copied into `process.env` and then deleted - not to be combined with the rest of config parsing. The list of keys being exported should be logged in a single debug message, but the values never logged.

RahulGautamSingh:
## Changes
- Adds new self-hosted config option `processEnv`

## Context
- Closes: https://github.com/renovatebot/renovate/issues/34132

- [x] Both unit tests + ran on a real repository (tested locally)

### DIFF
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -1879,6 +1879,14 @@ const options: RenovateOptions[] = [
     default: 'strict',
   },
   {
+    name: 'processEnv',
+    description: 'Environment variables to be used in global config only.',
+    type: 'object',
+    default: {},
+    globalOnly: true,
+    stage: 'global',
+  },
+  {
     name: 'prCreation',
     description: 'When to create the PR for a branch.',
     type: 'string',
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -1885,6 +1885,9 @@ const options: RenovateOptions[] = [
     default: {},
     globalOnly: true,
     stage: 'global',
+    additionalProperties: {
+      type: 'string',
+    },
   },
   {
     name: 'prCreation',
--- a/lib/config/types.ts
+++ b/lib/config/types.ts
@@ -127,6 +127,7 @@ export interface GlobalOnlyConfig {
   repositories?: RenovateRepository[];
   useCloudMetadataServices?: boolean;
   deleteConfigFile?: boolean;
+  processEnv?: Record<string, string>;
 }
 
 // Config options used within the repository worker, but not user configurable
--- a/lib/config/types.ts
+++ b/lib/config/types.ts
@@ -127,7 +127,6 @@ export interface GlobalOnlyConfig {
   repositories?: RenovateRepository[];
   useCloudMetadataServices?: boolean;
   deleteConfigFile?: boolean;
-  processEnv?: Record<string, string>;
 }
 
 // Config options used within the repository worker, but not user configurable
--- a/lib/workers/global/config/parse/file.ts
+++ b/lib/workers/global/config/parse/file.ts
@@ -6,6 +6,7 @@ import type { AllConfig, RenovateConfig } from '../../../../config/types';
 import { logger } from '../../../../logger';
 import { parseJson } from '../../../../util/common';
 import { readSystemFile } from '../../../../util/fs';
+import { regEx } from '../../../../util/regex';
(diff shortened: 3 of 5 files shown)


lib/config/options/index.ts
```diff
@@ -1878,6 +1878,14 @@ const options: RenovateOptions[] = [
     allowedValues: ['strict', 'flexible', 'none'],
     default: 'strict',
   },
+  {
+    name: 'processEnv',
+    description: 'Environment variables to be used in global config only.',
+    type: 'object',
```
rarkins, line 1884: Probably needs a subtype defined

lib/workers/global/config/parse/file.ts
```diff
@@ -78,6 +79,22 @@ export async function getConfig(env: NodeJS.ProcessEnv): Promise<AllConfig> {
     logger.debug('No config file found on disk - skipping');
   }
 
+  if (is.nonEmptyObject(config.processEnv)) {
+    // should begin and end with capital alphabets + use underscore as separator
+    const keyRegex = regEx(/^[A-Z]+(_[A-Z]+)*$/);
```
rarkins, line 84: Where is this from? Leave a reference to the spec

lib/workers/global/config/parse/file.ts
```diff
@@ -78,6 +79,22 @@ export async function getConfig(env: NodeJS.ProcessEnv): Promise<AllConfig> {
...
+    const exportedKeys = [];
+    for (const [key, value] of Object.entries(config.processEnv)) {
+      if (keyRegex.test(key) && is.nonEmptyString(value)) {
+        exportedKeys.push(key);
+        process.env[key] = value;
+      } else {
+        logger.debug({ key, value }, 'Could not export key to environment');
+      }
```
rarkins, line 92: Use separate checks, separate error messages, and never log the value

rarkins (changes requested):

lib/workers/global/config/parse/file.ts
RahulGautamSingh: I don't think there's a spec/ref to this. I took it from the env names we used for our config options.
like `labels` -> `RENOVATE_LABELS` : separated by `_`and all capitals

RahulGautamSingh: Just realized that if users use their own prefix instead of `RENOVATE_`, then they could include lowercase letter as well.

Don't think checking the key name makes sense now.

rarkins: This feature is not intended for renovate's own env. It's intended for other env so should allow all valid env variable name formats

rarkins (changes requested):

Boomtokn (approved):

lib/config/types.ts
```diff
@@ -127,6 +127,7 @@ export interface GlobalOnlyConfig {
   repositories?: RenovateRepository[];
   useCloudMetadataServices?: boolean;
   deleteConfigFile?: boolean;
+  processEnv?: Record<string, string>;
```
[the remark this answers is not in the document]
RahulGautamSingh: TS was showing error in my case , which disaapeared after I closed & re-opened VS Code :)

type-check passes for me too.

rarkins (approved):

nabeelsaabna (commented):
I see that `...process.env` is used only in a few `git` commands (not managers..)
and also with `exposeAllEnv` as expected

rarkins:
Can we test that the AWS libraries pick it up?

nabeelsaabna:
verified
image

nabeelsaabna (approved):

nabeelsaabna:
@rarkins do we want `exposeAllEnv` to also expose `processEnv` ?

rarkins:
I'm not sure. I don't think we plan on exposing all env in our hosted app anyway do we? It may be hard to block without some complicated logic to remember the keys

viceice (approved):

PR #34154, opened 2025-02-11
refactor(bazel-module): use bazel syntactic concepts in parser

gzm0:
## Changes

Instead of using a general purpose "record" type, we now parse into rules and extension tags (see comments in the relevant files for a bazel primer of what these are).

This change allows us to de-duplicate some of the parsing functionality and hopefully makes it easier to add support for more bazel packages. Notably, it should be possible to add support for a new bazel rule without having to understand parsing.

As an added benefit, this centralizes the way we recognize extensions, so drift between different types of extensions is less likely.

## Context

This refactor was triggered by #33670. To fix that, we need to keep the original syntactic strings during parsing. This is made easier by this cleanup.

### DIFF
--- a/lib/modules/manager/bazel-module/parser/extension-tags.ts
+++ b/lib/modules/manager/bazel-module/parser/extension-tags.ts
@@ -1,0 +1,50 @@
+import { query as q } from 'good-enough-parser';
+import { regEx } from '../../../../util/regex';
+import type { Ctx } from '../context';
+import { kvParams } from './common';
+
+import { mavenExtensionPrefix, mavenExtensionTags } from './maven';
+import { ociExtensionPrefix, ociExtensionTags } from './oci';
+
+// In bazel modules an extension tag is (roughly) a "member function application".
+// For example:
+//
+//     oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")
+//     ^^^ --> the extension definition (not parsed by this module)
+//
+//     oci.pull(<parameters>)
+//         ^^^^ --> the extension tag
+//
+// The name of the extension (`oci` in the example above) technically arbitrary.
+// However, in practice, there are conventions. We use this to simplify parsing
+// by assuming the extension names start with well-known prefixes.
+
+const supportedExtensionRegex = regEx(
+  `^(${ociExtensionPrefix}|${mavenExtensionPrefix}).*$`,
+);
+
+const supportedExtensionTags = [...mavenExtensionTags, ...ociExtensionTags];
+
+const supportedExtensionTagsRegex = regEx(
+  `^(${supportedExtensionTags.join('|')})$`,
+);
+
+export const extensionTags = q
+  .sym<Ctx>(supportedExtensionRegex, (ctx, token) => {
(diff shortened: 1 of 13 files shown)


rarkins (commented):
Impressive PR! This is a little "high risk" though because both code and tests are changing together, which can unfortunately mask regressions. @zharinov please take a careful look

gzm0:
> This is a little "high risk" though because both code and tests are changing together, which can unfortunately mask regressions.

Yes, some of this was unfortunately unavoidable.

Please do note though, that the overall extraction tests (https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/bazel-module/extract.spec.ts) do not change.

So we do have reasonable confidence that `extractPacakgeFile` for the manager has not changed. I'm happy to add tests there in another PR to be merged before this one, if you feel they are too weak.

Boomtokn (approved):

lib/modules/manager/bazel-module/parser/extension-tags.ts
```diff
@@ -0,0 +1,50 @@
...
+//
+// The name of the extension (`oci` in the example above) technically arbitrary.
+// However, in practice, there are conventions. We use this to simplify parsing
+// by assuming the extension names start with well-known prefixes.
+
+const supportedExtensionRegex = regEx(
+  `^(${ociExtensionPrefix}|${mavenExtensionPrefix}).*$`,
+);
+
+const supportedExtensionTags = [...mavenExtensionTags, ...ociExtensionTags];
```
fmeum, line 26: It would be cleaner to handle tags per extension rather than taking the union over all tags.

gzm0: Ah yes, I admit I have somewhat cargo-culted / not touched the `supportedX` regular expression concept.

The "true" definition of extension tags we handle are in the zod schemas for the actual package rules. For example for rules_oci:
https://github.com/renovatebot/renovate/blob/8806078e291f70b5591fc0365481c06d9b909549/lib/modules/manager/bazel-module/parser/oci.ts

My assumption here was that in parsing, the regular expressions exist for efficiency: There is no point in attempting to parse and load an entire rule / extension tag / sub-tree, if we are not even interested in it.

So in that sense, it is OK if this is an over-approximation. My `good-enough-parser` knowledge is limited, so I'm not sure we can have that cleaner parsing without duplicating larger parts of the extension tag logic.

Maybe a Renovate maintainer can chime in here for some insight in how this is usually handled in Renovate.
I will also try to look at how other managers handle this and I will also try and ask on the Renovate slack.

zharinov: > My assumption here was that in parsing, the regular expressions exist for efficiency: There is no point in attempting to parse and load an entire rule / extension tag / sub-tree, if we are not even interested in it.

Yes, it's correct, such regex checks are meant to limit the unnecessary tree walks by a parser.

zharinov (approved):

rarkins (approved):

PR #34160, opened 2025-02-11
feat: Allow using `packageFile` in regex manager templates

erezrokah:
## Changes

This PR adds support for using `packageFile` as a template variable in the custom regex manager.
This shouldn't be breaking if someone is already using `packageFile` as a named match since it the matching groups are spread after the newly added `packageFile`.
This could impact users if they are using `packageFile` in one of their templates without any matches, and now they will have that value injected.

## Context

I was expecting this to work based on https://github.com/renovatebot/renovate/discussions/13791#discussioncomment-2045123, to found out it doesn't due to trial and error.
The use case is having multiple version files structured as below:

```
component-1-version.txt
component-2-version.txt
```

Where each file has the version of the component, and the filename is the name of the dependency.

Not sure if docs are needed, I couldn't find a clear answer from the [docs](https://docs.renovatebot.com/templates/) which variables can be used for each template.


erezrokah:
Opened https://github.com/renovatebot/renovate/discussions/34161

PR #34212, opened 2025-02-13
test(bazel-module): remove Ctx success case tests

gzm0:
## Changes

Remove success path tests for bazel-module parser context.

## Context

The success paths of Ctx are already fully tested via the parser tests (parser/index.spec.ts).

This is triggered by #33670, for which we need to adjust parsing to retain the raw original strings so we can provide them as `replaceString` in the `PackageDepencency`.

This in turn requires adjustments to the context that become prohibitively hard to test at such a low level.

Since (as coverage shows), these tests are not necessary, we simply remove them.


gzm0:
I was a bit concerned by the coverage run in CI not triggering, which would make this PR quite high risk.

I have attached a local coverage result of `pnpm jest ./lib/modules/manager/bazel-module/*` showing that the `bazel-module` subtree still has 100% coverage after the changes in this PR.

My hope is that coverage runs as soon as I click "Ready for Review", but let's see :)

[coverage.tar.gz](https://github.com/user-attachments/files/18794335/coverage.tar.gz)

viceice (approved):

PR #34205, opened 2025-02-13
fix(lib/config/decrypt): support azure collection matching

nabeelsaabna:
## Changes

For azure the higher level org (collection) is part of the `endpoint` and encrypted tokens can't be created for a full org  

example:
```
platform: azure
endpoint: https://dev.azure.com/nabeelsaabna
repository: renovate-test/token-decrypt
```
| org    | repo | current | new |
| -------- | ------- | ------- | ------- |
| `renovate-test`  | `token-decrypt`   | match  | match |
| `renovate-test` |  | match | match |
| `nabeelsaabna/renovate-test`  | `token-decrypt`  | no match| match |
| `nabeelsaabna/renovate-test`  |  |no match| match |
| `nabeelsaabna`  |  | no match| match |
| `nabeelsaabna/*`  | | no match| match |

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [x] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

PR #34215, opened 2025-02-13
fix(lib/config/decrypt): support azure collection matching

nabeelsaabna:
## Changes

For azure the higher level org (collection) is part of the `endpoint` and encrypted tokens can't be created for a full org  

example:
```
platform: azure
endpoint: https://dev.azure.com/nabeelsaabna
repository: renovate-test/token-decrypt
```
| org    | repo | current | new |
| -------- | ------- | ------- | ------- |
| `renovate-test`  | `token-decrypt`   | match  | match |
| `renovate-test` |  | match | match |
| `nabeelsaabna/renovate-test`  | `token-decrypt`  | no match| match |
| `nabeelsaabna/renovate-test`  |  |no match| match |
| `nabeelsaabna`  |  | no match| not match |
| `nabeelsaabna/*`  | | no match| match |

**updated**

## Context

recreated for CLA and email issue https://github.com/renovatebot/renovate/pull/34205


lib/config/decrypt.spec.ts
```diff
@@ -53,4 +53,136 @@ describe('config/decrypt', () => {
...
+      it.each`
+        str                                                 | repo             | expected
+        ${'{"o":"abcd",         "r":"",     "v":"123#'}     | ${'abcd/edf'}    | ${null}
+        ${'{"o":"abcd",         "r":"",     "v":""}'}       | ${'abcd/edf'}    | ${null}
+        ${'{"o":"",             "r":"",     "v":"val"}'}    | ${'abcd/edf'}    | ${null}
+        ${'{"o":"abcd",         "r":"edf",  "v":"val-1"}'}  | ${'abcd/edf'}    | ${'val-1'}
+        ${'{"o":"abcd",         "r":"",     "v":"val-2"}'}  | ${'abcd/edf'}    | ${'val-2'}
+        ${'{"o":"abcd/fgh",     "r":"ef",   "v":"val-3"}'}  | ${'abcd/fgh/ef'} | ${'val-3'}
+        ${'{"o":"abcd/fgh",     "r":"",     "v":"val-4"}'}  | ${'abcd/fgh/ef'} | ${'val-4'}
+        ${'{"o":"a/b/c/d",      "r":"ef",   "v":"val-5"}'}  | ${'a/b/c/d/ef'}  | ${'val-5'}
```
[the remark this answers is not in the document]
nabeelsaabna: I know GitHub and Bitbucket have 2 levels and azure has 3 levels, but not sure about Gitlab and other platforms!!
These are function level tests that cover a variety of input/output cases even if some might never happen.
Like the invalid endpoint tests...

lib/config/decrypt.ts
```diff
@@ -77,6 +77,17 @@ function validateDecryptedValue(
...
     }
 
+    const repositories = [repository.toUpperCase()];
+
+    const azureCollection = getAzureCollection();
+    if (is.nonEmptyString(azureCollection)) {
+      // used for full 'org/project/repo' matching
+      repositories.push(`${azureCollection}/${repository}`.toUpperCase());
+      // used for org prefix matching without repo
+      repositories.push(azureCollection.toUpperCase());
```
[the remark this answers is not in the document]
nabeelsaabna: The `/*/` is repo-integrations specific, this one is the standard org name only.
I thought it's better to support it in Renovate rather than adding a special case documentation

lib/config/decrypt.spec.ts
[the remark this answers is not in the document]
rarkins: I had assumed these were azure tests!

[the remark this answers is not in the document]
nabeelsaabna: `describe('platforms non azure', () => {`

lib/config/decrypt.spec.ts
```diff
@@ -53,4 +53,136 @@ describe('config/decrypt', () => {
...
+          ${'{"o":"any",           "r":"",     "v":"wrong-123#'} | ${'fgh/rp1'} | ${null}
+          ${'{"o":"any",           "r":"",     "v":""}'}         | ${'fgh/rp1'} | ${null}
+          ${'{"o":"",              "r":"",     "v":"any"}'}      | ${'fgh/rp1'} | ${null}
+          ${'{"o":"fgh",           "r":"rp1",  "v":"zv-1"}'}     | ${'fgh/rp1'} | ${'zv-1'}
+          ${'{"o":"fgh",           "r":"",     "v":"zv-2"}'}     | ${'fgh/rp1'} | ${'zv-2'}
+          ${'{"o":"az123/fgh",     "r":"rp1",  "v":"zv-3"}'}     | ${'fgh/rp1'} | ${'zv-3'}
+          ${'{"o":"az123/fgh",     "r":"",     "v":"zv-4"}'}     | ${'fgh/rp1'} | ${'zv-4'}
+          ${'{"o":"az123/*",       "r":"",     "v":"zv-5"}'}     | ${'fgh/rp1'} | ${'zv-5'}
+          ${'{"o":"az123/",        "r":"",     "v":"zv-6"}'}     | ${'fgh/rp1'} | ${'zv-6'}
+          ${'{"o":"az123",         "r":"",     "v":"zv-7"}'}     | ${'fgh/rp1'} | ${'zv-7'}
```
rarkins, line 99: I didn't expect this test. It seems essentially the same as the first line, yet it returns true. In other words, is it intentional that if `o` is defined, then it can match the org OR the project name?

nabeelsaabna: the `/*/` is temporary for Azure repo-integrations
I prefer to support the correct (Renovate) way of scoping a token, once the hosted Azure repo-integrations is deprecated we should remove the token matching project without org

lib/config/decrypt.ts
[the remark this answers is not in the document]
nabeelsaabna: posting the same explanation from previous comment 
> the /*/ is temporary for Mend Azure repo-integrations
> I prefer to support the correct (Renovate) way of scoping a token, once the hosted Azure repo-integrations is deprecated we should remove the token matching project without org

lib/config/decrypt.spec.ts
rarkins: Are you sure we can really do this, considering there's others in the wild using `o`/`r` today.

I don't like it if users have encrypted secrets using `o` to mean "project name" and now we're widening the decrypting to allow them to be decrypted if `o` matches the endpoint org name too.

I don't mind the test above `az123/` - assuming that it would fail to match a project named `az123` today - but have a bad feeling about doubling up the meaning of "plain" `o` values.

nabeelsaabna: The issue is only related to NO Repo in the encrypted token keys

| #                    | collection | project  | repo | encrypted org value | current      | new       |
| :---------------- | :------:          | :------:     |:------: | :------:                           | :------:          | ----:        |
| 1                    | abc             | foo        |           | foo                              | match        | match   |
| 2                    | abc             | foo        |           | abc/*                          | no match   | match   |
| 3                    | abc             | foo        |           | abc/foo                      | no match   | match   |
| 4                    | abc             | foo        |           | abc                              | no match   | ??          | 
| 5                    | abc             | foo        |           | abc/                            | no match    | ??          | 

so you want to keep the first match of "only project" although its not unique and prevent the last 2 example of "clean collection name" without `/*`  at the end ?
the ability to "attack" the encrypted token is caused by the first example `#1` (existing) behavior of project matching and not the new behavior of unique collection name! 

Note: we are keeping `#1` only due to backward compatibility 

regardless, if you wish so I can edit the PR to exclude the last 2 use cases

rarkins: I think the last two confuse things, so if they're not _required_ for other reasons, such as compatibility with other Mend products, then drop them. Especially because you can achieve the same thing with (2).

nabeelsaabna: implemented the change
because we need to support `#3` its required to explicitly ignore `#4` and `#5`

viceice:
please force push to old branch next time, so we don't loose review history

lib/config/decrypt.ts
```diff
@@ -209,3 +228,29 @@ Refer to migration documents here: https://docs.renovatebot.com/mend-hosted/migr
...
+  if (endpointUrl === null) {
+    // should not happen
+    logger.warn({ endpoint }, 'Unable to parse endpoint for token decryption');
+    return undefined;
+  }
+
+  const azureCollection = endpointUrl.pathname
+    ?.split('/')
+    .filter(is.nonEmptyString)
+    .join('/');
```
viceice, line 249: why do we need to split and join here at all?

lib/config/decrypt.ts
```diff
@@ -209,3 +228,29 @@ Refer to migration documents here: https://docs.renovatebot.com/mend-hosted/migr
...
+  if (!is.nonEmptyString(azureCollection)) {
+    logger.warn({ endpoint }, 'Unable to find azure collection name from URL');
```
viceice, line 251: we should not warn for self hosted devops server. they don't have azure collections

viceice (changes requested):

lib/config/decrypt.ts
nabeelsaabna: downgraded to debug

lib/config/decrypt.ts
nabeelsaabna: to remove the slash at the start and at the end
e.g
* `https://dev.azure.com/` the URL pathname holds `/`
* `https://dev.azure.com/aaa/bbb/` the URL pathname holds `/aaa/bbb/`

viceice: use `trimSlashes` url util? i think `endpointUrl.pathname` should never undefined or null here?

rarkins (dismissed):

lib/config/decrypt.spec.ts
```diff
@@ -53,4 +57,190 @@ describe('config/decrypt', () => {
...
+            '{"o":"col/*", "r":"", "v":"any"}',
+            'proj/repo',
+          ),
+        ).toBeNull();
+      });
+
+      it('endpoint URL without collection', () => {
+        GlobalConfig.set({
+          platform: 'azure',
+          endpoint: 'https://dev.azure.com/',
```
viceice, line 162: test with custom domain?  eg `https://domain.test/some/devops-server/` and  `https://devops.domain.test/`

viceice (commented):
otherwise LGTM

viceice (dismissed):

lib/config/decrypt.spec.ts
nabeelsaabna: I tried to lookup details on azureDevOps self-host and couldn't find much details/examples.. 

`https://domain.test/some/devops-server/` will result in `pathname: some/devops-server` which will be added as top-level org/collection name and will affect the matching. 
In case `https://domain.test/some/devops-server/` this is a valid server name then I have to take into account and hardcod a check for `https://dev.azure.com/`

nabeelsaabna: added handling and tests

rarkins (approved):

viceice (approved):

PR #34218, opened 2025-02-13
feat(manager/composer): support updates with minimal changes

### LINKED ISSUE #34147: Support `--minimal-changes` CLI option for Composer >= 2.7
### Discussed in https://github.com/renovatebot/renovate/discussions/34140

<sup>Originally posted by **ste93cry** February 10, 2025</sup>
### Tell us more.

When Renovate updates a Composer dependency, it passes the `--with-dependencies` flag to the package manager, which also updates the dependencies of the packages in the argument list, unless the dependency itself is a root requirement. As a result, even when only the root package is being updated, its transitive dependencies are all updated in turn to the latest possible versions.

To solve this problem, while still retaining the flexibility of bumping the transitive dependencies if there's a strict need for, the `--minimal-changes` option has been added in [Composer `2.7`](https://github.com/composer/composer/pull/11665). To be fair, it's possible to set the `COMPOSER_MINIMAL_CHANGES=1` environment variable to get the same result, but using this option by default would reduce the scope of the changes in a merge request to the strict necessary, which is a good thing imho.

In [this](https://github.com/ste93cry/renovate-repro-34140/pull/3) little repro repository, you can see the amount of changes proposed to update one root requirement. You can try to run `composer update open-telemetry/sdk:1.2.2 --with-dependencies --minimal-changes` to see how it would look like instead with the proposed solution.</div>

ste93cry:
## Changes

I'm implementing a new function in the Composer package manager that, depending on the installed version, will append the `--minimal-changes` option to the command to update the dependencies. As the tool constraint can be either a single version or a range, like in the default case where the package manager version is not pinned explicitly, I had to expose the `intersects()` function of the `semver` package, which allows checking whether the given constraint falls under the conditions for which the package manager supports the option.

## Context

Closes #34147


lib/modules/manager/composer/utils.ts
```diff
@@ -46,6 +46,20 @@ export function getComposerArguments(
   return args;
 }
 
+export function getComposerUpdateArguments(
+  config: UpdateArtifactsConfig,
+  toolConstraint: ToolConstraint,
+): string {
+  let args = getComposerArguments(config, toolConstraint);
+
+  // TODO: toolConstraint.constraint can be null or undefined? (#22198)
```
rarkins, line 55: Yes it can be undefined

lib/modules/versioning/types.ts
```diff
@@ -125,6 +125,11 @@ export interface VersioningApi {
    */
   subset?(subRange: string, superRange: string): boolean | undefined;
 
+  /**
+   * Checks whether subRange intersects superRange.
+   */
+  intersects?(subRange: string, superRange: string): boolean;
```
rarkins, line 131: I prefer not to extend our versioning API (which applies to all versioning schemes) just for one narrow use case in one manager

lib/modules/manager/composer/utils.ts
rarkins: If it's not a string, return the args.

lib/modules/versioning/types.ts
ste93cry: I did this because it was done similarly for the `subset()` function used by the Poetry package manager. I evaluated to directly import the `semver` package, but then I would also need to export the [`composer2npm()`](https://github.com/renovatebot/renovate/blob/858f26b58670e732609367e6b23ebeab3b1080c6/lib/modules/versioning/composer/index.ts#L121) function. Does it sound good?

rarkins: Is there any chance that `subset()` can be used here? e.g. passing `2.7.0` or `=2.7.0` might work?

ste93cry: So, there are 4 different cases that can happen:

- The package manager version is pinned by requiring the `composer/composer` package. The version can be a single version or a range and is used as a constraint as is.
https://github.com/renovatebot/renovate/blob/030130ccc94471e44cda80fc4b84e437ad06e339/lib/modules/manager/composer/utils.ts#L90-L94

- The package manager version is pinned by requiring the `composer` package. The version can be a single version or a range and is used as a constraint as is.
https://github.com/renovatebot/renovate/blob/030130ccc94471e44cda80fc4b84e437ad06e339/lib/modules/manager/composer/utils.ts#L96-L99

- The package manager version is not pinned explicitly, so Renovate looks at the `plugin-api-version` prop in the `composer.lock` file. This is always a single version, but the bot transforms it to be a `^X.Y` range.
https://github.com/renovatebot/renovate/blob/030130ccc94471e44cda80fc4b84e437ad06e339/lib/modules/manager/composer/utils.ts#L102-L106

- If no `plugin-api-version` prop is present, Renovate falls back as last resort to the `composer-runtime-api` prop, which again is always a single version, transformed to a `^X.Y` range by the bot.
https://github.com/renovatebot/renovate/blob/030130ccc94471e44cda80fc4b84e437ad06e339/lib/modules/manager/composer/utils.ts#L108-L112

When the version is anything below `2.7`, even if the range uses the caret, `false` is returned by `subset()`:

| Tool Version Constraint | Minimum Required Tool Version | Result |
|-------------------------|-------------------------------|--------|
| `^2.6`                  | `^2.7`                        | `false`  |
| `^2.8`                  | `^2.7`                        | `true`   |
| `2.6.0`                 | `^2.7`                        | `false`  |
| `2.6.9`                 | `^2.7`                        | `false`  |
| `2.7.0`                 | `^2.7`                        | `true`   |
| `2.7.1`                 | `^2.7`                        | `true`   |
| `2.8.0`                 | `^2.7`                        | `true`   |

Instead, using the `intersect()` function yields these results, which are similar but also correctly resolves to the expected output when the caret is used with a version older than `2.7`:

| Tool Version Constraint | Minimum Required Tool Version | Result  |
|-------------------------|-------------------------------|---------|
| `^2.6`                  | `^2.7`                        | `true`  |
| `^2.8`                  | `^2.7`                        | `true`  |
| `2.6.0`                 | `^2.7`                        | `false` |
| `2.6.9`                 | `^2.7`                        | `false` |
| `2.7.0`                 | `^2.7`                        | `true`  |
| `2.7.1`                 | `^2.7`                        | `true`  |
| `2.8.0`                 | `^2.7`                        | `true`  |

In the end, to answer your question, `subset()` is not a viable solution in this case as far as I understand.

lib/modules/manager/composer/utils.ts
```diff
@@ -21,12 +22,16 @@ export function getComposerArguments(
...
+      if (is.string(toolConstraint.constraint)) {
+        const major = api.getMajor(toolConstraint.constraint);
+        const minor = api.getMinor(toolConstraint.constraint);
+
+        args += api.matches(`${major}.${minor}`, '^2.2')
+          ? " --ignore-platform-req='ext-*' --ignore-platform-req='lib-*'"
+          : ' --ignore-platform-reqs';
+      } else {
+        args += ' --ignore-platform-reqs';
+      }
```
viceice, line 34: please do this refactor in a separate PR and use intersects like in your addition

lib/modules/versioning/types.ts
rarkins: OK, good reasoning. Let's go with intersects

lib/modules/manager/composer/utils.ts
ste93cry: I extracted the changes in #34256. Once merged, I will update this pull request accordingly

lib/modules/manager/composer/utils.ts
```diff
@@ -46,6 +47,22 @@ export function getComposerArguments(
...
+  toolConstraint: ToolConstraint,
+): string {
+  let args = getComposerArguments(config, toolConstraint);
+
+  if (
+    is.string(toolConstraint.constraint) &&
+    api.intersects!(toolConstraint.constraint, '^2.7')
+  ) {
+      args += ' --minimal-changes';
+  }
```
rarkins, line 61: If `api.intersects()` can ever throw, it would be best to wrap this block in a try/catch and emit a `logger.warn()` if it ever throws (and then continue with default args)

rarkins (changes requested):

lib/modules/manager/composer/utils.ts
ste93cry: This function now never throws an exception, the `try`/`catch` block is inside it, so there is no need to do anything extra.

lib/modules/manager/composer/utils.ts
```diff
@@ -49,6 +49,22 @@ export function getComposerArguments(
...
 
+export function getComposerUpdateArguments(
+  config: UpdateArtifactsConfig,
+  toolConstraint: ToolConstraint,
+): string {
+  let args = getComposerArguments(config, toolConstraint);
+
+  if (
+    is.string(toolConstraint.constraint) &&
+    api.intersects!(toolConstraint.constraint, '^2.7')
```
viceice, line 60:
```suggestion
    api.intersects!(toolConstraint.constraint, '>=2.7')
```

it should also work for composer v3+

lib/modules/manager/composer/utils.spec.ts
```diff
@@ -316,6 +317,74 @@ describe('modules/manager/composer/utils', () => {
     });
   });
 
+  describe('getComposerUpdateArguments', () => {
+    it('does not request an update with minimal changes with 2.6.0', () => {
```
viceice, line 321: - use each to test multiple versions
- add tests for `>=2.6` and `>=3.0.0`

ste93cry: >use each to test multiple versions

This would make this test inconsistent with all the others of this file. Are you fine with this?

viceice: yes

rarkins (approved):

viceice (approved):

PR #34222, opened 2025-02-14
fix(bazel-module): Add replaceString to oci_pull PackageDependency

gzm0:
## Changes

- Adjust the parser to retain the underlying raw string.
- Return replaceString from `oci_pull` `PackageDependency`

## Context

By returning replaceString, we allow the auto replacer to replace both the tag and the digest.
- #33670 

Depends on #34212.

Pros of this PR:
- Minimal changes required to fix the oci_pull issue.

Cons of this PR:
- Parsing is somewhat inconsistent (only extension tags get `rawString` rules do not).
- The parser abstraction is starting to leak a lot of its internals.

- [x] Both unit tests + ran on a real repository (https://github.com/gzm0/bzlmod-oci-update-test/pull/2/files)


gzm0:
Note to the reviewers: IMHO, this PR is on the lower end of the quality spectrum (see description), but it fixes the issue.

I'm happy to discuss alternatives. One option to make this cleaner is to hide "fragments" entirely inside the parsing abstraction and expose something simpler (no "offset", no "complete", etc.).

rarkins (approved):

PR #34244, opened 2025-02-15
fix(presets): actually group Node.js packages

maxbrunet:
## Changes

Set `groupName` in `group:nodeJs` preset

## Context

The PRs for these packages were not grouped until now :confused:


rarkins (approved):

lib/config/presets/internal/group.ts
```diff
@@ -363,6 +363,7 @@ const staticGroups = {
     packageRules: [
       {
         commitMessageTopic: 'Node.js',
```
viceice, line 365: I think we should drop this line now? 🤔

PR #34256, opened 2025-02-16
fix(manager/composer): use `intersects()` instead of `matches()` to compare tool constraint

ste93cry:
## Changes

Within this pull request, I'm refactoring the `getComposerArguments()` function to make use of the newly added `intersects()` function to compare the tool version constraint.

## Context

While developing an unrelated feature, I realized that the `matches()` function that is being used to compare the tool version constraint in the Composer package manager returns unwanted results when the version is lower than the minimum required one, even if it's using the caret. For example, if the minimum required version is `^2.6` and the tool constraint is `^2.7`, the expected result should be `true`, but `false` is returned instead. The following table shows some examples of what's happening:

| Function  | Tool Version Constraint | Minimum Required Tool Version | Result  |
|-----------|-------------------------|-------------------------------|---------|
| `matches` | `^2.6`                  | `^2.7`                        | `false` |
| `matches` | `^2.8`                  | `^2.7`                        | `true`  |
| `matches` | `2.6.0`                 | `^2.7`                        | `false` |
| `matches` | `2.7.0`                 | `^2.7`                        | `true`  |
| `matches` | `2.8.0`                 | `^2.7`                        | `true`  |

By switching to `intersects()` instead, the intersection of two ranges is handled correctly, as can be seen in the first row of the following table:

|   Function   | Tool Version Constraint | Minimum Required Tool Version | Result  |
|--------------|-------------------------|-------------------------------|---------|
| `intersects` | `^2.6`                  | `^2.7`                        | `true`  |
| `intersects` | `^2.8`                  | `^2.7`                        | `true`  |
| `intersects` | `2.6.0`                 | `^2.7`                        | `false` |
| `intersects` | `2.7.0`                 | `^2.7`                        | `true`  |
| `intersects` | `2.8.0`                 | `^2.7`                        | `true`  |


rarkins (dismissed):

lib/modules/versioning/types.ts
```diff
@@ -125,6 +125,11 @@ export interface VersioningApi {
    */
   subset?(subRange: string, superRange: string): boolean | undefined;
 
+  /**
+   * Checks whether subRange intersects superRange.
+   */
+  intersects?(subRange: string, superRange: string): boolean;
```
viceice, line 131: Do we really need this on public versioning api? This is a very specific use case, so i would suggest to not do this

ste93cry: I did this because it's the same approach used for the `subset()` function. I though about directly calling `semver.intersects()`, but then I would need to export the `composer2npm()` function, which doesn't seem appropriate.

viceice: i've now seen the discussion

https://github.com/renovatebot/renovate/pull/34218/files#r1955589318

viceice (dismissed):

lib/modules/manager/composer/utils.ts
```diff
@@ -21,12 +22,13 @@ export function getComposerArguments(
...
-      const major = api.getMajor(toolConstraint.constraint!);
-      const minor = api.getMinor(toolConstraint.constraint!);
-      args += api.matches(`${major}.${minor}`, '^2.2')
-        ? " --ignore-platform-req='ext-*' --ignore-platform-req='lib-*'"
-        : ' --ignore-platform-reqs';
+      if (is.string(toolConstraint.constraint)) {
+        args += api.intersects!(toolConstraint.constraint, '^2.2')
+          ? " --ignore-platform-req='ext-*' --ignore-platform-req='lib-*'"
+          : ' --ignore-platform-reqs';
+      } else {
```
viceice, line 29:
```suggestion
      if (is.string(toolConstraint.constraint) && api.intersects!(toolConstraint.constraint, '^2.2'))
          args += " --ignore-platform-req='ext-*' --ignore-platform-req='lib-*'";
      } else {
```
simplify

viceice (dismissed):

ste93cry:
@viceice Sorry, I had to push one more commit because the linter was complaining again, and I'm having some difficulties running it locally. Also, the code coverage check is failing because of [these](https://github.com/renovatebot/renovate/pull/34256/files#diff-e2479ebc289578272884dee7ff7905d1009e80e33cce6a4c619f145b1d970c54R241-R242) two lines of code: similar lines are also not covered, so let me know if it's fine or I should try to add some tests for that case too.

viceice (dismissed):

viceice:
> @viceice Sorry, I had to push one more commit because the linter was complaining again, and I'm having some difficulties running it locally. Also, the code coverage check is failing because of [these](https://github.com/renovatebot/renovate/pull/34256/files#diff-e2479ebc289578272884dee7ff7905d1009e80e33cce6a4c619f145b1d970c54R241-R242) two lines of code: similar lines are also not covered, so let me know if it's fine or I should try to add some tests for that case too.

let's see what the CI says, but those lines need to be covered by tests.

ste93cry:
>those lines need to be covered by tests

I would first question whether the `try-catch` block is really necessary. Actually, there are functions in the same file that don't have any special error handling when calling functions of the `npm` package, so I wonder why this one should be different 🤔

ste93cry:
I added a test to cover the missing lines. Can you have another look please?

viceice (approved):

PR #34257, opened 2025-02-17
fix(devbox): use new --no-install flag to prevent devbox from using the nix store

burritobill:
## Changes

Devbox was causing artifact errors due to the nix store installed in the renovate docker container being installed in a strange location and not being mounted under /nix.

We worked with jetify to get a new `--no-install` flag added to `devbox update`. This updates the lockfile without installing the packages to the nix store, which is fine for this purpose.

In order to only update the lock file for the packages that have been updated I have chained together the commands with the depName as a specifier. e.g. `devbox update nodejs --no-install; devbox update ruby --no-install`. This is because if update is called without specifying the package it will update all dependencies with versions such as `latest` and `3.0` or `3`.

## Context

This PR is related to https://github.com/renovatebot/renovate/pull/34002 and all associated PRs


rarkins:
Do we need to worry about people using old devbox versions?

lib/modules/manager/devbox/artifacts.ts
```diff
@@ -34,9 +34,17 @@ export async function updateArtifacts(
...
-    cmd += 'devbox install';
+    const updateCommands = updateConfig.updatedDeps
+      .map((dep) => dep.depName && `devbox update ${dep.depName} --no-install`)
+      .filter((dep) => dep);
+    if (updateCommands.length) {
+      cmd += updateCommands.join('; ');
+    } else {
+      logger.trace('No updated devbox packages - returning null');
+      return null;
+    }
```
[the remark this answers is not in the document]
burritobill: This is a requirement for this PR, the new flag on update means we have to call update on each dependency instead of calling `devbox install`

lib/modules/manager/devbox/artifacts.ts
```diff
@@ -34,9 +34,17 @@ export async function updateArtifacts(
     updateConfig.config.isLockFileMaintenance ||
     updateConfig.config.updateType === 'lockFileMaintenance'
   ) {
-    cmd += 'devbox update';
+    cmd += 'devbox update --no-install';
```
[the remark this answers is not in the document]
burritobill: Where does that value get set? as part of install-tool?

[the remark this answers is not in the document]
rarkins: Users have the ability to manually configure/specify devbox version using `constraints.devbox`. If they've pinned devbox to an older version which doesn't support `--no-install` then it means you need to detect that and drop the parameter.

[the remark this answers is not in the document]
burritobill: Ok perfect, I'll add that in tomorrow

lib/modules/manager/devbox/artifacts.ts
```diff
@@ -17,6 +18,11 @@ export async function updateArtifacts(
     logger.debug('No devbox.lock found');
     return null;
   }
+
+  const supportsNoInstall = updateConfig.config.constraints?.devbox
+    ? new SemVer(updateConfig.config.constraints.devbox).compare('0.14.0') >= 0
```
rarkins, line 23: The constraint could be a range, not just a version

rarkins (changes requested):

lib/modules/manager/devbox/artifacts.ts
```diff
@@ -1,42 +1,64 @@
...
     return null;
   }
+
+  let supportsNoInstall = true;
+  if (constraints?.devbox) {
+    const constraintVersion = await resolveConstraint({
+      toolName: 'devbox',
+      constraint: constraints?.devbox,
+    });
+    supportsNoInstall = new SemVer(constraintVersion).compare('0.14.0') >= 0;
```
viceice, line 28: use use something like `semver.satisfies(version, '>=0.14.0')` so no need to resolve the real version used .

lib/modules/manager/devbox/artifacts.spec.ts
```diff
@@ -90,6 +144,94 @@ describe('modules/manager/devbox/artifacts', () => {
...
+          newPackageFileContent: devboxJson,
+          updatedDeps: [{}],
+          config: {},
+        }),
+      ).toBeNull();
+    });
+
+    it('returns null if no updatedDeps have depNames', async () => {
+      fs.getSiblingFileName.mockReturnValueOnce('devbox.lock');
+      fs.readLocalFile.mockResolvedValueOnce(codeBlock`{}`);
```
viceice, line 215:
```suggestion
      fs.readLocalFile.mockResolvedValueOnce('{}');
```
template not needed for this

lib/modules/manager/devbox/artifacts.spec.ts
```diff
@@ -90,6 +144,94 @@ describe('modules/manager/devbox/artifacts', () => {
...
+      const oldLockFileContent = Buffer.from('Old devbox.lock');
+      const newLockFileContent = Buffer.from('New devbox.lock');
+      fs.readLocalFile.mockResolvedValueOnce(oldLockFileContent as never);
+      fs.readLocalFile.mockResolvedValueOnce(newLockFileContent as never);
```
viceice, line 219:
```suggestion
      fs.readLocalFile.mockResolvedValueOnce(oldLockFileContent);
      fs.readLocalFile.mockResolvedValueOnce(newLockFileContent);
```

I think the never is no longer needed.

lib/modules/manager/devbox/artifacts.ts
```diff
@@ -1,42 +1,58 @@
...
+}: UpdateArtifact): Promise<UpdateArtifactsResult[] | null> {
+  const lockFileName = getSiblingFileName(packageFileName, 'devbox.lock');
   const existingLockFileContent = await readLocalFile(lockFileName, 'utf8');
   if (!existingLockFileContent) {
     logger.debug('No devbox.lock found');
     return null;
   }
+
+  const supportsNoInstall = constraints?.devbox
+    ? semver.satisfies(constraints.devbox, '>=0.14.0')
```
viceice, line 22:
```suggestion
    ? semver.intersects(constraints.devbox, '>=0.14.0')
```
maybe this is safer if constraint is a range. 🤔

lib/modules/manager/devbox/artifacts.ts
```diff
@@ -1,42 +1,58 @@
...
 
   let cmd = '';
-  if (
-    updateConfig.config.isLockFileMaintenance ||
-    updateConfig.config.updateType === 'lockFileMaintenance'
-  ) {
-    cmd += 'devbox update';
-  } else if (is.nonEmptyArray(updateConfig.updatedDeps)) {
-    cmd += 'devbox install';
+  if (isLockFileMaintenance || updateType === 'lockFileMaintenance') {
```
viceice, line 38:
```suggestion
  if (isLockFileMaintenance) {
```
I think you no longer need to check update type

lib/modules/manager/devbox/artifacts.ts
```diff
@@ -1,42 +1,58 @@
...
+    cmd += supportsNoInstall ? 'devbox update --no-install' : 'devbox update';
+  } else if (is.nonEmptyArray(updatedDeps)) {
+    if (supportsNoInstall) {
+      const updateCommands = updatedDeps
+        .map(
+          (dep) => dep.depName && `devbox update ${dep.depName} --no-install`,
```
viceice, line 44:
```suggestion
          (dep) => dep.depName && `devbox update ${quote(dep.depName)} --no-install`,
```

lib/modules/manager/devbox/artifacts.ts
```diff
@@ -1,42 +1,58 @@
...
-        constraint: updateConfig.config.constraints?.devbox,
+        constraint: constraints?.devbox,
       },
     ],
     docker: {},
-    userConfiguredEnv: updateConfig.config.env,
+    userConfiguredEnv: env,
   };
 
   let cmd = '';
```
viceice, line 37:
```suggestion
  const cmd = [];
```
use array form

lib/modules/manager/devbox/artifacts.ts
```diff
@@ -1,42 +1,58 @@
...
-  if (
-    updateConfig.config.isLockFileMaintenance ||
-    updateConfig.config.updateType === 'lockFileMaintenance'
-  ) {
-    cmd += 'devbox update';
-  } else if (is.nonEmptyArray(updateConfig.updatedDeps)) {
-    cmd += 'devbox install';
+  if (isLockFileMaintenance || updateType === 'lockFileMaintenance') {
+    cmd += supportsNoInstall ? 'devbox update --no-install' : 'devbox update';
```
viceice, line 39:
```suggestion
    cmd.push(supportsNoInstall ? 'devbox update --no-install' : 'devbox update');
```

lib/modules/manager/devbox/artifacts.ts
```diff
@@ -1,42 +1,58 @@
...
+  } else if (is.nonEmptyArray(updatedDeps)) {
+    if (supportsNoInstall) {
+      const updateCommands = updatedDeps
+        .map(
+          (dep) => dep.depName && `devbox update ${dep.depName} --no-install`,
+        )
+        .filter((dep) => dep);
+      if (updateCommands.length) {
+        cmd += updateCommands.join('; ');
```
viceice, line 48:
```suggestion
        cmd.push(...updateCommands);
```

lib/modules/manager/devbox/artifacts.ts
```diff
@@ -1,42 +1,58 @@
...
+      } else {
+        logger.trace('No updated devbox packages - returning null');
+        return null;
+      }
+    } else {
+      cmd += 'devbox install';
```
viceice, line 54:
```suggestion
      cmd.push( 'devbox install');
```

lib/modules/manager/devbox/artifacts.spec.ts
burritobill: This seems to be required as mockResovedValueOnce is expecting a string but old/newLockFileContent are Buffers. The test passes anyway so I think this is fine.

viceice (approved):

burritobill:
@rarkins Can you tick the requested change off so this one can be merged?

rarkins (approved):

PR #34267, opened 2025-02-17
fix(npm): include credentials without host type for lock file generation

languitar:
## Changes

So far, the post-update task for generating the NPM lock file did not pick up credentials from `hostRules` that lacked a `hostType`. This was inconsistent to the way renovate had previously determined updatable dependencies. The code path there also included `hostRules` without a `hostType` defined. As a result, PRs would potentially fail because the lock file generation did not have the same set of credentials available as the trigger for creating a PR at all.

Now, lock file generation also picks up credentials without a defined `hostType`.

## Context

Related discussion: https://github.com/renovatebot/renovate/discussions/34203


lib/modules/manager/npm/post-update/rules.ts
```diff
@@ -19,8 +19,13 @@ export function processHostRules(): HostRulesResult {
...
     hostType: 'npm',
   });
-  logger.debug(`Found ${npmHostRules.length} npm host rule(s)`);
-  for (const hostRule of npmHostRules) {
+  // Include host rules without specific type to mimic the behavior used when determining dependencies with updates.
+  const noTypeHostRules = hostRules
+    .getAll()
+    .filter((rule) => rule.hostType === null || rule.hostType === undefined);
+  const effectiveHostRules = npmHostRules.concat(noTypeHostRules);
+  logger.debug(`Found ${effectiveHostRules.length} npm host rule(s)`);
```
rarkins, line 27: Can you log how many were found with hostType=npm and then how many were found with empty hostType?

lib/modules/manager/npm/post-update/rules.ts
```diff
@@ -19,8 +19,13 @@ export function processHostRules(): HostRulesResult {
...
+  for (const hostRule of effectiveHostRules) {
```
rarkins, line 28: You need to dedupe first. Ones with hostType=npm should have higher priority

rarkins (changes requested):

lib/modules/manager/npm/post-update/rules.ts
languitar: Added. The implementation currently assumes that the list of rules is not terribly long. Otherwise, a more efficient data structure should be used for deduplication. Is that fine?

rarkins: It shouldn't be particularly long

rarkins (dismissed):

lib/modules/manager/npm/post-update/rules.spec.ts
```diff
@@ -142,5 +142,82 @@ describe('modules/manager/npm/post-update/rules', () => {
...
       );
     });
+
+    it('uses rules without host type', () => {
+      hostRules.add({
+        matchHost: 'registry.company.com',
+        token: 'sometoken',
+      });
+      const res = processHostRules();
+      expect(res).toMatchInlineSnapshot(
```
viceice, line 152: no snapshot, use toEqual

lib/modules/manager/npm/post-update/rules.ts
```diff
@@ -20,7 +20,27 @@ export function processHostRules(): HostRulesResult {
...
+    .getAll()
+    .filter((rule) => rule.hostType === null || rule.hostType === undefined);
+  logger.debug(
+    `Found ${noTypeHostRules.length} host rule(s) without host type`,
+  );
+  // Drop duplicates for the same matchHost while prefering the more specific rules with hostType npm.
+  const noTypeHostRulesWithoutDuplicates = noTypeHostRules.filter(
+    (rule) => !npmHostRules.some((elem) => elem.matchHost === rule.matchHost),
+  );
+  logger.debug(
```
viceice, line 34:
```suggestion
  logger.trace(
```
I think we can drop this log message

lib/modules/manager/npm/post-update/rules.ts
```diff
@@ -20,7 +20,27 @@ export function processHostRules(): HostRulesResult {
...
+    `Found ${noTypeHostRulesWithoutDuplicates.length} host rule(s) without host type after dropping duplicates`,
+  );
+  const effectiveHostRules = npmHostRules.concat(
+    noTypeHostRulesWithoutDuplicates,
+  );
+  logger.debug(
```
viceice, line 40: I think this can be dropped too, I see no real benefit for debugging 
```suggestion
  logger.trace(
```

lib/modules/manager/npm/post-update/rules.ts
rarkins: I'd prefer to log something like this so that this new feature is not "invisible". Perhaps the last message can be lowered

languitar: So what's the verdict here?

lib/modules/manager/npm/post-update/rules.spec.ts
languitar: Sorry for the delay. Changed now.

lib/modules/manager/npm/post-update/rules.ts
languitar: This should be resolved already?

viceice:
please avoid force push

viceice (dismissed):

viceice (approved):

PR #34277, opened 2025-02-18
feat(manager/kustomize): Support inflating helm charts

### LINKED ISSUE #14137: Support for inflating helmchart deptype upgrades in the kustomize manager

Right now when the kustomize manager resolves an upgrade for a dependency of the helmchart deptype the kustomization.yaml is upgraded but the chart dependency is not inflated which means that the resulting Renovate PR is incomplete for repositories where the inflated chart is vendored (as per kustomizes best practice: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/chart.md#best-practice).

Renovate would likely need to first be able to determine if the chart is vendored, first because it likely needs to pull out if the chart is not vendored, but also in order to delete it before running `kustomize build --enable-helm` on the upgraded kustomization.yaml as per https://github.com/kubernetes-sigs/kustomize/blob/master/examples/chart.md#how-does-the-pull-work (i.e kustomize doesn't pull down the chart if it's already inflated regardless of version).

No

micke:
## Changes

1. Add support for extracting `helmGlobals` options from kustomization files to support custom destination for inflated helm charts.
1. Add support for inflating helmchart deptypes in the kustomize manager.

## Context

Closes #14137 

The helm charts will be inflated if any of the following is true:

1. The `currentVersion` is inflated, OR
1. The `kustomizeInflateHelmCharts` `postUpdateOptions` is enabled.

If the `currentVersion` is inflated then they will be deleted.

If the chart is already inflated then Renovate won't try to inflate them again.

I decided to accomplish this by directly interfacing with helm commands to inflate the charts by calling `helm add` instead of relying on using `kustomize build` to have kustomize inflate the charts.

I took this approach because i've seen setups where a kustomization might need pre-processing to be able to be built, a common example of this might be encrypted secrets, where the secrets referenced in the kustomization need to be decrypted before using `kustomize build`.

I've tried to structure the code so that it would work when a dependency is upgraded but i also envision a scenario where Renovate could be used to inflate existing charts used in kustomize kind of like how Renovate today can generate a PR to pin docker digests with a "Pin dependencies" PR when using `docker:pinDigests`.
But i'm not yet comfortable enough with the codebase to dive deeper into what it would take to make that a possibility, so i would greatly appreciate guidance on that.


lib/modules/manager/kustomize/extract.ts
```diff
@@ -200,6 +200,9 @@ export function parseKustomize(
 
   pkg.kind ??= 'Kustomization';
 
+  pkg.helmGlobals ??= { chartHome: 'charts' };
```
viceice, line 203:
```suggestion
  pkg.helmGlobals ??= { };
```

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,217 @@
...
+        isUpdateOptionInflateChartArchives,
+      );
+    }
+
+    const status = await getRepoStatus();
+    const chartsAddition = status?.not_added ?? [];
+    const chartsDeletion = status?.deleted ?? [];
+
+    const fileChanges: UpdateArtifactsResult[] = [];
+    await writeLocalFile(packageFileName, newPackageFileContent);
```
viceice, line 173: Doesn't make sense to write a file to disk here. What's the intention?

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,217 @@
...
+    return;
+  }
+
+  if (dependency.depType !== 'HelmChart') {
+    return;
+  }
+
+  if (!is.nonEmptyArray(dependency.registryUrls)) {
+    return;
+  }
```
viceice, line 65: move this to outer calling loop to save funciton calls (perf)

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,217 @@
...
+  readLocalFile,
+  writeLocalFile,
+} from '../../../util/fs';
+import { getRepoStatus } from '../../../util/git';
+import type { UpdateArtifact, UpdateArtifactsResult, Upgrade } from '../types';
+import { parseKustomize } from './extract';
+
+function generateHelmEnvs(): ExtraEnv {
+  return {
+    HELM_EXPERIMENTAL_OCI: '1',
```
viceice, line 26: do we really need this? should only be set if user sets contraints which causes a helm version <3.8.0

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    const execOptions: ExecOptions = {
+      docker: {},
+      userConfiguredEnv: config.env,
+      extraEnv: generateHelmEnvs(),
+      toolConstraints: [helmToolConstraint],
+    };
+
+    for (const dependency of updatedDeps) {
+      await inflateHelmChart(
```
viceice, line 160: You're reading `project`, so use that to only call this function for updates which are chart references. return early if no chart was updated

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,217 @@
...
+  if (!flagEnabled && is.nullOrUndefined(currentChartExistingPath)) {
+    logger.debug(
+      `Not inflating Helm chart for ${dependency.depName} as kustomizeInflateHelmCharts is not enabled and the current version isn't inflated`,
+    );
+    return;
+  }
+
+  if (
+    is.nonEmptyString(currentChartExistingPath) &&
+    is.nonEmptyString(dependency.newVersion)
```
viceice, line 82: return early if current and new version are equal?

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,217 @@
...
+  const untarDir = upath.join(chartHome, folderName);
+  const registryUrl = dependency.registryUrls[0];
+  logger.debug(
+    `Pulling helm chart ${dependency.depName} version ${versionToPull} to ${untarDir}`,
+  );
+  const cmd =
+    `helm pull --untar --untardir ${quote(untarDir)} ` +
+    `--version ${quote(versionToPull)} --repo ${quote(registryUrl)} ${dependency.depName}`;
+
+  await exec([cmd], execOptions);
```
viceice, line 117:
```suggestion
  await exec(cmd, execOptions);
```

viceice (changes requested):

lib/modules/manager/kustomize/artifacts.ts
micke: Just inexperience with the project, i wasn't sure if the new packagefile content needed to be written to disk here or if it's done somewhere else.

lib/modules/manager/kustomize/artifacts.ts
micke: Am i understanding you correctly that i need to check that the `dependency` is referenced in the parsed kustomize `project`? Because there could be `HelmChart` deptypes present that aren't part of the kustomize project?

viceice: there could be docker image updates and you should not try to use them as helm chart update

lib/modules/manager/kustomize/artifacts.ts
micke: It looks like we don't need to write `newPackageFileContent` in the artifacts management.
I'm running https://github.com/renovatebot/renovate/pull/34277/commits/3f057936e99a43b1dc51f071dc6dc226903e1a40 now and the updated `kustomization.yaml` file is included in the PR without me having to write it in artifacts management, see this PR for example: https://github.com/micke/renovate-kustomize-artifacts/pull/8/files#diff-5c38001a592677b709f1e7de7702d683403356ce7824481d04ed5def445663ba

If that's not by design or if you think we should write it anyways then i'll add it back but we certainly don't need to write it before calling helm as we're not using the file to pull the helm charts...

lib/modules/manager/kustomize/artifacts.ts
micke: Let me know if you think i should remove it, otherwise i'm under the impression that it should be there as-is.

micke:
I've run this on a repository with two example kustomization.yaml files, one with "regular" helm charts and one with OCI chart: https://github.com/micke/renovate-kustomize-artifacts

The resulting "regular" helm chart PR: https://github.com/micke/renovate-kustomize-artifacts/pull/15
And the resulting OCI chart PR: https://github.com/micke/renovate-kustomize-artifacts/pull/8

A side effect of having the helm charts inflated is that renovate manages updates for sub-charts and images in these inflated charts.
How should we reason about those? Do we leave it up to the user to exclude/ignore the inflated charts or can and should we do something about that?

lib/modules/manager/kustomize/readme.md
```diff
@@ -20,7 +20,14 @@ This manager uses three `depType`s to allow fine-grained control of which depend
 - HelmChart
 - OCIChart
 
-**Limitations**
+### Helm charts inflation
+
+Renovate will inflate helm charts referenced in a kustomization if any of the following is true:
```
viceice, line 25: Add a note that you need to manually exclude the charts from helm managers

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,254 @@
...
+  readLocalFile,
+} from '../../../util/fs';
+import { getRepoStatus } from '../../../util/git';
+import { DockerDatasource } from '../../datasource/docker';
+import { HelmDatasource } from '../../datasource/helm';
+import type { UpdateArtifact, UpdateArtifactsResult } from '../types';
+import { parseKustomize } from './extract';
+
+function generateHelmEnvs(): ExtraEnv {
+  return {
```
viceice, line 26:
```suggestion
  const cacheDir = privateCacheDir();
  return {
```

don't call this multiple times (perf)

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,254 @@
...
+    HELM_EXPERIMENTAL_OCI: '1',
+    // set cache and config files to a path in privateCacheDir to prevent file and credential leakage
+    HELM_REGISTRY_CONFIG: `${upath.join(privateCacheDir(), 'registry.json')}`,
+    HELM_REPOSITORY_CONFIG: `${upath.join(
+      privateCacheDir(),
+      'repositories.yaml',
+    )}`,
+    HELM_REPOSITORY_CACHE: `${upath.join(privateCacheDir(), 'repositories')}`,
```
viceice, line 34:
```suggestion
    HELM_REGISTRY_CONFIG: upath.join(cacheDir, 'registry.json'),
    HELM_REPOSITORY_CONFIG: upath.join(cacheDir, 'repositories.yaml'),
    HELM_REPOSITORY_CACHE: upath.join(cacheDir, 'repositories'),
```

`upath.join` already returns a string

lib/modules/manager/kustomize/artifacts.ts
viceice: yes, only set if `config.constraints?.helm` is set and `semver.satisfies(config.constraints.helm, '<3.8.0')`

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,261 @@
...
+import { parseKustomize } from './extract';
+
+function generateHelmEnvs(config: UpdateArtifactsConfig): ExtraEnv {
+  const cacheDir = privateCacheDir();
+  const enableExperimentalOci =
+    config.constraints?.helm &&
+    semver.satisfies(config.constraints.helm, '<3.8.0');
+
+  return {
+    HELM_EXPERIMENTAL_OCI: enableExperimentalOci ? '1' : '0',
```
viceice, line 37: don't add this env at all on newer helm. needs a test.

you probably should move this function to common or utils file in this folder for easier testing

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,261 @@
...
+  repository: string,
+  depName: string,
+  datasource?: string,
+): string {
+  switch (datasource) {
+    case HelmDatasource.id:
+      return `--repo ${quote(repository)} ${depName}`;
+    case DockerDatasource.id:
+      return quote(`oci://${repository}`);
+    // istanbul ignore next: should never happen
```
viceice, line 67: no longer works, see updated development docs

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,575 @@
...
+        packageFileName,
+        updatedDeps,
+        newPackageFileContent,
+        config: {
+          ...config,
+          postUpdateOptions: [],
+        },
+      }),
+    ).toBeNull();
+    expect(execSnapshots).toMatchObject([]);
```
viceice, line 232: use tobeemptyarray matcher

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,237 @@
...
+        artifactError: {
+          stderr: 'Failed to parse new package file content',
+        },
+      },
+    ];
+  }
+
+  const chartHome = getSiblingFileName(
+    packageFileName,
+    project.helmGlobals?.chartHome as string, // eslint-disable-line @typescript-eslint/non-nullable-type-assertion-style
```
micke, line 129: `lint` complained about `@typescript-eslint/non-nullable-type-assertion-style` and `lint-fix` changed this to:
```typescript
project?.helmGlobals?.chartHome!
```
`lint` then complained about `@typescript-eslint/no-non-null-asserted-optional-chain`.

`chartHome` always has a value because we set the kustomize default if it's not explicitly set, and it wasn't optional in https://github.com/renovatebot/renovate/pull/34277/commits/c0761652c042e31124a2f4abbfe6870241a8cdfe but Michael wanted it changed in his comment [here](https://github.com/renovatebot/renovate/pull/34277/commits/c0761652c042e31124a2f4abbfe6870241a8cdfe#r1959600370).

I don't know what else to do but disable the lint at this point...

lib/modules/manager/kustomize/common.ts
```diff
@@ -0,0 +1,25 @@
...
+  const envs: ExtraEnv = {
+    // set cache and config files to a path in privateCacheDir to prevent file and credential leakage
+    HELM_REGISTRY_CONFIG: upath.join(cacheDir, 'registry.json'),
+    HELM_REPOSITORY_CONFIG: upath.join(cacheDir, 'repositories.yaml'),
+    HELM_REPOSITORY_CACHE: upath.join(cacheDir, 'repositories'),
+  };
+
+  if (
+    config.constraints?.helm &&
+    semver.satisfies(config.constraints.helm, '<3.8.0')
```
viceice, line 19:
```suggestion
    semver.intersects(config.constraints.helm, '<3.8.0')
```

i think we need this if `config.constraints.helm` itself is a range

lib/modules/manager/kustomize/common.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+        HELM_REPOSITORY_CONFIG:
+          '/tmp/cache/__renovate-private-cache/repositories.yaml',
+        HELM_REPOSITORY_CACHE:
+          '/tmp/cache/__renovate-private-cache/repositories',
+      });
+    });
+
+    it('generates envs with experimental OCI', () => {
+      const config = {
+        constraints: { helm: '3.7.0' },
```
viceice, line 36: add a sample for `>=3.7.0`

lib/modules/manager/kustomize/artifacts.ts
viceice:
```suggestion
    project.helmGlobals!.chartHome!
```
or 
```suggestion
    project.helmGlobals?.chartHome ?? 'charts';
```

lib/modules/manager/kustomize/extract.ts
```diff
@@ -200,6 +200,9 @@ export function parseKustomize(
 
   pkg.kind ??= 'Kustomization';
 
+  pkg.helmGlobals ??= {};
+  pkg.helmGlobals.chartHome ??= 'charts';
+
```
viceice, line 205:
```suggestion
```

lib/modules/manager/kustomize/common.ts
```diff
@@ -0,0 +1,25 @@
...
+  const envs: ExtraEnv = {
+    // set cache and config files to a path in privateCacheDir to prevent file and credential leakage
+    HELM_REGISTRY_CONFIG: upath.join(cacheDir, 'registry.json'),
+    HELM_REPOSITORY_CONFIG: upath.join(cacheDir, 'repositories.yaml'),
+    HELM_REPOSITORY_CACHE: upath.join(cacheDir, 'repositories'),
+  };
+
+  if (
+    config.constraints?.helm &&
+    semver.intersects(config.constraints.helm, '<3.8.0')
```
viceice, line 19:
```suggestion
    !semver.intersects(config.constraints.helm, '>=3.8.0')
```
so a constraint `>=3.7.0` will not set it, because it will be resoved to a much later helm version

lib/modules/manager/kustomize/common.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+      };
+      const envs = generateHelmEnvs(config);
+      expect(envs).toEqual({
+        HELM_REGISTRY_CONFIG:
+          '/tmp/cache/__renovate-private-cache/registry.json',
+        HELM_REPOSITORY_CONFIG:
+          '/tmp/cache/__renovate-private-cache/repositories.yaml',
+        HELM_REPOSITORY_CACHE:
+          '/tmp/cache/__renovate-private-cache/repositories',
+        HELM_EXPERIMENTAL_OCI: '1',
```
viceice, line 79:
```suggestion
```

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,237 @@
...
+        file: {
+          type: 'deletion',
+          path: file,
+        },
+      });
+    }
+
+    return fileChanges.length > 0 ? fileChanges : null;
+  } catch (err) {
+    /* v8 ignore next 3: not easily testable */
```
viceice, line 224:
```suggestion
```

throw `new Error(TEMPORARY_ERROR)` error from `localPathExists` in a test

lib/modules/manager/kustomize/artifacts.ts
```diff
@@ -0,0 +1,237 @@
...
+        artifactError: {
+          stderr: 'Failed to parse new package file content',
+        },
+      },
+    ];
+  }
+
+  const chartHome = getSiblingFileName(
+    packageFileName,
+    project.helmGlobals!.chartHome!,
```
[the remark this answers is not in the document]
micke: Sorry, i interpreted this suggestion as a choice for me to make and as we default chartHome to `'charts'` in extract i felt it was unnecessary to do it here aswell.

micke:
Would be great to get another review on this!

lib/modules/manager/kustomize/extract.spec.ts
```diff
@@ -55,6 +55,15 @@ describe('modules/manager/kustomize/extract', () => {
     expect(file?.kind).toBe('Kustomization');
   });
 
+  it('should extract chartHome', () => {
+    const file = parseKustomize(`
+      helmGlobals:
+        chartHome: customPathToCharts
+    `);
+    expect(file).not.toBeNull();
+    expect(file?.helmGlobals?.chartHome).toBe('customPathToCharts');
```
viceice, line 64:
```suggestion
    expect(file!.helmGlobals?.chartHome).toBe('customPathToCharts');
```

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,575 @@
...
+    fs.deleteLocalFile.mockResolvedValueOnce(undefined);
+    fs.privateCacheDir.mockReturnValue(
+      '/tmp/renovate/cache/__renovate-private-cache',
+    );
+  });
+
+  afterEach(() => {
+    GlobalConfig.reset();
+  });
+
```
viceice, line 48:
```suggestion
```
not needed, the set above is overriding all previous values

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,575 @@
...
+    ];
+    expect(
+      await kustomize.updateArtifacts({
+        packageFileName,
+        updatedDeps,
+        newPackageFileContent,
+        config,
+      }),
+    ).toBeNull();
+    expect(execSnapshots).toMatchObject([]);
```
viceice, line 265: use tobeemptyarray

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,575 @@
...
+        updatedDeps,
+        newPackageFileContent,
+        config: {
+          ...config,
+          postUpdateOptions: [],
+        },
+      }),
+    ).rejects.toThrowError(TEMPORARY_ERROR);
+    expect(fs.deleteLocalFile).not.toHaveBeenCalled();
+    expect(execSnapshots).toMatchObject([]);
```
viceice, line 573: see other comment

micke:
@viceice Alright, are we done now? If you got any more nitpicks then feel free to push them yourself directly.

micke:
@viceice @rarkins Is there anything stopping this from being merged? Would be great to get this in!

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,571 @@
+import { join } from 'upath';
+import { envMock, mockExecAll } from '../../../../test/exec-util';
+import { env, fs, git, partial } from '../../../../test/util';
+import { GlobalConfig } from '../../../config/global';
+import type { RepoGlobalConfig } from '../../../config/types';
+import { TEMPORARY_ERROR } from '../../../constants/error-messages';
+import * as docker from '../../../util/exec/docker';
+import type { StatusResult } from '../../../util/git/types';
+import * as hostRules from '../../../util/host-rules';
```
viceice, line 9:
```suggestion
```

you don't use host rules

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,571 @@
...
+
+const packageFileName = 'kustomization.yaml';
+const newPackageFileContent = 'kind: Kustomization';
+
+describe('modules/manager/kustomize/artifacts', () => {
+  beforeEach(() => {
+    env.getChildProcessEnv.mockReturnValue(envMock.basic);
+    GlobalConfig.set(adminConfig);
+    docker.resetPrefetchedImages();
+    hostRules.clear();
```
viceice, line 37:
```suggestion
```

not needed

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,571 @@
+import { join } from 'upath';
+import { envMock, mockExecAll } from '../../../../test/exec-util';
+import { env, fs, git, partial } from '../../../../test/util';
+import { GlobalConfig } from '../../../config/global';
+import type { RepoGlobalConfig } from '../../../config/types';
+import { TEMPORARY_ERROR } from '../../../constants/error-messages';
+import * as docker from '../../../util/exec/docker';
```
viceice, line 7:
```suggestion
```

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,571 @@
...
+const newPackageFileContent = 'kind: Kustomization';
+
+describe('modules/manager/kustomize/artifacts', () => {
+  beforeEach(() => {
+    env.getChildProcessEnv.mockReturnValue(envMock.basic);
+    GlobalConfig.set(adminConfig);
+    docker.resetPrefetchedImages();
+    hostRules.clear();
+    fs.getSiblingFileName.mockReturnValueOnce('charts');
+    fs.deleteLocalFile.mockResolvedValueOnce(undefined);
```
viceice, line 39:
```suggestion
```

default return value is undefined

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,571 @@
...
+        packageFileName,
+        updatedDeps,
+        newPackageFileContent,
+        config,
+      }),
+    ).toBeNull();
+  });
+
+  it('returns null if newVersion is not found and currentVersion is already inflated', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 174: is it important to return `null` instead of `undefined`?

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,571 @@
...
+        packageFileName,
+        updatedDeps,
+        newPackageFileContent,
+        config,
+      }),
+    ).toBeNull();
+  });
+
+  it('returns null if old version is not inflated and kustomizeInflateHelmCharts is not enabled', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 198: same

micke:
At this point i'm mostly chasing to keep up to date with changes made in upstream and really small nitpicks.
Like i said before, feel free to push nitpicks directly so we can get this merged please.

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -452,6 +457,115 @@ describe('modules/manager/kustomize/artifacts', () => {
...
+          contents: null,
+        },
+      },
+    ]);
+    expect(fs.deleteLocalFile).not.toHaveBeenCalled();
+    expect(execSnapshots).toMatchObject([
+      { cmd: 'docker pull ghcr.io/containerbase/sidecar' },
+      { cmd: 'docker ps --filter name=renovate_sidecar -aq' },
+      {
+        cmd: 'docker run --rm --name=renovate_sidecar --label=renovate_child -v "/tmp/github/some/repo":"/tmp/github/some/repo" -v "/tmp/renovate/cache":"/tmp/renovate/cache" -e HELM_REGISTRY_CONFIG -e HELM_REPOSITORY_CONFIG -e HELM_REPOSITORY_CACHE -e CONTAINERBASE_CACHE_DIR -w "/tmp/github/some/repo" ghcr.io/containerbase/sidecar bash -l -c "install-tool helm 3.17.0 && helm pull --untar --untardir charts/example-1.0.0 --version 1.0.0 oci://github.com/example/example/example"',
```
viceice, line 564: please format it like this:
https://github.com/renovatebot/renovate/blob/242ef8dd87e9e3ae3ca19a9eaeb9aa440509f3d4/lib/modules/manager/gomod/artifacts.spec.ts#L1438-L1460

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,691 @@
...
+          ...config,
+          postUpdateOptions: [],
+        },
+      }),
+    ).toBeNull();
+    expect(execSnapshots).toBeEmptyArray();
+  });
+
+  it('returns null if newVersion and currentVersion is the same', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 230: still needed? Default value is `undefined` which should also do it. So no need to mock a value here and below
```suggestion
```

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,691 @@
...
+        updatedDeps,
+        newPackageFileContent,
+        config,
+      }),
+    ).toBeNull();
+    expect(execSnapshots).toBeEmptyArray();
+  });
+
+  it('inflates new version if old version is inflated and kustomizeInflateHelmCharts is not enabled', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 262: still needed?
```suggestion
```

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,691 @@
...
+    expect(fs.deleteLocalFile).toHaveBeenCalledWith('charts/example-1.0.0');
+    expect(execSnapshots).toMatchObject([
+      {
+        cmd: 'helm pull --untar --untardir charts/example-2.0.0 --version 2.0.0 --repo https://github.com.com/example/example example',
+      },
+    ]);
+  });
+
+  it('inflates new version if old version is not inflated but kustomizeInflateHelmCharts is enabled', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 323: still needed?
```suggestion
```

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,691 @@
...
+    expect(fs.deleteLocalFile).not.toHaveBeenCalled();
+    expect(execSnapshots).toMatchObject([
+      {
+        cmd: 'helm pull --untar --untardir charts/example-2.0.0 --version 2.0.0 --repo https://github.com.com/example/example example',
+      },
+    ]);
+  });
+
+  it('inflates current version if no new version and kustomizeInflateHelmCharts is enabled', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 369: still needed?
```suggestion
```

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,691 @@
...
+    expect(fs.deleteLocalFile).not.toHaveBeenCalled();
+    expect(execSnapshots).toMatchObject([
+      {
+        cmd: 'helm pull --untar --untardir charts/example-1.0.0 --version 1.0.0 --repo https://github.com.com/example/example example',
+      },
+    ]);
+  });
+
+  it('handles OCI repositories', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 415: still needed?
```suggestion
```

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,691 @@
...
+    expect(execSnapshots).toMatchObject([
+      {
+        cmd: 'helm pull --untar --untardir charts/example-1.0.0 --version 1.0.0 oci://github.com/example/example/example',
+      },
+    ]);
+  });
+
+  it('installs binaries on install mode', async () => {
+    GlobalConfig.set({ ...adminConfig, binarySource: 'install' });
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 462: still needed?
```suggestion
```

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,691 @@
...
+    ]);
+  });
+
+  it('installs binaries on docker mode', async () => {
+    GlobalConfig.set({
+      ...adminConfig,
+      binarySource: 'docker',
+      dockerSidecarImage: 'ghcr.io/containerbase/sidecar',
+    });
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 518: still needed?
```suggestion
```

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,691 @@
...
+          'install-tool helm 3.17.0' +
+          ' && ' +
+          'helm pull --untar --untardir charts/example-1.0.0 --version 1.0.0 oci://github.com/example/example/example' +
+          '"',
+      },
+    ]);
+  });
+
+  it('does not inflate current version if kustomizeInflateHelmCharts is not enabled', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 584: still needed?
```suggestion
```

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,691 @@
...
+          postUpdateOptions: [],
+        },
+      }),
+    ).toBeNull();
+    expect(fs.deleteLocalFile).not.toHaveBeenCalled();
+    expect(execSnapshots).toBeEmptyArray();
+  });
+
+  it('catches errors', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 621: still needed?
```suggestion
```

lib/modules/manager/kustomize/artifacts.spec.ts
```diff
@@ -0,0 +1,691 @@
...
+          postUpdateOptions: [],
+        },
+      }),
+    ).toEqual([{ artifactError: { stderr: 'not found' } }]);
+    expect(fs.deleteLocalFile).not.toHaveBeenCalled();
+    expect(execSnapshots).toBeEmptyArray();
+  });
+
+  it('throws on TEMPORARY_ERROR', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(null);
```
viceice, line 660: still needed?
```suggestion
```

viceice (approved):

PR #34313, opened 2025-02-19
fix: append the static repo config to the repo config

Gabriel-Ladzaretti:
## Changes
directly merge `RENOVATE_STATIC_REPO_CONFIG` into the parsed `repo` file config. doing so gives the static config higher precedence over the repo config.

previously, the static config was applied before the `repo` config and passed via the runtime config, which meant it had lower priority _even compared to presets_ which was unintended.


lib/workers/repository/init/merge.spec.ts
```diff
@@ -32,6 +35,13 @@ const migrate = mocked(_migrate);
 const migrateAndValidate = mocked(_migrateAndValidate);
 const onboardingCache = mocked(_onboardingCache);
 
+const readLocalFileMock = mockedFunction(fs.readLocalFile);
+const fileListMock = mockedFunction(scm.getFileList);
+const migrateConfigMock = mockedFunction(migrate.migrateConfig);
+const migrateAndValidateMock = mockedFunction(
+  migrateAndValidate.migrateAndValidate,
+);
+
```
viceice, line 44: those are already mocks, so no need to do again. use the functions directly

lib/workers/repository/init/merge.ts
```diff
@@ -187,15 +188,20 @@ export async function mergeRenovateConfig(
...
+  const configFileAndEnv = await mergeStaticRepoEnvConfig(
+    configFileParsed,
+    process.env,
+  );
   if (is.nonEmptyArray(returnConfig.extends)) {
-    configFileParsed.extends = [
+    configFileAndEnv.extends = [
       ...returnConfig.extends,
-      ...(configFileParsed.extends || []),
+      // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
viceice, line 198: why this? it seems your local installation is not uptodate? run `pnpm i`?

Gabriel-Ladzaretti: https://github.com/renovatebot/renovate/actions/runs/13413637678/job/37469179680
github ci failed :thinking:

viceice: I wonder why we don't use `??` operator here should work because we already migrated this field to an array.

Gabriel-Ladzaretti: tbh, i think `||` is safer in this case regardless.

viceice: I don't think so. It's to be lazy. `configFileParsed.extends` should never be something different thang `null`, `undefined` or an `array` here. We should really get rid of such lazy safeguards.

viceice (dismissed):

viceice (approved):

PR #34319, opened 2025-02-19
Revert "chore(deps): update dependency @semantic-release/exec to v7"

secustor:
Reverts renovatebot/renovate#34302


travi:
happened to spot this in my notifications. any particular problems you ran into that is causing you to revert? i'm interested to understand if there is a problem that we need to look into on the semantic-release side

secustor:
We have a failure in the build process, but we are not sure yet if this major is the cause or the pnpm 10 upgrade.

PR #34324, opened 2025-02-19
feat: Repository cache git backend

zharinov:

### DIFF
--- a/lib/util/git/repo-store.ts
+++ b/lib/util/git/repo-store.ts
@@ -151,7 +151,7 @@ export async function set(
   }
   logger.debug(`Git commit created: ${commitSha}`);
 
-  const refName = `refs/renovate-data/${key}`;
+  const refName = `refs/renovate/data/${key}`;
   await createRef(refName, commitSha);
   logger.debug(`Git ref created: ${refName}`);
 
@@ -215,12 +215,12 @@ export async function get(
   await syncGit();
 
   try {
-    const ref = `refs/renovate-data/${key}`;
+    const ref = `refs/renovate/data/${key}`;
 
     await execGit([
       'fetch',
       'origin',
-      'refs/renovate-data/*:refs/renovate-data/*',
+      'refs/renovate/data/*:refs/renovate/data/*',
     ]);
 
     const commitSha = await getRefCommit(ref);
--- a/lib/util/git/repo-store.ts
+++ b/lib/util/git/repo-store.ts
@@ -1,0 +1,250 @@
+import { spawn } from 'child_process';
+import { GlobalConfig } from '../../config/global';
+import { logger } from '../../logger';
+import type { LongCommitSha } from './types';
+import { syncGit } from '.';
+
+interface ExecResult {
+  code: number | null;
+  stdout: string;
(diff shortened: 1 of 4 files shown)


lib/util/git/repo-store.ts
```diff
@@ -0,0 +1,250 @@
...
+  }
+  logger.debug(`Git tree created: ${treeSha}`);
+
+  const commitSha = await createCommit({ tree: treeSha, message: key });
+  if (!commitSha) {
+    return;
+  }
+  logger.debug(`Git commit created: ${commitSha}`);
+
+  const refName = `refs/renovate-data/${key}`;
```
rarkins, line 154: Do you think we should separate `refs/renovate` from `refs/renovate-data`? e.g. not `refs/renovate/data/*`?

lib/util/git/repo-store.ts
```diff
@@ -0,0 +1,250 @@
...
+  if (!treeSha) {
+    return;
+  }
+  logger.debug(`Git tree created: ${treeSha}`);
+
+  const commitSha = await createCommit({ tree: treeSha, message: key });
+  if (!commitSha) {
+    return;
+  }
+  logger.debug(`Git commit created: ${commitSha}`);
```
rarkins, line 152: Was this something you learned from previously, that it must be a _commit_ to be retained by github?

lib/util/git/repo-store.ts
```diff
@@ -0,0 +1,250 @@
...
+  return stdout.trim() as LongCommitSha;
+}
+
+interface CreateCommitOptions {
+  tree: LongCommitSha;
+  message: string;
+  parents?: LongCommitSha[];
+}
+
+async function createCommit({
```
rarkins, line 87: Is this an "orphan" commit we push?

lib/util/git/repo-store.ts
zharinov: Then we need to change the branches we do for platform-based commits

zharinov: Otherwise I guarantee we'll get a weird bug some day

lib/util/git/repo-store.ts
zharinov: Yes, I'm not sure it's okay or not

lib/util/git/repo-store.ts
rarkins: Yes, let's do that then. Have everything under `refs/renovate/<something>/*`

rarkins: BTW I thought you used `refs/renovate/branches/*` already

lib/util/git/repo-store.ts
rarkins: I wonder if it's better to create an orphan commit, with no parent, or to make it branch off the very first commit of the repo ever

lib/util/git/repo-store.ts
zharinov: Yes, but maybe let's have them under `refs/renovate/tmp/*`, to be clear they're subject to cleanup

PR #34326, opened 2025-02-20
fix: allow protobufjs as it is used in the hex datasource

secustor:
## Changes
Allow to build `protobufjs` as it is used in the `hex` datasource. 

### DIFF
--- a/package.json
+++ b/package.json
@@ -360,6 +360,7 @@
   "pnpm": {
     "onlyBuiltDependencies": [
       "better-sqlite3",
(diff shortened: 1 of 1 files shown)


rarkins (approved):

viceice:
@secustor this only checks some things and logs some warnings we don't care. so no need to call postinstall for this package

- https://github.com/protobufjs/protobuf.js/blob/master/scripts/postinstall.js
- #34286

PR #34344, opened 2025-02-20
chore: add prettier to pre-commit

viceice:
## Changes
We often forget to run prettier on changed files and it doesn't need much time on pre-commit

### DIFF
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,0 +1,3 @@
+{
+  "*": "prettier --ignore-unknown --write"
+}
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,6 +1,7 @@
 #!/bin/sh
 
-pnpm ls-lint
+ls-lint
+lint-staged
 
 BRANCH_NAME=$(git branch --show-current)
 
--- a/package.json
+++ b/package.json
@@ -336,6 +336,7 @@
     "jest-mock": "29.7.0",
     "jest-mock-extended": "3.0.7",
     "jest-snapshot": "29.7.0",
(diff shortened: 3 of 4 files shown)


rarkins (commented):
So if it's part of `pre-commit` then it will run when a user attempts to do a `git commit` (essentially delaying it)?

It's part of `lint-staged` so will it only run `prettier` on files which are staged?

If the `prettier` command is set `--write`, what happens next if something needs prettifying? Does it just get automatically added to staging and to the commit?

viceice:
> So if it's part of `pre-commit` then it will run when a user attempts to do a `git commit` (essentially delaying it)?

Yes, but the delay is a few ms only.

> It's part of `lint-staged` so will it only run `prettier` on files which are staged?

Yes
 
> If the `prettier` command is set `--write`, what happens next if something needs prettifying? Does it just get automatically added to staging and to the commit?

Required changes get staged and commited automatically. No user intervention required.

I use it on all my repos without issues.

.lintstagedrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "*": "prettier --ignore-unknown --write"
```
rarkins, line 2: Does this call `pnpm prettier` under the hood, or only work if `prettier` is globally installed?

viceice:
here's a more complex variant which also works fine
https://github.com/containerbase/base/blob/main/.lintstagedrc.json

rarkins:
What about for users who edit a markdown file using github UI or codespaces without node_modules. Will it fail silently and still let them submit a PR?

rarkins (approved):

PR #34357, opened 2025-02-20
feat(lib/modules/versioning): add Alpine (APK) and Chainguard tags versioning support

lyoung-confluent:
## Changes
Introduces a new versioning API `alpine` which can parse/compare the versioning schema produced by the [Alpine Package Keeper (APK)](https://wiki.alpinelinux.org/wiki/Alpine_Package_Keeper) used by [Alpine Linux](https://www.alpinelinux.org/) and [Wolfi/Chainguard OS](https://www.chainguard.dev/unchained/introducing-wolfi-the-first-linux-un-distro-designed-for-securing-the-software-supply-chain) _packages_ (not the OS version itself).

Additionally, it introduces a `chainguard` versioning API which extends the `alpine` versioning with support for the [unique tags](https://edu.chainguard.dev/chainguard/chainguard-images/features/unique-tags/) feature used by [Chainguard](https://www.chainguard.dev/) (`cgr.dev/...`) Docker/OCI images.

## Context
If you wish to use RenovateBot to upgrade the version of APKs, you can typically make it work using the `semver` or `regex` versioning schemes, but it won't handle more exotic versions correctly. At a high-level, APKs have a primary version (ex: `1.2.3`) as well as an "epoch" or "revision" (ex: `r1`) which is typically incremented each time the package maintainer makes a change that differs from the upstream.

For example, the [musl package](https://pkgs.alpinelinux.org/package/v3.21/main/x86_64/musl) currently has the version `1.2.5-r9`. This means the upstream "tag" of `musl` will be [v1.2.5](https://github.com/kraj/musl/tree/v1.2.5), however the alpine maintainers have made 9 subsequent additional patches/metadata changes ([in this case](https://gitlab.alpinelinux.org/alpine/aports/-/commit/efd4d5df7be0ea48d98512671f676fc5ceb14899) most recently to patch CVE-2025-26519). There are additional features/complexities supported by the versioning scheme such as release candidates (`_rc<num>`), proposed versions (`_p<num>`), and more. 

This PR adds `lib/modules/versioning/alpine` which implements the `VersioningApi` interface allowing RenovateBot to correctly parse and compare these APK versions. It uses [the same test data](https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/master/test/version.data) as the `apk` CLI, however a _few_ more exotic test cases are commented out as they do not correctly parse/sort right now.

In addition to the `alpine` versioning API, a `chainguard` versioning API is introduced which _extends_ the `alpine` versioning scheme. [Chainguard](https://www.chainguard.dev/) maintains distroless (and low/no CVE) Docker/OCI images for [many popular projects/images](https://images.chainguard.dev/). As these images have become quite popular we would like to be able to maintain/upgrade the tags via RenovateBot.

While Chainguard creates/updates the standard `latest` tag for each image which can be used via the existing Docker datasource, typically you want to specify an exact version/tag of an image to deploy/use. Chainguard images are generally versioned/tagged using the version of the core package the image is based on/for, as an example: The [cert-manager-webhook](https://images.chainguard.dev/directory/image/cert-manager-webhook/overview) image will be tagged directly based on the version (including the revision/epoch) of the `cert-manager` APK package it contains.

However, because an image contains multiple packages, not just the main package we need an additional version field that will be incremented anytime a dependency like `glibc` is updated but the package remains on the same version. This is handled by Chainguard’s [unique tags](https://edu.chainguard.dev/chainguard/chainguard-images/images-features/unique-tags/) feature which adds a (always incrementing) date timestamp when the image was built in the format -`%Y%m%d%H%M`. So using the [cert-manager-webhook](https://github.com/wolfi-dev/os/blob/main/cert-manager-1.14.yaml) image as an example, we could have the following tags:

* `1.14.2-r0-202402261115`: The first build of `1.14.2`
* `1.14.2-r0-202403060824`: A dependency like `glibc` is updated
* `1.14.2-r1-202403070335`: A CVE in `cert-manager` is identified/patched by Chainguard
* `1.14.2-r1-202403081212`: A dependency like `busybox` is updated

The `lib/modules/versioning/chainguard` API adds support for parsing/comparing this date suffix, as well as logic to avoid mixing `-dev` tags (contain `busybox` and other developer tools) with non `-dev` tags.

## Documentation

## How I've tested my work


MindTooth:
Regarding Chainguard, it’s important to preface that the unique tags are only for the commercial offerings (requires an account).  Free only usually have `latest` or `latest-dev`.

rarkins (changes requested):

PR #34379, opened 2025-02-21
fix(presets): Revert "fix(presets): actually group Node.js packages"

rarkins:
Reverts renovatebot/renovate#34244

### DIFF
--- a/lib/config/presets/internal/group.ts
+++ b/lib/config/presets/internal/group.ts
@@ -363,7 +363,6 @@ const staticGroups = {
     packageRules: [
       {
         commitMessageTopic: 'Node.js',
-        groupName: 'Node.js',
         matchDatasources: ['docker', 'node-version'],
         matchPackageNames: [
           '/(?:^|/)node$/', // node or ends with "/node, except those below"


rarkins:
@maxbrunet the original change causes an undesirable change of branch name

e.g. I saw a change from https://github.com/rarkins/nvm6/pull/5 to https://github.com/rarkins/nvm6/pull/7 in one of my test directories. Branch name changed from `renovate/node-20.x` to `renovate/node.js`

maxbrunet:
Yes, like the other groups, why is it undesirable?

rarkins:
Because it's lost the major version part

maxbrunet:
So if we set `branchTopic` on this preset, that'd be acceptable?

```json
"branchTopic": "node-{{{newMajor}}}{{#if separateMinorPatch}}{{#if isPatch}}.{{{newMinor}}}{{/if}}{{/if}}.x"
```

rarkins:
Why does it need to be a group name if it's all the same depName?

maxbrunet:
It's not al the same `depName`, the package can have the Docker registry as prefix: `/(?:^|/)node$/`

rarkins:
That can remain as packageName

maxbrunet:
hmm not sure what do you mean, let's say we have that:

```dockerfile
# Dockerfile
FROM node:22.14.0
FROM library/node:22.14.0
FROM docker.io/library/node:22.14.0
FROM public.ecr.aws/docker/library/node:22.14.0
```
And a `.nvmrc`
```
22.14.0
```
How would Renovate be able to group them?

rarkins:
Yes if depName is "Node.js" for all

viceice (approved):

PR #34392, opened 2025-02-22
chore: prepare prereleases

viceice:
## Changes
- create pre-releases on next branch
- create maintenance releases on `maint/...` branches

## Context
We want to publish some test packages for mend internal testing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,10 +26,11 @@ permissions:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.number || github.ref }}
-  cancel-in-progress: ${{ github.ref_name != 'main' }}
+  cancel-in-progress: true
 
 env:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -26,10 +26,11 @@ permissions:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.number || github.ref }}
-  cancel-in-progress: ${{ github.ref_name != 'main' }}
+  cancel-in-progress: true
```
rarkins, line 29: Would this rule be reverted later, or it's not actually necessary to begin with?

viceice: it's no longer necessary, semantic release will break anyways if a new commit is pushed to main, so we should cancel early. there is a small chance to cancel the job while a release is published, but i havn't seen it on our other repos.

rarkins (approved):

PR #34400, opened 2025-02-22
feat: new manager pixi

trim21:
## Changes

[Pixi](https://github.com/prefix-dev/pixi) is a manager that support conda ecosystem and pypi ecosystem, with lock file support.

This PR add pixi with currently only `isLockFileMaintenance` support.

I plan to add pypi package support and conda package support in following PRs, don't want this PR being to big

---

ref: https://github.com/renovatebot/renovate/discussions/30243

local tests with gitea repo:

image


trim21:
do we do not want to use `__fixtures__` directory? I saw ci error with `❌ Error: More files have been added to '__fixtures__' directories than deleted.`

secustor:
> I saw ci error with `❌ Error: More files have been added to '__fixtures__' directories than deleted.`
> 
> Does it mean we shoudl avoid using `__fixtures__` and embedding them in code?

We are trying to avoid fixtures in separate files as vulnerability detection tools often are triggered because of them. This as by nature Renovate will have outdated package files and lockfiles in its fixtures. 
Also IIRC there are performance gains while running jest tests.

lib/modules/manager/pixi/lockfile.ts
```diff
@@ -0,0 +1,32 @@
...
+  { lockVersion: 6, range: '>=0.39.2, <=0.41.4', cmd: commandLock },
+] as const;
+
+interface Config {
+  lockVersion: number;
+  range: string;
+  cmd: string;
+}
+
+export function pickConfig(lockVersion?: number): Config | undefined {
```
secustor, line 20:
```suggestion
export function pickConfig(lockVersion?: number): Config | null {
```
Please do not return `undefined`, we use `null` if it is returned explicitly

lib/modules/manager/pixi/extract.ts
```diff
@@ -0,0 +1,58 @@
...
+  const lockfileName = getSiblingFileName(packageFile, 'pixi.lock');
+  const lockFiles = [];
+  if (await localPathExists(lockfileName)) {
+    lockFiles.push(lockfileName);
+  }
+
+  return {
+    lockFiles,
+    deps: [], // TODO
+    fileFormat: 'toml',
```
secustor, line 56:
```suggestion
    deps: [],
```
`fileFormat` is only used in `customManagers`

lib/modules/manager/pixi/extract.ts
```diff
@@ -0,0 +1,58 @@
...
+): Promise<PackageFileContent | null> {
+  logger.trace(`pixi.extractPackageFile(${packageFile})`);
+
+  const config = getUserPixiConfig(content, packageFile);
+  if (!config) {
+    return null;
+  }
+
+  const lockfileName = getSiblingFileName(packageFile, 'pixi.lock');
+  const lockFiles = [];
```
secustor, line 48:
```suggestion
  const lockFiles: string[] = [];
```
no implicit any

lib/modules/manager/pixi/extract.ts
```diff
@@ -0,0 +1,58 @@
...
+import { logger } from '../../../logger';
+import { getSiblingFileName, localPathExists } from '../../../util/fs';
+import { Result } from '../../../util/result';
+import type { PackageFileContent } from '../types';
+import { type PixiConfigSchema, PixiToml, PyprojectToml } from './schema';
+
+function getUserPixiConfig(
+  content: string,
+  packageFile: string,
+): null | z.infer<typeof PixiConfigSchema> {
```
secustor, line 11:
```suggestion
): null | PixiConfigSchema {
```
Infer this in `schema.ts`

lib/modules/manager/pixi/index.ts
```diff
@@ -0,0 +1,32 @@
...
+};
+
+export const supportedDatasources = [
+  PypiDatasource.id,
+  GithubTagsDatasource.id,
+  GithubReleasesDatasource.id,
+  GitlabTagsDatasource.id,
+  GitRefsDatasource.id,
+  GitTagsDatasource.id,
+];
```
secustor, line 32:
```suggestion
export const supportedDatasources = [];
```

secustor (changes requested):

lib/modules/manager/pixi/index.ts
trim21: So we fo not need to define this, empty arrat works fine?

lib/modules/manager/pixi/schema.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export const PixiConfigSchema = z.union([z.object({}), z.undefined()]);
+
+export const PyprojectSchema = z
+  .object({
+    tool: z.object({ pixi: PixiConfigSchema }),
+  })
+  .transform(({ tool: { pixi } }) => {
+    return pixi ?? null;
+  });
```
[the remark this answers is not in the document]
trim21: I'd suggestion to not extend pep621 schema, even it's a pep621 tool section.

I plan to add pixi only transform in this schema in the future.

lib/modules/manager/pixi/index.ts
trim21: This suggestion failed tests:

```
   × modules/manager/index > supportedDatasources > has valid supportedDatasources for pixi 10ms
     → expected [] to be an array with at least one item
```

lib/modules/manager/pixi/schema.ts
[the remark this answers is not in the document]
trim21: I plan to use a AST based replacemanet for pixi manager so I need to store additional data in managerData and do a transform when pasing pyproject


https://github.com/trim21/renovate/blob/f9d820b8c2f6d18e1e01d58f8ce166de61eee212/lib/modules/manager/pixi/schema.ts#L394-L405

https://github.com/trim21/renovate/blob/f9d820b8c2f6d18e1e01d58f8ce166de61eee212/lib/modules/manager/pixi/update.ts#L8-L21

https://github.com/trim21/renovate/blob/f9d820b8c2f6d18e1e01d58f8ce166de61eee212/lib/util/toml.ts#L97

lib/util/exec/containerbase.ts
```diff
@@ -171,6 +171,11 @@ const allToolConfig: Record<string, ToolConfig> = {
     packageName: 'pnpm',
     versioning: npmVersioningId,
   },
+  pixi: {
+    datasource: 'github-releases',
+    packageName: 'prefix-dev/pixi',
+    versioning: pep440VersioningId,
```
viceice, line 177: are you sure with that versioning? 🤔 we usually use semver for such GitHub packages.

lib/modules/manager/pixi/extract.ts
```diff
@@ -0,0 +1,56 @@
...
+    const { val, err } = Result.parse(content, PixiToml).unwrap();
+    if (err) {
+      logger.debug({ packageFile, err }, `error parsing ${packageFile}`);
+      return null;
+    }
+
+    return val;
+  }
+
+  /* v8 ignore next 2 */
```
viceice, line 31: add a test instead

lib/modules/manager/pixi/extract.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const lockfileName = getSiblingFileName(packageFile, 'pixi.lock');
+  const lockFiles: string[] = [];
+  if (await localPathExists(lockfileName)) {
+    lockFiles.push(lockfileName);
+  }
+
+  return {
+    lockFiles,
+    deps: [],
```
viceice, line 54: renovate won't run update artifacts without dependencies.

viceice (changes requested):

lib/util/exec/containerbase.ts
trim21: yes: https://github.com/prefix-dev/pixi/releases/tag/v0.33.0rc1

lib/modules/manager/pixi/schema.ts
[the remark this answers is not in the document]
secustor: There is nothing prevent you from adding then a transform on the top object. 

The AST part has to be discussed, because I seen no reason why you have to go a separate route rather than other implementations.

lib/modules/manager/pixi/extract.ts
trim21: this is the code path that expected to failed, should I add a error here can check it?

lib/modules/manager/pixi/index.ts
secustor: This field is for documentation purposes https://docs.renovatebot.com/modules/manager/pep621/#supported-datasources
so it is fine if you add the pypi datasource

lib/modules/manager/pixi/extract.ts
trim21: This is not true.

image
image

lib/modules/manager/pixi/extract.ts
secustor: Simply test for a file which does not match the the pattern above. 

Users can override the fileMatch and will then run into this.

lib/modules/manager/pixi/extract.ts
viceice: okay 🤔

lib/modules/manager/pixi/extract.ts
trim21: can you explain how? I can think many cases here

For example, `fileMatch: ['pixi.*.toml', 'pyproject.*.toml']`? or a magic toml file pattern? should we try each case? or just still return null?

secustor: return `null` will do until the need arises as parsing every file is costly. 
If the need arises this can be changed later.

trim21: I add a fallback to first try to parse config as pyproject.toml then pixi

secustor (dismissed):

trim21:
Does this PR need something else from me？

secustor:
`@viceice`s review is blocking the PR at the moment. All maintainers which requested changes at one point have to approve the PR.

secustor (dismissed):

lib/modules/manager/pixi/index.ts
```diff
@@ -0,0 +1,21 @@
...
+export const categories: Category[] = ['python'];
+
+export const defaultConfig = {
+  fileMatch: [
+    '(^|/)pyproject\\.toml$', // `tool.pixi` section
+    '(^|/)pixi\\.toml$', // root object
+  ],
+};
+
+export const supportedDatasources = [PypiDatasource.id, CondaDatasource.id];
```
viceice, line 21:
```suggestion
```
unsed yet

lib/modules/manager/pixi/artifacts.ts
```diff
@@ -0,0 +1,100 @@
...
+  const existingLockFileContent = await readLocalFile(lockFileName, 'utf8');
+  if (!existingLockFileContent) {
+    logger.debug(`No lock file found`);
+    return null;
+  }
+  logger.debug(`Updating ${lockFileName}`);
+
+  const { val } = Result.parse(existingLockFileContent, LockfileYaml).unwrap();
+  const cfg = pickConfig(val?.version);
+  const constraint = cfg?.range;
```
viceice, line 45: prefer `config.constraints?.pixi` over this. so only read and parse lockfile if no user constrains are there

lib/modules/manager/pixi/artifacts.ts
```diff
@@ -0,0 +1,100 @@
...
+      userConfiguredEnv: config.env,
+      toolConstraints: [{ toolName: 'pixi', constraint }],
+    };
+    await exec(cmd, execOptions);
+    const newPixiLockContent = await readLocalFile(lockFileName, 'utf8');
+    if (existingLockFileContent === newPixiLockContent) {
+      logger.debug(`${lockFileName} is unchanged`);
+      return null;
+    }
+    logger.debug(`Returning updated ${lockFileName}`);
```
viceice, line 73:
```suggestion
```

lib/modules/manager/pixi/artifacts.ts
```diff
@@ -0,0 +1,100 @@
...
+          contents: newPixiLockContent,
+        },
+      },
+    ];
+  } catch (err) {
+    /* v8 ignore start */
+    if (err.message === TEMPORARY_ERROR) {
+      throw err;
+    }
+    /* v8 ignore end */
```
viceice, line 88: add a test, simply throw `new Error(TEMPORARY_ERROR)` from `writeLocalFile`

lib/modules/manager/pixi/extract.ts
```diff
@@ -0,0 +1,69 @@
...
+  };
```
viceice, line 68:
```suggestion
  if (!await localPathExists(lockfileName)) {
    return null;
  }

  return {
    lockFiles: [lockfileName],
    deps: [],
  };
```
return null if no deps and no lockfile

lib/modules/manager/pixi/extract.ts
```diff
@@ -0,0 +1,69 @@
...
+import type { PackageFileContent } from '../types';
+import { type PixiConfig, PixiToml } from './schema';
+
+const PyProjectToml = Toml.pipe(PyProjectSchema);
+
+function getUserPixiConfig(
+  content: string,
+  packageFile: string,
+): null | PixiConfig {
+  if (packageFile.endsWith('pyproject.toml')) {
```
viceice, line 15:
```suggestion
  if (packageFile === 'pyproject.toml' || packageFile.endsWith('/pyproject.toml')) {
```
make it safer to not match `some-pyproject.toml`

lib/modules/manager/pixi/extract.ts
```diff
@@ -0,0 +1,69 @@
...
+    const { val, err } = Result.parse(content, PyProjectToml).unwrap();
+    if (err) {
+      logger.debug({ packageFile, err }, `error parsing ${packageFile}`);
+      return null;
+    }
+
+    return val.tool?.pixi ?? null;
+  }
+
+  if (packageFile.endsWith('pixi.toml')) {
```
viceice, line 25:
```suggestion
  if (packageFile === 'pixi.toml' || packageFile.endsWith('/pixi.toml')) {
```

lib/modules/manager/pixi/extract.ts
```diff
@@ -0,0 +1,69 @@
...
+    content,
+    PyProjectToml,
+  ).unwrap();
+
+  if (maybePyprojectConfig?.tool?.pixi) {
+    return maybePyprojectConfig.tool.pixi;
+  }
+
+  const { val: maybePixiConfig } = Result.parse(content, PixiToml).unwrap();
+  return maybePixiConfig ?? null;
```
viceice, line 45:
```suggestion
  const { val, err } = Result.parse(
    content,
    z.union([PyProjectToml.transform(p => p.tool?.pixi),PixiToml])
  ).unwrap();

  if (err) {
    logger.debug({ packageFile, err }, `error parsing ${packageFile}`);
    return null;
  }
  return val ?? null;
```

lib/modules/manager/pixi/extract.ts
trim21: we will return deps in the futhre, and it will cause unnecessary diff then

lib/modules/manager/pixi/index.ts
trim21: also it can't be an empty array

lib/modules/manager/pixi/extract.ts
trim21: also it make us hard to test if it's unsupport toml or can't find anything.

lib/modules/manager/pixi/artifacts.ts
```diff
@@ -0,0 +1,106 @@
...
+    return null;
+  }
+
+  const lockFileName = getSiblingFileName(packageFileName, 'pixi.lock');
+  const existingLockFileContent = await readLocalFile(lockFileName, 'utf8');
+  if (!existingLockFileContent) {
+    logger.debug(`No lock file found`);
+    return null;
+  }
+  logger.debug(`Updating ${lockFileName}`);
```
viceice, line 40:
```suggestion
  logger.trace(`Updating ${lockFileName}`);
```

lib/modules/manager/pixi/extract.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  const { val, err } = Result.parse(
+    content,
+    z.union([PixiToml, PyProjectToml.transform((p) => p.tool?.pixi)]),
+  ).unwrap();
+
+  if (err) {
+    logger.debug({ packageFile, err }, `error parsing ${packageFile}`);
+    return null;
```
viceice, line 46: needs a test

lib/modules/manager/pixi/artifacts.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+          config: { ...config, isLockFileMaintenance: true },
+        }),
+      ).rejects.toThrow(new Error(TEMPORARY_ERROR));
+    });
+
+    it('returns updated pixi.lock using docker', async () => {
+      GlobalConfig.set({
+        ...adminConfig,
+        binarySource: 'docker',
+        dockerSidecarImage: 'ghcr.io/containerbase/sidecar',
```
viceice, line 145: move to `adminConfig` object, so it doesn't need to be repeated in future

lib/modules/manager/pixi/artifacts.spec.ts
```diff
@@ -0,0 +1,325 @@
+import { join } from 'upath';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { mockDeep } from 'vitest-mock-extended';
+import { envMock, mockExecAll } from '../../../../test/exec-util';
+import { env, fs } from '../../../../test/util';
+import { GlobalConfig } from '../../../config/global';
+import type { RepoGlobalConfig } from '../../../config/types';
+import { TEMPORARY_ERROR } from '../../../constants/error-messages';
+import * as docker from '../../../util/exec/docker';
+import * as _hostRules from '../../../util/host-rules';
```
viceice, line 10:
```suggestion
```

lib/modules/manager/pixi/artifacts.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+
+vi.mock('../../../util/exec/env');
+vi.mock('../../../util/fs');
+vi.mock('../../datasource', () => mockDeep());
+vi.mock('../../../util/host-rules', () => mockDeep());
+
+process.env.CONTAINERBASE = 'true';
+
+const datasource = vi.mocked(_datasource);
+const hostRules = vi.mocked(_hostRules);
```
viceice, line 52:
```suggestion
```

lib/modules/manager/pixi/artifacts.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+  dockerSidecarImage: 'ghcr.io/containerbase/sidecar',
+};
+
+const config: UpdateArtifactsConfig = {};
+
+describe('modules/manager/pixi/artifacts', () => {
+  describe('updateArtifacts', () => {
+    beforeEach(() => {
+      env.getChildProcessEnv.mockReturnValue(envMock.basic);
+      hostRules.getAll.mockReturnValue([]);
```
viceice, line 67:
```suggestion
```

do not mock if you don't need to 😏 I know it's copy'n'paste 🤗

lib/modules/manager/pixi/artifacts.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const lockFileName = getSiblingFileName(packageFileName, 'pixi.lock');
+  const existingLockFileContent = await readLocalFile(lockFileName, 'utf8');
+  if (!existingLockFileContent) {
+    logger.debug(`No lock file found`);
+    return null;
+  }
+  logger.trace(`Updating ${lockFileName}`);
+
+  const cmd: string[] = [commandLock];
```
viceice, line 42:
```suggestion
  const cmd = [commandLock];
```

typescript knows it's a string array

lib/modules/manager/pixi/extract.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+    it('returns parse pixi.toml', async () => {
+      expect(await extractPackageFile(pixiToml, 'pixi.toml')).toMatchObject({
+        deps: [],
+        lockFiles: [],
+      });
+    });
+
+    it('returns parse pixi section from pyproject.toml', async () => {
+      fs.getSiblingFileName.mockReturnValueOnce('pixi.lock');
+      fs.localPathExists.mockReturnValueOnce(Promise.resolve(true));
```
viceice, line 107:
```suggestion
      fs.localPathExists.mockResolvedValueOnce(true);
```

lib/modules/manager/pixi/artifacts.ts
trim21: yes it's a legacy type hint, it used to be `cmd.push(commandLock)`

viceice (approved):
please use batch commit from pr files tab to combine multiple suggestions next time

PR #34457, opened 2025-02-25
feat(presets): add `twig` monorepo

herndlm:
## Changes

groups the PHP twig packages, which seem to be a fake monorepo, see https://packagist.org/?query=twig%2F (e.g. https://packagist.org/packages/twig/twig and https://packagist.org/packages/twig/extra-bundle)

## Context

realized that renovate wanted to create 2 MRs for 2 different twig packages from the same versions a to b.


lib/data/monorepo.json
```diff
@@ -648,6 +648,7 @@
     "spock": "/^org\\.spockframework:spock-/",
     "syncfusion-dotnet": "/^Syncfusion\\./",
     "testing-library": "/^@testing-library//",
+    "twig": "/^twig//",
```
rarkins, line 651: This seems very generic as it will apply across ecosystems

rarkins: I think here makes most sense: https://github.com/renovatebot/renovate/blob/3169d1a9dd55e2003a25cc24943709b7ad63bc55/lib/data/monorepo.json#L619-L627

herndlm: looks like a good place indeed. adapted, thx!

rarkins (approved):

herndlm:
working, thx!

image

PR #34530, opened 2025-02-27
feat(util): a AST based toml string replacer

trim21:
## Changes

Add a new util to replace string value in toml based on AST to keey whitespace style.

It will be possible to replace a string value in toml based on object path.

for example:

```ts
import * as toml from './lib/util/toml';

const updated = toml.replaceString(
  `
[project]
dependencies = [
  'requests=1'
]
`,
  ['project', 'dependencies', 0],
  (currentSpec) => {
    return currentSpec.replace('=1', '=2');
  },
);


// OR

toml.replaceString(
  `
[project.dependencies]
requests = {version = '*'}
`,
  ['project', 'dependencies', 'requests', 'version'],
  () => {
    return '==0.2.7';
  },
);
```

I plan to use this in pixi manager like this:

```ts
import { replaceString as replaceTomlString } from '../../../util/toml';
import type { UpdateDependencyConfig } from '../types';
import type { PixiManagerData } from './schema';

type PixiUpgrade = UpdateDependencyConfig<{path: (string|number)[]}>;

export function updateDependency({
  fileContent,
  upgrade,
}: PixiUpgrade): string | null {
  if (is.nullOrUndefined(upgrade.managerData)) {
    return fileContent;
  }

  return replaceTomlString(
    fileContent,
    upgrade.managerData.path,
    () => upgrade.newValue!,
  );
}
```

### DIFF
--- a/lib/util/toml.spec.ts
+++ b/lib/util/toml.spec.ts
@@ -1,5 +1,6 @@
 import { codeBlock } from 'common-tags';
-import { parse as parseToml } from './toml';
+import { parseTOML } from 'toml-eslint-parser';
+import { getSingleValue, parse as parseToml, replaceString } from './toml';
 
 describe('util/toml', () => {
   it('works', () => {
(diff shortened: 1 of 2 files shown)


viceice (changes requested):
can you clarify why the normal auto replace from renovate doesn't work?

PR #34532, opened 2025-02-27
feat(git): Support baseBranches in fork mode

janhoy:
## Changes

Building upon #34467, this PR adds code that makes `baseBranches` feature work also in forking mode:

* Extend `scm` interface with `syncForkBranch()`
* Added new `util/git` methods:
  * `checkoutBranchFromRemote()`
  * `resetHardFromRemote()`
  * `forcePushToRemote()`
  * `localBranchExists()`
  * `syncForkBranch()` which syncs `<baseBranch>` from `upstream` to local and force-push to `origin`
* Refactor `syncGit()` upstream handling to use the new `syncForkBranch()` instead of using individual git commands
* Call `syncForkBranch()` in `extractDependencies()` inside the loop for each baseBranch
* Removed disclaimer from documentation that forkMode does not support baseBranchews

PS: This PR includes the commit from #34467 for completeness, i.e. that PR can be merged first if you so choose.

## Context

See #7850 for discussion and context.

TODO: Add unit tests

**Public repos**:

I used dummy repo: https://github.com/cominvent/dummy for testing. It contains two branches `main` and `dev`, both having a Dockerfile with out-of-date FROM tag. Ran renovate locally in my IDE with a token that forks to https://github.com/janhoy/cominvent-_-dummy . You can see how the forked repo has both branches, and that the upstream repo has two PR's created, one for each branch.


lib/workers/repository/process/index.ts
```diff
@@ -130,6 +130,16 @@ export async function extractDependencies(
     const extracted: Record<string, Record<string, PackageFile[]>> = {};
     for (const baseBranch of config.baseBranches) {
       addMeta({ baseBranch });
+
+      // TODO: initialize these two variables
+      const isForkMode = true;
+      const isGitHub = true;
```
janhoy, line 136: How to determine if we are in forkMode and that the scm in use in in fact git(hub)?

I think that if we are in forkMode but platform is non-git, we should log an error and abort since we cannot bring fork up to date. Perhaps this will happen automatically for scm implementations that do not implement `syncForkBranch()`?

janhoy: So I added a function `scm.hasUpstream()` as a means to determine if we are in forkingMode. It would be easier if we could just read some boolean from `config` object, but I looked but `config.isFork()` seems to indicate whether a discovered repo is a fork and should be skipped, while `config.forkProcessing` seems to be mostly for disabling fork processing. Also I cannot find the `upstreamUrl` variable used during git init from `LocalRepoConfig` does not seem to be available from the process worker. Anything I missed?

lib/util/git/index.ts
```diff
@@ -485,6 +488,24 @@ export async function syncGit(): Promise<void> {
     config.defaultBranch ??
     (await getDefaultBranch(git));
   delete getCache()?.semanticCommits;
+  // istanbul ignore if
+  if (config.upstreamUrl) {
+    logger.debug(
+      `Bringing default branch up to date with upstream to get latest config`,
```
HonkingGoose, line 494:
```suggestion
      `Bringing default branch up-to-date with upstream, to get latest config`,
```

lib/util/git/index.ts
```diff
@@ -1372,3 +1450,52 @@ export async function listCommitTree(
...
+/**
+ * Synchronizes a forked branch with its upstream counterpart.
+ *
+ * This function ensures that the specified branch in the forked repository is up-to-date
+ * with the corresponding branch in the upstream repository. It performs the following steps:
+ * 1. Checks if the branch exists locally.
+ * 2. If the branch exists locally, it checks out the branch.
+ * 3. If the branch does not exist locally, it checks out the branch from the upstream repository.
+ * 4. Resets the local branch to match the upstream branch.
+ * 5. Force pushes the updated branch to the origin repository.
```
HonkingGoose, line 1469:
```suggestion
 * The syncForkWithUpstream function ensures the fork's branch is up-to-date with the corresponding branch in the upstream repository.
* The steps are:
 * 1. Check if the branch exists locally.
 * 2. If the branch exists locally: check out the branch.
 * 3. If the branch does _not_ exist locally: checkout the upstream branch.
 * 4. Reset the local branch to match the upstream branch.
 * 5. Force push the updated branch to the origin repository.
```

### About my rewrite

Here's a small rewrite to the comment to:

- Use simple words and phrases
- Use one sentence per line (this makes it easier to move/change bits around later, and helps with the git diff)
- Write short sentences
- Put the name of the function in the comment
- Use `if: then` formatting for steps that rely on a condition to be true
- Emphasize the _not_ part in a sentence

### Questions

After reading just the comment, I don't understand:

- Does step 4 "Reset the local branch to match the upstream branch" _always_ happen, or does step 4 depend on the outcome of step 2/3?
- What is the "updated branch" that's pushed to the _origin_ repository in step 5. Is it the local branch, or the upstream branch?

lib/util/git/index.ts
```diff
@@ -1372,3 +1450,52 @@ export async function listCommitTree(
...
+  try {
+    await git.fetch(['upstream']);
+    if (await localBranchExists(branchName)) {
+      await checkoutBranch(branchName);
+    } else {
+      await checkoutBranchFromRemote(branchName, `upstream`);
+    }
+    await resetHardFromRemote(`upstream/${branchName}`);
+    await forcePushToRemote(branchName, 'origin');
+    // Get long git sha
```
HonkingGoose, line 1490:
```suggestion
    // Get long Git SHA
```

Git is a proper noun (a named thing), so should be capitalized. And abbreviations like SHA should be capitalized in full. 😉

lib/util/git/index.ts
```diff
@@ -1372,3 +1450,52 @@ export async function listCommitTree(
...
+ * This function ensures that the specified branch in the forked repository is up-to-date
+ * with the corresponding branch in the upstream repository. It performs the following steps:
+ * 1. Checks if the branch exists locally.
+ * 2. If the branch exists locally, it checks out the branch.
+ * 3. If the branch does not exist locally, it checks out the branch from the upstream repository.
+ * 4. Resets the local branch to match the upstream branch.
+ * 5. Force pushes the updated branch to the origin repository.
+ *
+ * @param {string} branchName - The name of the branch to synchronize.
+ * @returns {Promise<LongCommitSha>} - A promise that resolves to  if the synchronization is successful, or `false` if an error occurs.
```
HonkingGoose, line 1472:
```suggestion
 * @returns {Promise<LongCommitSha>} - A promise that resolves to True if the synchronization is successful, or `false` if an error occurs.
```

The comment is missing a bit of text. I suspect the promise resolves to `True?`

HonkingGoose (changes requested):

lib/util/git/index.ts
janhoy: > Does step 4 "Reset the local branch to match the upstream branch" always happen, or does step 4 depend on the outcome of step 2/3?

Currently we always do a git reset to upstream branch. If we for some reason had an existing local repo with a stale local branch, just checking it out won't bring it up to date. If we fetch from upstream for the first time, it is not necessary to "git reset", but again, it is mostly a NOOP so does not hurt.

> What is the "updated branch" that's pushed to the origin repository in step 5. Is it the local branch, or the upstream branch?

It will be the local branch, which is just reset to match "upstream". So we fetch/pull from "upstream" to local, then push (force) to origin which is our fork. It has same effect as calling the GitHub API we did before which brought the fork up to date on github, before pulling.

lib/workers/repository/process/index.ts
```diff
@@ -130,6 +130,13 @@ export async function extractDependencies(
     const extracted: Record<string, Record<string, PackageFile[]>> = {};
     for (const baseBranch of config.baseBranches) {
       addMeta({ baseBranch });
+
+      // If we are in forking mode, sync the baseBranch with upstream remote
+      if (await scm.hasUpstream()) {
+        await scm.syncForkWithUpstream(baseBranch);
+        logger.info(`Synced base branch ${baseBranch} of fork with upstream`);
+      }
```
rarkins, line 138:
```suggestion
      await scm.syncForkWithUpstream(baseBranch);
```

lib/modules/platform/local/scm.ts
```diff
@@ -59,4 +59,12 @@ export class LocalFs implements PlatformScm {
...
     return Promise.resolve();
   }
+
+  syncForkWithUpstream(baseBranch: string): Promise<LongCommitSha> {
+    throw new Error('Cannot sync fork in local mode');
+  }
+
+  hasUpstream(): Promise<boolean> {
+    return Promise.resolve(false);
+  }
```
rarkins, line 69:
```suggestion
```

lib/workers/repository/process/index.ts
rarkins: Push the "hasStream()" logic down to platform layer

lib/modules/platform/default-scm.ts
```diff
@@ -48,4 +48,13 @@ export class DefaultGitScm implements PlatformScm {
...
   }
+
+  async syncForkWithUpstream(branchName: string): Promise<LongCommitSha> {
+    return await git.syncForkWithUpstream(branchName);
+  }
+
+  async hasUpstream(): Promise<boolean> {
+    const remotes = await git.getRemotes();
+    return remotes.includes('upstream');
+  }
```
rarkins, line 59:
```suggestion
```

lib/modules/platform/default-scm.ts
```diff
@@ -48,4 +48,13 @@ export class DefaultGitScm implements PlatformScm {
   mergeToLocal(branchName: string): Promise<void> {
     return git.mergeToLocal(branchName);
   }
+
+  async syncForkWithUpstream(branchName: string): Promise<LongCommitSha> {
+    return await git.syncForkWithUpstream(branchName);
+  }
```
rarkins, line 54:
```suggestion
  async syncForkWithUpstream(branchName: string): Promise<void> {
    const remotes = await git.getRemotes();
    if (remotes?.includes('upstream') {
      return await git.syncForkWithUpstream(branchName);
    }
  }
```

lib/modules/platform/local/scm.ts
rarkins, line 69:
```suggestion
```

lib/modules/platform/local/scm.ts
```diff
@@ -59,4 +59,12 @@ export class LocalFs implements PlatformScm {
   mergeToLocal(branchName: string): Promise<void> {
     return Promise.resolve();
   }
+
+  syncForkWithUpstream(baseBranch: string): Promise<LongCommitSha> {
+    throw new Error('Cannot sync fork in local mode');
+  }
```
rarkins, line 65:
```suggestion
  syncForkWithUpstream(baseBranch: string): Promise<void> {
    return Promise.resolve();
  }
```

rarkins (changes requested):

lib/util/git/index.ts
```diff
@@ -1372,3 +1450,52 @@ export async function listCommitTree(
...
+* The steps are:
+ * 1. Check if the branch exists locally.
+ * 2. If the branch exists locally: check out the branch.
+ * 3. If the branch does _not_ exist locally: checkout the upstream branch.
+ * 4. Reset the local branch to match the upstream branch.
+ * 5. Force push the updated branch to the origin repository.
+ *
+ * @param {string} branchName - The name of the branch to synchronize.
+ * @returns {Promise<LongCommitSha>} - A promise that resolves to True if the synchronization is successful, or `false` if an error occurs.
+ */
```
HonkingGoose, line 1473:
```suggestion
/**
 * Synchronize a forked branch with its upstream counterpart.
 *
 * syncForkWithUpstream updates the fork's branch, to match the corresponding branch in the upstream repository.
* The steps are:
 * 1. Check if the branch exists locally.
 * 2. If the branch exists locally: checkout the local branch.
 * 3. If the branch does _not_ exist locally: checkout the upstream branch.
 * 4. Reset the local branch to match the upstream branch.
 * 5. Force push the (updated) local branch to the origin repository.
 *
 * @param {string} branchName - The name of the branch to synchronize.
 * @returns {Promise<LongCommitSha>} - A promise that resolves to True if the synchronization is successful, or `false` if an error occurs.
 */
```

Comments:

- Rewrite the intro bits again, to shorten the text
- Fix wrong whitespace in the verb checkout
- Clarify which branch is the object of the sentence in the list items
- Put `(updated)` in parenthesises (I think we only update the local branch if we actually need to)?

HonkingGoose (changes requested):

lib/util/git/index.ts
HonkingGoose: Thank you for explaining how the code works, and which branches each steps targets. My new suggestion puts this extra info in the code comment. 😄 

After reviewing and applying my new suggestion, a maintainer can mark this discussion as resolved. 😉

janhoy: Thanks for the attention to details and understandable docs.

janhoy:
Thanks for the reviews. I merged it all and ran prettier etc.
How do we approach testing?

lib/util/git/types.ts
```diff
@@ -19,6 +19,7 @@ export interface StorageConfig {
   currentBranch?: string;
   defaultBranch?: string;
   url: string;
+  upstreamUrl?: string;
```
janhoy, line 22: The ["build-docker" test](https://github.com/renovatebot/renovate/actions/runs/13776159815/job/38525771751?pr=34532) fails with a message 
> "message": "remote: Not Found\nfatal: repository 'https://github.com/undefined.git/' not found\n"
I believe that is because the `upstreamUrl` variable is always passed in to `git.initRepo()`, even if `undefined`, but the `upstreamUrl?: string` declaration here only catches cases where the property is not passed in at all.

I believe the fix is to modify the type here.
```suggestion
  upstreamUrl?: string | undefined;
```

Not tested, thoughts? There are subtle differences between not-provided, null and undefined in TS....

janhoy: Tested, but same error. Will need to Reproduce locally and find a working fix.

rarkins:
@RahulGautamSingh can you fork off and add tests to this PR?

lib/workers/repository/process/index.ts
```diff
@@ -130,6 +130,10 @@ export async function extractDependencies(
     const extracted: Record<string, Record<string, PackageFile[]>> = {};
     for (const baseBranch of config.baseBranches) {
       addMeta({ baseBranch });
+
+      if (scm.syncForkWithUpstream) {
+        await scm.syncForkWithUpstream(baseBranch);
+      }
```
RahulGautamSingh, line 136: We could also check `forkToken` here, so we do not to call `git.getRemotes` & fetch usptream each time when we are processing non-fork cases. 
```suggestion

      if (config.forkToken && scm.syncForkWithUpstream) {
        await scm.syncForkWithUpstream(baseBranch);
      }
```

PR #34564, opened 2025-03-01
fix(manager/gomod): perfer to use go version defined as toolchain to update artifacts

trim21:
## Changes

prefer use version from toolchain directive instead of go diective in go.mod, this will avoid change current toolchain directive.

## Context

As explained in https://github.com/renovatebot/renovate/discussions/34563, when running `go get ./...`, go will update toolchain directive in go.mod, which is a `depType=toolchain` change instead of expected `depType=require` change.

And there are 2 ways to avoid change of toolchain directive.

I don't think old version of go support `go get ./... toolchain@${current toolchain}` (just like we still didn't remove `-d` flag here), so it would be more backward compatible to just use version of current toolchain to avoid this.

If go.mod doesn't have such directive, we still use version frmo go directive.


rarkins (changes requested):
Need to consider the edge case where the toolchain version is being upgraded in the same PR

trim21:
So I think in that case we do not parse GoConstraints but use updated toolchain version from `updatedDeps`?

trim21:
Is it possible to update go.mod first then get constraints? this should avoid this problem naturally

rarkins:
By default renovate doesn't bump the go.mod directive. This was the recommendation/request of the golang team. So 99% of the time it's probably only the tool chain getting bumped, or none. I agree with you that we shouldn't "accidentally" bump the tool chain by running a latest go

viceice:
I think we simply need to use new file content to extract this information instead of reading old file from disk here.

https://github.com/renovatebot/renovate/blob/ad5d318c3d2fdd5ecbea2bac81b1d03860d1d802/lib/modules/manager/gomod/artifacts.ts#L194

trim21:
> I think we simply need to use new file content to extract this information instead of reading old file from disk here.
> 
> https://github.com/renovatebot/renovate/blob/ad5d318c3d2fdd5ecbea2bac81b1d03860d1d802/lib/modules/manager/gomod/artifacts.ts#L194

Yes, this is what I suggest

trim21:
> By default renovate doesn't bump the go.mod directive. This was the recommendation/request of the golang team. So 99% of the time it's probably only the tool chain getting bumped, or none. I agree with you that we shouldn't "accidentally" bump the tool chain by running a latest go

I don't mean go directive, I mean if we need to update `toolchain` directive, we update go.mod first then parse updated toolchain directive from go.mod.

I don't know if it's possible.

trim21:
it should be something like this?

```js
  const goConstraints =
    config.constraints?.go ?? getGoConstraints(massagedGoMod);
```

rarkins:
Yes. Any updates to go.mod would have been written out already

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -191,7 +191,7 @@ export async function updateArtifacts({
     }
   }
   const goConstraints =
-    config.constraints?.go ?? (await getGoConstraints(goModFileName));
+    config.constraints?.go ?? getGoConstraints(massagedGoMod);
```
viceice, line 194: let's do this refactor in a separate PR

trim21: I'm confused, do you want initial commit? ad5d318

trim21: What does this mean?
Screenshot_2025-03-02-04-33-24-035_com brave browser

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -451,6 +451,14 @@ async function getGoConstraints(
   if (!content) {
     return undefined;
   }
+
+  // prefer toolchain directive when go.mod has one
+  const toolchain = regEx(/^toolchain\s*go(?<gover>\d+\.\d+\.\d+)$/m);
+  const toolchainMatch = toolchain.exec(content);
+  if (toolchainMatch?.groups?.gover) {
+    return toolchainMatch.groups.gover;
+  }
```
rarkins, line 460:
```suggestion
  const toolchainVer = toolchainMatch?.groups?.gover;
  if (toolchainVer) {
    logger.debug(`Using go version ${toolchainVer) found in toolchain directive`);
    return toolchainVer;
  }
```

rarkins (changes requested):

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -451,6 +451,16 @@ async function getGoConstraints(
   if (!content) {
     return undefined;
   }
+
+  // prefer toolchain directive when go.mod has one
+  const toolchain = regEx(/^toolchain\s*go(?<gover>\d+\.\d+\.\d+)$/m);
+  const toolchainMatch = toolchain.exec(content);
+  const toolchainVer = toolchainMatch?.groups?.gover;
+  if (toolchainVer) {
+    logger.debug(`Using go version ${toolchainVer) found in toolchain directive`);
```
trim21, line 460:
```suggestion
    logger.debug(`Using go version ${toolchainVer} found in toolchain directive`);
```

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -451,6 +451,16 @@ async function getGoConstraints(
   if (!content) {
     return undefined;
   }
+
+  // prefer toolchain directive when go.mod has one
+  const toolchain = regEx(/^toolchain\s*go(?<gover>\d+\.\d+\.\d+)$/m);
+  const toolchainMatch = toolchain.exec(content);
+  const toolchainVer = toolchainMatch?.groups?.gover;
+  if (toolchainVer) {
+    logger.debug(`Using go version ${toolchainVer} found in toolchain directive`);
```
trim21, line 460:
```suggestion
    logger.debug(
      `Using go version ${toolchainVer} found in toolchain directive`
    );
```

rarkins (commented):
Does this definitely work if we are *upgrading* the tool chain version?

trim21:
> Does this definitely work if we are _upgrading_ the tool chain version?

I don't know, that's what I'm asking since https://github.com/renovatebot/renovate/pull/34564#issuecomment-2692108203

If you also don't know it for sure, I'd suggestion we go back again using `getGoConstraints(massagedGoMod)`? , which definitely works.

casaqori:
Interested in this feature.
Would that change then allow toolchain pinning?
For motivation see reply on https://github.com/renovatebot/renovate/discussions/34563

trim21:
> Interested in this feature. Would that change then allow toolchain pinning? For motivation see reply on #34563

In my understanding after this change only updateType=toolchain will change toolchain directive, package only change will not change it anymore

trim21:
There is acually a corner case, when a go.mod has `go major.minor.path` without toolchain directive for example `go 1.23.5`, renovate still pick latest go version and add a toolchain directive. but go itself will pick 1.23.5, in my understanding.

rarkins (dismissed):

lib/modules/manager/gomod/artifacts.spec.ts
```diff
@@ -2121,6 +2121,96 @@ describe('modules/manager/gomod/artifacts', () => {
...
+    fs.readLocalFile
+      .mockResolvedValueOnce('New go.sum')
+      .mockResolvedValueOnce('New go.mod');
+    datasource.getPkgReleases.mockResolvedValueOnce({
+      releases: [
+        { version: '1.17.0' },
+        { version: '1.23.6' },
+        { version: '1.24.0' },
+      ],
+    });
```
viceice, line 2143: drop it and add a should not be called expect after update artifacts 
```suggestion
```

lib/modules/manager/gomod/artifacts.spec.ts
```diff
@@ -2121,6 +2121,96 @@ describe('modules/manager/gomod/artifacts', () => {
...
+      .mockResolvedValueOnce('New go.sum')
+      .mockResolvedValueOnce('New go.mod');
+    datasource.getPkgReleases.mockResolvedValueOnce({
+      releases: [
+        { version: '1.17.0' },
+        { version: '1.23.5' },
+        { version: '1.23.6' },
+        { version: '1.24.0' },
+      ],
+    });
```
viceice, line 2189: same here

lib/modules/manager/gomod/artifacts.spec.ts
trim21: which function and arguments should not be called？

lib/modules/manager/gomod/artifacts.spec.ts
```diff
@@ -2144,7 +2225,7 @@ describe('modules/manager/gomod/artifacts', () => {
       updatedDeps: [
         { depName: 'github.com/google/go-github/v24', newVersion: 'v28.0.0' },
       ],
-      newPackageFileContent: gomod1,
+      newPackageFileContent: 'someText\n\ngo 1.17\n\n' + gomod1,
```
viceice, line 2228: use string template

lib/modules/manager/gomod/artifacts.spec.ts
```diff
@@ -2121,6 +2121,87 @@ describe('modules/manager/gomod/artifacts', () => {
...
+    );
+    fs.readLocalFile
+      .mockResolvedValueOnce('New go.sum')
+      .mockResolvedValueOnce('New go.mod');
+
+    const res = await gomod.updateArtifacts({
+      packageFileName: 'go.mod',
+      updatedDeps: [{ depName: 'golang.org/x/crypto', newVersion: '0.35.0' }],
+      newPackageFileContent:
+        'someText\n\ngo 1.13\n\ntoolchain go1.23.6\n\n' + gomod1,
```
viceice, line 2142: use string template

lib/modules/manager/gomod/artifacts.spec.ts
```diff
@@ -2121,6 +2121,87 @@ describe('modules/manager/gomod/artifacts', () => {
...
+      }),
+    );
+    fs.readLocalFile
+      .mockResolvedValueOnce('New go.sum')
+      .mockResolvedValueOnce('New go.mod');
+
+    const res = await gomod.updateArtifacts({
+      packageFileName: 'go.mod',
+      updatedDeps: [{ depName: 'golang.org/x/crypto', newVersion: '0.35.0' }],
+      newPackageFileContent: 'someText\n\ngo 1.23.5\n\n' + gomod1,
```
viceice, line 2182: use string template

rarkins:
I think I'd be happy _not_ to add the toolchain directive in the last case

trim21:
> I think I'd be happy _not_ to add the toolchain directive in the last case

In that case we need to know which go version we will pick finally

trim21:
> > I think I'd be happy _not_ to add the toolchain directive in the last case
> 
> In that case we need to know which go version we will pick finally, it need `go get ./... toolchain@none` to do this.

I think this is out of scope of this PR, this PR just want to resolve first 2 case.

It should be done with the removing of -d argument, they need same mechanism

rarkins (approved):

viceice (approved):

trim21:
👀 need anything else before merge?

PR #34568, opened 2025-03-01
feat: add RENOVATE_ prefix to Github com token and env host rules

secustor:
## Changes

Adds support for prefixing `GITHUB_COM_TOKEN` and host rules with `RENOVATE_`. 
This will align the usages of these configuration with other options. 

### DIFF
--- a/docs/usage/getting-started/private-packages.md
+++ b/docs/usage/getting-started/private-packages.md
@@ -154,7 +154,11 @@ It can be confusing for people who host their own source code privately to be as
 Currently the preferred way to configure `github.com` credentials for self-hosted Renovate is:
 
 - Create a read-only Personal Access Token (PAT) for a `github.com` account. This can be any GitHub account, but we recommend you create an "empty" account for this purpose.
-- Add the PAT to Renovate using the environment variable `GITHUB_COM_TOKEN`
+- Add the PAT to Renovate using the environment variable `RENOVATE_GITHUB_COM_TOKEN`
+
+<!-- prettier-ignore -->
+!!! note
+    `GITHUB_COM_TOKEN` is still parsed and takes precedence over `RENOVATE_GITHUB_COM_TOKEN`, but is considered deprecated and will be removed in a future major update.
 
 ## Package Manager Credentials for Artifact Updating
 
@@ -605,7 +609,7 @@ If you need to provide credentials to the Mend Renovate App, please do this:
      "hostRules": [
        {
          "matchHost": "github.com",
-         "token": "{{ secrets.GITHUB_COM_TOKEN }}"
+         "token": "{{ secrets.RENOVATE_GITHUB_COM_TOKEN }}"
        }
      ]
    }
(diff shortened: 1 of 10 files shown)


docs/usage/getting-started/private-packages.md
```diff
@@ -154,7 +154,11 @@ It can be confusing for people who host their own source code privately to be as
 Currently the preferred way to configure `github.com` credentials for self-hosted Renovate is:
 
 - Create a read-only Personal Access Token (PAT) for a `github.com` account. This can be any GitHub account, but we recommend you create an "empty" account for this purpose.
-- Add the PAT to Renovate using the environment variable `GITHUB_COM_TOKEN`
+- Add the PAT to Renovate using the environment variable `RENOVATE_GITHUB_COM_TOKEN`
+
+<!-- prettier-ignore -->
+!!! note
+    `GITHUB_COM_TOKEN` is still parsed and takes precedence over `RENOVATE_GITHUB_COM_TOKEN`, but is considered deprecated and will be removed in a future major update.
```
viceice, line 161: shouldn't we do it the other way around? 🤔 we do it for log variables

docs/usage/mend-hosted/migrating-secrets.md
```diff
@@ -32,7 +32,7 @@ This is the new method that you should start using:
   "hostRules": [
     {
       "matchHost": "github.com",
-      "token": "{{ secrets.GITHUB_COM_TOKEN }}"
+      "token": "{{ secrets.RENOVATE_GITHUB_COM_TOKEN }}"
```
viceice, line 35: 🤔 secret names are irrelevant for this

lib/workers/global/config/parse/env.ts
```diff
@@ -202,12 +202,13 @@ export async function getConfig(
     }
   }
 
-  if (env.GITHUB_COM_TOKEN) {
+  const githubComToken = env.GITHUB_COM_TOKEN ?? env.RENOVATE_GITHUB_COM_TOKEN;
```
viceice, line 205: shouldn't we invert this? 🤔

docs/usage/getting-started/private-packages.md
secustor: My idea has been that if somebody has this variable set for whatever reason, maybe because the person has expected the prefix, then it would break. So this would be technically a breaking change. 

With this we give users the possibility to migrate and then in v40 we can remove support for it.

lib/workers/global/config/parse/env.ts
secustor: See above.

docs/usage/mend-hosted/migrating-secrets.md
secustor: True, removing it from here.

docs/usage/getting-started/private-packages.md
rarkins: Sure, let's change precedence in v40

viceice (approved):

PR #34569, opened 2025-03-01
feat(bitbucket-server): add support for changelogs

### LINKED ISSUE #3903: Support changelogs from bitbucket-server

Churro:
## Changes

Implements fetching of changelogs for Bitbucket Server / Data Center.

There are barely any changes to existing code but, if desired, the following changes could be extracted into separate PRs:
* `bitbucket-server-tags` datasource
* adding a `maxPages` option to `lib/util/http/bitbucket-server.ts`
* sourceURL extraction in `lib/workers/repository/update/pr/body/index.ts`

## Context

- Closes https://github.com/renovatebot/renovate/issues/3903

Test repo: https://github.com/renovate-demo/3903-bitbucket-server-changelogs/pull/1
-> Demonstrates changelog fetching from publicly available (but slightly outdated) Bitbucket Server instances


lib/workers/repository/update/pr/changelog/release-notes.spec.ts
```diff
@@ -100,6 +100,12 @@
   baseUrl: 'https://bitbucket.org/',
 });
 
+const bitbucketServerProject = partial<ChangeLogProject>({
+  type: 'bitbucket-server',
+  apiBaseUrl: 'https://bitbucket.domain.org/rest/api/1.0/',
+  baseUrl: 'https://bitbucket.domain.org/',
```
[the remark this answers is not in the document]
Churro: Any hints on how to deal with this finding would be appreciated 😊

lib/workers/repository/update/pr/body/index.ts
```diff
@@ -32,22 +32,20 @@ function massageUpdateMetadata(config: BranchConfig): void {
       depNameLinked = `[${depNameLinked}](${primaryLink})`;
     }
 
-    let sourceRootPath = 'tree';
+    let sourceRootPath = 'tree/HEAD';
     if (sourceUrl) {
       const sourcePlatform = detectPlatform(sourceUrl);
       if (sourcePlatform === 'bitbucket') {
-        sourceRootPath = 'src';
+        sourceRootPath = 'src/HEAD';
```
rarkins, line 39: Is this a fix for the existing bitbucket cloud functionality?

Churro: No, this doesn't change existing behavior. `src/HEAD` is already used with bitbucket but `HEAD` is currently hard-coded a few lines later:
https://github.com/renovatebot/renovate/blob/c4242426bcf43e0bfa5eaeb3fcceb4a0d5edda6a/lib/workers/repository/update/pr/body/index.ts#L48

All other platforms use `tree/HEAD`

Churro:
Please let me know if there's anything I could do on my end to help move this PR forward 😊

rarkins (dismissed):

lib/modules/datasource/bitbucket-server-tags/index.ts
```diff
@@ -0,0 +1,194 @@
...
+  BitbucketServerTag,
+  BitbucketServerTags,
+} from './schema';
+
+export class BitbucketServerTagsDatasource extends Datasource {
+  static readonly id = 'bitbucket-server-tags';
+
+  override http = new BitbucketServerHttp(BitbucketServerTagsDatasource.id);
+
+  static readonly defaultRegistryUrls = ['https://bitbucket.org'];
```
viceice, line 23:
```suggestion
  static readonly defaultRegistryUrls = [];
```
this would crash and never work. `bitbucket.org` has a totally different platform / api

lib/modules/datasource/bitbucket-server-tags/index.ts
```diff
@@ -0,0 +1,194 @@
...
+      )
+      .transform((versions): ReleaseResult => {
+        return {
+          sourceUrl: BitbucketServerTagsDatasource.getSourceUrl(
+            projectKey,
+            repositorySlug,
+            registryUrl,
+          ),
+          registryUrl:
+            BitbucketServerTagsDatasource.getRegistryURL(registryUrl),
```
viceice, line 103: why do we need to compute the registry url again? can't we reuse `registryUrl` ?

lib/util/http/bitbucket-server.ts
```diff
@@ -3,6 +3,7 @@ import { HttpBase, type InternalJsonUnsafeOptions } from './http';
 import type { HttpMethod, HttpOptions, HttpResponse } from './types';
 
 const MAX_LIMIT = 100;
+const MAX_PAGES = Number.MAX_SAFE_INTEGER;
```
viceice, line 6: shouldn't we set it to a more conservative value like here
https://github.com/renovatebot/renovate/blob/6dee68227cca36214c0fcf6b17dc2c82ee9d5bf8/lib/util/http/bitbucket.ts#L6

viceice (commented):
otherwise lgtm

lib/modules/datasource/bitbucket-server-tags/index.ts
Churro: I've removed this fallback now and added an explicit check for `registryUrl`. It wouldn't have crashed but arguably not have been very useful either, since requests would have always errored.

lib/modules/datasource/bitbucket-server-tags/index.ts
Churro: Depending on how the datasource is used (directly by users or during changelog retrieval) `registryUrl` could be `bitbucket.server.org/rest/api/1.0` or `bitbucket.server.org`. 

Calling `getRegistryURL()` here ensures that the value returned for releases is always without `/rest/api/1.0` suffix, which is how this value also looks like in other datasources.

lib/modules/datasource/bitbucket-server-tags/index.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    if (newValue?.length) {
+      return this.getTagCommit(baseUrl, newValue);
+    }
+
+    const result = Result.parse(config, DigestsConfig)
+      .transform(() => {
+        const url = `${baseUrl}/commits?ignoreMissing=true`;
+
+        return this.http.getJsonSafe(
```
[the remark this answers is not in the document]
Churro: They are the same. Only difference is that `getJson` throws an error if the response is not valid and `getJsonSafe` wraps the result in the `Result`class.

I tried to follow the "good example" mentioned [here](https://github.com/renovatebot/renovate/blob/main/docs/development/zod.md) as much as possible:
> [The cdnjs datasource](https://github.com/renovatebot/renovate/blob/main/lib/modules/datasource/cdnjs/index.ts) is a good example of using Zod schema validations on API responses from external sources.

secustor (approved):

viceice (approved):

PR #34584, opened 2025-03-03
build: compile with nodenext target (still cjs)

viceice:
## Changes
Some preparations to easier allow importing esm from cjs later.
Code will still be generated as cjs but some dynamic imports are no longer compiled to require, they use import instead.


lib/util/exec/containerbase.ts
```diff
@@ -230,7 +230,7 @@ async function getPkgReleases(
   toolConfig: ToolConfig,
 ): Promise<ReleaseResult | null> {
   if (_getPkgReleases === null) {
-    _getPkgReleases = import('../../modules/datasource');
+    _getPkgReleases = import('../../modules/datasource/index.js');
```
rarkins, line 233: JS reference in a TS non-test file seems hacky?

viceice: no, TS extension isn't allowed for typescript 🤗

rarkins: How does this behave in dev?

viceice: vitest loads the TS file transparently

tools/static-data/generate-distro-info.mjs
```diff
@@ -52,8 +51,13 @@ function csvToJson(raw) {
  * @param {string} file File path to update
  */
 async function update(url, file) {
-  const res = await got(url);
-  const csv = res.body;
+  // TODO: will be fixed with got upgrade
```
viceice, line 54:
```suggestion
```

package.json
```diff
@@ -338,7 +338,7 @@
     "semantic-release": "24.2.3",
     "tar": "7.4.3",
     "tmp-promise": "3.0.3",
-    "ts-node": "10.9.2",
+    "tsx": "4.19.3",
```
secustor, line 341: Now that we are having Node 22.13.0 as minimum version we could also use the experimental flag to remove types and drop this 3rd party dep. 🤔

lib/util/exec/containerbase.ts
secustor: Can we enable [allowImportingTsExtensions](https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions) to make this more transparent?

package.json
viceice: it's not yet in main and tsx solves some import inconsistencies

viceice:
reverted some changes, because main already has ts v5.8 now 😉

secustor (dismissed):
See open comment, else LGTM

lib/util/exec/containerbase.ts
trim21: allowImportingTsExtensions only works when you do not build js file with tsc

trim21: need something like this:

```diff
diff --git a/tsconfig.app.json b/tsconfig.app.json
index 58c915367..d51d9f5e2 100644
--- a/tsconfig.app.json
+++ b/tsconfig.app.json
@@ -5,9 +5,12 @@
     "sourceMap": true,
     "inlineSources": true,
     "importHelpers": true,
+    "allowImportingTsExtensions": false,
+    "noEmit": false,
     "types": ["node"],
     "paths": {} // no paths for production
   },
   "files": ["lib/renovate.ts", "lib/config-validator.ts"],
+  "exclude": ["*.spec.ts"],
   "include": ["lib/**/*.d.ts"]
 }
diff --git a/tsconfig.json b/tsconfig.json
index 04d5a09a0..726b34bad 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,6 +10,8 @@
     "module": "nodenext",
     "sourceMap": true,
     "allowSyntheticDefaultImports": true,
+    "allowImportingTsExtensions": true,
+    "noEmit": true,
     "esModuleInterop": true,
     "resolveJsonModule": true,
     "noUnusedLocals": true,

```

rarkins (approved):

PR #34595, opened 2025-03-03
fix(datasource/composer):  update default registry url.

Seldaek:
## Changes

Packagist.org & Composer maintainer here.. We are trying to phase out the packagist.org/p2/... URLs as we are serving the repo.packagist.org domain via a CDN which is much more efficient for everyone involved. These URLs were historically present but right now I see a lot of the requests for these fils on the main domain come from RenovateBot.

I have no idea if this URL needs to be changed in other places, so please take over here, but essentially packagist.org (without *repo* prefix) SHOULD NOT be used as a Composer repository anymore. Composer even internally rewrites that domain to repo.packagist.org, but third party clients obviously do not.

If you fail to take action here RenovateBot will stop working at some point when we disable these routes on the main domain. I don't have a firm timeline yet but I am trying to migrate whoever I can to avoid trouble.


viceice (dismissed):
it needs some test changes, we'll take care of it. thanks for notification

viceice (approved):

PR #34630, opened 2025-03-05
fix: resolve registry aliases for gitlabci manager

TimKnight-DWP:
Proposal for: https://github.com/renovatebot/renovate/discussions/34629


viceice (changes requested):
no, this behavior was explict reverted. registry aliases should not be used for dep names

viceice:
it was replaced some time ago, but we stopped to replace depname with registry aliases. that caused too many bad side effects.

- #33911

rarkins (changes requested):
We don't want to "resolve" them. It's intentional that depName reflects the original

PR #34701, opened 2025-03-08
fix: Pass the lock files of the processed packages to the manager

chludwig-haufe:
This PR addresses the issue discussed in #34015 that lock files might not be updated if the same dependency is updated by multiple managers, some of which use lock files and some do not.

## Changes

The change is limited to the function `patchConfigForArtifactsUpdate()` in `lib/workers/repository/update/branch/get-updated.ts`:

1. The result type is changed to `UpdateArtifactsConfig`: The result is exclusively used as the `config` attribute of an `UpdateArtifact`. This way it is clearer which attributes are expected by the managers that use the result.
2. Removed the attribute `lockFiles` from the local copy of the branch config argument. Whatever the value of this attribute (if any), I don't see any reliable connection to the packageFile about to be processed.
3. Copy the `lockFiles` attribute from the package file object into the result even if the branch config has no non-empty `lockFiles` attribute.

## Context

- #34015
- As discussed, I don't see why the absence of a non-empty `lockFiles` array in the branch config object should pre-empt the update of a manager-specific lock file.


viceice (changes requested):
this needs tests

chludwig-haufe:
> this needs tests

I tried my hand on a regression test. Since I usually do not develop in TypeScript or JavaScript, I am not familiar with the test framework. This might show...

chludwig-haufe:
Just in case it is not obvious: I am quite sure that the documentation build failure (a timeout while downloading `https://unpkg.com/mermaid@11/dist/mermaid.min.js`) has nothing to do with my change.

Besides making gratuitous changes to the PR, I am not aware of any way how I could re-trigger the pipeline, though.

rarkins:
@chludwig-haufe FYI you can usually select the "Update branch" button to merge from main, which triggers more tests. I did it just now

rarkins (approved):

viceice (approved):

PR #34713, opened 2025-03-09
docs(gomod): add info on go.mod directive

rarkins:


viceice (approved):

trim21:
There is a special case that when go directive is a full version, for example `go 1.23.5`, and if there is no toolchain, go will use go directive as toolchain directive. in this case the go directive means both "Compatible with this version or later" and "Use this exact version of go when it's main module"

PR #34720, opened 2025-03-09
fix: pnpm install flags on non-workspace project

trim21:
## Changes

remove `--recursive` if current package.json is not a workspace

## Context

- https://github.com/renovatebot/renovate/discussions/34718


rarkins (dismissed):

lib/modules/manager/npm/post-update/pnpm.ts
```diff
@@ -70,7 +75,14 @@ export async function generateLockFile(
...
     }
 
-    let args = '--recursive --lockfile-only';
+    let args = '--lockfile-only';
+    if (
+      await localPathExists(
+        getSiblingFileName(lockFileName, 'pnpm-workspace.yaml'),
+      )
+    ) {
+      args += ' --recursive';
```
viceice, line 84: I wonder if we need this flag at all. 🤔

at least pnpm v9+ is doing recursive install automatically when it detects workspaces.

trim21: that will be a breaking change for pnpm@8 I guess?

trim21: That's hard to implement in semver and too many pnpm source, I don't very want to do it 😂

viceice: what? isn't it simply if not intersects with `>=9`?

trim21: Ok I don't know there is a semver.intersects, I was looking on versioning/semver and I thought  we only have that on versioning/composer.

rarkins (approved):

lib/modules/manager/npm/post-update/pnpm.ts
```diff
@@ -70,7 +76,19 @@ export async function generateLockFile(
...
+      'pnpm-workspace.yaml',
+    );
+
+    let args = '--lockfile-only';
+    if (
+      pnpmToolConstraint.constraint &&
+      !semver.intersects(pnpmToolConstraint.constraint, '>=9') &&
+      (await localPathExists(pnpmWorkspaceFilePath))
+    ) {
+      args += ' --recursive';
```
rarkins, line 90: Please add a code comment describing why this is, for future maintainers/contributors

rarkins (changes requested):

viceice (commented):
otherwise LGTM

rarkins (approved):

PR #34747, opened 2025-03-11
feat(gitlab): implement pr cache

RahulGautamSingh:
## Changes
Cache pull requests using repository cache

## Context
- PartOf: https://github.com/renovatebot/renovate/issues/33290


lib/modules/platform/gitlab/pr-cache.ts
```diff
@@ -0,0 +1,175 @@
...
+import type { GitLabMergeRequest, GitlabPr, GitlabPrCacheData } from './types';
+import { prInfo } from './utils';
+
+export class GitlabPrCache {
+  private items: GitlabPr[] = [];
+  private cache: GitlabPrCacheData;
+
+  private constructor(
+    private repo: string,
+    private author: string | null,
```
zharinov, line 19:
```suggestion
    author: string | null,
```

zharinov (approved):

lib/modules/platform/gitlab/pr-cache.ts
RahulGautamSingh: Why?

lib/modules/platform/gitlab/pr-cache.ts
```diff
@@ -0,0 +1,175 @@
...
+      per_page: '100',
+    } as any;
+    if (!this.ignorePrAuthor) {
+      searchParams.scope = 'created_by_me';
+    }
+    const query = getQueryString(searchParams);
+    const url = `/projects/${this.repo}/merge_requests?${query}`;
+
+    const res = await http.getJsonUnchecked<GitLabMergeRequest[]>(url, {
+      paginate: true,
```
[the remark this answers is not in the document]
zharinov: _Yo Dawg i herd you like cache so we put a cache in yo cache so you can cache while u cache_

I think we just need to remove it here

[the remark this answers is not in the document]
RahulGautamSingh: Yes, that was unnecessary.

I also think `reconcile` doesn't make much sense here as other platforms since it's only called from within the `sync` method unlike on other platforms where it could be called from `initRepo`, should I remove it too?

zharinov (approved):

lib/modules/platform/gitlab/index.ts
```diff
@@ -136,6 +137,7 @@ export async function initPlatform({
       platformConfig.gitAuthor = `${user.name} <${
         user.commit_email ?? user.email
       }>`;
+      config.author = user.name;
```
viceice, line 140: we should probably allow using the manual configured username if we can't determine.
can be done in a followup pr, gitea doesn't fallback either.

lib/modules/platform/gitlab/index.ts
```diff
@@ -548,36 +550,14 @@ export async function getBranchStatus(
...
-      throw new Error(PLATFORM_AUTHENTICATION_ERROR);
-    }
-    throw err;
-  }
-}
-
 export async function getPrList(): Promise<Pr[]> {
   if (!config.prList) {
-    config.prList = await fetchPrList();
+    config.prList = await GitlabPrCache.getPrs(
```
viceice, line 555: wonder if we still need `config.prList` 🤔 we don't have it on gitea.

lib/modules/platform/gitlab/pr-cache.ts
```diff
@@ -0,0 +1,176 @@
...
+      searchParams.scope = 'created_by_me';
+    }
+    let query: string | null = getQueryString(searchParams);
+
+    while (query) {
+      const res = await http.getJsonUnchecked<GitLabMergeRequest[]>(
+        `/projects/${this.repo}/merge_requests?${query}`,
+        {
+          memCache: false,
+          paginate: false,
```
viceice, line 151:
```suggestion
```

Isn't it the default value?

lib/modules/platform/gitlab/index.ts
RahulGautamSingh: We don't

viceice: let's remove it to keep things simple

RahulGautamSingh:
Tests failed due GH actions issue. Will need to re-run the jobs once the main is ahead .

lib/modules/platform/gitlab/index.ts
```diff
@@ -321,7 +325,7 @@ export async function initRepo({
   endpoint,
   includeMirrors,
 }: RepoParams): Promise<RepoResult> {
-  config = {} as any;
+  config ??= {} as any;
```
RahulGautamSingh, line 328: @viceice Should I separate this out in a dffierent PR?

Because of this the config gets reset which causes the `author` set during `initPlatform` to be lost after `initRepo`

viceice: no, use a separate variable for username, like we do on other plarforms.

lib/modules/platform/gitlab/pr-cache.ts
```diff
@@ -0,0 +1,174 @@
...
+        },
+      );
+
+      const needNextPage = this.reconcile(res.body);
+      if (!needNextPage) {
+        break;
+      }
+
+      const uri = parseUrl(parseLinkHeader(res.headers.link)?.next?.url);
+      query = uri ? uri.search : null;
```
viceice, line 159: We need to take `GITLAB_IGNORE_REPO_URL` env into account
https://github.com/renovatebot/renovate/blob/fffd986fa784bc0297062dba2fe7a53616775fc5/lib/util/http/gitlab.ts#L51-L55

lib/modules/platform/gitlab/index.ts
```diff
@@ -165,6 +167,8 @@ export async function initPlatform({
     ? DRAFT_PREFIX_DEPRECATED
     : DRAFT_PREFIX;
 
+  config.author ??= username!;
```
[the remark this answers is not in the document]
RahulGautamSingh: Will create a separate PR for this 🫡

lib/modules/platform/gitlab/pr-cache.ts
```diff
@@ -0,0 +1,174 @@
...
+
+      const cacheOldTime = updated_at ? DateTime.fromISO(updated_at) : null;
+      if (!cacheOldTime || itemNewTime > cacheOldTime) {
+        updated_at = rawItem.updated_at;
+      }
+    }
+
+    this.cache.updated_at = updated_at;
+    return needNextPage;
+  }
```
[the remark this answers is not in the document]
RahulGautamSingh: Will add it to the //todolist

lib/modules/platform/gitlab/pr-cache.ts
RahulGautamSingh: It's deprecated

lib/modules/platform/gitlab/index.ts
```diff
@@ -78,14 +78,14 @@ import { DRAFT_PREFIX, DRAFT_PREFIX_DEPRECATED, prInfo } from './utils';
...
   email: string;
-  prList: GitlabPr[] | undefined;
   issueList: GitlabIssue[] | undefined;
   mergeMethod: MergeMethod;
   defaultBranch: string;
   cloneSubmodules: boolean | undefined;
   cloneSubmodulesFilter: string[] | undefined;
   ignorePrAuthor: boolean | undefined;
   squash: boolean;
+  author: string | null;
```
viceice, line 88: drop and use `botUserName` consistently

lib/modules/platform/gitlab/index.ts
```diff
@@ -138,6 +140,8 @@ export async function initPlatform({
       platformConfig.gitAuthor = `${user.name} <${
         user.commit_email ?? user.email
       }>`;
+      config.author = user.name;
```
viceice, line 143:
```suggestion
```

lib/modules/platform/gitlab/index.ts
```diff
@@ -165,6 +169,9 @@ export async function initPlatform({
     ? DRAFT_PREFIX_DEPRECATED
     : DRAFT_PREFIX;
 
+  config.author ??= username!;
```
viceice, line 172:
```suggestion
```

lib/modules/platform/gitlab/index.ts
```diff
@@ -841,22 +825,13 @@ export async function updatePr({
...
-      );
-      config.prList.push(updatedPr);
-    } /* v8 ignore stop */ else {
-      config.prList[existingIndex] = updatedPr;
-    }
-  }
+  await GitlabPrCache.setPr(
+    gitlabApi,
+    config.repository,
+    config.author,
```
viceice, line 831:
```suggestion
    botUserName,
```

lib/modules/platform/gitlab/pr-cache.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+  resetCache as repoCacheReset,
+} from '../../../util/cache/repository';
+import type { LongCommitSha } from '../../../util/git/types';
+import { GitlabHttp, setBaseUrl } from '../../../util/http/gitlab';
+import { GitlabPrCache } from './pr-cache';
+import type { GitLabMergeRequest } from './types';
+import { prInfo } from './utils';
+
+const http = new GitlabHttp();
+const ignorePrAuthor = false;
```
viceice, line 14: does it really makes sense to store it globally? 🤔

lib/modules/platform/gitlab/pr-cache.ts
```diff
@@ -0,0 +1,181 @@
...
+      }
+
+      const uri = parseUrl(parseLinkHeader(res.headers.link)?.next?.url);
+      if (uri && http.baseUrl) {
+        if (process.env.GITLAB_IGNORE_REPO_URL) {
+          const defaultEndpoint = new URL(http.baseUrl);
+          uri.protocol = defaultEndpoint.protocol;
+          uri.host = defaultEndpoint.host;
+        }
+      }
```
viceice, line 165: wait, you're only using `search`, so this doesn't change anything on `query`
```suggestion
```

lib/util/http/gitlab.ts
```diff
@@ -16,7 +16,7 @@ export interface GitlabHttpOptions extends HttpOptions {
 }
 
 export class GitlabHttp extends HttpBase<GitlabHttpOptions> {
-  protected override get baseUrl(): string | undefined {
+  override get baseUrl(): string | undefined {
```
viceice, line 19: revert, not needed

lib/modules/platform/gitlab/pr-cache.ts
RahulGautamSingh: 😅

lib/modules/platform/gitlab/pr-cache.ts
```diff
@@ -0,0 +1,174 @@
...
+
+      const oldItem = oldItems[id];
+      const newItem = prInfo(rawItem);
+
+      const itemNewTime = DateTime.fromISO(rawItem.updated_at);
+
+      if (dequal(oldItem, newItem)) {
+        needNextPage = false;
+        continue;
+      }
```
viceice, line 121: do we really need to do this? is this for the case we need new pr properties?

lib/modules/platform/gitlab/pr-cache.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    this.cache.updated_at = updated_at;
+    return needNextPage;
+  }
+
+  private async sync(http: GitlabHttp): Promise<GitlabPrCache> {
+    logger.debug('Syncing PR list');
+    const searchParams = {
+      per_page: this.items.length ? '20' : '100',
+    } as any;
```
viceice, line 139:
```suggestion
    } as Record<string,string>;
```

lib/modules/platform/gitlab/pr-cache.ts
viceice: then we can simply compare the `updated_at` with the old item, that would be much faster

viceice (commented):
see comment

lib/modules/platform/gitlab/pr-cache.ts
RahulGautamSingh: It is the same in other implementations, maybe better to refactor this together in followup

viceice: ok, it would be a huge performance improvement because comaring a lot of those big objects is more expensive than comparing a date 🙃

viceice (approved):

PR #34779, opened 2025-03-13
feat(gomod): preserve go/toolchain directives

### LINKED ISSUE #34097: Skip -d flag for "go get" if go 1.18 or later is run

As originally proposed in https://github.com/renovatebot/renovate/pull/32506 but not completed, the `-d` flag is not necessary for go 1.18 or later.

It's important that Renovate to keep the `-d` flag if an older version of `go` is run, although that is rare.

The go toolchain directive was introduced in Go 1.21 so it's not possible for it to be a scenario where <1.18 is used.

The only way which Go could be pinned to a lower version is if the user configures `constraints.go` in Renovate config, so this edge case should be considered.

trim21:
## Changes

If go.mod has go>=1.21, add `toolchain@{toolchain directive} go@${go directive}` to preserve the directives. use `toolchain@none` as a special case to not add toolchain directive.

Only add `-d` flag if go.mod has old go version < 1.17.

close https://github.com/renovatebot/renovate/issues/34097


lib/modules/manager/gomod/artifacts.ts
```diff
@@ -190,8 +190,20 @@ export async function updateArtifacts({
...
-    config.constraints?.go ?? getGoConstraints(newGoModContent);
+    config.constraints?.go ?? goMod.toolchain ?? goMod.constraints;
+
+  const flags: string[] = [];
+
+  if (semver.satisfies(goMod.go, '^1.21.0')) {
+    flags.push(`toolchain@${goMod.toolchain ?? 'none'}`);
+  }
+
+  if (!semver.satisfies(goMod.go, '^1.17.0')) {
```
viceice, line 204:
```suggestion
  if (!semver.satisfies(goMod.go, '>=1.17.0')) {
```

should also allow go v2

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -190,8 +190,20 @@ export async function updateArtifacts({
...
   }
+
+  const goMod = getGoConfig(newGoModContent);
   const goConstraints =
-    config.constraints?.go ?? getGoConstraints(newGoModContent);
+    config.constraints?.go ?? goMod.toolchain ?? goMod.constraints;
+
+  const flags: string[] = [];
+
+  if (semver.satisfies(goMod.go, '^1.21.0')) {
```
viceice, line 200:
```suggestion
  if (semver.satisfies(goMod.go, '>=1.21.0')) {
```

trim21:
github UI is somehow broken, I can't apply them as batch, hope you don't mind:

image

trim21:
@viceice I changed the implementation due to I didn't fully understand this problem, please check the PR description for currently  implementation .

trim21:
@rarkins can you take a look?

lib/modules/manager/gomod/artifacts.spec.ts
```diff
@@ -123,7 +123,7 @@ describe('modules/manager/gomod/artifacts', () => {
     ).toBeNull();
     expect(execSnapshots).toMatchObject([
       {
-        cmd: 'go get -d -t ./...',
+        cmd: 'go get -t -d ./...',
```
viceice, line 126: can you try to avoid this reorder of flags?

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -443,37 +456,30 @@ export async function updateArtifacts({
...
-  const match = re.exec(content);
-  if (!match?.groups?.gover) {
-    return undefined;
+  // go mod spec says if go directive is missing it's 1.16
+  const goDirective = goMatch?.groups?.gover ?? '1.16';
+  let goVersion = goDirective;
+
+  // partial semver version without patch
+  if (/^\d+\.\d+$/.test(goVersion)) {
+    goVersion = `${goVersion}.0`;
```
viceice, line 477: why do we need to add the zero? 🤔

viceice (commented):
this needs extensive testing on existing repos

viceice:
I think we should revert https://github.com/renovatebot/renovate/pull/34564 and start new here.


@rarkins wdyt?

lib/modules/manager/gomod/artifacts.ts
trim21: `semver.satisfies('1.17', '>=1.17.0')` is false

trim21: the regex here only match `*.*`

rarkins:
> I think we should revert #34564 and start new here.
> 
> @rarkins wdyt?

I thought that #34564 was an improvement over what we had before though. i.e. using the toolchain version instead of the latest?

viceice:
> > I think we should revert #34564 and start new here.
> > 
> > @rarkins wdyt?
> 
> I thought that #34564 was an improvement over what we had before though. i.e. using the toolchain version instead of the latest?

sure, but the hosted app was reverted because of issues with that fix. see slack discussion

rarkins:
Definitely revert then

trim21:
> Definitely revert then

With or without #34564, the PR (issues mentioned is slack) is always broken 🤔

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -443,37 +456,30 @@ export async function updateArtifacts({
   }
 }
 
-function getGoConstraints(content: string): string | undefined {
-  // prefer toolchain directive when go.mod has one
+function getGoConfig(content: string): {
+  toolchain?: string;
+  minialGoVersion: string;
```
casaqori, line 461: typo

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -443,37 +456,30 @@ export async function updateArtifacts({
...
+  go: string;
```
casaqori, line 462: naming: goVersion to be consistent with minimalGoVersion ?

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -443,37 +456,30 @@ export async function updateArtifacts({
   }
 }
 
-function getGoConstraints(content: string): string | undefined {
-  // prefer toolchain directive when go.mod has one
+function getGoConfig(content: string): {
+  toolchain?: string;
```
casaqori, line 460: naming: toolchainVersion to be consistent with minimalGoVersion ?

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -190,8 +190,20 @@ export async function updateArtifacts({
       );
     }
   }
-  const goConstraints =
-    config.constraints?.go ?? getGoConstraints(newGoModContent);
+
+  const goMod = getGoConfig(newGoModContent);
```
casaqori, line 194: naming: goDirectives - because its just the go version numbers parsed from the whole .mod file

lib/modules/manager/gomod/artifacts.ts
trim21: It's goDirective actually. Looks like I should name it better

lib/modules/manager/gomod/artifacts.ts
trim21: make sense

trim21:
good suggestions to me, but don't have time to fix it revently. 😅

 Don't know would you like to work on this PR? @casaqori

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -190,8 +190,20 @@ export async function updateArtifacts({
...
+  const goConstraints = config.constraints?.go ?? `^${goMod.minialGoVersion}`;
+
+  const getFlags = ['-t'];
+  if (!semver.satisfies(goMod.minialGoVersion, '>=1.17.0')) {
+    getFlags.unshift('-d');
+  }
+
+  const extraGetArguments: string[] = [];
+  if (semver.satisfies(goMod.minialGoVersion, '>=1.21.0')) {
```
casaqori, line 203: A question here - if toolchain and go directive are always passed to the cmd, can we still update to a newer version, when a dependency requires to ???
That would be the vanilla go default behavior.

trim21: no, you can will get this: https://github.com/trim21-bot/oapi-codegen/pull/6/files

trim21: oh sorry, didn't understand your qurstion. If renovate want to upgrade to new toolchain version we will pass new toolchain version i command, so it will upgrade to a new version naturally

trim21: to summarize it:

1. go directive and toolchain directive are always preserved, unless renovate want to upgrade it.
2. it won't failed at `updateArtifacts`
3. if renovate want to upgrade a package that requires new go version, **it won't work as expected**. golang will downgrade that package to latest version that work with gomod's go directive

trim21: > A question here - if toolchain and go directive are always passed to the cmd, can we still update to a newer version of them, when a dependency requires to (declaring newer go directive in its mod file) ??? That would be the vanilla go default behavior. Test Case:
> 
> ```
> installed go tools: 1.23.7
> mod: go1.23.0 ; toolchain: none
> update to: golang.org/x/crypto@v0.34.0 (req. go1.23.0)
> 
> => upgrades dependency and does not add toolchain
> ---
> installed go tools: 1.23.7
> mod: go1.23.0 ; toolchain: none
> update to: golang.org/x/crypto@v0.40.0 (req. go1.24.0)
> 
> => does NOT upgrade dependency (go.mod unchanged)
> ---
> installed go tools: 1.24.1
> mod: go1.23.0 ; toolchain: none
> update to: golang.org/x/crypto@v0.40.0 (req. go1.24.0)
> 
> => upgrades dependency and bumps to go1.24.0 and does not add toolchain
> ---
> installed go tools: 1.23.7
> mod: go1.23.0 ; toolchain: go1.24.1
> update to: golang.org/x/crypto@v0.40.0 (req. go1.24.0)
> 
> => installs newer go tools, upgrades dependency and bumps to go1.24.0 and does not remove toolchain
> ```

We go back to pick latest go version as toolchain, so  go itself  should not install any go toolchain with this PR

casaqori: Got it, so basically with this PR renovate suggests dependency updates within the go version defined in go.mod of the project (e.g. 1.23).
The installed toolchain coming with renovate image would define, whether upgrades to newer versions of go (e.g. 1.24) are possible.
As far as I understood the Docker image of renovate will include always the latest go version. Which would mean that these deployments cannot benefit from this PR?
Or is it that the constraints of renovate shall actually control the "upgrade minor" vs "stay on minor" behavior @viceice ?
```json
  "constraints": {
    "go": "1.21"
  },
```

casaqori:
> good suggestions to me, but don't have time to fix it recently. 😅
> 
> Would you like to work on this PR? @casaqori

I see this going into the right direction. Can you add me to your Fork https://github.com/trim21/renovate/tree/fix/go-mod-toolchain so I can implement the renaming and add some additional test-cases? Thanks.

trim21:
> > good suggestions to me, but don't have time to fix it recently. 😅
> > Would you like to work on this PR? @casaqori
> 
> I see this going into the right direction. Can you add me to your Fork https://github.com/trim21/renovate/tree/fix/go-mod-toolchain so I can implement the renaming and add some additional test-cases? Thanks.

done. please don't touch other branches

lib/modules/manager/gomod/artifacts.ts
casaqori: @trim21 When I try the assembled command locally, with enforcement of versions, all I get is go exiting the upgrade, instead of falling back to a compatible version. What could I be missing?
On a go1.22 module having go 1.22.0 installed:
```shell
go get -t sigs.k8s.io/controller-runtime toolchain@1.22.0 go@1.22.0
go: sigs.k8s.io/controller-runtime@v0.20.4 requires go >= 1.23.0; switching to go1.23.7
go: sigs.k8s.io/controller-runtime@upgrade (v0.20.4) requires go@1.23.0, not go@1.22.0

go get -t sigs.k8s.io/controller-runtime toolchain@1.22.0
go: sigs.k8s.io/controller-runtime@v0.20.4 requires go >= 1.23.0; switching to go1.23.7
go: sigs.k8s.io/controller-runtime@upgrade (v0.20.4) indirectly requires toolchain@go1.23.0, not toolchain@1.22.0 (go1.22.0)
```

casaqori: thank you, with that a new observation:
installed go@1.22.0; disable automatic toolchain downloads => aborts
```
GOTOOLCHAIN=local go get -t ./... toolchain@1.22.0 go@1.22.0
go: golang.org/x/net@v0.38.0: module golang.org/x/net@v0.38.0 requires go >= 1.23.0 (running go 1.22.0; GOTOOLCHAIN=local)
```

installed go@1.22.0; allow automatic toolchain downloads => succeeds
```
GOTOOLCHAIN=auto go get -t ./... toolchain@1.22.0 go@1.22.0
go: module golang.org/x/net@v0.38.0 requires go >= 1.23.0; switching to go1.23.7
go: downgraded golang.org/x/crypto v0.36.0 => v0.33.0
go: downgraded golang.org/x/net v0.38.0 => v0.35.0
go: downgraded golang.org/x/sync v0.12.0 => v0.11.0
go: downgraded golang.org/x/sys v0.31.0 => v0.30.0
go: downgraded golang.org/x/term v0.30.0 => v0.29.0
go: downgraded golang.org/x/text v0.23.0 => v0.22.0
```

So, to downgrade, we still need the more recent version of go required by the dependency as installed local toolchain.
Since GOTOOLCHAIN=auto is the default, this is ensured for renovate.
Yet a constraint in renovate config would NOT limit the toolchain in use. It's just the "starter" toolchain!
Instead, forthgoing it would be an explicit limit for max go directive that can end up in go.mod after downgrading.
An example, with renovate config...
```json
  "constraints": {
    "go": "1.22"
  },
```
... and current go.mod ...
```text
go 1.21.0

require (
        golang.org/x/net v0.33.0
}
```
... the command will be `go get ./... toolchain@1.22.0 go@1.22.0` and the go.mod will become:
```text
go 1.22.0

require (
        golang.org/x/net v0.35.0
}
```
Which is a downgrade from the more recently golang.org/x/net v0.38.0 which would require go1.23.0

Is that okay with everyone @trim21 @viceice @rarkins  ?

trim21: > Yet a constraint in renovate config would NOT limit the toolchain in use. It's just the "starter" toolchain!


yes that's how go>=1.21 works. I have no opinion on how renovate use go if they doesn't touch the directives unexpectedly.

But renovatebot can consider to set `GOTOOLCHAIN=local` to disable the go toolchain version switch if they prefer.

casaqori: > renovatebot doesn't ensure `GOTOOLCHAIN=auto` (in my understanding), it's just the default value.
> 
> they can set `GOTOOLCHAIN=local` to disable the go toolchain version switch if they prefer, like some package manager.

That is right, it CAN be overridden in self-hosting scenario. But then it is reflecting exactly the expected behavior. If the required toolchain is not pre-installed locally then renovate will fail updating respective newer dependencies and no downgrade is possible.

The important part in my question is however to change the significance of renovate config's constraint for go.

casaqori:
@trim21 the new logic is in. The readme explains it. Can you test that on your repo forks please and come back with results? Thank you very much.

trim21:
> @trim21 the new logic is in. The readme explains it. Can you test that on your repo forks please and come back with results? Thank you very much.

you can fork it and test it by yourself...

casaqori:
uff, ... what would happen when you retrigger all here: https://github.com/trim21-bot/oapi-codegen/issues/3
does that make use of the renovate feature branch?

trim21:
> uff, ... what would happen when you retrigger all here: [trim21-bot/oapi-codegen#3](https://github.com/trim21-bot/oapi-codegen/issues/3) does that make use of the renovate feature branch?

This issue is generated from a renovatebot cli I executed manually. You will need to fork https://github.com/trim21-bot/oapi-codegen and run renovatebot locally to test it.

viceice:
> after you commit your change, run `pnpm run compile:ts` and `node dist/renovate.js ${github_owner}/${github_repo}` with renovate self-hosted config.

no, run `pnpm start` and pass required args. no need for compile step.

lib/modules/manager/gomod/artifacts.ts
viceice: > oh sorry, didn't understand your question at first.
> 
> If renovate want to upgrade to new go version we will use the new go version in command, so it will upgrade go.mod's directives to a new version naturally.
> 
> Then if the new package version work with given go version, it also upgrade package to expected version
> 
> Otherwise, no, you will get this: https://github.com/trim21-bot/oapi-codegen/pull/6/files , go will downgrade the package requires new go version

This isn't good. We need to detect if the upgrade failed! Otherwise it could be silentently gone.

Eg:
1. Renovate detects security update
2. go reverts because of constraints
3. automerge
4. bang no upgrade at all!

this must be an artifact error if the package was downgraded. all other would be very unexpected to a user.

trim21: if we are going to raise an artifact error, I think actually we do not need this PR, we just need to add `GOTOOLCHAIN=local` and the `go get` command will failed naturally 😁

trim21:
another way is to add updated package to commands:

```text
$ go get go@1.21.0 toolchain@none golang.org/x/tools@v0.31.0 ./...
go: golang.org/x/tools@v0.31.0 requires go@1.23.0, not go@1.21.0
```


we will get a clear error message from go itself

casaqori:
thanks, understood how to proceed

casaqori:
Test Results with a constraint on `go1.21` in https://github.com/casaqori/oapi-codegen/blob/main/renovate.json

Please see https://github.com/casaqori/oapi-codegen/pull/9
The PR title says "fix(deps): update module golang.org/x/tools to v0.31.0"
But since the actual `go get` used `go@1.21.0` the finally chosen version is "golang.org/x/tools v0.24.0"
Same with any other dependency that got downgraded, like https://github.com/casaqori/oapi-codegen/pull/8

Please see https://github.com/casaqori/oapi-codegen/pull/7
The update to "github.com/speakeasy-api/openapi-overlay v0.10.1" did not downgrade, the dep requires "go 1.22"
Please see https://github.com/casaqori/oapi-codegen/pull/6
Not downgraded, although "go 1.22.5" required: https://github.com/getkin/kin-openapi/blob/v0.131.0/go.mod
I will investigate.

Please see https://github.com/casaqori/oapi-codegen/pull/12
There are 2 times the same version in the .mod after update.
That seems unrelated and is presumably a (known) renovate bug? @viceice

trim21:
> Test Results with a constraint on `go1.21` in https://github.com/casaqori/oapi-codegen/blob/main/renovate.json
> 
> Please see [casaqori/oapi-codegen#9](https://github.com/casaqori/oapi-codegen/pull/9) The PR title says "fix(deps): update module golang.org/x/tools to v0.31.0" But since the actual `go get` used `go@1.21.0` the finally chosen version is "golang.org/x/tools v0.24.0" Same with any other dependency that got downgraded, like [casaqori/oapi-codegen#8](https://github.com/casaqori/oapi-codegen/pull/8)
> 
> Please see [casaqori/oapi-codegen#7](https://github.com/casaqori/oapi-codegen/pull/7) The update to "github.com/speakeasy-api/openapi-overlay v0.10.1" did not downgrade, the dep requires "go 1.22" Please see [casaqori/oapi-codegen#6](https://github.com/casaqori/oapi-codegen/pull/6) Not downgraded, although "go 1.22.5" required: https://github.com/getkin/kin-openapi/blob/v0.131.0/go.mod I will investigate.
> 
> Please see [casaqori/oapi-codegen#12](https://github.com/casaqori/oapi-codegen/pull/12) There are 2 times the same version in the .mod after update. That seems unrelated and is presumably a (known) renovate bug? @viceice

at here: https://github.com/renovatebot/renovate/pull/34779#discussion_r2024527437

lib/modules/manager/gomod/artifacts.ts
casaqori: Downgrade can result in 2 situations:
1. a PR that wants latest version, the latest compatible version is already in use => no changed files
2. a PR that wants latest version, another newer compatible version is available => changed go.mod uses other version than present in PR title

How to handle each?
1. add comment and auto-close?
2. add notice, update PR title?

Note that the downgrades would not be necessary at all when we would have the go.mod of the dependency available for inspection and discard those that are incompatible with the configured constraint prior to invoking updateArtifacts.

casaqori: see this PR for a Notice
https://github.com/casaqori/oapi-codegen/pull/14

However, I do not understand why renovate included the wrong go.mod into the PR.
It is the original one, after detecting available updates but before executing the updateArtifacts function.
The modified one would not contain any changes, which I did verify using log statements.

Some hints where to look at @viceice ?

casaqori:
@trim21 getting it right to do the dependency downgrade compatible to a certain constraint requires major restructuring of the gomanager. We'd need something like a dry run to detect what will finally change. 
Therefore, I'd propose to limit this PR solely on not writing the toolchain directive, if it was not present before.
There is no way of retaining the value it currently has, without defining a constraint, as go will always want to bump deps to its latest and requires at least a toolchain compatible with that latest.
Imho, it's not worth to try to revert the new toolchain value after upgrade to a value that it was before, because that counterworks the initial purpose that the go-team did foresee.
The solution to not have toolchain bumps will simply be to never introduce the directive manually.

Another PR can take that constraint logic to limit the go directive bumps, but if its only me, it will have to wait.
Are we okay?

trim21:
> @trim21 getting it right to do the dependency downgrade compatible to a certain constraint requires major restructuring of the gomanager. We'd need something like a dry run to detect what will finally change. Therefore, I'd propose to limit this PR solely on not writing the toolchain directive, if it was not present before. There is no way of retaining the value it currently has, without defining a constraint, as go will always want to bump deps to its latest and requires at least a toolchain compatible with that latest. Imho, it's not worth to try to revert the new toolchain value after upgrade to a value that it was before, because that counterworks the initial purpose that the go-team did foresee. The solution to not have toolchain bumps will simply be to never introduce the directive manually.
> 
> Another PR can take that constraint logic to limit the go directive bumps, but if its only me, it will have to wait. Are we okay?

no, there is no need to do a major refactor or define a constraint.

Just check go doesn't downgrade any updated package after we run `go get`. And if it doesn't, raise a artifacts updating error.

casaqori:
But the user will end up with a HUGE bunch of PRs that will never be mergeable...
SO why create the PRs in the first place? Skipping them is what I'd strive for.

trim21:
> But the user will end up with a HUGE bunch of PRs that will never be mergeable... And they always come up again when a new version is available ... SO why create the PRs in the first place? Skipping them is what I'd strive for.

Users can add package rule to disable them, and what if it's security PR?

casaqori:
How would artifactUpdate error behave? Would it NOT create the PR? But the err would appear in the dashboard?

trim21:
> How would artifactUpdate error behave? Would it NOT create the PR? But the err would appear in the dashboard?

https://github.com/oapi-codegen/oapi-codegen/pull/1641#issuecomment-2730263577

trim21:
I suggest adding updated package to `go get` command, go command provide a clear enough error message

trim21:
I get my other PRs merged, I can go back work on this if you don't want to go this direction.

casaqori:
```
$ go get go@1.21.0 toolchain@none golang.org/x/tools@v0.31.0 ./...
go: golang.org/x/tools@v0.31.0 requires go@1.23.0, not go@1.21.0
```
Also, without explicitly adding package name to command we see that error message.
Just currently I do not see an example of capturing the std output. And further, it's multiple commands stacked up that are being called.
So, maybe you can commit your ideas and we will see further.
What still causes headaches is that the PR is left in a state where it has 'golang.org/x/tools@v0.31.0' in the go.mod
When I "cleaned" go.mod to the state it was before the PR (since no upgrade was done) then I get warnings in console about having detected an empty commit which will not be pushed.

trim21:
image

trim21:
we need to add a little more code to handle the major upgrade case, because we need to also update package name

lib/modules/manager/gomod/artifacts.ts
```diff
@@ -190,8 +190,38 @@ export async function updateArtifacts({
       );
     }
   }
-  const goConstraints =
-    config.constraints?.go ?? getGoConstraints(newGoModContent);
+
+  const goMod = getGoConfig(newGoModContent);
+  const goConstraints = config.constraints?.go ?? `^${goMod.minimalGoVersion}`;
```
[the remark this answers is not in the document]
trim21: that's why https://github.com/renovatebot/renovate/pull/34779/files#diff-fd367c04a5c8c9b0dcbe22bb60da84f2b674ce21dadc840799ab1c2dd9b1a251R217-R221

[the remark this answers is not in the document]
casaqori: https://github.com/renovatebot/renovate/pull/34779/files#diff-4c37b8edc62aa49295d264c221717bd0361ccff0a791d4327f4421a0ea23521fR2214

Consider x/crypto would specify go1.23.7 in it's go.mod

go@1.23.5 would not allow that ... 
And users can not control that behavior in renovate. They would explicitly need to bump go before that. Why not allow the bump in same PR?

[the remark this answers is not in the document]
casaqori: The question is what a sane default is:
1) allow automated go version bumping that is supported by local toolchain when dependency requires that (have PRs that can be merged, toolchain always latest, current state of affairs)
2) avoid that (have PRs that break), enforce users to maintain their go directive separately
3) allow users to choose from 1) and 2) - e.g. if config.constraints.go is set, use strategy 1) for allowing updates within the constrained minor only

Who would be able to take such decision @viceice ?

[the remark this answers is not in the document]
casaqori: > ..., but they can't suppress updating go directive caused by library if they want
That is only the case for auto-merges of PRs. Otherwise users can transparently review and close unwanted PRs.

Please also consider that the changes currently WILL ALLOW go directive bumping. Here is the observation:
```
go.mod is at 1.22.0

go get -t ./... toolchain@none go@1.22.0 github.com/someorg/somepackage@v1.0.1-20250120212554
go: github.com/someorg/somepackage@v1.0.1-20250120212554 requires go >= 1.22.2; switching to go1.23.8
go: github.com/someorg/somepackage@v1.0.1-20250120212554 requires go@1.22.2, not go@1.22.0
```

```
go.mod is at 1.22

go get -t ./... toolchain@none go@1.22 github.com/someorg/somepackage@v1.0.1-20250120212554
go: github.com/someorg/somepackage@v1.0.1-20250120212554 requires go >= 1.22.2; switching to go1.23.8
go: upgraded go 1.22.0 => 1.22.12

go.mod is now at 1.22.12
```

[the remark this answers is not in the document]
trim21: Thanks, I don't know this case.

A workaround would be using `go 1.23` instead of `go 1.23.0` in go mod, but that requires user to change their code ... 

A better solution would be disabling go version switching and pick `1.23.0` when it has `go 1.23` directive 😅

trim21:
to get what we want, we need:

1. add env `GOTOOLCHAIN=local` to env
2. add `go@${go directive}` , `toolchain@${toolchain directive}` and all updated package to `go get` commands
3. pick the minimal supported go version defined in go.mod.

lib/modules/manager/gomod/artifacts.ts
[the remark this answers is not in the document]
trim21: > The question is what a sane default is:
> 
> 1. allow automated go version bumping that is supported by local toolchain when dependency requires that (have PRs that can be merged, toolchain always latest, current state of affairs)
> 2. avoid that (have PRs that break), enforce users to maintain their go directive separately
> 3. allow users to choose from 1) and 2) - e.g. if config.constraints.go is set, use strategy 1) for allowing updates within the constrained minor only
> 
> Who would be able to take such decision @viceice ?

IMO, if users want renovate to touch go version, they can manually enable it.

So renovate will update the go directive to expected version based on user config.

If they group go directive with packages, they get one working PR.
And if they have seprated PRs for go direcive and package, they can still get a working PR for package, they just need to merge the PR for go directive first!

rarkins (changes requested):
This goes far beyond the #34097 that it closes. It's also a pretty fundamental change to how Renovate handles go compatibility, and needs its own discussion, which I don't want to do here in a PR review comment

trim21:
it close #34097 by the way， but it's not the main target of this pr

trim21:
discussion： https://github.com/renovatebot/renovate/discussions/35344

lib/modules/manager/gomod/artifacts.ts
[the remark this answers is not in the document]
casaqori: `if (preserveGoDirectives === 'enabled'`

... would activate the stricter handling introduced by this PR
... that would also set the env GOTOOLCHAIN

`if (!preserveGoDirectives)`

... updates happen as go would do them out-of-the box (automatic toolchain switching & upgrading)

Either of the two modes would be the default, and that is to be discussed in thread https://github.com/renovatebot/renovate/discussions/35344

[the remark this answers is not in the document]
trim21: I'm ok to not change default but simply add new option to enable it

viceice:
> github UI is somehow broken, I can't apply them as batch, hope you don't mind:
> 
> image

you need to add each suggestion to the batch first

viceice (changes requested):
converting to draft, because it needs more discussion.

Feel free to suggest the `-d` handling in a new separate PR.

PR #34780, opened 2025-03-13
fix(versioning/hashicorp): Fully support semver in hashicorp constraint

neumantm:
## Changes

- In the hashicorp constraint versioning converter: Fully support semver (with prerelease and build)

## Context

- #34664

   - Documentation already links to <https://semver.org>

  - Real repository is on private GitLab Instance


lib/modules/versioning/hashicorp/convertor.ts
```diff
@@ -1,11 +1,18 @@
 import { logger } from '../../../logger';
 import { regEx } from '../../../util/regex';
 
+// Taken from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string. Licensed under CC BY 3.0
+// Removed the ^ and $.
+// Made minor and patch versions optional by surrounding them in parentheses followed by a question mark.
+const semverRegex =
```
viceice, line 7: use `!!semver.valid(value)` instead

neumantm: I see the point and initially agreed, that that is the cleaner option.
However, while trying to get this work, I noticed a problem:

`semver.valid` does not match partial semver like (`1.0`) even with `loose: true`.
We would need to use `semver.valid(semver.corece(value))`.
However, `semver.coerce` is much more liberal then just allowing partial semver.
I am not sure if we really want this.

What do you think?
1. Should we use `semver.coerce` here?
2. If we use `semver.coerce`: Should we return the coerced semver or the raw input?
3. Do you have any other ideas on using the semver package without needing to coerce?

lib/modules/versioning/hashicorp/convertor.ts
```diff
@@ -18,7 +25,11 @@ export function hashicorp2npm(input: string): string {
     .map((single) => {
       const r = single.match(
         regEx(
-          /^\s*(|=|!=|>|<|>=|<=|~>)\s*v?((\d+)(\.\d+){0,2}[\w-+]*(\.\d+)*)\s*$/,
+          /^\s*(?<operator>(|=|!=|>|<|>=|<=|~>))\s*v?/.source +
```
viceice, line 28: use string template syntax

lib/modules/versioning/hashicorp/convertor.ts
viceice: still not happy 🤔

lib/modules/versioning/hashicorp/convertor.ts
```diff
@@ -72,14 +83,20 @@ export function npm2hashicorp(input: string): string {
     .split(' ')
     .map((single) => {
       const r = single.match(
-        regEx(/^(|>|<|>=|<=|~|\^)v?((\d+)(\.\d+){0,2}[\w-]*(\.\d+)*)$/),
+        regEx(
+          /^(?<operator>(|>|<|>=|<=|~|\^))v?/.source +
```
viceice, line 87: same

viceice (approved):

PR #34791, opened 2025-03-13
feat(github-actions): add some community action support

### LINKED ISSUE #34767: Add support for `pnpm/action-setup` with `version:` updates
### Discussed in https://github.com/renovatebot/renovate/discussions/34724

<sup>Originally posted by **alexfaxe** March 10, 2025</sup>
### Tell us more.

Similar to the recent feature addition done in https://github.com/renovatebot/renovate/issues/34658

The `github-actions` manager should be updated to extract the version value when using [pnpm/action-setup](https://github.com/pnpm/action-setup).

Example syntax:
```
- name: Setup pnpm
      uses: pnpm/action-setup@v4
      with:
        version: 10.6.1
```</div>


Use datasource=npm and packageName=pnpm

### PR DESCRIPTION
**Pull request by trim21**
## Changes

add support for github actions
- `astral-sh/setup-uv` 
- `jaxxstorm/action-install-gh-release` 
-  `pnpm/action-setup`
- `prefix-dev/setup-pixi`

## Context

- close https://github.com/renovatebot/renovate/issues/34767

### CODE REVIEW

**Review by rarkins — DISMISSED** [review]

File: lib/modules/manager/github-actions/extract.spec.ts [hunk 1]
```diff
@@ -705,5 +706,109 @@ describe('modules/manager/github-actions/extract', () => {
         },
       ]);
     });
+
+    it('extracts package from community actions', () => {
+      const yamlContent = dump({
```
viceice, line 711: I would prefer to use inline yaml codeblocks as we do on other tests.

lib/modules/manager/github-actions/extract.ts
```diff
@@ -246,6 +334,17 @@ function extractWithYAMLParser(
     };
 
     for (const step of coerceArray(job?.steps)) {
+      if (step.uses) {
+        for (const action of communityActions) {
```
viceice, line 338: please create a zod union type and parse once.

lib/modules/manager/github-actions/extract.ts
```diff
@@ -246,6 +334,17 @@ function extractWithYAMLParser(
...
+          if (action.use.test(step.uses)) {
+            const val = Result.parse(step, action.schema).unwrapOrNull();
```
viceice, line 340: this hides all errors, you need to log them as debug, so users are aware of wrongly configured actions.

lib/modules/manager/github-actions/extract.ts
```diff
@@ -193,6 +198,89 @@ function extractRunners(runner: unknown): PackageDependency[] {
...
+          depName: val.repo,
+          packageName: val.repo,
+          currentValue: val.tag,
+          depType: 'uses-with',
+        };
+      }),
+  },
+  {
+    // https://github.com/astral-sh/setup-uv
+    use: /^astral-sh\/setup-uv@.*$/,
```
viceice, line 219: this manager supports gitea / forgejo actions too. there this could be the full repo url.

viceice (changes requested):

lib/modules/manager/github-actions/extract.ts
trim21: I don't think we need error log for case actions without `with`? This warning are useless to users.

trim21: > I don't think we need error log for case like actions without `with`? This warning are useless to users IMO

trim21: And how to do this when I rewrite it with a zod union?

package.json
```diff
@@ -197,6 +197,7 @@
     "emoji-regex": "10.4.0",
     "emojibase": "16.0.0",
     "emojibase-regex": "16.0.0",
+    "escape-string-regexp": "4.0.0",
```
trim21, line 200: like other nodejs package of sindresorhu, escape-string-regexp@4 is cjs package and escape-string-regexp@5 is esm package

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import type { PackageDependency } from '../types';
+
+function matchAction(action: string): z.ZodString {
+  return z
+    .string()
+    .regex(
+      new RegExp(
+        String.raw`^(https://github\.com/)?` +
+          escapeStringRegexp(action) +
```
viceice, line 18:
```suggestion
          escapeRegExp(action) +
```

from `utils/regex`

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,110 @@
...
+import * as pep440versioning from '../../versioning/pep440';
+
+import type { PackageDependency } from '../types';
+
+function matchAction(action: string): z.ZodString {
+  return z
+    .string()
+    .regex(
+      new RegExp(
+        String.raw`^(https://github\.com/)?` +
```
viceice, line 17: why `String.raw` ? can't we use a simple string template?

package.json
viceice: drop it, see other comment. 😉

viceice (commented):
conflicted

lib/modules/manager/github-actions/community.ts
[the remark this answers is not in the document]
trim21: https://gitea.com/actions only contains github official actions

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,118 @@
...
+import * as pep440versioning from '../../versioning/pep440';
+
+import type { PackageDependency } from '../types';
+
+function matchFullUrl(domain: string, action: string): z.ZodString {
+  return z
+    .string()
+    .regex(
+      regEx(
+        escapeRegExp('https://' + domain + '/') +
```
viceice, line 18: use string templates

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,118 @@
...
+ * there may be some actions use env as arguments version.
+ *
+ * each type should return `PackageDependency | undefined`
+ */
+export const communityActions = z.union([
+  z
+    .object({
+      // https://github.com/astral-sh/setup-uv
+      uses: matchAction('astral-sh/setup-uv'),
+      with: z.object({ version: z.string().refine((s) => s !== 'latest') }),
```
viceice, line 45: what happens if the version isn't defined or `latest`? shouldn't cause an error debug message because of failed zod schema

lib/modules/manager/github-actions/extract.ts
```diff
@@ -182,6 +184,13 @@ function extractSteps(
   deps: PackageDependency<Record<string, any>>[],
 ): void {
   for (const step of steps) {
+    if (step.uses) {
+      const val = Result.parse(step, communityActions).unwrapOrNull();
+      if (val) {
+        deps.push(val);
```
viceice, line 190:
```suggestion
        deps.push(val);
        continue;
```
I don't think we need to check further actions below?

lib/modules/manager/github-actions/schema.ts
```diff
@@ -8,7 +8,9 @@ import {
 
 const StepsSchema = z.object({
   uses: z.string(),
-  with: LooseRecord(z.string()),
+  with: LooseRecord(
+    z.union([z.string(), z.number().transform((s) => s.toString())]),
```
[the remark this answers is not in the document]
trim21: maybe, using bool is valid in some case, but it's currently not used by renovatebot, so I didn't add it.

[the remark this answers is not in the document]
trim21: If someone need to oneday we can add it then?

lib/modules/manager/github-actions/community.ts
trim21: Fo you mean a skipped package with reason `unspecified-version`?

I'm not sure how to handle `latest` here becasue it's not a valid version spec but a special case of this action.

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,118 @@
...
+      ),
+    );
+}
+
+function matchAction(action: string): z.Schema {
+  return z.union([
+    z.string().regex(regEx(escapeRegExp(action) + '(@.+)?$')),
+    matchFullUrl('github.com', action),
+    matchFullUrl('code.forgejo.org', action),
+    matchFullUrl('data.forgejo.org', action),
```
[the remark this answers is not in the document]
trim21: so you want to match `pnpm/setup-pnpm` and `https://*/pnpm/setup-pnpm` ?

lib/modules/manager/github-actions/community.ts
trim21: do you mean a nexted string template ``  `${escapeRegExp(`https://${domain}/`)}${escapeRegExp(action)}(@.+)?$` ``?

trim21:
there are some lines not covered, I'm not sure what the implement maintainers want in these case, so I didn't add tests yet. I'll add them later.

lib/modules/manager/github-actions/community.ts
rarkins: Normally we'd pass it through as a "dep" but with either empty `currentValue` or set to `"latest"`. Then the versioning would take care of it and it would have `skipReason: "invalid-value"`.

trim21: Yeah that's why I skip it in the first place, it's invalid in versioning but it's valid here.

rarkins: It can be good for users to see it in logs even if it's disabled (instead of silently dropping it)

trim21:
any more progress?  👀

lib/modules/manager/github-actions/community.ts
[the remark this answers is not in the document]
rarkins: Can we try to allow _any_ host, or does that result in too much false positives? If not, can we allow the list of known public endpoints + the user's `endpoint` somehow? Maybe this validation shouldn't be in Zod if so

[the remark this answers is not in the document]
trim21: Yeah I think we can, and also I don't think there will be many false positives if we match any hosts.

trim21:
👀

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,134 @@
...
+    .regex(regEx(`(?:https?://[^/]+/)?${escapeRegExp(action)}(@.+)?$`));
+}
+
+/**
+ * schema here should match the whole step,
+ * there may be some actions use env as arguments version.
+ *
+ * each type should return `PackageDependency | undefined`
+ */
+export const communityActions = z.union([
```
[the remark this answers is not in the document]
trim21: Do you mean to split it into `SetupPdmAction` and `SetupUVAction` and check steps on each schemas to see if it match?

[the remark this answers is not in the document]
zharinov: Maybe not exactly with these names, but yes

trim21:
any new code review? 👀

lib/modules/manager/github-actions/extract.ts
```diff
@@ -182,6 +184,14 @@ function extractSteps(
   deps: PackageDependency<Record<string, any>>[],
 ): void {
   for (const step of steps) {
+    if (step.uses) {
```
zharinov, line 187: Won't parsing work without this explicit check?

zharinov (approved):

lib/modules/manager/github-actions/extract.ts
trim21: yes, remove this still work. 

do you want to this check?

zharinov: Checks like these are why we use schemas, so I would prefer to remove it

rarkins (dismissed):

trim21:
@viceice can you approve this PR？ your previous review is blocking the auto-mergo

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,144 @@
...
+      skipReason = 'unspecified-version';
+    }
+
+    return {
+      datasource: NpmDatasource.id,
+      depName: 'pnpm',
+      versioning: npmVersioning.id,
+      packageName: 'pnpm',
+      skipReason,
+      skipStage,
```
viceice, line 67:
```suggestion
      ...(skipStage && { skipStage }),
      ...(skipReason && { skipReason }),
```

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,144 @@
...
+      skipReason = 'unspecified-version';
+    }
+
+    return {
+      datasource: GithubReleasesDatasource.id,
+      depName: 'astral-sh/uv',
+      versioning: npmVersioning.id,
+      packageName: 'astral-sh/uv',
+      skipReason,
+      skipStage,
```
viceice, line 40: ```suggestion```suggestion
      ...(skipStage && { skipStage }),
      ...(skipReason && { skipReason }),
```
```

lib/modules/manager/github-actions/extract.ts
```diff
@@ -182,6 +184,12 @@ function extractSteps(
   deps: PackageDependency<Record<string, any>>[],
 ): void {
   for (const step of steps) {
+    const val = Result.parse(step, communityActions).unwrapOrNull();
+    if (val) {
+      deps.push(val);
+      continue;
+    }
```
viceice, line 191:
```suggestion
    const res = communityActions.safeParse(step);
    if (res.success) {
      deps.push(res.data);
      continue;
    }
```
simplify

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,144 @@
...
+      currentValue: val.version,
+      depType: 'uses-with',
+    };
+  });
+
+const setupPDMSchema = z
+  .object({
+    uses: matchAction('pdm-project/setup-pdm'),
+    with: z.object({ version: z.string().optional() }),
```
viceice, line 76:
```suggestion
    with: z.object({ version: z.string().optional() }).optional(),
```

lib/modules/manager/github-actions/schema.ts
```diff
@@ -8,7 +8,9 @@ import {
...
+  ),
```
viceice, line 13:
```suggestion
  ).optional(),
```
🤔  should `with` optional here?

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,144 @@
...
+  return z
+    .string()
+    .regex(regEx(`(?:https?://[^/]+/)?${escapeRegExp(action)}(@.+)?$`));
+}
+
+const setupUVSchema = z
+  .object({
+    // https://github.com/astral-sh/setup-uv
+    uses: matchAction('astral-sh/setup-uv'),
+    with: z.object({ version: z.string().optional() }),
```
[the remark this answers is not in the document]
trim21: I think we should not add optional here, it does not match step without `with`.

lib/modules/manager/github-actions/schema.ts
trim21: not really, we do not use it yet

wyardley: Thanks -- as @viceice mentioned, seems like this PR will also solve the issue described in #36129 when the version is a float or int, not a string.

I scanned this quickly and didn't see a test for a case like that - would it make sense to add a test here (see diff in https://github.com/renovatebot/renovate/discussions/36129#discussioncomment-13265736) to cover the case where the input version is a number? And might be good to call out this fix in the description and / or break out this update first separately)?

trim21:
I'm waiting for response from reviewers

lib/modules/manager/github-actions/extract.spec.ts
```diff
@@ -852,4 +853,234 @@ describe('modules/manager/github-actions/extract', () => {
...
+          depName: 'astral-sh/uv',
+          depType: 'uses-with',
+          packageName: 'astral-sh/uv',
+          versioning: 'npm',
+        },
+      ],
+    },
+    {
+      step: {
+        name: 'Pinning a minor version of uv',
```
viceice: remove or update, it doesn't match the test

```suggestion
```

viceice (commented):
otherwise LGTM

trim21:
I'm not sure why e2e test failed....

trim21:
the e2e tests failure is caused by a upstream bug https://github.com/fb55/css-select/issues/1592

rarkins:
Should we add a resolutions override to main?

trim21:
> Should we add a resolutions override to main?

yes

rarkins:
@RahulGautamSingh can you?

trim21:
it's fixed in upstream, so I think we do not need it now.

But i can't re-trigger the e2e ci.

trim21:
can we have some code review on this?

lib/modules/manager/github-actions/extract.spec.ts
RahulGautamSingh, line 1: I think we can use a fixture file here instead of the big json object. It will make it more readable.

RahulGautamSingh (commented):
~Could you also add some explanation in the readme file related tot he actions in the `community.ts` file?~

↑ Not needed

lib/modules/manager/github-actions/extract.spec.ts
trim21: can you explain what kind of fixture you would like?

RahulGautamSingh: A normal one which has all the actions in which are in the test.?

RahulGautamSingh: Actually, it's fine without it, too.

RahulGautamSingh (approved):

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,144 @@
...
+import { PypiDatasource } from '../../datasource/pypi';
+import * as condaVersioning from '../../versioning/conda';
+import * as npmVersioning from '../../versioning/npm';
+import * as pep440versioning from '../../versioning/pep440';
+import type { PackageDependency } from '../types';
+
+function matchAction(action: string): z.Schema {
+  return z
+    .string()
+    .regex(regEx(`(?:https?://[^/]+/)?${escapeRegExp(action)}(@.+)?$`));
```
viceice:
```suggestion
    .regex(regEx(`(?:https?://[^/]+/)?${escapeRegExp(action)}(?:@.+)?$`));
```

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+const setupUVSchema = z
```
viceice:
```suggestion
const SetupUVSchema = z
```

We use `PascalCase` for all out schema.

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,144 @@
...
+      versioning: npmVersioning.id,
+      packageName: 'astral-sh/uv',
+      ...(skipStage && { skipStage }),
+      ...(skipReason && { skipReason }),
+      currentValue: val.version,
+      depType: 'uses-with',
+    };
+  });
+
+const setupPnpmSchema = z
```
viceice:
```suggestion
const SetupPnpmSchema = z
```

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,144 @@
...
+      versioning: npmVersioning.id,
+      packageName: 'pnpm',
+      ...(skipStage && { skipStage }),
+      ...(skipReason && { skipReason }),
+      currentValue: val.version,
+      depType: 'uses-with',
+    };
+  });
+
+const setupPDMSchema = z
```
viceice:
```suggestion
const SetupPDMSchema = z
```

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,144 @@
...
+      versioning: pep440versioning.id,
+      packageName: 'pdm',
+      currentValue: val.version,
+      ...(skipStage && { skipStage }),
+      ...(skipReason && { skipReason }),
+      depType: 'uses-with',
+    };
+  });
+
+const installBianrySchema = z
```
viceice:
```suggestion
const InstallBinarySchema = z
```

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,144 @@
...
+    return {
+      datasource: GithubReleasesDatasource.id,
+      depName: val.repo,
+      packageName: val.repo,
+      currentValue: val.tag,
+      depType: 'uses-with',
+    };
+  });
+
+const setupPixiSchema = z
```
viceice:
```suggestion
const SetupPixiSchema = z
```

lib/modules/manager/github-actions/community.ts
```diff
@@ -0,0 +1,144 @@
...
+ * there may be some actions use env as arguments version.
+ *
+ * each type should return `PackageDependency | undefined`
+ */
+export const communityActions = z.union([
+  setupUVSchema,
+  setupPnpmSchema,
+  setupPDMSchema,
+  installBianrySchema,
+  setupPixiSchema,
```
viceice:
```suggestion
export const CommunityActions = z.union([
  InstallBinarySchema,
  SetupPDMSchema,
  SetupPixiSchema,
  SetupPnpmSchema,
  SetupUVSchema,
```

Please sort schema by name, so they can be easier found and maintained

lib/modules/manager/github-actions/extract.ts
```diff
@@ -15,6 +15,7 @@ import type {
   PackageDependency,
   PackageFileContent,
 } from '../types';
+import { communityActions } from './community';
```
viceice:
```suggestion
import { CommunityActions } from './community';
```

lib/modules/manager/github-actions/extract.ts
```diff
@@ -182,6 +183,12 @@ function extractSteps(
   deps: PackageDependency<Record<string, any>>[],
 ): void {
   for (const step of steps) {
+    const res = communityActions.safeParse(step);
```
viceice:
```suggestion
    const res = CommunityActions.safeParse(step);
```

viceice (approved):

Issue #34127, opened 2025-02-10
Abandoned packages detection


rarkins:
### Describe the proposed change(s).

This is a parent Issue for a feature to:
- Detect the "abandonment" status of packages (ones which seem abandoned, even if not officially deprecated)
- Provide ways of reporting on this

A package can be "deprecated" OR "abandoned", but not both, because abandonment is a way of saying "seems deprecated but not officially".

joebnb:
pls consider support npm first https://github.com/renovatebot/renovate/discussions/33199


![Image](https://github.com/user-attachments/assets/beb53dba-71d7-4d10-9cb8-4dc00317e50c)


need information  like this
![Image](https://github.com/user-attachments/assets/f81c1210-ef6a-4bef-b106-064c6f889c0d)

rarkins:
That's deprecated packages, not abandoned. Please start a new discussion.

kivra-pauoli:
👍 thanks for the new feature

We're getting, from https://github.com/imranismail/setup-kustomize/releases, an abandonment date of Oct 24, 2023 (`v2.1.0-rc`), but we're already at the latest version, which from that list should be seen as Feb 26, 2024 (`v2.1.0`).

I don't know the inners of the algo, but is this what's expected?

rarkins:
No, that doesn't sound right. Please create a discussion with a reproduction repo and we'll debug it

zharinov:
Reproduction: https://github.com/renovate-testing/test-abandonment/

zharinov:
Just found the root cause: that's because `v2` tag points to `v2.1.0-rc` instead of `v2.1.0`, while our versioning comparison prioritizes `v2` over `v2.1.0` when selecting the most recent version.

kivra-pauoli:
👍 
that'd also explain why:
- https://github.com/minchao/setup-gator/tree/v2 is shown as latest, Jun 5, 2024, and not
- https://github.com/minchao/setup-gator/tree/v2.0.0, Feb 15, 2024, though in that case it just seems wrong that they don't point to the same commit (and Renovate is effectively picking the most recent date, even if "by chance")

PR #34794, opened 2025-03-13
feat: Add `bumpedAt` field

zharinov:
## Context

- Ref: #34127


viceice (dismissed):

lib/workers/repository/process/lookup/timestamps.ts
```diff
@@ -0,0 +1,33 @@
+import type { ReleaseResult } from '../../../../modules/datasource/types';
+import type { VersioningApi } from '../../../../modules/versioning/types';
+import { type Timestamp, asTimestamp } from '../../../../util/timestamp';
+
+export function calculateLatestReleaseTimestamp(
```
rarkins, line 5: For npm do you define "latestReleaseTimestamp" as the timestamp of the release tagged as `latest`? Or do you simply mean "the most recent timestamp of any release"? Because I think to change the name if it's the latter, e.g. to `mostRecentReleaseTimestamp` or `newestReleaseTimestamp`.

zharinov:
It's the `releaseTimestamp` of the highest version available. I decided not to select max timestamp, because I think if the datasource stop receive timestamps for some reason, it's unacceptable to return the timestamp of older versions.

lib/workers/repository/process/lookup/timestamps.ts
zharinov: I would go further and name it `bumpedAt`

zharinov:
So the safest we could do, is the to take the timestamp of latest version and make sure it's the highest of all other timestamps. Any alternatives?

zharinov:
Maybe we could log such cases and then ignore them via preset or hardcoded list?

rarkins:
Is there any scenario where we _don't_ want to use the newest timestamp? I'm not sure how likely it is that a package goes from having timestamps to not having them.

zharinov:
It could happen when used with merge datasource strategy

zharinov:
I would suggest to start with the safest logic, see the logs and relax it if there is a need. It'll help avoid false positives, while for most datasources it will just work fine.

lib/workers/repository/process/lookup/timestamps.ts
rarkins, line 1: Please document the logic in this file.

Also, if the `latest` release is a higher timestamp than the highest release timestamp, we definitely should use that

rarkins (changes requested):

lib/workers/repository/process/lookup/timestamps.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    try {
+      if (
+        versioningApi.isGreaterThan(release.version, highestRelease.version)
+      ) {
+        highestRelease = release;
+        continue;
+      }
+      // eslint-disable-next-line no-empty
+    } catch /* v8 ignore next */ {}
```
viceice, line 28: shouldn't we at least log a trace message if it fails? so you can remap it for debugging purposes?

it's also easily testable because you can pass a mock versioning which throws

lib/workers/repository/process/lookup/timestamps.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  if (!highestRelease) {
+    return releaseResult;
```
viceice, line 32: should we add additional logging (trace) so users can more easily see what's going on?

lib/workers/repository/process/lookup/timestamps.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  if (highestRelease.isDeprecated) {
+    return releaseResult;
```
viceice, line 36: same here

lib/workers/repository/process/lookup/timestamps.ts
```diff
@@ -0,0 +1,57 @@
...
+      const releaseTimestamp = asTimestamp(release.releaseTimestamp);
+      if (!releaseTimestamp) {
+        return false;
+      }
+
+      return DateTime.fromISO(releaseTimestamp) > highestReleaseDatetime;
+    });
+
+    if (!higherTimestampExists) {
+      releaseResult.bumpedAt = highestReleaseTimestamp;
```
viceice, line 52: same here

lib/workers/repository/process/lookup/timestamps.ts
zharinov: I think this couldn't happen because we're calculating the highest version of all, and the `latest` could only be lower than this.

viceice (approved):

rarkins (approved):

PR #34805, opened 2025-03-14
feat(presets): group Puppeteer packages together

wereHamster:
## Changes

See https://github.com/renovatebot/renovate/discussions/34784

### DIFF
--- a/lib/config/presets/internal/group.ts
+++ b/lib/config/presets/internal/group.ts
@@ -457,6 +457,7 @@ const staticGroups = {
           'puppeteer-core'
         ],
       },
+    ],
   },
   react: {
     description: 'Group React and corresponding `@types` packages together.',
--- a/lib/config/presets/internal/group.ts
+++ b/lib/config/presets/internal/group.ts
@@ -446,6 +446,18 @@ const staticGroups = {
       },
     ],
   },
+  puppeteer: {
+    description: 'Group Puppeteer packages together.',
+    packageRules: [
+      {
+        groupName: 'Puppeteer',
+        matchDatasources: ['npm'],
+        matchPackageNames: [
+          'puppeteer',
+          'puppeteer-core'
+        ],
+      },
+  },
   react: {
     description: 'Group React and corresponding `@types` packages together.',
     packageRules: [
@@ -485,6 +497,7 @@ const staticGroups = {
       'group:micrometer',
       'group:phpstan',
       'group:polymer',
+      'group:puppeteer',
       'group:react',
       'group:remark',
       'group:resilience4j',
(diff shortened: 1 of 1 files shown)


rarkins (changes requested):
Please add the repo to https://github.com/renovatebot/renovate/blob/main/lib/data/monorepo.json instead

wereHamster:
@rarkins even though the monorepo contains packages with different versions? wouldn't that mean renovate may create immortal PRs?

wereHamster:
https://github.com/renovatebot/renovate/pull/34808

viceice:
> @rarkins even though the monorepo contains packages with different versions? wouldn't that mean renovate may create immortal PRs?

No. we don't group them by default when not needed and they've published with different versions. This would cause imortal PR's , which is bad.

rarkins:
Our preference:
- Update monorepo packages together if they have the same version (preferred)
- Update monorepo packages together even if they don't have the same version, IF they need to be updated together
- No monorepo grouping, or selective grouping

If you think it's the third option which applies (i.e. the packages with a different version can be updated independently) then you're right and this PR is best

wereHamster:
The https://github.com/puppeteer/puppeteer/ repository contains 5 packages:

 - `@puppeteer/browsers`: 2.8.0
 - `@puppeteer/ng-schematics`: 0.7.0
 - `@pptr/testserver`: 0.6.1
 - `puppeteer-core`: 24.4.0
 - `puppeteer`: 24.4.0

Most people will only install `puppeteer`. Fewer will install `puppeteer-core` directly. Fewer still both at the same time. In my case, I have `puppeteer-core` as dependency, and `puppeteer` as dev dependency.

The `@puppeteer/browsers` is a dependency of both `puppeteer-core` and `puppeteer`. The `@pptr/testserver` package is used to test puppeteer itself, I doubt any real project seriously uses it as a dependency. The `@puppeteer/ng-schematics` package is some kind of integration that Angular projects can use in their E2E tests.

IMHO, it makes sense to only group `puppeteer-core` and `puppeteer`, since they share the same version and are published in tandem. These two packages will then pull in whatever version of `@puppeteer/browsers` they need, but the majority of users won't care about what exact version that is.

rarkins:
@viceice wdyt? Our choice is:
- Keep it simple with a URL because it seems unlikely people have packages with mismatched versions, or
- Have a manual rule for just `puppeteer` and `puppeteer-core`

viceice (approved):
I think in this case I think we should use the manual rule

rarkins (approved):

wereHamster:
Thanks :) this change is much appreciated.

image

PR #34820, opened 2025-03-14
feat: add LOG_FILE_FORMAT

bdovaz:
## Changes

I have added the behavior of the new environment variable `LOG_FILE_FORMAT`:
* If the variable is not defined, it will behave as before.
* If this variable is defined, it is not a empty or whitespace string and the value is equals to `pretty` it will write a pretty log as it is already done with stdout.

## Context

https://github.com/renovatebot/renovate/discussions/34810


lib/logger/index.ts
```diff
@@ -63,11 +63,22 @@ function createLogFileStream(logFile: string): bunyan.Stream {
...
     name: 'logfile',
     path: logFile,
     level: validateLogLevel(getEnv('LOG_FILE_LEVEL'), 'debug'),
   };
+
+  const logFileFormat = getEnv('LOG_FILE_FORMAT');
+
+  if (
+    is.nonEmptyStringAndNotWhitespace(logFileFormat) &&
+    logFileFormat !== 'json'
```
viceice, line 76: no, we should not support arbitrary values as valid

viceice (changes requested):

lib/logger/index.ts
bdovaz: So what do I do, `logFileFormat == 'pretty'`? It's just that in the other case of `LOG_FORMAT` it doesn't really check against a fixed list of formats either, it just looks to see if it's different from `json`.

That is, in the code that already is, the check is arbitrary, just look that it is different from json, not a concrete value like `pretty` for example.

https://github.com/renovatebot/renovate/blob/470f03ccfc9efbd3492a848c91ac8c8d2b0acf8d/lib/logger/index.ts#L37

bdovaz: I mean that for `LOG_FORMAT` is not documented the valid values other than `json`, it should be documented the current two possible values.

rarkins:
Should we make this one `RENOVATE_LOG_FILE_FORMAT`?

bdovaz:
> Should we make this one `RENOVATE_LOG_FILE_FORMAT`?

If you are referring to renaming this variable, I have relied on the convention of existing variables:

https://docs.renovatebot.com/config-overview/#logging-variables

rarkins:
Yes, but we want to _change_ those in future so that we always use RENOVATE_ prefixes

bdovaz:
> Yes, but we want to _change_ those in future so that we always use RENOVATE_ prefixes

Then I change it? Because it will be strange that "temporarily" there are some with this prefix (this PR) and the rest that do not have it, no?

rarkins:
@viceice do you think we should keep it consistent for now, and change all later?

lib/logger/index.ts
viceice: Fallback to `pretty` if the given value isn't `json`. I'm not sure if we can already warn the user here about an invalid config option. 🤔 

@rarkins WDYT? Shcouzld we exit early when an invalid value is passed? Would probably my preference but could make trouble if `LOG_FILE_FORMAT` is used for something else.

bdovaz:
> > @viceice do you think we should keep it consistent for now, and change all later?
> 
> I would change the docs in a separate PR merged before this, so we have the prefixed version only in official docs. We can still fallback to un-prefixed as we do now.

I understand that some of you (@rarkins  / @viceice) are going to create that PR you mention and I simply wait for that? As far as I am concerned, unless there are any changes to be made (like the prefix), it is ready to merge if you see fit.

bdovaz:
I'm still waiting for renovate team

RahulGautamSingh:
> I'm still waiting for renovate team

My bad. I understood the comments differently

lib/logger/index.ts
rarkins: We should aim to retain existing behavior, which means using JSON format unless we're sure it should be pretty.

Therefore,
If pretty then pretty
Else json

rarkins (dismissed):

bdovaz:
Is there anything else needed to be merged? I see that it has already been approved and auto merge enabled by @rarkins.

rarkins:
All previous reviewers must approve

lib/logger/index.spec.ts
```diff
@@ -32,6 +32,10 @@ vi.mock('nanoid', () => ({
 const bunyanDebugSpy = vi.spyOn(bunyan.prototype, 'debug');
 
 describe('logger/index', () => {
+  beforeEach(() => {
+    process.env = {};
```
viceice:
```suggestion
    delete process.env.LOG_FILE_LEVEL;
```

don't override full env

viceice (dismissed):

bdovaz:
Why the cancelled tests?

lib/logger/index.spec.ts
```diff
@@ -202,6 +206,58 @@ describe('logger/index', () => {
         { name: 'logfile' },
       ]);
     });
+
+    it.each([
+      {
+        logFileLevel: null,
```
viceice, line 212: it seems this test fails.

bdovaz: Maybe your last change? I think that all test passed when I requested a review

viceice: maybe, but resetting the full process env is a no-go. only delete env values which were added by tests

lib/logger/index.spec.ts
bdovaz: I relied on other tests that do that, there are several places where there are references:

https://github.com/search?q=repo%3Arenovatebot%2Frenovate%20process.env%20%3D%20%7B%7D%3B&type=code

viceice (approved):

PR #34823, opened 2025-03-14
fix(versioning/docker): pure semver prerelease support

felipecrs:
## Changes

Fix support for pure semver prerelease versions in docker versioning.

## Context

Docker images can be tagged with pure semver prerelease versions, e.g. `1.2.3-beta.1`. Renovate currently doesn't support this and disregards everything after the `-` in the tag.

Note that none of the existing tests were modified, only new test cases where added. This means the new code should not break any existing behavior.


felipecrs:
@viceice or @rarkins, do you know why upgrading major within prereleases are not allowed?

https://github.com/felipecrs/renovate/blob/9b2fb90e74eed151d55942aaf1bf97ab09215a93/lib/workers/repository/process/lookup/filter.ts#L155

This prevents upgrading versions like `1.2.3-4` to `2.0.0-1`.

felipecrs:
build-docs is failing for some unrelated reason.

felipecrs:
Ok, this has one issue. It would parse:

```
11.0.0-alpine-3.12
```

As if it was a semver prerelease. There are no current tests to cover this case.

PR #34827, opened 2025-03-15
chore: disable strict mkdocs

rarkins:

### DIFF
--- a/tools/mkdocs.ts
+++ b/tools/mkdocs.ts
@@ -23,7 +23,7 @@ program
     logger.info('* running mkdocs build');
     const args = ['run', 'mkdocs', 'build'];
     if (opts.strict) {
-      args.push('--strict');
+      // args.push('--strict');
     }
     const res = exec('pdm', args, {
       cwd: 'tools/mkdocs',
@@ -43,7 +43,7 @@ program
     logger.info('* running mkdocs serve');
     const args = ['run', 'mkdocs', 'serve'];
     if (opts.strict) {
-      args.push('--strict');
+      // args.push('--strict');
     }
     const res = exec('pdm', args, {
       cwd: 'tools/mkdocs',


viceice (approved):

viceice:
we should ask in mkdocs material repo if there's a better option. or maybe to not use cdn for js files.

@HonkingGoose you suggested us to enable the bundling flag for privacy, which now causes the build issues. do you have an idea how to move forward?

HonkingGoose:
From the mkdocs-material version 9.6.8 release notes:

> Fixed https://github.com/squidfunk/mkdocs-material/issues/8076: Privacy plugin crashes if HTTP download fails

Maybe we should try this?

1. Update to mkdocs-material 9.6.8.
2. Revert PR #34827.
3. _Maybe_ our docs build starts working again?

viceice:
> From the mkdocs-material version 9.6.8 release notes:
> 
> > Fixed https://github.com/squidfunk/mkdocs-material/issues/8076: Privacy plugin crashes if HTTP download fails
> 
> Maybe we should try this?
> 
> 1. Update to mkdocs-material 9.6.8.
> 2. Revert PR #34827.
> 3. _Maybe_ our docs build starts working again?
> 
> Related upstream issue report, which the new mkdocs-material version should resolve:
> 
> - https://github.com/squidfunk/mkdocs-material/issues/8076

it doesn't solve it, docs build doesn't crash. it fails because of download errors/ warnings on strict mode. 😔

HonkingGoose:
The strict mode stops us from having broken links in the docs. When the CDN failure is fixed, we should start using strict mode again.

It's okay to use the normal mode docs builds, to get us unstuck. But that should only be temporary. :wink: 

I think we can either:

1. Wait for CDN to get fixed, then re-enable strict mode.
2. Wait for some kind of privacy plugin refactor, where a failure to download does not affect strict mode builds, and then enable strict mode again.

I prefer option 1, as that gets us using strict mode sooner. As always, I'll you Renovate maintainers weigh the pros/cons, and decide. :wink: 

## Relevant discussions on the Material for MkDocs repo

The Material for MkDocs maintainer confirms that "strict mode build error on failure to download" is _intended_ behavior:

> Thanks for reporting. That is, in fact, intended behavior. See the reasoning in https://github.com/squidfunk/mkdocs-material/issues/8076#issuecomment-2718478154.

I think the maintainer wants to change something about the "download failure leads to strict mode failure", the next time they work on the privacy plugin:

> Fixed in [74a7835](https://github.com/squidfunk/mkdocs-material/commit/74a7835b356b5e06257244c5bf6aba3f1398b883). We should probably track the state in the list of futures that we manage for downloading, but for now this should be fine. The plugin will now print a warning, so if you want to fail the build on assets that can't be downloaded, [use `strict` mode](https://www.mkdocs.org/user-guide/configuration/#strict). We'll refactor it the next time we work on the privacy plugin.

Sources:

- https://github.com/squidfunk/mkdocs-material/issues/8091#issuecomment-2724855649
- https://github.com/squidfunk/mkdocs-material/issues/8076#issuecomment-2718478154

Issue #34829, opened 2025-03-15
tj-actions/changed-files compromise


rarkins:
This is not a Renovate _issue_ per se, but I'm posting this here for maximum visibility.

1. An attacker had enough access to the tj-actions/changed-files repo to be able to configure/reconfigure tags somehow

2. The attacker added malicious code to their fork of the repo, and chose to spoof Renovate as the committer, potentially because it was a Renovate commit which was the most recent commit in the repo. GitHub has always allowed git commits to be spoofed - it's a known feature/limitation depending on who you ask

3. Important: this spoofing of commits wasn't done to "trick" a maintainer into accepting any PR, instead it was just to obfuscate it a little. It was an orphan commit and not on top of main or any other branch

4. As you'd expect, the commit showed up as Unverified, although if we're being realistic, most people don't look at that or enforce signed commits only (the real bot signs its commits)

5. Kind of unrelated, but the "real" Renovate Bot - just like Dependabot presumably - then started proposing PRs to update the action, like it does any other outdated dependency

6. Some people had automerging of such updates enabled, but this is not Renovate's default behavior. Even without automerging, an action like this might be able to achieve its aim only with a PR, if it's run as part of PR builds

7. This incident has reminded that many people mistakenly assume that git tags are immutable, especially if they are in semver format. Although it's rare for such tags to be changed, they are not immutable by design

jamietanna:
Want to say thanks for the super speedy release of https://github.com/renovatebot/renovate/pull/34825 to protect folks from getting auto-updates - makes a big difference 💜

janderssonse:
Thanks for the quick update, just a quick note also, I dont think the container image was updated from 39.202.. to 39.203.. with the fix yet, looking at https://github.com/renovatebot/renovate/pkgs/container/renovate.

viceice:
it seems the issue is fixed upstream, so we can drop the workaround 

- https://github.com/tj-actions/changed-files/issues/2470

rarkins:
Yes let's drop it so that we can at least upgrade anyone who was impacted

JossWhittle:
Posting here in the hopes of some cross visibility since the safest thing to do is start pinning actions digests to make them immutable.

https://github.com/renovatebot/renovate/discussions/34889 

[...]

PR #34837, opened 2025-03-16
docs(upgrade best practices): update preset code plus small rewrite

HonkingGoose:
## Changes

- Drop the intro text (as the sidebar has the table of contents already, and I want the reader to focus on the general recommendations)
- Small grammar fixes
- Update the JSON for the `config:best-practices` preset, so it matches the code of the preset again
- Recommend users automerging third-party dependencies tell Renovate to wait 14 days. This way upstream registries have time to pull the malicious packages before Renovate automerges it

## Context

Update the JSON in our docs to match the presets code. Plus some small rewrites. I'm not closing any issue with my PR.


docs/usage/upgrade-best-practices.md
```diff
@@ -45,17 +42,17 @@ The [`config:best-practices` preset](./presets-config.md#configbest-practices) h
...
+    ":configMigration",
     ":pinDevDependencies"
   ]
 }
 ```
 
-The next sections explain each part of the preset.
+The next sections explain what each part of the preset does.
 
 #### Config migration
```
HonkingGoose, line 57: Do we still want to recommend users automerge third-party dependencies (like Prettier or ESLint)? Full quote of the current docs:

> ### You'll look for ways to automate the updates
>
> Finally, when you're updating often, you'll start looking for ways to automate the updates. You may start to [`automerge`](./configuration-options.md#automerge) development dependencies like Prettier, or ESLint when the linter passes. Or you may decide to automerge any `patch` type upgrades, by using the [`default:automergePatch`](./presets-default.md#automergepatch) preset.
>
> You may also start using [GitHub's pull request merge queues](./key-concepts/automerge.md#github-merge-queue) to speed up the merge process.
> Renovate does not support GitLab's Merge Trains, see [issue #5573](https://github.com/renovatebot/renovate/issues/5573).

I'm thinking of removing the mention of automerging dependencies, because of things like:

- #34829

I think the safest thing is:

- For a human to review each PR
- For a human to decide whether to merge the PR
- Optionally: use Merge Queues to speed up the merge process

Therefore I want to drop the bit about automerging, but leave in the Merge Queue bits. What do you think maintainers? :wink:

HonkingGoose: > It's no safer if humans realistically either:
> 
>     * don't look at the code, or
> 
>     * don't have the security ability to determine if it's a risk

Good point!

> Human review of the tj-actions one would have done practically nothing at all, for example. A much more pragmatic approach is to delay updates by a week like we do in this repo because the majority of malicious incidents are found within the first week.

Hmm... So which option should we go for in our recommendation then? 😄  

### Option 1. Automerge, but wait a week?

Are we happy to assume the upstream project, or GitHub, or the registry will have blocked the malicious dependency by that time?

The failure state I'm thinking of is:

1. Renovate sees update that turns out to be malicious.
2. Renovate waits a week, the project _doesn't_ pull the release in time for Renovate to skip the malicious update.
3. Renovate automerges malicious update anyway...

Waiting a week will probably help prevent merging bad dependency updates. But waiting does _not_ guarantee the update will be "clean/safe" by that time.

### Option 2: No automerges, but set minumumReleaseAge to wait a week?

To nudge humans into doing a manual review: stop recommending automerge in the best practices guide. Instead recommend users set `minimumReleaseAge` to 7 days, so Renovate is more likely to create a safe PR?

### Option 3: something else?

If you have a better idea, please let me know. 😉

rarkins: I think that if you automerge, for third party dependencies then ideally wait a week or two. You can always open them earlier on-demand through the dashboard if you really wish.

rarkins (approved):

PR #34863, opened 2025-03-17
fix(nix): urldecode gitlab subgroups

xanderio:
Do to a bug in nix one needs to uri encode the owner of a gitlab repository that resides in a subgroups, in this case the `/` as `%20F`. This however breaks the `git ls-remotes` which renovate runs to discover if the input can be updated.

This commit passes the owner attribute through `decodeURIComponent` to ensure that `git` is unware of this workaround on nixs side.

Nix PR: https://github.com/NixOS/nix/pull/9163

### DIFF
--- a/lib/modules/manager/nix/extract.spec.ts
+++ b/lib/modules/manager/nix/extract.spec.ts
@@ -602,4 +602,45 @@ describe('modules/manager/nix/extract', () => {
       ],
     });
   });
+
+  const flake12Lock = `{
+    "nodes": {
+      "subgroup-project": {
+        "locked": {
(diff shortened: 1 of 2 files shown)


rarkins (approved):

lib/modules/manager/nix/extract.spec.ts
```diff
@@ -602,4 +602,45 @@ describe('modules/manager/nix/extract', () => {
       ],
     });
   });
+
+  const flake12Lock = `{
```
viceice, line 606: should be moved inside the test and use `codeBlock` template helper

Issue #34949, opened 2025-03-23
Add long term cache for Maven Central results


rarkins:
### Describe the proposed change(s).

This feature is primarily for Maven Central, but it's OK for it to be enabled for any public Maven datasource which also supports etags.

We should implement caching for Maven Central results similar for npm:
- Use a one-week "hard" TTL for caching to Redis etc
- Use the regular cache expiry as the "soft" TTL

The approach for npm seems to have worked well, including the cache revision for cache busting, so let's reuse it for Maven too.

PR #34956, opened 2025-03-23
feat: Add auth header check for package cache provider

zharinov:
## Changes

- Prevent caching when `Authorization` header is present (could be disabled via global config)
- Rename `checkCacheControl` to `checkCacheControlHeader`
  - **Changed**: this check will be skipped if the `Cache-Control` is missing
- Refactor the private method responsible for that

## Context

- Ref: #34949


lib/util/http/cache/abstract-http-cache-provider.ts
```diff
@@ -66,11 +66,11 @@ export abstract class AbstractHttpCacheProvider implements HttpCacheProvider {
         timestamp,
       });
 
-      /* v8 ignore next 4 -- should never happen */
+      /* v8 ignore start: should never happen */
```
viceice, line 69: why this change 🤔

viceice (commented):
otherwise LGTM

lib/util/http/cache/abstract-http-cache-provider.ts
zharinov: I don't like the line count approach

rarkins: I dislike both approaches, but start/stop seems better of the two

viceice (approved):

rarkins (commented):
Docker Hub and GitHub are two examples where we are likely to find both:
- Credentials
- Public packages

Will this make any changes to the caching of either of those?

viceice (dismissed):

zharinov:
| `cachePrivatePackages` | `checkCacheControlHeader` | `Cache-Control~=public` | `resp.authorization` | `cacheAllowed` (result) |
|------------------------|---------------------------|-----------------------------------|----------------------|----------------------|
| ✅ | * | * | * | ✅ |
| ❌ | ✅ | ✅ | ✅ | ❌ |
| ❌ | ✅ | ✅ | ❌ | ✅ |
| ❌ | ✅ | ❌ | * | ❌ |
| ❌ | ❌ | * | ✅ | ❌ |
| ❌ | ❌ | * | ❌ | ✅ |

rarkins:
I care most about any changes to github or Docker Hub datasources, where they are mostly public but we have auth

zharinov:
Okay, then I guess we must have flag for disabling this check

viceice (approved):

PR #34999, opened 2025-03-25
feat(devcontainer): extract known feature versions

viceice:
## Changes
Extract some known feature versions like `node` and `python`


lib/modules/manager/devcontainer/extract.ts
```diff
@@ -43,6 +47,39 @@ export function extractPackageFile(
           featureDep.depType = 'feature';
           featureDep.pinDigests = false;
           deps.push(featureDep);
+          if (!value.version) {
+            continue;
```
secustor, line 51: We should return here a dependency with skipReason `no-version`.
So that they are showing up in the logs, reports and such

rarkins (approved):

PR #35010, opened 2025-03-25
feat(bun): workspaces

### LINKED ISSUE #24459: bun: workspace support

Use the `workspaces` field in `package.json` to detect workspace package.json files

The goal would be that a reproduction like https://github.com/renovate-reproductions/24459b/issues/3 would show all three package files as "bun" and none as "npm".

Likely all work is contained to https://github.com/renovatebot/renovate/tree/main/lib/modules/manager/bun

Today Renovate's bun manager:
- is triggered by a lock file
- locates the adjacent package.json
- parses the package.json
- returns results for package.json + bun.lock(b)

What needs to be done is:
- After parsing the package.json adjacent to the lock file, look for "workspaces" config inside the package.json
- Locate any matching package.json deeper in the repo
- For each of those, parse them as package.json and put the results into the returned "packageFiles", ensuring that the `lockFiles` correctly points to the top level lock file

This alone may be enough for it to work.

maxdewald:
## Changes

Renovate now respects the `workspaces` field in the `package.json`. It now finds the `bun.lock` in the project root and applies package updates there.

## Context

Closes #24459 
Based on the work of @rarkins 
https://github.com/renovatebot/renovate/issues/24459#issuecomment-2655807439

- [X] Both unit tests + ran on a real repository ( Tested on my companies repo, which is quite complex)


Norbiros:
Are we waiting for something? Or it can be merged?

rarkins (dismissed):

lib/modules/manager/bun/extract.spec.ts
```diff
@@ -125,4 +125,76 @@ describe('modules/manager/bun/extract', () => {
...
+        }),
+      );
+      // Only the main package.json file is processed
+      const result = await extractAllPackageFiles({}, [
+        'bun.lock',
+        'package.json',
+        'packages/pkg1/package.json',
+      ]);
+      expect(result).toHaveLength(1);
+      expect(result[0].packageFile).toBe('package.json');
```
secustor, line 197: See above

lib/modules/manager/bun/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+  fileName: string,
+  workspaces: string[],
+): boolean {
+  if (!fileName.startsWith(pwd)) {
+    return false;
+  }
+  const relativeFile = fileName
+    .slice(pwd.length)
+    .replace(/^\//, '')
+    .replace(/\/package\.json$/, '');
```
secustor, line 14: Use the upath library

lib/modules/manager/bun/index.ts
```diff
@@ -13,7 +13,7 @@ export const supersedesManagers = ['npm'];
 export const supportsLockFileMaintenance = true;
 
 export const defaultConfig = {
-  fileMatch: ['(^|/)bun\\.lockb?$'],
+  fileMatch: ['(^|/)bun\\.lockb?$', '(^|/)package\\.json$'],
```
secustor, line 16: @rarkins  Do you remember why the lockfile is here in the filematch and not the package file?

rarkins: @secustor my goal originally had been to trigger bun logic only if there's a bun lockfile, and not for the 99% of cases where there's non-bun package.json files. However, I think this PR is the most pragmatic way of supporting workspaces in bun for now.

The alternative to changing `fileMatch` I considered is:
- Passing the include/ignore files list through to the extractAllPackageFiles function
- Letting the extractAllPackageFiles function search for related files within that filtered list
- etc

(If we didn't pass the result if include/ignore per-manager then we can end up with extracted results which don't honor the config)

But that seemed quite complicated too.

Norbiros:
@rarkins @secustor, sorry for pings, but this PR is very important for our project...

rarkins (dismissed):

secustor (approved):

PR #35024, opened 2025-03-26
fix: log disabled updates count

rarkins:
## Changes

Checks explicitly for enabled=false plus adds logging of the filtered count

## Context

If a package rule disables a particular update, it doesn't seem to be reflected in the logs anywhere

### DIFF
--- a/lib/workers/repository/updates/flatten.ts
+++ b/lib/workers/repository/updates/flatten.ts
@@ -217,20 +217,13 @@ export async function flattenUpdates(
       update.semanticCommits = semanticCommits;
     }
   }
-  let filteredOutCount = 0;
   const filteredUpdates = updates
-    .filter((update) => {
-      if (update.enabled === false) {
-        filteredOutCount += 1;
-        return false;
-      }
-      return true;
-    })
+    .filter((update) => update.enabled !== false)
     .map(({ vulnerabilityAlerts, ...update }) => update)
     .map((update) => filterConfig(update, 'branch'));
-  if (filteredOutCount) {
+  if (filteredUpdates.length < updates.length) {
     logger.debug(
-      `Filtered out ${filteredOutCount} disabled update(s). ${filteredUpdates.length} update(s) remaining.`,
+      `Filtered out ${updates.length - filteredUpdates.length} disabled update(s). ${filteredUpdates.length} update(s) remaining.`,
     );
   }
   return filteredUpdates;
--- a/lib/workers/repository/updates/flatten.ts
+++ b/lib/workers/repository/updates/flatten.ts
@@ -4,6 +4,7 @@ import {
   mergeChildConfig,
 } from '../../../config';
 import type { RenovateConfig } from '../../../config/types';
+import { logger } from '../../../logger';
 import { getDefaultConfig } from '../../../modules/datasource';
 import { get } from '../../../modules/manager';
 import { detectSemanticCommits } from '../../../util/git/semantic';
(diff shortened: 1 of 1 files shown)


lib/workers/repository/updates/flatten.ts
```diff
@@ -216,8 +217,21 @@ export async function flattenUpdates(
       update.semanticCommits = semanticCommits;
     }
   }
-  return updates
-    .filter((update) => update.enabled)
```
rarkins, line 220: This original code assumes that `enabled` is always boolean and never missing. My new code does not assume that.

lib/workers/repository/updates/flatten.ts
```diff
@@ -216,8 +217,21 @@ export async function flattenUpdates(
...
+  let filteredOutCount = 0;
+  const filteredUpdates = updates
+    .filter((update) => {
+      if (update.enabled === false) {
```
rarkins, line 223: This original code assumes that `enabled` is always boolean and never missing. My new code does not assume that. I assume this wasn't a bug I'm fixing, but my new approach is more defensive

lib/workers/repository/updates/flatten.ts
```diff
@@ -216,8 +217,21 @@ export async function flattenUpdates(
...
+        filteredOutCount += 1;
+        return false;
+      }
+      return true;
+    })
     .map(({ vulnerabilityAlerts, ...update }) => update)
     .map((update) => filterConfig(update, 'branch'));
+  if (filteredOutCount) {
```
viceice, line 231: why not simply
```suggestion
  if (filteredUpadates.length <  updates.length) {
```
so above becomes simpler.

lib/workers/repository/updates/flatten.ts
```diff
@@ -216,8 +217,14 @@ export async function flattenUpdates(
...
   }
-  return updates
-    .filter((update) => update.enabled)
+  const filteredUpdates = updates
+    .filter((update) => update.enabled !== false)
     .map(({ vulnerabilityAlerts, ...update }) => update)
     .map((update) => filterConfig(update, 'branch'));
+  if (filteredUpdates.length < updates.length) {
+    logger.debug(
+      `Filtered out ${updates.length - filteredUpdates.length} disabled update(s). ${filteredUpdates.length} update(s) remaining.`,
```
secustor, line 226:
```suggestion
      `Filtered out disabled ${updates.length - filteredUpdates.length} disabled update(s). ${filteredUpdates.length} update(s) remaining.`,
```
IMO this makes it clearer for users why they are filtered out.

lib/workers/repository/updates/flatten.ts
```diff
@@ -216,8 +217,14 @@ export async function flattenUpdates(
       update.semanticCommits = semanticCommits;
     }
   }
-  return updates
-    .filter((update) => update.enabled)
+  const filteredUpdates = updates
+    .filter((update) => update.enabled !== false)
```
secustor, line 221: Is that attribute optional, if yes it changes the behavior

lib/workers/repository/updates/flatten.ts
viceice: duplicate disabled ☝️

lib/workers/repository/updates/flatten.ts
viceice: see previous comments, it seems `enabled` is never `null` or `undefined`

viceice (approved):

secustor (approved):

PR #35025, opened 2025-03-26
docs: add a note how to fix minimum release age with GAR's maven

iamstarkov:
## Context

Earlier today I opened a discussion https://github.com/renovatebot/renovate/discussions/35017

> ### Tell us more.

> after tj-actions incident we decided reevaluate our stance on immediate bumping of dependencies and we want to introduce minimum release age requirement. We happen to use java and GAR and apparently GAR's maven implementation doesn't support `last-modified` header. Which renders minimum release age option less useful than we want it to be.
>
> We will be talking to GAR support shortly to find a way to remediate it, but I dont think we should have expectations to get it fixed in timely manner.
>
> We notice renovate having confidence badges and one of them is badge with age. And it appears to be internal to renovate database. Because in the logs `[org.junit.jupiter:junit-jupiter-api](https://junit.org/junit5/)@5.12.1` mentioned as `DEBUG: No currentVersionTimestamp for org.junit.jupiter:junit-jupiter-api` while at the same time featuring 11 days old on a age badge [age](https://docs.renovatebot.com/merge-confidence/).
>
> I understand minimum release age feature and merge confidence feature probably got developed independently and there is no intersection between them right now.
>
> Will it hurt to use merge confidence age information as a fallback source for `currentVersionTimestamp` if underlying registry misses to deliver such information itself? Obviously, this behavior can be behind another option.

After some thoughts we discovered that we can add maven central on before GAR's maven and it will make it work, which I documented in the discussion and then decided to contribute to documentation, so more people are aware of the workaround.

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2447,7 +2447,7 @@ In those cases a feature request needs to be implemented.
     For `minimumReleaseAge` to work, the Maven source must return reliable `last-modified` headers.
 
     <!-- markdownlint-disable MD046 -->
-    If your custom maven source registry doesnt support `last-modified` header, like GAR (Google Artifact Registry's Maven implementation, then feasible option is to extend maven source registry url with https://repo1.maven.org/maven2 as the first item, then `currentVersionTimestamp` via `last-modified` will be taken from maven central for public dependencies.
+    If your custom Maven source registry does _not_ support the `last-modified` header, like GAR (Google Artifact Registry's Maven implementation) then you can extend the Maven source registry url with `https://repo1.maven.org/maven2` as the first item. Then the `currentVersionTimestamp` via `last-modified` will be taken from Maven central for public dependencies.
 
     ```json
     "registryUrls": [
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2447,7 +2447,7 @@ In those cases a feature request needs to be implemented.
     For `minimumReleaseAge` to work, the Maven source must return reliable `last-modified` headers.
 
     <!-- markdownlint-disable MD046 -->
-    If your custom Maven source registry does _not_ support the `last-modified` header, like GAR (Google Artifact Registry's Maven implementation) then you can extend the Maven source registry url with `https://repo1.maven.org/maven2` as the first item. Then the `currentVersionTimestamp` via `last-modified` will be taken from Maven central for public dependencies.
+    If your custom Maven source registry is **full proxy or pull-through** and does _not_ support the `last-modified` header, like GAR (Google Artifact Registry's Maven implementation) then you can extend the Maven source registry url with `https://repo1.maven.org/maven2` as the first item. Then the `currentVersionTimestamp` via `last-modified` will be taken from Maven central for public dependencies.
 
     ```json
     "registryUrls": [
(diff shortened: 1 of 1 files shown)


docs/usage/configuration-options.md
```diff
@@ -2446,6 +2446,18 @@ In those cases a feature request needs to be implemented.
 !!! warning "Warning for Maven users"
     For `minimumReleaseAge` to work, the Maven source must return reliable `last-modified` headers.
 
+    If your custom maven source registry doesnt support `last-modified` header, like GAR (Google Artifact Registry's Maven implementation, then feasible option is to extend maven source registry url with https://repo1.maven.org/maven2 as the first item, then `currentVersionTimestamp` via `last-modified` will be taken from maven central for public dependencies.
+    
+    <!-- markdownlint-disable MD046 -->
```
iamstarkov, line 2451: I disabled the rule, because it seems to be false negative. because it expects fenced code, and it is, but it thinks its indented
```
MD046/code-block-style Code block style [Expected: fenced; Actual: indented] https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md046.md
```

docs/usage/configuration-options.md
```diff
@@ -2446,6 +2446,17 @@ In those cases a feature request needs to be implemented.
 !!! warning "Warning for Maven users"
     For `minimumReleaseAge` to work, the Maven source must return reliable `last-modified` headers.
 
+    <!-- markdownlint-disable MD046 -->
```
iamstarkov, line 2449: I disabled the rule, because it seems to be false negative. because it treats continuation of admonition as a indented code which it is not intended to be, but at the same time MkDocs documentation on admonitions is not clear if multi paragraph admonitions are even possible https://squidfunk.github.io/mkdocs-material/reference/admonitions/#inline-blocks-inline-end

```
MD046/code-block-style Code block style [Expected: fenced; Actual: indented] https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md046.md
```

docs/usage/configuration-options.md
```diff
@@ -2446,6 +2446,17 @@ In those cases a feature request needs to be implemented.
...
+    If your custom maven source registry doesnt support `last-modified` header, like GAR (Google Artifact Registry's Maven implementation, then feasible option is to extend maven source registry url with https://repo1.maven.org/maven2 as the first item, then `currentVersionTimestamp` via `last-modified` will be taken from maven central for public dependencies.
```
rarkins, line 2450: Are you sure this doesn't make a _functional_ difference? Are you sure that it applies so widely? e.g. does it only apply if your maven registry is a full proxy/pull-through of maven central in the first place? Otherwise you could get versions on Maven Central which aren't in your own registry

HonkingGoose, line 2450:
```suggestion
    If your custom Maven source registry does _not_ support the `last-modified` header, like GAR (Google Artifact Registry's Maven implementation) then you can extend the Maven source registry url with `https://repo1.maven.org/maven2` as the first item. Then the `currentVersionTimestamp` via `last-modified` will be taken from Maven central for public dependencies.
```

HonkingGoose (changes requested):

docs/usage/configuration-options.md
iamstarkov: you are right, it makes functional difference if custom maven registry is not full proxy or pull through. although in that case it doesn't make sense to combine both in the same `match` rule. perhaps its worth it to specify explicitly?

iamstarkov: clarified that `[`45370dc` (#35025)](https://github.com/renovatebot/renovate/pull/35025/commits/45370dc182c3330c1b672daf8b58da0353ef5db0)`

rarkins: If your proxy is only for pull-through then you could leave it out of registryUrls completely, right? You'd only need to keep it in if it's a combination of public Maven Central + private packages.

iamstarkov: yes, that's what we have, it will be good to clarify that. I've removed full proxy and left only pull-through

rarkins (approved):

iamstarkov:
I'm glad new changes were approved. As you are probably aware, I'm not authorized to push to `main`, so I'm not sure what happens next

iamstarkov:
sweet, thank you!

PR #35048, opened 2025-03-28
fix(host-rules): respect `hostRules.enabled=false` even when `noAuth=true`

yueki-work:
## Changes

* Previously, when `options.noAuth = true`, `hostRules.enabled = false` setting was not respected
* This caused problems in cases where access to certain hosts—such as index.docker.io—was meant to be blocked by disabling specific hostRules.
* This change ensures that `hostRules.enabled = false` is properly respected, even when `options.noAuth = true`.

The below screenshot is to show
* trying to access index.docker.io
* `hostRule.enabled = false`
* `options.noAuth = true`
* Thus else-if block (l.146) won't be executed
image

https://github.com/yueki-work/renovate-issue-hostRules-disable/blob/main/renovate.json

### DIFF
--- a/lib/util/http/host-rules.ts
+++ b/lib/util/http/host-rules.ts
@@ -122,7 +122,7 @@ export function applyHostRule<GotOptions extends HostRulesGotOptions>(
   options: GotOptions,
   hostRule: HostRule,
 ): GotOptions {
-  const { username, password, token, enabled, authType } = hostRule;
+  const { username, password, token, authType } = hostRule;
   const host = parseUrl(url)?.host;
   if (options.noAuth) {
     logger.trace({ url }, `Authorization disabled`);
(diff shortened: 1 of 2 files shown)


lib/util/http/host-rules.ts
```diff
@@ -157,6 +155,10 @@ export function applyHostRule<GotOptions extends HostRulesGotOptions>(
     options.abortIgnoreStatusCodes = hostRule.abortIgnoreStatusCodes;
   }
 
+  if (hostRule.enabled === false) {
+    options.enabled = false;
+  }
+
```
viceice, line 161: I think this can be moved to start of function and an early return added. if it's disabled all other code is irrelevant and can be skipped.

viceice (approved):

PR #35052, opened 2025-03-28
feat: added support for package.yaml in npm

Dracks:
## Changes

Adds support to package.yaml files in npm

## Context

With the last versions of pnpm you can get your package.json in yaml format, since I hate modifying the json file because almost allways I get an error, I start using in my projects yaml format, which is more friendly (at least to me). But renovate doesn't support the yaml.

- [x] No documentation update is required (Not 100% sure is correct, but I hope if I'm wrong somebody will point me in the right direction ;) )


lib/modules/manager/npm/index.ts
```diff
@@ -20,7 +20,11 @@ export const url = 'https://docs.npmjs.com';
 export const categories: Category[] = ['js'];
 
 export const defaultConfig = {
-  fileMatch: ['(^|/)package\\.json$', '(^|/)pnpm-workspace\\.yaml$'],
+  fileMatch: [
+    '(^|/)package\\.json$',
+    '(^|/)package\\.yaml$',
```
rarkins, line 25: I would prefer to revert this change and keep it as-is initially. You and any others can modify `fileMatch` for now and if we find no unintended consequences then it can be added as a default file match.

If we keep the fileMatch unchanged, we can merge this if we're confident it doesn't cause regressions for JSON. But otherwise we need to be more careful / slower to approve.

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -30,6 +31,18 @@ function renameObjKey(
   }, {} as DependenciesMeta);
 }
 
+function getStrValue(isYaml: boolean, text: string): string {
+  if (!isYaml) {
+    return `"${text}"`;
+  }
+  const simpleString = new RegExp(/^[a-z0-9]/i).exec(text);
+  if (simpleString) {
```
rarkins, line 39: Please document (with code comments) why this special case is necessary.

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -30,6 +31,18 @@ function renameObjKey(
   }, {} as DependenciesMeta);
 }
 
+function getStrValue(isYaml: boolean, text: string): string {
+  if (!isYaml) {
+    return `"${text}"`;
+  }
+  const simpleString = new RegExp(/^[a-z0-9]/i).exec(text);
```
rarkins, line 38:
```suggestion
  const simpleString = new RegExp(/^[a-z0-9]/i).exec(text);
```

Use `regEx()` instead

rarkins: So `foo` or `1` or `foo1` are simple strings, but `1.2` would not be?

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -76,20 +90,21 @@ function replaceAsString(
...
+  let newString = getStrValue(isYaml, newValue);
 
   const escapedDepName = escapeRegExp(depName);
   const patchRe = regEx(`^(patch:${escapedDepName}@(npm:)?).*#`);
   const match = patchRe.exec(oldValue);
   if (match && depType === 'resolutions') {
     const patch = oldValue.replace(match[0], `${match[1]}${newValue}#`);
     parsedContents[depType]![depName] = patch;
-    newString = `"${patch}"`;
+    newString = getStrValue(isYaml, patch);
```
rarkins, line 102: In theory, this could result in us going from quoted to non-quoted or vice versa? I wonder if this would break things if you go from a quoted simple string to a non-simple string. I think you'd get doubled quotes.

rarkins (changes requested):

lib/modules/manager/npm/update/dependency/index.ts
Dracks: no, I only check that starts with any letter or number.

lib/modules/manager/npm/update/dependency/index.ts
Dracks: theorically this is only for yaml files, And since any yaml generator usually tends to remove quotes when are not needed, keeping the quotes always will generate constant git changes any time that you run pnpm install or similar stuff, as the new one will drop the quotes.

lib/modules/manager/npm/index.ts
Dracks: One question, should I add somewhere in the logs that this is possible? Otherwise any other person using the yaml and trying to use renovate will think yaml is not supported.

lib/modules/manager/npm/update/dependency/index.ts
rarkins: But it's entirely possible that someone will have a "simple" value in their yaml wrapped in quotes (you cannot assume that the quotes will be removed/non-present). And then update to a non-simple value. I think that from `1` to `1.2.3` is an example (e.g. dependency pinning). Please add a test case for that to check

Dracks: I realise while fixing the other comments, that in fact, for versioning the only part which are necessary is for the dependency name, I changed everything in the case of the yaml to search without the quotes, and replace without the quotes, that should keep it as the same than before. (no add or remove quotes)

For your example 1 and 1.2.3 both cases in my old code was without quotes.

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -129,8 +137,20 @@ export function updateDependency({
   newValue = getNewNpmAliasValue(newValue, upgrade) ?? newValue;
 
   logger.debug(`npm.updateDependency(): ${depType}.${depName} = ${newValue}`);
+
+  let isYaml = false;
```
rarkins, line 141: Can we instead update the `updateDependency()` function to include an `isYaml` parameter so we don't have to try/catch/try/catch?

rarkins (changes requested):

viceice (commented):
@rarkins I think we should first move the pnpm manager out of the npm manager. like we've implementd bun. this would make the implementation much simpler,

rarkins:
@viceice I think that's going to be a lot of work, and I'm not sure anyone has time for it right now

Dracks:
@viceice @rarkins should I try to get this changes done? Or this will be fully dropped?

rarkins:
@Dracks we'll discuss amongst the maintainers and get back to you, thanks for your patience

Dracks:
> @Dracks we'll discuss amongst the maintainers and get back to you, thanks for your patience

All good, I did a change yesterday, which is a better solution, but having the good solution I wished, will be a lot of work, which I'm afraid can broke things. For the moment I created a custom workflow to my project to get the renovate working with my code

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -113,10 +121,13 @@ function replaceAsString(
...
-export function updateDependency({
-  fileContent,
-  upgrade,
-}: UpdateDependencyConfig): string | null {
+export function updateDependency(
+  { fileContent, upgrade }: UpdateDependencyConfig,
+  fileFormat?: string,
+): string | null {
+  const isYaml = fileFormat
+    ? fileFormat.endsWith('.yaml') || fileFormat.endsWith('.yml')
```
rarkins: What is "fileFormat"? is it the file _name_?

Dracks: Right now I don't remember, I think you are right, It was really long time ago, and this ended evolving from somewere else.

lib/modules/manager/npm/index.ts
RahulGautamSingh: Add in the `npm/readme.md` file

lib/modules/manager/npm/extract/index.spec.ts
```diff
@@ -113,6 +114,33 @@ describe('modules/manager/npm/extract/index', () => {
       });
     });
 
+    it('returns an array of dependencies when is a yaml version too', async () => {
```
RahulGautamSingh: Unclear description

lib/modules/manager/npm/extract/index.spec.ts
```diff
@@ -113,6 +114,33 @@ describe('modules/manager/npm/extract/index', () => {
...
+      const res = await npmExtract.extractPackageFile(
+        input01YamlContent,
+        'package.yaml',
+        defaultExtractConfig,
+      );
+      expect(res).toMatchSnapshot({
```
RahulGautamSingh: Use `toMatchObject` instead. We prefer not to add new snapshots as it affects performance

RahulGautamSingh (changes requested):
Thanks for the PR and your patience.  From initial review, the changes seem fine and non-breaking to me.

Please fix the following though:

1. The two fixtures 011 and 012 can be removed and replaced with string samples since they only interact with one dep at a time from the whole file.

2. Imports are not in order, and some failing tests.

lib/modules/manager/npm/extract/index.spec.ts
Dracks: the test is the same as previous test, and has a similar text, I changed it a little bit, but since is the same test for the different format, I think it should have the same (or similar description)

Dracks:
> Thanks for the PR and your patience. From initial review, the changes seem fine and non-breaking to me.
> 
> Please fix the following though:
> 
>     1. The two fixtures 011 and 012 can be removed and replaced with string samples since they only interact with one dep at a time from the whole file.
> 
>     2. Imports are not in order, and some failing tests.

Should I remove also the fixture 01.yaml ? I see the pipeline complains about that one.

RahulGautamSingh:
> Should I remove also the fixture 01.yaml ? I

No, it's good to have one fixture for each file type

lib/modules/manager/npm/extract/index.spec.ts
```diff
@@ -117,6 +127,33 @@ describe('modules/manager/npm/extract/index', () => {
...
     });
 
+    it('returns an array of dependencies in yaml', async () => {
+      const res = await npmExtract.extractPackageFile(
+        input01YamlContent,
+        'package.yaml',
+        defaultExtractConfig,
+      );
+      expect(res).toMatchObject({
+        deps: [
```
RahulGautamSingh, line 147: Also check the dependencyType so we verify the devDependencies and normal ones.

Would be also good if we can add devEngines, engines, packageManager and volta fields here, so we can verify all work

lib/modules/manager/npm/update/dependency/index.spec.ts
```diff
@@ -24,6 +26,148 @@ describe('modules/manager/npm/update/dependency/index', () => {
       expect(testContent).toEqual(outputContent);
     });
 
+    it('replaces a dependency of a yaml value', () => {
+      const expectedOutput = codeBlock`
```
RahulGautamSingh: remove the untested deps here like in the test after it

lib/modules/manager/npm/update/dependency/index.spec.ts
```diff
@@ -24,6 +26,148 @@ describe('modules/manager/npm/update/dependency/index', () => {
...
+        {
+          fileContent: input,
+          upgrade,
+        },
+        'outputs/011.yml',
+      );
+      expect(testContent).toEqual(expectedOutput);
+    });
+
+    it('replaces a dependency of a yaml value when the source do not have quotes', () => {
```
RahulGautamSingh, line 87: same here, replace untested deps

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -133,8 +144,11 @@ export function updateDependency({
   newValue = getNewNpmAliasValue(newValue, upgrade) ?? newValue;
 
   logger.debug(`npm.updateDependency(): ${depType}.${depName} = ${newValue}`);
+
```
RahulGautamSingh:
```suggestion
```
revert

RahulGautamSingh (changes requested):

lib/modules/manager/npm/update/dependency/index.spec.ts
Dracks: What do you mean? to only check the line of the dependency that has changed?

lib/modules/manager/npm/extract/index.spec.ts
Dracks: Ok, do you mean to put the full object, right?

lib/modules/manager/npm/update/dependency/index.spec.ts
RahulGautamSingh: Yes, since we are updating cheerio only the codeblock string doesn't need the other deps

lib/modules/manager/npm/index.ts
Dracks: libs/modules/manager... ? right? 
Is added. Thanks!

lib/modules/manager/npm/update/dependency/index.spec.ts
Dracks: should be done, should I resolve your comment?

RahulGautamSingh: yes

lib/modules/manager/npm/extract/index.spec.ts
RahulGautamSingh: yes

lib/modules/manager/npm/extract/index.ts
```diff
@@ -255,7 +265,7 @@ export async function extractAllPackageFiles(
           });
         }
       } else {
-        if (packageFile.endsWith('json')) {
+        if (packageFile.startsWith('package.')) {
           logger.trace({ packageFile }, `Extracting as a package.json file`);
```
RahulGautamSingh: Change this too

lib/modules/manager/npm/extract/index.ts
```diff
@@ -255,7 +265,7 @@ export async function extractAllPackageFiles(
           });
         }
       } else {
-        if (packageFile.endsWith('json')) {
+        if (packageFile.startsWith('package.')) {
```
RahulGautamSingh, line 277: @viceice I wonder if this is suitable. should we rather match using regex for fullname i.e. `/package\.(json|yaml|yml)/`

lib/modules/manager/npm/readme.md
```diff
@@ -40,3 +40,18 @@ If set, and Renovate detects Yarn 2+, and one or both of those variables are pre
 This will result in the `~/.yarnrc.yml` file being created or modified with these settings, and the settings are not removed afterwards.
 
 Configuration/conversion of `NO_PROXY` to Yarn config is not supported.
+
+### pnpm
```
RahulGautamSingh, line 44: Is the yaml format supported only by pnpm manager or all (npm, pnpm, yarn)?

RahulGautamSingh (changes requested):

lib/modules/manager/npm/readme.md
Dracks: As far as I know, only pnpm gives support to that. (together with json5)

lib/modules/manager/npm/extract/index.ts
Dracks: Tbh, I found today that with the last version was not working anymore the yaml, and I did the simple version, but... Since there was asked explicitly to not have support directly to package.yaml, I was not sure about having a regex. Also, since PnPm supports json5 and this should have a different extension too, I'm not sure if then we should add json5 too.

lib/modules/manager/npm/readme.md
```diff
@@ -40,3 +40,18 @@ If set, and Renovate detects Yarn 2+, and one or both of those variables are pre
...
+
+When using yaml as package format, you need to modify your renovate.json to add a regex to your yaml to `managerFilePatterns`
```
RahulGautamSingh:
```suggestion
Renovate supports updating the dependencies in your `package.yaml` file.
To enable it you need to extend the `managerFilePatterns` for the npm manager as shown below:
```

lib/modules/manager/npm/readme.md
```diff
@@ -40,3 +40,18 @@ If set, and Renovate detects Yarn 2+, and one or both of those variables are pre
...
+
+Example:
+
+```json
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
```
RahulGautamSingh:
```suggestion
```

RahulGautamSingh (changes requested):

RahulGautamSingh (commented):
LGTM

lib/modules/manager/npm/extract/index.ts
Dracks: What's your opinion? should I use the regex? Or is it fine for the moment

lib/modules/manager/npm/readme.md
```diff
@@ -40,3 +40,18 @@ If set, and Renovate detects Yarn 2+, and one or both of those variables are pre
 This will result in the `~/.yarnrc.yml` file being created or modified with these settings, and the settings are not removed afterwards.
 
 Configuration/conversion of `NO_PROXY` to Yarn config is not supported.
+
+### pnpm
+
+Renovate supports updating the dependencies in your `package.yaml` file.
+To enable it you need to extend the `managerFilePatterns` for the npm manager as shown below:
+
+Example:
```
RahulGautamSingh:
```suggestion
```

RahulGautamSingh (approved):
LGTM except the unresolved comment. @viceice @secustor wdyt?? I think it's okay to have the `package.` string matching for now it should not break anything.

Also, please link the test repo here.

lib/modules/manager/npm/extract/index.ts
```diff
@@ -255,8 +265,8 @@ export async function extractAllPackageFiles(
           });
         }
       } else {
-        if (packageFile.endsWith('json')) {
-          logger.trace({ packageFile }, `Extracting as a package.json file`);
+        if (packageFile.startsWith('package.')) {
```
secustor, line 277: As you have described users could add a JSON5 file in here and then your changes to parsing and  replacement logic will no longer work. 

```suggestion
        if (regEx('package\\.(json?|ya?ml)$').test(packageFile)) {
```

Can we move these kind of logic into the pnpm directory. 
This looks messy.

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -174,6 +187,7 @@ export function updateDependency({
         depName,
         oldVersion!,
         `npm:${upgrade.newName}@${newValue}`,
+        isYaml,
```
secustor: I'd rather forward the packageFile here, which can be used for different purposes here too.

secustor (changes requested):
I don't like the complexity this is introducing :/

lib/modules/manager/npm/extract/index.ts
Dracks: Maybe I'm blind, but I don't see any directory for pnpm, I only see a pnpm.ts in the same folder as this index.ts

I'm not sure I'm following you about "adding a json5 file will invalidate my changes." Maybe I missunderstand you.

secustor: > Maybe I'm blind, but I don't see any directory for pnpm, I only see a pnpm.ts in the same folder as this index.ts

Sry, I have meant the `pnpm.ts` file. 

> I'm not sure I'm following you about "adding a json5 file will invalidate my changes." Maybe I missunderstand you.

As you have said that pnpm also supports `package.json5` it is expected that users would try to add `package.json5` to the fileMatch. 
With `packageFile.startsWith('package.')`, the `extractPackageFile` file function would be still called and then JSON parser and the replacement logic, which do not support JSON5 ATM.

Dracks: ok, you mean that it will crash trying to parse the json5 file, right? I will try to test something, maybe we can give support to json5 also.

Dracks: Hey, I tried to run the json5, I was hoping (as renovate config supports json5) it will work. But that was not the case :( I don't know, but maybe does  makes sense that I add support for json5? At the end, the error was that he didn't know which package manager was using the [repo](https://github.com/Dracks/renovate-tests-json5)

The error: 
``` 
      "loggerErrors": [
         {
           "name": "renovate",
           "level": 50,
           "logContext": "5-5ppHRpKT8LQdDV1xsi5",
           "repository": "dracks/renovate-test-json5",
           "err": {
             "message": "platform-unknown-error",
             "stack": "Error: platform-unknown-error\n    at Proxy.initRepo (/home/dracks/Documents/devel/node/renovate/lib/modules/platform/github/index.ts:2:15061)\n    at processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at getPlatformConfig (/home/dracks/Documents/devel/node/renovate/lib/workers/repository/init/apis.ts:2:3101)\n    at initApis (/home/dracks/Documents/devel/node/renovate/lib/workers/repository/init/apis.ts:2:3291)\n    at initRepo (/home/dracks/Documents/devel/node/renovate/lib/workers/repository/init/index.ts:2:2882)\n    at Object.renovateRepository (/home/dracks/Documents/devel/node/renovate/lib/workers/repository/index.ts:2:3323)\n    at attributes.repository (/home/dracks/Documents/devel/node/renovate/lib/workers/global/index.ts:2:6602)\n    at start (/home/dracks/Documents/devel/node/renovate/lib/workers/global/index.ts:2:6256)\n    at /home/dracks/Documents/devel/node/renovate/lib/renovate.ts:3:447"
           },
           "msg": "Repository has unknown error"
         }
       ]
```

lib/modules/manager/npm/extract/index.ts
```diff
@@ -41,11 +42,20 @@ export async function extractPackageFile(
...
+  if (packageFile.endsWith('.yaml') || packageFile.endsWith('.yml')) {
+    try {
+      packageJson = parseSingleYaml(content);
+    } catch {
+      logger.debug({ packageFile }, `Invalid YAML`);
+      return null;
+    }
+  } else {
+    try {
+      packageJson = JSON.parse(content);
```
RahulGautamSingh: If we use the existing utility function `parseJson` here then we can support `json5` as well.

lib/modules/manager/npm/extract/index.ts
RahulGautamSingh: I forked it and it was working.

Dracks: then, maybe is my repo that is wrong :(  I didn't get time last weeks. 

OMG, I'm blind, my test repo is renovate-tests-json5 no renovate-test-json5

lib/modules/manager/npm/extract/index.ts
Dracks: the function you comments returns nullish content, which mean a more complex code, but there is the fallback version that returns the json directly. I used that function. Also I needed to change the JSON.parse in multiple places. Now it work and generate the P/Rs in my test json5 project :)

lib/modules/manager/npm/update/dependency/index.ts
Dracks: I needed the fileName to add support to the json5, which I changed it.

lib/modules/manager/npm/extract/index.ts
Dracks: Hey Now that I added support for json5, you still wishes to have that regex? And you wishes that regex provided by the pnpm.ts file right? 

I needed to use some regex extra (I changed the endsWith to a regex) where is the best place to put it? since is used in extractor and update. Thanks!

lib/modules/manager/npm/__fixtures__/inputs/01.yaml
secustor, line 1: Please inline this in the test. 

The reason is that security tools will detect this fixture as outdated dependencies.
Also consider reducing the number of dependencies in your fixture

lib/modules/manager/npm/extract/index.ts
```diff
@@ -41,11 +43,20 @@ export async function extractPackageFile(
   logger.trace(`npm.extractPackageFile(${packageFile})`);
   logger.trace({ content });
   let packageJson: NpmPackage;
-  try {
-    packageJson = JSON.parse(content);
-  } catch {
-    logger.debug({ packageFile }, `Invalid JSON`);
-    return null;
+  if (/\.ya?ml$/.test(packageFile)) {
```
secustor, line 42: move this if/else into a helper function for better readability

lib/modules/manager/npm/readme.md
```diff
@@ -40,3 +40,26 @@ If set, and Renovate detects Yarn 2+, and one or both of those variables are pre
...
+
+```json
+{
+  "npm": {
+    "managerFilePatterns": ["/(^|/)package\\.json5$/"]
+  }
+}
+```
+
+This ire the sample repo: [package.yaml test repo](https://github.com/Dracks/renovate-tests) & [package.json5 test repo](https://github.com/Dracks/renovate-tests-json5)
```
secustor:
```suggestion
```

If you want to self promote you can add yourself to the contributor https://github.com/renovatebot/renovate/blob/main/package.json#L94

lib/modules/manager/npm/extract/index.ts
```diff
@@ -255,8 +266,8 @@ export async function extractAllPackageFiles(
           });
         }
       } else {
-        if (packageFile.endsWith('json')) {
-          logger.trace({ packageFile }, `Extracting as a package.json file`);
+        if (packageFile.startsWith('package.')) {
+          logger.trace({ packageFile }, `Extracting as a ${packageFile} file`);
```
secustor: Please revert that duplicates information and is not helpful as `packageFile` is the full path.

lib/modules/manager/npm/extract/index.ts
secustor: Though this will produce warnings if a users uses json5. 

Why are you thinking it will increase complexity of the code?

lib/modules/manager/types.ts
```diff
@@ -300,6 +300,7 @@ export interface ManagerApi extends ModuleApi {
 
   updateDependency?(
     updateDependencyConfig: UpdateDependencyConfig,
+    packageFile?: string,
```
secustor: Add `packageFile` to `UpdateDependencyConfig` instead

lib/modules/manager/npm/extract/index.ts
Dracks: well, if the parseJson can return null, I need to handle the null, which is more conditions, in contrary the parseJsonCallback returns directly the result which ends in a cleaner Code (at least from my point of view) But Maybe I'm using wrong functions to parse the json5.

lib/modules/manager/npm/readme.md
Dracks: No, I simply wished to show some repos as example . I don't wish to promote myself.

RahulGautamSingh:
@secustor Needs attention 🤗

Dracks:
Hey, I don't know if somebody else can take care of it, for the next 4~5 weeks I doubt having any time if something else is requested.

lib/modules/manager/types.ts
```diff
@@ -236,6 +236,7 @@ export interface UpdateArtifact<T = Record<string, unknown>> {
 export interface UpdateDependencyConfig<T = Record<string, any>> {
   fileContent: string;
   upgrade: Upgrade<T>;
+  packageFile?: string;
```
secustor, line 239:
```suggestion
  packageFile: string;
```

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -105,7 +110,14 @@ function replaceAsString(
         newString,
       );
       // Compare the parsed JSON structure of old and new
-      if (dequal(parsedContents, JSON.parse(testContent))) {
+      if (
+        dequal(
+          parsedContents,
+          isYaml
+            ? parseSingleYaml(testContent)
+            : (parseJsonWithFallback(testContent, fileName) as NpmPackage),
```
secustor, line 118: https://github.com/renovatebot/renovate/pull/35052/changes#r2431199116
```suggestion
          parsePackageContent(fileName, testContent)
```

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -134,7 +148,9 @@ export function updateDependency({
 
   logger.debug(`npm.updateDependency(): ${depType}.${depName} = ${newValue}`);
   try {
-    const parsedContents: NpmPackage = JSON.parse(fileContent);
+    const parsedContents: NpmPackage = isYaml
+      ? parseSingleYaml(fileContent)
+      : (parseJsonWithFallback(fileContent, packageFile) as NpmPackage);
```
secustor, line 153:
```suggestion
    const parsedContents = parsePackageContent(packageFile, fileContent);
```

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -77,20 +81,21 @@ function replaceAsString(
     parsedContents[depType]![depName] = newValue;
   }
   // Look for the old version number
-  const searchString = `"${oldValue}"`;
-  let newString = `"${newValue}"`;
+  const searchString = isYaml ? oldValue : `"${oldValue}"`;
```
secustor, line 84: Move this and other occurrences to a helper function. 
```suggestion
  const searchString = massageValue(filename, oldValue);
```

```ts
export function massageValue(filename: string, value: string): string {
  const isYaml = regEx(/\.ya?ml$/).test(filename); // caching regex 
  return isYaml ? oldValue : `"${oldValue}"`
}
```

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -117,7 +129,9 @@ function replaceAsString(
 export function updateDependency({
   fileContent,
   upgrade,
+  packageFile = 'package.json',
```
secustor, line 132:
```suggestion
packageFile
fileContent,
upgrade,
```
Ordering the logically and remove the optional

secustor (commented):
Sry for the late reply, but your review request got lost somewhere in my notifications

lib/modules/manager/types.ts
Dracks: If I remember correctly, I made this optional, because otherwise was clashing with other packages, like bun and deno.

RahulGautamSingh:
@Dracks Closing this for now, Feel free to open  when you can work on the suggestions.

PR #35056, opened 2025-03-28
fix(gerrit): memCache being used all the time

felipecrs:
## Changes

Fixes memCache being used all the time in Gerrit platform.

Turns out that `memCache` option when not specified is assumed to be `true` by default.

This PR will make sure to properly and explicitly set `memCache` to `false` always, and use `memCacheProvider` only in the situations where `memCache` was previously allowed.

Note it also aligns the implementation with BitBucket platform to make the code better understandable, but it doesn't change the behavior there.

## Context

This fixes a regression of https://github.com/renovatebot/renovate/pull/33901, as reported [here](https://github.com/renovatebot/renovate/pull/33901#issuecomment-2758897013).

Such regression was causing several severe issues which was rendering the whole platform unusable.

### DIFF
--- a/lib/modules/platform/bitbucket-server/index.ts
+++ b/lib/modules/platform/bitbucket-server/index.ts
@@ -340,9 +340,10 @@ export async function getPr(
     return null;
   }
 
-  const opts: HttpOptions = refreshCache
-    ? { memCache: false }
-    : { cacheProvider: memCacheProvider };
+  const opts: HttpOptions = { memCache: false };
+  if (!refreshCache) {
+    opts.cacheProvider = memCacheProvider;
+  }
 
   const res = await bitbucketServerHttp.getJsonUnchecked<BbsRestPr>(
     `./rest/api/1.0/projects/${config.projectKey}/repos/${config.repositorySlug}/pull-requests/${prNo}`,
--- a/lib/modules/platform/bitbucket-server/index.ts
+++ b/lib/modules/platform/bitbucket-server/index.ts
@@ -340,7 +340,10 @@ export async function getPr(
     return null;
   }
 
+  // Disables memCache (which is enabled by default) to be replaced by
+  // memCacheProvider.
   const opts: HttpOptions = { memCache: false };
+  // TODO: should refresh the cache rather than just ignore it
   if (!refreshCache) {
     opts.cacheProvider = memCacheProvider;
   }
--- a/lib/modules/platform/gerrit/client.ts
+++ b/lib/modules/platform/gerrit/client.ts
@@ -62,7 +62,7 @@ class GerritClient {
     findPRConfig: GerritFindPRConfig,
     refreshCache?: boolean,
   ): Promise<GerritChange[]> {
-    const opts: HttpOptions = {};
+    const opts: HttpOptions = { memCache: false };
     if (!refreshCache) {
       opts.cacheProvider = memCacheProvider;
     }
--- a/lib/modules/platform/gerrit/client.ts
+++ b/lib/modules/platform/gerrit/client.ts
@@ -62,7 +62,10 @@ class GerritClient {
     findPRConfig: GerritFindPRConfig,
     refreshCache?: boolean,
   ): Promise<GerritChange[]> {
+    // Disables memCache (which is enabled by default) to be replaced by
+    // memCacheProvider.
     const opts: HttpOptions = { memCache: false };
+    // TODO: should refresh the cache rather than just ignore it
     if (!refreshCache) {
       opts.cacheProvider = memCacheProvider;
     }


felipecrs:
@viceice, @rarkins, I would recommend bumping the priority of this PR and merging it ASAP. Whoever updates Renovate without this PR would have potentially catastrophic consequences, like Renovate creating PRs limitless (like I had).

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -340,9 +340,10 @@ export async function getPr(
     return null;
   }
 
-  const opts: HttpOptions = refreshCache
-    ? { memCache: false }
-    : { cacheProvider: memCacheProvider };
+  const opts: HttpOptions = { memCache: false };
+  if (!refreshCache) {
+    opts.cacheProvider = memCacheProvider;
+  }
```
rarkins, line 346: Can this be documented? Both before and after, the intent of this is a little confusing. 

Before: if refreshCache then memCache=false, else cacheProvider=memCacheProvider
After: if refreshCache then memCache=false, else memCache=false and cacheProvider=memCacheProvider

felipecrs: Do you mean documented by a comment? I agree, one sec.

viceice (commented):
looks good

lib/modules/platform/bitbucket-server/index.ts
felipecrs: How do you like it?

```ts
    // Replaces memCache with memCacheProvider. Note that memCache needs to be
    // explicitly disabled as it's enabled by default.
    const opts: HttpOptions = { memCache: false };
    // refreshCache means not to read this request from cache or save it to
    // cache, thus, when refreshCache is true, cacheProvider is not set.
    if (!refreshCache) {
      opts.cacheProvider = memCacheProvider;
    }
```

felipecrs: Actually there is one discrepancy with the old (prior to memCacheProvider) behavior and now.

Previously, even if memCache was disabled, the cache would still be saved. I.e. it would disregard existing cache but would update it.

How do we retain the same behavior with memCacheProvider?

Right now when cacheProvider is unset, the response will not be saved to cache. And otherwise it would both be read from cache and saved to cache.

felipecrs: Sorry, I got things wrong. The previous behavior was not as I described.

The option `refreshCache` was poorly implemented though, at least in Gerrit and BitBucket, given they were not actually refreshing cache, but only ignoring it.

felipecrs: Please see https://github.com/renovatebot/renovate/pull/35056/commits/f6ebf374e3690b004007a682ab22a58a3e0c757b

lib/modules/platform/bitbucket-server/index.ts
zharinov, line 1: We're about to drop `memCache` parameter, so it' simpler to not use neither of them

felipecrs: If I didn't get your comment right, can you please elaborate?

zharinov: Yeah, you're right, force the old one disabled via false param, but don't use new implementation at all. When we switch, it won't cache, but it's okay

zharinov: My PR lacks that `memCache: false`

felipecrs: Use only the new implementation when `refreshCache == false`.

> When we switch, it won't cache, but it's okay

I'm confused. Haven't we switched already to `memCacheProvider`? The cache is happening now, through `memCacheProvider`.

zharinov: No, it falls back to old one unless the false is explicitly for older param. We plan to drop it soon, it will dramatically reduce amount of responses we cache in memory.

zharinov: This is why it's simpler to drop the whole refresh flag thing

felipecrs: > This is why it's simpler to drop the whole refresh flag thing

I don't understand. Are you assuming it's useless?

zharinov: Usually we cache to avoid rate limits. If it's not the concern, it's simpler to just drop caching.

felipecrs: Well. It was working for more than a year by now. I disagree and think it should be more carefully thought.

zharinov: Okay, it makes sense too.
Here is more broad fix of the similar bug across many platforms: #35060

felipecrs:
This is now superseded by https://github.com/renovatebot/renovate/pull/35060. But as #35060 may take more time to be reviewed, I'll let maintainers decide whether to close this or not.

rarkins (approved):

viceice (approved):

PR #35057, opened 2025-03-28
feat(azure): impement pr cache

RahulGautamSingh:
## Changes
Cache pull requests using repository cache

## Context
- Last Of: https://github.com/renovatebot/renovate/issues/33290


lib/modules/platform/azure/pr-cache.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+describe('modules/platform/azure/pr-cache', () => {
+  let azureApi: Mocked<typeof import('./azure-got-wrapper')>;
+  let cache = getCache();
+
+  beforeAll(() => {
+    vi.useFakeTimers();
+  });
+
+  beforeEach(async () => {
+    azureApi = await vi.importMock('./azure-got-wrapper');
```
viceice, line 44: do we really need to import it again and again? (perf)

I don't see any reset modules call, so i assume we can statically import it for better test perf.

lib/modules/platform/azure/pr-cache.ts
```diff
@@ -0,0 +1,152 @@
+import type { GitPullRequest } from 'azure-devops-node-api/interfaces/GitInterfaces';
+import { dequal } from 'dequal';
+import { DateTime } from 'luxon';
+import { logger } from '../../../logger';
+import * as memCache from '../../../util/cache/memory';
+import { getCache } from '../../../util/cache/repository';
+// import * as azureApi from './azure-got-wrapper';
```
viceice, line 7:
```suggestion
```

drop it if not needed

lib/modules/platform/azure/pr-cache.ts
```diff
@@ -0,0 +1,152 @@
...
+      if (!cacheOldTime || itemNewTime > cacheOldTime) {
+        updated_at = itemNewTime.toISO();
+      }
+    }
+
+    this.cache.updated_at = updated_at;
+    return needNextPage;
+  }
+
+  private async sync(azureApi: any): Promise<AzurePrCache> {
```
viceice, line 114: No any. To easy to create bugs.

viceice (changes requested):

lib/modules/platform/azure/pr-cache.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    repoCacheReset();
+    cache = getCache();
+  });
+
+  it('fetches cache', async () => {
+    azureApi.gitApi.mockImplementationOnce(
+      () =>
+        ({
+          getPullRequests: vi.fn(() => [pr1]),
+        }) as any,
```
viceice, line 55: use partial helper

lib/modules/platform/azure/pr-cache.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+          updated_at: '2023-01-01T00:00:00.000Z',
+        },
+      },
+    };
+
+    azureApi.gitApi.mockImplementationOnce(
+      () =>
+        ({
+          getPullRequests: vi.fn(() => [pr2, pr1]),
+        }) as any,
```
viceice, line 96: same

lib/modules/platform/azure/pr-cache.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    let needNextPage = true;
+
+    for (const rawItem of rawItems) {
+      const id = rawItem.pullRequestId!;
+
+      const oldItem = oldItems[id];
+      const newItem = getRenovatePRFormat(rawItem);
+
+      // Using current date because azure pullrequest responses do not return updated date
```
viceice, line 94: does it mean we query all PRs on each run? 🤔

lib/modules/platform/azure/pr-cache.ts
```diff
@@ -0,0 +1,152 @@
...
+    return needNextPage;
+  }
+
+  private async sync(azureApi: typeof AzureApi): Promise<AzurePrCache> {
+    logger.debug('Syncing PR list');
+    const azureApiGit = await azureApi.gitApi();
+    let fetchedPrs: GitPullRequest[];
+    let skip = 0;
+    do {
+      fetchedPrs = await azureApiGit.getPullRequests(
```
viceice, line 120: is there no way to include last updated date / order PR return list?

lib/modules/platform/azure/pr-cache.ts
RahulGautamSingh: No. We do one fetch request and compare it with cache. If we found any one PR similar between the fetched and cached we stop fetching PRs.

lib/modules/platform/azure/pr-cache.ts
RahulGautamSingh: Sorry, I do not understand.

viceice: the second parameter is search criteria, it supports at least `creatorId`, so we could add the author filter.

I see there is no order option. they are sorted by pr id descending
https://developercommunity.visualstudio.com/t/get-pull-requests-sorted-by-an-field/795654#T-N801706

lib/modules/platform/azure/pr-cache.ts
viceice: do we need the `updated_at` at all then? looks like it doesn't make sense if the pr's don't have an updated date

😞 we need to load all pr's each time because we never know which of the PR's have changed because they're always returned in pr id order decending.

lib/modules/platform/azure/pr-cache.ts
RahulGautamSingh: I couldn't find the api endpoint to fetch the creatorId from the `username` or something like `/me` which could use the token to fetcht the user details.

PR #35060, opened 2025-03-28
fix: Explicitly disable `memCache` when caching is switched on/off

zharinov:

### DIFF
--- a/lib/modules/platform/gerrit/client.ts
+++ b/lib/modules/platform/gerrit/client.ts
@@ -64,6 +64,7 @@ class GerritClient {
   ): Promise<GerritChange[]> {
     const opts: HttpOptions = {};
     /* v8 ignore start: temporary code */
+    // TODO: should refresh the cache rather than just ignore it
     if (refreshCache) {
       opts.memCache = false;
     } else {
--- a/lib/modules/platform/gerrit/client.ts
+++ b/lib/modules/platform/gerrit/client.ts
@@ -63,7 +63,9 @@ class GerritClient {
     refreshCache?: boolean,
   ): Promise<GerritChange[]> {
     const opts: HttpOptions = {};
-    if (!refreshCache) {
+    if (refreshCache) {
+      opts.memCache = false;
+    } else {
       opts.cacheProvider = memCacheProvider;
     }
 
--- a/lib/modules/platform/gerrit/client.ts
+++ b/lib/modules/platform/gerrit/client.ts
@@ -63,11 +63,13 @@ class GerritClient {
     refreshCache?: boolean,
   ): Promise<GerritChange[]> {
     const opts: HttpOptions = {};
+    /* v8 ignore start: temporary code */
     if (refreshCache) {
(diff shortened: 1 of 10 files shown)


lib/modules/platform/gitlab/index.ts
```diff
@@ -1121,12 +1121,14 @@ export async function getIssue(
   useCache = true,
 ): Promise<Issue | null> {
   try {
-    const opts: GitlabHttpOptions = { memCache: false };
-    /* v8 ignore next: temporary code */
+    const opts: GitlabHttpOptions = {};
+    /* v8 ignore start: temporary code */
     if (useCache) {
       opts.cacheProvider = memCacheProvider;
       delete opts.memCache;
```
felipecrs, line 1128:
```suggestion

```

zharinov: Yeah, it's better to rewrite for clarity

lib/modules/platform/gerrit/client.ts
```diff
@@ -63,9 +63,13 @@ class GerritClient {
     refreshCache?: boolean,
   ): Promise<GerritChange[]> {
     const opts: HttpOptions = {};
-    if (!refreshCache) {
+    /* v8 ignore start: temporary code */
+    if (refreshCache) {
```
felipecrs, line 67: I suggest bringing this comment from #35056, to address the confusion around this option name.

```suggestion
    // TODO: should refresh the cache rather than just ignore it
    if (refreshCache) {
```

felipecrs: The comment applies to bitbucket too, btw.

lib/util/http/github.ts
```diff
@@ -345,6 +345,7 @@ export class GithubHttp extends HttpBase<GithubHttpOptions> {
     const result = await super.requestJsonUnsafe<T>(method, opts);
     if (httpOptions.paginate) {
       delete httpOptions.cacheProvider;
+      httpOptions.memCache = false;
```
viceice, line 348: don't we need it on other http platform implementations too?

viceice (commented):
otherwise LGTM

lib/util/http/github.ts
zharinov: Maybe, I'll double check. Was searching for `cacheProvider` usage only...

felipecrs:
@zharinov you'll need to pull the latest changes as #35056 got merged.

lib/modules/platform/gerrit/client.ts
felipecrs: @zharinov please update this section to match what you did for Gerrit:

https://github.com/renovatebot/renovate/blob/f2fffbf8f5823434f2112bdf7dc2c02c18108b33/lib/modules/platform/bitbucket-server/index.ts#L343-L349

The code was exactly the same, would be nice if they were kept the same.

viceice (approved):

PR #35065, opened 2025-03-29
feat(gitea): find prs from other authors

RahulGautamSingh:
## Changes
- Use the sourceBranch and targetBranch to find PRs from other authors using the [getPullRequestByBaseHead](https://docs.gitea.com/api/1.23/#tag/repository/operation/repoGetPullRequestByBaseHead) endpoint of Gitea API

## Context
- Useful to find reconfigure PRs

- [x] Both unit tests + ran on a real repository https://gitea.com/RahulGautamSingh/labels/pulls/7

### DIFF
--- a/lib/modules/platform/gitea/gitea-helper.ts
+++ b/lib/modules/platform/gitea/gitea-helper.ts
@@ -1,3 +1,4 @@
+import { logger } from '../../../logger';
 import type { BranchStatus } from '../../../types';
 import type { GiteaHttpOptions } from '../../../util/http/gitea';
 import { GiteaHttp } from '../../../util/http/gitea';
@@ -193,7 +194,11 @@ export async function getPRByBranch(
   try {
     const res = await giteaHttp.getJsonUnchecked<PR>(url, options);
     return res.body;
-  } catch {
+  } catch (err) {
+    logger.trace({ err }, 'Error while fetching PR');
+    if (err.statusCode !== 404) {
+      logger.debug({ err }, 'Error while fetching PR');
+    }
     return null;
   }
 }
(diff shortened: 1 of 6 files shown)


lib/modules/platform/gitea/gitea-helper.ts
```diff
@@ -183,6 +183,21 @@ export async function getPR(
...
+  base: string,
+  head: string,
+  options?: GiteaHttpOptions,
+): Promise<PR | null> {
+  const url = `${API_PATH}/repos/${repoPath}/pulls/${base}/${head}`;
+  try {
+    const res = await giteaHttp.getJsonUnchecked<PR>(url, options);
+    return res.body;
+  } catch {
+    return null;
```
viceice, line 197: log error as trace for debug purpose. only ignore 404. if any other errors occur then log debug.

we should also use zod schema for new functions

viceice (commented):
otherwise LGTM

viceice:
API also available on Forgejo 😅

https://code.forgejo.org/api/swagger#/repository/repoGetPullRequestByBaseHead

lib/modules/platform/gitea/gitea-helper.ts
RahulGautamSingh: 1st one done..

For the zod schema, we will need some discussion before about how to handle the `LongCommitSha` branded type. Because, we will first need to create a branded shcema in zod and infer type from it and then use that type everywhere instead if the TS type we use currently, for zod validation to be possible.

RahulGautamSingh: Should I create an issue for that?

viceice: yes please

viceice (approved):

lib/modules/platform/gitea/gitea-helper.ts
RahulGautamSingh: Created a discussion: https://github.com/renovatebot/renovate/discussions/35099

PR #35078, opened 2025-03-30
feat(gomod): support tool directive

maxbrunet:
## Changes

* Extract `tool` directives from `go.mod`
* Extracted `tool` directives are not added to the final dependency list, they are only used to alter matching `indirect` dependencies
* Matching is performed by a simple loop at the moment looking for the longest matching prefix
* Promote matching `indirect` dependencies to `tool` depType and enable them. Non-`indirect` (`require`) modules are not changed in anyway

Reference: https://go.dev/doc/modules/gomod-ref#tool

## Context

Discussion #33867


maxbrunet:
I realized that changing the `depType` might not be adequate and could be considered a breaking change, so I limited the change to be re-enabling the `indirect` deps for tools

viceice (approved):

PR #35080, opened 2025-03-30
feat: allow manager to set `registryStrategy` for a package

### LINKED ISSUE #35029: Allow manager set registryStrategy for a package
### Discussed in https://github.com/renovatebot/renovate/discussions/35023

<sup>Originally posted by **trim21** March 26, 2025</sup>
### Tell us more.

The default strategy of pixi is hunt which is same with our conda data source, but the pixi package manager allow different project use different registryStrategy, users can use hunt or merge if they prefer, and this option is stored in pixl.toml.

So when this option is set, i think renovate should respect it, and therefore pixi manager need to override default strategy of conda source </div>

trim21:
## Context

- closes https://github.com/renovatebot/renovate/issues/35029

### DIFF
--- a/lib/modules/datasource/index.ts
+++ b/lib/modules/datasource/index.ts
@@ -312,7 +312,7 @@ async function fetchReleases(
   );
   let dep: ReleaseResult | null = null;
   const registryStrategy =
-    config.strategy ?? datasource.registryStrategy ?? 'hunt';
+    config.registryStrategy ?? datasource.registryStrategy ?? 'hunt';
   try {
     if (is.nonEmptyArray(registryUrls)) {
       if (registryStrategy === 'first') {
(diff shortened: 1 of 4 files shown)


rarkins (dismissed):

lib/modules/datasource/index.ts
```diff
@@ -311,7 +311,8 @@ async function fetchReleases(
     config.additionalRegistryUrls,
   );
   let dep: ReleaseResult | null = null;
-  const registryStrategy = datasource.registryStrategy ?? 'hunt';
+  const registryStrategy =
+    config.strategy ?? datasource.registryStrategy ?? 'hunt';
```
viceice, line 315: I think we should create a test for this

viceice (commented):
otherwise LGTM

lib/modules/datasource/index.ts
trim21: yeah, where should I put it?

trim21: ok

viceice (approved):

trim21:
need another review from rarkins?

PR #35103, opened 2025-03-31
refactor(util): move user `env` to util folder

### LINKED ISSUE #35036: Set/get user env

After #35035 has been implemented, add these functions:
- setUserEnv(userEnv)
- getUserEnv()

The set function should be called after repository config has been parsed and validated, and should contain the contents of `config.env`. It should be stored in the memory cache so that it is automatically removed at the end of the repo processing. The value of `env` in config should be stripped out after it has been set.

Remove all `userConfiguredEnv` logic which was added in https://github.com/renovatebot/renovate/pull/27028

Instead, the lib/util/exec logic should call `getUserEnv()`. Workers code from that PR also needs updating.

RahulGautamSingh:
## Changes
- To the env.ts in the lib/util folder add these two functions: `setUserEnv` & `getUserEnv`
- Modify affected files, types and tests accordingly

## Context
- Closes: #35036 

### DIFF
--- a/lib/config/types.ts
+++ b/lib/config/types.ts
@@ -219,7 +219,7 @@ export const allowedStatusCheckStrings = [
   'artifactError',
 ] as const;
 export type StatusCheckKey = (typeof allowedStatusCheckStrings)[number];
-
+type UserEnv = Record<string, string>;
 // TODO: Proper typings
 export interface RenovateConfig
   extends LegacyAdminConfig,
@@ -305,7 +305,10 @@ export interface RenovateConfig
   customizeDashboard?: Record<string, string>;
 
   statusCheckNames?: Record<StatusCheckKey, string | null>;
-  env?: Record<string, string>;
+  /**
+   * User configured environment variables that Renovate uses when executing package manager commands
+   */
+  env?: UserEnv;
   logLevelRemap?: LogLevelRemap[];
 
   branchTopic?: string;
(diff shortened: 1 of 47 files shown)


rarkins (commented):
Do you have a repo to test on, e.g. from when `env` was originally added?

lib/config/types.ts
```diff
@@ -306,7 +305,7 @@ export interface RenovateConfig
   customizeDashboard?: Record<string, string>;
 
   statusCheckNames?: Record<StatusCheckKey, string | null>;
-  env?: UserEnv;
+  env?: Record<string, string>;
```
viceice, line 308: why this type change? I like that type alias, you can add tsdoc to it.

viceice (commented):
otherwise LGTM

viceice:
missing coverage

viceice:
it's in cargo rust manager

RahulGautamSingh:
> Do you have a repo to test on, e.g. from when `env` was originally added?

Did some test runs locally, where I used a test repo with `env` config option and some test variables and thet were passed to the `exec` function.

rarkins (approved):

lib/config/types.ts
```diff
@@ -219,8 +219,7 @@ export const allowedStatusCheckStrings = [
   'artifactError',
 ] as const;
 export type StatusCheckKey = (typeof allowedStatusCheckStrings)[number];
-export type UserEnv = Record<string, string>;
-
+type UserEnv = Record<string, string>;
```
viceice, line 222:
```suggestion
export type UserEnv = Record<string, string>;

```

viceice (approved):

lib/config/types.ts
viceice: we can improve this type in a follow up PR

PR #35135, opened 2025-04-02
fix(gomod): fix potential false-positive when matching tools

maxbrunet:
## Changes

Avoid matching a module named `github.com/foo/bar/sub1/sub2/cmd/hell` with a tool named `github.com/foo/bar/sub1/sub2/cmd/hello` by appending trailing slashes before comparing

## Context

Follow up to #35078

### DIFF
--- a/lib/modules/manager/gomod/extract.ts
+++ b/lib/modules/manager/gomod/extract.ts
@@ -7,11 +7,12 @@ function findMatchingModule(
   deps: PackageDependency[],
 ): PackageDependency | undefined {
   let bestMatch: PackageDependency | undefined;
+  const normalizedTool = tool.depName! + '/';
 
   // Find the longest matching prefix for the tool within the dependencies
   for (const dep of deps) {
     if (
-      tool.depName!.startsWith(dep.depName!) &&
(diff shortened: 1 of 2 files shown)


rarkins (approved):

lib/modules/manager/gomod/extract.ts
```diff
@@ -7,11 +7,12 @@ function findMatchingModule(
   deps: PackageDependency[],
 ): PackageDependency | undefined {
   let bestMatch: PackageDependency | undefined;
+  const normalizedTool = tool.depName! + '/';
```
viceice, line 10: use string template

lib/modules/manager/gomod/extract.ts
```diff
@@ -7,11 +7,12 @@ function findMatchingModule(
...
 
   // Find the longest matching prefix for the tool within the dependencies
   for (const dep of deps) {
     if (
-      tool.depName!.startsWith(dep.depName!) &&
+      normalizedTool.startsWith(dep.depName! + '/') &&
```
viceice, line 15: same

viceice (changes requested):

Issue #35141, opened 2025-04-02
Support  `helm-git` powered repositories for helmfile manager


viceice:
### Discussed in https://github.com/renovatebot/renovate/discussions/34928

<div type='discussions-op-text'>

<sup>Originally posted by **MaxymVlasov** March 20, 2025</sup>
### How are you running Renovate?

A Mend.io-hosted app

### Please tell us more about your question or problem

We using `helmfile` and  `helm-git` plugin [offical docs](https://helmfile.readthedocs.io/en/latest/#configuration) for manage ours and 3rd party helm charts.

I not sure is that issue always exist or was introduced in new Renovate versions, as I didn't check Renovate logs for long time. All logs for last 14 days includes that error, and we introduced this specific helm chart only 4 months ago. 
But for sure, current Renovate start failing right after addition of 
```yaml
repositories:
  - name: "gitops-external-cluster"
    url: "git+https://github.com/codefresh-io/csdp-official@add-cluster/helm"
```

![Screenshot from 2025-03-20 18-35-16](https://github.com/user-attachments/assets/19016a9a-d94c-409b-bcc4-4996319df4e3)

Minimum reproducible repository - https://github.com/MaxymVlasov/renovate-helmfile-url-issue/tree/main

### Logs (if relevant)

<details><summary>Logs</summary>

ERROR-only:
```
ERROR: Request Error: cannot parse url
{
  "url": "index.yaml"
  "baseUrl": "git+https://github.com/codefresh-io/csdp-official@add-cluster/helm/"
  "resolvedUrl": "git+https://github.com/codefresh-io/csdp-official@add-cluster/helm/index.yaml"
}
```

DEBUG:

```

DEBUG: Using RE2 regex engine
DEBUG: Parsing configs
DEBUG: Checking for config file in /runner/renovate/job_config.json
DEBUG: Detected config in env RENOVATE_CONFIG
{
  "config": {
    "extends": [
      "mergeConfidence:all-badges"
    ],
    "prFooter": "This PR was generated by [Mend Renovate](https://mend.io/renovate/). View the [repository job log](https://developer.mend.io/{{platform}}/{{repository}}).",
    "redisUrl": "redis://developer-platform-bare-metal-prd.aqffol.ng.0001.use1.cache.amazonaws.com:6379",
    "binarySource": "install",
    "globalExtends": [
      "global:safeEnv"
    ],
    "platformCommit": "enabled",
    "allowedCommands": [
      "^git add --all$",
[...]

PR #35142, opened 2025-04-02
fix(helmfile): skip helm-git repositories

viceice:
## Changes
Currently renovate will fail with an error on http because of an invalid url.
So we skip those repositories for now.

## Context
- #34928
- #35141


secustor (commented):
Why not simply removing the prefix?  

We are doing this in the ansible manager https://github.com/renovatebot/renovate/blob/a87dea425527bea4d3e3da668fedf2b3d465f96b/lib/modules/manager/ansible-galaxy/roles.ts#L59

secustor (approved):

PR #35150, opened 2025-04-02
fix(pep621): make name for uv index optional

nikstuckenbrock:
## Changes

```
name: z.string(),
```

```
name: z.string().optional(),
```

## Context

Regarding to the official `uv` [documentation](https://docs.astral.sh/uv/configuration/indexes/#defining-an-index) a name for a new defined index is optional: `# Optional name for the index.`. Therefor I've modified the zod schema for the uv index accordingly.


rarkins (commented):
This `name` field is used in these three places:
- https://github.com/renovatebot/renovate/blob/f17caf4c1da65d63c359727cd24fcef770d9b8ff/lib/modules/manager/pep621/processors/uv.ts#L42
- https://github.com/renovatebot/renovate/blob/f17caf4c1da65d63c359727cd24fcef770d9b8ff/lib/modules/manager/pep621/processors/uv.ts#L69
-https://github.com/renovatebot/renovate/blob/f17caf4c1da65d63c359727cd24fcef770d9b8ff/lib/modules/manager/pep621/processors/uv.ts#L349-L357

We need to assess the impact on this field being undefined.

viceice (dismissed):

nikstuckenbrock:
Hey @viceice @rarkins,

thanks for the quick response. In my opinion the first two references shouldn't be a problem. For the `getUvIndexCredentials` to work a name for the index is required. Therefor if no name exists I would assume skipping it is the most stable option. Let me know what you think. :)

rarkins:
> thanks for the quick response. In my opinion the first two references shouldn't be a problem. For the `getUvIndexCredentials` to work a name for the index is required. Therefor if no name exists I would assume skipping it is the most stable option. Let me know what you think. :)

How about the other usage? It seems to be logic relating to "is this the default index or not?" so what's the impact if one or both of them are undefined? If both undefined then they _match_ as default.

nikstuckenbrock:
> > thanks for the quick response. In my opinion the first two references shouldn't be a problem. For the `getUvIndexCredentials` to work a name for the index is required. Therefor if no name exists I would assume skipping it is the most stable option. Let me know what you think. :)
> 
> How about the other usage? It seems to be logic relating to "is this the default index or not?" so what's the impact if one or both of them are undefined? If both undefined then they _match_ as default.

Well, in this case all indizes missing a name are considered default and `uv` for example will try to check if it finds the package in one of the registries one after another. In my case I'm using GitLab and auth using a `.netrc` file containing a `CI_JOB_TOKEN` for my pipeline which has access to all the private package registries. But since this is default `uv` behaviour I don't see a problem here?

nikstuckenbrock:
@viceice Can you tell me were to add a test so the codecov step does not fail anymore?

rarkins:
Yes the failure appears to be because of coverage only

nikstuckenbrock:
> Yes the failure appears to be because of coverage only

@rarkins i think we have misunderstood each other here. my question is at what point do i have to add the tests so that the coverage step runs through?

rarkins:
If you look in https://github.com/renovatebot/renovate/pull/35150/files you should see an annotation of the line which is missed

nikstuckenbrock:
@rarkins @viceice Coverage should be good now.

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -344,6 +344,11 @@ async function getUvIndexCredentials(
       continue;
     }
 
+    // If no name is provided for the index, authentication information must be passed through alternative methods
+    if (!name) {
+      continue;
+    }
```
rarkins, line 350: I don't understand this part. I don't see `name` used further, so what prevents its use?

nikstuckenbrock: If you take a look at line 354 in the same file you'll see that the name is refrenced and `.toUpperCase().replace(/[^A-Z0-9]/g, '_')` gets applied to it. If it is `undefined` this will lead to `Cannot read properties of undefined (reading 'toUpperCase')`. Therefor I've added the check.

lib/modules/manager/pep621/schema.ts
```diff
@@ -80,7 +80,7 @@ const UvSchema = z.object({
   index: z
     .array(
       z.object({
-        name: z.string(),
+        name: z.string().optional(),
```
rarkins, line 83: Does this have any impact on the extract part?

nikstuckenbrock: No, because the extraction and uprage process is done by `uv` itself which should be able to handle it's own specification. Therefor handling errors related to this are in the scope of `uv` not `renovate`.

rarkins (approved):

viceice (approved):

PR #35152, opened 2025-04-02
fix(datasource/deb): use only suite as parameter

viceice:
## Changes
`release` is redundant to `suite`, so we should only use the official name which is `suite`.

## Context
- https://github.com/renovatebot/renovate/pull/34902#discussion_r2006314616
- https://manpages.debian.org/testing/apt/sources.list.5.en.html


secustor (changes requested):
This would breaking anyone who is using `release` right now. 

I think we should defer this to v40

viceice:
> This would breaking anyone who is using `release` right now. 
> 
> I think we should defer this to v40

not breaking, `release` is still supported but no longer documented. there are still tests to verify this.

secustor (approved):

PR #35161, opened 2025-04-03
feat(util): getEnv

### LINKED ISSUE #35037: Add getEnv() util and use in place of `process.env`

Following on from #35036

There should be a new function `getEnv()` which returns the combination of `process.env`, `customEnvVariables`, and user `env`.

This should be used in place of existing calls to `process.env` and should fix issues like #33062 (GOPROXY in datasource).

RahulGautamSingh:
## Changes

- Added  function `getEnv()` which returns the combination of process.env, customEnvVariables, and user env
- Replaces most usage of `process.env` with `getEnv()`
- Modified tests as necessary

## Context
- Closes: #35037 


tools/test-shards.ts
```diff
@@ -1,5 +1,6 @@
 import crypto from 'node:crypto';
 import { minimatch } from 'minimatch';
+import { getEnv } from '../lib/util/env';
```
viceice, line 3: no, revert this. this should use real env and isn't part of renovate package / distribution.

tools/docs/github-query-items.ts
```diff
@@ -1,6 +1,7 @@
 import { DateTime } from 'luxon';
 import { z } from 'zod';
 import { logger } from '../../lib/logger';
+import { getEnv } from '../../lib/util/env';
```
viceice, line 4: same here

lib/util/exec/env.ts
RahulGautamSingh, line 1: Should we use combined env in this file too?

lib/workers/global/index.ts
RahulGautamSingh, line 1: Using `getEnv` won't makes much sense here, as the value returned will be same as `process.env` neither repo/file config have been parsed yet. Should I revert the changes?

lib/util/exec/env.ts
RahulGautamSingh, line 1: Should I use `getEnv` here?

lib/workers/repository/init/merge.ts
RahulGautamSingh, line 1: I think we do not need to use combined env here as static repo config is meant to be in the env var and not file/repo config

RahulGautamSingh (commented):
Need instructions for the following files.

lib/util/exec/env.ts
viceice: @rarkins i'm not sure here 🤔

lib/workers/global/index.ts
viceice: 🤔 maybe leave it for easier mocking?

lib/util/exec/env.ts
```diff
@@ -39,6 +39,7 @@ const basicEnvVars = [
   'COREPACK_ROOT',
 ];
 
+// should I use the combined env here as well ?
```
viceice, line 42: I'm not sure, maybe? We need to exclude `process.env` if `exposeAllEnv!==true`

lib/workers/global/index.ts
```diff
@@ -123,12 +124,13 @@ export async function start(): Promise<number> {
   }
 
   let config: AllConfig;
+  const env = getEnv();
```
viceice, line 127: is the global env already configured here?

lib/util/exec/env.ts
rarkins: I think we should use it here as well

lib/workers/global/index.ts
rarkins: The only possible env available at this moment would be from `process.env`

rarkins (changes requested):
Needs deconflicting

viceice (approved):

rarkins (approved):

PR #35177, opened 2025-04-04
fix(util): parse jsonc as jsonc

risu729:
## Changes

Parses `.jsonc` files using `jsonc-parser`, not with JSON5 fallback.

## Context

This avoids warnings: `File contents are invalid JSON but parse using JSON5`, while parsing JSONC files.
- #35176.

### DIFF
--- a/lib/util/common.ts
+++ b/lib/util/common.ts
@@ -1,4 +1,6 @@
 import JSON5 from 'json5';
+import * as JSONC from 'jsonc-parser';
+import type { JsonValue } from 'type-fest';
 import {
   BITBUCKET_API_USING_HOST_TYPES,
   BITBUCKET_SERVER_API_USING_HOST_TYPES,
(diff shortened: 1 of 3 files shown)


lib/util/common.ts
```diff
@@ -108,3 +116,12 @@ export function parseJsonWithFallback(
 
   return parsedJson;
 }
+
+export function parseJsonc(content: string): JsonValue {
```
viceice, line 120: add context parameter

lib/util/common.ts
```diff
@@ -108,3 +116,12 @@ export function parseJsonWithFallback(
...
+  const errors: JSONC.ParseError[] = [];
+  const value = JSONC.parse(content, errors, { allowTrailingComma: true });
+  if (errors.length === 0) {
+    return value;
+  }
+  throw new Error('Invalid JSONC');
```
viceice, line 126: log a warning like the other parse functions and don't throw

viceice (changes requested):

lib/util/common.ts
risu729: Should I return `null` for that case?
I guess other parse functions like `JSON.parse` or `JSON5.parse` throw errors if the content is invalid.

viceice: you're right, I misread the function above. sorry.

viceice (approved):

PR #35180, opened 2025-04-04
feat(submodules): use env vars from getGitEnvironmentVariables 

RGunning:
Replacement PR for #31934 with imports fixed

## Changes

This change uses the existing getGitEnvironmentVariables function during the cloning of git submodules.

This function has been used to include support for updating private git ssh submodules on github.com, but it was not added in the function triggered when the cloneSubmodules config file option is set to true.

see: https://github.com/renovatebot/renovate/pull/22061

## Context

https://github.com/renovatebot/renovate/discussions/30942

Private submodules on the github.com app cannot currently be cloned with the cloneSubmodules option when the .gitmodules file is configured with ssh.

This is because the GitHub App token does not have SSH permissions, it can only auth via HTTP.

My use case is private Go submodules on github.com, for the "postUpdateOptions": ["gomodTidy", "gomodUpdateImportPaths"], options to work, I need my private git submodules cloned.


rarkins (dismissed):

lib/util/git/index.ts
```diff
@@ -358,6 +359,13 @@ export async function cloneSubmodules(
     return;
   }
   submodulesInitizialized = true;
+  const gitSubmoduleAuthEnvironmentVariables = getGitEnvironmentVariables();
+  const gitEnv = {
+    // pass all existing env variables
+    ...process.env,
```
viceice, line 365: @rarkins I would prefer to use `getChildProcessEnv` for security reasons.

viceice: - #35764

viceice: @rarkins use `getEnv` or `getChildEnv` here? 🤔

rarkins: Not sure. `getEnv()` maybe?

viceice: i think `getChildEnv`should be used here:
https://github.com/renovatebot/renovate/blob/bb9faed9c3754c316bea5415a3100df4284ce639/lib/modules/datasource/git-refs/base.ts#L37

RGunning: Updated to use getChildEnv and confirmed that it still works after the change

RGunning:
Also, it appears the previous workaround we were using to get this working with gitSubmodules (https://qiita.com/sijiaoh/items/7d31904a4329d644b05d) stopped working due to https://github.com/renovatebot/renovate/pull/35764, so we are now blocked from using renovate until this PR is merged

rarkins:
Needs a test fix

viceice (commented):
needs test fix

RGunning:
I'm unable to run these tests locally, so can't fix the failing test. Trying to run locally I encounter a whole host of git issues that block the tests running in this branch or on main. Is there some docker environment where I can run the tests?

Screenshot 2025-05-26 at 8 31 15

RGunning:
I'm unable to run these tests locally even in the devContainer. I'm getting a git error. Is there some documentation on how to setup and run these tests?

Screenshot 2025-05-26 at 8 59 11

viceice (commented):
otherwise LGTM

viceice (dismissed):

lib/util/git/index.ts
```diff
@@ -1,10 +1,10 @@
-import URL from 'node:url';
```
viceice: lint issue: import order is wrong

rsahrakorpi-cambri: import order fixed, but tests still fail (but they also fail for me on main, so no idea what is wrong with them)

RGunning: import order fixed, but tests still fail (but they also fail for me on main, so no idea what is wrong with them)

rarkins:
The failing test is in https://github.com/renovatebot/renovate/actions/runs/15993362696/job/45111113890?pr=35180

lib/util/git/index.spec.ts
```diff
@@ -208,6 +211,10 @@ describe('util/git/index', { timeout: 10000 }, () => {
       beforeEach(async () => {
         const repo = Git(base.path);
 
+        modifiedAuth.getGitEnvironmentVariables.mockReturnValue({
+          GIT_ALLOW_PROTOCOL: 'file',
```
RGunning, line 216: This was the minimal git config required for the tests to pass. Otherwise, this error is thrown 
> Error: fatal: not a git repository: ../.git/modules/submodule
fatal: could not reset submodule index

without this `getGitEnvironmentVariables` returns {}, which messes up the tests. This only seems to be required when running the unit tests, I guess, when running the code for real git inherits the required environment, or there are always HOST rules added, which might be sufficient to inform git that this is actually a git repository. I'm not sure on the underlying requirements of git

lib/util/git/index.spec.ts
```diff
@@ -10,24 +10,27 @@ import {
...
 } from '../../constants/error-messages';
 import { newlineRegex, regEx } from '../regex';
+import * as _modifiedAuth from './auth';
 import * as _behindBaseCache from './behind-base-branch-cache';
 import * as _conflictsCache from './conflicts-cache';
 import * as _modifiedCache from './modified-cache';
 import type { FileChange } from './types';
-import * as git from '.';
 import { setNoVerify } from '.';
+import * as git from '.';
```
viceice: please avoid this change, drop the line above this line and prefix that function call with `git.`

lib/util/git/index.spec.ts
```diff
@@ -10,24 +10,27 @@ import {
   UNKNOWN_ERROR,
 } from '../../constants/error-messages';
 import { newlineRegex, regEx } from '../regex';
+import * as _modifiedAuth from './auth';
```
viceice:
```suggestion
import * as _auth from './auth';
```
please no unnecessary renaming

lib/util/git/index.spec.ts
```diff
@@ -10,24 +10,27 @@ import {
...
 vi.mock('./modified-cache');
 vi.mock('timers/promises');
 vi.mock('../cache/repository');
+vi.mock('./auth');
 vi.unmock('.');
 
 const behindBaseCache = vi.mocked(_behindBaseCache);
 const conflictsCache = vi.mocked(_conflictsCache);
 const modifiedCache = vi.mocked(_modifiedCache);
+const modifiedAuth = vi.mocked(_modifiedAuth);
```
viceice:
```suggestion
const auth = vi.mocked(_auth);
```

lib/util/git/index.spec.ts
```diff
@@ -209,6 +212,10 @@ describe('util/git/index', { timeout: 10000 }, () => {
       beforeEach(async () => {
         const repo = Git(base.path);
 
+        modifiedAuth.getGitEnvironmentVariables.mockReturnValue({
```
viceice:
```suggestion
        auth.getGitEnvironmentVariables.mockReturnValue({
```

viceice (commented):
otherwise LGTM

viceice (approved):

RGunning:
@viceice what is the status on this? What is it waiting on to merge? Just says "Merging is blocked" for me

viceice:
> @viceice what is the status on this? What is it waiting on to merge? Just says "Merging is blocked" for me

waiting for another reviewer

viceice:
it's important to note that's a small breaking change, because now the global env vars are trimmed to only allowed values. that's why I want a second pair of 👀

RGunning:
ping @secustor could this get a second review please

secustor (approved):
Theoretically breaking if users inject different credentials for submodule clones via config file, but I see this rather as a fix as it aligns the behaviour.

PR #35181, opened 2025-04-04
feat: allow user to pass git extra config

Djiit:
# Changes

Add a way to pass extra git config to be stacked **after** host rules processing.

I'm not sure it's the "right" way to do so. Any help or feedback appreciated.

# Context

Before this change, any user env var are kept; but renovate-added one are taking precedence and thus "nulling" previous config that would target the same urls.

This is a simple attempt at allowing users to pass a JSON-formatted var to add arbitrary git config that would override Renovate ones.

Our specific use case: we want git operations to target a git CDN. Without this change, Renovate always bypass our config with hostrules-generated ones.

# Documentation (please check one with an [x])

### DIFF
--- a/lib/util/git/auth.spec.ts
+++ b/lib/util/git/auth.spec.ts
@@ -113,11 +113,7 @@ describe('util/git/auth', () => {
       expect(
         getGitAuthenticatedEnvironmentVariables(
           'https://github.com/',
-          {
-            token: 'token1234',
-            hostType: 'github',
-            matchHost: 'github.com',
-          },
+          { token: 'token1234', hostType: 'github', matchHost: 'github.com' },
           { GIT_CONFIG_COUNT: '1' },
         ),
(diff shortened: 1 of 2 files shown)


Djiit:
Hey @rarkins ; while this is not 100% ready, could you give me your thoughts on this when you get a chance? Thank you.

rarkins (commented):
@Djiit sorry, I didn't notice this review request as it was in draft.

It would be preferred to have started this as a Discussion, not a PR, so we can discuss requirements and possible solutions.

Requirement: users want to pass custom git config through to child processes?

Best solution: Renovate should add/append to any existing git config instead of blowing away the original?

As well as a question about this:

> Our specific use case: we want git operations to target a git CDN. Without this change, Renovate always bypass our config with hostrules-generated ones.

Is there something stopping you configuring hostRules for this CDN?

PR #31425, opened 2024-09-16
feat(manager): CNB support for Spring Boot maven plugin

nicolasbender:
## Changes

Adds support for updating Cloud Native Buildpacks in the maven plugin for Spring Boot

## Context

Maven plugin for Spring Boot allows to create an OCI image from a jar or war file using [Cloud Native Buildpacks](https://buildpacks.io/). The plugin offers to provide remote docker resources for builder and run images including custom buildpacks in `pom.xml`. This PR includes a manager to keep these resources updated.

- [Buildpacks](https://docs.spring.io/spring-boot/maven-plugin/build-image.html#build-image.examples.custom-image-builder) in Spring Boot

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

PR #35183, opened 2025-04-04
feat(manager): CNB support for Spring Boot maven plugin

c0d1ngm0nk3y:
## Changes

Adds support for updating Cloud Native Buildpacks in the maven plugin for Spring Boot

## Context

Maven plugin for Spring Boot allows to create an OCI image from a jar or war file using [Cloud Native Buildpacks](https://buildpacks.io/). The plugin offers to provide remote docker resources for builder and run images including custom buildpacks in pom.xml. This PR includes a manager to keep these resources updated.

* [Buildpacks](https://docs.spring.io/spring-boot/maven-plugin/build-image.html#build-image.examples.custom-image-builder) in Spring Boot

This was originally done in a separate manager in #31425, but was requested to be moved to the existing `maven` manager with [this comment](https://github.com/renovatebot/renovate/pull/31425#pullrequestreview-2307077195).


lib/modules/manager/buildpacks/extract.spec.ts
```diff
@@ -103,13 +103,15 @@ id = "not/valid"`,
           datasource: 'buildpacks-registry',
           currentValue: '1.0.0',
           packageName: 'example/foo',
+          depName: 'example/foo',
```
rarkins, line 106: Duplicating packageName into depName is undesirable. Is this change avoidable?

lib/modules/manager/maven/update.ts
```diff
@@ -69,6 +69,20 @@ export function updateAtPosition(
     // TODO: validate newValue (#22198)
     const replacedPart = versionPart.replace(version, newValue!);
     return leftPart + replacedPart + restPart;
+  } else {
```
rarkins, line 72: Add code comment documenting when/why this logic branch would occur

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +78,75 @@ function containsPlaceholder(str: string | null | undefined): boolean {
...
+    return (
+      pluginNode.valueWithPath('groupId')?.trim() ===
+        'org.springframework.boot' &&
+      pluginNode.valueWithPath('artifactId')?.trim() ===
+        'spring-boot-maven-plugin'
+    );
+  });
+  if (pluginNodes.length !== 1) {
+    return null;
+  }
```
rarkins, line 121: If you only handle the case where there's one result, use `.find()` instead of `.filter()`.

rarkins (changes requested):
We'll want to split this into two PRs - refactoring + feature, but can keep it here until it's finalized/greed

lib/modules/manager/buildpacks/extract.spec.ts
mvach: We switched to use depName now.

lib/modules/manager/maven/update.ts
mvach: I guess that is explaining enough ;-).

lib/modules/manager/maven/extract.ts
mvach: Done

lib/modules/manager/buildpacks/extract.spec.ts
rarkins: Why switch? If you have only one, then packageName is preferred.

mvach:
> We'll want to split this into two PRs - refactoring + feature, but can keep it here until it's finalized/greed

Hmmm I'm wondering what you see as a refactoring in this PR ...

rarkins:
> > We'll want to split this into two PRs - refactoring + feature, but can keep it here until it's finalized/greed
> 
> Hmmm I'm wondering what you see as a refactoring in this PR ...

100% of the changes in the buildpack manager and perhaps 25% of the changes in maven

mvach:
> > Hmmm I'm wondering what you see as a refactoring in this PR ...
> 
> 100% of the changes in the buildpack manager and perhaps 25% of the changes in maven

I'm still not 100% convinced that this one should move the named changes out of this PR. They are actually useless without the changes in the for the CNB support. However if that are the rules then let's just follow them ;-).

I created the needed refactoring PR: #35492. As soon that PR is merged I'll cleanup this PR accordingly.

mvach:
After PR: #35492 was merged into main, I updated this PR accordingly.

mvach:
We added the test fixtures since all other tests in the maven manager a using such resources as well. But still the "Check for Undesirable Code" is failing.

rarkins:
Text fixtures are ok - we will override that check

rarkins (dismissed):

lib/config/options/index.ts
```diff
@@ -1148,6 +1148,7 @@ const options: RenovateOptions[] = [
       'helmv3',
       'kubernetes',
       'kustomize',
+      'maven',
```
Churro, line 1151: IMHO this would be misleading, since `registryAliases` in `maven` manager will only work with CNB image references, not with registries that are defined for regular dependencies.

lib/modules/manager/maven/extract.spec.ts
```diff
@@ -274,6 +274,169 @@ describe('modules/manager/maven/extract', () => {
...
+          depName: 'paketobuildpacks/builder-jammy-base',
+          packageName: 'paketobuildpacks/builder-jammy-base',
+          replaceString: 'paketobuildpacks/builder-jammy-base:0.4.316',
+          fileReplacePosition: 1273,
+          registryUrls: [],
+        },
+      ]);
+    });
+
+    it('returns no buildpack dependencies when image tag is missing in spring boot plugin', () => {
```
Churro, line 405: please further reduce this test and the (lengthy) fixture. Extracting `spring-boot-start-parent` here only redundantly tests the `maven` manager but it's unrelated to CNB.

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +84,84 @@ function containsPlaceholder(str: string | null | undefined): boolean {
...
+  const deps: PackageDependency[] = [];
+  nodes.forEach((node) => {
+    const depString = node.val.trim();
+    if (isDockerRef(depString)) {
+      const dep = getDockerDep(
+        depString.replace(DOCKER_PREFIX, ''),
+        true,
+        config.registryAliases,
+      );
+      dep.registryUrls = [];
```
Churro, line 100: why is this needed here? `buildpacks/extract` doesn't set it either?

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +84,84 @@ function containsPlaceholder(str: string | null | undefined): boolean {
...
+      if (dep.currentValue || dep.currentDigest) {
+        deps.push(dep);
+      }
+    } else if (isBuildpackRegistryRef(depString)) {
+      const dep = getBuildpackDep(
+        depString.replace(BUILDPACK_REGISTRY_PREFIX, ''),
+      );
+
+      if (dep && (dep.currentValue || dep.currentDigest)) {
+        dep.registryUrls = [];
```
Churro, line 111: same with registryUrls here

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +84,84 @@ function containsPlaceholder(str: string | null | undefined): boolean {
...
+  const buildpacks: PackageDependency[] = getCNBDependencies(
+    imageNode.childNamed('buildpacks')?.childrenNamed('buildpack') ?? [],
+    config,
+  );
+  deps.push(...builder, ...runImage, ...buildpacks);
+
+  if (!deps.length) {
+    return null;
+  }
+  return deps;
```
Churro, line 162:
```suggestion
  return deps.length ? deps : null;
```

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +84,84 @@ function containsPlaceholder(str: string | null | undefined): boolean {
...
+      dep.fileReplacePosition = node.position;
+      if (dep.currentValue || dep.currentDigest) {
+        deps.push(dep);
+      }
+    } else if (isBuildpackRegistryRef(depString)) {
+      const dep = getBuildpackDep(
+        depString.replace(BUILDPACK_REGISTRY_PREFIX, ''),
+      );
+
+      if (dep && (dep.currentValue || dep.currentDigest)) {
```
Churro, line 110: Why is the `(dep.currentValue || dep.currentDigest)` check needed here? `currentValue` is always set and `currentDigest` always undefined, see https://github.com/renovatebot/renovate/blob/1484c54019bbfee1812543033b4154ee4889c61c/lib/modules/manager/buildpacks/extract.ts#L156-L158

Churro (changes requested):

lib/modules/manager/maven/extract.ts
mvach: if dep.registryUrls is not set, extract fails here: https://github.com/renovatebot/renovate/blob/c1d59fde1d831c6c8833c4d9774731fd7a9c55eb/lib/modules/manager/maven/extract.ts#L585

lib/modules/manager/maven/extract.ts
mvach: if dep.registryUrls is not set, extract fails here: https://github.com/renovatebot/renovate/blob/c1d59fde1d831c6c8833c4d9774731fd7a9c55eb/lib/modules/manager/maven/extract.ts#L585

Churro: Wouldn't that be wrong anyway or why would one want to add maven registries that were extracted from `settings.xml` to extrated docker image references? 

Perhaps it makes sense to make this safer and prevent adding additional registries by changing it to:
```ts
`if (dep.registryUrls) {
  dep.registryUrls.unshift(...additionalRegistryUrls);
}
```
...untested, so please verify

lib/config/options/index.ts
nicolasbender: Is this just for information purposes or a whitelist-type structure? Because for the latter we would need it to be included here since we rely on registry aliases for CNBs in the spring-boot plugin

lib/modules/manager/maven/extract.ts
nicolasbender: `currentValue` is not always set, there are cases [like this](https://github.com/renovatebot/renovate/commit/1d3733fb878389a043b0398ffe2e1173b34dde96)

lib/config/options/index.ts
Churro: Only information purposes. How about keeping it here but adding a note to https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/maven/readme.md that `registyAliases` works only for container image references?

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +84,84 @@ function containsPlaceholder(str: string | null | undefined): boolean {
   return !!str && regEx(/\${[^}]*?}/).test(str);
 }
 
+function getCNBDependencies(
+  nodes: XmlElement[],
+  config: ExtractConfig,
+): PackageDependency[] {
+  const deps: PackageDependency[] = [];
+  nodes.forEach((node) => {
```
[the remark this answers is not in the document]
mvach: Done

lib/modules/manager/maven/update.ts
```diff
@@ -69,6 +69,22 @@ export function updateAtPosition(
...
```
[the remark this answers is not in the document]
mvach: Done

lib/modules/manager/maven/update.spec.ts
```diff
@@ -141,6 +142,65 @@ describe('modules/manager/maven/update', () => {
 
       expect(res).toBe(simpleContent);
     });
+
+    it('should update a cloud native buildpack version', () => {
```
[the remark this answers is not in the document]
mvach: Done

lib/config/options/index.ts
mvach: Done - Added a readme comment.

lib/modules/manager/maven/extract.spec.ts
mvach: Done

lib/modules/manager/maven/extract.ts
mvach: Done

lib/modules/manager/maven/extract.ts
mvach: Done, we had to touch this as well: https://github.com/renovatebot/renovate/blob/d02cc7028d9fc4bffd8e49db2dfd1a62a4942b1b/lib/modules/manager/maven/extract.ts#L553

lib/modules/manager/maven/extract.ts
mvach: Done

lib/modules/manager/maven/update.ts
```diff
@@ -69,6 +69,25 @@ export function updateAtPosition(
...
+    upgrade.datasource === 'docker' ||
+    upgrade.datasource === 'buildpacks-registry'
+  ) {
+    // In contrast to maven dependencies, cloud native buildpacks are not contained in specific version tags.
+    // Instead they are contained in the value of the buildpack tag and we have to update it differently.
+    let replacedPart = version;
+    if (currentValue) {
+      replacedPart = version.replace(currentValue, newValue!);
+    }
+    if (upgrade.currentDigest) {
```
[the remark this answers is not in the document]
mvach: We could easily add a check for that corner case but asking ourself how that could happen?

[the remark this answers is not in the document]
mvach: I added the requested check but still would be interested how that case could happen.

nicolasbender:
Is there anything missing or something left we should address?

mvach:
Hi @c0d1ngm0nk3y ,
sadly I cannot resolve the conversations. Since you created the PR, can you please try to take care?

Churro (changes requested):
Running renovate with the code from this PR on the `full_cnb.pom.xml` fixture (extracted into a test repo) fails due to `dep.depName` being undefined somewhere:
> TypeError: Cannot read properties of undefined (reading 'replace')
https://github.com/renovatebot/renovate/blob/29de7b9afba7f4661f2c7cb903749464326da082/lib/modules/manager/maven/extract.ts#L206

... likely during the buildpack image extraction, which only sets `packageName` and not `depName`. Probably not the only place because `getDep()` for docker refs could also return `dep.depName = undefined` and a `skip-reason`, so this needs to be handled accordingly.

The newly added tests in `extracts.spec.ts` only call `extractPackage()` but the mentioned exception is thrown in the outer `extractAllPackageFiles()` -> `resolveParents(packages)`. 

Could you please ensure that this works end-to-end and then provide a small repo that demonstrates this?

mvach:
> Running renovate with the code from this PR on the `full_cnb.pom.xml` fixture (extracted into a test repo) fails due to `dep.depName` being undefined somewhere:
> 
> > TypeError: Cannot read properties of undefined (reading 'replace')
> > https://github.com/renovatebot/renovate/blob/29de7b9afba7f4661f2c7cb903749464326da082/lib/modules/manager/maven/extract.ts#L206
> 
> ... likely during the buildpack image extraction, which only sets `packageName` and not `depName`. Probably not the only place because `getDep()` for docker refs could also return `dep.depName = undefined` and a `skip-reason`, so this needs to be handled accordingly.
> 
> The newly added tests in `extracts.spec.ts` only call `extractPackage()` but the mentioned exception is thrown in the outer `extractAllPackageFiles()` -> `resolveParents(packages)`.
> 
> Could you please ensure that this works end-to-end and then provide a small repo that demonstrates this?

You're right, build pack dependencies do not have a version necessarily. I added the required check for that now.

... from my perspective this is just another indicator that this should be better implemented in its own manager. But we have been asked to implement it here ... which has pros and cons: https://github.com/renovatebot/renovate/pull/31425#pullrequestreview-2307077195

With regards to `packageName` vs `depName`, we've been advised to use `packageName`: https://github.com/renovatebot/renovate/pull/35183/files/1984bd8d3ab72409ffc5b4a1ced0905bd0ecc606#r2053997324

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +84,79 @@ function containsPlaceholder(str: string | null | undefined): boolean {
...
 }
 
+function getCNBDependencies(
+  nodes: XmlElement[],
+  config: ExtractConfig,
+): PackageDependency[] {
+  const deps: PackageDependency[] = [];
+  for (const node of nodes) {
+    const depString = node.val.trim();
+    if (isDockerRef(depString)) {
```
Churro, line 94: Checking for `isDockerRef(depString)` and later `isBuildpackRegistryRef(depString)` means that a directive like `<builder>${some.prop.with.the.image}</builder>` will not work with renovate while it does with `mvn`. 

This is a limitation of the current implementation (which is okay IMHO) but a note in the readme would be good.

Churro (changes requested):
Tested `full_cnb.pom` in a real repo and seems to work now.

What is likely not intended and should still be addressed is to explicitly set `dep.versioning` to `dockerVersioning`. At the moment, renovate falls back to `maven` as it is the default for maven manager.

```json
{
  "depName": "paketobuildpacks/run-noble-full",
  "packageName": "paketobuildpacks/run-noble-full",
  "currentValue": "0.0.28",
  "datasource": "docker",
   ...
  "versioning": "maven",
   ...
}
```

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +85,81 @@ function containsPlaceholder(str: string | null | undefined): boolean {
...
+        deps.push(dep);
+      }
+    } else if (isBuildpackRegistryRef(depString)) {
+      const dep = getBuildpackDep(
+        depString.replace(BUILDPACK_REGISTRY_PREFIX, ''),
+      );
+
+      if (dep?.currentValue) {
+        dep.fileReplacePosition = node.position;
+        deps.push(dep);
```
Churro, line 112: What about versioning for buildpacks? It would implicitly use  `maven` here, while the buildpack manager relies on the default `semver`. Probably not an issue in most cases but an inconsistency that should be easy to fix. If `maven` versioning is expected for buildpack refs here, then of course no need to change.

Otherwise LGTM

lib/modules/manager/maven/__fixtures__/basic_cnb.pom.xml
```diff
@@ -0,0 +1,47 @@
...
+				<configuration>
+					<image>
+						<name>repository.local/demo-spring-boot</name>
+						<builder>paketobuildpacks/builder-jammy-base:0.4.316</builder>
+					</image>
+				</configuration>
+			</plugin>
+		</plugins>
+	</build>
+</project>
```
viceice: add trailing newlines to fixtures

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +85,81 @@ function containsPlaceholder(str: string | null | undefined): boolean {
...
+  const deps: PackageDependency[] = [];
+  for (const node of nodes) {
+    const depString = node.val.trim();
+    if (isDockerRef(depString)) {
+      const dep = getDockerDep(
+        depString.replace(DOCKER_PREFIX, ''),
+        true,
+        config.registryAliases,
+      );
+      dep.versioning = dockerVersioningId;
```
viceice: drop this, docker datasource will automatically default to docker versioning if not overridden by user
```suggestion
```

lib/modules/manager/maven/extract.ts
viceice: no, it's using docker versioning, see above. maven manager doesn't set any default versioning, it's set by datasource

lib/modules/manager/maven/index.ts
```diff
@@ -17,4 +18,4 @@ export const defaultConfig = {
   versioning: mavenVersioning.id,
```
viceice: needs to be removed, otherwise it will be set for docker images, see other comments. versioning is set by datasource anyways

```suggestion
```

https://github.com/renovatebot/renovate/blob/eb9c35f0bfc665e73041c5decbda689aa3a6caca/lib/modules/datasource/maven/index.ts#L75

lib/modules/manager/maven/extract.ts
mvach: As far as we got it @Churro requested that: https://github.com/renovatebot/renovate/pull/35183#pullrequestreview-2894208976

mvach: Ah, I got your point. By removing it from [index.ts#L1](https://github.com/renovatebot/renovate/blob/f6224a94c82dcda3608db941cc248134a9338c3a/lib/modules/manager/maven/index.ts#L17), we don't need to set it here.

Churro (approved):

rarkins (dismissed):

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +84,80 @@ function containsPlaceholder(str: string | null | undefined): boolean {
...
+  if (!pluginNode) {
+    return null;
+  }
+
+  const deps: PackageDependency[] = [];
+  const imageNode = pluginNode.childNamed('configuration')?.childNamed('image');
+  if (!imageNode) {
+    return null;
+  }
+  const builder: PackageDependency[] = getCNBDependencies(
```
viceice:
```suggestion
  const builder = getCNBDependencies(
```
not needed here

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +84,80 @@ function containsPlaceholder(str: string | null | undefined): boolean {
...
+    imageNode.childrenNamed('builder'),
+    config,
+  );
+  const runImage: PackageDependency[] = getCNBDependencies(
```
viceice:
```suggestion
  const runImage = getCNBDependencies(
```

lib/modules/manager/maven/extract.ts
```diff
@@ -76,6 +84,80 @@ function containsPlaceholder(str: string | null | undefined): boolean {
...
+    imageNode.childrenNamed('runImage'),
+    config,
+  );
+  const buildpacks: PackageDependency[] = getCNBDependencies(
```
viceice:
```suggestion
  const buildpacks = getCNBDependencies(
```

viceice (approved):

PR #35188, opened 2025-04-04
feat(gitlab): add reviewers from an existing approval rule in GitLab

nvnoorloos:
## Changes

Inspired by the `bbUseDefaultReviewers` option in Renovate (https://docs.renovatebot.com/configuration-options/#bbusedefaultreviewers) implemented in the `createPr` methods for both BitBucket platforms i've introduced an implementation to add 'default reviewers' for GitLab based on the concept of Approval Rules (https://docs.gitlab.com/user/project/merge_requests/approvals/rules/).

In GitLab approval rules can be added (with a custom name) containing users like this:
image

This feature introduces a property called `gitLabReviewersFromApprovalRule` where the name of an approval rule can be configured, Renovate will query the users from this approval rule, and adds them to the reviewers list of the Merge Request in GitLab.

Run with Debug log enabled:
image


.python-version
rarkins, line 1: Revert

lib/modules/platform/gitlab/index.ts
```diff
@@ -757,6 +758,51 @@ export async function createPr({
   }
   const description = sanitize(rawDescription);
   logger.debug(`Creating Merge Request: ${title}`);
+
+  let reviewerIds;
+  if (platformPrOptions?.gitLabReviewersFromApprovalRule) {
+    logger.info(
```
rarkins, line 764:
```suggestion
    logger.debug(
```

lib/modules/platform/gitlab/index.ts
```diff
@@ -757,6 +758,51 @@ export async function createPr({
...
+      {
+        gitLabReviewersFromApprovalRule:
+          platformPrOptions.gitLabReviewersFromApprovalRule,
+      },
+      'Fetching reviewers from GitLab approval rule',
+    );
+    try {
+      const approvalRules = await gitlabApi.getJsonUnchecked<
```
rarkins, line 772: Please use a zod schema

lib/modules/platform/gitlab/index.ts
```diff
@@ -757,6 +758,51 @@ export async function createPr({
...
+          { ruleName: platformPrOptions.gitLabReviewersFromApprovalRule },
+          'Found matching approval rule',
+        );
+        for (const approver of matchingApprovalRule.eligible_approvers) {
+          approvalRuleReviewerIds.add(approver.id);
+        }
+        reviewerIds = Array.from(approvalRuleReviewerIds);
+      } else {
+        logger.debug(
+          { ruleName: platformPrOptions.gitLabReviewersFromApprovalRule },
```
rarkins, line 793:
```suggestion
```

No need to repeat this if it was in the previous log message

lib/modules/platform/gitlab/index.ts
```diff
@@ -757,6 +758,51 @@ export async function createPr({
...
   const description = sanitize(rawDescription);
   logger.debug(`Creating Merge Request: ${title}`);
+
+  let reviewerIds;
+  if (platformPrOptions?.gitLabReviewersFromApprovalRule) {
+    logger.info(
+      {
+        gitLabReviewersFromApprovalRule:
+          platformPrOptions.gitLabReviewersFromApprovalRule,
+      },
```
rarkins, line 768: Inline this string into the log message itself

rarkins (changes requested):

nvnoorloos:
Sorry, the force-push occurred due to a 'Update with rebase' action

lib/modules/platform/gitlab/index.ts
```diff
@@ -757,6 +757,44 @@ export async function createPr({
   }
   const description = sanitize(rawDescription);
   logger.debug(`Creating Merge Request: ${title}`);
+
+  let reviewerIds;
```
viceice, line 761:
```suggestion
  let reviewerIds: number[];
```

nvnoorloos:
[@rarkins](https://gitlab.com/rarkins) I've processed all your requested changes, so a new review would be welcome. Thank you!

docs/usage/configuration-options.md
```diff
@@ -1655,6 +1655,27 @@ Under the hood, it creates a MR-level approval rule where `approvals_required` i
...
+For example, to add reviewers from the "default reviewers" approval rule to all dependency updates:
+
+```json
+{
+  "gitLabReviewersFromApprovalRule": "default reviewers"
+}
+```
+
+<!-- prettier-ignore -->
+!!! warning
```
rarkins, line 1676: Make this a note, not warning

lib/modules/platform/gitlab/index.ts
```diff
@@ -757,6 +757,44 @@ export async function createPr({
...
+        GitLabApprovalRules,
+      );
+
+      const approvalRuleReviewerIds = new Set<number>();
+      const matchingApprovalRule = approvalRules.body.find(
+        (rule) =>
+          rule.name === platformPrOptions.gitLabReviewersFromApprovalRule,
+      );
+
+      if (matchingApprovalRule?.eligible_approvers?.length) {
```
rarkins, line 778: Split this into 2 if's so that we can have separate log messages if:
- rule not found
- rule has no approvers

lib/modules/platform/gitlab/index.ts
```diff
@@ -757,6 +757,44 @@ export async function createPr({
...
+        for (const approver of matchingApprovalRule.eligible_approvers) {
+          approvalRuleReviewerIds.add(approver.id);
+        }
+        reviewerIds = Array.from(approvalRuleReviewerIds);
+      } else {
+        logger.debug(
+          'No matching approval rule found or rule has no eligible approvers',
+        );
+      }
+      logger.debug(
```
rarkins, line 789: Move this inside the "if" above so that it's only logged if the array is non empty

lib/modules/platform/gitlab/index.ts
```diff
@@ -768,6 +806,7 @@ export async function createPr({
         description,
         labels: (labels ?? []).join(','),
         squash: config.squash,
+        reviewer_ids: reviewerIds,
```
rarkins, line 809: How does this conflict with renovate's existing reviewers capability?

rarkins (changes requested):

lib/modules/platform/gitlab/index.ts
nvnoorloos: This implementation is based on the BitBucket variant enabled with `bbUseDefaultReviewers` and is implemented when creating the initial Merge Request (where the array is by default already empty).

The other Renovate options for reviewers are processed in a later stage and will add/remove reviewers accordingly.

nvnoorloos:
@rarkins i've processed all your requested changes and added some extra tests, codecov should be fine now.

lib/modules/platform/gitlab/index.ts
```diff
@@ -757,6 +757,49 @@ export async function createPr({
   }
   const description = sanitize(rawDescription);
   logger.debug(`Creating Merge Request: ${title}`);
+
+  let reviewerIds: number[] = [];
+  if (platformPrOptions?.gitLabReviewersFromApprovalRule) {
+    logger.debug(
```
viceice, line 763: test that those new log messages are called for better verification

viceice (changes requested):

lib/modules/platform/gitlab/index.ts
nvnoorloos: @viceice Thanks! Added asserts for all the log statements

rarkins (commented):
Normally reviewers are added to PRs/MRs using the `addReviewers()` function. Adding them here as part of `createPr()` changes the semantics and needs further thought to make sure it doesn't introduce problems

nvnoorloos:
For BitBucket the implementation is the same. Therefore i've used the same setup in GitLab. But i do understand your concerns.

nvnoorloos:
> One _good_ reason to do it as part of createPr() is that it means we'll query this API at most once per MR. If we put it with our usual reviewer logic, and supported updating existing MRs, it would mean that we'd need to keep querying the API each Renovate run.

What do you want to do with this Pull Request giving your concerns?

viceice (dismissed):

rarkins (changes requested):
Please note in the docs that this setting is only applied at MR creation time. 

We also need to work out what happens if other reviewers are set, either via direct config or via code owners. Maybe this feature isn't compatible and we need to note that too?

nvnoorloos:
> Please note in the docs that this setting is only applied at MR creation time.
> 
> We also need to work out what happens if other reviewers are set, either via direct config or via code owners. Maybe this feature isn't compatible and we need to note that too?

I've updated the docs to make clear that it's only on MR creation, i've also added this to the BitBucket option that has the same behaviour.

Other reviewers are set (also when using the `reviewersFromCodeOwners` option) after creation time, not during.

In the UI GitLab helps the users by adding /assign_reviewer xxxx by default in the description field when creating an MR, but when creating MR's via the API there are no reviewers set. If you enable the "Require code owner approval" in the Branch Rules it will still not add reviewers to the reviewer list on creation, it will inject a separate approval section within MR itself.

In short: GitLab seems to create MR's without any reviewers by default, it's something that has to be set explicitly during or after creation, i've not found any options that will implicitly add reviewers to the list on creation therefore this feature does not conflict with other options.

nvnoorloos:
> > Please note in the docs that this setting is only applied at MR creation time.
> > We also need to work out what happens if other reviewers are set, either via direct config or via code owners. Maybe this feature isn't compatible and we need to note that too?
> 
> I've updated the docs to make clear that it's only on MR creation, i've also added this to the BitBucket option that has the same behaviour.
> 
> Other reviewers are set (also when using the `reviewersFromCodeOwners` option) after creation time, not during.
> 
> In the UI GitLab helps the users by adding /assign_reviewer xxxx by default in the description field when creating an MR, but when creating MR's via the API there are no reviewers set. If you enable the "Require code owner approval" in the Branch Rules it will still not add reviewers to the reviewer list on creation, it will inject a separate approval section within MR itself.
> 
> In short: GitLab seems to create MR's without any reviewers by default, it's something that has to be set explicitly during or after creation, i've not found any options that will implicitly add reviewers to the list on creation therefore this feature does not conflict with other options.

@rarkins is the above explanation enough to consider implementing this feature? 

If not, i will close this and we are going to find another way to achieve the functionality within our company because the MR's created by Renovate are not easy to find within our GitLab platform for our end-users (_they do not show up in the "to review" lists etc. because the MR's are not assigned to any team members. A static assignment is no option because we service a lot of projects that each have their own owners/developers_)

docs/usage/configuration-options.md
```diff
@@ -353,7 +353,7 @@ Configuring this to `true` means that Renovate will mark all PR Tasks as complet
 
 ## bbUseDefaultReviewers
 
-Configuring this to `true` means that Renovate will detect and apply the default reviewers rules to PRs (Bitbucket only).
+Configuring this to `true` means that Renovate will detect and apply the default reviewers rules to PRs (Bitbucket only). Renovate will add the reviewers only at time of PR creation and not update them after that.
```
rarkins:
```suggestion
Configuring this to `true` means that Renovate will detect and apply the default reviewers rules to PRs (Bitbucket only).
Renovate will add the reviewers only at time of PR creation and not update them after that.
```

docs/usage/configuration-options.md
```diff
@@ -1655,6 +1655,27 @@ Under the hood, it creates a MR-level approval rule where `approvals_required` i
 This option works only when `automerge=true` and either `automergeType=pr` or `automergeType=branch`.
 Also, approval rules overriding should not be [prevented in GitLab settings](https://docs.gitlab.com/ee/user/project/merge_requests/approvals/settings.html#prevent-editing-approval-rules-in-merge-requests).
 
+## gitLabReviewersFromApprovalRule
+
+There are cases where you want to automatically add reviewers to Renovate PRs based on existing GitLab approval rules (<https://docs.gitlab.com/user/project/merge_requests/approvals/rules/>)
```
rarkins:
```suggestion
There are cases where you want to automatically add reviewers to Renovate PRs based on existing GitLab [approval rules](https://docs.gitlab.com/user/project/merge_requests/approvals/rules/).
```

docs/usage/configuration-options.md
```diff
@@ -1655,6 +1655,27 @@ Under the hood, it creates a MR-level approval rule where `approvals_required` i
...
+Renovate can use a GitLab's project approval rule to determine who should review dependency updates.
+This is particularly useful for teams that already have well-defined approval processes in GitLab.
+
+To enable this functionality, use the gitLabReviewersFromApprovalRule option to specify the name of the GitLab approval rule to use for selecting reviewers.
```
rarkins:
```suggestion
To enable this functionality, use the `gitLabReviewersFromApprovalRule` option to specify the name of the GitLab approval rule to use for selecting reviewers.
```

docs/usage/configuration-options.md
```diff
@@ -1655,6 +1655,27 @@ Under the hood, it creates a MR-level approval rule where `approvals_required` i
...
+Renovate will then add all reviewers from that rule to the created merge requests. It will add the reviewers only at time of MR creation and does not update them after that.
```
rarkins:
```suggestion
Renovate will then add all reviewers from that rule to merge requests it creates.
It will add the reviewers only at time of MR creation and does not update them after that.
```

docs/usage/configuration-options.md
```diff
@@ -1655,6 +1655,27 @@ Under the hood, it creates a MR-level approval rule where `approvals_required` i
...
+
+```json
+{
+  "gitLabReviewersFromApprovalRule": "default reviewers"
+}
+```
+
+<!-- prettier-ignore -->
+!!! note
+GitLab approval rules are only available for the 'Premium' or 'Ultimate' GitLab Tier
```
rarkins: I'm pretty sure this needs indentation

lib/modules/platform/gitlab/index.ts
```diff
@@ -757,6 +757,49 @@ export async function createPr({
...
+      `Fetching reviewers from GitLab approval rule: ${platformPrOptions.gitLabReviewersFromApprovalRule}`,
+    );
+    try {
+      const approvalRules = await gitlabApi.getJson(
```
rarkins, line 755: Memory cache this so that it only needs to query gitlab once per job

rarkins (changes requested):

docs/usage/configuration-options.md
nvnoorloos: Fixed

lib/modules/platform/gitlab/index.ts
nvnoorloos: Implemented

rarkins (commented):
I still have a general concern here - until now, reviewers were always added (using multiple logic) after the MR creation. This way of defining reviewers is done during MR creation and I suspect may conflict with the others. Unless it's impossible to do it like the others, the logic should be moved

nvnoorloos:
> I still have a general concern here - until now, reviewers were always added (using multiple logic) after the MR creation. This way of defining reviewers is done during MR creation and I suspect may conflict with the others. Unless it's impossible to do it like the others, the logic should be moved

@rarkins i've opened a discussion about this implementation to prevent spending more time on a solution that will not be accepted:

https://github.com/renovatebot/renovate/discussions/36892

PR #35234, opened 2025-04-07
fix(gomod): disable automaticly toolchain switching

trim21:
## Changes

go>=1.21 will switch toolchain be default, the initial toolchain may download a new version of go and use it, we should disable toolchain switching, it make toolchain picking meanless and make `config.constraints.go` not working as expected in some case.

Add `GOTOOLCHAIN=local` to disable it.

self-hosted app can still add such env to override it.

### DIFF
--- a/lib/modules/manager/gomod/artifacts.ts
+++ b/lib/modules/manager/gomod/artifacts.ts
@@ -200,6 +200,9 @@ export async function updateArtifacts({
     const execOptions: ExecOptions = {
       cwdFile: goModFileName,
       extraEnv: {
+        // disable go toolchain switch by default,
+        // but user can override it in self-hosted app.
+        GOTOOLCHAIN: process.env.GOTOOLCHAIN ?? 'local',
         GOPATH: await ensureCacheDir('go'),
         GOPROXY: process.env.GOPROXY,
         GOPRIVATE: process.env.GOPRIVATE,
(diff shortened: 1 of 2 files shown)


lib/modules/manager/gomod/artifacts.ts
```diff
@@ -200,6 +200,9 @@ export async function updateArtifacts({
     const execOptions: ExecOptions = {
       cwdFile: goModFileName,
       extraEnv: {
+        // disable go toolchain switch by default,
+        // but user can override it in self-hosted app.
```
astellingwerf, line 204: What's the specific reason why it can only be overridden in self-hosted app instances?

trim21: I think there is no readon that mend hosted APP can't also override it? 

I just make it configurable and a default value.

casaqori: Shouldn't we set it to local only if there is a  `config.constraints.go` set?
Imho, renovate shall behave like the go defaults out of the box. So, allowing version bumps of go when not constrained.

trim21: now I think it should be `local` in all case:

https://github.com/renovatebot/renovate/pull/34779#discussion_r2040204012

otherwise we can't avoid touch go directive when it doesn't include patch version...

trim21:
this should be merged with 34779

PR #35265, opened 2025-04-08
fix(workers/auto-replace): correctly handle prefix replacements

jonasrutishauser:
## Changes

This PR fixes at least one bug in auto replacement if the replacement adds a prefix (e.g. `example.org/foo` -> `docker.example.org/foo` or `1.2.3` -> `v1.2.3`).

Additionally the implementation is faster as it uses less iterations.

## Context

With the current implementation if the replacement (packageName or version) only adds a prefix, the replacement does not work and returns an error.
The additional tests are showing the issue.

### DIFF
--- a/lib/workers/repository/update/branch/auto-replace.spec.ts
+++ b/lib/workers/repository/update/branch/auto-replace.spec.ts
@@ -757,6 +757,54 @@ describe('workers/repository/update/branch/auto-replace', () => {
       );
     });
 
+    it('updates with helm value image/repository prefix replacement', async () => {
+      const yml = codeBlock`
+        parser:
+          image:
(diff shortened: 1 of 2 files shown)


rarkins:
Can you test this on a real repo to confirm it definitely works?

jonasrutishauser:
I add a small change to ensure that there is no endless loop (it shouldn't be possible before, but this way we can really be sure).

I tested it against a test repo here: https://github.com/jonasrutishauser/renovate-35265/pull/1

I found a bug in the helmfile manager, which doesn't support replacements with oci repositories. I will open a new bug for this.

viceice (approved):

secustor (approved):

PR #35296, opened 2025-04-09
fix(manager/helmfile): Allow replacements of oci:// charts

jonasrutishauser:
## Changes

Allow replacements of oci chart references in helmfile.

## Context

At the moment, it is not possible to do a replacement of a oci chart refence in a helmfile.yaml like the following:
```yaml
releases:
  name: example
  version: 1.2.3
  chart: oci://ghcr.io/example/oci-repo/url-example
```
This issue was found while testing #35265 against a repository.

- [x] Both unit tests + ran on a real repository (see https://github.com/jonasrutishauser/renovate-35265-helmfile/pull/2)

### DIFF
--- a/lib/modules/manager/helmfile/extract.spec.ts
+++ b/lib/modules/manager/helmfile/extract.spec.ts
@@ -339,7 +339,7 @@ describe('modules/manager/helmfile/extract', () => {
           stable: 'https://charts.helm.sh/stable',
         },
       });
-      expect(result).toMatchObject({
+      expect(result).toStrictEqual({
         datasource: 'helm',
         deps: [
           {
@@ -347,6 +347,7 @@ describe('modules/manager/helmfile/extract', () => {
             depName: 'example',
             datasource: 'docker',
             packageName: 'ghcr.io/example/oci-repo/example',
+            registryUrls: ['ghcr.io/example/oci-repo'],
           },
           {
             currentValue: '3.3.0',
@@ -355,9 +356,10 @@ describe('modules/manager/helmfile/extract', () => {
           },
           {
             currentValue: '0.4.2',
-            depName: 'url-example',
+            depName: 'ghcr.io/example/oci-repo/url-example',
             datasource: 'docker',
             packageName: 'ghcr.io/example/oci-repo/url-example',
+            registryUrls: [],
           },
         ],
       });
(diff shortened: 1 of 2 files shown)


lib/modules/manager/helmfile/extract.spec.ts
```diff
@@ -355,9 +356,10 @@ describe('modules/manager/helmfile/extract', () => {
           },
           {
             currentValue: '0.4.2',
-            depName: 'url-example',
+            depName: 'ghcr.io/example/oci-repo/url-example',
```
rarkins, line 359: Is this change really necessary? The full name is already in `packageName`

rarkins (dismissed):

lib/modules/manager/helmfile/extract.spec.ts
jonasrutishauser: In the case of a package name replacement it is needed as the `auto-replace` uses the `depName` for the replacement (which is correct as in the case of the indirection via the repository, only the `depName` can be replaced).

lib/modules/manager/helmfile/extract.spec.ts
```diff
@@ -339,14 +339,15 @@ describe('modules/manager/helmfile/extract', () => {
...
-      expect(result).toMatchObject({
+      expect(result).toStrictEqual({
         datasource: 'helm',
         deps: [
           {
             currentValue: '0.1.0',
             depName: 'example',
             datasource: 'docker',
             packageName: 'ghcr.io/example/oci-repo/example',
+            registryUrls: ['ghcr.io/example/oci-repo'],
```
viceice, line 350: registry urls should not be set for docker / oci charts

lib/modules/manager/helmfile/extract.ts
```diff
@@ -109,20 +107,22 @@ export async function extractPackageFile(
...
       if (kustomizationsKeysUsed(dep)) {
         needKustomize = true;
       }
 
       if (isOCIRegistry(dep.chart)) {
         res.datasource = DockerDatasource.id;
-        res.packageName = `${repoName}/${depName}`;
-      } else if (registryData[repoName]?.oci) {
+        res.packageName = depName;
+      } else if (registryData[repoName!]?.oci) {
```
viceice, line 123: how do you ensure `repoName` isn't `undefined` or `null` here?

viceice (approved):

PR #35320, opened 2025-04-11
fix(git): add workaround for git fetch not pulling automerged commit

felipecrs:
## Changes

This adds a workaround for a rare case where `git fetch` could not return a commit that was just auto-merged.

## Context

In my environment, Gerrit hosts the git repositories behind a replication system with several mirrors.

This means that, sometimes, running `git pull` may not return the actual latest commit (because it may not yet be replicated to the mirror that is chosen by `git pull`).

This happens super frequently when Renovate auto-merges a change and then starts a new branch for handling another dependency. Humans barely notice it, but Renovate merges and `git fetch`es in a few seconds.

This is an environment-specific issue, but nevertheless a real case that may happen for others under similar environments.

If Renovate do `git fetch` and the auto-merged commit doesn't come, it will keep going as if nothing happened, and this will cause some unexpected issues like changes being pushed behind the actual base branch.

Another unexpected behavior is that Renovate merges commits which are behind base branch, while it should not.

~I have been running this workaround successfully in my own Renovate instance for months.~ EDIT: I re-implemented it, and I need to test it better now.

However, my intention is not to merge this PR (yet), but to open a discussion about it:

- What do you think about this?
  - ~I dislike the part that it requires one extra API call to Gerrit~. EDIT: solved.
  - Injecting an environment variable to share information from one function to another seems super hacky, but I couldn't find a better way.
- How likely is it that something like this can be merged? Provided I make all necessary changes, of course.


felipecrs:
I reimplemented this PR:

- It's no longer platform-dependent, it should work for all platforms
- No longer adds an extra API call

PR #35321, opened 2025-04-11
feat(presets): add `mvdan-sh` replacement `sh-syntax`

JounQin:
> [!note]
> it's not a 1:1 replacement

## Changes

As title, real case replacement: https://github.com/un-ts/prettier/pull/430

## Context

https://github.com/mvdan/sh/issues/1145

### DIFF
--- a/lib/data/replacements.json
+++ b/lib/data/replacements.json
@@ -1,6 +1,11 @@
 {
   "$schema": "../../tools/schemas/replacements-schema.json",
   "all": {
+    "description": "Apply all replacement rules.",
+    "extends": ["replacements:drop-in", "replacements:breaking"],
+    "ignoreDeps": []
+  },
+  "drop-in": {
     "description": "Apply crowd-sourced package replacement rules.",
     "extends": [
       "replacements:airbnb-prop-types-to-prop-types-tools",
@@ -49,6 +54,11 @@
       "replacements:xmldom-to-scoped",
       "replacements:zap"
     ],
+    "ignoreDeps": []
+  },
+  "breaking": {
+    "description": "Apply breaking package replacement rules.",
+    "extends": ["replacements:mvdan-sh-to-sh-syntax"],
     "ignoreDeps": []
   },
   "airbnb-prop-types-to-prop-types-tools": {
@@ -761,6 +771,16 @@
         "matchPackageNames": ["middie"],
         "replacementName": "@fastify/middie",
         "replacementVersion": "8.0.0"
+      }
+    ]
+  },
+  "mvdan-sh-to-sh-syntax": {
+    "description": "`mvdan-sh` uses unmaintained `gopherjs` so it is very outdated, and now `mvdan-sh` is deprecated and `sh-syntax` is recommended by the original author.",
+    "packageRules": [
+      {
+        "matchDatasources": ["npm"],
+        "matchPackageNames": ["mvdan-sh"],
+        "replacementName": "sh-syntax"
       }
     ]
   },
--- a/tools/schemas/replacements-schema.json
+++ b/tools/schemas/replacements-schema.json
@@ -16,6 +16,36 @@
         }
       },
       "required": ["description", "extends"]
+    },
+    "breaking": {
+      "type": "object",
+      "properties": {
+        "description": { "type": "string" },
+        "extends": {
+          "type": "array",
+          "items": { "type": "string" }
+        },
+        "ignoreDeps": {
+          "type": "array",
+          "items": { "type": "string" }
+        }
+      },
+      "required": ["description", "extends"]
+    },
+    "drop-in": {
+      "type": "object",
+      "properties": {
+        "description": { "type": "string" },
+        "extends": {
+          "type": "array",
+          "items": { "type": "string" }
+        },
+        "ignoreDeps": {
+          "type": "array",
+          "items": { "type": "string" }
+        }
+      },
+      "required": ["description", "extends"]
     }
   },
   "patternProperties": {
--- a/tools/schemas/schema.ts
+++ b/tools/schemas/schema.ts
@@ -41,7 +41,7 @@ const RuleSetSchema = z.object({
     ),
 });
 
-const AllSchema = z.object({
+const ExtendsSchema = z.object({
   description: z.string(),
   extends: z.array(z.string()),
   ignoreDeps: z.array(z.string()).optional(),
@@ -50,7 +50,9 @@ const AllSchema = z.object({
 export const ReplacementsSchema = z
   .object({
     $schema: z.string(),
-    all: AllSchema,
+    all: ExtendsSchema,
+    'drop-in': ExtendsSchema,
+    breaking: ExtendsSchema,
   })
   .catchall(RuleSetSchema);
 


rarkins:
We normally only approve drop-in replacements so need to think about the implications of this. 

Without this change, users would still get a deprecation warning and see the deprecation message

JounQin:
Indeed, make your decision freely.

rarkins:
@viceice @secustor perhaps we could have a different type of workaround group - not included with config:recommended - which is for these type of valid replacements which aren't drop-in?

viceice:
> @viceice @secustor perhaps we could have a different type of workaround group - not included with config:recommended - which is for these type of valid replacements which aren't drop-in?

makes sense 👍

secustor:
Having it as opt in and make it clear that these are community suggestions then this makes sense to me. 

I'm thinking of something like `replacements: community-breaking`.

rarkins:
So we already have `workarounds:all`: https://github.com/renovatebot/renovate/blob/5091b14184a38d239de2619b7acd02dc0eade3da/lib/config/presets/internal/workarounds.ts#L6

Perhaps we should have:
- `workarounds:drop-in` (contains all of the ones so far)
- `workarounds:breaking` (the one in this PR will be the first)
- `workarounds:all` can extend both of those
- `config:recommended` can extend `workarounds:drop-in` instead of `workarounds:all`

lib/data/replacements.json
```diff
@@ -751,6 +751,15 @@
       }
     ]
   },
+  "mvdan-sh": {
+    "description": "`mvdan-sh` uses unmaintained `gopherjs` so itself is very outdated",
```
astellingwerf, line 755:
```suggestion
    "description": "`mvdan-sh` uses unmaintained `gopherjs` so it is very outdated.",
```

rarkins:
So this PR adds a workaround but not adds it to "workarounds:all"?

JounQin:
> So this PR adds a workaround but not adds it to "workarounds:all"?

@rarkins Oh, sorry, I'm not familiar with the code base, where should I change the location>

lib/data/replacements.json
```diff
@@ -751,6 +751,15 @@
       }
     ]
   },
+  "mvdan-sh": {
+    "description": "`mvdan-sh` uses unmaintained `gopherjs` so it is very outdated.",
+    "packageRules": [
+      {
+        "matchDatasources": ["npm"],
```
astellingwerf, line 758: This is much too wide of a predicate. We need a `"matchPackageNames"` condition on it as well.

JounQin: Isn't the key name as the original package name? Sorry I'm not familiar with this code base yet.

astellingwerf: No, you need to specify the original package names that you want to replace explicitly.

The key name convention appears to be "<orignal name>-to-<new name>", so in this case "mvdan-sh-to-sh-syntax".

JounQin: @astellingwerf I did a rebase and addressed that!

JounQin:
What need I do to make it merged?

rarkins:
See https://github.com/renovatebot/renovate/pull/35321#issuecomment-2798811417

JounQin:
@rarkins Are my latest changes correct? https://github.com/renovatebot/renovate/pull/35321/files

lib/config/presets/internal/workarounds.ts
```diff
@@ -45,6 +24,16 @@ export const presets: Record<string, Preset> = {
       },
     ],
   },
+  breaking: {
```
rarkins, line 27: This should be similar to the "drop-in" approach. It shouldn't have its own packageRules

lib/config/presets/internal/workarounds.ts
```diff
@@ -260,6 +268,17 @@ export const presets: Record<string, Preset> = {
       },
     ],
   },
+  replacementsMapping: {
```
astellingwerf, line 271: I'd expect this name to at least contain "mvdan-sh".

JounQin: So the rule name vs rules is 1-1? I was expecting to add more rules into the list.

lib/data/replacements.json
```diff
@@ -764,6 +764,16 @@
...
+  "mvdan-sh-to-sh-syntax": {
+    "description": "`mvdan-sh` uses unmaintained `gopherjs` so it is very outdated, and now `mvdan-sh` is deprecated and `sh-syntax` is recommended by the original author.",
+    "packageRules": [
+      {
+        "matchDatasources": ["npm"],
+        "matchPackageNames": ["mvdan-sh"],
+        "replacementName": "sh-syntax"
+      }
+    ]
+  },
```
astellingwerf, line 776: We'd get rid of this change entirely, won't we?

lib/config/presets/internal/workarounds.ts
astellingwerf: The "breaking" and "drop-in" presets are on the same level. In my understanding they would both have an `extends` array with the full set of presets that are either breaking or drop in, all named to generally indicate what they're doing.

lib/data/replacements.json
astellingwerf: You'd put the package rule in lib/config/presets/internal/workarounds.ts lines 276-278.

JounQin: I don't know they're same thing previously. The typing is not great at `workaround.ts`.

rarkins (changes requested):
I'm sorry, I think I may have mixed things up earlier. This should change `replacements:all` and not `workarounds:all`, if this is a replacement

lib/config/presets/internal/workarounds.ts
astellingwerf: I'd suggest "mvdan-sh-to-sh-syntax"

JounQin:
@rarkins Just updated the zod schema.

JounQin:
Friendly ping @@rarkins

lib/data/replacements.json
```diff
@@ -48,8 +53,11 @@
       "replacements:vsts-task-lib-to-azure-pipelines-task-lib",
       "replacements:xmldom-to-scoped",
       "replacements:zap"
-    ],
-    "ignoreDeps": []
```
rarkins, line 52: Not sure this should be removed

JounQin: Should I add it back? I don't know what does this property mean.

rarkins: Yes, avoid removing things if you don't understand!

JounQin: @rarkins Done!

JounQin:
Friendly ping @viceice and @secustor

lib/data/replacements.json
```diff
@@ -764,6 +774,16 @@
       }
     ]
   },
+  "mvdan-sh-to-sh-syntax": {
+    "description": "`mvdan-sh` uses unmaintained `gopherjs` so it is very outdated, and now `mvdan-sh` is deprecated and `sh-syntax` is recommended by the original author.",
+    "packageRules": [
+      {
+        "matchDatasources": ["npm"],
+        "matchPackageNames": ["mvdan-sh"],
+        "replacementName": "sh-syntax"
```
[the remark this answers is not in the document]
viceice: yes, sounds good

viceice (commented):
I think this won't work as is and needs further settings.

I'm also not sure if the breaking replacements should be enabled by default 🤔.

They at least need the mentioned body notes.

JounQin:
Nope.

PR #35322, opened 2025-04-11
fix: also increase commits per run counter in dry-run mode

felipecrs:
## Changes

This changes dry-run mode to also increase and therefore respect `prCommitsLimitPerRun` limit.

## Context

I use Renovate in dry-run to test changes to real projects, and noticed it wasn't respecting the `prCommitsLimitPerRun` limit.


viceice (commented):
I'm not sure if we want this

felipecrs:
It is completely fine if you say no. I don't have any strong opinion either.

rarkins:
Dry run mode is a global setting though, and by definition should never do any commits, so I'm not sure if this does any harm?

felipecrs:
You are right. This PR is a no-op.

I created this PR from a wrong assumption that this would cause Renovate to stop early when the limit is reached, but this is a wrong assumption from my part.

Sorry for the confusion.

PR #35335, opened 2025-04-11
feat: Support multi-version python-version files

### LINKED ISSUE #31850: Support `.python-version` with multiple versions for `pyenv` manager
### Discussed in https://github.com/renovatebot/renovate/discussions/31837

<sup>Originally posted by **mkniewallner** October  7, 2024</sup>
### Tell us more.

Renovate [implements `pyenv` manager](https://github.com/renovatebot/renovate/tree/main/lib/modules/manager/pyenv) to update Python version defined in a `.python-version` file. Per [`pyenv` documentation](https://github.com/pyenv/pyenv/blob/d201daeb3366cae98cca88654b4c52018175ded5/README.md?plain=1#L155-L164), it is also possible to set multiple Python versions in a single `.python-version` file, separated by newlines:
> You can also specify multiple versions in a `.python-version` file by hand,
separated by newlines. Lines starting with a `#` are ignored.

Would it make sense for Renovate to support that, by extracting all possible Python versions defined in the file? Currently, [the content of the file is trimmed](https://github.com/renovatebot/renovate/blob/03a85316b20c8eb21dad2e8a0d14cfe2b739e43d/lib/modules/manager/pyenv/extract.ts#L8), assuming that only one version can be specified.

If it makes sense, I'm willing to implement support for it.</div>

carstencodes:
## Changes

If a `.python-version` file contains multiple
lines, each line, that contains a valid version
string is returned.
A version is here an arbitrary string, that
does not start with a hash-symbol.
A line may contain comments starting with hashes.
In this case, the hash symbol and the trailing
characters will be ignored.
Empty lines will also be ignored.

The position of the version within the file must
now be taken into account.

If a version has major, minor or patch version
parts, the specific bump version property is
set.

## Context

This is especially useful, when environments like
tox shall test multiple python versions at once.

For more details refer to #31850.

Closes #31850

### DIFF
--- a/lib/modules/manager/pyenv/extract.spec.ts
+++ b/lib/modules/manager/pyenv/extract.spec.ts
@@ -10,6 +10,8 @@ describe('modules/manager/pyenv/extract', () => {
           commitMessageTopic: 'Python',
           currentValue: '3.7.1',
           datasource: 'docker',
+          fileReplacePosition: 0,
+          bumpVersion: 'patch',
         },
       ]);
     });
@@ -22,6 +24,8 @@ describe('modules/manager/pyenv/extract', () => {
           commitMessageTopic: 'Python',
           currentValue: '3.8',
           datasource: 'docker',
+          fileReplacePosition: 0,
+          bumpVersion: 'minor',
         },
       ]);
     });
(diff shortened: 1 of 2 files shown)


rarkins (changes requested):
`bumpVersion` is not correct to use here. Let's go back to #31850 to discuss requirements and implementation. In the meantime I will mark this as a draft PT

carstencodes:
Sure, no problem.

PR #35347, opened 2025-04-12
feat(preset): dont split params between quotes

elbandi:
## Changes

Don't split preset params inside quote.

The split regexp come from here: https://stackoverflow.com/questions/73121226/javascript-regex-split-a-string-by-commas-but-ignore-commas-within-double-quote
And the remove quote replace from here: https://stackoverflow.com/questions/26156292/trim-specific-character-from-a-string

## Context

we have preset args with comma:

    "defaultRegistryUrlTemplate": "https://www.somesite.com/query?code={{arg0}}"

We call this:  `mypreset("foo,bar")` so the defaultRegistryUrlTemplate will be: "https://www.somesite.com/query?code=foo,bar"

### DIFF
--- a/lib/config/presets/parse.ts
+++ b/lib/config/presets/parse.ts
@@ -45,7 +45,8 @@ export function parsePreset(input: string): ParsedPreset {
   if (str.includes('(')) {
     params = str
       .slice(str.indexOf('(') + 1, -1)
-      .split(',')
+      .split(/,\s*(?=(?:(?:[^"]*"){2})*[^"]*$)/)
+      .map((elem) => elem.replace(/^\"+|\"+$/g, ''))
       .map((elem) => elem.trim());
     str = str.slice(0, str.indexOf('('));
   }
--- a/lib/config/presets/parse.spec.ts
+++ b/lib/config/presets/parse.spec.ts
@@ -332,6 +332,18 @@ describe('config/presets/parse', () => {
       });
     });
 
+    it('returns scope with repo and params with quote and default', () => {
+      expect(
+        parsePreset('@somescope/somepackagename("param1, param2", param3)'),
+      ).toEqual({
+        repo: '@somescope/somepackagename',
+        params: ['param1, param2', 'param3'],
+        presetName: 'default',
+        presetPath: undefined,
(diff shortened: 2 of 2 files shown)


lib/config/presets/parse.ts
```diff
@@ -45,7 +45,8 @@ export function parsePreset(input: string): ParsedPreset {
   if (str.includes('(')) {
     params = str
       .slice(str.indexOf('(') + 1, -1)
-      .split(',')
+      .split(/,\s*(?=(?:(?:[^"]*"){2})*[^"]*$)/)
```
viceice, line 48: this can cause harm because of backtracking, so a no-go.

viceice (changes requested):

lib/config/presets/parse.ts
viceice: @zharinov Idea? Would like to avoid lookahead /loockbehind because of possible backtracking. This can cause harm because it's user input.

Elbandi:
Another idea: use a different character for no-split commas, split the params, and replace specila char to comma (and no need 
 quote):
preset(foo,bar§baz) -> `.split().map(v=> v.replace("§",","))` -> ["foo", "bar,baz"] and the special char was the '§'.

Second idea: escape the comma, but still need a special char to avoid regexp lookahead /loockbehind:
preset(foo,bar\\,baz) -> `replace("\\,", "§").split().map(v=> v.replace("§",","))` ->["foo", "bar,baz"]

Or use a better regexp! (my knowledge is less for that... )

RahulGautamSingh:
I think it would be simpler to do the extraction of quoted and non-quoted params one by one rather than using a single complex regex.

for eg.

```js
params = getParams(str);

fn  getParams(str){
let params = []
const newStr = str.slice(str.indexOf('(') + 1, -1)

const quotedMatches = [...input.matchAll(/"([^"]*)"/g)].map(m => m[0]);
params = params.concat(quotedMatches);

// logic to remove quoted matches from the newStr
...

const remainingParams = str.trim().split(",").map(m => m.trim())
params = params.concat(remainingParams);

return params;
}
```

PR #35349, opened 2025-04-13
fix(git): optimize `syncGit`

felipecrs:
## Changes

Optimizes the `syncGit` function:

1. Avoid contacting remote in at least 2 situations, but maybe even 3:
   1. By removing redundant `git pull` (`git fetch` && `git reset --hard origin/main` is enough)
   2. By combining `git remote prune origin` into the `git fetch` command through the `--prune` flag
   3. By avoiding an unnecessary `getDefaultBranch` call before the `git fetch` command
2. By avoiding an unnecessary `resetToLocal` call before the `git fetch` command

## Context

While working in #35320 (that PR is completely WIP by the way, non mergeable), I noticed that the `syncGit` function was doing some unnecessary work, then I thought it would be a good idea to optimize it.


lib/util/git/index.ts
```diff
@@ -250,8 +250,8 @@ export async function initRepo(args: StorageConfig): Promise<void> {
 async function resetToBranch(branchName: string): Promise<void> {
   logger.debug(`resetToBranch(${branchName})`);
   await git.raw(['reset', '--hard']);
-  await gitRetry(() => git.checkout(branchName));
-  await git.raw(['reset', '--hard', 'origin/' + branchName]);
+  await git.checkout(branchName);
```
viceice, line 253: wrong assumption. git contacts remote if blob isn't yet downloaded. we clone with filter `blob:none` for perf

viceice (changes requested):

lib/util/git/index.ts
felipecrs: I'd like to try reproducing that, but for now I'll just revert this piece, it doesn't make much difference anyway.

viceice: please update PR description accordingly

viceice (dismissed):

viceice:
changed to `fix` prefix to get a release. @rarkins maybe we should configure `perf` to do a patch release?

felipecrs:
BTW I do that in my projects. Example:

https://github.com/vscode-shellcheck/vscode-shellcheck/blob/49f499f82ae8a300587cd7874ca4e1697d7b7eaf/common.release.config.mjs#L10-L13

lib/util/git/index.ts
felipecrs: I had done this:

image

But I removed it now.

lib/util/git/index.ts
```diff
@@ -249,9 +249,9 @@ export async function initRepo(args: StorageConfig): Promise<void> {
 
 async function resetToBranch(branchName: string): Promise<void> {
   logger.debug(`resetToBranch(${branchName})`);
-  await git.raw(['reset', '--hard']);
+  await git.raw(['reset', '--hard', 'HEAD']);
```
rarkins, line 252: I believe this change is functionally identical to the original. I'm ok with changing it, but have you identified any reason for this change?

felipecrs: No, that's correct. It's identical. The reason why I added `HEAD` is to make it explicit and the purpose clearer.

I can revert if you want.

rarkins: I think let's revert if we think it's working as is

felipecrs, line 252:
```suggestion
  await git.raw(['reset', '--hard']);
```

lib/util/git/index.ts
```diff
@@ -251,7 +251,7 @@ async function resetToBranch(branchName: string): Promise<void> {
   logger.debug(`resetToBranch(${branchName})`);
   await git.raw(['reset', '--hard']);
   await gitRetry(() => git.checkout(branchName));
-  await git.raw(['reset', '--hard', 'origin/' + branchName]);
+  await git.raw(['reset', '--hard', `origin/${branchName}`]);
```
felipecrs, line 254:
```suggestion
  await git.raw(['reset', '--hard', `origin/${branchName}`]);
```

felipecrs:
```suggestion
  await git.raw(['reset', '--hard', 'origin/' + branchName]);
```

lib/util/git/index.ts
felipecrs: Done.

rarkins (approved):

PR #35350, opened 2025-04-13
feat(config): variables

### LINKED ISSUE #35219: Non-masked "variables" object to complement "secrets"

Create a new configuration option called "variables" which behaves the same way as `secrets` except that its values are _not_ masked in logs. It should be available in templates like this: `{{ variables.FOO }}`.

It is OK - but not recommended - for variables to have the same name as secrets.

RahulGautamSingh:
## Changes
- Implement  self-hosted config option variables
- Introduce a `util/interpolator.ts` to house the repeated logic between `secrets` and `variables`

## Context
- Closes: #35219 

- [x] Both unit tests + ran on a real repository https://github.com/Rahul-renovate-testing/secrets-variables/pulls


viceice (commented):
would be nice to have the refactor in separate PR, otherwise lgtm

rarkins (changes requested):
Create refactor PR first

viceice (commented):
failed checks

rarkins (commented):
What was the reason for keeping variables and secrets "apply" function separate?

viceice (dismissed):

RahulGautamSingh:
> What was the reason for keeping variables and secrets "apply" function separate?

No reason :)

rarkins:
Is there any place where applying secrets and applying variables _aren't_ done at the same time? If there's never any separation of them, these functions should be combined.

lib/config/secrets.ts
```diff
@@ -41,11 +64,55 @@ export function applySecretsToConfig(
...
+  variables = config.variables,
+  deleteVariables = true,
+): RenovateConfig {
+  return replaceInterpolatedValuesInObject(
+    config,
+    variables ?? {},
+    options.variables,
+    deleteVariables,
+  );
+}
```
viceice: do we still need this public? i think we can drop those wrapper now and only use the single function below?

lib/config/secrets.ts
```diff
@@ -9,27 +9,50 @@ import { addSecretForSanitizing } from '../util/sanitize';
...
 }
 
+function validateConfigSecrets(config: AllConfig): void {
+  validateNestedInterpolatedValues(config, 'secrets');
+}
+
+function validateConfigVariables(config: AllConfig): void {
+  validateNestedInterpolatedValues(config, 'variables');
+}
+
```
viceice: can't we just inline this? i don't see a benefit for this small wrappers

lib/config/secrets.ts
```diff
@@ -41,11 +64,55 @@ export function applySecretsToConfig(
...
+  variables = config.variables,
+  deleteVariables = true,
+): RenovateConfig {
+  return replaceInterpolatedValuesInObject(
+    config,
+    variables ?? {},
+    options.variables,
+    deleteVariables,
+  );
+}
```
viceice: i would also inline these functions for simpler code

RahulGautamSingh:
Needs some attention 🤒

lib/config/secrets.ts
```diff
@@ -9,43 +9,85 @@ import { addSecretForSanitizing } from '../util/sanitize';
 import type { AllConfig, RenovateConfig } from './types';
 
 const secretNamePattern = '[A-Za-z][A-Za-z0-9_]*';
-
 const secretNameRegex = regEx(`^${secretNamePattern}$`);
 const secretTemplateRegex = regEx(`{{ secrets\\.(${secretNamePattern}) }}`);
+const variableNamePattern = '[A-Za-z][A-Za-z0-9_]*';
+
+const variableNameRegex = regEx(`^${variableNamePattern}$`);
```
viceice: can't we use the secrets pattern because they're the same?

viceice (commented):
otherwise LGTM

viceice (approved):

rarkins (approved):

PR #35374, opened 2025-04-14
feat(yaml): Improve YAML parser on Jinja2 templates

MGSousa:
## Changes

- Added new tests that should fail at the moment, since the YAML Parser is not fixed, at first
image

- Removing the regex pattern will allow parsing this example file, without replacing almost the whole file, in the following situations:

```yaml
networks: 
  {{ custom_network }}: 
    external: true

services:
  service_xpto: 
    hostname: {{ custom_value }}
    image: "{{ image }}:{{ version }}"
    deploy:
      labels: 
        label: {{ custom_value }}
        {{ prefix }}image:{{ custom_value }}
        ...
```

## Context

The workaround is to use double quotes, which does not make sense since Jinja does not need quotes to parse a template correctly.

Added new tests to cover the observed issue.

Without the regex pattern, the `docker-compose` manager works, and all images are correctly discovered and updated without any issues.

 - https://github.com/renovatebot/renovate/discussions/32967

### DIFF
--- a/lib/util/yaml.spec.ts
+++ b/lib/util/yaml.spec.ts
@@ -207,12 +207,12 @@ describe('util/yaml', () => {
         {
           myObject: {
             aString: null,
-            anotherString: 'value'
+            anotherString: 'value',
           },
         },
         {
           foo: null,
-          bar: 'value:v2'
+          bar: 'value:v2',
         },
       ]);
     });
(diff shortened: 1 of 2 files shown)


MGSousa:
The first commit with two new tests were intended to fail as expected: https://github.com/renovatebot/renovate/actions/runs/14453217026/job/40533023397?pr=35374

viceice (approved):

PR #35389, opened 2025-04-15
fix(manager/pixi): do not pick pixi version based on lockfile

trim21:
## Changes

stop pick pixi version based on pixi.lock version

## Context

pixi may add new feature to manifest and it still support loading old lock file version, so it's impossible to get a valid pixi version range based on lock file, we should stop pick pixi version based on lock file verison.


viceice (commented):
I don't like this. 

what's the current lock file version?

in other manager we use it to only limit older lockfile versions to use older tool versions to not silently update the lockfile.

why this behavior is wrong here?

trim21:
Yeah, I don't like it either.

There are currently two problems here:

1. We should not add the latest lock file version here, since `pixi` is still adding new features, this is resolvable by removing the latest version of lock file from mapping.
2. Pixi can still install projects using the old version, and at the same time, the project may contain new features in the manifest, this is not resolvable, because when we a old version of pixi, it can't generate lock file for it.

viceice (approved):

PR #35398, opened 2025-04-15
fix: constraints type in json schema

trim21:
## Changes

constraints in user config should be `Record<string, string>`, this avoid this error:

image

before:
```json
    "constraints": {
      "description": "Configuration object to define language or manager version constraints.",
      "type": "object",
      "default": {},
      "$ref": "#"
    },
```

after:
```json
    "constraints": {
      "description": "Configuration object to define language or manager version constraints.",
      "type": "object",
      "default": {},
      "additionalProperties": {
        "type": "string"
      }
    },
```

### DIFF
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -2439,7 +2439,11 @@ const options: RenovateOptions[] = [
       'pep621',
       'pipenv',
       'poetry',
+      'pixi',
     ],
+    additionalProperties: {
+      type: 'string',
+    },
   },
   {
     name: 'hostRules',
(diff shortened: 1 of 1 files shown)


lib/config/options/index.ts
```diff
@@ -2439,7 +2439,12 @@ const options: RenovateOptions[] = [
       'pep621',
       'pipenv',
       'poetry',
+      'pixi',
```
viceice, line 2442: we should add all supported manager in a follow up PR

viceice (dismissed):

viceice (approved):

PR #35412, opened 2025-04-16
feat: captureGroups can be used in autoReplaceStringTemplate

oemel09:
## Changes

This change makes the values of the capture groups used in the `matchStrings` array available to later be used in the `autoReplaceStringTemplate` for regex managers.

This was not yet the case and there was already a discussion opened about this feature.

## Context

See discussion
 - https://github.com/renovatebot/renovate/discussions/30475


viceice (changes requested):
it's not enough and needs tests. it's also not agreed

oemel09:
I wasn't aware that it has to be agreed on before submitting a PR. Does it make sense to add tests for this or do you not want this to be implemented?

PR #35427, opened 2025-04-17
fix(repositoryCache): skip loading empty cache files

Gabriel-Ladzaretti:
## Changes
Skips loading empty repository cache files.

## Context
whenever a new repo is processed, its cache file is empty. this, in turn, causes Renovate to raise a warning in the dependency dashboard due to the cache file failing schema validation, which is expected for newly added scanned repos.

### DIFF
--- a/lib/util/cache/repository/impl/base.ts
+++ b/lib/util/cache/repository/impl/base.ts
@@ -58,6 +58,11 @@ export abstract class RepoCacheBase implements RepoCache {
         return;
       }
 
+      if (!is.nonEmptyString(oldCache)) {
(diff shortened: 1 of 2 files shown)


rarkins (approved):

lib/util/cache/repository/impl/base.ts
```diff
@@ -58,6 +58,11 @@ export abstract class RepoCacheBase implements RepoCache {
         return;
       }
 
+      if (!is.nonEmptyString(oldCache)) {
```
viceice, line 61: combine with previous condition

PR #35479, opened 2025-04-22
feat(cache/repository): implement Redis client

querti:
- implement redis client for repo caching
- add new cleanup method for repo cache interface
- call cleanup method in finalize repository
- implement cleanup for Redis which disconnects the client - otherwise renovate keeps hanging indefinitely at the end
- set cache TTL using Redis EX parameter

## Changes

Add Redis support to repository cache in addition to the existing S3 support. Add a cleanup stage for repository cache in case some backend (like Redis) needs to perform some operations upon closing.

## Context

Adding a new backend will give users more options for caching, and more importantly, will allow the users to use the same database for package and repo cache. An attempt was made to implement this in the past, but was never finished due to refactoring that happened at the same time https://github.com/renovatebot/renovate/pull/16735 https://github.com/renovatebot/renovate/discussions/16497


lib/util/cache/repository/impl/redis.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  beforeEach(() => {
+    GlobalConfig.set({ cacheDir: '/tmp/cache', platform: 'github' });
+    vi.mocked(normalizeRedisUrl).mockImplementation((url) => url);
+    redisCache = new RepoCacheRedis(repository, fingerprint, url);
+  });
+
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
```
viceice, line 30: I think this is already enabled globally, so no need to call it

lib/util/cache/repository/impl/redis.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  describe('read()', () => {
+    it('successfully reads from redis', async () => {
+      const mockClient = {
+        get: vi.fn().mockResolvedValue(JSON.stringify(repoCache)),
+        connect: vi.fn(),
+      };
+      vi.mocked(createClient).mockReturnValue(mockClient as any);
```
viceice, line 38: no any , use partial helper

lib/util/cache/repository/impl/redis.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.redisClient = null;
+  }
+
+  async read(): Promise<string | null> {
+    this.redisClient ??= await this.getRedisClient();
+    const cacheKey = this.getCacheKey();
+    try {
+      return (await this.redisClient?.get(cacheKey)) ?? null;
+    } catch {
+      logger.debug({ cacheKey }, 'Repository cache not found');
```
viceice, line 34: trace log error, otherwise debugging issues from logs are not possible

viceice (changes requested):

PR #35480, opened 2025-04-22
feat: Implement hashicorp/go-version completely

ademmers:
## Changes

This change will support exclusion patterns in Hashicorp versioning.

## Context

Previous implementation of Hashicorp versioning scheme was a derivation of semver with some npm range mappings, hence was unable to support the version exclusion pattern that Hashicorp supports in it's constraint definition.


rarkins (commented):
How much of this code is copy/pasted/transformed from  https://github.com/hashicorp/go-version ?

ademmers:
contraint.ts and version.ts and their respective spec.ts files: 90+%. Some changes were done (rather last minute during testing) to toString() requirements mostly of Version and Constraints classes. Spec changes have even been put in comments of respective tests.

rarkins:
In that case we need to check the licensing and copyright implications of that. Probably at least including full acknowledgements at the top of such files

rarkins:
If this is based so directly from the Hashicorp lib then it would be better if this is created as a standalone library (with direct attribution) and used as a dependency here, like we do with some other versioning libraries like pep440. I'm not sure if you'll have the time for this, but perhaps someone will one day. Thanks!

PR #35508, opened 2025-04-23
fix(manager/gomod): use regex to remove version suffix from name

Shegox:
## Changes

This PR uses a regular expression to remove the `/v2` or `/v3` suffix from go packages. This is later used to determine if suffix can be updated or appended. 

## Context

The change is required for packages which don't contain a `org/repo` but only a package name like `sigs.k8s.io/structured-merge-diff`. Currently Renovate appends here the new major version to construct the wrong `sigs.k8s.io/structured-merge-diff/v4/v6` rather than the expected `sigs.k8s.io/structured-merge-diff/v6` 

fixes #35507

Tests are added both for the short example as well as an example with multiple (3+) name components.

### DIFF
--- a/lib/modules/manager/gomod/update.ts
+++ b/lib/modules/manager/gomod/update.ts
@@ -4,7 +4,9 @@ import { newlineRegex, regEx } from '../../../util/regex';
 import type { UpdateDependencyConfig } from '../types';
 
 function getNameWithNoVersion(name: string): string {
-  let nameNoVersion = name.split('/').slice(0, 3).join('/');
+  // remove version suffixes like /v1 or /v2
+  let nameNoVersion = name.replace(/\/v\d+$/, '');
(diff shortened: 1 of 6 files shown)


rarkins (approved):

lib/modules/manager/gomod/update.ts
```diff
@@ -4,7 +4,9 @@ import { newlineRegex, regEx } from '../../../util/regex';
 import type { UpdateDependencyConfig } from '../types';
 
 function getNameWithNoVersion(name: string): string {
-  let nameNoVersion = name.split('/').slice(0, 3).join('/');
+  // remove version suffixes like /v1 or /v2
+  let nameNoVersion = name.replace(/\/v\d+$/, '');
```
viceice, line 8: should use our regEx helper

Shegox: will open a PR for that 👍

PR #35515, opened 2025-04-24
feat(datasource/maven)!: use latest and release values as tags

### LINKED ISSUE #35503: Include latest and release tags with Maven datasource results

Maven metadata can be like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<metadata>
  <groupId>my-groupid</groupId>
  <artifactId>artifact-id</artifactId>
  <versioning>
    <latest>110401-5e96bb1</latest>
    <release>110401-5e96bb1</release>
    <versions>
      <version>110401-5e96bb1</version>
    </versions>
    <lastUpdated>20250410115847</lastUpdated>
  </versioning>
</metadata>
```

The two fields `latest` and `release`, if present, should be returned as `tags`, per this type: https://github.com/renovatebot/renovate/blob/57426756de78f9c49f638663b30b6d3a891087c2/lib/modules/datasource/types.ts#L83

rarkins:
## Changes

Uses "latest" and "release" values in `maven-metadata.xml` as tags if present.

This could have an effect on some exist package lookups because prior to this PR, the Renovate concept of `respectLatest` was not applicable, whereas now it is.

## Context

- Closes #35503

### DIFF
--- a/lib/modules/datasource/index.ts
+++ b/lib/modules/datasource/index.ts
@@ -42,8 +42,6 @@ export const getDatasourceList = (): string[] => Array.from(datasources.keys());
 
 const cacheNamespace = 'datasource-releases';
 
-const mavenVersioning = versioning.get('maven');
-
 type GetReleasesInternalConfig = GetReleasesConfig & GetPkgReleasesConfig;
 
 // TODO: fix error Type
@@ -167,6 +165,7 @@ async function mergeRegistries(
   let combinedRes: ReleaseResult | undefined;
   let lastErr: Error | undefined;
   let singleRegistry = true;
+  const releaseVersioning = versioning.get(config.versioning);
   for (const registryUrl of registryUrls) {
     try {
       const res = await getRegistryReleases(datasource, config, registryUrl);
@@ -206,20 +205,20 @@ async function mergeRegistries(
           for (const tag of ['release', 'latest']) {
             const existingTag = combinedRes?.tags?.[tag];
             const newTag = res.tags?.[tag];
-            if (is.string(newTag) && mavenVersioning.isVersion(newTag)) {
+            if (is.string(newTag) && releaseVersioning.isVersion(newTag)) {
               if (
                 is.string(existingTag) &&
-                mavenVersioning.isVersion(existingTag)
+                releaseVersioning.isVersion(existingTag)
               ) {
                 // We need to compare them
-                if (mavenVersioning.isGreaterThan(newTag, existingTag)) {
+                if (releaseVersioning.isGreaterThan(newTag, existingTag)) {
                   // New tag is greater than the existing one
-                  tags![tag] = newTag;
+                  tags[tag] = newTag;
                 }
               } else {
                 // Existing tag was not present or not a version
                 // so we can just use the new one
-                tags![tag] = newTag;
+                tags[tag] = newTag;
               }
             }
           }
--- a/lib/modules/datasource/index.ts
+++ b/lib/modules/datasource/index.ts
@@ -12,6 +12,7 @@ import { filterMap } from '../../util/filter-map';
 import { AsyncResult, Result } from '../../util/result';
 import { DatasourceCacheStats } from '../../util/stats';
 import { trimTrailingSlash } from '../../util/url';
+import * as versioning from '../versioning';
 import datasources from './api';
 import {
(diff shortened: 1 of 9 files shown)


rarkins:
Because the Maven datasource can use "merge" strategy, we need to think about the implications of this for tags. Different registries may have different `latest` and `release` tags. Should we take the highest of the tags if multiple apply?

Churro:
> Because the Maven datasource can use "merge" strategy, we need to think about the implications of this for tags. Different registries may have different `latest` and `release` tags. Should we take the highest of the tags if multiple apply?

If I read it correctly, maven resolves `latest` and `release` by merging metadata from all registries and selecting the value with the most recent `lastUpdated` timestamp for each tag - not by picking the numerically highest version. If two registries propose different `latest` values, the one whose metadata has the most recent `lastUpdated` timestamp for the tag will "win".

The evidence that leads to this conclusion can be found in maven's `DefaultVersionResolver.java` ([ref](https://github.com/apache/maven/blob/08177a38eb86cac7b7d14d788003cad6b140fe2e/impl/maven-impl/src/main/java/org/apache/maven/impl/resolver/DefaultVersionResolver.java#L295-L338)) for regular deps and `DefaultPluginVersionResolver.java` ([ref](https://github.com/apache/maven/blob/08177a38eb86cac7b7d14d788003cad6b140fe2e/impl/maven-core/src/main/java/org/apache/maven/plugin/version/internal/DefaultPluginVersionResolver.java#L342-L371)) for plugins.

Generally, taking the highest of the tags sounds reasonable to me. In some registries like artifactory it is possible to trigger a rebuild of Maven metadata, so linking the use of `latest` to other criteria, such as `lastUpdated`, could actually lead to wrong selections.

viceice (dismissed):

lib/modules/datasource/index.ts
```diff
@@ -172,24 +175,64 @@ async function mergeRegistries(
...
       }
 
       combinedRes.releases.push(...releases);
+
+      // Merge the tags from the two results
+      let tags = combinedRes.tags;
+      if (tags) {
+        if (res.tags) {
+          // Both results had tags, so we need to compare them
+          ['release', 'latest'].forEach((tag) => {
```
viceice, line 206: please use for-of loop

lib/modules/datasource/index.ts
```diff
@@ -172,24 +175,64 @@ async function mergeRegistries(
...
+            const newTag = res.tags?.[tag];
+            if (is.string(newTag) && mavenVersioning.isVersion(newTag)) {
+              if (
+                is.string(existingTag) &&
+                mavenVersioning.isVersion(existingTag)
+              ) {
+                // We need to compare them
+                if (mavenVersioning.isGreaterThan(newTag, existingTag)) {
+                  // New tag is greater than the existing one
+                  tags![tag] = newTag;
```
viceice, line 217:
```suggestion
                  tags[tag] = newTag;
```

lib/modules/datasource/index.ts
```diff
@@ -172,24 +175,64 @@ async function mergeRegistries(
...
+                }
+              } else {
+                // Existing tag was not present or not a version
+                // so we can just use the new one
+                tags![tag] = newTag;
```
viceice, line 222:
```suggestion
                tags[tag] = newTag;
```

lib/modules/datasource/index.ts
```diff
@@ -172,24 +175,64 @@ async function mergeRegistries(
...
+
+      // Merge the tags from the two results
+      let tags = combinedRes.tags;
+      if (tags) {
+        if (res.tags) {
+          // Both results had tags, so we need to compare them
+          for (const tag of ['release', 'latest']) {
+            const existingTag = combinedRes?.tags?.[tag];
+            const newTag = res.tags?.[tag];
+            if (is.string(newTag) && mavenVersioning.isVersion(newTag)) {
```
viceice, line 209: so this combining is only done for maven? It would also happen if another datasource is used and the versions are maven compatible.

Should we guard this to only work on maven datasource?

nuget has also hunt strategy, but doesn't return tags yet. it may in future and nobody would probably think about this maven thing here.

rarkins: Good catch, I changed it to be generic

viceice (approved):

PR #35521, opened 2025-04-24
fix(config): merge globalExtends before default config

csmuller:
## Changes

Moves globalExtends resolving/merging earlier so that it occurs before defaults merging.

## Context

- https://github.com/renovatebot/renovate/discussions/32307


lib/workers/global/config/parse/index.spec.ts
```diff
@@ -165,6 +166,33 @@ describe('workers/global/config/parse/index', () => {
...
+      vi.doMock('../../../../../config.js', () => ({
+        default: {
+          globalExtends: ['http://example.com/config.json', ':pinVersions'],
+          dryRun: 'extract', // This should overwrite the one from the globalExtends
+        },
+      }));
+      httpMock
+        .scope('http://example.com/')
+        .get('/config.json')
+        .reply(200, { repositories: ['g/r1', 'g/r2'], druRun: 'full' });
```
rarkins, line 179:
```suggestion
        .reply(200, { repositories: ['g/r1', 'g/r2'], dryRun: 'full' });
```

rarkins (dismissed):

csmuller:
The tests behave differently on GitHub Actions than when run locally. Apparently, the mock that is set up with `httpMock` doesn't get used. Thus, the test fails and `httpMock` complains about an unused mock.
@rarkins Does this ring a bell? If not, then I guess I'll have to mock the call to a remote preset differently.

rarkins:
Actually the reason I paused before testing was because wasn't sure if http mocking had changed when we switched to vitest. @viceice any tips?

csmuller:
@rarkins I replaced the `httpMock` with a mock of `getPreset` in `/lib/config/presets/http`.
I assume the tests will tests run smoothly now.

csmuller:
Ok, another try. Actually, the problem seems to be the mock of the config.js. I replaced that with a real config file in the fixtures. Let's see if this time it works independently of where the tests run 🙄
Sorry, I know I shouldn't, but had to force push because of wrong setting of the commit author.

lib/workers/global/config/parse/__fixtures__/config-global-extends.js
rarkins, line 1: Convert this to an inline string in the test, instead of a separate file

rarkins (changes requested):

rarkins (dismissed):

viceice (dismissed):

lib/workers/global/config/parse/index.ts
```diff
@@ -33,6 +50,18 @@ export async function parseConfigs(
 
   config = await codespaces.setConfig(config);
 
+  let resolvedGlobalExtends: AllConfig | undefined;
+
+  if (is.nonEmptyArray(config?.globalExtends)) {
+    // resolve global presets immediately
+    resolvedGlobalExtends = await resolveGlobalExtends(
```
viceice, line 57: we don't have any global config yet, so things like host rules are not applied while resolving

csmuller: When quickly checking, I don't see any authentication happening in the preset-fetching code, but maybe I'm mistaken.
For us it probably always worked because in our internal network the repo with the presets is not private.

Either way, it looks like this was problem even before the globalExtends bug came up, which we are trying to fix here.

rarkins: If globalExtends previously didn't use auth then we can merge this then

csmuller: Here's why I think that the host rules were not considered for resolving the global presets even before the changes of this PR: https://github.com/renovatebot/renovate/blob/e09764fceceb55cbea88ab8653039e1e7aa8b974/lib/workers/global/index.ts#L139-L154
`globalInitialize(config)` is called only after `resolveGlobalExtends(...)` is called. The `globalInitialize` method contains the [initialization](https://github.com/renovatebot/renovate/blob/6946d5b0ecab7706f2eb804beb82b1a6c98a7510/lib/workers/global/initialize.ts#L85) of the host rules.

rarkins: Confirmed via our _existing_ docs:

image

rarkins (approved):

PR #35537, opened 2025-04-24
feat(terraform): add support for `docker_registry_image` data source

interad-woergoetter:
## Changes

Adds support for the data source docker_registry_image to the GenericDockerImageRefExtractor of teh terraform manager.

## Context

The docker_registry_image data source is a minimal resource that only checks if the image in that version is available on the registry. With a (recent change to the kreuzwerker/docker terraform provider)[https://github.com/kreuzwerker/terraform-provider-docker/pull/703] it is now possible to use this data source without having to have a docker host available at all.
There are use cases where you only need to reference docker images in your terraform configuration without actually pulling them or even having docker installed, e.g. for Azure App Service site containers.
Lazzu also requested support for this in discussion #26783.
This change will enable users to keep these references updated using Renovate and the kreuzwerker/docker provider without needing to configure complicated custom matching rules.

I have some concerns that this could interfere with users who auto-update their docker images with kreuzwerker/docker as in (this example usage)[https://github.com/kreuzwerker/terraform-provider-docker/pull/703], but disabling individual depTypes is supported, as far as I know, which mitigates the issue.


lib/modules/manager/terraform/extractors/resources/utils.ts
```diff
@@ -35,6 +35,10 @@ const KubernetesJobTemplateInit = [
   'image',
 ];
 
+export const generic_image_data_source: GenericImageResourceDef[] = [
```
viceice, line 38: please move to generic image resource below. then you need less changes

interad-woergoetter: Just merging this array with the generic_image_resource would not achieve the goal of my change, because docker_registry_image is used as a terraform data source and not a resource in my use case.

I could of course check the terraform data sources for all the other generic images also and not just docker_registry_image.
Is that what you want? It's less code changes, for sure, but a more significant change in behavior. On the other hand, it would make it more consistent than what I wrote now 🤔

interad-woergoetter: Okay, I did what you asked for. Let me know if you prefer this. We can also just revert the latest commit if you prefer the previous solution :)

lib/modules/manager/terraform/extractors/resources/utils.ts
```diff
@@ -40,6 +40,7 @@ export const generic_image_resource: GenericImageResourceDef[] = [
   { type: 'docker_image', path: ['name'] },
   { type: 'docker_container', path: ['image'] },
   { type: 'docker_service', path: ['task_spec', 'container_spec', 'image'] },
+  { type: 'docker_registry_image', path: ['name'] },
```
viceice, line 43: sorry, but i see now the issue. we should not mix resources and datasources. can you split them again 🙏 

It'll be very expensive to run alll definitions over datasources and resources

lib/modules/manager/terraform/__fixtures__/docker.tf
```diff
@@ -8,7 +13,7 @@ resource "docker_image" "invalid" {
 }
 
 resource "docker_image" "ignore_variable" {
-  name          = "${data.docker_registry_image.ubuntu.name}"
+  name          = data.docker_registry_image.ubuntu.name
```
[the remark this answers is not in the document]
interad-woergoetter: yeah sorry, that was my terraform auto formatting on save. That syntax is officially deprecated, so I didn't bother changing it back ;)

lib/modules/manager/terraform/extractors/resources/utils.ts
interad-woergoetter: No worries :)
I reverted to the original state 👍

lib/modules/manager/terraform/extractors/resources/utils.ts
viceice, line 38:
```suggestion
export const generic_image_datasource: GenericImageResourceDef[] = [
```

viceice (approved):

secustor (approved):

PR #35561, opened 2025-04-25
fix(gradle): Correctly download the new Gradle wrapper files after updating Gradle

eygraber:
## Changes

When updating Gradle, there is a requirement that the wrapper task be run twice; first to update the version that the wrapper uses, and second to update the wrapper and associated files correctly.

## Context

https://github.com/renovatebot/renovate/discussions/34273

### DIFF
--- a/lib/modules/manager/gradle-wrapper/artifacts.ts
+++ b/lib/modules/manager/gradle-wrapper/artifacts.ts
@@ -194,6 +194,48 @@ export async function updateArtifacts({
       );
     }
 
+    let writeLocksCmd = await prepareGradleCommand(gradlewFile);
+    if (!writeLocksCmd) {
+      logger.info('No gradlew found - skipping Artifacts update');
+      return null;
+    }
+    writeLocksCmd += ' :wrapper --write-locks';
+
+    logger.debug(`Downloading the new gradle wrapper and updating the locks: "${writeLocksCmd}"`);
+    try {
+      await exec(writeLocksCmd, execOptions);
+    } catch (err) {
+      // istanbul ignore if
+      if (err.message === TEMPORARY_ERROR) {
+        throw err;
+      }
+      logger.warn(
+        { err },
+        'Error executing gradle wrapper download and lock update command. It can be not a critical one though.',
+      );
+    }
+
+    let updateWrapperCmd = await prepareGradleCommand(gradlewFile);
+    if (!updateWrapperCmd) {
+      logger.info('No gradlew found - skipping Artifacts update');
+      return null;
+    }
+    updateWrapperCmd += ' :wrapper';
+
+    logger.debug(`Updating the gradle wrapper files: "${writeLocksCmd}"`);
+    try {
+      await exec(updateWrapperCmd, execOptions);
+    } catch (err) {
+      // istanbul ignore if
+      if (err.message === TEMPORARY_ERROR) {
+        throw err;
+      }
+      logger.warn(
+        { err },
+        'Error executing gradle wrapper update wrapper files. It can be not a critical one though.',
+      );
+    }
+
     const buildFileName = await updateBuildFile(localGradleDir, {
       gradleVersion: config.newValue,
       distributionSha256Sum: checksum,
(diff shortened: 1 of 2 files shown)


lib/modules/manager/gradle-wrapper/artifacts.ts
```diff
@@ -194,6 +194,48 @@ export async function updateArtifacts({
       );
     }
 
+    let writeLocksCmd = await prepareGradleCommand(gradlewFile);
```
rarkins, line 197: This function should only need to call `prepareGradleCommand()` once and after that the value should be reused.

lib/modules/manager/gradle-wrapper/artifacts.ts
```diff
@@ -194,6 +194,48 @@ export async function updateArtifacts({
...
+        'Error executing gradle wrapper download and lock update command. It can be not a critical one though.',
+      );
+    }
+
+    let updateWrapperCmd = await prepareGradleCommand(gradlewFile);
+    if (!updateWrapperCmd) {
+      logger.info('No gradlew found - skipping Artifacts update');
+      return null;
+    }
+    updateWrapperCmd += ' :wrapper';
```
rarkins, line 223: Do I understand correctly that this function will now:
- Run gradlew :wrapper
- Run gradlew :wrapper --write-locks
- Run gradlew :wrapper again?

Even if this is correct, we need to document the reasoning behind this logic for future developers/maintainers

rarkins (changes requested):

lib/modules/manager/gradle-wrapper/artifacts.ts
```diff
@@ -194,6 +194,48 @@ export async function updateArtifacts({
       );
     }
 
+    let writeLocksCmd = await prepareGradleCommand(gradlewFile);
+    if (!writeLocksCmd) {
+      logger.info('No gradlew found - skipping Artifacts update');
+      return null;
+    }
+    writeLocksCmd += ' :wrapper --write-locks';
```
Churro, line 202: This ignores gradle subprojects

Churro (changes requested):
The changes proposed in this PR are not relevant for most gradle-wrapper updates. 
* `--write-locks` is specific to the use of gradle lockfiles, which is an optional feature. Calling this during wrapper updates, refreshes the locks of all dependencies, i.e., direct dependencies with dynamic versions and transitive dependencies. From a user perspective, this can be unexpected, since gradle itself doesn't include such behavior. The only other place where this happens is during gradle lock file maintenance but only if lockfiles exist at all.
* Calling `:wrapper` re-builds the entire project against the downloaded gradle distribution and replaces `gradle-wrapper.jar` and the `gradlew` / `gradlew.bat` scripts. This is superfluous here since the build output is not used in any way and bootstrap files don't need regular updates.

Generally, running these gradle tasks would at least double the runtime of gradle-wrapper updates for all renovate users, regardless if they use lockfiles or feel that they want renewed bootstrap files. Therefore, I don't think this is a universally reasonable change.

lib/modules/manager/gradle-wrapper/artifacts.ts
eygraber: If we move forward with leaving out the `--write-locks` portion of the PR, I can probably just reuse `cmd` as is since including the `gradle-distribution-url` for the second execution would work as well.

lib/modules/manager/gradle-wrapper/artifacts.ts
eygraber: Pending the resolution of the discussion in https://github.com/renovatebot/renovate/discussions/34273 the plan would be to run gradlew :wrapper twice. I can add in documentation explaining why that is the case.

lib/modules/manager/gradle-wrapper/artifacts.ts
eygraber: Can you elaborate? AFAIK the wrapper task is only present on the root project. The difference between `:wrapper` and `wrapper` is semantics, and since Renovate was already using `:wrapper` I didn't want to change that, even though the convention is to use `wrapper`.

PR #24538, opened 2023-09-20
feat: separateMultipleMinor

phyzical:
## Changes

Attempt at supporting the `seperateMultipleMinor`

* added logic to adjust branch topic when enabled
* added logic to pull all minor versions to buckets when enabled

Will follow up with one for patch also when/is this is merged

## Context

My usecase is seperating eks versions as individual prs as they are "minor versions" but you must go 1 by 1 during an update

- closes #6576

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

below is it run for my exact usecase happy to try others if suggested

https://github.com/phyzical/renovate-tester/pulls

It works for my tiny usecase.

but I think ive done this wrong though... as i felt like i was fighting the ```{{#if separateMinorPatch}}{{#if isPatch}}``` logic so open to feedback

i also couldn't find an example of `separateMultipleMajor` being run from a multiple branch test context, there just seems to be hardcoded tests against static templates. or have i not looked hard enough

Thanks!

PR #35606, opened 2025-04-28
feat(config)!: add minor version to default branchTopic to support separateMultipleMinor

colinodell:
## Changes

Changes the default `branchTopic` to include the minor version when `separateMultipleMinor` is enabled, ensuring that `separateMultipleMinor` works properly when `hashedBranchLength` is also set.

(This is an alternative solution to #35607 - only one or the other should be merged)

### Breaking Change

**This will be a breaking change** for any users that have customized their `branchTopic` and use `separateMultipleMinor=true`. In those cases, the minor version version will no longer be included in the branch names until they make a similar modification to their `branchTopic`.

For all other users (who haven't set `branchTopic` or `separateMultipleMinor`) the behavior should be unchanged.

## Context

- #35553

The current implementation (from #24538) uses hard-coded logic to add the minor version to the branch name:

https://github.com/renovatebot/renovate/blob/8c78a7e1ceb238845760d851affc5fabffc86e05/lib/workers/repository/updates/branch-name.ts#L123-L125

But this code never runs if `hashedBranchLength` is set, thus breaking this feature!

We could fix this either by:

- Duplicating the existing hardcoded logic into the other `if` branch (#35607); or,
- Moving the implementation of this feature into the `branchTopic` template

This PR implements the latter.

Closes #35553


colinodell:
I wasn't sure if this should be considered a `fix` (because it fixes #35553) or a `feat`ure (because it improves the implementation and changes a default option), so please let me know if I should relabel this.

lib/workers/repository/updates/branch-name.spec.ts
```diff
@@ -305,13 +305,32 @@ describe('workers/repository/updates/branch-name', () => {
...
+        depNameSanitized: 'jest',
+        newMajor: '42',
+        separateMultipleMinor: true,
+        isMinor: true,
+        newMinor: '3',
+        group: {},
+      };
+      generateBranchName(upgrade);
+      expect(upgrade.branchName).toBe('dep-2e27927800');
+    });
```
rarkins, line 332: Can you add a non-hashed version of this test?

colinodell: > How then does this change not impact any other tests?

Oh, that's because the added logic in `branchTopic` mirrors the original logic I removed from the non-hashing part of `generateBranchName()`!

> image

Put another way...

Prior to this PR, the extra `.${newMinor}` string was only inserted in the `else` branch:

```ts
// over-simplified for brevity
if (update.hashedBranchLength) {
  // Compile the branchTopic, hash it, and use it as the branch name
} else {
  // Compile the branchName (based on the branchTopic)
  if (update.updateType === 'minor' && update.separateMultipleMinor) {
    update.branchName = update.branchName.replace('.x', `.${newMinor}.x`);
  }
  // Use that as the branch name
}
```

And the existing tests (like `separateMultipleMinor=true` on lines 92-110) covered this.

**I've shifted that hard-coded logic out of the `else` branch and into the `branchTopic` template.** Because both `if (update.hashedBranchLength)` and `else` already use `branchTopic`, both now get that logic. This means the `else` (non-hashing) part **retains its original behavior**, thus the original tests are unaffected.

rarkins: In that case, I'm not sure we need to treat it as a breaking change? If it fixes a problem with hash branch lengths, and other use cases are unchanged, that seems reasonable as a fix instead of a breaking feature. What do you think?

colinodell: > it fixes a problem with hash branch lengths, and other use cases are unchanged

Yeah, this will be true for _**most**_ users. However, if somebody has set their own custom `branchTopic`, they won't get this new default, and so `separateMultipleMinor=true` will stop working for them (until they make a similar change in their own custom `branchTopic`).

rarkins (approved):

PR #35607, opened 2025-04-28
fix: support separateMultipleMinor when using hashed branches

colinodell:
## Changes

Ensures that `separateMultipleMinor` is respected when `hashedBranchLength` is set by incorporating the minor version into the hash input.

(This is an alternative solution to #35606 - only one or the other should be merged)

## Context

- #35553

The current implementation (from #24538) uses hard-coded logic to add the minor version to the branch name:

https://github.com/renovatebot/renovate/blob/8c78a7e1ceb238845760d851affc5fabffc86e05/lib/workers/repository/updates/branch-name.ts#L123-L125

But this code never runs if `hashedBranchLength` is set, thus breaking this feature!

We could fix this either by:

- Duplicating the existing hardcoded logic into the other `if` branch (**this PR**); or,
- Moving the implementation of this feature into the `branchTopic` template (#35606)

This PR implements the former.

Closes #35553

### DIFF
--- a/lib/workers/repository/updates/branch-name.spec.ts
+++ b/lib/workers/repository/updates/branch-name.spec.ts
@@ -322,10 +322,13 @@ describe('workers/repository/updates/branch-name', () => {
         branchPrefix: 'dep-',
         depNameSanitized: 'jest',
         newMajor: '42',
+        updateType: 'minor',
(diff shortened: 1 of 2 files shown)


colinodell:
Closed in favor of #35606

PR #35622, opened 2025-04-29
fix(docker): Check cacheability for fetched tags

zharinov:
## Context

- Ref: #35467


lib/modules/datasource/docker/index.ts
```diff
@@ -712,6 +713,14 @@ export class DockerDatasource extends Datasource {
     namespace: 'datasource-docker-tags',
     key: (registryHost: string, dockerRepository: string) =>
       `${registryHost}:${dockerRepository}`,
+    cacheable: (registryHost: string, _dockerRepository: string) => {
+      const { password, token } = hostRules.find({
+        hostType: dockerDatasourceId,
+        url: registryHost,
+      });
+      const secret = token ?? password;
+      return !secret;
```
rarkins, line 722: Unfortunately.. we have a token for all Docker Hub, to increase API limits. So I think this code would mean we don't cache Docker Hub tag lookups anymore.

zharinov: This function never fetches docker hub

rarkins: In that case, this could be OK

viceice (dismissed):

PR #35628, opened 2025-04-29
feat(cache): Track repeating package cache read access

zharinov:
## Changes

We deliberately leak all the `Promise` objects returned by `packageCache.get(...)` or passed to `packageCache.set(...)` which makes it the biggest contributor to the memory footprint.

We need to discover the keys with repeating access and apply such caching to the limited set of namespaces/keys.

## Context

- #6202

### DIFF
--- a/lib/util/stats.ts
+++ b/lib/util/stats.ts
@@ -548,20 +548,19 @@ export class ObsoleteCacheHitLogger {
       'Cache fallback URLs',
     );
   }
-}
-/* v8 ignore stop: temporary code */
+} /* v8 ignore stop: temporary code */
 
 /* v8 ignore start: temporary code */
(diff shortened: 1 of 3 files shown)


lib/util/stats.ts
```diff
@@ -547,5 +548,44 @@ export class ObsoleteCacheHitLogger {
       'Cache fallback URLs',
     );
   }
-}
-/* v8 ignore stop: temporary code */
+} /* v8 ignore stop: temporary code */
+
+/* v8 ignore start: temporary code */
+export class RepeatingPackageCacheKeyLogger {
+  static getData(): Record<string, Record<string, number>> {
```
viceice, line 555: make private?

zharinov:
Not necessary now since we clean up these keys after fetching is done

PR #35634, opened 2025-04-30
fix(memcache): http requests being garbage collected

felipecrs:
## Changes

This changes the `memCache` to use an in-memory SQLite database for storing strings. Such database is not subject to garbage collection.

## Context

It took me a long while to understand this issue.

To begin, this issue started with:

- https://github.com/renovatebot/renovate/pull/33901

I thought I had fixed everything about it for Gerrit platform in:

- https://github.com/renovatebot/renovate/pull/35056

However, a colleague noticed that all `prNo` were being set as `null` in our Renovate reports. We identified it started happening (consistently) after some Renovate upgrade. Then we pinned down to #33901.

Now that http responses are being cached in a JavaScript object in memory, such object is subject to garbage collection.

The garbage collector starts erasing some properties like `httpResponse.body` when memory usage gets high.

Because of that, the memory cache starts serving broken requests as if they were properly cached.

This also seems to be the root cause of:

- https://github.com/renovatebot/renovate/discussions/35367

Note I also considered saving a hash of the `httpResponse` together with the response itself so that the cache could be validated upon read, but in that case, it would mean more http requests because instead of ensuring the cache is not erased, it only ensures the cache is valid.

To be honest I have not profiled Renovate to ensure the culprit is actually the garbage collector, however I verified that this solution fixes the issue.

### DIFF
--- a/lib/util/cache/memory/index.spec.ts
+++ b/lib/util/cache/memory/index.spec.ts
@@ -5,16 +5,45 @@ describe('util/cache/memory/index', () => {
     expect(memCache.get('key1')).toBeUndefined();
   });
 
-  it('sets and gets repo cache', () => {
+  it('sets and gets repo cache for strings', () => {
     memCache.init();
     memCache.set('key2', 'value');
     expect(memCache.get('key2')).toBe('value');
   });
 
+  it('sets and gets repo cache for objects', () => {
+    memCache.init();
+    const obj = { some: 'thing' };
+    memCache.set('key3', obj);
+    expect(memCache.get('key3')).toBe(obj);
+  });
+
+  it('sets and gets repo cache for promises', () => {
+    memCache.init();
+    const promise = new Promise(() => undefined);
+    memCache.set('key4', promise);
+    expect(memCache.get('key4')).toBe(promise);
+  });
+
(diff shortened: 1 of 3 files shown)


felipecrs:
I can also explore using these libraries to check if they also fix the issue:

https://www.npmjs.com/package/cache-manager
https://www.npmjs.com/package/safe-memory-cache

rarkins:
I think we need a Discussion first.

I would be very surprised if Node just deletes ("garbage collection") variables we put in memory. What is your source for this claim?

felipecrs:
Debugging step by step and noticing some properties gets deleted from `repoCache` without invoking the set method. They seem to be deleted to free up space for more entries.

I can consistently and relatively easily replicate this and would be ok even showing it in a call.

If you'd like to take a more conservative approach for this, I can implement the hashing mechanism I mentioned, and then log every time the hashes don't match.

This would allow you to keep track of situations where the cache got tampered with, so that you can consider this fix.

felipecrs:
Again, I am not a Node.js expert but am assuming the garbage collector is the culprit.

Regardless of this assumption being correct or not, this PR fixes the problem.

PR #35642, opened 2025-04-30
fix(bitbucket-server): do not persist bearer token in git config as it leads to issues behind aws elb

mwiede:
## Changes

As described in #30180 when a repository is cloned with bearer token, it will be written to the `.git/config` file. Now then another call on this repo is done, the bearer token is taken twice in the http call, which is invalid and therefore rejected by the AWS ELB.
This PR changed the clone options and moves the extraheader to the git command options (<https://git-scm.com/docs/git#Documentation/git.txt--cltnamegtltvaluegt>) rather than the clone command options. These are not written to the repository config. 

## Context

This change is only done for the bitbucket-server and azure platform, as these are the only ones using the extra options.

### DIFF
--- a/lib/util/git/types.ts
+++ b/lib/util/git/types.ts
@@ -19,6 +19,7 @@ export interface StorageConfig {
   currentBranch?: string;
   defaultBranch?: string;
   url: string;
+  extraGitOptions?: GitOptions;
   upstreamUrl?: string | undefined;
   extraCloneOpts?: GitOptions;
   cloneSubmodules?: boolean;
--- a/lib/modules/platform/azure/index.ts
+++ b/lib/modules/platform/azure/index.ts
@@ -239,7 +239,7 @@ export async function initRepo({
   await git.initRepo({
     ...config,
     url,
-    extraCloneOpts: getStorageExtraCloneOpts(opts),
+    extraGitOptions: getStorageExtraCloneOpts(opts),
     cloneSubmodules,
     cloneSubmodulesFilter,
   });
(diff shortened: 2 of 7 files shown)


rarkins (commented):
The refactoring/renaming done makes it harder to review for the _logic changes_

mwiede:
sorry I needed to force push this as my companies user was in the way.

mwiede:
@rarkins what I could do to make the code clean, is, that I could remove all the `extraCloneOpts` because they are not needed any more

rarkins (changes requested):
It would be best to either:
- Revert the `extraCloneOpts`-> `extraGitOptions` renaming, or
- Refactor code with that renaming separately/first to separate it from the logic change

lib/util/git/types.ts
```diff
@@ -19,6 +19,7 @@ export interface StorageConfig {
   currentBranch?: string;
   defaultBranch?: string;
   url: string;
+  extraGitOptions?: GitOptions;
   extraCloneOpts?: GitOptions;
```
rarkins, line 23: Is this still used?

mwiede: yes, this is what the relevant change of this pr is, see lib/util/git/index.ts line 199 and line 444

viceice (commented):
please do the `extraGitOptions` refactor in a separate PR

conflicted

mwiede:
@viceice from my point of view it does not make sense to split this into a separate PR. The whole point is, that the property extraCloneOpts is only used for authentication and with my change, extraCloneOpts is not needed any more.

viceice:
> @viceice from my point of view it does not make sense to split this into a separate PR. The whole point is, that the property extraCloneOpts is only used for authentication and with my change, extraCloneOpts is not needed any more.

we almost always do such refactorings before it's really used. so why we should make a difference here? you need unit tests anyways to get full coverage.

PR #35647, opened 2025-04-30
fix(memCache): disallow http requests in cache to be mutable

felipecrs:
## Changes

Fixes an issue with the memoery http cache provider where the cached http reponses could be mutated after being read from the cache.

This could cause several unexected issues, like returning a cache hit with a different body than the one that was cached.

So far, only the Gerrit platform was identified as being affected by this issue.

## Context

- Closes https://github.com/renovatebot/renovate/discussions/35638
- Supersedes https://github.com/renovatebot/renovate/pull/35634

### DIFF
--- a/lib/util/http/cache/memory-http-cache-provider.spec.ts
+++ b/lib/util/http/cache/memory-http-cache-provider.spec.ts
@@ -36,4 +36,38 @@ describe('util/http/cache/memory-http-cache-provider', () => {
       body: { msg: 'Hello, world!' },
     });
   });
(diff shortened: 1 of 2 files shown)


zharinov (approved):

viceice (approved):
this causes performance, not sure if it's measurable. 🤔

felipecrs:
> this causes performance, not sure if it's measurable. 🤔

Yeah, but that is not worse than the otherwise side effect.

PR #35652, opened 2025-04-30
fix(workers/auto-replace): Correctly replace with digest and without replaceString

jonasrutishauser:
## Changes
Fix replacements if a digest it present (and no `replaceString` exists).

## Context
There are some managers which can't set a `replaceString`. If there is at the same time a digest present and the update is a replacement, it won't work without the fix.


viceice (approved):

viceice:
need a second pair of 👀

secustor (approved):
🤞

PR #35656, opened 2025-04-30
fix(workers/auto-replace): Correctly replace with digest

jonasrutishauser:
## Changes
Fix replacements if a digest it present and version is not changed.

## Context
If the version is not changed in the replacement and as the `currentDigestShort` is present and it is not the same as the `newDigest` (which is equal to the `currentDigest`), the current implementation replaces the short digest, which results in an invalid digest.


lib/workers/repository/update/branch/auto-replace.ts
```diff
@@ -288,6 +288,7 @@ export async function doAutoReplace(
           newDigest,
         );
       } else if (
+        !currentDigest &&
         currentDigestShort &&
```
viceice, line 292: i wonder when we don't have current digest but a short current digest 🤔

jonasrutishauser: It seems that this is possible with github actions: https://github.com/renovatebot/renovate/blob/bc7d0595d0c42f84669ceb7884bef3f54cd6bd42/lib/modules/manager/github-actions/extract.ts#L109-L112

viceice (commented):
please demonstrate on a real repo

RahulGautamSingh (commented):
@jonasrutishauser LGTM. If you're inactive for a week I will try to recreate & test this on a real repo and create a new PR and close this one. As it seems like a bug/issue which should be fixed.

lib/workers/repository/update/branch/auto-replace.ts
RahulGautamSingh: Makes sense.

PR #35661, opened 2025-05-01
fix(gerrit): various fixes and optimizations

felipecrs:
## Changes

I spent the day today improving the Gerrit platform implementation. It would be unfeasible to send a PR for each of the changes I made, so I am sending a single PR with all the changes. Here's a summary of the changes:

- When searching for changes, only request extra details as needed instead of by default. This severely reduces the load under the Gerrit backend
- Remove `SUBMITTABLE`, `CHECK`, `CURRENT_ACTIONS` from default requested details
- Replace `CURRENT_COMMIT` request detail with `COMMIT_WITH_FOOTERS` which is a lot more lightweight
- When searching for a single change, set the limit to 1 in the request so that Gerrit backend can stop searching after the first match
- Pass explicit `no-limit` query to Gerrit as Renovate is not currently able to handle pagination (otherwise the limit is 500), and add a `TODO` to code to implement pagination
- Avoid an API call in `setBranchStatus()` in case the change does not support the label
- Avoid two Gerrit API calls (including a very heavy one that lists all comments of a change) during `createPr()` while ensuring the PR body is added as a comment
- Improve commit message escaping logic to follow official Gerrit recommendation
- Fix Gerrit `status:closed` filter being used to search for closed PRs, `status:abandoned` should have been used instead because `status:closed` in Gerrit also includes merged changes
- Implement a safety check in `createPr()` to ensure Renovate will not mistakenly think it had just created change in the unexpected situations this happens
- Implement `refreshPr()` which is very important given Gerrit has memory cache enabled for requests
- Always pass `-is:wip` and `-is:private` when searching for changes to allow better backend search optimization, Renovate does not currently handle _WIP_ or _Private_ changes anyway
- Move `refreshCache` parameter into `FindPRConfig`, which already had support for a `refreshCache` option
- Fix the Gerrit change into a PR conversion logic which was tolerating a change not having its source branch inferred from the commit message.
  - It was falling back to use the base branch as source branch, which I think should never happen
  - But I added some smartness to reuse already known source branches
- Implement `targetBranch` parameter for `getBranchPr`. It doesn't change anything functionality-wise, but at least optimizes the search.
- Fix `getBranchStatus()` returning `green` if change was submittable but had problems
- Refactor `getBranchStatus()` to avoid taking multiple changes into account.
  - It is not possible that a source branch can map to more than one open change. This just doesn't make sense.
- Fix `getBranchStatusCheck()` returning `green` even if label is marked as _blocking_.
  - This can happen if the label range is from `-2` to `+2` and there's already some `+1` approval.
- Add `sha` property to PRs inferred from Gerrit changes
- Refactored all `.then()` to `async`/`await`
- Remove repository fallback in `getRawFile()` which makes no sense. For example, Azure does the same.
- Remove fallback to `all` when mapping a Gerrit change into PR. A PR state from a real PR should never be `all`.
- Add `TODO`s in code to use newer and better optimized Gerrit API features once more mature
- Remove some useless Gerrit types and improve some existing ones, including some JSDocs

## Context

I use Renovate with Gerrit at work daily, and the lack of optimization of the Gerrit platform implementation is becoming a big problem overloading our Gerrit instance, as we increase the adoption of Renovate.

This PR improves things a lot. But there is a lot more to do. :)


viceice (commented):
this PR is really big,so review will take time

please avoid changes to unrelated code. even if it's only a comment.

felipecrs:
Sure, thank you.

Despite being a big PR, there is not a single big feature shipping with it. It's a collection of small fixes that would otherwise be very time consuming to break into small PRs.

Given  the Gerrit platform is still marked as _experimental_, I think we can afford making these non-trivial changes while working towards a future stabilization of the platform.

Should I revert the comment change in bitbucket-server? The very same comment exists in Gerrit, my intention was to keep them in sync.

lib/modules/platform/gerrit/index.spec.ts
```diff
@@ -252,36 +274,50 @@ describe('modules/platform/gerrit/index', () => {
         `the change should be created automatically from previous push to refs/for/source`,
       );
     });
+
+    it('createPr() - found existing but not created in the last 5 minutes => rejects', async () => {
+      const change = partial<GerritChange>({
+        _number: 123456,
+        // 6 minutes ago
+        created: new Date(Date.now() - 6 * 60 * 1000).toISOString(),
```
viceice, line 282: use luxon, it's much easier to read 

you should use static dates and use vitest fake timers

lib/modules/platform/gerrit/index.spec.ts
```diff
@@ -252,36 +274,50 @@ describe('modules/platform/gerrit/index', () => {
...
     beforeAll(() => {
       gerrit.writeToConfig({ labels: {} });
     });
 
-    const message = 'some subject\n\nsome body\n\nChange-Id: some-change-id';
-
     const change = partial<GerritChange>({
       _number: 123456,
       current_revision: 'some-revision',
+      created: new Date().toISOString(),
```
viceice, line 310: same

lib/modules/platform/gerrit/utils.ts
```diff
@@ -89,14 +109,15 @@ export function mapGerritChangeStateToPrState(
     case 'ABANDONED':
       return 'closed';
   }
-  return 'all';
```
viceice, line 92: are you sure this can never happen in future?

lib/modules/platform/gerrit/utils.ts
```diff
@@ -44,43 +54,53 @@ export function getGerritRepoUrl(repository: string, endpoint: string): string {
   return url.toString();
 }
 
-export function mapPrStateToGerritFilter(state?: PrState): string {
+export function mapPrStateToGerritFilter(state?: PrState): string | undefined {
```
viceice, line 57: can we prefer null instead of undefined?

lib/modules/platform/gerrit/types.ts
```diff
@@ -30,39 +32,54 @@ export interface GerritBranchInfo {
...
+  _number: number;
   branch: string;
   change_id: string;
   subject: string;
   status: GerritChangeStatus;
   created: string;
-  submittable?: boolean;
-  _number: number;
+  /** Requires o=LABELS. */
+  reviewers?: {
```
viceice, line 53: can you avoid reorder please? I don't see a clear reason for new order.

viceice (commented):
first part.

is this still supporting Gerrit 3.0? we currently support all v3 versions. please update readme if something in this PR changes the minimum required Gerrit version.

lib/modules/platform/gerrit/utils.ts
felipecrs: Yes. The Gerrit REST API documentation is pretty clear, `status` is always set (because it is not marked as _optional_, and it can only have `NEW`, `MERGED`, or `ABANDONED`.

https://gerrit-review.googlesource.com/Documentation/rest-api-changes.html#change-info

If it ever happens in the future for some very unexpected reason, we better not mask it in my opinion.

lib/modules/platform/gerrit/types.ts
felipecrs: The new order puts the non-optional properties at the top, and the optional properties at the end.

Do you still want me to revert it?

felipecrs:
> is this still supporting Gerrit 3.0? we currently support all v3 versions. please update readme if something in this PR changes the minimum required Gerrit version.

Yes, nothing changes compatibility-wise. I have added TODOs in the code for improvements we can leverage from newer Gerrit versions, but the implementation remains as compatible as before.

lib/modules/platform/gerrit/types.ts
viceice: we prefer to sort properties alphabetical even if some are optional

felipecrs: Got it.

The new order is more alphabetical than before.

Should I restore the older order, sort everything alphabetically, or do nothing?

viceice: better restrore and do ordering in a followup pr. that would be much easier and faster to review

lib/modules/platform/gerrit/index.ts
```diff
@@ -316,23 +365,39 @@ export async function setBranchStatus(
   const labelValue =
     label && mapBranchStatusToLabel(branchStatusConfig.state, label);
   if (branchStatusConfig.context && labelValue) {
-    const pr = await getBranchPr(branchStatusConfig.branchName);
-    if (pr === null) {
+    const change = (
+      await client.findChanges(config.repository!, {
```
viceice, line 366: why fetching this new? isn't it already cached on pr?

lib/modules/platform/gerrit/index.ts
```diff
@@ -177,41 +191,75 @@ export async function createPr(prConfig: CreatePRConfig): Promise<Pr | null> {
...
-    .findChanges(config.repository!, { branchName: '' })
-    .then((res) => res.map((change) => mapGerritChangeToPr(change)));
+export async function getPrList(): Promise<Pr[]> {
+  const changes = await client.findChanges(config.repository!, {
+    branchName: '',
+    requestDetails: REQUEST_DETAILS_FOR_PRS,
+  });
+  return changes
+    .map((change) => mapGerritChangeToPr(change))
+    .filter((pr) => pr !== null);
```
viceice:
```suggestion
    .filter(isTruthy);
```

needs
```ts
import { isTruthy } from '@sindresorhus/is';
```

lib/modules/platform/gerrit/index.ts
```diff
@@ -177,41 +191,75 @@ export async function createPr(prConfig: CreatePRConfig): Promise<Pr | null> {
...
+  if (change === undefined) {
     throw new Error(
       `the change should be created automatically from previous push to refs/for/${prConfig.sourceBranch}`,
     );
   }
+  const createdMs = DateTime.fromISO(change.created.replace(' ', 'T'), {
+    zone: 'utc',
+  }).toMillis();
+  const fiveMinutesAgoMs = DateTime.utc().minus({ minutes: 5 }).toMillis();
+  if (createdMs < fiveMinutesAgoMs) {
```
viceice:
```suggestion
  });
  const fiveMinutesAgoMs = DateTime.utc().minus({ minutes: 5 });
  if (createdMs < fiveMinutesAgoMs) {
```

should also work 😉

lib/modules/platform/gerrit/index.spec.ts
```diff
@@ -247,7 +270,23 @@ describe('modules/platform/gerrit/index', () => {
     });
   });
 
-  describe('createPr() - error', () => {
+  describe('createPr()', () => {
+    const t0 = DateTime.fromISO('2025-04-14T16:33:37.000000000', {
+      zone: 'utc',
+    }) as DateTime<true>;
```
viceice:
```suggestion
    });
```

should not be needed

https://github.com/renovatebot/renovate/blob/3b6b00bf66e9134c0cfdc96a7d8793fbfdafb0ee/lib/util/package-rules/current-age.spec.ts#L8

lib/modules/platform/gerrit/index.spec.ts
```diff
@@ -247,7 +270,23 @@ describe('modules/platform/gerrit/index', () => {
     });
   });
 
-  describe('createPr() - error', () => {
+  describe('createPr()', () => {
+    const t0 = DateTime.fromISO('2025-04-14T16:33:37.000000000', {
```
viceice: do it for all tests once?

lib/modules/platform/gerrit/index.ts
felipecrs: Previously, this function was already fetching the change/PR from Gerrit through `getBranchPr()`.

It was doing that to decide whether a change/PR exists for this branch and whether to call `setLabel()` or not.

I changed from `getBranchPr` to `findChanges` so that the response includes an additional field `labels`, which I can use to decide better whether `setLabel` should be called or not.

lib/modules/platform/gerrit/index.spec.ts
felipecrs: do you mean for the whole index.spec.ts? Why would I do that if it's only needed by `createPr()`?

lib/modules/platform/gerrit/index.spec.ts
felipecrs: If I don't do that:

image

felipecrs: > https://github.com/renovatebot/renovate/blob/3b6b00bf66e9134c0cfdc96a7d8793fbfdafb0ee/lib/util/package-rules/current-age.spec.ts#L8

This example does not use `t0` within tests.

lib/modules/platform/gerrit/index.spec.ts
viceice: it's simpler and you don't need to call `vi.useRealTimers()`

viceice (commented):
see last open conversation

lib/modules/platform/gerrit/index.spec.ts
felipecrs: @viceice I think that happens because `DateTime.fromISO()` will not always returns an actual date, like `DateTime.fromISO("felipe")`.

And I believe this `as` assertion makes sense as we know it will always work.

lib/modules/platform/gerrit/index.spec.ts
felipecrs: Fair enough, it's indeed simpler.

felipecrs: FYI I am planning to refactor this logic to avoid depending on dates whatsoever, but I will keep that for a future PR as I already have too many enqueued.

lib/modules/platform/gerrit/index.spec.ts
```diff
@@ -247,7 +278,11 @@ describe('modules/platform/gerrit/index', () => {
     });
   });
 
-  describe('createPr() - error', () => {
+  describe('createPr()', () => {
+    beforeEach(() => {
+      vi.setSystemTime(t0.toMillis());
```
viceice: set it in before all where you set fake timers

felipecrs: I changed it, but that only works because the tests don't take 5 minutes to execute.

Not big deal anyway if I'll refactor it in the future.

viceice (approved):

secustor (approved):

PR #35668, opened 2025-05-01
feat(parser): support JSON5 parsing for URLs containing “.json5”

willswire:
Use `path.includes('.json5')` (instead of `endsWith`) so URLs like “…/renovate.json5/raw?ref=…” are correctly detected

## Changes

- Switch from `endsWith('.json5')` to `path.includes('.json5')`, so we catch URLs like `…/foo.json5/raw?ref=…`
- Retain the `.jsonc` check first via `path.includes('.jsonc')` for backwards compatibility

## Context

Some GitLab “raw” file URLs append a `/raw` segment and include query parameters (e.g. `?ref=0.6.0`), which prevented our old `endsWith('.json5')` check from matching. This change makes our JSON5 detection more robust and avoids warnings about invalid JSON in renovate’s parser when handling `.json5` files served via raw URLs.

### DIFF
--- a/lib/config/presets/gitlab/index.ts
+++ b/lib/config/presets/gitlab/index.ts
@@ -72,7 +72,7 @@ export async function fetchJSONFile(
  * @param fileName - The filename or path to extract from
  * @returns The extracted filename without path or query parameters
  */
-function extractFilenameFromGitLabPath(fileName: string): string {
+export function extractFilenameFromGitLabPath(fileName: string): string {
   try {
     const url = parseUrl(fileName);
     if (url) {
@@ -90,7 +90,7 @@ function extractFilenameFromGitLabPath(fileName: string): string {
       // Extract filename from the path
       return pathWithoutQuery.split('/').slice(-1)[0];
     }
-  } catch (error) {
+  } catch {
     // If parseUrl fails, it's not a URL, so return as is
   }
 
--- a/lib/config/presets/gitlab/index.ts
+++ b/lib/config/presets/gitlab/index.ts
@@ -73,25 +73,21 @@ export async function fetchJSONFile(
  * @returns The extracted filename without path or query parameters
  */
 export function extractFilenameFromGitLabPath(fileName: string): string {
-  try {
-    const url = parseUrl(fileName);
-    if (url) {
-      // If it's a valid URL, extract the pathname and process it
-      let pathWithoutQuery = url.pathname;
+  const url = parseUrl(fileName);
+  if (url) {
+    // If it's a valid URL, extract the pathname and process it
+    let pathWithoutQuery = url.pathname;
 
-      // Handle GitLab "/raw" suffix that's appended to the filename
-      if (pathWithoutQuery.endsWith('/raw')) {
-        pathWithoutQuery = pathWithoutQuery.substring(
-          0,
-          pathWithoutQuery.length - 4,
-        );
-      }
-
-      // Extract filename from the path
-      return pathWithoutQuery.split('/').slice(-1)[0];
+    // Handle GitLab "/raw" suffix that's appended to the filename
+    if (pathWithoutQuery.endsWith('/raw')) {
+      pathWithoutQuery = pathWithoutQuery.substring(
+        0,
+        pathWithoutQuery.length - 4,
+      );
     }
-  } catch {
-    // If parseUrl fails, it's not a URL, so return as is
+
+    // Extract filename from the path
+    return pathWithoutQuery.split('/').slice(-1)[0];
   }
 
   // Return as is - should be a normal filename with extension
--- a/lib/config/presets/gitlab/index.ts
+++ b/lib/config/presets/gitlab/index.ts
@@ -79,12 +79,7 @@ export function extractFilenameFromGitLabPath(fileName: string): string {
     let pathWithoutQuery = url.pathname;
 
     // Handle GitLab "/raw" suffix that's appended to the filename
-    if (pathWithoutQuery.endsWith('/raw')) {
-      pathWithoutQuery = pathWithoutQuery.substring(
-        0,
-        pathWithoutQuery.length - 4,
-      );
-    }
+ pathWithoutQuery= pathWithoutQuery.replace(regEx(/\/raw$/), '');
 
     // Extract filename from the path
     return pathWithoutQuery.split('/').pop();
--- a/lib/config/presets/gitlab/index.ts
+++ b/lib/config/presets/gitlab/index.ts
@@ -6,6 +6,7 @@ import type { GitlabProject } from '../../../types/platform/gitlab';
 import { GitlabHttp } from '../../../util/http/gitlab';
 import type { HttpResponse } from '../../../util/http/types';
 import { parseUrl } from '../../../util/url';
+import { regEx } from '../../../util/regex';
 import type { Preset, PresetConfig } from '../types';
 import { PRESET_DEP_NOT_FOUND, fetchPreset, parsePreset } from '../util';
 
@@ -79,7 +80,7 @@ export function extractFilenameFromGitLabPath(fileName: string): string {
     let pathWithoutQuery = url.pathname;
 
     // Handle GitLab "/raw" suffix that's appended to the filename
-    pathWithoutQuery = pathWithoutQuery.replace(/\/raw$/, '');
+    pathWithoutQuery = pathWithoutQuery.replace(regEx(/\/raw$/), '');
 
     // Extract filename from the path
     return pathWithoutQuery.split('/').pop() || '';
--- a/lib/config/presets/gitlab/index.ts
+++ b/lib/config/presets/gitlab/index.ts
@@ -83,7 +83,7 @@ export function extractFilenameFromGitLabPath(fileName: string): string {
     pathWithoutQuery = pathWithoutQuery.replace(regEx(/\/raw$/), '');
 
     // Extract filename from the path
-    return pathWithoutQuery.split('/').pop() || '';
+    return pathWithoutQuery.split('/').pop() ?? '';
   }
 
   // Return as is - should be a normal filename with extension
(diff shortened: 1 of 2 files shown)


lib/util/common.ts
```diff
@@ -87,11 +87,11 @@ export function parseJson(content: string | null, filename: string): JsonValue {
     return null;
   }
 
-  if (filename.endsWith('.jsonc')) {
+  if (filename.includes('.jsonc')) {
```
rarkins, line 90: I think this is too fuzzy/hacky, even if it should usually work. What's an example file name which doesn't currently work?

rarkins (changes requested):

lib/util/common.ts
willswire: When extending configuration files on a self-hosted GitLab instance, users must use the [Repository files API](https://docs.gitlab.com/api/repository_files/), which results in filenames like `https://gitlab.example.com/api/v4/projects/13083/repository/files/src%2Frenovate%2Fconfig%2Ejson5?ref=main`

Because the end of the URL is not the file extension, the `parseJsonWithFallback` method is called.

viceice: in this case we should parse the url in presets / caller and extract filename without query parameters. then pass cleaned filename instead of this fuzzyness

willswire: I've implemented the solution as suggested. Instead of using a fuzzy includes approach, I've made these changes:

1. Added a generic `getFilenameFromPath` utility in `lib/util/url.ts` that extracts filenames from paths and URLs by getting the part after the last slash and removing query parameters.

2. Implemented GitLab-specific filename extraction logic in the `fetchJSONFile` function in `lib/config/presets/gitlab/index.ts` that:
   - Removes query parameters
   - Handles the "/raw" suffix that GitLab adds
   - Properly decodes URL-encoded paths (like `src%2Frenovate%2Fconfig.json5` → `config.json5`)
   - Uses the generic utility as a fallback for regular paths

3. Added comprehensive tests to confirm this works with all GitLab URL formats.

This approach extracts the actual filename from GitLab repository API URLs and passes the clean filename to `parsePreset`, ensuring proper file format detection. It's more robust than the previous approach and avoids any "fuzzy" detection in the parser.

URLs like `https://gitlab.example.com/api/v4/projects/13083/repository/files/src%2Frenovate%2Fconfig%2Ejson5?ref=main` now correctly extract `config.json5` as the filename, so it uses the JSON5 parser instead of falling back to the JSON with fallback parser.

lib/util/url.ts
```diff
@@ -133,6 +133,24 @@ export function parseLinkHeader(
...
 }
 
+/**
+ * Extract the filename from a file path or URL.
+ * Works with paths that may have query parameters.
+ *
+ * @param pathOrUrl - The path or URL to extract the filename from
+ * @returns The extracted filename without query parameters
+ */
+export function getFilenameFromPath(pathOrUrl: string): string {
```
rarkins, line 143: We should try to use a URL parsing library instead of this

rarkins (changes requested):
Use an off the shelf url parser instead

lib/config/presets/gitlab/index.ts
```diff
@@ -52,7 +53,33 @@ export async function fetchJSONFile(
     throw new Error(PRESET_DEP_NOT_FOUND);
   }
 
-  return parsePreset(res.body, fileName);
+  // Extract the actual filename from the path without query parameters or URL encoding
+  let cleanFileName = fileName;
+
+  // First, remove any query parameters
+  cleanFileName = cleanFileName.split('?')[0];
```
viceice, line 57: Use `parseUrl` and extract `pathname`

lib/util/url.ts
```diff
@@ -133,6 +133,40 @@ export function parseLinkHeader(
...
+  try {
+    // Try to parse as URL first
+    const url = new URL(pathOrUrl);
```
viceice, line 146: use `parseUrl` which doesn't throw and return null instead

lib/util/url.spec.ts
```diff
@@ -10,220 +11,190 @@ import {
   parseUrl,
   replaceUrlPath,
   resolveBaseUrl,
+  trimLeadingSlash,
   trimSlashes,
   trimTrailingSlash,
 } from './url';
 
 describe('util/url', () => {
-  it.each`
```
viceice, line 18: Why do you remove all those tests? they are unrelated to your changes, so leave them as is. otherwise it's more complicated to see unwanted side effects from your changes.

RahulGautamSingh:
Checking my understading:

1. `filename` can be either the path to a file ie. `/path/to/file.txt` or some url like `https://gitlab.com/path/filename.txt`
2. We want to handle the case where, there's some query and an extra `/raw` present in the filename string.
3. The queries and `/raw` is generally only present in the URL like `filename`.

Based on the above, I think it will be simpler to do this:

1. First check if the filename is a URL or not.
2. If it is a URL, use the `parseUrl` utility fn to extract `pathname` from it (queries won't be present in `pathname`)
3. Use regex to remove `/raw` from the `pathname`

WDYT?? @willswire

lib/config/presets/gitlab/index.ts
```diff
@@ -52,7 +53,73 @@ export async function fetchJSONFile(
...
+  const cleanFileName = extractFilenameFromGitLabPath(fileName);
+
+  return parsePreset(res.body, cleanFileName);
+}
+
+/**
+ * Extracts the actual filename from a GitLab path, which could be:
+ * - A simple filename: "renovate.json5"
+ * - A path: "path/to/renovate.json5"
+ * - A URL-encoded path: "src%2Frenovate%2Fconfig.json5"
```
viceice, line 70: when this will happen?

lib/config/presets/gitlab/index.ts
```diff
@@ -52,7 +53,73 @@ export async function fetchJSONFile(
...
+ * - A GitLab API URL: "https://gitlab.example.com/api/v4/projects/13083/repository/files/renovate.json5?ref=main"
+ *
+ * @param fileName - The filename or path to extract from
+ * @returns The extracted filename without path or query parameters
+ */
+function extractFilenameFromGitLabPath(fileName: string): string {
+  // Fast path for simple filenames without path separators
+  if (
```
viceice: do:
- parseUrl(filename)
- if it returns a url -> use `pathname` and remove mentioned stuff like `/raw`
- if not -> return as is. it should be a normal filename with extension

lib/config/presets/gitlab/index.ts
```diff
@@ -52,7 +53,46 @@ export async function fetchJSONFile(
...
+      // Handle GitLab "/raw" suffix that's appended to the filename
+      if (pathWithoutQuery.endsWith('/raw')) {
+        pathWithoutQuery = pathWithoutQuery.substring(
+          0,
+          pathWithoutQuery.length - 4,
+        );
+      }
+
+      // Extract filename from the path
+      return getFilenameFromPath(pathWithoutQuery);
```
RahulGautamSingh: Is the util fn `getFilenameFromPath` still needed?

Can't we just split `pathWithoutQuery` at `/` and then `slice(-1)`?

willswire: correct; this is better since we're reducing the footprint of changes to only `lib/config/presets/gitlab/index.ts`

RahulGautamSingh:
Thanks! Please, add tests for the new urls/paths you introduced handling for

RahulGautamSingh (commented):
Please, fix the linting and coverage issues.

lib/config/presets/gitlab/index.ts
```diff
@@ -52,7 +53,42 @@ export async function fetchJSONFile(
...
+    // If it's a valid URL, extract the pathname and process it
+    let pathWithoutQuery = url.pathname;
+
+    // Handle GitLab "/raw" suffix that's appended to the filename
+    if (pathWithoutQuery.endsWith('/raw')) {
+      pathWithoutQuery = pathWithoutQuery.substring(
+        0,
+        pathWithoutQuery.length - 4,
+      );
+    }
```
RahulGautamSingh:
```suggestion
 pathWithoutQuery= pathWithoutQuery.replace(regEx(/\/raw$/), '');
```

lib/config/presets/gitlab/index.ts
```diff
@@ -52,7 +53,42 @@ export async function fetchJSONFile(
...
+
+    // Extract filename from the path
+    return pathWithoutQuery.split('/').slice(-1)[0];
```
RahulGautamSingh:
```suggestion
    return pathWithoutQuery.split('/').pop();
```
Easier to read

RahulGautamSingh (approved):
LGTM, except the formatting. Please run prettier locally and commit the changes.

RahulGautamSingh:
Wrap the regex in `regEx` please.

willswire:
> Wrap the regex in `regEx` please.

All set

rarkins (dismissed):

rarkins (dismissed):

willswire:
@rarkins any chance we can run this thru CI again?

lib/config/presets/gitlab/index.ts
viceice: ☝️ i expect an answer or a code change describing this case.

configuring `src%2Frenovate%2Fconfig.json5` as preset path isn't valid, so we shouldn't handle it at all.

viceice (changes requested):

lib/config/presets/gitlab/index.ts
willswire: @viceice when referencing files via the GitLab API, the file paths [must be URL encoded](https://docs.gitlab.com/api/rest/#file-path-branches-and-tags-name).

Example `renovate.json5` file:

```
{
  "extends": [
    // renovate: depName=gitlab-group/components/renovate
    "https://gitlab.example.com/api/v4/projects/1234/repository/files/src%2Frenovate.json5/raw?ref=0.9.2"
  ]
}
```

viceice: there is no test case for the full url which by the way doesn't use that code path

PR #35676, opened 2025-05-02
feat(gerrit): implement pr cache

felipecrs:
## Changes

This PR implements [PR cache](https://github.com/renovatebot/renovate/issues/33290) for the Gerrit platform, which delivers massive performance improvements.

PR cache in Gerrit works similarly to other platforms:

- Fetches all Renovate changes and stores them as `PR`s in the repo cache
- The repo cache may be [persisted](https://docs.renovatebot.com/self-hosted-configuration/#repositorycache) between runs
- When there's no cache, the initial sync fetches all Renovate changes through the Gerrit API, 100 changes per page
- When there's cache, the sync fetches changes until it hits a change whose `updated` timestamp matches what's already cached, 20 changes per page. **Gerrit API returns changes ordered by their update date**, meaning no further changes would have been modified.
- The PR cache is only synced once per execution
- When a change is created, updated, or merged, the PR cache for Gerrit is updated on the fly without requiring a re-sync.

This PR is a follow-up to the PR below, which completes the optimization cycle for Gerrit:

- https://github.com/renovatebot/renovate/pull/38966

[Here](https://github.com/renovatebot/renovate/pull/35676#issuecomment-4195881545)'s a comparative benchmark.

## Context

- [ ] This closes an existing Issue:

## AI assistance disclosure

I initially wrote this PR by myself, but then used Claude Opus 4.6 to rewrite it following the PR cache pattern from other platforms. It was also used to implement tests and to help review the code.

The public repository: https://review.gerrithub.io/q/project:felipecrs/renovate-repro-spring-framework


lib/modules/platform/gerrit/pr-cache.ts
```diff
@@ -0,0 +1,83 @@
...
+      continue;
+    }
+    if (limit === 1) {
+      return clone(pr);
+    }
+    matches.push(pr);
+  }
+  if (limit === 1) {
+    return null;
+  }
```
RahulGautamSingh: What does this do?

RahulGautamSingh (commented):
Thanks for this.

And, as you mentioned this could reuse some logic from the pr-cache implementation in other platforms.

lib/modules/platform/gerrit/pr-cache.ts
felipecrs: When `limit: 1`, it means we are searching for a single PR (only in Gerrit platform, see [GerritFindPRConfig](https://github.com/felipecrs/renovate/blob/e7bc8ac91bf90dd47cfa58003b10bd80c82c6a7e/lib/modules/platform/gerrit/types.ts#L5)).

This function was crafted so that when a single PR is being searched for, the return type is `Pr | null`.

If it reached this line, it means it didn't return earlier, which means no single PR was found.

felipecrs:
Yes, I surely plan to complete it.

felipecrs:
I plan to resume working on this next week.

felipecrs:
@RahulGautamSingh, I still plan to finish this. Currently I'm waiting for https://github.com/renovatebot/renovate/pull/39250 to be merged first. Would you mind reopening it, please?

felipecrs:
This PR is in a good, functional state (tested in against a few repositories).

But it shouldn't be reviewed as it currently depends on these other two to be merged first:

- https://github.com/renovatebot/renovate/pull/39250
- https://github.com/renovatebot/renovate/pull/38966

felipecrs:
Here is a recent performance comparison test with this PR (which also contains https://github.com/renovatebot/renovate/pull/38966):

|                      | Before this PR | **After this PR** |
| -------------------- | -------------- | ----------------- |
| Gerrit HTTP requests | 2443           | **8**             |
| Run duration         | 46m9s          | **2m46s**         |

https://github.com/user-attachments/assets/ecc52159-6c1a-4969-b611-75c5440e968c

<details><summary>Test details</summary>
<p>

- _Before this PR_ commit SHA: aa56c57a1ec570f26fe0a226f5dc6a4a58975806
- _After this PR_ commit SHA: bec063d540803822f4063bb2596a6f0b961870a9
- Repository: https://gerrithub.io/q/project:felipecrs/renovate-repro-spring-framework+status:open
- Disabled rate limits (concurrent PRs, hourly PRs, concurrent branches, hourly commits)
- Valkey pre-populated with (very) recent lookup data from data sources (e.g., Maven)
- HTTP requests against Gerrit throttled to 1 per second (to avoid overloading GerritHub)
- No prior repository cache (which would potentially save even more time and HTTP requests otherwise)
- All PRs/changes are up-to-date already (no actual work to do)
- Built Renovate before and after using `pnpm build && pnpm build:docker build`, and tagged with `unoptimized`/`optimized`
- Test command used in the video:
  ```console
  $ variant=optimized # or unoptimized
  $ time docker run --rm -t -u "$(id -u):$(id -g)" -v "$PWD:$PWD" -e RENOVATE_CONFIG_FILE="$PWD/config.js" --env-file=.env -e RENOVATE_REPORT_PATH="$PWD/renovate-$variant-$(date -u +"%Y-%m-%dT%H:%M:%SZ").json" -e LOG_FILE="$PWD/renovate-$variant-$(date -u +"%Y-%m-%dT%H:%M:%SZ").log" --link=valkey renovate/renovate:43.102.8-$variant-full
  ```

</p>
</details>

felipecrs:
Hey @RahulGautamSingh, could you please remove the _auto:inactive-pr_ label from this PR?

EDIT: Thank you!

RahulGautamSingh (commented):
Conflicted.

@felipecrs are you still working on it? If it will be ready in a while, you should draft it.

felipecrs:
@RahulGautamSingh this PR is still very relevant and I'm still maintaining it.

I have been daily driving it at my company for 500+ repositories for more than 3 months at this point.

I would super appreciate any help for getting it reviewed and merged, but since it depends on https://github.com/renovatebot/renovate/pull/38966, I'll mark it as draft for now.

Perhaps you can help me with https://github.com/renovatebot/renovate/pull/38966 instead?

lib/modules/platform/gerrit/index.ts
```diff
@@ -214,50 +223,97 @@ export async function initRepo({
...
+    await client.addMessage(
       prConfig.number,
-      prConfig.prBody,
+      prConfig.prBody!,
       TAG_PULL_REQUEST_BODY,
     );
+    pr.bodyStruct = {
+      hash: newBodyHash,
+    };
+    pr.updatedAt = new Date().toISOString();
```
RahulGautamSingh: Two problems here:
- Raw Date violates the Luxon rule in `best-practices.md`
- It produces `…000Z`, but everywhere else the cache stores Gerrit's format via `convertGerritDateToISO (…000000000, no Z)`. reconcile compares `updatedAt.replace('T',' ') === change.updated`, so a body-updated PR never matches on the next sync and its early-stop is skipped. It self-heals (the change just gets re-mapped), so not a correctness bug, but the format diverges. The abandon and merge branches already do the right thing using `change.updated` from the server response. `addMessage` returns void, so there's no server timestamp here, but at minimum use `DateTime` and keep the format consistent.

lib/modules/platform/gerrit/pr-cache.spec.ts
RahulGautamSingh, line 1: if a Renovate change exists on the server and isn't in the cache for any reason, a merge silently fails. Is this no-fallback behavior intended, or should these fall back to a direct lookup?

lib/modules/platform/gerrit/client.spec.ts
```diff
@@ -245,24 +245,6 @@ describe('modules/platform/gerrit/client', () => {
...
-        );
-      await expect(
-        client.findChanges('repo', {
-          branchName: 'dependency-xyz',
-          startOffset: 5,
-        }),
-      ).resolves.toEqual([gerritChange({ _number: 1 })]);
-    });
-
     it('sets query.S as 0 if startOffset is not provided', async () => {
```
RahulGautamSingh: 'sets query.S as 0 if startOffset is not provided' still names the removed startOffset. Rename to "on first page".

RahulGautamSingh (changes requested):

lib/modules/platform/gerrit/pr-cache.ts
```diff
@@ -0,0 +1,165 @@
...
+    this.cache = platformCache.gerrit.pullRequestsCache;
+    this.updateItems();
+  }
+
+  private static async init(repository: string): Promise<GerritPrCache> {
+    const res = new GerritPrCache(repository);
+    const isSynced = memCache.get<true | undefined>('gerrit-pr-cache-synced');
+
+    if (!isSynced) {
+      await res.sync();
```
viceice, line 61: this could run twice it init is called in parallel for some reason. use mutex or lazy

lib/modules/platform/gerrit/index.ts
felipecrs: Thank you. Got rid of that entirely via 3406750d74f154966eccb5813319054eef1542a6.

lib/modules/platform/gerrit/pr-cache.spec.ts
felipecrs: Well, it's not actually silent. It emits a warning.

And that situation should never happen, so I don't think implementing a fallback here is necessary.

lib/modules/platform/gerrit/client.spec.ts
felipecrs: Thank you. Done.

lib/modules/platform/gerrit/pr-cache.ts
felipecrs: @viceice, that is theoretically true, but given Renovate doesn't parallelize any platform call, this doesn't really happen in practice.

Also, this is following the pr-cache logic from all other platforms. All of them are subject to the same issue.

How about you log this as a future improvement instead? (not restricted to Gerrit)

RahulGautamSingh:
@felipecrs as a practice we do not review draft PRs. Can you mark it Ready for review, please?

felipecrs:
> @felipecrs as a practice we do not review draft PRs. Can you mark it Ready for review, please?

Oops, my bad. Forgot to mark is a ready. 😅

lib/modules/platform/gerrit/index.ts
felipecrs: Actually, I had to restore it. After testing this in production I realized that only Gerrit >=3.10 includes the update date in the response.

https://github.com/renovatebot/renovate/pull/35676/commits/5afb577098a4d6c756b987a09b217acc6f82488b

lib/modules/platform/gerrit/index.ts
```diff
@@ -305,50 +365,74 @@ export async function createPr(prConfig: CreatePRConfig): Promise<Pr | null> {
...
+export async function mergePr(mergeConfig: MergePRConfig): Promise<boolean> {
   logger.debug(
-    `mergePr(${config.id}, ${config.branchName!}, ${config.strategy!})`,
+    `mergePr(${mergeConfig.id}, ${mergeConfig.branchName!}, ${mergeConfig.strategy!})`,
   );
+  const prs = await GerritPrCache.getPrs(config.repository!);
+  const cached = prs.find((pr) => pr.number === mergeConfig.id);
+  if (!cached) {
+    logger.warn({ number: mergeConfig.id }, 'mergePr: PR not found in cache');
+    return false;
```
RahulGautamSingh: when the PR is missing from cache this returns false without calling submitChange, so a stale cache turns a real merge into a silent failure Renovate will misreport. previously this hit Gerrit directly. is cache always populated?

lib/modules/platform/gerrit/pr-cache.ts
```diff
@@ -0,0 +1,165 @@
...
+    let changeCount = 0;
+
+    for (const change of changes) {
+      const cachedPr = items[change._number];
+      const cachedUpdated = cachedPr?.updatedAt?.replace('T', ' ');
+
+      // If this change exists in cache and has the same updated timestamp,
+      // it hasn't been modified. Since changes are sorted by update time,
+      // all subsequent changes are also unchanged.
+      if (cachedUpdated === change.updated) {
```
RahulGautamSingh: this stops mid-page, unlike bitbucket-server which finishes reconciling the already-fetched page and only skips the next page. Gerrit's update-desc sort makes it safe in practice, but a never-cached older change on the same page gets skipped until the next run. consider matching bitbucket server's pattern

lib/modules/platform/gerrit/pr-cache.ts
```diff
@@ -0,0 +1,165 @@
...
+   * @param changes - Page of changes from Gerrit API
+   * @returns true if more pages can be fetched, false if not needed
+   */
+  private reconcile(changes: GerritChange[]): boolean {
+    const { items } = this.cache;
+    let changeCount = 0;
+
+    for (const change of changes) {
+      const cachedPr = items[change._number];
+      const cachedUpdated = cachedPr?.updatedAt?.replace('T', ' ');
```
RahulGautamSingh: this hand-reverses convertGerritDateToISO. prefer convertGerritDateToISO(change.updated) === cachedPr.updatedAt so the format logic stays in one place.

lib/modules/platform/gerrit/pr-cache.spec.ts
```diff
@@ -0,0 +1,259 @@
...
+  });
+
+  describe('updateItems() sorting edge cases', () => {
+    it('handles PRs without updatedAt', async () => {
+      clientMock.findChanges.mockResolvedValueOnce([]);
+      await GerritPrCache.getPrs('test/repo');
+
+      const pr1 = mapGerritChangeToPr(makeChange({ _number: 1 }))!;
+      const pr2 = mapGerritChangeToPr(makeChange({ _number: 2 }))!;
+      pr1.updatedAt = undefined as never;
```
RahulGautamSingh: `as never` forces undefined onto a required field. use partial<Pr>() instead.

lib/modules/platform/gerrit/pr-cache.ts
felipecrs: Yeah, this is a good idea. The page's data is already there anyway. 83cdf0bb80ec7ecc227a3b5a42e22e335c854351

lib/modules/platform/gerrit/pr-cache.spec.ts
felipecrs: Turns out I don't need `as never` since updatedAt is already optional in `Pr`.

lib/modules/platform/gerrit/pr-cache.ts
felipecrs: That is so much better. Thanks for catching it. 7833987ac23b5a755277287623b00cefcae9a35f

lib/modules/platform/gerrit/index.ts
felipecrs: My initial mindset was to avoid making requests to Gerrit, given our cache **should** be populated correctly.

But in this case we already have the PR/change number, which is a big indicator that the change exists, and if it's not in cache, it's an issue with the cache. It's harmless to self-heal in such case.

So, thanks for the feedback. Implemented via https://github.com/renovatebot/renovate/pull/35676/commits/a2a72a91bb370b3edc8e7cf882d6e94b0724ecf4.

This aligns the implementation with forgejo/gitea as well.

lib/modules/platform/gerrit/index.ts
```diff
@@ -305,50 +390,68 @@ export async function createPr(prConfig: CreatePRConfig): Promise<Pr | null> {
...
     prConfig.prBody,
     TAG_PULL_REQUEST_BODY,
   );
-  return mapGerritChangeToPr(change, {
+
+  const pr = mapGerritChangeToPr(change, {
     sourceBranch: prConfig.sourceBranch,
     prBody: prConfig.prBody,
-  });
+  })!;
```
RahulGautamSingh, line 398: the `!` hides that `mapGerritChangeToPr` can return `null` when `sourceBranch` is missing. check the result and throw a clear error instead of asserting non-null.

lib/modules/platform/gerrit/index.ts
```diff
@@ -214,50 +223,126 @@ export async function initRepo({
...
+      return false;
+    }
+    if (findPRConfig.prTitle && pr.title !== findPRConfig.prTitle) {
+      return false;
+    }
+    if (
+      !isUndefined(findPRConfig.state) &&
+      findPRConfig.state !== 'all' &&
+      pr.state !== findPRConfig.state &&
+      !(findPRConfig.state === '!open' && pr.state !== 'open')
```
RahulGautamSingh: this state-matching condition is hard to read, especially the `!open` double negative. extract a `matchesState(pr.state, findPRConfig.state)` helper with early returns.

lib/modules/platform/gerrit/index.ts
felipecrs: It's impossible that `sourceBranch` is missing since we pass it as parameter. Checking and throwing the error here would just make the code more complex.

Please let me know if you'd like me to implement it anyway.

lib/modules/platform/gerrit/index.ts
felipecrs: Done via https://github.com/renovatebot/renovate/pull/35676/commits/4d8a9a5646ebfc1a1e2654bbe5620620f993cadb.

lib/modules/platform/gerrit/index.ts
RahulGautamSingh: Yes, I got that it will be impossible since you explicitly pass it. Its okay to ignore this.

RahulGautamSingh (approved):

PR #35679, opened 2025-05-02
refactor(config): set and delete keys early

rarkins:
## Changes

Refactors config parsing to set and delete keys early, so they're not needed in GlobalConfig.

### DIFF
--- a/lib/config/decrypt/legacy.spec.ts
+++ b/lib/config/decrypt/legacy.spec.ts
@@ -1,5 +1,5 @@
 import { CONFIG_VALIDATION } from '../../constants/error-messages';
-import { decryptConfig } from '../decrypt';
+import { decryptConfig, setPrivateKeys } from '../decrypt';
 import { GlobalConfig } from '../global';
 import type { RenovateConfig } from '../types';
 import { Fixtures } from '~test/fixtures';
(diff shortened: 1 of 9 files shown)


lib/config/global.ts
```diff
@@ -27,8 +27,7 @@ export class GlobalConfig {
     'localDir',
     'migratePresets',
     'presetCachePersistence',
-    'privateKey',
-    'privateKeyOld',
+
```
viceice, line 30:
```suggestion
```

viceice (commented):
just a small nit

lib/config/decrypt/legacy.spec.ts
```diff
@@ -14,28 +14,26 @@ describe('config/decrypt/legacy', () => {
...
         CONFIG_VALIDATION,
       );
     });
 
     it('replaces npm token placeholder in npmrc', async () => {
-      GlobalConfig.set({
-        privateKey: 'invalid-key',
-        privateKeyOld: privateKey,
-      }); // test old key failover
+      setPrivateKeys('invalid-key', privateKey);
```
rarkins, line 36:
```suggestion
      setPrivateKeys('invalid-key', privateKey); // test old key failover
```

viceice (approved):

PR #35761, opened 2025-05-06
feat(versioning): Add Unity packages

bdovaz:
## Context

https://github.com/renovatebot/renovate/pull/35298#pullrequestreview-2809757922


lib/modules/versioning/unity3d-packages/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    const { major, minor, patch, label } = matches.groups;
+
+    const release = [
+      parseInt(major, 10),
+      parseInt(minor, 10),
+      parseInt(patch, 10),
+    ];
+    const isStable = !Unity3dPackagesVersioningApi.unstableRegex.test(label);
+
+    return { release, prerelease: isStable ? undefined : label, suffix: label };
```
viceice, line 35:
```suggestion
    return { release, prerelease: isStable ? undefined : label, suffix: label };
```
do not use prerelease and suffix

lib/modules/versioning/unity3d-packages/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  protected override _compareOther(
+    _left: GenericVersion,
+    _right: GenericVersion,
+  ): number {
+    if (is.nonEmptyString(_left.suffix) && is.nonEmptyString(_right.suffix)) {
+      return _left.suffix.localeCompare(_right.suffix);
```
viceice, line 43: add numeric true, so `pre.2` < `.pre.10`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#numeric_sorting

lib/modules/versioning/unity3d-packages/index.ts
bdovaz: I don't understand, do you mean don't use both at the same time or don't use either one?

I need the `suffix` for the `_compareOther` method.

And why can't I use `prerelease` even if I don't use it directly? Aren't we supposed to try to fill in as much information as possible so that it can be useful to whoever wants to extend the functionality with custom managers, etc?

bdovaz: @viceice can you answer? Thanks

viceice: they should not have same values. use pre release if the version is unstable. use suffix for some stable compatibility string.

so best is to drop suffix entirely.

viceice (commented):
See comment

bdovaz:
> See comment

Ready!

viceice (commented):
otherwise LGTM

lib/modules/versioning/unity3d-packages/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    const { major, minor, patch, label } = matches.groups;
+
+    const release = [
+      parseInt(major, 10),
+      parseInt(minor, 10),
+      parseInt(patch, 10),
+    ];
+    const isStable = !Unity3dPackagesVersioningApi.unstableRegex.test(label);
+
+    return { release, prerelease: isStable ? undefined : label };
```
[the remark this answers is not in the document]
bdovaz: When you commented in https://github.com/renovatebot/renovate/pull/35761#discussion_r2075575221 not to use prerelease / suffix at the same time, this problem was generated.

I had both at the same time in case there were labels that were not of type prerelease (exp.|pre.|preview.) so that in that case prerelease would be `false` but it would interpret that label and order the versions based on it.

Is it ok to use the suffix field again at the same time? Because I need the string of the label to be able to contemplate that `1.2.3-4` is greater than `1.2.3` because as I say, that `-4` is not prerelease, it is simply a label.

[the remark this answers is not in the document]
viceice: if the label is always `-<number>` then you can simply add the number to the `release` array. then the base class can handle it as forth part of the version.

viceice (commented):
see comment

lib/modules/versioning/unity3d-packages/index.ts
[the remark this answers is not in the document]
bdovaz: Is not always a number, as I added in tests or documentation. It can be also -preview.{number}, -pre.{number}, ...

https://github.com/renovatebot/renovate/blob/b400a218269085a9584973330ab98f7f48a43aa1/lib/modules/versioning/unity3d-packages/index.spec.ts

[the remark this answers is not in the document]
bdovaz: @viceice can you give me feedback please? As far as I am concerned, everything is ready

bdovaz:
On my side everything is ready and I am waiting for a new review.

bdovaz:
On my side everything is ready and I am waiting for a new review.

I am waiting for it to be merged as soon as possible so I can continue working on (depends on this PR): https://github.com/renovatebot/renovate/pull/35298

rarkins (approved):

PR #35808, opened 2025-05-08
feat(gitlab): add gitLabNotesMerge option to support merge trains

### LINKED ISSUE #5573: Support GitLab "merge trains"

Preisschild:
- closes https://github.com/renovatebot/renovate/issues/5573

- prior PR: https://github.com/renovatebot/renovate/pull/34102

Gitlab's /merge API does not support merge trains, but sending "/merge" over the MR comment (notes) API does work:
https://gitlab.com/gitlab-org/gitlab/-/issues/32665

This introduces a configuration option to use this API as a workaround until Gitlab introduces support to the merge API for Merge Trains.

## Changes

When using automerge with Gitlab and `gitLabNotesMerge` is enabled, renovate will use the `/merge` quick reactions api, which supports adding MRs to merge trains. 
Gitlab Docs:

- https://docs.gitlab.com/api/notes/#create-new-merge-request-note
- https://docs.gitlab.com/user/project/quick_actions/

This can be understood of as [a workaround](https://gitlab.com/gitlab-org/gitlab/-/issues/32665#note_974671680) until Gitlab implements support for Merge Trains in the [merge API](https://gitlab.com/gitlab-org/gitlab/-/issues/32665). This option can then be removed/deprecated.

## Context

- https://github.com/renovatebot/renovate/issues/5573

### DIFF
--- a/lib/modules/platform/gitlab/index.ts
+++ b/lib/modules/platform/gitlab/index.ts
@@ -596,6 +596,8 @@ async function tryPrAutomerge(
         250,
       );
 
+      let diffHeadSha = '';
+
       // Check for correct merge request status before setting `merge_when_pipeline_succeeds` to  `true`.
       for (let attempt = 1; attempt <= retryTimes; attempt += 1) {
         const { body } = await gitlabApi.getJsonUnchecked<{
@@ -604,9 +606,13 @@ async function tryPrAutomerge(
           pipeline: {
             status: string;
           };
+          sha: string;
         }>(`projects/${config.repository}/merge_requests/${pr}`, {
           memCache: false,
         });
+
+        diffHeadSha = body.sha;
+
         // detailed_merge_status is available with Gitlab >=15.6.0
         const use_detailed_merge_status = !!body.detailed_merge_status;
         const detailed_merge_status_check =
@@ -620,12 +626,31 @@ async function tryPrAutomerge(
         if (
           (detailed_merge_status_check || deprecated_merge_status_check) &&
           body.pipeline !== null &&
+          diffHeadSha !== '' &&
           desiredPipelineStatus.includes(body.pipeline.status)
         ) {
           break;
         }
         logger.debug(`PR not yet in mergeable state. Retrying ${attempt}`);
         await setTimeout(mergeDelay * attempt ** 2); // exponential backoff
+      }
+
+      if (platformPrOptions?.gitLabNotesMerge) {
+        // Use the /merge "quick action" to automerge, which will work for
+        // both merge trains or non-train merges
+        // Associated docs:
+        //  - https://docs.gitlab.com/api/notes/#create-new-merge-request-note
+        //  - https://docs.gitlab.com/user/project/quick_actions/
+        await gitlabApi.postJson(
+          `projects/${config.repository}/merge_requests/${pr}/notes`,
+          {
+            body: {
+              body: '/merge',
+              merge_request_diff_head_sha: diffHeadSha,
+            },
+          },
+        );
+        return;
       }
 
       // Even if Gitlab returns a "merge-able" merge request status, enabling auto-merge sometimes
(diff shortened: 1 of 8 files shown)


marc-jasinski:
Hi @Preisschild 
Dropping information you may find helpful for this PR in the context of final workaround implementation.
Merge trains have their own API for adding a merge request to the train: https://docs.gitlab.com/api/merge_trains/#add-a-merge-request-to-a-merge-train. If a project has merge trains configured, it seems like calls could just be updated to use that API.

marc-jasinski:
> Ideally we'd just use the `/merge` API and gitlab adds to the train if merge train support is enabled and if not uses the other merge strategy. This seems to be the plan anyways, but gitlab hasn't implemented this yet.
From what I learnt from Gitlab this week is that they're not planning working on extending that functionality in their merge API in this year.
They've passed me however that info about possibility of using this merge-train API https://docs.gitlab.com/api/merge_trains/#add-a-merge-request-to-a-merge-train hence I thought to drop it here.

lib/modules/platform/gitlab/index.ts
```diff
@@ -688,27 +694,44 @@ async function tryPrAutomerge(
...
             },
-          );
-          break;
-        } catch (err) {
-          logger.debug(
-            { err },
-            `Automerge on PR creation failed. Retrying ${attempt}`,
-          );
+          },
+        );
```
rarkins: I think you could return here and save the "else" below

rarkins (changes requested):

lib/modules/platform/gitlab/index.ts
Preisschild: done

rarkins (dismissed):

BoomchainLabs (approved):

lib/modules/platform/gitlab/index.ts
```diff
@@ -688,6 +694,24 @@ async function tryPrAutomerge(
...
+        // both merge trains or non-train merges
+        // Associated docs:
+        //  - https://docs.gitlab.com/api/notes/#create-new-merge-request-note
+        //  - https://docs.gitlab.com/user/project/quick_actions/
+        await gitlabApi.postJson(
+          `projects/${config.repository}/merge_requests/${pr}/notes`,
+          {
+            body: {
+              body: '/merge',
+              merge_request_diff_head_sha: diffHeadSha,
```
viceice, line 649: `diffHeadSha` can be emptyif all attempts are failing. you need to handle that case

viceice (commented):
otherwise lgtm

lib/modules/platform/gitlab/index.ts
Preisschild: wouldn't it just then be set to an empty string? This would result in the /notes POST call failing anyways.

Should I log an error and return false before that if diffHeadSha is empty?

Preisschild: Ok, I've added an empty string check in the `if { break }` statement

viceice: it can still be an empty string if all attempts above are failing.

what happens when `diffHeadSha=''`?

viceice (commented):
see comment

lib/modules/platform/gitlab/index.ts
Preisschild: Ah it seems I misunderstood the logic then. 

The API responds 202 with an error message that the sha is missing and then just `return`s

Preisschild: Would this make sense? (I'm a typescript noob, sorry)
```
      if (diffHeadSha === '') {
        throw new Error('Diff head SHA of the merge request is not set');
      }
```

PR #35815, opened 2025-05-08
refactor: Make package caching params explicit

zharinov:
## Changes

Make options explicit in order to avoid confusion when using package cache provider class.


rarkins (changes requested):
Needs explaining / documenting

zharinov:
If we enable all the private package checks (which is default), it will break all users (NPM and Maven).
The reality is, each of the two checks is specific to the corresponding datasource.
This PR makes particular datasource enable their own check and explicitly disable every other.
I also decided to make them non-nullable in order to be explicit and double-check when some new check mechanism is eventually added.

rarkins (approved):

PR #35844, opened 2025-05-09
fix(manager/docker-compose): update images used as anchors

### LINKED ISSUE #34369: Renovate doesn't update docker-compose with yaml anchors anymore
### Discussed in https://github.com/renovatebot/renovate/discussions/32918

<sup>Originally posted by **mueller-ma** December  5, 2024</sup>

A Mend.io-hosted app

_No response_

In the past Renovate had no problems when a Docker compose file had yaml anchors. Now it doesn't extract the image anymore:
- https://github.com/mueller-ma/renovate-reproduce/blob/main/docker-compose.yml
- https://github.com/mueller-ma/renovate-reproduce/issues/2

`docker compose` can use and parse this docker-compose file. You can see that docker includes the `image` in both services:

````
$ docker compose -f docker-compose.yml config
name: tmp
services:
  service-a:
    environment:
      SERVICE: a
    image: debian:11
    networks:
      default: null
  service-b:
    environment:
      SERVICE: b
    image: debian:11
    networks:
      default: null
networks:
  default:
    name: tmp_default
x-shared_setting:
  image: debian:11
````

### Logs (if relevant)

```

DEBUG: packageFiles with updates
{
  "baseBranch": "main"
  "config": {
    "docker-compose": [
      {
        "deps": [],
        "packageFile": "docker-compose.yml"
      }
    ]
  }
}

```

</details>
</div>

Reproduction: renovate-reproductions/32918

### PR DESCRIPTION
**Pull request by RahulGautamSingh**
## Changes
- Extract the images which are used as anchors from the `extensions`

## Context
- Closes: https://github.com/renovatebot/renovate/issues/34369

- [x] Both unit tests + ran on a real repository https://github.com/Rahul-renovate-testing/repro-34369/pulls

### DIFF
--- a/lib/modules/manager/docker-compose/schema.ts
+++ b/lib/modules/manager/docker-compose/schema.ts
@@ -24,7 +24,9 @@ const DockerComposeFileModern = z
     version: z.string().optional(),
     services: z.record(DockerComposeService),
   })
-  .catchall(z.unknown()) // allow unknown fields
+  // using catchall to capture fields starting with `x-` and collecting them in `extensions` field
+  // might need to replace this with something better once zodv4 is stable
+  .catchall(z.unknown())
   .transform(
     (
       obj,
--- a/lib/modules/manager/docker-compose/schema.ts
+++ b/lib/modules/manager/docker-compose/schema.ts
@@ -29,13 +29,11 @@ const DockerComposeFileModern = z
     ): {
       version?: string;
       services: Record<string, z.infer<typeof DockerComposeService>>;
-      extensions?: Record<string, z.infer<typeof DockerComposeService>>;
+      extensions?: Record<string, { image: string }>;
     } => {
       const { version, services, ...rest } = obj;
 
-      let extensions:
-        | Record<string, z.infer<typeof DockerComposeService>>
-        | undefined = undefined;
+      let extensions: Record<string, { image: string }> | undefined;
 
       // collect extensions which have image field
       // https://docs.docker.com/reference/compose-file/extensions/
@@ -44,7 +42,7 @@ const DockerComposeFileModern = z
           const value = rest[key];
           if (is.object(value) && 'image' in value && is.string(value.image)) {
             extensions ??= {};
-            extensions[key] = value as z.infer<typeof DockerComposeService>;
+            extensions[key] = { image: value.image };
           }
         }
       }
--- a/lib/modules/manager/docker-compose/schema.ts
+++ b/lib/modules/manager/docker-compose/schema.ts
@@ -25,7 +25,7 @@ const DockerComposeFileModern = z
     services: z.record(DockerComposeService),
   })
   // using catchall to capture fields starting with `x-` and collecting them in `extensions` field
-  // might need to replace this with something better once zodv4 is stable
+  // might need to replace this with something better once zod v4 is stable
   .catchall(z.unknown())
   .transform(
     (
--- a/lib/modules/manager/docker-compose/schema.ts
+++ b/lib/modules/manager/docker-compose/schema.ts
@@ -38,6 +38,8 @@ const DockerComposeFileModern = z
         | undefined = undefined;
 
       // collect fragments which have image field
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/modules/manager/docker-compose/schema.ts [hunk 1]
```diff
@@ -14,13 +15,48 @@ const DockerComposeService = z.object({
...
 const DockerComposeFileV1 = z.record(DockerComposeService);
-const DockerComposeFileModern = z.object({
-  // compose does not use this strictly, so we shouldn't be either
-  // https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element
-  version: z.string().optional(),
-  services: z.record(DockerComposeService),
-});
+const DockerComposeFileModern = z
+  .object({
+    // compose does not use this strictly, so we shouldn't be either
```
viceice, line 20: use tsdoc style comment

lib/modules/manager/docker-compose/schema.ts
```diff
@@ -14,13 +15,48 @@ const DockerComposeService = z.object({
...
+    // https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element
+    version: z.string().optional(),
+    services: z.record(DockerComposeService),
+  })
+  .catchall(z.unknown()) // allow unknown fields
```
viceice, line 25: i don't like that. why not simply add extensions to the root schema?

RahulGautamSingh: `extensions` is a name I have given to collect all of them in one object
They can have any name as long as they `begin` with `x-` prefix hence we cannot use them directly as field.

viceice: can we define the root schema like:
```ts
type DExtensionKey = `x-${string}`
const isValidDKey = (key: string): key is DExtensionKey => key.startsWith("x-")
const DCFile = z.record(z.string().refine(isValidDKey)).extend(
  z.object({
    version: z.string().optional(),
    services: z.record(DockerComposeService),
  })
);
```

viceice: - https://github.com/colinhacks/zod/issues/2195

viceice: ☝️

viceice (commented):
see open conversation

lib/modules/manager/docker-compose/schema.ts
RahulGautamSingh: It asks to use `extend` with `record` which is not correct as per the Zod documentation. And, the suggested approach as per their AI is to use `catchall` as well

viceice: maybe simply add a filter to only allow keys with `x-` 🤔 .

if not possible add a comment to the code that we need a newer (v4?) zod to make this better.

viceice (commented):
see comment

lib/modules/manager/docker-compose/schema.ts
secustor: Could we passthrough the on the root level and then simply loop over they object keys?

RahulGautamSingh: Kinda doing something with the `catchall` in the current implemenation

lib/modules/manager/docker-compose/schema.ts
```diff
@@ -14,13 +15,52 @@ const DockerComposeService = z.object({
...
+  .object({
+    /**
+     *  compose does not use this strictly, so we shouldn't be either
+     *  https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element
+     */
+    version: z.string().optional(),
+    services: z.record(DockerComposeService),
+  })
+  // using catchall to capture fields starting with `x-` and collecting them in `extensions` field
+  // might need to replace this with something better once zodv4 is stable
```
viceice:
```suggestion
  // might need to replace this with something better once zod v4 is stable
```

viceice (approved):

PR #35847, opened 2025-05-10
fix(helmv3): helm charts with lockfiles never gets automerged

felipecrs:
## Changes

This fixes the issue with the Chart.lock comparison which always considers the lockfile as changed, and thus always pushes a new update to the PR, which causes the PR to be never automerged (since automerged is attempted after processing the lockfile).

## Context

I was investigating why some of my dependencies were never automerged despite no apparent issue, until I found this was why.

When calling `helm dependency update` on a chart that doesn't already have a `Chart.lock` (Renovate's case), it will create a new `Chart.lock` file with a `generated` key with a timestamp.

Note that `helm dependency update` doesn't change the `generated` key if the `Chart.lock` file already exists (and if there are no further changes), but this is not applicable to Renovate since `helmv3` performs `helm dependency update` in an isolated directory, and therefore there will never be a `Chart.lock` file to begin with.


felipecrs:
What's with this snapshot check thing? What should I do to fix it?

lib/modules/manager/helmv3/__fixtures__/oci_1_generated_changed.lock
```diff
@@ -0,0 +1,12 @@
+dependencies:
+- name: renovate-test
```
viceice, line 2: please inline and only use one dep

lib/modules/manager/helmv3/__snapshots__/artifacts.spec.ts.snap
```diff
@@ -306,6 +306,55 @@ exports[`modules/manager/helmv3/artifacts > log into private registries and repo
 ]
 `;
 
+exports[`modules/manager/helmv3/artifacts > returns null if only "generated" is changed 1`] = `
```
viceice, line 309: no snapshots, use toEqual ot toMatchObject

lib/modules/manager/helmv3/artifacts.ts
```diff
@@ -218,3 +221,12 @@ export async function updateArtifacts({
     ];
   }
 }
+
+function isHelmLockChanged(oldContent: string, newContent: string): boolean {
+  const regex = regEx(/^generated: ".+"$/gm);
```
viceice, line 226:
```suggestion
  const regex = regEx(/^generated: ".+"$/m);
```
i don't think you need the global flag, there is only one occurrence of it.

lib/modules/manager/helmv3/artifacts.spec.ts
```diff
@@ -111,6 +114,27 @@ describe('modules/manager/helmv3/artifacts', () => {
     expect(execSnapshots).toMatchSnapshot();
   });
 
+  it('returns null if only "generated" is changed', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(ociLockFile1 as any);
```
viceice, line 118:
```suggestion
    fs.readLocalFile.mockResolvedValueOnce(ociLockFile1);
```

lib/modules/manager/helmv3/artifacts.spec.ts
```diff
@@ -111,6 +114,27 @@ describe('modules/manager/helmv3/artifacts', () => {
...
+    fs.getSiblingFileName.mockReturnValueOnce('Chart.lock');
+    const execSnapshots = mockExecAll();
+    fs.readLocalFile.mockResolvedValueOnce(ociLockFile1GeneratedChanged as any);
```
viceice, line 121:
```suggestion
    fs.readLocalFile.mockResolvedValueOnce(ociLockFile1GeneratedChanged);
```

felipecrs:
@viceice I think I fixed everything as you wanted, please take a look again.

lib/modules/manager/helmv3/artifacts.spec.ts
```diff
@@ -111,6 +112,45 @@ describe('modules/manager/helmv3/artifacts', () => {
     expect(execSnapshots).toMatchSnapshot();
   });
 
+  it('returns null if only "generated" is changed', async () => {
+    fs.readLocalFile.mockResolvedValueOnce(stripIndent`
```
viceice, line 116:
```suggestion
    fs.readLocalFile.mockResolvedValueOnce(codeBlock`
```

we prefer `codeBlock`

lib/modules/manager/helmv3/artifacts.spec.ts
```diff
@@ -1,6 +1,7 @@
 import type { GetAuthorizationTokenCommandOutput } from '@aws-sdk/client-ecr';
 import { ECRClient, GetAuthorizationTokenCommand } from '@aws-sdk/client-ecr';
 import { mockClient } from 'aws-sdk-client-mock';
+import { stripIndent } from 'common-tags';
```
viceice, line 4:
```suggestion
import { codeBlock } from 'common-tags';
```

lib/modules/manager/helmv3/artifacts.spec.ts
```diff
@@ -111,6 +112,45 @@ describe('modules/manager/helmv3/artifacts', () => {
...
+      dependencies:
+      - name: renovate-test
+        repository: oci://registry.gitlab.com/user/oci-helm-test
+        version: 0.1.0
+      digest: sha256:886f204516ea48785fe615d22071d742f7fb0d6519ed3cd274f4ec0978d8b82b
+      generated: "2022-01-20T17:48:47.610371241+01:00"
+      `);
+    fs.getSiblingFileName.mockReturnValueOnce('Chart.lock');
+    const execMocks = mockExecAll();
+    fs.readLocalFile.mockResolvedValueOnce(stripIndent`
```
viceice, line 126:
```suggestion
    fs.readLocalFile.mockResolvedValueOnce(codeBlock`
```

lib/modules/manager/helmv3/artifacts.ts
```diff
@@ -152,7 +153,9 @@ export async function updateArtifacts({
 
     if (is.truthy(existingLockFileContent)) {
       const newHelmLockContent = await readLocalFile(lockFileName, 'utf8');
-      const isLockFileChanged = existingLockFileContent !== newHelmLockContent;
+      const isLockFileChanged =
+        !is.truthy(newHelmLockContent) ||
```
viceice, line 157:
```suggestion
        !is.string(newHelmLockContent) ||
```

lib/modules/manager/helmv3/artifacts.ts
```diff
@@ -218,3 +221,12 @@ export async function updateArtifacts({
     ];
   }
 }
+
+function isHelmLockChanged(oldContent: string, newContent: string): boolean {
+  const regex = regEx(/^generated: ".+"$/m);
+  const replacement = 'generated: "stubbed"';
+  return (
+    newContent.replace(regex, replacement) !==
+    oldContent.replace(regex, replacement)
```
viceice, line 230:
```suggestion
    newContent.replace(regex, '') !==
    oldContent.replace(regex, '')
```
just replace with empty string

felipecrs:
@viceice, done.

viceice (approved):

PR #35856, opened 2025-05-11
fix(versioning/npm): expand replace ranges support

RahulGautamSingh:
## Changes
Add support to handle `greaterThan` and `lessThan` ranges for `replace` range strategy ie.e `>= 1.2.0 < 2.0.0`
- Previously, the range changed to: `< {newVersion}`
- Now, it changes to: `>= {newVersion} < {upperLimit}`

`upperLimit` is decided based on existing range. If the existing range:
i) spans across the same major, the `upperLimit` will be `newVersion` incremented to next minor
ii) else, `uppperLimit` will be `newVersion` incremented by to major

Example 1:

existingRange => `>= 1.0.0 < 1.5.0` (spanning across same major)

if newVersion => `1.6.0` ... upperLimit => `1.7.0`
if newVersion => `2.0.0` ... upperLimit => `2.1.0`

Example 2:

existingRange => `>= 1.0.0 < 2.0.0` (spanning beyond major)

if newVersion => `2.0.0` ... upperLimit => `3.0.0`
if newVersion => `3.0.0` ... upperLimit => `4.0.0`


RahulGautamSingh:
Drafting it cause, I forgot to anticipate the effects on other versionings which use `npm` versioning. Hence, they broke.

PR #35866, opened 2025-05-11
feat: Add `abandonmentThreshold` and detect `isAbandoned` flag

### LINKED ISSUE #34128: Flag packages as abandoned in lookup results

For each package we look up, we should calculate the most recent release date of all releases in the result  (where `releaseTimestamp` is present). This should not be "the release date of the release flagged as latest" - instead it should be the latest/most recent release timestamp of _all_ releases in the package, including beta/non-latest releases. This field could be stored as `lastReleaseTimestamp` at the package level. It should be calculated and stored _before_ filtering such as `allowedVersions` is applied. i.e. in the datasource/index processing before passing back to the worker lookup process.

In addition, a new configuration option should be added: `abandonmentThreshold`, which supports the same date range syntax as `minimumReleaseAge`. It should default to `null` (no threshold) but if it _is_ defined then it should be used to calculate a new `isAbandoned` boolean field to be added to packages, based on the `lastReleaseTimestamp` value calculated earlier. The field `isAbandoned` should be present (true or false) only if both `lastReleaseTimestamp` and `abandonmentThreshold` are defined for a package - otherwise undefined. Such logic could be done in the worker and is not essential in the datasource itself.

zharinov:
## Context

- Closes: #34128
- Ref: #34127


docs/usage/configuration-options.md
```diff
@@ -49,6 +49,43 @@ When an array or object configuration option is `mergeable`, it means that value
...
 ---
 
+## abandonmentThreshold
+
+The `abandonmentThreshold` option allows Renovate to flag packages as abandoned when they haven't received updates for a specified period of time.
+
+Renovate adds an `isAbandoned` boolean property to the package lookup result when:
+
+- `abandonmentThreshold` is defined (not `null`)
+- The package has a `bumpedAt` timestamp available
```
rarkins, line 59: This needs explanation

rarkins (changes requested):

rarkins (dismissed):

lib/config/options/index.ts
```diff
@@ -1890,6 +1890,13 @@ const options: RenovateOptions[] = [
     type: 'string',
     default: null,
   },
+  {
+    name: 'abandonmentThreshold',
+    description:
+      'The package will be marked as abandoned if it has not been updated for this duration.',
```
HonkingGoose, line 1896:
```suggestion
      'Flags packages that have not been updated within this period as abandoned.',
```

How about something like this?

docs/usage/configuration-options.md
```diff
@@ -49,6 +49,43 @@ When an array or object configuration option is `mergeable`, it means that value
 
 ---
 
+## abandonmentThreshold
```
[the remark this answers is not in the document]
rarkins: Actually I've received positive feedback on the term "abandoned" (neglected) as a complement to "deprecated" (officially declared).

lib/workers/repository/process/lookup/abandonment.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    });
+
+    beforeEach(() => {
+      vi.setSystemTime(mockTime);
+    });
+
+    afterAll(() => {
+      vi.useRealTimers();
+    });
+
```
viceice, line 23:
```suggestion
```
no need, doesn't influence other tests

lib/workers/repository/process/lookup/abandonment.ts
```diff
@@ -0,0 +1,32 @@
...
+  releaseResult: ReleaseResult,
+  config: LookupUpdateConfig,
+): ReleaseResult {
+  const { abandonmentThreshold } = config;
+  if (!abandonmentThreshold) {
+    return releaseResult;
+  }
+
+  const abandonmentThresholdMs = toMs(abandonmentThreshold);
+  if (!abandonmentThresholdMs) {
```
viceice, line 16: is there a warning /error somewhere about invalid values?

lib/workers/repository/process/lookup/abandonment.ts
```diff
@@ -0,0 +1,32 @@
...
+    return releaseResult;
+  }
+
+  const { bumpedAt } = releaseResult;
+  if (!bumpedAt) {
+    return releaseResult;
```
viceice, line 22: should we debug or trace log this case? so users are more aware of this case?

lib/workers/repository/process/lookup/abandonment.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+  const bumpedAtDate = DateTime.fromISO(bumpedAt);
+
+  const abandonmentDate = bumpedAtDate.plus({
+    milliseconds: abandonmentThresholdMs,
+  });
+  const isAbandoned = abandonmentDate < DateTime.now();
```
viceice, line 29: should we use local time? like we do in other places?

lib/workers/repository/process/lookup/abandonment.ts
rarkins: Trace is best. People are likely to apply a default threshold for all so there'd be too many debug messages if we logged for every package without release timestamps

lib/workers/repository/process/lookup/abandonment.ts
rarkins: I don't think datasource packages are I. Local time zone so that doesn't seem to make sense. It's all relative (also unlikely to be impacted by hours)

lib/workers/repository/process/lookup/abandonment.ts
zharinov: The error is logged in the `toMs` exception handler

zharinov: But added trace anyway

rarkins (changes requested):
Can you rename "bumpedAt" to "lastReleaseTimestamp"?

viceice (approved):

rarkins (approved):

PR #35869, opened 2025-05-11
feat(manager/npm): use volta pins as constraints

### LINKED ISSUE #35340: Renovate should detect node and npm constraints from Volta pins
### Discussed in https://github.com/renovatebot/renovate/discussions/34946

<sup>Originally posted by **felipecrs** March 22, 2025</sup>
### Tell us more.

When a repository has a package.json with Volta key, that means whatever version of Node and NPM is set there is the version that should be used in that repository.

Currently, Renovate won't take the Volta pins into account for deciding which version of node or npm to use for that repository, and the users need to either add an extra `engines.npm` field or set constraints in the renovate.json, which is an extra step and duplication to be kept up to date.

I think Renovate should simply detect the constraint from the Volta pins too.</div>

RahulGautamSingh:
## Changes
- Use the extracted volta pins as constraints, prioritizing them over `packageManager` and `engines` values

## Context
- Closes: https://github.com/renovatebot/renovate/issues/35340


lib/modules/manager/npm/extract/__snapshots__/index.spec.ts.snap
```diff
@@ -525,6 +525,9 @@ exports[`modules/manager/npm/extract/index > .extractPackageFile() > extracts vo
   ],
   "extractedConstraints": {
     "node": "8.9.2",
+    "npm": "5.9.0",
+    "pnpm": "6.11.2",
+    "yarn": "1.12.3",
```
rarkins, line 530: What's the benefit of extracting these?
- I don't think they are used during lookups (e.g. for compatibility like node is)
- If we did use them in post-upgrades then we should extract them from the latest/current version of the file, because they may have been updated

It might be better that these ones which are only used in post-update steps into `managerData`

rarkins (changes requested):

lib/modules/manager/npm/extract/common/dependency.ts
```diff
@@ -207,16 +207,21 @@ export function getExtractedConstraints(
...
-  for (const dep of deps) {
-    if (
-      !dep.skipReason &&
-      (dep.depType === 'engines' || dep.depType === 'packageManager') &&
-      dep.depName &&
-      constraints.includes(dep.depName) &&
-      is.string(dep.currentValue)
-    ) {
-      extractedConstraints[dep.depName] = dep.currentValue;
+  const constraints = new Set(['node', 'yarn', 'npm', 'pnpm', 'vscode']);
```
viceice, line 210: extracting here makes no sense, they need to be extracted in post update

lib/modules/manager/npm/post-update/pnpm.spec.ts
```diff
@@ -284,6 +285,61 @@ describe('modules/manager/npm/post-update/pnpm', () => {
...
 
+  it('uses volta version and puts it into constraint', async () => {
+    const execSnapshots = mockExecAll();
+    const configTemp = partial<PostUpdateConfig>();
+    const fileContent = codeBlock`
+    {
+  "name": "parent",
+  "version": "1.0.0",
+  "engines": {
+    "pnpm": "=6.15.0"
```
viceice, line 296:
```suggestion
    "pnpm": "^6.0.0"
```
this is usually used as range

lib/modules/manager/npm/extract/common/dependency.ts
RahulGautamSingh: Yeah, we do not these anyways, reverting.

lib/modules/manager/npm/extract/__snapshots__/index.spec.ts.snap
RahulGautamSingh: 1. Yes, they aren't used during lookups
2. We do extract them during post-update as use those values. This extraction here is redundant

lib/modules/manager/npm/extract/common/dependency.ts
```diff
@@ -208,6 +208,7 @@ export function getExtractedConstraints(
 ): Record<string, string> {
   const extractedConstraints: Record<string, string> = {};
   const constraints = ['node', 'yarn', 'npm', 'pnpm', 'vscode'];
+
```
rarkins, line 211:
```suggestion
```

rarkins (approved):

lib/modules/manager/npm/post-update/utils.ts
```diff
@@ -32,6 +32,12 @@ export function getPackageManagerVersion(
   name: string,
   pkg: PackageJsonSchema,
 ): string | null {
+  if (pkg.volta?.[name]) {
```
viceice, line 35: btw: do we pass the workspace root package here or the local? 🤔

viceice (approved):

rarkins (approved):

lib/modules/manager/npm/post-update/utils.ts
RahulGautamSingh: Local

viceice: so we probably need to check the parent package.json if we're in a workspace

viceice: That's why we sometimes see wrong tool version usage 🤦‍♂️

viceice (approved):

lib/modules/manager/npm/post-update/utils.ts
felipecrs: Ideally Renovate should use the workspace member tools for the workspace member itself, but I agree relying on the workspace parent tools is miles better than the current behavior.

PR #35879, opened 2025-05-12
feat: search for version in changelog body if not found in header

Techn1x:
## Changes

When parsing a changelog file to determine release notes for a PR, allows for the version to be found in the body of the changelog, if the line also contains the name of the package being released. This can be common in monorepo changelogs, where the changelog title doesn't necessarily contain the version, as there may be multiple packages being released of multiple versions.

Example, [changelogs built with release-plan](https://github.com/embroider-build/embroider/blob/ce3535d5a887f450cdf55b0f3fe85bd9de698822/CHANGELOG.md)

See discussion https://github.com/renovatebot/renovate/discussions/35877 for more info

## Context

This is a potential solution to resolve https://github.com/renovatebot/renovate/discussions/35877


lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -366,33 +366,45 @@ export async function getReleaseNotesMd(
             .replace(regEx(/^\s*#*\s*/), '')
             .split(' ')
             .filter(Boolean);
-          let body = section.replace(regEx(/.*?\n(-{3,}\n)?/), '').trim();
+          const body = section.replace(regEx(/.*?\n(-{3,}\n)?/), '').trim();
```
Techn1x, line 369: With the changes made here, we are able to make the body variable a `const` - a bit nicer.

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -366,33 +366,45 @@ export async function getReleaseNotesMd(
...
-                baseUrl,
-                repository,
-                project,
-                changelogFile,
-              );
-              const mdHeadingLink = title
-                .filter((word) => !isHttpUrl(word))
-                .join('-')
-                .replace(regEx(/[^A-Za-z0-9-]/g), '');
-              const url = `${notesSourceUrl}#${mdHeadingLink}`;
```
Techn1x, line 383: Lifted these up out of the loop over the title words. Allows them to also be used in the block where we search the body for the version number being released

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -366,33 +366,45 @@ export async function getReleaseNotesMd(
...
-              body = massageBody(body, baseUrl);
-              if (body?.length) {
-                try {
-                  body = await linkify(body, {
-                    repository: `${baseUrl}${repository}`,
-                  });
-                } catch (err) /* istanbul ignore next */ {
-                  logger.warn({ body, err }, 'linkify error');
-                }
-              }
```
Techn1x, line 393: moved to a function `linkifyBody` so that it can be used in the other block as well

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -366,33 +366,45 @@ export async function getReleaseNotesMd(
...
+            // Now check if any line contains both the package name and the version
+            const bodyLines = body.match(/^.*([\n\r]+|$)/gm) ?? [];
+            if (
+              bodyLines.some(
+                (line) =>
+                  line.includes(packageName) &&
+                  line.includes(version) &&
+                  !isHttpUrl(line),
+              )
+            ) {
```
Techn1x, line 404: This is the extent of the new code.

If a line in the body contains the package name and the version we're searching for, it'll return the same `ChangeLogNotes` object (body + url + notesSourceUrl)

lib/workers/repository/update/pr/changelog/__fixtures__/release-plan.md
```diff
@@ -0,0 +1,146 @@
...
+
+* release-plan 0.16.0 (minor)
+
+#### :rocket: Enhancement
+* `release-plan`
+  * [#155](https://github.com/embroider-build/release-plan/pull/155) add ability to set tag per package
+
+## Release (2025-03-03)
+
+* release-plan 0.15.0 (minor)
```
Techn1x, line 13: Fixture copied from a real changelog here, but removed committers and user tags
https://github.com/embroider-build/release-plan/blob/a7ffd1f8634bf0fdcb98397a7528000db6b51db7/CHANGELOG.md

rarkins (changes requested):
Please split this PR into two. First the refactoring PR, which should retain same logic, and then after that this PR can be merged to just show the logic _change_

Techn1x:
Ok this is updated with main, fixed up and good to go.

Unfortunately the code coverage is complaining about _one line_, I believe it's this one;
`const bodyLines = body.match(/^.*([\n\r]+|$)/gm) ?? [];`

Even though technically `.match()` returns nullish (hence the `?? []`), I cannot write a test to check the else case of this, because it is technically impossible with the regex given for any string there to return null. And typescript requires me to have the else case.

even empty string returns an array here
`"".match(/^.*([\n\r]+|$)/gm)` === `[""]`

Techn1x:
Friendly ping @rarkins - would like to get this across the line

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -389,6 +389,27 @@ export async function getReleaseNotesMd(
               };
             }
           }
+          // Look for version in body - useful for monorepos. First check for heading with "(yyyy-mm-dd)"
+          const releasesRegex = regEx(/([0-9]{4}-[0-9]{2}-[0-9]{2})/);
+          if (packageName && heading.search(releasesRegex) !== -1) {
+            // Now check if any line contains both the package name and the version
+            const bodyLines = body.match(/^.*([\n\r]+|$)/gm);
```
rarkins: What is the purpose of this match? Is it the same as body.split()?

Techn1x: It's just to split the body into an array of its lines, so that we can iterate over each line.

You are totally right this can be done with split, I forgot that split can accept regex.

This should work
`const bodyLines = body.split(/[\n\r]+/)`

It's 1am where I am right now but I can do this first thing in the morning when I start for the day

rarkins: We use `split('\n')` quite extensively in the code already. We don't support windows line endings either (it's a prerequisite to use git to massage them back and forth).

Techn1x: Good to know, thanks, that makes sense

Techn1x:
Done 🚀 body.match() replaced for body.split() - way better. Updated from main. CI passing. thanks for your patience @rarkins

rarkins (approved):

PR #35891, opened 2025-05-12
fix(datasource/maven): Reduce cache TTL to 15 minutes

zharinov:
## Context

- Ref: https://github.com/renovatebot/renovate/discussions/35812

### DIFF
--- a/lib/modules/datasource/maven/util.ts
+++ b/lib/modules/datasource/maven/util.ts
@@ -66,7 +66,7 @@ function isUnsupportedHostError(err: HttpError): boolean {
 
 const cacheProvider = new PackageHttpCacheProvider({
   namespace: 'datasource-maven:cache-provider',
-  ttlMinutes: 7 * 24 * 60,
+  ttlMinutes: 15,
   checkAuthorizationHeader: true,
   checkCacheControlHeader: false, // Maven doesn't respond with `cache-control` headers
 });


lib/modules/datasource/maven/util.ts
```diff
@@ -66,7 +66,7 @@ function isUnsupportedHostError(err: HttpError): boolean {
 
 const cacheProvider = new PackageHttpCacheProvider({
   namespace: 'datasource-maven:cache-provider',
-  ttlMinutes: 7 * 24 * 60,
+  ttlMinutes: 15,
```
rarkins, line 69: Can we cache-bust the existing?

Shegox: I agree, that we should cache bust this. Otherwise already cached dependencies are only reevaluated after 7 days of the update.

To bust the cache, we need to change the namespace, right?

Additionally, because the standard ttl for PackageHttpCacheProvider is 15min, we can drop the explicit number here (similar to how the npm cache does it): 

```suggestion
```

zharinov: We store timestamp of when data is cached and not when it expires, so if we change ttl parameter, the expiry date is recalculated and the request is made to the upstream.

rarkins (approved):

PR #35896, opened 2025-05-12
fix(kubernetes): add underscore tag support

tlecomte:
## Changes

Adjust the Kubernetes manager to add support for underscores in Docker image tags.

## Context

We have observed that Renovate would not us update automatically the Docker image for LiteLLM when used in a Kubernetes deployment file, even after specifying a custom regex versioning pattern. LiteLLM uses a unique versioning scheme, where the Docker tag is, for example: litellm_stable_release_branch-v1.69.0-stable ([link to registry](https://github.com/berriai/litellm/pkgs/container/litellm/412669894?tag=litellm_stable_release_branch-v1.69.0-stable)). With such a tag, Renovate debug output shows "currentValue = litellm", where we would expect "currentValue = litellm_stable_release_branch-v1.69.0-stable".

The root cause is a slight miss in the Regex characters used to parse Docker image references. The '\w' metacharacter includes the underscore, as described in https://github.com/google/re2/wiki/Syntax.

So here we regenerate the regex for the Docker images to allow the underscore, and we add a unit test for that scenario.

### DIFF
--- a/lib/modules/manager/kubernetes/__fixtures__/underscore-tag.yaml
+++ b/lib/modules/manager/kubernetes/__fixtures__/underscore-tag.yaml
@@ -1,0 +1,16 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: litellm
+  namespace: kube-system
+  labels:
+    app: litellm
+spec:
+  selector:
+    matchLabels:
+      app: litellm
+  template:
+    spec:
+      containers:
+      - name: litellm
+        image: "ghcr.io/berriai/litellm:litellm_stable_release_branch-v1.67.0-stable"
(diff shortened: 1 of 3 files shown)


rarkins (approved):

tlecomte:
@rarkins FYI I've had to force push in my branch to fix the commit user email, because there was a mismatch with my Github email, which was preventing the CLA from seeing my signature. The diff has not changed at all.

rarkins (approved):

lib/modules/manager/kubernetes/__fixtures__/underscore-tag.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: apps/v1
```
viceice, line 1: please inline to test with codeBlock template helper

rarkins: I think the others in this file are all Fixtures, so it's ok

PR #35900, opened 2025-05-12
feat(gerrit): add support for labels through hashtags

felipecrs:
## Changes

This PR adds support for labels in the `renovate.json` config file for the Gerrit platform. In Gerrit, labels are done through hashtags.

Example:

image

```json
{
  "packageRules": [
    {
      "matchPackageNames": ["dependency-name"],
      "labels": [
        "hi there",
        "hi+there",
        "hi-there",
        "hi_there",
        "hi/there",
        "hi%there",
        "hi=there",
        "hi.there",
        "hi;there",
        "hi^there",
        "hi~there",
        "hi@there",
        "hi$there",
        "hi#there"
      ]
    }
  ]
}
```

Therefore, this PR also adds support for `rebaseLabel`, `stopUpdatingLabel`, and `keepUpdatedLabel` for the Gerrit platform.

**EDIT: the following is no longer part of this PR.**

~Furthermore, it changes the rebase instructions to prefer `rebaseLabel` instead. This is especially important in Gerrit because otherwise updating the commit message will cause the CI pipelines to be triggered.~

image

## Context

This is a missing feature in the Gerrit platform, and I thought it would be nice to have it.


felipecrs:
@viceice, @rarkins, this PR has no dependency on #35661, and it is simpler to review.

lib/util/git/index.ts
```diff
@@ -1158,21 +1158,24 @@ export async function pushCommit({
...
 }: PushFilesConfig): Promise<boolean> {
   await syncGit();
   logger.debug(`Pushing refSpec ${sourceRef}:${targetRef ?? sourceRef}`);
   let result = false;
   try {
-    const pushOptions: TaskOptions = {
-      '--force-with-lease': null,
-      '-u': null,
-    };
+    const extraArgs = ['--force-with-lease', '-u'];
```
rarkins, line 1167: Why was this refactoring necessary?

lib/modules/platform/types.ts
```diff
@@ -242,7 +242,7 @@ export interface Platform {
   ensureIssue(
     issueConfig: EnsureIssueConfig,
   ): Promise<EnsureIssueResult | null>;
-  massageMarkdown(prBody: string): string;
+  massageMarkdown(prBody: string, rebaseLabel?: string): string;
```
rarkins, line 245: I don't like this (seems like bad design), but I'm also not sure if there's a better way.

lib/util/git/index.ts
felipecrs: That was the only way I found to pass multiple of the same options (various `--options`).

lib/modules/platform/types.ts
felipecrs: @rarkins I'm not sure about that. The tickbox in GitHub for example is way better than having to add a label.

Also, if editing a PR title (to add rebase!) is as simple in other platforms as they are in GitHub, it is probably not worth it.

Maybe it's indeed worth for some platform beyond Gerrit. I'd rather let their "maintainers" know and decide whether it's good or not.

Also, I had tried several approaches to get this done. This was by far the most cleanest.

Would it improve if I add jsdocs to this function mentioning it's only used by Gerrit?

lib/util/git/index.ts
felipecrs: `TaskOptions` type only supports `string: null | string | number`. I would need it to support ~` string[]` for:

```js
options: ["some=option", "another=option"]
```

And since that is not supported by simple-git, this was the only alternative approach.

lib/modules/platform/gerrit/index.ts
```diff
@@ -397,7 +404,15 @@ export function massageMarkdown(prBody: string): string {
     )
     .replace(
       'you tick the rebase/retry checkbox',
-      'add "rebase!" at the beginning of the commit message.',
+      rebaseLabel
```
rarkins, line 407: `rebaseLabel` in Renovate has a default value. I don't think this optionality is necessary here, unless it's your intention that a user _intentionally_ sets it to `null` to disable the capability?

lib/util/git/index.ts
rarkins: @viceice any thoughts?

lib/modules/platform/types.ts
```diff
@@ -242,7 +242,14 @@ export interface Platform {
...
     issueConfig: EnsureIssueConfig,
   ): Promise<EnsureIssueResult | null>;
-  massageMarkdown(prBody: string): string;
+  massageMarkdown(
+    prBody: string,
+    /**
+     * Useful for suggesting the use of rebase label when there is no better
+     * way, e.g. for Gerrit.
+     */
+    rebaseLabel?: string,
```
rarkins, line 251: I've thought about it some more, and I'd like to refactor the markdown text so that it mentions the rebase label for all platforms

rarkins (dismissed):

lib/modules/platform/types.ts
rarkins: I've split this off into a separate discussion: https://github.com/renovatebot/renovate/discussions/36130

lib/modules/platform/gerrit/index.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { stripIndent } from 'common-tags';
```
viceice, line 1:
```suggestion
import { codeBlock } from 'common-tags';
```

we use `codeBlock` in all other places

lib/modules/platform/gerrit/index.ts
```diff
@@ -338,6 +338,13 @@ export async function getJsonFile(
   return parseJson(raw, fileName);
 }
 
+export async function deleteLabel(
```
viceice, line 341: please move to original position below to have less changes in this pr

lib/modules/platform/gerrit/scm.ts
```diff
@@ -134,10 +134,16 @@ export class GerritScm extends DefaultGitScm {
         if (commit.autoApprove) {
           pushOptions.push('label=Code-Review+2');
         }
+        if (commit.labels) {
+          for (const label of commit.labels) {
+            pushOptions.push(`hashtag=${label}`);
```
viceice, line 139: should the values be url encoded?

lib/util/git/index.ts
```diff
@@ -1241,21 +1241,24 @@ export async function pushCommit({
...
```
viceice, line 1250: seems ok, would prefer to have a separate PR, but changes are really small, so 🤷‍♂️

lib/modules/platform/gerrit/index.ts
felipecrs: Yes, that's what I had in mind. But considering `rebaseLabel` documentation, there's no option to disable it.

image

Thus, I will just remove the optionality, please let me know otherwise.

lib/modules/platform/gerrit/scm.ts
felipecrs: No, they should not. If you try to encode them, Gerrit will create a literal hashtag with the percentage symbols.

Also, this is the main reason why I changed from `git push origin HEAD:refs/for/master%pushOptionsHere...` to `--push-option`s. `--push-option` allow pretty much any character to be passed and will be turned into a hashtag without any escaping need.

lib/util/git/index.ts
viceice: we should wait for the change in simple-git or do this refactor in a separate pr to merge before this pr

lib/modules/platform/types.ts
viceice: let's remove this refacto from this pr?

viceice (commented):
lgtm beside the two conversations

lib/modules/platform/types.ts
felipecrs: @viceice, do you mean the jsdoc?

lib/util/git/index.ts
felipecrs: I personally don't think a new PR is worth, given it's so small and also temporary. I will refactor again anyway as soon as simple-git releases a new version.

But you decide, @viceice, just say the word.

lib/modules/platform/types.ts
felipecrs: Oh, do you mean I should create a PR addressing https://github.com/renovatebot/renovate/discussions/36130 generically first?

lib/util/git/index.ts
felipecrs: This is now resolved. The new version of simple-git with this feature was released, and I refactored this PR with it.

lib/modules/platform/types.ts
rarkins: Yes

felipecrs: @rarkins, @viceice, I undid all related changes. Can you please review this PR again?

felipecrs:
@rarkins feel free to merge this as well. I will send a follow up PR adapting to #36374.

felipecrs:
@rarkins @viceice my understanding is that this PR is ready to merge. Can you please look into it? This is a gentle ping.

viceice (approved):

PR #34696, opened 2025-03-08
feat: interactive reconfigure pr

RahulGautamSingh:
## Changes
- check if reconfigure pr exists 
- if reconfigure pr found, add a comment on the pr with expected pr's list and config description
- expand reconfigure cache to store extraction result so we can skip processing the reconfigure branch if it has not changed
- make slight changes `mergeRenovateConfig` fn such that it can pass the branchName to the `detectConfigFile` fn which comes in handy when trying to read the config file from reconfigure branch using api endpoint
- added tests

## Context
- Closes: #547 

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [ ] No documentation update is required (todo)

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository 

1. https://github.com/Rahul-renovate-testing/recon-pr/pull/1
2. https://github.com/Rahul-renovate-testing/repro-10869-new/pull/6

PR #35941, opened 2025-05-15
refactor(workers): enhance some fns

RahulGautamSingh:
## Changes
1. Added `overwriteCache` param to the `extractDependecnies` & `extract` fn so that we prevent using the cached extract from the same run while processing the reconfigure PR
2. Pass the `branchName` to the `mergeRenovateConfig` and `detectRenovateConfig` fn inorder to fetch the config file  from correct branch ie. reconfigure branch instead of default branch

## Context
- Ref: https://github.com/renovatebot/renovate/pull/34696/files#r2075329471

### DIFF
--- a/lib/workers/repository/process/extract-update.ts
+++ b/lib/workers/repository/process/extract-update.ts
@@ -132,6 +132,7 @@ export function isCacheExtractValid(
 
 export async function extract(
   config: RenovateConfig,
+  overwriteCache = true,
 ): Promise<Record<string, PackageFile[]>> {
   logger.debug('extract()');
   const { baseBranch } = config;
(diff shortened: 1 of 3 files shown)


lib/workers/repository/process/extract-update.ts
```diff
@@ -142,7 +143,10 @@ export async function extract(
   const cachedExtract = cache.scan[baseBranch!];
   const configHash = fingerprint(generateFingerprintConfig(config));
   // istanbul ignore if
-  if (isCacheExtractValid(baseBranchSha!, configHash, cachedExtract)) {
+  if (
+    overwriteCache &&
+    isCacheExtractValid(baseBranchSha!, configHash, cachedExtract)
```
rarkins, line 148: So if overwriteCache=false then also don't _read_ the cache either?

BTW shouldn't we be trying to cache the reconfigure branch too though?

viceice: caching seems to make sense 🤔

viceice (commented):
see comment, otherwise LGTM

lib/workers/repository/process/extract-update.ts
RahulGautamSingh: We do use caching for the reconfigure Pr. If the branchSha of reconfigre branch hasn't changed we do not process it. 

> So if overwriteCache=false then also don't read the cache either?

Yes, both. For reconfigure pr we do not have intelligent caching like we have for normal extraction using fingerprints. 
A simple branchSha comparison is used to prevent the reconfigure branch processing.

rarkins: That seems ok then

viceice (approved):

PR #35993, opened 2025-05-17
feat: add option `skipArtifactsUpdate`

### LINKED ISSUE #33118: New option: skipArtifactUpdating

Add new config option `skipArtifactUpdating` which defaults to `false`. If it's set to `true` then artifacts updating should be skipped, regardless of package manager.

For some managers, artifact updating is done per-branch, and for others it's per-update. Where possible, it should be possible to support this new feature per-update, e.g. check update.disableArtifactUpdating and not branch.disableArtifactUpdating, and filter out any updates which have the value `true`. If all updates are skipped, artifact updating should be skipped. A debug log message should be printed any time skipping occurs.

The idea behind this is to allow users to skip/bypass Renovate's automatic artifact updating in cases where it's problematic, so that they can update artifacts via postUpgradeTasks or their own commit via CI.

reduckted:
## Changes

This adds a new option called `skipArtifactsUpdate`. When set to true, artifact updating will be skipped.

This new option replaces the existing `updateLockFiles` option. A migration will replace `updateLockFiles=false` with `skipArtifactsUpdate=true`.

## Context

- Closes #33118 


lib/modules/manager/npm/post-update/index.ts
reduckted, line 1: ℹ️ When testing the changes to `get-updated.ts`, I ran an end-to-end test against a real repository and just happened to use the `npm` manager. I was surprised to find that the `package-lock.json` file was still being updated!

Stepping through the code I found this, which seems to be a special case for `npm`. I have two questions:

1. Do you think this is the best way to handle it? I considered changing the caller of `getAdditionalFiles` (lib/workers/repository/update/branch/index.ts`) to filter out any `upgrades` that had `skipArtifactUpdating` set to true, but decided that I'd do that in here instead.
2. Are there any other special cases like this that might update artifacts? I had a look and couldn't find any, but I'm not very familiar with the code base, so I may have missed some.

lib/workers/repository/update/branch/get-updated.ts
```diff
@@ -472,6 +472,39 @@ async function managerUpdateArtifacts(
...
+    if (updateArtifact.config.skipArtifactUpdating) {
+      logger.debug(
+        { manager, packageFileName: updateArtifact.packageFileName },
+        'skipping artifact updating because skipArtifactUpdating=true',
+      );
+      return null;
+    }
+  } else {
+    // When there are updated dependencies, use the dependency's value.
+    const skipArtifactUpdatingCount = updateArtifact.updatedDeps.filter(
```
reduckted, line 487: ℹ️ Instead of using `updateArtifact.updatedDeps.every(...)` to determine if all of the updated deps want to skip artifact updating, I decided to get the number of updates deps with `skipArtifactUpdating` set to true so that a log message could indicate why artifact updating was going to run. For example, instead of just _not_ seeing the "skipping artifact updating" message, you will see a message like:

> running artifact updating because 2 out of 3 updated dependencies have skipArtifactUpdating=false

I thought this would be better for debugging because if you're expecting artifact updating to be skipped, but it's still running, then seeing _why_ it's running will make it easier to work out what's going on.

lib/config/options/index.ts
```diff
@@ -3120,6 +3120,13 @@ const options: RenovateOptions[] = [
     default: false,
     globalOnly: true,
   },
+  {
+    name: 'skipArtifactUpdating',
+    description: 'If set, artifact updating will be skipped.',
```
viceice, line 3125: this will be the first line in docs above, so it should integrate with the text above

lib/modules/manager/npm/post-update/index.ts
viceice: yes, npm manager is a special case and needs a refactor to be same as any other manager.

but that's not your concern. you probably need to duplicate some code for the upgrade filter and skipping.

lib/modules/manager/npm/post-update/index.spec.ts
```diff
@@ -559,6 +559,34 @@ describe('modules/manager/npm/post-update/index', () => {
...
+              },
+            ],
+          },
+          additionalFiles,
+        ),
+      ).toStrictEqual({
+        artifactErrors: [],
+        updatedArtifacts: [],
+      });
+      expect(spyNpm).not.toHaveBeenCalled();
```
viceice, line 587:
```suggestion
      expect(spyNpm).not.toHaveBeenCalled();
      expect(logger.debug).toHaveBeenCalledWith('Skipping lock file generation');
```

lib/workers/repository/updates/generate.ts
```diff
@@ -466,6 +466,11 @@ export function generateBranchConfig(
     (upgrade) => upgrade.skipInstalls !== false,
   );
 
+  // Artifact updating will only be skipped if every upgrade wants to skip it.
```
viceice, line 469: This should be added to the config option docs as a note, so users are awar of grouping behavior

docs/usage/configuration-options.md
```diff
@@ -4275,6 +4275,10 @@ By default, Renovate skips versions in between, like `python@v3.10.x`.
 
 But if you set `separateMultipleMinor=true` then you get separate PRs for each minor stream, like `python@3.9.x`, `python@v3.10.x` and `python@v3.11.x`, etc.
 
+## skipArtifactUpdating
+
+Use this option in cases where automatic artifact updating can be problematic, or when you would like to customize how artifacts (such as lock files) are updated. When this option is set, you should update artifacts (such as lock files), via `postUpgradeTasks` or another method such as a CI workflow.
```
viceice, line 4280: Add a note that all upgrades in a branch (aka grouping) needs this have to be true to really skip.

rarkins (changes requested):

lib/modules/manager/npm/post-update/index.ts
```diff
@@ -371,7 +371,7 @@ export async function getAdditionalFiles(
   if (!packageFiles.npm?.length) {
     return { artifactErrors, updatedArtifacts };
   }
-  if (!config.updateLockFiles) {
+  if (!config.updateLockFiles || config.skipArtifactUpdating) {
```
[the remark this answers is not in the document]
rarkins: Why does Renovate need both options now? Aren't they achieving the same thing?

viceice (dismissed):

docs/usage/configuration-options.md
```diff
@@ -4312,6 +4312,15 @@ By default, Renovate skips versions in between, like `python@v3.10.x`.
 
 But if you set `separateMultipleMinor=true` then you get separate PRs for each minor stream, like `python@3.9.x`, `python@v3.10.x` and `python@v3.11.x`, etc.
 
+## skipArtifactUpdating
+
+Use this option in cases where automatic artifact updating by Renovate fails or is incorrect, or you have custom needs.
```
RahulGautamSingh, line 4317:
```suggestion
Use this option when automatic artifact updating fails, is incorrect, or not needed.  
This is useful if you have custom workflows handling artifact updates.

```

docs/usage/configuration-options.md
```diff
@@ -4312,6 +4312,15 @@ By default, Renovate skips versions in between, like `python@v3.10.x`.
...
+When this option is set, Renovate won't attempt to update artifacts such as lock files, so you will need to update them yourself, either manually or through secondary automation such as CI workflows.
```
RahulGautamSingh, line 4318:
```suggestion
When this option is set, Renovate won't attempt to update artifacts such as lock files.
 You will need to update them yourself, either manually or via secondary automation such as CI workflows.
```

docs/usage/configuration-options.md
reduckted: @rarkins Do you want this change, or stick with what you previously suggested?
https://github.com/renovatebot/renovate/pull/35993/files/98a1e917c59b34176c3795abcdb834d9a95c6b6d#r2105764636

lib/modules/manager/npm/post-update/index.ts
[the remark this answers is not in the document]
rarkins: I would like to have only `updateLockFiles` and no new `skipArtifactUpdating` in this PR. We can later rename/migrate the name in a future PR.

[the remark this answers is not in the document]
reduckted: So `updateLockFiles` will be changed to support _all_ managers? Doesn't that have the possibility of causing undesirable behavior if it's currently defined at the root of someone's config with a value of false? They're using it to disable updating of npm lock files, but now it will disable the updating of _all_ lock files.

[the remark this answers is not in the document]
rarkins: If you configure something at the root then it applies to all managers, now and future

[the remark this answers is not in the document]
reduckted: OK, I'll make the change.

RahulGautamSingh (approved):

lib/workers/repository/update/branch/get-updated.ts
```diff
@@ -474,7 +474,14 @@ async function managerUpdateArtifacts(
 ): Promise<UpdateArtifactsResult[] | null> {
   const updateArtifacts = get(manager, 'updateArtifacts');
   if (updateArtifacts) {
-    return await updateArtifacts(updateArtifact);
+    if (updateArtifact.config.updateLockFiles === false) {
+      logger.debug(
+        { manager, packageFileName: updateArtifact.packageFileName },
+        'skipping artifact updating',
```
RahulGautamSingh, line 480:
```suggestion
        'Skipping artifact update',
```

lib/workers/repository/updates/generate.ts
```diff
@@ -466,6 +466,11 @@ export function generateBranchConfig(
     (upgrade) => upgrade.skipInstalls !== false,
   );
 
+  // Lock file updating will only be disabled if every upgrade wants to disable it.
+  config.updateLockFiles = !config.upgrades.every(
+    (upgrade) => upgrade.updateLockFiles === false,
+  );
```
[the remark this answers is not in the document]
reduckted: What's the reason for doing it that way?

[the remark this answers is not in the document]
reduckted: This suggested change produces inconsistent results because the config is initialized from the first update. If the first update has `updateLockFiles = false`, then this won't revert it back to true.

lib/config/options/index.ts
```diff
@@ -982,7 +982,6 @@ const options: RenovateOptions[] = [
     description: 'Set to `false` to disable lock file updating.',
     type: 'boolean',
     default: true,
-    supportedManagers: ['npm'],
```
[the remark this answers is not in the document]
reduckted: Is there anything that I need to do here?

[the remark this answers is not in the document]
rarkins: Not for this PR

viceice (dismissed):
@rarkins I think we should rename this option to `skipLockfileUpdate` in a follow up PR. So we can leave that field `undefined` and skip if it's set explicit to true.

rarkins:
@viceice 90% of the time it's lock files, but 10% it's truly "artifacts". Should we name it after the 90% and be wrong 10% of the time, or give it a less understandable generic "artifacts" name which is correct 100%?

viceice:
`skipArtifactsUpdate` is my new proposal

reduckted:
> `skipArtifactsUpdate` is my new proposal

That's almost what this change started as:
https://github.com/renovatebot/renovate/pull/35993/commits/9640339873e4db219dc5e8552219d49ac0429f2c

Should I just change it in this PR?

viceice:
> > `skipArtifactsUpdate` is my new proposal
> 
> That's almost what this change started as:
> https://github.com/renovatebot/renovate/pull/35993/commits/9640339873e4db219dc5e8552219d49ac0429f2c
> 
> Should I just change it in this PR?

interesting 🤔, I would support that if you also remove the old option and add a migration.

@rarkins wdyt?

rarkins:
Yes sounds good

docs/usage/configuration-options.md
```diff
@@ -4338,6 +4338,15 @@ By default, Renovate skips versions in between, like `python@v3.10.x`.
 
 But if you set `separateMultipleMinor=true` then you get separate PRs for each minor stream, like `python@3.9.x`, `python@v3.10.x` and `python@v3.11.x`, etc.
 
+## skipArtifactsUpdate
+
```
viceice, line 4413: add a note that this was formerly known as `updateLockFiles`

lib/config/presets/internal/default.ts
```diff
@@ -175,7 +175,7 @@ export const presets: Record<string, Preset> = {
   },
   disableLockFiles: {
```
viceice: @rarkins should we mark this as deprecated and add a new preset or also migrate it to a new name 🤔

lib/modules/manager/npm/post-update/index.ts
```diff
@@ -377,7 +377,7 @@ export async function getAdditionalFiles(
   if (!packageFiles.npm?.length) {
     return { artifactErrors, updatedArtifacts };
   }
-  if (!config.updateLockFiles) {
```
viceice, line 386: was that never strongly typed? 🤔

reduckted: Unfortunately not. `PostUpdateConfig` extends `Record<string, any>`.

https://github.com/renovatebot/renovate/blob/4047d64bc06f101956275db9c2750c6e07a4861f/lib/modules/manager/types.ts#L311-L312

lib/workers/repository/update/branch/get-updated.ts
```diff
@@ -474,7 +474,14 @@ async function managerUpdateArtifacts(
 ): Promise<UpdateArtifactsResult[] | null> {
   const updateArtifacts = get(manager, 'updateArtifacts');
   if (updateArtifacts) {
-    return await updateArtifacts(updateArtifact);
+    if (updateArtifact.config.skipArtifactsUpdate) {
```
RahulGautamSingh: This will be easier to read.
```ts
  if (!updateArtifacts) {
    return null;
  }

  if (updateArtifact.config.skipArtifactsUpdate) {
    logger.debug(
      { manager, packageFileName: updateArtifact.packageFileName },
      'Skipping artifact update',
    );
    return null;
  }

  return await updateArtifacts(updateArtifact);
  ```

RahulGautamSingh (approved):
LGTM apart from one comment above.

lib/config/presets/internal/default.ts
reduckted: @rarkins Any opinions on this? The migration is very simple (also, kudos on the migration infrastructure - very nice and easy to use! 😀)

rarkins: Migrate it to the new name

viceice (dismissed):

reduckted:
Is there anything else needed for this to be merged?

rarkins:
@RahulGautamSingh @viceice have either of you had the chance to test this on some real examples? This has the chance to break a lot of people if it has a bug

RahulGautamSingh:
I will test tonight _+5.30 GMT_

RahulGautamSingh (approved):
Tested

RahulGautamSingh:
**Note:**

When user disable artifact updates they will involuntarily receive wrong updates for the `packageManager` field in their `package.json` files if the version string contains a `hash` as we trigger `corepack` update in the `updateArtifacts` function for the npm manager module.

Example:

`packageManager: "pnpm@1.2.0"` this will get updated correctly ✅


`packageManager: "yarn@3.0.0+sha224.deadbeef"` this will get updated to `"yarn@3.1.0"` ❌

rarkins:
Please add that warning to the docs

RahulGautamSingh (approved):

reduckted:
Anything else needed?

rarkins (approved):

PR #35995, opened 2025-05-17
fix(manager/nuget): deduplicate registryUrls

rarkins:
## Changes

Adds a deduplication step in nuget's registry detection to avoid having essentially identical registryUrls together due to protocolVersion.

## Context

This otherwise results in duplicate lookups.

### DIFF
--- a/lib/modules/manager/nuget/util.ts
+++ b/lib/modules/manager/nuget/util.ts
@@ -142,12 +142,18 @@ export async function getConfiguredRegistries(
   }
 
   // Deduplicate registries
-  const registryUrls = registries.map((r) => r.url);
-  registries = registries.filter(
-    (r) =>
-      !registryUrls.includes(r.url + '#protocolVersion=2') &&
-      !registryUrls.includes(r.url + '#protocolVersion=3'),
+  const registryUrlsWithSourceMappings = registries.map(
+    (r) => r.sourceMappedPackagePatterns + r.url,
   );
+  registries = registries.filter((r) => {
+    const lookupKey = r.sourceMappedPackagePatterns + r.url;
+    return (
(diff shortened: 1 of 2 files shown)


lib/modules/manager/nuget/util.ts
```diff
@@ -141,6 +141,20 @@ export async function getConfiguredRegistries(
...
 
+  // Deduplicate registries
+  const registryUrlsWithSourceMappings = registries.map(
+    (r) => r.sourceMappedPackagePatterns + r.url,
+  );
+  registries = registries.filter((r) => {
+    const lookupKey = r.sourceMappedPackagePatterns + r.url;
+    return (
+      !registryUrlsWithSourceMappings.includes(
+        lookupKey + '#protocolVersion=2',
```
viceice, line 152: I don't know a protocol version 2 🤔 have you seen it in the hosted app?

rarkins: No, I just was trying to be safe. Drop it?

viceice: yes, that version doesn't exist

lib/modules/manager/nuget/util.ts
```diff
@@ -141,6 +141,18 @@ export async function getConfiguredRegistries(
     }
   }
 
+  // Deduplicate registries with #procolVersion=3
+  // Keep any which include sourceMappedPackagePatterns
+  const plainRegistryUrls = registries
+    .filter((r) => !r.sourceMappedPackagePatterns)
+    .map((r) => `${r.url}`);
```
viceice, line 148:
```suggestion
    .map((r) => r.url);
```

it's already a string (perf)

viceice (approved):

PR #35999, opened 2025-05-17
feat(cache): Use repository cache instead of memory cache for GitHub presets

### LINKED ISSUE #35991: Cache config preset lookups in repository cache

To use this repo as an example:
- This repo extends a shared config: https://github.com/renovatebot/renovate/blob/f187692ab8191d3b8e4b1bd23d555e64da4c6d25/renovate.json#L3
- Each run on this repo results in a 200 OK for `https://api.github.com/repos/renovatebot/.github/contents/default.json`

We should cache the results of this preset lookup (and any others). GitHub seems to return both an etag and last-modified header for this type of request so we should use the last-modified as it's more reliable.

This same principle should apply to inheritConfig caching too.

zharinov:
## Changes

This PR won't reduce HTTP requests, but such requests will be made with cache-related headers (`etag`/`last-modified`), which helps with rate limits.

We don't want to cache it in the package cache because it's hard to differentiate between private/public repos. Hence we use repo cache.

Unlike package cache, repository caching doesn't implement soft/hard caching separation.
We probably don't want it anyways, as we always want to get fresh files without 15 minutes delay.

## Context

- Closes: #35991

### DIFF
--- a/lib/config/presets/github/index.ts
+++ b/lib/config/presets/github/index.ts
@@ -1,7 +1,7 @@
 import is from '@sindresorhus/is';
 import { logger } from '../../../logger';
 import { ExternalHostError } from '../../../types/errors/external-host-error';
-import { memCacheProvider } from '../../../util/http/cache/memory-http-cache-provider';
+import { repoCacheProvider } from '../../../util/http/cache/repository-http-cache-provider';
 import { GithubHttp } from '../../../util/http/github';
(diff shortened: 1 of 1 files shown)


zharinov:
`inheritConfig` relies on `platform.getRawFile()` which already uses repo cache, but only for the same org. I don't remember exact reasoning behind "same org only" caching.

viceice (approved):

PR #36018, opened 2025-05-19
chore: add `eslint-config-standard` maintained fork as replacement

JounQin:
## Changes

https://www.npmjs.com/package/eslint-config-standard is unmaintained for 2+ years, so I 

## Context

https://github.com/standard/eslint-config-standard/issues/351

https://github.com/standard/standard/issues/1948

### DIFF
--- a/lib/data/replacements.json
+++ b/lib/data/replacements.json
@@ -240,6 +240,18 @@
       }
     ]
   },
+  "eslint-config-standard-to-maintained-fork": {
+    "description": "Replace `eslint-config-standard` with a maintained fork: `@eslinter/eslint-config-standard`.",
+    "packageRules": [
+      {
+        "matchDatasources": ["npm"],
+        "matchPackageNames": ["eslint-config-standard"],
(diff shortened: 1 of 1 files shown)


JounQin:
Understood, thanks for the clarification!

voxpelli:
And `neostandard` is the maintained fork, by some of the same maintainers as `standard` itself, and reason why we haven’t submitted a request like this is that we still think we can reconcile eventually with the main `standard` project

JounQin:
`neostandard` is not a drop-in fork, right?

voxpelli:
@JounQin It’s as drop-in as is sensible. It’s essentially a new major version of `standard` from many of the very maintainers of `standard` but published under a different name due to governance reasons.

@rarkins If there’s ever a maintenance fork of an ESLint plugin or similar, then the only official home for that is https://github.com/eslint-community – eg. `@eslinter` or similar is not official, despite using the eslint name in ways that makes it easy to mix up with officially governed projects

JounQin:
I never say it's official...

PR #36060, opened 2025-05-21
fix(helm): helm login should not use host path

rarkins:
## Changes

Strips path part from host for "helm login"

## Context

- #36052

### DIFF
--- a/lib/modules/manager/helmfile/utils.ts
+++ b/lib/modules/manager/helmfile/utils.ts
@@ -47,5 +47,5 @@ export async function generateRegistryLoginCmd(
     }),
   };
 
-  return await generateLoginCmd(repositoryRule, 'helm registry login');
+  return await generateLoginCmd(repositoryRule);
 }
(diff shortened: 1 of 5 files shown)


secustor (commented):
I think this will break, registries on sub paths. Like Artifactory or Nexus

rarkins:
@secustor see https://github.com/renovatebot/renovate/discussions/36052#discussioncomment-13215886

Are there other commands which we use this for where the subpath is valid?

viceice (dismissed):

viceice:
i think this schould only be done for oci charts. then they use the docker protocoll and that requires to be `https://<host>/v2/` base url.

secustor:
Not that I'm aware of. 

`helm registry login` is already only for OCI images. 
"standard" helm registries use [`helm repo add`](https://helm.sh/docs/helm/helm_repo_add/)

rarkins:
Here's what I've now changed:
- `helm registry login` was the only way this function was called
- `helm registry login` _needs_ the path stripped
- I removed the generic `loginCMD` parameter so that now it's hardcoded to `helm registry login` and we are safe to strip path. Nobody should change this by accident to something non-functional

viceice (approved):

secustor (approved):

PR #36084, opened 2025-05-22
feat: add `replace` method for `useBaseBranch` option

reduckted:
#### New `useBaseBranch` method

This adds a `replace` method for the `useBaseBranch` option. When set to `replace`, instead of merging the config from the base branch into the config from the default branch, the config from the base branch is used to create a new config that is used for the "extract" and "lookup" phases.

This proved to be a bit more complicated than I anticipated. We cannot simply _replace_ the config with the one we load from the base branch, because Renovate applies defaults, environment variables and command line arguments to the config that it loads from the default branch. We need to do the same thing to the config that we load from the base branch.

#### Base Branch Config File Name Detection

The discovery of the config from the base branch has also been changed. Previously, the config on the base branch had to have the exact same name as the config from the default branch, which meant you could not test a different type of config file - for example, you might have `renovate.json` on the default branch, but you are changing it to `renovate.json5` and want to test that config before it is merged into the default branch.

To allow this, while also keeping the existing behavior, I've added a new option called `detectBaseBranchConfigFileName`. When this is false or unspecified, the config file name from the default branch is used when looking for the config on the default branch. If a file with that name does not exist on the base branch, then an error occurs.
When the value is true, Renovate will look for each allowed config file name on the base branch. If one of those files is found, that file is used. If none of those files are found, Renovate will fall back to looking for the file name from the default branch.

## Context

- #22264

Also related to:
- #16108
- #19916

The primary goal of these changes is to allow testing config changes without having to first merge them into the default branch. The workflow would be something like this:
1. You already have a config on the default branch, and you want to make changes to it.
2. You create a branch - let's call it `renovate-changes`.
3. You edit the Renovate config on that branch, adding new package rules, editing existing package rules, and so on.
4. You commit the changes on that branch.
5. You run Renovate with these options:
    - `dryRun: true`
    - `baseBranches: ['renovate-changes']`
    - `useBaseBranchConfig: 'replace'`
6. You check the logs to confirm that the changes you made to the config produce the expected results. Iterate through steps 3 to 5 until you are happy with the config changes.
7. Merge the `renovate-changes` branch into the default branch.

No more broken Renovate configs merged to the default branch. 🎉 

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -1384,6 +1384,10 @@ The description field can be used inside any configuration object to add a human
 A description field embedded within a preset is also collated as part of the onboarding description unless the preset only consists of presets itself.
 Presets which consist only of other presets have their own description omitted from the onboarding description because they will be fully described by the preset descriptions within.
 
+## detectBaseBranchConfigFileName
+
+Set this option to true if the config file name in the base branches is different to the config file name in the default branch.
+
 ## digest
 
 Add to this object if you wish to define rules that apply only to PRs that update digests.
@@ -4361,8 +4365,9 @@ You can set this option to `false` if you wish to disable updating for pinned (s
 
 By default, Renovate will read config file from the default branch only and will ignore any config files in base branches.
 You can configure `useBaseBranchConfig=merge` to instruct Renovate to merge the config from each base branch over the top of the config in the default branch.
+You can also configure `useBaseBranchConfig=replace` to instruct Renovate to use the config from each base branch as is and not use the config in the default branch.
 
-The config file name in the base branch must be the same as in the default branch and cannot be `package.json`.
+By default, the config file name in the base branch must be the same as in the default branch and cannot be `package.json`. To allow a different config file name in the base branches, use the `detectBaseBranchConfigFileName` option.
 This scenario may be useful for testing the config changes in base branches instantly.
 
 ## userStrings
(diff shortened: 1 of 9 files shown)


RahulGautamSingh:
Also could you link the real repository you tested these changes against?

docs/usage/configuration-options.md
```diff
@@ -1384,6 +1384,10 @@ The description field can be used inside any configuration object to add a human
 A description field embedded within a preset is also collated as part of the onboarding description unless the preset only consists of presets itself.
 Presets which consist only of other presets have their own description omitted from the onboarding description because they will be fully described by the preset descriptions within.
 
+## detectBaseBranchConfigFileName
```
viceice, line 1387: move this feature to a separate PR

rarkins: I'm happy for this to be default behavior actually

viceice: sure, but its unrelated to the feature this PR is trying to implement. that's why I suggested to move it to a separate PR.

rarkins: Good point, we should have "detection of a different file name in base branch" before or after this

reduckted:
Closing because this isn't possible and doesn't solve an important use case. See https://github.com/renovatebot/renovate/discussions/22264#discussioncomment-13342293

PR #36105, opened 2025-05-23
feat(manager/github-actions): update version string comments for container jobs

iainlane:
Add support for updating comments after image references for container jobs.

A job like:

```yaml
myjob:
  contianer: foo/bar@sha256:1234567890abcdef # 1.2.3
```

Won't currently have the version string updated. This is because we use a YAML parser which does not keep track of comments.

Here we add that support. This requires us to move to a regex-based approach, similar to how we already handle comments for `uses:` references to other GitHub Actions.

These references can also be given in another form:

```yaml
myjob:
  container:
    image: foo/bar@sha256:1234567890abcdef # 1.2.3
```

which we also handle, as well as the form which _already_ works:

```yaml
myjob:
  container:
    image: foo/bar:1.2.3@sha256:1234567890abcdef
```

Either approach is common in the wild, so it would be good for Renovate to support both.


RahulGautamSingh:
Isn't this already supported via custom managers?

PR #36106, opened 2025-05-23
feat(presets): Add replacement for `netlify`

eduardoboucas:
## Changes

Adds a replacement for the `netlify` package, which historically has been used to publish the JavaScript client for the Netlify API. That client is now published as [`@netlify/api`](https://www.npmjs.com/package/@netlify/api) and [`netlify`](https://www.npmjs.com/package/netlify) is used by the Netlify CLI.

### DIFF
--- a/lib/data/replacements.json
+++ b/lib/data/replacements.json
@@ -764,6 +764,18 @@
       }
     ]
   },
+  "netlify-to-scoped": {
+    "description": "The Netlify API client has been renamed to `@netlify/api` (`netlify` is now used by the Netlify CLI).",
+    "packageRules": [
+      {
+        "matchCurrentVersion": ">=13.3.5",
+        "matchDatasources": ["npm"],
+        "matchPackageNames": ["netlify"],
+        "replacementName": "@netlify/api",
+        "replacementVersion": "14.0.1"
+      }
+    ]
+  },
   "now-to-vercel": {
     "description": "`now` was renamed to `vercel`.",
     "packageRules": [


rarkins:
Is there public documentation explaining this? and which of these is it:

- `netlify` is still usable as an API but isn't being updated since version x.y.z
- `netlify` is still usable as an API and is still being updated
- `netlify` was no longer functional as an API since version x.y.z

eduardoboucas:
@rarkins It's the second one. I've added some details to https://github.com/netlify/cli/releases/tag/v21.5.0, let me know if that fully answers your question. Thanks!

RahulGautamSingh:
> The netlify package will continue to export the same interface as before, so there are no breaking changes; but if you're only interested in the API client, you should update your imports to use @netlify/api instead.

[This comment](https://github.com/netlify/cli/releases/tag/v21.5.0) on the linked release, means that `@netlify/api` is subset of `netlify` ie. users can keep using `netlify` for their usage even if `@netlify/api` exists. So, it makes it the decision of user rather than the bot to replace the name.

It will also be wrong to move users to `@netlify/api` given that they could be using other services from the `netlify` library that `@netlify/api` does not offer.

I think, we should close this one

lib/data/replacements.json
```diff
@@ -764,6 +764,18 @@
...
   },
+  "netlify-to-scoped": {
+    "description": "The Netlify API client has been renamed to `@netlify/api` (`netlify` is now used by the Netlify CLI).",
+    "packageRules": [
+      {
+        "matchCurrentVersion": ">=13.3.5",
+        "matchDatasources": ["npm"],
+        "matchPackageNames": ["netlify"],
+        "replacementName": "@netlify/api",
+        "replacementVersion": "14.0.1"
```
viceice, line 775: current and replacement version should usually be equal

rarkins:
I'm ok with this replacement as it comes from a Netlify staff member but I think it should be aligned on compatible major versions

rarkins (changes requested):
Needs matching of from/to versions

PR #36141, opened 2025-05-26
feat(config)!: support `jsonc` in existing config + preset files

### LINKED ISSUE #29535: Support jsonc syntax for JSON config and preset files

Renovate should support JSONC for .json files plus any config files without file extension (e.g. `.renovaterc`).

RahulGautamSingh:
## Changes
- If json parsing fails, fallback to jsonc parsing and then to `json5`. There, is no waya tobe sure which format `json | jsonc | json5` is present in a file before parsing it. 

## Context
- Closes: https://github.com/renovatebot/renovate/issues/29535


rarkins (changes requested):
I prefer we just parse with JSONC instead of JSON.parse

lib/util/common.ts
```diff
@@ -105,8 +105,9 @@ export function parseJsonWithFallback(
   let parsedJson: JsonValue;
 
   try {
-    parsedJson = JSON.parse(content);
+    parsedJson = parseJsonc(content);
   } catch {
+    // warn if json5 format used in json
```
rarkins, line 110: Change message below to say JSONC

rarkins:
I don't think this is technically a _breaking change_, but it's a very notable one so let's include it in v41 so there's a higher chance everyone notices.

rarkins (changes requested):
Definitely needs some docs updating!

RahulGautamSingh:
> Definitely needs some docs updating!

Where should it be added?

rarkins:
Probably a few places where we mention coding file types. Alone should recommend using jsonc instead of json5

reduckted:
Is there any chance of `.jsonc` being added as a valid file extension for Renovate configs here?:
https://github.com/renovatebot/renovate/blob/cdbaa1ff421231223cf35b0f2bb8db2915efeaac/lib/config/app-strings.ts#L1-L12

viceice (dismissed):
is this really a breaking change?

docs/usage/config-presets.md
```diff
@@ -12,8 +12,7 @@ Shareable config presets must use the JSON or JSON5 formats, other formats are n
 
 <!-- prettier-ignore -->
 !!! tip
-    Describe what your preset does in the `"description"` field.
-    This way your configuration is self-documenting.
+    Describe what your preset does in the `"description"` field or add comments as Renovate supports `JSONC` syntax within its preset files.
```
viceice, line 15: Add a note that comment's will be stripped on automatic config migration

rarkins:
@viceice I think it's strictly not a breaking change, but it's a significant one that I'd like to bundle in a major release note so increase the chance of getting more attention. I'd like to recommend jsonc instead of json5 too in future (because we can more easily write migration code)

secustor (dismissed):

docs/usage/config-presets.md
RahulGautamSingh: It's documented in the config migration section.

lib/util/common.spec.ts
```diff
@@ -36,6 +37,16 @@ const validJsoncString = `
 }
 `;
 
+const jsoncString = codeBlock`
```
viceice: move to test, it's only used once

viceice (commented):
otherwise LGTM

reduckted:
Just repeating this in case it was missed:

> Is there any chance of `.jsonc` being added as a valid file extension for Renovate configs here?:
> 
> https://github.com/renovatebot/renovate/blob/cdbaa1ff421231223cf35b0f2bb8db2915efeaac/lib/config/app-strings.ts#L1-L12

VS Code treats a `.json` file as strictly JSON, so it does not allow comments.
image

To get rid of those errors, I can rename the file to `renovate.jsonc`, but then (unless I've misunderstood something) Renovate won't automatically find it.

rarkins:
@reduckted it's not necessary to add in v41.0.0 as it would be a new capability instead of something _changed_ about existing functionality

lib/workers/repository/reconfigure/validate.ts
```diff
@@ -42,6 +46,69 @@ export async function validateReconfigureBranch(
...
+      branchName,
+      'Validation Failed - Unparsable config file',
+      'red',
+      context,
+    );
+    setReconfigureBranchCache(branchSha, false);
+    await scm.checkoutBranch(config.baseBranch!);
+    return;
+  }
+
```
RahulGautamSingh:
```suggestion
```

reduckted:
> @reduckted it's not necessary to add in v41.0.0 as it would be a new capability instead of something _changed_ about existing functionality

I don't understand what you mean. What a the point of supporting `.jsonc` files if Renovate won't automatically find them?

rarkins:
This not about supporting .jsonc files. It's about using JSONC parsing instead of JSON parsing for JSON files

karlhorky:
> > I don't understand what you mean. What a the point of supporting `.jsonc` files if Renovate won't automatically find them?
>
> This not about supporting .jsonc files. It's about using JSONC parsing instead of JSON parsing for JSON files

Interesting, why would Renovate support the JSONC syntax without supporting the `.jsonc` file extension?

Seems like that's probably confusing and has bad DX, eg. [editors showing errors in the file](https://github.com/renovatebot/renovate/pull/36141#issuecomment-2976186671), other tooling which expects strict JSON in `.json` files, etc.

I can understand it for "any config files without file extension (e.g. `.renovaterc`)" mentioned in #29535, but supporting it for `.json` files seems anti-interop.

reduckted:
> Seems like that's probably confusing and has bad DX

Totally agree. I know the issue says "Renovate should support JSONC for .json files", but if I didn't know this and I see that Renovate supports JSONC syntax and then I get error highlighting in my `.json` file, then I'm very likely going to change the file extension to `.jsonc`.

I assume that at some point in the future you're going to support `.jsonc` _files_, but why not do it as part of this change? It's only four lines that need to be added to that array and it's done, right?

rarkins:
We do not want to change more than is necessary in this PR. You're welcome to submit a PR after v41 which adds .jsonc support.

rarkins (approved):

PR #36177, opened 2025-05-26
feat(presets): add `react-spring` monorepo

terrymun:
## Changes

This PR groups the [`react-spring` packages](https://github.com/pmndrs/react-spring) into a single monorepo.

## Context

I discovered `@react-spring/*` packages are being listed to be upgraded individually and therefore missing from the monorepo settings. This was after I have to configure pnpm overrides due to a vendor bug, which triggered the following entries in the renovate bot dashboard:

Screenshot 2025-05-27 at 01 35 27

As these packages are all published with the same version, it is justified that they are considered a monorepo group.

### DIFF
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -509,6 +509,7 @@
       "https://github.com/remix-run/react-router"
     ],
     "react-spectrum": "https://github.com/adobe/react-spectrum",
+    "react-spring": "https://github.com/pmndrs/react-spring",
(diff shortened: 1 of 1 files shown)


rarkins (approved):

terrymun:
Thanks for looking into this @rarkins 🥇 I noticed that the build docs workflow is failing, but I doubt it's the changes I've introduced that's causing it. Do let me know if there's something I need to change tho 👍

PR #36199, opened 2025-05-28
feat: structured logger

rarkins:
## Changes

Adds new typed logging for fatal messages

## Context

I would like to consolidate and strongly type all warn, error and fatal messages

### DIFF
--- a/lib/logger/index.ts
+++ b/lib/logger/index.ts
@@ -6,9 +6,16 @@ import upath from 'upath';
 import cmdSerializer from './cmd-serializer';
 import configSerializer from './config-serializer';
 import errSerializer from './err-serializer';
+import { LOG_CODES } from './error-codes';
 import { RenovateStream } from './pretty-stdout';
 import { RenovateLogger } from './renovate-logger';
-import type { BunyanRecord, Logger } from './types';
+import type {
+  AdditionalFieldsMap,
+  BunyanRecord,
+  Exact,
+  FatalCode,
+  Logger,
+} from './types';
 import {
   ProblemStream,
   getEnv,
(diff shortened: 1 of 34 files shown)


lib/logger/index.ts
```diff
@@ -159,3 +166,11 @@ export function getProblems(): BunyanRecord[] {
 export function clearProblems(): void {
   return problems.clearProblems();
 }
+
+export function logFatal<C extends FatalCode, A extends AdditionalFieldsMap[C]>(
```
rarkins, line 170: I wasn't able to make this fit into `logger.fatal()` but I'm open to that.

lib/types/exact.ts
rarkins, line 1: This type came from ChatGPT. I tried using the `Exact` from `type-fest` instead, but it wasn't the same

secustor (commented):
I'm for this, but maybe there should a default handler for `errorMessage`, if error has been provided as a field. 

So that `errorMessage: err.message`  is the default, but can be overwritten.

rarkins:
So make err:Error an optional field for all?

viceice (commented):
failed CI checks

lib/logger/utils.ts
```diff
@@ -315,7 +315,11 @@ export function validateLogLevel(
       },
     ],
   });
-  logger.fatal({ logLevel: logLevelToCheck }, 'Invalid log level');
+  // Emulate logFatal convention
```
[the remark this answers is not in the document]
rarkins: This is the one log which is needed before the new logger is initialized, so it was always "special". Not sure if possible to refactor

Issue #34130, opened 2025-02-10
Create new abandoned package preset and add to config:best-practices


rarkins:
### Describe the proposed change(s).

`config:best-practices` should have a new default of `1 year` for `abandonmentThreshold`, but in addition we should add a new file in https://github.com/renovatebot/renovate/tree/main/lib/data which allows for community-sourced overrides which get converted into `packageRules`. This overrides file can contain a single entry to begin with / use as an example: npm's lodash. Although it's had no release in ~4 years, it's not fully abandoned - for example the maintainers recently added a SECURITY.md file to the repo.

rarkins:
Completed in https://github.com/renovatebot/renovate/pull/36210

PR #36210, opened 2025-05-28
feat(presets): Package abandonment preset

zharinov:
## Context

- Ref: #34130


data/abandonment-config.schema.json
```diff
@@ -0,0 +1,92 @@
...
+  "type": "object",
+  "definitions": {
+    "datasourceConfig": {
+      "type": "object",
+      "additionalProperties": {
+        "type": "string"
+      }
+    }
+  },
+  "properties": {
```
rarkins, line 14: This seems non-ideally static. I'd prefer to keep it simple with just any string allowed and rely on human review, than to have it needing to be statically maintained

lib/config/presets/internal/abandonments.ts
```diff
@@ -0,0 +1,38 @@
...
+  const data: Record<string, Record<string, string>> = JSON.parse(
+    dataFiles.get('data/abandonment-config.json')!,
+  );
+  delete data.$schema;
+
+  const packageRules: PackageRule[] = [
+    {
+      matchPackageNames: ['*'],
+      abandonmentThreshold: '1 year',
+    },
```
rarkins, line 10: Maybe a higher one for npm `@types/*`? @viceice wdyt

viceice: sounds good

lib/config/presets/internal/abandonments.ts
```diff
@@ -0,0 +1,38 @@
+import dataFiles from '../../../data-files.generated';
+import type { PackageRule } from '../../types';
+import type { Preset } from '../types';
+
+/* v8 ignore start: not testable */
```
viceice: why this isn't testable?

data/abandonment-config.json
viceice, line 1: can we please store it here: https://github.com/renovatebot/renovate/blob/cfa0990d50df7a08f21538787385eee5a32ed72a/lib/data

and directly import it in typescript, so we don't need the json blob. I plan to do it for the other files later too.

viceice: https://github.com/renovatebot/renovate/blob/3b6b00bf66e9134c0cfdc96a7d8793fbfdafb0ee/lib/config/presets/internal/replacements.ts#L1

tools/schemas/schema.ts
```diff
@@ -65,3 +65,9 @@ export const SourceUrlsSchema = z
     $schema: z.string(),
   })
   .catchall(z.record(z.string(), z.string().url()));
+
+export const AbandonmentsSchema = z
```
viceice, line 69: I think we should add a test which validates the file against this Schema?

viceice (commented):
otherwise LGTM

tools/schemas/schema.ts
zharinov: In fact, I added this schema only because we have such a test 🙂

viceice: I can't see where it's used 🤔

is there some generic test which uses this?

zharinov: https://github.com/renovatebot/renovate/blob/main/test/other/validate-schemas.spec.ts

viceice (dismissed):

zharinov:
Merging this leads to a spike of Dashboard edits

rarkins (changes requested):
Needs docs/examples on how to use

docs/usage/presets-default.md
rarkins, line 1: Two concerns:
- it's a generic doc, but contains only one subtopic
- it's not linked from anywhere, so will essentially be hidden in public docs

zharinov: Obviously enough, I just vibe-coded it.

Are there any existing `.md` files worth adding to instead of creating new one? LLM didn't find any, that's why it created a separate file.

Let's decide on that, then I'll link it with the relevent config options docs.

rarkins: Maybe a new doc combining just deprecated and abandoned concepts

zharinov: Okay, will colocate it with `Warnings for deprecated dependencies` in the `docs/usage/key-concepts/dashboard.md` then. Actually when I asked LLM to find a relevant doc file to add the doc, it was too much.

docs/usage/key-concepts/dashboard.md
```diff
@@ -66,6 +66,29 @@ Here is an example of how this can look:
...
+Renovate includes the `abandonments:recommended` preset (automatically included in `config:best-practices`) to help identify potentially abandoned packages. This preset:
+
+- Sets a default [`abandonmentThreshold`](../configuration-options.md#abandonmentthreshold) of `1 year` for all packages
+- Includes community-sourced overrides for packages that appear abandoned but are still maintained
+- Prevents updates to truly abandoned packages while allowing updates to packages with irregular release schedules
+
+**Key overrides:**
+
+- `@types/*` packages (npm): Set to `eternal` - TypeScript definitions don't need frequent releases
+- `lodash` (npm): Set to `6 years` - Maintainers still respond to security issues despite infrequent releases
```
rarkins: This list will require manual updating in docs, so I'm not sure if it's a good idea. Maybe just the `@types/` one?

zharinov: Removed this part. Those who contribute to `.json` will figure it out from the contents.

rarkins (approved):

rarkins (approved):

rarkins (approved):

viceice (approved):

PR #36220, opened 2025-05-29
feat(git): support x.509 commit signing

rarkins:
## Changes

Adds support for x.509 commit signing

## Context

- Closes #36219

### DIFF
--- a/lib/util/git/private-key.ts
+++ b/lib/util/git/private-key.ts
@@ -127,7 +127,7 @@ class X509Key extends PrivateKey {
   protected async importKey(): Promise<string | undefined> {
     const keyFileName = upath.join(os.tmpdir() + '/git-private-x509.key');
     await fs.outputFile(keyFileName, this.key);
-    const { stdout, stderr } = await exec(`gpgsm --import ${keyFileName}`);
+    const { stdout, stderr } = await exec(['gpgsm', '--import', keyFileName]);
     logger.debug({ stdout, stderr }, 'X.509 private key import result');
     await fs.remove(keyFileName);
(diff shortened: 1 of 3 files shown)


zharinov (changes requested):
The array arguments would help to avoid errors with path containing whitespaces

lib/util/git/private-key.ts
```diff
@@ -114,8 +117,32 @@ class SSHKey extends PrivateKey {
...
+  protected readonly gpgFormat = 'x509';
+
+  constructor(key: string) {
+    super(key.trim());
+  }
+
+  protected async importKey(): Promise<string | undefined> {
+    const keyFileName = upath.join(os.tmpdir() + '/git-private-x509.key');
+    await fs.outputFile(keyFileName, this.key);
+    const { stdout, stderr } = await exec(['gpgsm', '--import', keyFileName]);
```
viceice, line 130: is the `gpgsm` part of a git / gnupg installation?

it seems to be a separate package, maybe we need to add it to our images first?

https://packages.debian.org/de/sid/gpgsm

viceice (commented):
otherwise LGTM

lib/util/git/private-key.ts
rarkins: It seems already in our image

viceice (approved):

Frulfump:
This was closed without a comment, any follow up?

zharinov:
> This was closed without a comment, any follow up?

It doesn't seem to work, so needs to be re-implemented

rarkins:
Yes, we need to change it to use a different key format which is more widely used, so it's easier to start over

PR #36257, opened 2025-06-01
docs: updated matchSourceUrls config option

bernardo-martinez:
make matchSourceUrls more explicit

## Changes

Include more info of possible values for `matchSourceUrls`

## Context

The current docs were confusing to me (and also made AI agents confuse) cause as it is now it can be think that only exact matches are allowed.

See discussion: https://github.com/renovatebot/renovate/discussions/36218

### DIFF
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -1445,7 +1445,7 @@ const options: RenovateOptions[] = [
   },
   {
     name: 'matchSourceUrls',
-    description: 'A list of source URLs to match against. Each element can be a string, glob pattern or regex',
+    description: 'A list of exact match URLs or URL patterns to match sourceUrl against.',
     type: 'array',
     subType: 'string',
     allowString: true,
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -1445,7 +1445,7 @@ const options: RenovateOptions[] = [
   },
   {
     name: 'matchSourceUrls',
-    description: 
+    description:
       'A list of exact match URLs or URL patterns to match sourceUrl against.',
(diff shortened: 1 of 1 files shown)


lib/config/options/index.ts
```diff
@@ -1445,7 +1445,7 @@ const options: RenovateOptions[] = [
   },
   {
     name: 'matchSourceUrls',
-    description: 'A list of source URLs to exact match against.',
+    description: 'A list of source URLs to match against. Each element can be a string, glob pattern or regex',
```
viceice: I think this file should be charged instead 

https://github.com/renovatebot/renovate/blob/main/docs/usage/configuration-options.md#matchsourceurls

there's already a note about string matching and a link to the additional docs

rarkins:
```suggestion
    description: 'A list of exact match URLs or URL patterns to match sourceUrl against.',
```

RahulGautamSingh (approved):

rarkins (dismissed):

lib/config/options/index.ts
```diff
@@ -1445,7 +1445,8 @@ const options: RenovateOptions[] = [
   },
   {
     name: 'matchSourceUrls',
-    description: 'A list of source URLs to exact match against.',
+    description:
+      'A list of exact match URLs or URL patterns to match sourceUrl against.',
```
HonkingGoose:
```suggestion
      'A list of exact match URLs (or URL patterns) to match sourceUrl against.',
```

HonkingGoose (changes requested):

rarkins (approved):

PR #36261, opened 2025-06-01
refactor!: remove rangeStrategy=pin from versioning modules

rarkins:
## Changes

Refactors getNewValue logic to replace it with an optional getPinnedValue() where rangeStrategy=pin. This way each versioning module doesn't need to add and test extensive "pin" logic.

## Context

Refactor prior to restoring `pinVersions` logic.


lib/modules/versioning/debian/index.ts
rarkins, line 1: I have most concerns about this debian versioning. I'm not sure why it needed to be so complex in the first place, so maybe this simplification is completely fine. Ideally should be tested on a real repo

rarkins:
I noticed something curious: with go.mod values, it's impossible to pin to a version, so I was wondering how rangeStrategy=pin worked for it. Turns out we use `semver` versioning so everything looks like an exact version already and doesn't need pinning.

lib/modules/versioning/conda/index.ts
```diff
@@ -64,17 +63,17 @@ function isSingleVersion(input: string): boolean {
   return isValidVersion(input.replace(/^==/, '').trimStart());
 }
 
+function getPinnedValue(newVersion: string): string {
+  return '==' + newVersion;
```
viceice: use string template

RahulGautamSingh:
```suggestion
  return `==${newVersion}`;
```

lib/modules/versioning/pep440/range.ts
```diff
@@ -98,6 +98,10 @@ interface Range {
   version: string;
 }
 
+export function getPinnedValue(newVersion: string): string {
+  return '==' + newVersion;
```
RahulGautamSingh: Just to be consistent as we use string templates in most places.
```suggestion
  return `==${newVersion}`;
```

lib/modules/versioning/debian/index.ts
RahulGautamSingh: After these changes we won't pin codenames to exact version is that, okay?

lib/modules/versioning/hex/index.ts
RahulGautamSingh, line 1: We should test this in a real repo too

lib/modules/versioning/ruby/index.ts
```diff
@@ -123,9 +122,6 @@ const getNewValue = ({
           });
         }
         break;
-      case 'pin':
-        newValue = pin({ to: vtrim(newVersion) });
```
viceice, line 127: we probably need to override the getpinnedvalue to do the vtrim?

rarkins: Added in a getPinnedValue() using vtrim

lib/modules/versioning/debian/index.ts
```diff
@@ -40,31 +40,7 @@ export class DebianVersioningApi extends GenericVersioningApi {
...
-      }
-
-      // current value is codename or [oldold|old|]stable
-      if (
-        this._distroInfo.isCodename(currentValue) ||
-        this._rollingReleases.has(currentValue)
-      ) {
-        return newVer;
-      }
-    }
```
viceice, line 64: I think we lose to pin a codename to number, as it happens now with this range strategy.

Did you tested it on real repo?

viceice (approved):
LGTM beside the open question

lib/modules/versioning/debian/index.ts
rarkins: Yes, let's test it, because I think it should work automatically. e.g. currentValue will be a codename, currentVersion will be a number, so pinning will mean it's replaced with a number

secustor (approved):

rarkins:
Let's merge in v42

lib/modules/versioning/debian/index.ts
RahulGautamSingh: https://github.com/Rahul-renovate-testing/debian-new

It doesn't

viceice (approved):

jamietanna:
We've missed some documentation changes as part of this that lead to some confusion - I'll raise a follow-up shortly

PR #36269, opened 2025-06-02
feat: Introduce data file for post-upgrade commands

mmorhun:
## Changes

Introduce data file for post-upgrade commands.
Data file is created using the same template as the commands in OS temp directory. Each post-upgrade command can access the data file by reading its full path from `RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE` environment variable.
Example configuration (snippet):
```json
{
  "postUpgradeTasks": {
    "commands": [
      "my-script.py --data-file \"$RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE\""
    ],
    "dataFileTemplate": "[{{#each upgrades}}{\"depName\": \"{{{depName}}}\", \"currentValue\": \"{{{currentValue}}}\", \"newValue\": \"{{{newValue}}}\"}{{#unless @last}},{{\/unless}}{{\/each}}]"
  }
}
```

## Context

Sometimes, post-upgrade commands need to have information about changed by renovate dependencies in order to perform required custom actions. Currently, the only way of achieving that is to configure the post-upgrade command like:
```
my-script.py --data '[{{#each upgrades}}{\"depName\": \"{{{depName}}}\", \"currentValue\": \"{{{currentValue}}}\", \"newValue\": \"{{{newValue}}}\"}{{#unless @last}},{{\/unless}}{{\/each}}]'
```
However, when there are a lot of updates are made by renovate, the template is expanded into a very long line which could exceed the maximum allowed command length and renovate will fail to run post-upgrade command (`spawn E2BIG` in case of Linux host).

To overcome the above problem, this PR introduces data file for post-upgrade commands.
Also, transferring such a big data chunk as an argument(s) doesn't look right and the commands list looks more neat when using a file.

## Related discussions

https://github.com/renovatebot/renovate/discussions/36040


docs/usage/configuration-options.md
```diff
@@ -3768,6 +3768,30 @@ They will be compiled _prior_ to the comparison against [`allowedCommands`](./se
 !!! note
     Do not use `git add` in your commands to add new files to be tracked, add them by including them in your [`fileFilters`](#filefilters) instead.
 
+### dataFileTemplate
+
+A template to create data file from.
+The template uses the same fromat as `commands`.
+The data file is created as a temporary file and the path to the data file is stored in the `RENOVATE_POST_UPGRADE_COMMAND_DATA_FILE` environment variable avaliable to each post-upgrade command.
+The primary purpose of the data file is to store some update information in a file which would be consumed from a post-upgrade command.
+This is particulary useful if a post-upgrade command need to have a long line of arguments.
```
```suggestion
This is particularly useful if a post-upgrade command needs to have a long line of arguments.
```

lib/config/types.ts
```diff
@@ -194,6 +194,7 @@ export type ExecutionMode = 'branch' | 'update';
 
 export interface PostUpgradeTasks {
   commands?: string[];
+  dataFileTemplate?: string;
```
astellingwerf, line 198: I think we could grant ourselves more flexibility if we make this an array with multiple string for multiple files. Would that make sense?

lib/util/fs/index.ts
```diff
@@ -56,6 +56,13 @@ export async function readLocalSymlink(
   }
 }
 
+export async function writeFile(
+  fileName: string,
+  fileContent: string | Buffer,
+): Promise<void> {
+  await fs.outputFile(fileName, fileContent);
+}
+
```
astellingwerf: What's the reason for the addition if there already is `writeSystemFile` that does the exact same thing?

docs/usage/configuration-options.md
```diff
@@ -3768,6 +3768,30 @@ They will be compiled _prior_ to the comparison against [`allowedCommands`](./se
 !!! note
     Do not use `git add` in your commands to add new files to be tracked, add them by including them in your [`fileFilters`](#filefilters) instead.
 
+### dataFileTemplate
+
+A template to create data file from.
+The template uses the same fromat as `commands`.
```
```suggestion
The template uses the same format as `commands`.
```

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -61,6 +67,40 @@ export async function postUpgradeCommandsExecutor(
...
+        const dataFileContent = compile(
+          dataFileTemplate,
+          mergeChildConfig(config, upgrade),
+        );
+        logger.debug(
+          { dataFileTemplate },
+          'Processed post-upgrade commands data file template.',
+        );
+
+        const tmpDir = os.tmpdir();
```
astellingwerf: Can I suggest we use `fs.mkdtemp` instead (with `import fs from 'fs-extra'`), maybe best abstracted inside `lib/util/fs`?

astellingwerf (changes requested):

lib/config/types.ts
mmorhun: It has been discussed in https://github.com/renovatebot/renovate/discussions/36040 and the resolution is to use a single file.

lib/util/fs/index.ts
mmorhun: Thanks for pointing this out, I've overlooked that such a function already exists. Also renamed `deleteFile`  to `deleteSystemFile` for consistency.

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
mmorhun: Thank you for the idea. Could you elaborate more why do we need a separate folder for a single file? In case of many updates it would create twice more file descriptors / cleanup load.

lib/workers/repository/update/branch/execute-post-upgrade-commands.spec.ts
```diff
@@ -114,6 +116,122 @@ describe('workers/repository/update/branch/execute-post-upgrade-commands', () =>
...
+      };
+      git.getRepoStatus.mockResolvedValueOnce(
+        partial<StatusResult>({
+          modified: [],
+          not_added: [],
+          deleted: [],
+        }),
+      );
+      GlobalConfig.set({
+        localDir: __dirname,
```
viceice: use a temp folder

https://github.com/renovatebot/renovate/blob/3b6b00bf66e9134c0cfdc96a7d8793fbfdafb0ee/lib/modules/datasource/crate/index.spec.ts#L114

lib/workers/repository/update/branch/execute-post-upgrade-commands.spec.ts
```diff
@@ -114,6 +116,122 @@ describe('workers/repository/update/branch/execute-post-upgrade-commands', () =>
...
```
viceice: same

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
viceice: we probably should use our private cache dir instead, which is already cleaned after each repo.

https://github.com/renovatebot/renovate/blob/3b6b00bf66e9134c0cfdc96a7d8793fbfdafb0ee/lib/modules/manager/helmv3/common.ts#L53-L59

lib/config/options/index.ts
```diff
@@ -172,6 +172,15 @@ const options: RenovateOptions[] = [
     default: [],
     cli: false,
   },
+  {
+    name: 'dataFileTemplate',
+    description: 'A template to create post-upgrade command data file from.',
+    type: 'string',
+    parents: ['postUpgradeTasks'],
+    default: '',
```
secustor:
```suggestion
```

noop

secustor (dismissed):

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -61,6 +67,39 @@ export async function postUpgradeCommandsExecutor(
...
+          dataFileTemplate,
+          mergeChildConfig(config, upgrade),
+        );
+        logger.debug(
+          { dataFileTemplate },
+          'Processed post-upgrade commands data file template.',
+        );
+
+        const dataFileName = `renovate-post-upgrade-data-file-${crypto.randomBytes(8).toString('hex')}.tmp`;
+        dataFilePath = path.join(privateCacheDir(), dataFileName);
```
astellingwerf: Wouldn't it be better to use `ensureCacheDir` over `privateCacheDir` here?

lib/workers/repository/update/branch/execute-post-upgrade-commands.spec.ts
```diff
@@ -114,6 +116,130 @@ describe('workers/repository/update/branch/execute-post-upgrade-commands', () =>
...
+        partial<StatusResult>({
+          modified: [],
+          not_added: [],
+          deleted: [],
+        }),
+      );
+      const tmpDir = await dir({ unsafeCleanup: true });
+      const cacheDir = path.join(tmpDir.path, 'cache');
+      GlobalConfig.set({
+        localDir: __dirname,
```
viceice: no, use temp folder

lib/workers/repository/update/branch/execute-post-upgrade-commands.spec.ts
```diff
@@ -114,6 +116,130 @@ describe('workers/repository/update/branch/execute-post-upgrade-commands', () =>
...
+        baseBranch: 'base',
+      };
+      git.getRepoStatus.mockResolvedValueOnce(
+        partial<StatusResult>({
+          modified: [],
+          not_added: [],
+          deleted: [],
+        }),
+      );
+      const tmpDir = await dir({ unsafeCleanup: true });
```
viceice: ensure cleanup in afterEach or try-finally

lib/workers/repository/update/branch/execute-post-upgrade-commands.spec.ts
```diff
@@ -114,6 +116,130 @@ describe('workers/repository/update/branch/execute-post-upgrade-commands', () =>
...
```
viceice: same here

lib/workers/repository/update/branch/index.spec.ts
```diff
@@ -2125,6 +2139,108 @@ describe('workers/repository/update/branch/index', () => {
...
+      scm.isBranchModified.mockResolvedValueOnce(true);
+      git.getRepoStatus.mockResolvedValueOnce(
+        partial<StatusResult>({
+          modified: ['modified_file', 'modified_then_deleted_file'],
+          not_added: [],
+          deleted: ['deleted_file', 'deleted_then_created_file'],
+        }),
+      );
+
+      const tmpDir = await dir({ unsafeCleanup: true });
```
viceice, line 2173: same here

lib/workers/repository/update/branch/execute-post-upgrade-commands.spec.ts
```diff
@@ -8,6 +10,19 @@ vi.mock('../../../../util/fs');
...
+
+    beforeEach(async () => {
+      GlobalConfig.reset();
+
+      tmpDir = await dir({ unsafeCleanup: true });
+    });
+
+    afterEach(async () => {
+      await tmpDir?.cleanup();
+      tmpDir = undefined;
```
viceice:
```suggestion
```
not needed

lib/workers/repository/update/branch/execute-post-upgrade-commands.spec.ts
```diff
@@ -8,6 +10,19 @@ vi.mock('../../../../util/fs');
 
 describe('workers/repository/update/branch/execute-post-upgrade-commands', () => {
   describe('postUpgradeCommandsExecutor', () => {
+    let tmpDir: DirectoryResult | undefined;
```
viceice:
```suggestion
    let tmpDir: DirectoryResult;
```
we know it's initialized before each test. simplifies the logic.

lib/workers/repository/update/branch/execute-post-upgrade-commands.spec.ts
```diff
@@ -114,6 +129,128 @@ describe('workers/repository/update/branch/execute-post-upgrade-commands', () =>
...
+        baseBranch: 'base',
+      };
+      git.getRepoStatus.mockResolvedValueOnce(
+        partial<StatusResult>({
+          modified: [],
+          not_added: [],
+          deleted: [],
+        }),
+      );
+      const cacheDir = path.join(tmpDir!.path, 'cache');
```
viceice:
```suggestion
      const cacheDir = path.join(tmpDir.path, 'cache');
```

lib/workers/repository/update/branch/execute-post-upgrade-commands.spec.ts
```diff
@@ -114,6 +129,128 @@ describe('workers/repository/update/branch/execute-post-upgrade-commands', () =>
...
+      GlobalConfig.set({
+        localDir: path.join(tmpDir!.path, 'local'),
```
viceice:
```suggestion
        localDir: path.join(tmpDir.path, 'local'),
```

lib/workers/repository/update/branch/execute-post-upgrade-commands.spec.ts
```diff
@@ -114,6 +129,128 @@ describe('workers/repository/update/branch/execute-post-upgrade-commands', () =>
...
```
viceice:
```suggestion
      const cacheDir = path.join(tmpDir.path, 'cache');
      GlobalConfig.set({
        localDir: path.join(tmpDir.path, 'local'),
```

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -105,6 +148,24 @@ export async function postUpgradeCommandsExecutor(
         }
       }
 
+      if (dataFilePath) {
+        // Clean up data file after post-upgrade commands ran.
+        try {
+          await deleteSystemFile(dataFilePath);
```
viceice: not needed, private cache dir is cleaned when repo is finished 

https://github.com/renovatebot/renovate/blob/a181496be47c1395e8020b68b2b5f7290797f1e4/lib/workers/repository/index.ts#L141

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -61,6 +67,42 @@ export async function postUpgradeCommandsExecutor(
...
+          dataFileTemplate,
+          mergeChildConfig(config, upgrade),
+        );
+        logger.debug(
+          { dataFileTemplate },
+          'Processed post-upgrade commands data file template.',
+        );
+
+        const dataFileName = `data-file-${crypto.randomBytes(8).toString('hex')}.tmp`;
+        const dataFilesCacheDir = await ensureCacheDir(
```
viceice: didn't we told you to use private cache dir? 🤔

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
viceice: wrong, private cache dir is for repo specific files which are cleaned after repo is finished.
that's why I initially suggested to use that.

https://github.com/renovatebot/renovate/blob/a181496be47c1395e8020b68b2b5f7290797f1e4/lib/workers/repository/index.ts#L141

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
mmorhun: Looks like there were controversy suggestions. Will revert back to use the private cache directory.

mmorhun:
Could someone trigger retest, seems the failure is infra related.

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -61,6 +66,38 @@ export async function postUpgradeCommandsExecutor(
...
+          logger.debug(
+            { dataFilePath, dataFileContent },
+            'Created post-upgrade commands data file.',
+          );
+        } catch (error) {
+          dataFilePath = '';
+
+          artifactErrors.push({
+            stderr: sanitize(
+              `Failed to create post-upgrade commands data file at ${dataFilePath}, reason: ${error.message}`,
```
astellingwerf, line 93: Why do we use dataFilePath in this string, if we know it's an empty string 100% of the time?

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -61,6 +66,38 @@ export async function postUpgradeCommandsExecutor(
...
+        logger.debug(
+          { dataFileTemplate },
+          'Processed post-upgrade commands data file template.',
+        );
+
+        const dataFileName = `post-upgrade-data-file-${crypto.randomBytes(8).toString('hex')}.tmp`;
+        dataFilePath = upath.join(privateCacheDir(), dataFileName);
+
+        try {
+          await writeSystemFile(dataFilePath, dataFileContent);
```
astellingwerf: It'd be safer to use outputCacheFile instead, right?:
```suggestion
          await outputCacheFile(dataFilePath, dataFileContent);
```

astellingwerf (changes requested):

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
mmorhun: Yeah, thanks, wrong order.

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -61,6 +66,38 @@ export async function postUpgradeCommandsExecutor(
         }
       }
 
+      let dataFilePath = '';
+      if (dataFileTemplate) {
+        const dataFileContent = compile(
+          dataFileTemplate,
+          mergeChildConfig(config, upgrade),
+        );
```
rarkins: We need to make sure this is checked for accidental secrets leakage

rarkins (changes requested):

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
mmorhun: Looking at [templates](https://docs.renovatebot.com/templates/), I don't see any possibly sensitive information there.
And a follow up question: is there any utility I can use to do the check or maybe there is a code which already does the check?
Also, I've found that in a few other places, like [this](https://github.com/renovatebot/renovate/blob/5ad015812e4ef840c756ec5df883971db499f632/lib/modules/manager/helmv3/common.ts#L53), just private cache folder is used. What did I miss?

rarkins: Look for where calls to sanitize() are used, you can just do the same

astellingwerf (approved):

rarkins (dismissed):

rarkins:
@secustor or @viceice would you like to review before merging?

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -61,6 +66,37 @@ export async function postUpgradeCommandsExecutor(
         }
       }
 
+      let dataFilePath = '';
```
secustor:
```suggestion
      let dataFilePath: string | null = null;
```
The code base convention is to use `null` for representation of not set values when initializing.

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -61,6 +66,37 @@ export async function postUpgradeCommandsExecutor(
...
+            'Created post-upgrade commands data file.',
+          );
+        } catch (error) {
+          artifactErrors.push({
+            stderr: sanitize(
+              `Failed to create post-upgrade commands data file at ${dataFilePath}, reason: ${error.message}`,
+            ),
+          });
+
+          dataFilePath = '';
```
secustor:
```suggestion
          dataFilePath = null;
```

secustor (commented):
else LGTM

rarkins (approved):

mmorhun:
Thank you all!

viceice (approved):

PR #36274, opened 2025-06-02
fix: Update surveyjs monorepo url

alexander7161:
## Changes

Adds https://github.com/surveyjs/survey-creator and https://github.com/surveyjs/survey-library to monorepos to ensure they are updated at the same time.

## Context

Surveyjs updates currently aren't grouped correctly, see screenshot from my dependency dashboard:
image

https://github.com/surveyjs/surveyjs appears to no longer exists and redirects to https://github.com/surveyjs/survey-library

https://github.com/surveyjs/survey-creator/tags and https://github.com/surveyjs/survey-library/tags are always released together with the same version.

### DIFF
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -560,7 +560,11 @@
     "strum": "https://github.com/Peternator7/strum",
     "stryker-js": "https://github.com/stryker-mutator/stryker-js",
     "stylex-swc": "https://github.com/Dwlad90/stylex-swc-plugin",
-    "surveyjs": "https://github.com/surveyjs/survey-library",
+    "surveyjs": [
+      "https://github.com/surveyjs/surveyjs",
+      "https://github.com/surveyjs/survey-library",
+      "https://github.com/surveyjs/survey-creator"
+    ],
     "swashbuckle-aspnetcore": "https://github.com/domaindrivendev/Swashbuckle.AspNetCore",
     "system.io.abstractions": "https://github.com/System-IO-Abstractions/System.IO.Abstractions/",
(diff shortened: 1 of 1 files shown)


lib/data/monorepo.json
```diff
@@ -560,7 +560,7 @@
     "strum": "https://github.com/Peternator7/strum",
     "stryker-js": "https://github.com/stryker-mutator/stryker-js",
     "stylex-swc": "https://github.com/Dwlad90/stylex-swc-plugin",
-    "surveyjs": "https://github.com/surveyjs/surveyjs",
+    "surveyjs": "https://github.com/surveyjs/survey-library",
```
rarkins: We probably want to keep both for some time. How long ago was it renamed?

alexander7161: I believe they have retained the git history so it's difficult to say when it was renamed, but it has definitely been some time.
How would you suggest to keep both? E.g have `surveyjs` and `survey-library`?

Ideally https://github.com/surveyjs/survey-creator would also be included in the group, do you have any idea how that could be possible?

viceice: keep both urls, the older is probably renamed, so it follows with redirect.

the other url should only be included if the must be updated together or they share same release versions.

alexander7161: Thanks, I updated now to group all three under `surveyjs`.

`survey-library` and `survey-creator` share release versions and must be updated together, see their release tags: 
- https://github.com/surveyjs/survey-library/tags
- https://github.com/surveyjs/survey-creator/tags

viceice (approved):

PR #14734, opened 2022-03-20
fix(manager/gitlabci): strip Gitlab Dependecy Proxy variable from depName

dsch:
## Changes

Remove image prefix variable from images pulled through the Dependency Proxy.

## Context

The predefined variables `CI_DEPENDENCY_PROXY_GROUP_IMAGE_PREFIX` and `CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX` can be used to pull images from Docker Hub through Gitlab Dependency Proxy. Since the part after the prefix is a resolvable image name, only this part use used as `depName`. Also the `replaceString` does not contain the variable, since it's not relevant for replacement.

Closes #9958

_Note:_ This change does not solve the two additionally mentioned use-cases in #9958: [CI_REGISTRY](https://github.com/renovatebot/renovate/issues/9958#issuecomment-919866846) and [Dependency Proxy in Dockerfiles](https://github.com/renovatebot/renovate/issues/9958#issuecomment-954967674).

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please tick one)

I have verified these changes via:

- [ ] Code inspection only, or
- [x] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

PR #36293, opened 2025-06-03
fix(manager): Handle variables with default values in registry aliases

Wayneoween:
## Changes

Inspired by https://github.com/renovatebot/renovate/pull/14734 and driven by my own discussion (https://github.com/renovatebot/renovate/discussions/35833) I set out to try and expand upon the already existing `registryAlias` capabilities.

I added code to handle default variables with empty defaults, necessary for how we allow local development at the same time as using defaults for variables in a CI context.

As there was no further response on my discussion I tried to take matters in my own hands and here we are.

## Caveats

[From here](https://github.com/renovatebot/renovate/pull/36293#discussion_r3149446034). The `sanitizeDepName` change in `lib/workers/repository/updates/flatten.ts` strips `${...}` placeholders from every depName, not only ones resolved through `registryAliases`. Any existing depName which happens to contain a literal `${...}` substring will see a onetime change to the generated branch name on the next run.

This unconditional form is necessary because of the `${VAR:-}image:1.0` case (no slash between alias and image). It means the variable can end up in `dep.depName` even on the alias resolution path and we have no clean signal at the sanitiser to tell "applied an alias" apart from "the source name literally has a `$`".

## Context

- https://github.com/renovatebot/renovate/discussions/35833


lib/modules/manager/gitlabci/utils.spec.ts
```diff
@@ -50,6 +50,8 @@ describe('modules/manager/gitlabci/utils', () => {
       ${'multiple aliases'}        | ${{ foo: 'foo.registry.com', bar: 'bar.registry.com' }} | ${'foo/image:1.0'}            | ${{ depName: 'foo/image', packageName: 'foo.registry.com/image', currentValue: '1.0', autoReplaceStringTemplate: `foo/image${versionAndDigestTemplate}` }}
       ${'aliased variable'}        | ${{ $CI_REGISTRY: 'registry.com' }}                     | ${'$CI_REGISTRY/image:1.0'}   | ${{ depName: '$CI_REGISTRY/image', packageName: 'registry.com/image', currentValue: '1.0', autoReplaceStringTemplate: `$CI_REGISTRY/image${versionAndDigestTemplate}` }}
       ${'variables with brackets'} | ${{ '${CI_REGISTRY}': 'registry.com' }}                 | ${'${CI_REGISTRY}/image:1.0'} | ${{ depName: '${CI_REGISTRY}/image', packageName: 'registry.com/image', currentValue: '1.0', autoReplaceStringTemplate: `$\{CI_REGISTRY}/image${versionAndDigestTemplate}` }}
+      ${'variables with default and slash'} | ${{ '${CI_REGISTRY:-}': 'registry.com' }}      | ${'${CI_REGISTRY:-}/image:1.0'} | ${{ currentValue: '1.0' }}
```
RahulGautamSingh: Expand this test to include `depName`, `packageName` and `autoReplaceStringTemplate`.

I think the expected `depName` will be: `${CI_REGISTRY:-}/image`.
But when running the tests I get: `"depName": "${CI_REGISTRY",`

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -179,16 +179,24 @@ export function getDep(
...
   for (const [name, value] of Object.entries(registryAliases ?? {})) {
-    if (currentFrom.startsWith(`${name}/`)) {
-      const depName = currentFrom.substring(name.length + 1);
+    // Check for two possible formats:
+    // 1. Variable followed by slash: "${VAR}/image"
+    // 2. Variable including a slash: "${VAR}image"
+    if (currentFrom.startsWith(`${name}/`) || currentFrom.startsWith(name)) {
+      const depNameStartIndex = currentFrom.startsWith(`${name}/`) ? name.length + 1 : name.length;
+      const depName = currentFrom.substring(depNameStartIndex);
+      const valueWithSlash = value.endsWith('/') ? value : `${value}/`;
```
viceice:
```suggestion
      const valueWithSlash = ensureTrailingSlash(value);
```

lib/modules/manager/gitlabci/utils.spec.ts
viceice: please fill expected `depName`, `packageName` and `autoReplaceStringTemplate`

viceice (commented):
please run `pnpm markdown-lint-fix`

Wayneoween:
I have updated the MR, is there anything left for me to do in order to get another review? :)

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -179,16 +180,35 @@ export function getDep(
...
+    // 2. Variable including a slash: "${VAR}image"
+    if (currentFrom.startsWith(`${name}/`) || currentFrom.startsWith(name)) {
+      const depNameStartIndex = currentFrom.startsWith(`${name}/`)
+        ? name.length + 1
+        : name.length;
+      const depName = currentFrom.substring(depNameStartIndex);
+      const valueWithSlash = ensureTrailingSlash(value);
       const dep = {
-        ...getDep(`${value}/${depName}`, false),
+        ...getDep(`${valueWithSlash}${depName}`, false),
```
RahulGautamSingh:
```suggestion
    const match =
      currentFrom.startsWith(`${name}/`) || currentFrom.startsWith(name);
    if (!match) {
      continue;
    }

    const depName = currentFrom.slice(
      currentFrom.startsWith(`${name}/`) ? name.length + 1 : name.length,
    );
    const resolved = ensureTrailingSlash(value) + depName;
    const dep = { ...getDep(resolved, false)
```

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -179,16 +180,35 @@ export function getDep(
...
         replaceString: currentFrom,
       };
       // retain depName, not sure if condition is necessary
-      if (dep.depName?.startsWith(value)) {
+      if (dep.depName?.startsWith(valueWithSlash)) {
```
RahulGautamSingh:
```suggestion
      if (dep.depName?.startsWith(value)) {
```

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -179,16 +180,35 @@ export function getDep(
...
+          // Split on @ first (for digest), then find the last : (for tag)
+          const [imageAndTag] = currentFrom.split('@');
+          const lastColonIndex = imageAndTag.lastIndexOf(':');
+          if (lastColonIndex > 0) {
+            dep.depName = imageAndTag.substring(0, lastColonIndex);
+          }
+        } else {
+          // No tag/digest, so the entire currentFrom is the depName
+          dep.depName = currentFrom;
+        }
```
RahulGautamSingh:
```suggestion
```

RahulGautamSingh (changes requested):

RahulGautamSingh:
Hey, 

Sorry for the delay. Could you please add the link of the repo you tested you changes on?

Wayneoween:
My org settings at GitLab don't allow me to set repositories to public :|

After more testing I found out that branches with names like `renovate/${dependency_proxy}python-3.x` would have been created. I think my changes are working now, but the way I achieve it is suboptimal.

with this state the tests should be happy and the branches are seemingly also named correctly (e.g. `renovate/python-3.x`).

RahulGautamSingh (changes requested):
Failing Tests

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -179,22 +180,40 @@ export function getDep(
...
-      };
-      // retain depName, not sure if condition is necessary
-      if (dep.depName?.startsWith(value)) {
-        dep.packageName = dep.depName;
-        dep.depName = `${name}/${dep.depName.substring(value.length + 1)}`;
-      }
-      if (specifyReplaceString) {
-        dep.autoReplaceStringTemplate = getAutoReplaceTemplate(dep);
+    const match =
+      currentFrom.startsWith(`${name}/`) || currentFrom.startsWith(name);
```
RahulGautamSingh: This overmatches, please use regex with boundaries, so we have a exact match.

Example:

```
{
    "registryAliases": {
      "$CI_REGISTRY": "registry.example.com",
      "$CI_REGISTRY_IMAGE": "registry.example.com/project"
    }
  }
```

RahulGautamSingh (changes requested):

Wayneoween:
Apologies for the long radio silence, picking this back up now. For transparency: I used Claude Code to help me on this.

**Requested fix (your last review, @RahulGautamSingh).**
The bare-prefix `currentFrom.startsWith(name)` check is now behind the alias key ending in a non-identifier boundary character (typically `}` from `${VAR:-...}`). Concretely:

- `$CI_REGISTRY` ends in `Y` → still requires the `/` boundary, so it can no longer eat `$CI_REGISTRY_IMAGE/...`.
- `${CI_REGISTRY:-}` ends in `}` → both `${VAR:-}/image` and `${VAR:-}image` resolve, which is the use case I needed.

The example from the review is added in as a regression test (`overlapping prefixes` row in both `dockerfile/extract.spec.ts` and `gitlabci/utils.spec.ts`):

```ts
{
  registryAliases: {
    $CI_REGISTRY: 'registry.example.com',
    $CI_REGISTRY_IMAGE: 'registry.example.com/project',
  }
}
// '$CI_REGISTRY_IMAGE/image:1.0' → packageName 'registry.example.com/project/image'
```

**Reproducer repo, as requested:** https://github.com/Wayneoween/renovate-36293-repro

A `Dockerfile` + `renovate.json` (and an optional `.gitlab-ci.yml`) that demonstrates the `${DEPENDENCY_PROXY:-}python:3.13` pattern.

Local verification done via `pnpm vitest --run lib/modules/manager/dockerfile/extract.spec.ts lib/modules/manager/gitlabci/utils.spec.ts lib/workers/repository/updates/flatten.spec.ts` pass.

RahulGautamSingh (commented):
The rename of the regex group `prefix` to `rawPrefix` in the gitlabci utils looks unrelated to this fix. If there is a follow up that needs it, please mention it, otherwise consider dropping the rename to keep the diff focused.

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -179,20 +180,43 @@ export function getDep(
...
-        dep.autoReplaceStringTemplate = getAutoReplaceTemplate(dep);
-      }
-      return dep;
+    // Match the alias key as either:
+    //   - "${name}/..." — bare identifier keys like `$CI_REGISTRY` need a slash
+    //     boundary so they don't overmatch (e.g. `$CI_REGISTRY` vs `$CI_REGISTRY_IMAGE`).
+    //   - "${name}..." — only when `name` already ends in a non-identifier
+    //     boundary character (typically `}` from `${VAR:-...}` defaults), so
+    //     e.g. `${VAR:-}image` resolves correctly without requiring a slash.
+    const matchedWithSlash = currentFrom.startsWith(`${name}/`);
```
RahulGautamSingh, line 186: This boundary check matches any alias key ending in a non-word char, not only `${VAR:-}`. It could do with a trailing `}` so the broadened match stays narrow to the documented case? Also, this regex is rebuilt per alias on every call, hoist to a module scope?

lib/workers/repository/updates/flatten.ts
```diff
@@ -22,6 +22,8 @@ const upper = (str: string): string =>
 
 export function sanitizeDepName(depName: string): string {
   return depName
+    .replace(regEx(/\$\{[^}]+\}/g), '')
+    .replace(regEx(/[${}]/g), '')
```
RahulGautamSingh, line 26: This runs on every depName, not only aliased ones. Anyone whose depName today contains a literal `${...}` will see a one time branch name change. Could you call this out in the PR description?

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -179,20 +180,43 @@ export function getDep(
...
+    const valueWithSlash = ensureTrailingSlash(value);
+    const dep = getDep(`${valueWithSlash}${depName}`, false);
+    // retain depName, not sure if condition is necessary
+    if (dep.depName?.startsWith(valueWithSlash)) {
+      dep.packageName = dep.depName;
+      // Restore the original alias-rooted depName, dropping any tag/digest
+      // suffix. We use the part of currentFrom before `@` so a digest never
+      // leaks into the depName, and only trim the tag colon when the inner
+      // getDep actually parsed a currentValue (otherwise a `:` inside the
+      // alias key, e.g. `${VAR:-}` or `localhost:5000/...`, would be mistaken
```
RahulGautamSingh: This block could be shorter. The test cases make the behavior clear from the code.

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -179,20 +180,43 @@ export function getDep(
...
+      !matchedWithSlash && endsAtBoundary && currentFrom.startsWith(name);
+    if (!matchedWithSlash && !matchedAtBoundary) {
+      continue;
+    }
+    const depName = currentFrom.slice(
+      matchedWithSlash ? name.length + 1 : name.length,
+    );
+    const valueWithSlash = ensureTrailingSlash(value);
+    const dep = getDep(`${valueWithSlash}${depName}`, false);
+    // retain depName, not sure if condition is necessary
```
RahulGautamSingh: A pre-existing edge case worth a TODO here: when the inner `getDep` rewrites something like `docker.io/foo` to `foo`, this check fails and the alias rooted depName is not restored. Not for this PR to fix.

lib/workers/repository/updates/flatten.ts
Wayneoween: Added a `Caveat` section in the PR desc.

RahulGautamSingh (approved):

Wayneoween:
Is there anything left for me to do on this or do I just wait?

Wayneoween:
Hey again, what can I do to support this? I'd like to see this merged :)

lib/workers/repository/updates/flatten.spec.ts
```diff
@@ -22,6 +22,12 @@ describe('workers/repository/updates/flatten', () => {
         'https-some.host.name-a-path-to.git',
       );
     });
+
+    it('strips shell-style variable placeholders', () => {
+      expect(sanitizeDepName('${DEPENDENCY_PROXY}python')).toBe('python');
+      expect(sanitizeDepName('${DEPENDENCY_PROXY:-}python')).toBe('python');
+      expect(sanitizeDepName('$CI_REGISTRY/image')).toBe('ci_registry-image');
```
jamietanna, line 31: In this test, why does this get resolved as the result?

Wayneoween: My intention was to strip the characters that would break branch naming rules.

I think because of my changes [in extract.ts 202-205](https://github.com/renovatebot/renovate/pull/36293/changes/BASE..1a2bc31ba8f2030c2fdb8efd00583bc0a6e9768c#diff-6d282bf444c6345641f0621120c6fd565294d2e34b07fd17ef762092c943114dR202-R205) it's now possible that `$`, `{`, `}` reach this point because we rebuild `depName` from the original `FROM` string.

The examples from the `extract.spec.ts` show this:
  - `$CI_REGISTRY_IMAGE/image:1.0` → `depName: '$CI_REGISTRY_IMAGE/image'`
  - `${CI_REGISTRY:-}image:1.0` → `depName: '${CI_REGISTRY:-}image'`

Am I correct in understanding your question that you ask "Why is this not `toBe('image')`?"
I don't think (it's been a while) that it is strictly necessary to keep the `ci_registry-` part like this.

Wayneoween: Anything else I need to do here? Did I misunderstand?

viceice: @jamietanna wdyt?

viceice (commented):
LGTM beside the open conversation

lib/workers/repository/updates/flatten.spec.ts
Wayneoween: Would love some feedback here, if this is not the right place to update or what you intended woth yout initial question :)

lib/workers/repository/updates/flatten.ts
RahulGautamSingh, line 1: `sanitizeDepName('${CI_REGISTRY:-}/image')` returns `-image`, so the branch name keeps a leading hyphen. 

Make the regex consume a following slash too, `/\$\{[^}]+\}\/?/g`, and add the slash form to the test.

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -179,20 +180,35 @@ export function getDep(
...
+    const matchedWithSlash = currentFrom.startsWith(`${name}/`);
+    const matchedAtBoundary =
+      !matchedWithSlash && name.endsWith('}') && currentFrom.startsWith(name);
+    if (!matchedWithSlash && !matchedAtBoundary) {
+      continue;
+    }
+    const depName = currentFrom.slice(
+      matchedWithSlash ? name.length + 1 : name.length,
+    );
+    const valueWithSlash = ensureTrailingSlash(value);
```
RahulGautamSingh: An empty alias value, the natural config for "empty default means Docker Hub", becomes `getDep('/image:1.0')` and a `packageName` of `/image`. 

Use `value ? ensureTrailingSlash(value) : ''` so an empty value resolves to Docker Hub, and add a test.

lib/modules/manager/dockerfile/readme.md
```diff
@@ -29,6 +29,25 @@ ARG TAG=3.19.4
 FROM alpine:${TAG}
 ```
 
+Renovate supports variables with default values (using the `:-` syntax) when they're configured with registry aliases:
```
RahulGautamSingh:
```suggestion
Renovate supports variables with default values (using the `:-` syntax) when they are configured with registry aliases:
```

lib/modules/manager/docker-compose/readme.md
```diff
@@ -1,3 +1,28 @@
 Extracts all Docker images from with Docker Compose YAML files.
 
 If you need to change the versioning format, read the [versioning](../../versioning/index.md) documentation to learn more.
+
+### Registry Aliases with Variable Defaults
```
RahulGautamSingh:
```suggestion
### Registry aliases with variable defaults
```

lib/modules/manager/docker-compose/readme.md
```diff
@@ -1,3 +1,28 @@
...
+  "registryAliases": {
+    "${CI_REGISTRY:-}": "my-registry.io"
+  }
+}
+```
+
+This works with both formats, regardless of whether the registry alias ends with a slash or not:
+
+- Without slash in the variable: `${CI_REGISTRY:-}/image:1.0`
+- With slash in the variable: `${CI_REGISTRY:-}image:1.0`
```
RahulGautamSingh: The bullets show the two image reference formats, but "whether the registry alias ends with a slash" is about the alias value. 

These are separate facts, state them separately. Same sentence appears in the dockerfile readme.

lib/modules/manager/gitlabci/readme.md
```diff
@@ -18,3 +18,18 @@ If you use predefined GitLab CI variables like `CI_REGISTRY` or `CI_SERVER_FQDN`
   }
 }
 ```
+
+Variables with default values (using the `:-` syntax) are also supported, which is useful for Docker Compose files or scripts:
```
RahulGautamSingh:
```suggestion
Variables with default values (using the `:-` syntax) are also supported:
```

The Docker Compose mention doesn't fit this manager's readme

RahulGautamSingh (changes requested):

lib/workers/repository/updates/flatten.ts
Wayneoween: Oh, thanks for spotting that! Will add your suggestion and add also both `${VAR}/` and `${VAR:-}/` to testing.

lib/modules/manager/dockerfile/extract.ts
Wayneoween: Thank you! Will do!

lib/modules/manager/docker-compose/readme.md
Wayneoween: Fair, thank you!

lib/workers/repository/updates/flatten.spec.ts
Wayneoween: Trying to read into the question what you could have meant isn't easy, so not sure I'm answering the right question here:

`${DEPENDENCY_PROXY}` has a clear end marker, so the whole placeholder gets stripped. `$CI_REGISTRY` has no end marker. There's no way for us to know where the variable name stops and the image name starts. A greedy strip would consume the image name.

So I thought for the bare form we may only drop the `$`. The `/` then becomes `-` like any other path separator which then results in `ci_registry-image`. `sanitizeDepName` feeds `depNameSanitized` which is used for branch names:

| input | before | after |
| --- | --- | --- |
| `$CI_REGISTRY/image` | `$ci_registry-image` | `ci_registry-image` |
| `${CI_REGISTRY:-}/image` | `${ci_registry-}-image` | `image` |
| `${DEPENDENCY_PROXY}python` | `${dependency_proxy}python` | `python` |

With the default `branchNameStrict: false`, those braces ended up in the branch name. `${DEPENDENCY_PROXY}python` resulted in `renovate/${dependency_proxy}python-3.x` and now results in `renovate/python-3.x`.

lib/modules/manager/docker-compose/readme.md
Wayneoween: I also applied the same split to `docs/usage/configuration-options.md` which also had the same points.

lib/modules/manager/docker-compose/readme.md
```diff
@@ -1,3 +1,30 @@
...
 If you need to change the versioning format, read the [versioning](../../versioning/index.md) documentation to learn more.
+
+### Registry aliases with variable defaults
+
+Docker Compose files often use environment variables with default values:
+
+```yaml
+services:
+  foobar:
+    image: ${CI_REGISTRY:-}image:1.0
```
RahulGautamSingh, line 12: The compose spec has no `${VAR:-}` alias test. The dockerfile spec covers the shared `getDep`, so add one compose-level test for this.

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -179,20 +180,36 @@ export function getDep(
...
-        dep.replaceString = currentFrom;
-        dep.autoReplaceStringTemplate = getAutoReplaceTemplate(dep);
-      }
-      return dep;
+    // Allow `${VAR}`/`${VAR:-...}` keys to match without a trailing slash
+    // (`}` is an unambiguous boundary). Bare identifier keys like
+    // `$CI_REGISTRY` still require `/` so they can't eat `$CI_REGISTRY_IMAGE/`.
+    const matchedWithSlash = currentFrom.startsWith(`${name}/`);
+    const matchedAtBoundary =
+      !matchedWithSlash && name.endsWith('}') && currentFrom.startsWith(name);
```
RahulGautamSingh:
```suggestion
    const matchedAtBoundary =
      name.endsWith('}') && currentFrom.startsWith(name);
```

the if already handles `matchedWithSlash`

RahulGautamSingh (changes requested):

viceice (commented):
conflicted

lib/modules/manager/dockerfile/extract.ts
```diff
@@ -180,20 +181,36 @@ export function getDep(
...
+    }
+    const depName = currentFrom.slice(
+      matchedWithSlash ? name.length + 1 : name.length,
+    );
+    // An empty alias value means "no registry prefix", i.e. Docker Hub.
+    const valueWithSlash = value ? ensureTrailingSlash(value) : '';
+    const dep = getDep(`${valueWithSlash}${depName}`, false);
+    // TODO: when the inner getDep strips a `library/` prefix (or similar)
+    // the depName no longer starts with `valueWithSlash` and the alias-rooted
+    // depName is not restored.
```
viceice, line 201: when this will be fixed?

add separate tests, so this is visible to anybody who tries to change the behavior here.

Wayneoween: The behavior this addresses predates this PR. On `main` `$CI_REG/foo:1.0` with `registryAliases: {"$CI_REG": "library"}` gives `depName: foo` instead of `$CI_REG/foo`, because the inner `getDep` strips `library/` as a special prefix and the `startsWith(value)` check then fails. This PRs `${VAR:-}` path behaves the same way.

I added test rows for both so the current behavior is pinned. [Rahul asked for the TODO here](https://github.com/renovatebot/renovate/pull/36293#discussion_r3149446310), since fixing it changes behavior for aliases that I think are unrelated to variable defaults. Happy to open a follow-up PR.

viceice (commented):
failing checks

viceice (approved):
transient CI failure

jamietanna (approved):
Thanks for your patience!

PR #36308, opened 2025-06-04
feat(presets): `@tsconfig/bases` node replacements

setchy:
## Changes

Add new replacement preset to help with [@tsconfig/bases](https://github.com/tsconfig/bases) node updates, both within `package.json` and `tsconfig.json`

Demo repository: https://github.com/setchy/renovate-tsconfig-bases


setchy:
The build began failing once I added the corresponding `customManager` to help keep the `tsconfig.json` file in-sync with the replacement updates.

rarkins (commented):
Won't this start a cascading series of replacements? e.g. 10 -> 12 -> 14 -> 16 -> 18 etc.

setchy:
> Won't this start a cascading series of replacements? e.g. 10 -> 12 -> 14 -> 16 -> 18 etc.

It would, but I figured users could decide if they applied the replacement, declined it, or left it open. 

It would be nice if it could be kept in sync with the repositories node version - or: have these replacements dependant on another dependency version

setchy:
I would still like to find a solution to help automate the `@tsconfig/bases` updates.

Open to other suggestions on ways to implement this.  Are there other packages that follow a similar need?

RahulGautamSingh (commented):
Why would we need the replacements? Won't the custom manager alone be enough

setchy:
> Why would we need the replacements? Won't the custom manager alone be enough

It's been a while since I looked into this, but to my knowledge the replacements were needed as the `depName` itself changes

lib/data/replacements.json
```diff
@@ -1009,6 +1010,90 @@
...
+      {
+        "matchDatasources": ["npm"],
+        "matchPackageNames": ["@tsconfig/node14"],
+        "replacementName": "@tsconfig/node16",
+        "replacementVersion": "16.0.0"
+      },
+      {
+        "matchDatasources": ["npm"],
+        "matchPackageNames": ["@tsconfig/node16"],
+        "replacementName": "@tsconfig/node17",
```
viceice: why node 17? we should only migrate to node lts versions

viceice (commented):
otherwise LGTM

lib/data/replacements.json
setchy: I was just trying to have a replacement chain for each node version currently within https://github.com/tsconfig/bases/tree/main/bases

viceice: ok, but we should only migrate to lts versions. it's ok to have migrations from non-lts to lts is ok

setchy: Sure thing.

So you'd only like to have replacements which target these versions `10`, `12`, `14`, `16`, `18`, `20`, `22` and `24`?

Which means we would not have `17`, `19`, `21` and `23` as "to version"

setchy: Updated to only recommend LTS versions as target

lib/data/replacements.json
```diff
@@ -1009,6 +1010,66 @@
...
+        "replacementVersion": "22.0.0"
+      },
+      {
+        "matchDatasources": ["npm"],
+        "matchPackageNames": ["@tsconfig/node22", "@tsconfig/node23"],
+        "replacementName": "@tsconfig/node24",
+        "replacementVersion": "24.0.0"
+      }
+    ],
+    "customManagers": [
```
setchy: Looks like the schema checks are now stricter than when i first started this PR.

Should I move this part out to the customManagers presets?

What would be the best way to ensure users are using the replacements + custom manager together?

setchy: Moved to a customManager preset

lib/config/presets/internal/custom-managers.ts
```diff
@@ -147,4 +147,20 @@ export const presets: Record<string, Preset> = {
     ],
     description: 'Update `*_version` variables in `.tfvars` files.',
   },
+  tsconfigNodeVersions: {
+    customManagers: [
+      {
+        autoReplaceStringTemplate:
+          '"extends": "@tsconfig/node{{{major}}}/tsconfig.json"',
```
RahulGautamSingh, line 154: Won't this replace string be enough to handle the `depName` change as well .. so we do not need the replacements.

setchy: It isn't... 

We need updates to both
* package.json deps
* tsconfig.json extends statements

setchy: https://github.com/renovatebot/renovate/pull/36308#issuecomment-3628277458

setchy: https://github.com/renovatebot/renovate/pull/36308#issuecomment-3628277458

setchy:
@viceice - does this now make sense, splitting the logic between a set of replacements, and then a customManager for keeping the tsconfig extends updated

viceice (dismissed):

lib/config/presets/internal/custom-managers.ts
```diff
@@ -147,4 +147,20 @@ export const presets: Record<string, Preset> = {
...
+        currentValueTemplate: '{{{major}}}',
+        customType: 'regex',
+        datasourceTemplate: 'npm',
+        managerFilePatterns: ['/(^|/)tsconfig\\.json$/'],
```
secustor:
```suggestion
        managerFilePatterns: ['**/tsconfig.json'],
```
Glob is preferable for readability

secustor (commented):
else LGTM

secustor (approved):

lib/config/presets/internal/custom-managers.ts
```diff
@@ -147,4 +147,20 @@ export const presets: Record<string, Preset> = {
...
   },
+  tsconfigNodeVersions: {
+    customManagers: [
+      {
+        autoReplaceStringTemplate:
+          '"extends": "@tsconfig/node{{{major}}}/tsconfig.json"',
+        currentValueTemplate: '{{{major}}}',
+        customType: 'regex',
+        datasourceTemplate: 'npm',
+        managerFilePatterns: ['**/tsconfig.json'],
```
viceice, line 158: this doesn't match tsconfig in repo root

setchy: Glob vs regex.

Reran in the sample repo and looks OK to me https://github.com/setchy/renovate-tsconfig-bases/pull/18

viceice: interesting, `**/` also matched empty string 🤔

viceice:
@setchy it seems something is wrong 

```
Progress: resolved 1, reused 0, downloaded 0, added 0
 ERR_PNPM_NO_MATCHING_VERSION  No matching version found for @tsconfig/node20@20.0.0 while fetching it from https://npm.xxx.test/

This error happened while installing a direct dependency of /tmp/renovate/repos/forgejo/org/infra

The latest release of @tsconfig/node20 is "20.1.8". Published at 11/18/2025

If you need the full list of all 12 published versions run "$ pnpm view @tsconfig/node20 versions".
```

lib/config/presets/internal/custom-managers.ts
viceice: @setchy it seems this regex manager doesn't work as intended if extends is an array 😕 

- https://github.com/containerbase/base/pull/5622

setchy: Thanks, yes, I hadn't considered that use case

setchy: https://github.com/renovatebot/renovate/pull/40030

PR #36311, opened 2025-06-04
fix(util/host-rules): compare normalized URLs when matching hostRules

Shegox:
## Changes

This PR modifies the matching behavior of `http(s)://` hostRules. Instead of using a basic string comparison, it now creates a URL object from the hostRule and compares the normalized `.href` value.

## Context

This change prevents hostRule confusion and prevents unintended application to incorrect hosts.

For example, `https://test` will no longer match `https://test.example.com` because the `.href` adds a trailing `/`, making the comparison `https://test/` vs. `https://test.example.com/`. 
Similarly, `https://example.com` will not match `https://example.com:8080`, although it will match the standard port (`https://example.com:443`). Consequently, changes were made to the `queue` and `throttle` tests to set domain-level hosts appropriately.

### DIFF
--- a/lib/util/host-rules.ts
+++ b/lib/util/host-rules.ts
@@ -84,8 +84,8 @@ export function matchesHost(url: string, matchHost: string): boolean {
   }
 
   const parsedMatchHost = parseUrl(matchHost);
-  if (isHttpUrl(url) && parsedMatchHost && isHttpUrl(matchHost)) {
-    return parsedUrl.href.startsWith(parsedMatchHost.href);
+  if (isHttpUrl(parsedUrl) && isHttpUrl(parsedMatchHost)) {
+    return parsedUrl.href.startsWith(parsedMatchHost!.href);
   }
 
   const { hostname } = parsedUrl;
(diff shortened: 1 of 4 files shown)


rarkins (dismissed):

lib/util/host-rules.ts
```diff
@@ -78,15 +78,16 @@ export interface HostRuleSearch {
...
-    return url.startsWith(matchHost);
-  }
-
   const parsedUrl = parseUrl(url);
   if (!parsedUrl) {
     return false;
   }
 
+  const parsedMatchHost = parseUrl(matchHost);
+  if (isHttpUrl(url) && parsedMatchHost && isHttpUrl(matchHost)) {
```
viceice:
```suggestion
  if (isHttpUrl(url) && isHttpUrl(parsedMatchHost)) {
```

can be simplified, it will safe some additional parsings in `isHttpUrl` which accepts `unkown` (including `null` or `undefined`)

viceice (commented):
otherwise LGTM

viceice (approved):

PR #36330, opened 2025-06-05
feat(git): gitCredentialPassing

stephan2012:
## Changes

Allow using the Git Credentials Store instead of passing them within the URL by setting `gitCredentialPassing` to `store` (defaults to `url` to maintain the previous behavior). This enables authentication with a Personal Access Token against a Kerberos-enabled GitLab instance.

Please refer to the discussion mentioned below for details.

## Context

- See https://github.com/renovatebot/renovate/discussions/36009


stephan2012:
How do I get started with the tests? 

E.g., the following test fails when I enable the credential store:

```console
 FAIL  lib/modules/platform/gitlab/index.spec.ts > modules/platform/gitlab/index > initRepo > should fall back respecting when GITLAB_IGNORE_REPO_URL is set
Error: Snapshot `modules/platform/gitlab/index > initRepo > should fall back respecting when GITLAB_IGNORE_REPO_URL is set 1` mismatched

- Expected
+ Received

@@ -5,9 +5,9 @@
        "cloneSubmodulesFilter": undefined,
        "defaultBranch": "master",
        "ignorePrAuthor": undefined,
        "mergeMethod": "merge",
        "repository": "some%2Frepo%2Fproject",
-       "url": "http://oauth2:123test@mycompany.com/gitlab/some/repo/project.git",
+       "url": "http://mycompany.com/gitlab/some/repo/project.git",
      },
    ],
  ]

 ❯ lib/modules/platform/gitlab/index.spec.ts:483:39
    481|         repository: 'some/repo/project',
    482|       });
    483|       expect(git.initRepo.mock.calls).toMatchSnapshot();
       |                                       ^
    484|     });
    485|   });
```

docs/usage/self-hosted-configuration.md
rarkins, line 1: Use one sentence per line in markdown files

lib/workers/global/initialize.ts
```diff
@@ -91,6 +92,22 @@ export async function globalInitialize(
...
+    const lines: string[] = [];
+    for (const host of config.hostRules ?? []) {
+      lines.push(`https://oauth2:${host.token}@${host.matchHost}`);
+    }
+    const credentialsPath = upath.join(homedir(), '.git-credentials');
+    await fs.writeFile(credentialsPath, lines.join('\n') + '\n', {
+      mode: 0o600,
+      encoding: 'utf-8',
+    });
+  }
```
rarkins: Move this logic to lib/util/git

rarkins (changes requested):
@viceice perhaps this is best to move the logic to `initPlatform()`?

docs/usage/self-hosted-configuration.md
stephan2012: Done.

lib/workers/global/initialize.ts
stephan2012: Done.

RahulGautamSingh:
> E.g., the following test fails when I enable the credential store:

Meaning when you set `gitCredentialsPassing=store`? If so then yes, some existing test should fail as they expect the default behavior when `gitCredentialsPassing=url`.

You need to add new tests if you want to test the `store` behavior and make sure the old tests pass with `url` behavior

stephan2012:
> Meaning when you set `gitCredentialsPassing=store`? If so then yes, some existing test should fail as they expect the default behavior when `gitCredentialsPassing=url`.

Yes, exactly. When `gitCredentialsPassing` is set to `url`, test results do not change.

> You need to add new tests if you want to test the `store` behavior and make sure the old tests pass with `url` behavior

When I got the docs correctly, changing existing tests shouldn't be in the same PR, although it'd be required to let them pass for `gitCredentialsPassing=store`. How do you think you could deal with it?

stephan2012:
Please let me know what else is required to merge this PR.

docs/usage/self-hosted-configuration.md
```diff
@@ -690,6 +690,12 @@ Renovate will then create branches on the fork and opens Pull Requests on the pa
 !!! note
     Forked repositories will always be skipped when `forkToken` is set, even if `includeForks` is true.
 
+## gitCredentialPassing
+
+The method of passing repository credentials to Git. Uses the URL-embedded credentials by default or when set to `url`, or the [Git Credential Store](https://git-scm.com/docs/git-credential-store) when set to `store`.
```
rarkins:
```suggestion
Uses the URL-embedded credentials by default or when set to `url`, or the [Git Credential Store](https://git-scm.com/docs/git-credential-store) when set to `store`.
```

lib/util/git/index.ts
```diff
@@ -199,6 +200,59 @@ export async function validateGitVersion(): Promise<boolean> {
...
+  } catch (err) {
+    if ((err as NodeJS.ErrnoException).code !== 'ENOENT') {
+      throw err;
+    }
+  }
+
+  const newUrl = new URL.URL(url);
+  let updated = false;
+
+  const result = lines.map((line) => {
```
rarkins, line 230: Add some debug logging to this

docs/usage/self-hosted-configuration.md
```diff
@@ -690,6 +690,12 @@ Renovate will then create branches on the fork and opens Pull Requests on the pa
 !!! note
     Forked repositories will always be skipped when `forkToken` is set, even if `includeForks` is true.
 
+## gitCredentialPassing
```
rarkins, line 728: It looks like the .gitcredentials file in the home directory will be modified by Renovate (and not reverted at the end of a job). Both these should be in the docs

rarkins (changes requested):

docs/usage/self-hosted-configuration.md
stephan2012: Done.

lib/util/git/index.ts
stephan2012: Refactored code now includes debug logging.

docs/usage/self-hosted-configuration.md
stephan2012: Added.

package.json
```diff
@@ -259,6 +259,7 @@
     "url-join": "5.0.0",
     "validate-npm-package-name": "6.0.1",
     "vuln-vects": "1.1.0",
+    "write-file-atomic": "6.0.0",
```
rarkins: Why is a regular fs write not ok? Is there some concurrency to be worried about?

stephan2012: In general, it is not a good practice to overwrite an existing file (assume a power outage during a pending write operation, which will leave a file corrupted); instead, perform an atomic operation so that it remains valid at all times. I assume that the bot mostly runs in a build container, so neither concurrent updates nor race conditions (while reading the file) should occur. However, if the bot runs from the CLI, there may be parallel actions.

stephan2012: POSIX does not make crash consistency guarantees. :)

rarkins: I don't think that such a rare edge case is worth a new dependency

stephan2012: Okay, I've traded safe writes for one less dependency.

docs/usage/self-hosted-configuration.md
```diff
@@ -690,6 +690,12 @@ Renovate will then create branches on the fork and opens Pull Requests on the pa
...
+
+Uses the URL-embedded credentials by default or when set to `url`, or the [Git Credential Store](https://git-scm.com/docs/git-credential-store) when set to `store`. The Renovate Bot creates or updates the `~/.git-credentials` file when it does not contain the GitLab URL with the current username and password. Please note that the bot does not revert the changes when finishing.
```
rarkins:
```suggestion
Uses the URL-embedded credentials by default or when set to `url`, or the [Git Credential Store](https://git-scm.com/docs/git-credential-store) when set to `store`.
The Renovate Bot creates or updates the `~/.git-credentials` file when it does not contain the GitLab URL with the current username and password.
Please note that the bot does not revert the changes when finishing.
```

rarkins (changes requested):

lib/modules/platform/gitlab/index.ts
```diff
@@ -127,6 +129,19 @@ export async function initPlatform({
   const platformConfig: PlatformResult = {
     endpoint: defaults.endpoint,
   };
+
+  if (gitCredentialPassing === 'store') {
+    logger.debug('Enable the Git credential store');
+    await git.enableCredentialStore();
+    const data = new URL.URL(defaults.endpoint);
```
viceice:
```suggestion
    const data = parseUrl(defaults.endpoint);
```
use our url parse helper

lib/modules/platform/gitlab/index.ts
```diff
@@ -127,6 +129,19 @@ export async function initPlatform({
...
+    try {
+      await git.updateCredentialStore(`https://oauth2:${token}@${data.host}`);
+    } catch (err) {
+      logger.fatal({ err }, `Cannot update the Git credentials store: ${err}`);
+      process.exit(1);
```
viceice: throw an error instead, see other auth related throws

lib/modules/platform/gitlab/index.ts
```diff
@@ -263,7 +278,7 @@ export async function getJsonFile(
   return parseJson(raw, fileName);
 }
 
-function getRepoUrl(
+export function getRepoUrl(
```
viceice, line 286: why exporting?

lib/util/git/index.ts
```diff
@@ -199,6 +200,75 @@ export async function validateGitVersion(): Promise<boolean> {
...
+  try {
+    const content = await fs.readFile(credentialsPath, 'utf-8');
+    lines = content.split('\n').filter((line) => line.trim() !== '');
+  } catch (err) {
+    if ((err as NodeJS.ErrnoException).code !== 'ENOENT') {
+      throw err;
+    }
+  }
+
+  const newUrl = new URL.URL(url);
```
viceice: use parse url helper

lib/util/git/index.ts
```diff
@@ -199,6 +200,75 @@ export async function validateGitVersion(): Promise<boolean> {
...
+    'credential.helper',
+    'global',
+  );
+
+  if (!currentHelper.values.includes('store')) {
+    await simpleGit().addConfig('credential.helper', 'store', true, 'global');
+  }
+}
+
+export async function updateCredentialStore(url: string): Promise<void> {
```
viceice: pass as url

lib/util/git/index.ts
```diff
@@ -199,6 +200,75 @@ export async function validateGitVersion(): Promise<boolean> {
...
+    logger.debug(`Adding ${url} to the Git credential store`);
+    result.push(url);
+    storeUpdated = true;
+  }
+
+  if (storeUpdated) {
+    logger.debug(
+      `Updating the Git credential store file at ${credentialsPath}`,
+    );
+    await fs.writeFile(credentialsPath, result.join('\n') + '\n', {
```
viceice: use string template instead of concat

lib/util/git/index.ts
```diff
@@ -199,6 +200,75 @@ export async function validateGitVersion(): Promise<boolean> {
...
+    'global',
+  );
+
+  if (!currentHelper.values.includes('store')) {
+    await simpleGit().addConfig('credential.helper', 'store', true, 'global');
+  }
+}
+
+export async function updateCredentialStore(url: string): Promise<void> {
+  const credentialsPath = upath.join(homedir(), '.git-credentials');
```
viceice, line 215: @rarkins this won't work when home is read only.

rarkins: In that case, try/catch and a fatal error?

stephan2012: The latter write attempt fails in the case of a read-only home, not due to the determination of `credentialsPath`. The `readFile()` is already wrapped in a try/catch block, so adding another one for `writeFile()` should be sufficient. Do you agree?

Anyway, I will also add a note in the docs about write permissions for `~/.git-credentials`.

rarkins: Yes, try/catch on write is enougb

docs/usage/self-hosted-configuration.md
```diff
@@ -690,6 +690,14 @@ Renovate will then create branches on the fork and opens Pull Requests on the pa
 !!! note
     Forked repositories will always be skipped when `forkToken` is set, even if `includeForks` is true.
 
+## gitCredentialPassing
+
+Uses the URL-embedded credentials by default or when set to `url`, or the [Git Credential Store](https://git-scm.com/docs/git-credential-store) when set to `store`.
+The Renovate Bot creates or updates the `~/.git-credentials` file when it does not contain the GitLab URL with the current username and password.
```
rarkins:
```suggestion
The Renovate Bot creates or updates the `~/.git-credentials` file (which much be writeable) when it does not contain the GitLab URL with the current username and password.
```

rarkins (dismissed):

chrisongthb:
Is here something missing for the merge?

lib/modules/platform/gitlab/index.spec.ts
```diff
@@ -14,9 +15,12 @@ import type { BranchStatus } from '../../../types';
 import * as memCache from '../../../util/cache/memory';
 import * as repoCache from '../../../util/cache/repository';
 import type { LongCommitSha } from '../../../util/git/types';
+import type { HttpResponse } from '../../../util/http/types';
 import { toBase64 } from '../../../util/string';
 import * as prBodyModule from '../utils/pr-body';
+import type { RepoResponse } from './types';
 import * as gitlab from '.';
+import { getRepoUrl } from './index';
```
viceice, line 23:
```suggestion
import { getRepoUrl } from '.';
```

lib/modules/platform/gitlab/index.ts
viceice: move to `./utils.ts` in a separate refactor to be merged before this

lib/util/git/index.ts
```diff
@@ -1,4 +1,5 @@
 import URL from 'node:url';
+import { homedir } from 'os';
```
viceice, line 2:
```suggestion
import { homedir } from 'node:os';
```

lib/util/git/index.ts
viceice, line 1: missing test coverage

PR #36340, opened 2025-06-05
feat: add changelogContent property

bdovaz:
## Changes

Add `changelogContent` property

## Context

- https://github.com/renovatebot/renovate/pull/35769#pullrequestreview-2899142419


viceice (commented):
also the worker logic should be in this pr

bdovaz:
> also the worker logic should be in this pr

I understood with your comment that I should only add the fields and nothing else...

In the other PR, it's mixed with the Azure Pipelines data source changes because otherwise I have no way of testing it in a real way... I can run tests, but they will always be unit tests (which already exist), not integration tests.

viceice:
the unit tests are required anyways, so you should be able to test it with them alone.

bdovaz:
> the unit tests are required anyways, so you should be able to test it with them alone.

@viceice I am trying to make the unit tests for the missing coverage but the `lookupUpdates` method has 800 lines so it is not easy to test and it is coupled with data sources, I see that all the tests mock data source responses.....

But `changelogContent` property is something new, if I do not make a data source implement this new property I will not be able to test it... That is why in the previous PR https://github.com/renovatebot/renovate/pull/35769 (that you made me separate in this one) I had mixed with the support of this new property in Azure Pipelines data source, to be able to make the unit test and test it with `pnpm start` apart as I put in the comments on the other PR...

Can you think of a way to test it without adding the Azure Pipelines implementation in the same PR? I can't think of a way

**Also, before I start unit testing, I would like you to give me feedback on whether the changes in lookupUpdates are correct. In my tests in the other PR they are satisfactory but I don't know if the implementation is correct.**

viceice:
you can mock the datasource response like here:

https://github.com/renovatebot/renovate/blob/52bb0564ebf2173255903f751678d273252d68a0/lib/workers/repository/process/lookup/index.spec.ts#L4557-L4574

then check if the result contains the right properties

viceice (commented):
see comment, looks good so fard but needs unit test

bdovaz:
Test added

lib/config/types.ts
```diff
@@ -49,6 +49,7 @@ export interface RenovateSharedConfig {
   commitMessagePrefix?: string;
   commitMessageTopic?: string;
   confidential?: boolean;
+  changelogContent?: string;
```
rarkins: This doesn't seem right here. Content is per upgrade, while these fields are not

lib/modules/datasource/types.ts
```diff
@@ -55,6 +55,7 @@ export interface GetPkgReleasesConfig {
 }
 
 export interface Release {
+  changelogContent?: string;
```
rarkins, line 58: This looks right

rarkins (commented):
What is the relationship here?
- A release can have a content string
- A branch has.. an array of strings? Or a concatenation of them?

lib/config/types.ts
bdovaz: But then why does the field just below `changelogUrl` make sense? Wouldn't the same reasoning apply?

bdovaz:
> What is the relationship here?
> 
> * A release can have a content string
> * A branch has.. an array of strings? Or a concatenation of them?

I understand that with the comments you have posted and the change I have made, this comment is resolved in turn.

bdovaz:
On my side everything is ready and I am waiting for a new review.

viceice (approved):

PR #36341, opened 2025-06-05
feat: add support for .hcl, .hcl.json, .tf.json and .tofu.json

### LINKED ISSUE #11997: Support .tf.json Terraform JSON

Support Terraform's JSON syntax: https://www.terraform.io/docs/language/syntax/json.html

Ideally within the same manager.

Would need to be driven by someone who has this requirement, so far I'm not aware of anyone requesting it.

No

STollenaar:
## Changes

Added support for the terraform json file format. Expanded the parseHCL function to determine if a given file is a .hcl(.json), .tf(.json) of .tofu(.json) and do a JSON parse into HCL before passing that output into the HCL parser, which format is further supported by the code base.

Because a .hcl file can be the actual ".terraform.lock.hcl" file, these are ignored by the file suffix check.

## Context

- Closes #11997


viceice (commented):
I don't think we need all those new fixtures. please reduce to minimum to satisfy coverage

viceice:
I also think think you shouldn't convert to tf and parse again. you should directly transform to the parsed hcl structure.

isn't the hcl lib from terraform supporting those json files out of the box?

viceice:
you can check if there's an additional library to `@cdktf/hcl2json` which support it

STollenaar:
> I don't think we need all those new fixtures. please reduce to minimum to satisfy coverage

Ok, I will remove a few fixtures. I tried to keep the same tests as we have for the regular .tf files to ensure complete coverage for both types

> I also think think you shouldn't convert to tf and parse again. you should directly transform to the parsed hcl structure.
> 
> isn't the hcl lib from terraform supporting those json files out of the box?

so this is a bit tricky. When executing terraform directly it does support the json files. However, the syntax is a bit different. With the `@cdktf/hcl2json` when parsed there are some distinct differences on what the output will be. This is why a simple `JSON.parse` doesn't work as the rest of the renovate code would break.

 Including a few screenshots on the subtle differences:
image
image
image


> you can check if there's an additional library to `@cdktf/hcl2json` which support it

I checked around and I couldn't find a supported node library that would support the json to hcl parse. There are instances of GO packages, but that wouldn't work for this project.

lib/modules/manager/terraform/hcl/index.ts
```diff
@@ -6,12 +6,236 @@ export async function parseHCL(
...
 ): Promise<TerraformDefinitionFile | null> {
   try {
-    return await parse(fileName, content);
+    if (fileName.endsWith('.tf') || fileName.endsWith('.hcl')) {
+      return await parse(fileName, content);
+    } else if (
+      fileName.endsWith('.tf.json') ||
+      fileName.endsWith('.hcl.json')
+    ) {
+      return await parse(fileName, terraformJsonToHcl(JSON.parse(content)));
```
RahulGautamSingh: use `parseJson` from the `util` folder here if it doesn't cause any issue

lib/modules/manager/terraform/hcl/index.ts
```diff
@@ -6,12 +6,236 @@ export async function parseHCL(
...
+  if (Array.isArray(maybeArray)) {
+    return maybeArray;
+  }
+  if (maybeArray === undefined || maybeArray === null) {
+    return [];
+  }
+  return [maybeArray];
+}
+
+function terraformJsonToHcl(terraformJson: any): string {
```
RahulGautamSingh: Put this and its helper fns in a util file

lib/modules/manager/terraform/hcl/index.ts
```diff
@@ -6,12 +6,236 @@ export async function parseHCL(
...
+      return null;
+    }
   } catch /* istanbul ignore next */ {
     return null;
   }
 }
 
-export function parseJSON(content: string): TerraformDefinitionFile | null {
-  return JSON.parse(content);
+// take a JS value and return valid HCL literal)
```
RahulGautamSingh:
```suggestion
// take a JS value and return valid HCL literal
```

viceice:
ok, in this case you should directly convert the Json to the parsed hcl format and don't generate the hcl source code.

lib/modules/manager/terraform/hcl/index.ts
STollenaar: moved into a util file

STollenaar:
> ok, in this case you should directly convert the Jain to the parsed hcl format and don't generate the hcl source code.

From the .tf.json spec you don't really know when something is an array vs an object. in the kubernetes resources there is the metadata block for example. This is seen as an array, but written as an object. There are many of these little nuances that make it harder to go directly from JSON to that parsed hcl format

RahulGautamSingh:
Could you reduce the number of deps extracted in the tests? I can see many deps being extracted and most are similar. Best will be to only keeps deps which are unique somewhat

RahulGautamSingh (commented):
Could you reduce the number of deps extracted in the tests? I can see many deps being extracted and most are similar. Best will be to only keeps deps which are unique somewhat.

If something similar can be done with the fixtures too, please consider doing it. We prefer adding fixtures only when its necessary.

Thanks!

lib/modules/manager/terraform/__fixtures__/modules.tf
```diff
@@ -143,7 +143,7 @@ module "addons_aws" {
 
 
 
-module "relative" {
```
viceice, line 146: please avoid unrelated changes, do in separate PR if it's useful

lib/modules/manager/terraform/__fixtures__/modules.tf.json
```diff
@@ -0,0 +1,129 @@
+{
```
viceice: reduce fixtures to minimum size

viceice:
> > ok, in this case you should directly convert the Jain to the parsed hcl format and don't generate the hcl source code.
> 
> From the .tf.json spec you don't really know when something is an array vs an object. in the kubernetes resources there is the metadata block for example. This is seen as an array, but written as an object. There are many of these little nuances that make it harder to go directly from JSON to that parsed hcl format

I don't agree. we know we only support specific types with specific properties, so we can directly use zod schema to parse and transform the known supported objects and discard all other things.

lib/modules/manager/terraform/__fixtures__/modules.tf.json
STollenaar: I initially resolved this by removing some coverage. But I do want to ask, why do we want to reduce the coverage which would make sure the behavior between the `.tf` and the `.tf.json` remains the same?

STollenaar:
> > > ok, in this case you should directly convert the Jain to the parsed hcl format and don't generate the hcl source code.
> > 
> > 
> > From the .tf.json spec you don't really know when something is an array vs an object. in the kubernetes resources there is the metadata block for example. This is seen as an array, but written as an object. There are many of these little nuances that make it harder to go directly from JSON to that parsed hcl format
> 
> I don't agree. we know we only support specific types with specific properties, so we can directly use zod schema to parse and transform the known supported objects and discard all other things.

I changed it to use the zod schema. Tests pass so I think it does work correctly now. Might need to expand the test cases to cover all the supported terraform resources

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,228 @@
...
+    name: z.string().optional(),
+    task_spec: oneOrMany(DockerTaskSpec).optional(),
+    endpoint_spec: oneOrMany(DockerEndpointSpec).optional(),
+  })
+  .catchall(tfLiteral);
+
+const GenericResourceInstance = z.record(tfLiteral);
+const GenericResourceSchema = z.record(oneOrMany(GenericResourceInstance));
+
+const KubernetesInstance = z.any().transform((orig) => {
```
secustor: Why is this `any`?

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,228 @@
...
+  'metadata',
+  'spec',
+  'template',
+  'job_template',
+  'container',
+  'init_container',
+  'selector',
+]);
+
+function normalise(node: any): any {
```
secustor:
```suggestion
function normalise(node: unknown): unknown {
```
Do not use `any`.

If we are going this route we should do this for every block and not only the Kubernetes ones. 

> When a JSON object property is named after a nested block type, the value of this property represents one or more blocks of that type. The value of the property must be either a JSON object or a JSON array.

https://developer.hashicorp.com/terraform/language/syntax/json#nested-block-mapping

secustor (commented):
[`@cdktf/hcl2json`](https://github.com/hashicorp/terraform-cdk/blob/c7c1144103eae481dd6292a716d88d593e199624/packages/%40cdktf/hcl2json/lib/bridge.ts#L115) directly parses JSON files while converting. 

> From the .tf.json spec you don't really know when something is an array vs an object. in the kubernetes resources there is the metadata block for example. This is seen as an array, but written as an object. There are many of these little nuances that make it harder to go directly from JSON to that parsed hcl format

This is how the spec describes it [in their syntax docs.](https://developer.hashicorp.com/terraform/language/syntax/json#nested-block-mapping)

> When a JSON object property is named after a nested block type, the value of this property represents one or more blocks of that type. The value of the property must be either a JSON object or a JSON array.

To implement this it should be also possible to do a simply normalization from objects to array starting from the second or third level of the JSON objects

STollenaar:
> [`@cdktf/hcl2json`](https://github.com/hashicorp/terraform-cdk/blob/c7c1144103eae481dd6292a716d88d593e199624/packages/%40cdktf/hcl2json/lib/bridge.ts#L115) directly parses JSON files while converting.
> 

If you read that function, it actually only does a JSON.parse on it and then does a deep merge with the `.tf` files. If you were to do that with renovate you get errors since the extractor doesn't recognize that kind of syntax

lib/modules/manager/terraform/hcl/schema.ts
STollenaar: > If we are going this route we should do this for every block and not only the Kubernetes ones.

Agreed, however renovate currently only supports the docker, helm and kubernetes resources for Terraform.

lib/modules/manager/terraform/hcl/schema.ts
STollenaar: sorry, an oversight. Removed for a typed value

secustor:
> > [`@cdktf/hcl2json`](https://github.com/hashicorp/terraform-cdk/blob/c7c1144103eae481dd6292a716d88d593e199624/packages/%40cdktf/hcl2json/lib/bridge.ts#L115) directly parses JSON files while converting.
> 
> If you read that function, it actually only does a JSON.parse on it and then does a deep merge with the `.tf` files. If you were to do that with renovate you get errors since the extractor doesn't recognize that kind of syntax

I have read it and meant this in [context of this comment](https://github.com/renovatebot/renovate/pull/36341#pullrequestreview-3003192986) and it is feasible to me to adapt the extractor logic to expect an array here, if it simplifies the code.

RahulGautamSingh:
Confliced

docs/usage/docker.md
RahulGautamSingh, line 1: Revert

docs/usage/nuget.md
RahulGautamSingh, line 1: Revert

lib/modules/manager/terraform/__fixtures__/modules.tf.json
RahulGautamSingh: We prefer to inline small fixtures wherever possible as it's better for performance. But it should not come at cost of coverage

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,235 @@
...
+export const KubernetesDeployment = KubernetesInstance;
+export const KubernetesDeploymentV1 = KubernetesInstance;
+export const KubernetesJob = KubernetesInstance;
+export const KubernetesJobV1 = KubernetesInstance;
+export const KubernetesPod = KubernetesInstance;
+export const KubernetesPodV1 = KubernetesInstance;
+export const KubernetesReplicationController = KubernetesInstance;
+export const KubernetesReplicationControllerV1 = KubernetesInstance;
+export const KubernetesStatefulSet = KubernetesInstance;
+export const KubernetesStatefulSetV1 = KubernetesInstance;
```
RahulGautamSingh: These are only used in the `TerraformResources` schema i.e. do not need to be exported and in the `KubernetesInstance` you can just use the `KubernetesInstance` schema instead of renaming them all

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,235 @@
...
+export type TerraformProvider = z.infer<typeof TerraformProvider>;
+export type TerraformResources = z.infer<typeof TerraformResources>;
+export type TerraformHelmRelease = z.infer<typeof TerraformHelmRelease>;
+export type TerraformWorkspace = z.infer<typeof TerraformWorkspace>;
+export type TerraformRequiredProvider = z.infer<
+  typeof TerraformRequiredProvider
+>;
+export type TerraformRequiredProviderBlock = z.infer<
+  typeof TerraformRequiredProviderBlock
+>;
```
RahulGautamSingh: not needed if not used anywhere

lib/modules/manager/terraform/extract.spec.ts
```diff
@@ -816,4 +825,587 @@ describe('modules/manager/terraform/extract', () => {
...
+          currentValue: '1.7.8',
+          datasource: 'docker',
+          depName: 'nginx',
+          depType: 'docker_image',
+          replaceString: 'nginx:1.7.8',
+        },
+        {
+          depType: 'docker_image',
+          skipReason: 'invalid-dependency-specification',
+        },
```
RahulGautamSingh, line 1422: no depName?

lib/modules/manager/terraform/extract.spec.ts
```diff
@@ -816,4 +825,587 @@ describe('modules/manager/terraform/extract', () => {
...
+        {
+          autoReplaceStringTemplate:
+            '{{depName}}{{#if newValue}}:{{newValue}}{{/if}}{{#if newDigest}}@{{newDigest}}{{/if}}',
+          datasource: 'docker',
+          depType: 'docker_image',
+          replaceString: '${data.docker_registry_image.ubuntu.name}',
+          skipReason: 'contains-variable',
+        },
```
RahulGautamSingh, line 1430: no depName?

RahulGautamSingh (changes requested):
Also, we should add `.tf.json` to the `managerFilePatterns`

lib/modules/manager/terraform/__fixtures__/kubernetes.tf.json
RahulGautamSingh, line 1: Since kubernetes_cron_job_v1 and others are similar in structure, won't it suffice to test only one?

lib/modules/manager/terraform/extract.spec.ts
STollenaar: It does not return a depName. Same as with [line 479](https://github.com/renovatebot/renovate/blob/237f336e1311ada84a6654961908dfdfad707fe3/lib/modules/manager/terraform/extract.spec.ts#L479)

lib/modules/manager/terraform/extract.spec.ts
STollenaar: It does not return a depName. Same as with [line 471](https://github.com/renovatebot/renovate/blob/237f336e1311ada84a6654961908dfdfad707fe3/lib/modules/manager/terraform/extract.spec.ts#L471)

lib/modules/manager/terraform/__fixtures__/kubernetes.tf.json
STollenaar: that would be correct, but should we also remove those duplicates from the [kuberntes.tf](https://github.com/renovatebot/renovate/blob/237f336e1311ada84a6654961908dfdfad707fe3/lib/modules/manager/terraform/__fixtures__/kubernetes.tf) file?

lib/modules/manager/terraform/hcl/schema.ts
STollenaar: removed

lib/modules/manager/terraform/hcl/schema.ts
STollenaar: removed

STollenaar:
> Also, we should add `.tf.json` to the `managerFilePatterns`

added the `.json` items to this, also updated to include the `.tofu` files

lib/modules/manager/terraform/__fixtures__/kubernetes.tf.json
RahulGautamSingh: Not now as that is not related to change in this PR. You can create a new PR for that

lib/modules/manager/terraform/hcl/index.ts
```diff
@@ -1,17 +1,31 @@
...
+      fileName.endsWith('.hcl.json') ||
+      fileName.endsWith('.tf.json') ||
+      fileName.endsWith('.tofu.json')
+    ) {
+      return TerraformDefinitionFileJSON.parse(
+        parseJson(content, fileName),
+      ) as TerraformDefinitionFile;
+    } else {
+      return null;
+    }
```
RahulGautamSingh: add a test for this case

RahulGautamSingh (changes requested):

lib/modules/manager/terraform/hcl/index.ts
STollenaar: added

RahulGautamSingh (approved):
LGTM

viceice (commented):
looks good but still too many fixtures which will slow down tests too much

STollenaar:
> looks good but still too many fixtures which will slow down tests too much

Is this latest change to inline fixtures better?

lib/modules/manager/terraform/hcl/index.spec.ts
```diff
@@ -109,6 +105,30 @@ describe('modules/manager/terraform/hcl/index', () => {
 
   describe('parseHCL() for .tf.json', () => {
     it('should return flat modules', async () => {
+      const modulesTFJSON = JSON.stringify({
```
RahulGautamSingh: use `codeBlock`...

You can check other test files for example

RahulGautamSingh (changes requested):

lib/modules/manager/terraform/hcl/index.spec.ts
STollenaar: ok. changed it all to codeblocks

RahulGautamSingh (approved):

RahulGautamSingh (commented):
conflicted

STollenaar:
> conflicted

resolved the conflicts

lib/modules/manager/terraform/hcl/index.spec.ts
```diff
@@ -97,13 +104,173 @@ describe('modules/manager/terraform/hcl/index', () => {
...
+            source: 'hashicorp/consul/aws',
+            version: '0.1.0',
+          },
+        },
+      }
+      `;
+      const res = await parseHCL(modulesTFJSON, 'file.tf.json');
+      expect(res?.module).toBeDefined();
+      expect(Object.keys(res!.module!)).toBeArrayOfSize(6);
       expect(res).toMatchObject({
```
RahulGautamSingh:
```suggestion
      expect(res).toEqual({
```

Then we won't need the `Object.keys` assertion above

lib/modules/manager/terraform/hcl/index.ts
```diff
@@ -1,17 +1,31 @@
...
+    ) {
+      return await parse(fileName, content);
+    } else if (
+      fileName.endsWith('.hcl.json') ||
+      fileName.endsWith('.tf.json') ||
+      fileName.endsWith('.tofu.json')
+    ) {
+      return TerraformDefinitionFileJSON.parse(
+        parseJson(content, fileName),
+      ) as TerraformDefinitionFile;
```
RahulGautamSingh: We should derive the `TerraformDefinitionFile` from the schema `TerraformDefinitionFileJSON`. 

As such:
`type TerraformDefinitionFile = z.infer<typeof TerraformDefinitionFileJSON>;`


Maintaining two sources for same type def is not good.

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,183 @@
+import { z } from 'zod/v3';
+
+type TfLiteral =
```
RahulGautamSingh: This should ne derived from the schema for tfLiteral

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,183 @@
...
+});
+
+const TerraformWorkspace = z.object({
+  terraform_version: z.string().optional(),
+});
+
+const TerraformWorkspaceArray = oneOrMany(TerraformWorkspace);
+
+const DockerSimpleInstance = z.record(tfLiteral);
+const DockerContainerSpec = z.record(tfLiteral);
```
RahulGautamSingh: Why two diff variables for the same schema

RahulGautamSingh (changes requested):

lib/modules/manager/terraform/hcl/schema.ts
STollenaar: I am unsure how to do this cleanly. Since the tfLiteral below has an recursive type call for it

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    module: z.record(oneOrMany(TerraformModule)).optional(),
+
+    resource: TerraformResources.optional(),
+
+    data: z.record(tfLiteral).optional(),
+
+    provider: z.record(oneOrMany(TerraformProvider)).optional(),
+  })
+  .strict();
```
RahulGautamSingh: `.strict()` will reject any `.tf.json` that contains valid top-level Terraform JSON keys not in the schema, like `variable`, `output`, `locals`, `moved`, `check`, or `import`. Combined with the swallowed `catch` in `parseHCL`, those files are silently dropped. either drop `.strict()` and rely on the per-key shapes, or add the missing top-level keys.

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,182 @@
...
+    z.string(),
+    z.number(),
+    z.boolean(),
+    z.null(),
+    z.array(oneOrMany(tfLiteral)),
+    z.record(tfLiteral),
+  ]),
+);
+
+const NORMALISE_KEYS = new Set([
```
RahulGautamSingh: `NORMALISE_KEYS` and the `normalise` function below use British spelling. project style is American English, use `NORMALIZE_KEYS` and `normalize`.

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,182 @@
...
+      }
+      out[k] = val;
+    }
+    return out;
+  }
+
+  return node;
+}
+
+const oneOrMany = <T extends z.ZodTypeAny>(
```
RahulGautamSingh: prefer a `function oneOrMany<T extends z.ZodTypeAny>(schema: T)` declaration over a `const` arrow.

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,182 @@
...
+  'job_template',
+  'container',
+  'init_container',
+  'selector',
+]);
+
+function normalise(node: TfLiteral): TfLiteral {
+  if (Array.isArray(node)) {
+    // Recursively flatten nested arrays
+    const flatten = (arr: TfLiteral[]): TfLiteral[] =>
```
RahulGautamSingh: `const flatten = (arr) => ...` should be a nested `function flatten(arr: TfLiteral[]): TfLiteral[]` declaration. same reason as above, also better stack traces.

lib/modules/manager/terraform/index.ts
```diff
@@ -20,7 +20,14 @@ export const categories: Category[] = ['iac', 'terraform'];
 
 export const defaultConfig = {
   commitMessageTopic: 'Terraform {{depName}}',
-  managerFilePatterns: ['**/*.tf', '**/*.tofu'],
+  managerFilePatterns: [
+    '**/*.hcl',
```
RahulGautamSingh, line 25: this PR adds `**/*.hcl` and `**/*.hcl.json` to the file patterns, but the title and body only mention `.tf.json`. matching `**/*.hcl` will pull in `.terraform.lock.hcl` lockfiles, which the regular extractors are not meant to process. either narrow this to `.tf.json`/`.tofu.json` only, or call out the `.hcl` expansion in the description and verify lockfile handling.

lib/modules/manager/terraform/hcl/index.spec.ts
```diff
@@ -1,13 +1,20 @@
...
+import { parseHCL } from './index.ts';
 
 const modulesTF = Fixtures.get('modules.tf');
 const resourcesTF = Fixtures.get('resources.tf');
-const resourcesTFJSON = Fixtures.get('resources.tf.json');
 const lockedVersion = Fixtures.get('lockedVersion.tf');
 
 describe('modules/manager/terraform/hcl/index', () => {
-  describe('parseHCL()', () => {
+  describe('null for invalid name', () => {
```
RahulGautamSingh: this top-level `describe` is unrelated to `parseHCL()`. either nest it inside the `parseHCL()` block, or rename it to something like `describe(`unsupported file extension`, ...)` so the intent is clear.

lib/modules/manager/terraform/hcl/index.ts
```diff
@@ -1,17 +1,29 @@
...
+    } else if (
+      fileName.endsWith('.hcl.json') ||
+      fileName.endsWith('.tf.json') ||
+      fileName.endsWith('.tofu.json')
+    ) {
+      return TerraformDefinitionFileJSON.parse(parseJson(content, fileName));
+    } else {
+      return null;
+    }
   } catch {
```
RahulGautamSingh: the swallowed `catch` hides zod validation failures. with the new schema this masks real bugs, see the `.strict()` comment on schema.ts. add a `logger.debug({ err, packageFile: fileName }, ...)` so users can see why parsing failed.

lib/modules/manager/terraform/extract.spec.ts
```diff
@@ -816,4 +825,550 @@ describe('modules/manager/terraform/extract', () => {
...
+          "data": {
+            "sops_file": {
+              "secrets": {
+                "source_file": "\${path.module}/secrets.enc.json"
+              }
+            }
+          }
+        }
+        `;
+
```
RahulGautamSingh, line 977: typo: double space in `extracts  modules`.

lib/modules/manager/terraform/hcl/types.ts
```diff
@@ -1,10 +1,9 @@
...
-  provider?: Record<string, TerraformProvider[]>;
-}
+import type z from 'zod/v3';
+import type { TerraformDefinitionFileJSON } from './schema.ts';
+
+export type TerraformDefinitionFile = z.infer<
+  typeof TerraformDefinitionFileJSON
+>;
 
 export interface TerraformBlock {
```
RahulGautamSingh: `TerraformDefinitionFile` is now derived from the zod schema, but `TerraformBlock`, `TerraformResources`, `TerraformProvider`, etc. below are still hand-written and now diverge from the schema (for example the schema accepts `string` as a value in `TerraformRequiredProviderBlock`). either drop these and infer them from `schema.ts`, or remove the ones that are no longer used.

lib/modules/manager/terraform/hcl/schema.ts
RahulGautamSingh: recursive zod v3 schemas cannot have their type inferred, the cycle is unavoidable. the `z.ZodType<TfLiteral>` annotation pattern already in use is the standard workaround.

you can still reduce duplication by sourcing the non-recursive primitives from the schema, then keep the recursive shape manual:

```ts
const tfPrimitive = z.union([z.string(), z.number(), z.boolean(), z.null()]);
type TfPrimitive = z.infer<typeof tfPrimitive>;

type TfLiteral = TfPrimitive | { [k: string]: TfLiteral } | TfLiteral[];

const tfLiteral: z.ZodType<TfLiteral> = z.lazy(() =>
  z.union([tfPrimitive, z.record(tfLiteral), z.array(tfLiteral)]),
);
```

that removes the `string | number | boolean | null` duplication.

STollenaar: ah, yes that makes sense after that explanation

lib/modules/manager/terraform/hcl/types.ts
STollenaar: removed the old schema

lib/modules/manager/terraform/hcl/index.ts
STollenaar: added the debug logger

lib/modules/manager/terraform/hcl/index.spec.ts
STollenaar: moved this into the parseHCL describe and changed it to an it

lib/modules/manager/terraform/index.ts
STollenaar: updated the description and file suffix handler

lib/modules/manager/terraform/hcl/schema.ts
STollenaar: dropped the `.strict()`. the mentioned top-level Terraform keys are not required for renovate to check

lib/modules/manager/terraform/hcl/types.ts
STollenaar: > removed the old schema

reverted this, and have them be inferred from `schema.ts`

RahulGautamSingh (commented):
Failing coverage

STollenaar:
> Failing coverage

I have resolved the coverage issues, but for some reason the oxlint is now complaining about zod/v3. Is this a new change that it needs to be updated to v4?

secustor:
> > Failing coverage
> 
> I have resolved the coverage issues, but for some reason the oxlint is now complaining about zod/v3. Is this a new change that it needs to be updated to v4?

Yes, we migrated to `zod/v4` in the meantime

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,219 @@
+import { z } from 'zod/v4';
+
+const tfPrimitive = z.union([z.string(), z.number(), z.boolean(), z.null()]);
+type tfPrimitive = z.infer<typeof tfPrimitive>;
```
RahulGautamSingh:
```suggestion
type TfPrimitive = z.infer<typeof tfPrimitive>;
```

RahulGautamSingh (commented):
@viceice  keep `.hcl` or narrow to `.tf.json`/`.tofu.json` wdyt?

lib/modules/manager/terraform/index.ts
secustor: Do you have a reference that Terraform / OpenTofu support these `.hcl` or `.hcl.json` file extensions. 
With a quick search I couldn't find anything

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,219 @@
...
+  'spec',
+  'template',
+  'job_template',
+  'container',
+  'init_container',
+  'selector',
+]);
+
+function normalize(node: TfLiteral): TfLiteral {
+  if (Array.isArray(node)) {
```
secustor: Use the `is` helpers

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,219 @@
...
+    // Recursively flatten nested arrays
+    function flatten(arr: TfLiteral[]): TfLiteral[] {
+      return arr.flatMap((item) =>
+        Array.isArray(item) ? flatten(item) : [normalize(item)],
+      );
+    }
+    return flatten(node);
+  }
+
+  if (node && typeof node === 'object') {
```
secustor:
```suggestion
  if (is.plainObject(node)) {
```

RahulGautamSingh (changes requested):
@STollenaar can you fix the conflict and the failing lint

STollenaar:
> @STollenaar can you fix the conflict and the failing lint

resolved these issues

lib/modules/manager/terraform/index.ts
RahulGautamSingh: Terraform and OpenTofu load `.tf`, .`tf.json`, `.tofu`, `.tofu.json` only. `.hcl` is not a Terraform config extension. 

These two patterns also collide with existing managers:
- `terragrunt.hcl` → terragrunt manager (`/(^|/)terragrunt\.hcl$/`)
- `.tflint.hcl` → tflint-plugin manager (`/\.tflint\.hcl$/`)
- `.terraform.lock.hcl`` → already in `lockFileNames`

Every terragrunt repo would get duplicate deps with the wrong `depType`. 

Drop both patterns; that also removes the need for the `.lock.hcl` guard in `parseHCL`, which is filtering in the wrong layer anyway.

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,220 @@
...
+);
+
+const KubernetesInstance = tfLiteral.transform((orig) => {
+  const n = normalize(orig);
+  return isArray(n) ? n : [n];
+});
+
+const TerraformResources = z
+  .object({
+    helm_release: z.record(z.string(), TerraformHelmRelease).optional(),
```
RahulGautamSingh: This is missing `oneOrMany`, valid files get silently dropped.

Every other resource uses `oneOrMany`, `helm_release` does not. Terraform JSON allows a block to be written as an array, so this is valid and currently parses to null:

lib/modules/manager/terraform/hcl/index.spec.ts
RahulGautamSingh, line 1: All five `codeBlock` inputs use unquoted keys and single quotes. `jsonc-weaver` rejects them, so they only parse via the JSON5 fallback in `parseJson`, which fires `logger.warn('...Support for this will be removed in a future release...')`.

 Real `.tf.json` is strict JSON, so these tests check a deprecated path rather than the target format.

lib/modules/manager/terraform/hcl/index.ts
```diff
@@ -1,17 +1,31 @@
...
+): Promise<TerraformDefinitionFileJSON | null> {
   try {
-    return await parse(fileName, content);
-  } catch {
+    if (
+      (fileName.endsWith('.hcl') && !fileName.endsWith('.lock.hcl')) ||
+      fileName.endsWith('.tf') ||
+      fileName.endsWith('.tofu')
+    ) {
+      return await parse(fileName, content);
```
RahulGautamSingh: `parse()` returns `Promise<Record<string, any>>`, assigned straight to `TerraformDefinitionFileJSON`. 

The type system is bypassed. It also happens to be wrong: `hcl2json` yields `helm_release: Record<string, TerraformHelmRelease[]>` while the schema type says `Record<string, TerraformHelmRelease>`. 

Only `Object.values(...).flat()` in the extractor hides it.

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    docker_image: z
+      .record(z.string(), oneOrMany(DockerSimpleInstance))
+      .optional(),
+
+    docker_service: z
+      .record(z.string(), oneOrMany(DockerServiceInstance))
+      .optional(),
+    kubernetes_cron_job_v1: z
+      .record(z.string(), oneOrMany(KubernetesInstance))
```
RahulGautamSingh: from 159 - 201

`KubernetesInstance` already ends in `.transform()` that returns an array, so `oneOrMany`'s wrap is always a no-op and its `z.array` branch is unreachable. 

Use `KubernetesInstance` directly, repeated 14 times.

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,220 @@
...
+  'job_template',
+  'container',
+  'init_container',
+  'selector',
+]);
+
+function normalize(node: TfLiteral): TfLiteral {
+  if (isArray(node)) {
+    // Recursively flatten nested arrays
+    function flatten(arr: TfLiteral[]): TfLiteral[] {
```
RahulGautamSingh: no nested functions only when the scope is used. Hoist it to module level.

lib/modules/manager/terraform/hcl/schema.ts
```diff
@@ -0,0 +1,220 @@
...
+  module: z.record(z.string(), oneOrMany(TerraformModule)).optional(),
+
+  resource: TerraformResources.optional(),
+
+  data: z.record(z.string(), tfLiteral).optional(),
+
+  provider: z.record(z.string(), oneOrMany(TerraformProvider)).optional(),
+});
+
+export type TerraformDefinitionFileJSON = z.infer<
```
RahulGautamSingh: It types the output of both the HCL and the JSON path. `TerraformDefinitionFile` reads better than `TerraformDefinitionFileJSON`

RahulGautamSingh (changes requested):
@STollenaar 

1. [this comment](https://github.com/renovatebot/renovate/pull/36341#discussion_r3574416435) from @secustor is still unadressed
2. there are still 7 fixtures, we asked them to be reduced eg. lockedVersion.tf.json and tfeWorkspace.tf.json are small enough to inline as codeBlock.
3. readme does not mention JSON syntax support, and the PR body checks "No documentation update is required". A short note that `.tf.json` and `.tofu.json` are supported is needs documentation there.

viceice (approved):
long way, thanks for patience

viceice:
please resolve open conversations

STollenaar:
> please resolve open conversations

all should be resolved now

PR #36370, opened 2025-06-07
feat(bazel): Add tag support for `git_repository` rules

zharinov:


rarkins (commented):
Is there such a thing as tag but no digest supported?

zharinov:
According to docs, both `commit` and `tag` are optional, so I think yes

rarkins:
Can you add a test case for that to make sure we support it?

rarkins (approved):

PR #36371, opened 2025-06-07
fix(ci): Skip coverage jobs when tests are failed

zharinov:
## Changes

When unit tests fail, we also see the coverage failure, which makes it harder to track down the problem.

This PR replaces the `success() || failure()` conditions to just `success()` ones.

The _"catch all"_ job named `test-success` still makes sure none of the required steps is missing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -389,7 +389,7 @@ jobs:
           done
 
       - name: Move coverage files
-        if: (success() || failure()) && github.event.pull_request.draft != true && matrix.coverage
+        if: github.event.pull_request.draft != true && matrix.coverage
         run: |
           mkdir -p ./coverage/lcov
           mkdir -p ./coverage/json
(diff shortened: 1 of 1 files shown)


zharinov:
It's so tempting to remove these conditions, but I'm afraid it could lead to unexpected consequences in the `test-success` job, due to how "skipped" result works. I model Github Actions as something following the principle of maximum surprise.

.github/workflows/build.yml
```diff
@@ -412,7 +412,7 @@ jobs:
     needs: [test]
     runs-on: ubuntu-latest
     timeout-minutes: 3
-    if: (success() || failure()) && github.event_name != 'merge_group' && github.event.pull_request.draft != true
+    if: success() && github.event_name != 'merge_group' && github.event.pull_request.draft != true
```
viceice, line 415: will codecov set an error state if no coverage was send? 🤔

viceice (approved):

.github/workflows/build.yml
zharinov: Codecov sends the check from their side, and it's required on our side. If the sending silently fails on the CI side, we just don't get the status from Codecov.

PR #36412, opened 2025-06-10
fix(manager/cargo): Revert "feat(manager/cargo): support git dependencies"

rarkins:
Reverts renovatebot/renovate#36289

Reverted due to two regressions found:
- https://github.com/renovatebot/renovate/discussions/36397
- https://github.com/renovatebot/renovate/discussions/36408

PR #36444, opened 2025-06-11
feat(manager/cargo): support git dependencies

### LINKED ISSUE #26531: feat(manager/cargo): support for git dependencies
### Discussed in https://github.com/renovatebot/renovate/discussions/26521

<sup>Originally posted by **richvdh** January  5, 2024</sup>
### Tell us more.

https://github.com/richvdh/renovate-cargo-git-test is an example of a Cargo project with a git dependency. [`Cargo.lock`](https://github.com/richvdh/renovate-cargo-git-test/blob/main/Cargo.lock#L8) is out of date, and can be updated with `cargo update -p empty-git-lib`.

Renovate reports:

```json
{
  "baseBranch": "main"
  "config": {
    "cargo": [
      {
        "deps": [
          {
            "currentValue": "0.1.0",
            "datasource": "crate",
            "depName": "empty-git-lib",
            "depType": "dependencies",
            "lockedVersion": "0.1.0",
            "managerData": {
              "nestedVersion": true
            },
            "packageName": "empty-git-lib",
            "skipReason": "git-dependency",
            "updates": []
          }
        ],
        "lockFiles": [
          "Cargo.lock"
        ],
        "packageFile": "Cargo.toml"
      }
    ]
  }
}
```

My understanding, looking at the [source](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/cargo/extract.ts#L49-L92), is that Renovate will refuse to touch any dependency that comes from a git repo. It would be great if support could be added!</div>

Reproduction: https://github.com/renovate-reproductions/26531

okkero:
## Changes

This PR reintroduces Cargo git dependencies, reverted in #36412. The [original PR](https://github.com/renovatebot/renovate/pull/36289) adding support for git dependencies also introduced two regressions. This PR addresses both:

- #36397 is addressed by not using `cargo` versioning by default for git dependencies, instead making a best effort decision between `semver` and `loose` based on the current value's shape.
- #36408 is addressed by not trying to read `currentValue` if it is not defined, effectively skipping git dependencies without a specified ref, leaving it up to the lock file maintenance schedule to update the lock file.

## Context

- Closes #26531 


lib/modules/manager/cargo/extract.ts
```diff
@@ -284,10 +284,15 @@ export async function extractPackageFile(
     for (const dep of deps) {
       const packageName = dep.packageName ?? dep.depName!;
       const versions = coerceArray(versionsByPackage.get(packageName));
-      const lockedVersion = versioning.getSatisfyingVersion(
-        versions,
-        dep.currentValue!,
-      );
+
+      let lockedVersion;
```
viceice:
```suggestion
      let lockedVersion: string | null =  null;
```

lib/modules/manager/cargo/index.ts
```diff
@@ -20,4 +24,10 @@ export const defaultConfig = {
   versioning: cargoVersioning.id,
```
viceice:
```suggestion
```
i think we need to drop this, this will set cargo vasioning for all deps, the datasource can handle the default:

https://github.com/renovatebot/renovate/blob/f479c33974f80e3932e81483c036eb8d7747b5b1/lib/modules/datasource/crate/index.ts#L40

lib/modules/manager/cargo/schema.ts
```diff
@@ -56,6 +60,21 @@ const CargoDep = z.union([
...
+          skipReason = 'path-dependency';
+        } else if (workspace) {
+          skipReason = 'inherited-dependency';
+        } else if (git) {
+          applyGitSource(dep, git, rev, tag, branch);
+          if (semver.parse(dep.currentValue)) {
+            dep.versioning = semanticVersioning.id;
+          } else {
+            dep.versioning = looseVersioning.id;
+          }
```
viceice: @rarkins I wonder if we really should set versioning here. 🤔 

I would suggest to drop the manager default versioning (see other comment) and let it use the defalt database versioning with the `semver-coerce` fallback.

lib/modules/manager/util.ts
```diff
@@ -21,7 +21,9 @@ export function applyGitSource(
           ? GithubTagsDatasource.id
           : GitlabTagsDatasource.id;
       const { protocol, source, full_name } = parseGitUrl(git);
-      dep.registryUrls = [`${protocol}://${source}`];
+      dep.registryUrls = [
+        `${protocol === 'ssh' ? 'https' : protocol}://${source}`,
```
viceice: wouldn't this cause broken urls if you simply change the protocol?

lib/modules/manager/cargo/index.ts
```diff
@@ -1,5 +1,9 @@
 import type { Category } from '../../../constants';
 import { CrateDatasource } from '../../datasource/crate';
+import { GitRefsDatasource } from '../../datasource/git-refs';
+import { GitTagsDatasource } from '../../datasource/git-tags';
+import { GithubTagsDatasource } from '../../datasource/github-tags';
+import { GitlabTagsDatasource } from '../../datasource/gitlab-tags';
 import * as cargoVersioning from '../../versioning/cargo';
```
viceice:
```suggestion
```

see below

lib/modules/manager/cargo/artifacts.ts
```diff
@@ -120,14 +121,24 @@ async function updateArtifactsImpl(
...
-          'Missing locked version for dependency',
-        );
+      const nonCrateDep = updatedDeps.find(
+        (dep) => dep.datasource !== CrateDatasource.id,
+      );
+      const crateDepWithoutLockedVersion = updatedDeps.find(
+        (dep) => !dep.lockedVersion && dep.datasource === CrateDatasource.id,
+      );
+      // Non-crate dependencies (like git ones) do not have locked versions.
+      // For crate dependencies, not having a locked version is not expected.
```
RahulGautamSingh:
```suggestion
      // For crate dependencies, a locked version is expected.
```

lib/modules/manager/cargo/schema.ts
```diff
@@ -1,8 +1,12 @@
+import semver from 'semver';
 import { z } from 'zod';
 import type { SkipReason } from '../../../types';
 import { Toml, withDepType } from '../../../util/schema-utils';
 import { CrateDatasource } from '../../datasource/crate';
+import * as looseVersioning from '../../versioning/loose';
+import * as semanticVersioning from '../../versioning/semver';
```
RahulGautamSingh:
```suggestion
import * as semverVersioning from '../../versioning/semver';
```

Lets use consistent naming

lib/modules/manager/cargo/artifacts.ts
```diff
@@ -120,14 +121,24 @@ async function updateArtifactsImpl(
...
+      // In both situations, perform a regular workspace lockfile update.
+      if (nonCrateDep || crateDepWithoutLockedVersion) {
+        if (crateDepWithoutLockedVersion) {
+          // Only warn when a crate dependency has no locked version, as this is
+          // not an expected situation.
```
RahulGautamSingh:
```suggestion
          // Only warn when a crate dependency has no locked version
```

lib/modules/manager/util.ts
okkero: My thought process was this:

Cargo (maybe others as well) allows you to use ssh protocol for git dependencies (see https://doc.rust-lang.org/cargo/appendix/git-authentication.html#ssh-authentication). However, if you do so, that's still not the correct protocol for the registry URL (that would be `https`). Running on my test repository failed without this, so I thought coercing `ssh` to `https` was a sane default in this case.

However, I'm really unsure about the best approach to this, and I'm open to suggestions. Maybe something should be done at the cargo manager level instead...?

lib/modules/manager/cargo/schema.ts
okkero: The problem was this #36397; `cargo` versioning makes little sense for git dependencies. But this is perhaps not the correct place to do this?

I'm looking through the code to try and make sense of how this all fits together, but there's a lot, so maybe you could help me? I'm trying to understand how versioning is selected based on manager vs config, which values take precedence over others, and where in the code I might look to see that logic. Would you be able to point me in the right direction?

lib/modules/manager/cargo/schema.ts
```diff
@@ -56,6 +60,21 @@ const CargoDep = z.union([
...
+          skipReason = 'path-dependency';
+        } else if (workspace) {
+          skipReason = 'inherited-dependency';
+        } else if (git) {
+          applyGitSource(dep, git, rev, tag, branch);
+          if (semver.parse(dep.currentValue)) {
+            dep.versioning = semverVersioning.id;
+          } else {
+            dep.versioning = looseVersioning.id;
+          }
```
RahulGautamSingh:
```suggestion
            dep.versioning = semver.parse(dep.currentValue)
                 ? semverVersioning.id
                 : looseVersioning.id;
```
cleaner

lib/modules/manager/cargo/artifacts.ts
```diff
@@ -120,14 +121,23 @@ async function updateArtifactsImpl(
...
     } else {
-      const missingDep = updatedDeps.find((dep) => !dep.lockedVersion);
-      if (missingDep) {
-        // If there is a dependency without a locked version then log a warning
-        // and perform a regular workspace lockfile update.
-        logger.warn(
-          { dependency: missingDep.depName },
-          'Missing locked version for dependency',
-        );
+      const nonCrateDep = updatedDeps.find(
```
RahulGautamSingh:
```suggestion
      const hasNonCrateDep = updatedDeps.some(
```
Since, we only need to check presence

RahulGautamSingh (commented):
LGTM, except the comments and the protocol change situation

lib/modules/manager/cargo/schema.ts
RahulGautamSingh: You can remove this if-check and also remove the ` versioning: cargoVersioning.id,` from the `defaultConfig` in the `manager/cargo/index.ts` file.

Now, the versioning for each dependency will be its datasource's default versioning.

You can try this in your test repos, and if all works fine and the regressions are fixed. We can use this approach.

lib/modules/manager/util.ts
RahulGautamSingh: This helper function can be useful here:
https://github.com/renovatebot/renovate/blob/03542cb78972d3b19efc25a6c89d83437dd1ccaf/lib/util/git/url.ts#L11

Btw, if you have difficulty understanding something, you can check the `poetry` and `vendir` managers. We extract git dependencies there.

lib/modules/manager/cargo/artifacts.ts
```diff
@@ -120,14 +121,23 @@ async function updateArtifactsImpl(
...
+        (dep) => dep.datasource !== CrateDatasource.id,
+      );
+      const crateDepWithoutLockedVersion = updatedDeps.find(
+        (dep) => !dep.lockedVersion && dep.datasource === CrateDatasource.id,
+      );
+      // Non-crate dependencies (like git ones) do not have locked versions.
+      // For crate dependencies, a locked version is expected.
+      // In both situations, perform a regular workspace lockfile update.
+      if (nonCrateDep || crateDepWithoutLockedVersion) {
```
okkero: connected to above suggestion

```suggestion
      if (hasNonCrateDep || crateDepWithoutLockedVersion) {
```

lib/modules/manager/util.ts
okkero: Thank you, that's very helpful

lib/modules/manager/cargo/schema.ts
okkero: This seems to work. It does not try to pin git dependencies, so the regression is still fixed, and everything else still seems to work.

lib/modules/manager/util.ts
```diff
@@ -20,7 +20,8 @@ export function applyGitSource(
         platform === 'github'
           ? GithubTagsDatasource.id
           : GitlabTagsDatasource.id;
-      const { protocol, source, full_name } = parseGitUrl(git);
+      const httpUrl = getHttpUrl(git);
```
RahulGautamSingh: Is this working fine on your test repo?

okkero: Yeah. I tested with git dependencies both using URLs of the form `https://github.com/...` and `ssh://git@github.com/...`, and they both work. Did you have any concerns about this? An edge case I may not be aware of?

lib/modules/manager/cargo/__snapshots__/extract.spec.ts.snap
```diff
@@ -282,23 +282,63 @@ exports[`modules/manager/cargo/extract > extractPackageFile() > extracts multipl
...
-    "currentValue": "0.1.0",
-    "datasource": "crate",
+    "currentValue": undefined,
+    "datasource": "git-refs",
     "depName": "git_dep_with_version",
     "depType": "dependencies",
     "managerData": {
       "nestedVersion": true,
     },
-    "skipReason": "git-dependency",
```
RahulGautamSingh, line 292: The version was detected before, but now it is unspecified? 🤔

rarkins: Good catch. Why is `currentValue` now undefined?

okkero: Right, hmm. So, for cargo git dependencies, the `version` field does not specify which version of the crate to fetch, but may exist alongside `branch`, `rev` or `tag` as a sort of sanity check: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#the-role-of-the-version-key

So, to me it seems correct that `currentValue` would be unspecified in this case, but I'm not exactly sure what to do with the version key for git dependencies, or whether we should do anything at all.

lib/modules/manager/util.ts
RahulGautamSingh: I was wondering why you case didn't occur earlier as we use the same function at other places for the same purpose.

lib/modules/manager/cargo/__snapshots__/extract.spec.ts.snap
RahulGautamSingh: Could you please point to me which change cause this? I am unable to find the change that triggered this

lib/modules/manager/cargo/artifacts.spec.ts
RahulGautamSingh, line 1: Why?

lib/modules/manager/cargo/__snapshots__/extract.spec.ts.snap
okkero: I think it's this: https://github.com/renovatebot/renovate/pull/36444/commits/8bbf20350551e2279a3390b26f080eabf224a458#diff-09657ce916b4a05cf4f9d8b64a7c5fc2eb4e4d6078cf7c45583656dc230899c1R65

@mkniewallner Can you confirm?

lib/modules/manager/util.ts
okkero: That I don't know. The dependency update breaks without this when using `ssh://git@github.com/...` style URLs with tags in the `Cargo.toml`. [Here is the Cargo.toml file used](https://github.com/okkero/renovate-test/blob/main/Cargo.toml), and this is the error message I get:

```
 INFO: Renovate is exiting with a non-zero code due to the following logged errors
       "loggerErrors": [
         {
           "name": "renovate",
           "level": 50,
           "logContext": "h4eoLGnOiFF5APflydRsn",
           "repository": "okkero/renovate-test",
           "url": "/graphql",
           "baseUrl": "ssh://github.com/api/",
           "resolvedUrl": "ssh://github.com/api/graphql",
           "msg": "Request Error: cannot parse url"
         }
       ]
```

I see that [massageUrl](https://github.com/renovatebot/renovate/blob/main/lib/modules/datasource/metadata.ts#L15) does something similar, although it seems more thorough and that it covers more cases.

Should we reuse that here instead? Or should this replacement be done at a different level entirely? I mean, it seems to me that this has to be done at some point or other, but I don't know what would be the best way to go about it...

lib/modules/manager/cargo/artifacts.spec.ts
okkero: I don't know. These were added before I took over the PR. Perhaps @mkniewallner can fill us in?

I tried removing these, and then two tests broke. It's not clear to me exactly what's happening here.

```
 FAIL  lib/modules/manager/cargo/artifacts.spec.ts > modules/manager/cargo/artifacts > returns updated Cargo.lock with precise version update
AssertionError: expected [ { …(2) } ] to match object [ …(2) ]

- Expected
+ Received

  [
    {
      "cmd": "cargo update --config net.git-fetch-with-cli=true --manifest-path Cargo.toml --workspace",
+     "options": {
+       "cwd": "/tmp/github/some/repo",
+       "encoding": "utf-8",
+       "env": {
+         "HOME": "/home/user",
+         "HTTPS_PROXY": "https://example.com",
+         "HTTP_PROXY": "http://example.com",
+         "LANG": "en_US.UTF-8",
+         "LC_ALL": "en_US",
+         "NO_PROXY": "localhost",
+         "PATH": "/tmp/path",
        },
-   {
-     "cmd": "cargo update --config net.git-fetch-with-cli=true --manifest-path Cargo.toml --package dep1@1.0.0 --precise 1.0.1",
+       "maxBuffer": 10485760,
+       "timeout": 900000,
+     },
    },
  ]

 ❯ lib/modules/manager/cargo/artifacts.spec.ts:142:27
    140|       { file: { contents: undefined, path: 'Cargo.lock', type: 'addition' } },
    141|     ]);
    142|     expect(execSnapshots).toMatchObject([
       |                           ^
    143|       {
    144|         cmd:
```

```
 FAIL  lib/modules/manager/cargo/artifacts.spec.ts > modules/manager/cargo/artifacts > returns updated Cargo.lock when a preceding dependency triggers an update in a later dependency
AssertionError: expected [ Array(1) ] to deeply equal [ { file: { …(3) } } ]

- Expected
+ Received

  [
    {
      "file": {
-       "contents": undefined,
+       "contents": "
+       [[package]]
+       name = \"dep1\"
+       version = \"1.0.1\"
+
+       [[package]]
+       name = \"dep2\"
+       version = \"1.0.2\"
+
+       [[package]]
+       name = \"dep3\"
+       version = \"1.0.0\"
+     ",
        "path": "Cargo.lock",
        "type": "addition",
      },
    },
  ]

 ❯ lib/modules/manager/cargo/artifacts.spec.ts:310:7
    308|         config,
    309|       }),
    310|     ).toEqual([
       |       ^
    311|       { file: { contents: undefined, path: 'Cargo.lock', type: 'addition' } },
    312|     ]);
```

lib/modules/manager/util.ts
okkero: Do you think that would work, @RahulGautamSingh , or do you recommend another approach? It just seems that the protocol for the registry URL can't be `ssh`, even though cargo supports it. So, the GitHub API must be called through `https`, which makes sense, so it has to be converted at some point. I am just unsure where exactly it would be best to do that.

lib/modules/manager/cargo/extract.ts
```diff
@@ -284,10 +284,15 @@ export async function extractPackageFile(
     for (const dep of deps) {
       const packageName = dep.packageName ?? dep.depName!;
       const versions = coerceArray(versionsByPackage.get(packageName));
-      const lockedVersion = versioning.getSatisfyingVersion(
-        versions,
-        dep.currentValue!,
-      );
+
+      let lockedVersion: string | null = null;
+      if (dep.currentValue) {
```
viceice, line 285: why this? can't it have a locked version without a pinned current value?

lib/modules/manager/cargo/index.ts
```diff
@@ -17,7 +20,12 @@ export const categories: Category[] = ['rust'];
 export const defaultConfig = {
   commitMessageTopic: 'Rust crate {{depName}}',
   managerFilePatterns: ['/(^|/)Cargo\\.toml$/'],
-  versioning: cargoVersioning.id,
```
viceice, line 20: ensure it the default on crate datasource

okkero: I'm pretty sure it is

lib/modules/manager/cargo/extract.ts
okkero: What do I pass as the second argument to `versioning.getSatisfyingVersion` if `dep.currentValue` is not set?

Would this make more sense?
```ts
      const lockedVersion = versioning.getSatisfyingVersion(
        versions,
        dep.currentValue ?? '*',
      );
```

lib/modules/manager/util.ts
okkero: Looks like this has been addressed on main now.

lib/modules/manager/cargo/extract.ts
RahulGautamSingh: No need I think if there is no currentVaule the getSatisfyingVersion will return `null` so this code can be kept as original.

Were there any errors with the original code that you encountered?

lib/modules/manager/cargo/__fixtures__/Cargo.1.toml
```diff
@@ -20,6 +20,9 @@ pcap-sys = { version = "=0.1", path = "pcap-sys" }
 pnet = { version = "0.21.0", optional = true, default-features = false}
 git_dep_with_version = { version = "0.1.0", git = "https://github.com/foo/bar" }
```
RahulGautamSingh, line 24: This dep was skipped before. And, I would assume that after your changes it wuld be updated, right? But the snaps say that it will be skipped with an unspecified version.

I don't mind approving the PR as is since the behaviour is same for such deps but we should look into this as why is is being skipped with an incorrect reason of `unspecified-version`

lib/modules/manager/cargo/__snapshots__/extract.spec.ts.snap
RahulGautamSingh: Yes, this seems to be the reason. 

But we might need to addresss this here as we are no introducing support for git deps, which were ignored when the function was written.

Maybe a new function for the crate manaer needs to be written, similar to applyGitSource which doesn't ignore git deps with specified version but no tag/rev

lib/modules/manager/cargo/extract.ts
okkero: I do remember a crash related to this, yes. I'll do some experimentation and report back.

lib/modules/manager/cargo/__fixtures__/Cargo.1.toml
okkero: Good catch. That behavior doesn't make much sense. I'll look into that.

lib/modules/manager/cargo/__snapshots__/extract.spec.ts.snap
okkero: Okay, I think I need an explanation of the relationship between `currentValue`, `currentDigest` and `currentVersion`. The latter two are obvious enough, but I'm not sure what `currentValue` is supposed to represent and how it differs from and/or relates to the other two.

`applyGitSource` seems to set `currentValue` to the specified `tag` or `branch`, whichever applies, but forgoes setting `currentValue` entirely when `rev` is selected, and I'm not sure why.

okkero: I think the crux of the issue here is that, if neither `tag` or `rev` were specified, `applyGitSource` will always fall back to assuming `branch` is set, overriding `currentValue` to `undefined` if `branch` isn't set after all.

I wonder if a change like this would be sufficient: https://github.com/renovatebot/renovate/commit/acd720082d93d6ef6e623b8a83421ff084d8a9b5
This fixes this particular issue in this snapshot, and doesn't seem to cause any other tests to break (unless the `currentValue: undefined` vs absent field in two of the other tests is important)

lib/modules/manager/cargo/extract.ts
okkero: > No need I think if there is no currentVaule the getSatisfyingVersion will return `null` so this code can be kept as original.

Hmm, I'm not sure this is true in general. The [interface for `getSatisfyingVersion`](https://github.com/renovatebot/renovate/blob/67935500d72c23d212b6d767adcb8b956b4bc6bf/lib/modules/versioning/types.ts#L94) does not specify the second argument as nullable or optional, so passing `dep.currentValue` without first checking would be disallowed by TypeScript. The function as declared requires the argument to be supplied and non-null.

If we force it anyway, which is what this code did previously, there is no guarantee that the actual versioning implementation won't just assume the argument will be an actual string (which, again, due to the types as declared, should be well within its rights), and it seems in the case of Cargo versioning, that's exactly what it does.

However, since the cargo manager seems to always default `currentValue` to empty string, I don't know how much of a problem this is in practice. I can reproduce the crash I was experiencing previously on this current branch by removing this check and adding a path dependency without version (very common in multi-crate projects). I'll add/modify a test to cover that case. With my suggested fix to `applyGitSource` (separate comment thread, see [my comment](https://github.com/renovatebot/renovate/pull/36444#discussion_r2307346503)), the crash goes away. Which makes sense, since `applyGitSource` no longer overrides `currentValue` to `undefined` then.

As a principle, I still wouldn't want to pass a potentially undefined value (by forcing it with `!`) to a function that explicitly declares that it requires a defined non-null string, so I think keeping this check is well justified.

Does that all make sense?

lib/modules/manager/cargo/__fixtures__/Cargo.1.toml
okkero: No, this should still be skipped I think, since a git dependency doesn't make sense without a rev, tag or branch. Specifying `version` for a git dependency is more of a supplementary sanity check.

Perhaps the skip reason should be something else. Would it be better to use 'git-dependency'?

RahulGautamSingh: Your reasoning sounds good. I think a comment/note somewhere will do the job if not then we can keep it as `git-dep-with-no-tag`.

RahulGautamSingh (commented):
Added comment

lib/modules/manager/cargo/__snapshots__/extract.spec.ts.snap
RahulGautamSingh: I don't think any change is needed now after [this comment ](https://github.com/renovatebot/renovate/pull/36444#discussion_r2309783005)

lib/modules/manager/cargo/__fixtures__/Cargo.1.toml
okkero: I'm not sure if this is what you meant, but I added a comment in the test fixture, explaining the behavior.

lib/modules/manager/cargo/__fixtures__/Cargo.1.toml
```diff
@@ -18,8 +18,12 @@ libc = "=0.2.43"
 bitflags = "1.0.4"
 pcap-sys = { version = "=0.1", path = "pcap-sys" }
 pnet = { version = "0.21.0", optional = true, default-features = false}
+
+# Git dependencies should always have one of tag, rev or branch defined.
+# If not, they should be skipped, even if version is defined.
 git_dep_with_version = { version = "0.1.0", git = "https://github.com/foo/bar" }
 git_dep = { git = "https://github.com/foo/bar" }
+
```
RahulGautamSingh, line 26:
```suggestion
```

RahulGautamSingh (approved):

lib/modules/manager/cargo/__fixtures__/Cargo.1.toml
okkero: Do you disagree with the grouping? It was to make it clear that the comment applies only to those two deps.

RahulGautamSingh: Got it. I thought it was by mistake did't see the grouping.

viceice (approved):

PR #36466, opened 2025-06-11
fix(gerrit): add codeowners support for gerrit

jutzig:
## Changes

Fixes CODEOWNERS support for the gerrit platform

## Context

I tried using CODEOWNERS in gerrit to automatically add reviewers, but found out it's not working. I came to the conclusion, that the problem is in code-owner#getBranchFiles, because it uses the sourceBranch of the Pr to diff.
With Gerrit, the value in sourceBranch isn't a real branch, so the diff fails. To fix that, I use the commit sha as a diff base for the gerrit platform.

### DIFF
--- a/lib/util/git/index.ts
+++ b/lib/util/git/index.ts
@@ -1030,35 +1030,27 @@ export async function getBranchLastCommitTime(
 export async function getBranchFiles(
   branchName: string,
 ): Promise<string[] | null> {
-  await syncGit();
-  try {
-    const diff = await gitRetry(() =>
-      git.diffSummary([`origin/${branchName}`, `origin/${branchName}^`]),
-    );
-    return diff.files.map((file) => file.file);
-    /* v8 ignore next 8 -- TODO: add test */
-  } catch (err) {
-    logger.warn({ err }, 'getBranchFiles error');
-    const errChecked = checkForPlatformFailure(err);
-    if (errChecked) {
-      throw errChecked;
-    }
-    return null;
-  }
+  return getBranchFilesFromRef(`origin/${branchName}`);
 }
 
 export async function getBranchFilesFromCommit(
   referenceCommit: LongCommitSha,
 ): Promise<string[] | null> {
+  return getBranchFilesFromRef(referenceCommit);
+}
+
+export async function getBranchFilesFromRef(
+  refName: string,
+): Promise<string[] | null> {
   await syncGit();
   try {
     const diff = await gitRetry(() =>
-      git.diffSummary([referenceCommit, referenceCommit + `^`]),
+      git.diffSummary([refName, `${refName}^`]),
     );
     return diff.files.map((file) => file.file);
     /* v8 ignore next 8 -- TODO: add test */
   } catch (err) {
-    logger.warn({ err }, 'getBranchFiles error');
+    logger.warn({ err }, 'getBranchFilesFromRef error');
     const errChecked = checkForPlatformFailure(err);
     if (errChecked) {
       throw errChecked;
--- a/lib/util/git/index.ts
+++ b/lib/util/git/index.ts
@@ -1027,13 +1027,11 @@ export async function getBranchLastCommitTime(
   }
 }
 
-export async function getBranchFiles(
-  branchName: string,
-): Promise<string[] | null> {
+export function getBranchFiles(branchName: string): Promise<string[] | null> {
   return getBranchFilesFromRef(`origin/${branchName}`);
 }
 
-export async function getBranchFilesFromCommit(
+export function getBranchFilesFromCommit(
   referenceCommit: LongCommitSha,
 ): Promise<string[] | null> {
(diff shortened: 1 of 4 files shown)


rarkins:
@felipecrs could you approve if ok?

felipecrs:
For sure. Allow me some time.

lib/workers/repository/update/pr/code-owners.ts
```diff
@@ -98,9 +99,14 @@ export async function codeOwnersForPr(pr: Pr): Promise<string[]> {
...
     logger.debug(`Found CODEOWNERS file: ${codeOwnersFile}`);
+    const pl = GlobalConfig.get('platform');
 
     // Get list of modified files in PR
-    const prFiles = await getBranchFiles(pr.sourceBranch);
+    //on the gerrit platform, the source branch does not exists. Instead of the branch, we can take the commit sha and diff against the parent
+    const prFiles =
+      'gerrit' === pl && pr.sha
+        ? await getBranchFilesFromCommit(pr.sha)
+        : await getBranchFiles(pr.sourceBranch);
```
felipecrs: I don't think this needs to be platform specific.

Why not use `getBranchFilesFromCommit()` whenever `pr.sha` is set? There are only benefits, it makes the function sharper.

And I wonder why not use it always: can't we make `sha` a mandatory property in the `Pr` type?

@rarkins what do you think?

jutzig: @felipecrs thank you very much for taking the time to review!

> I don't think this needs to be platform specific.

I wanted to change as little as possible, but if it makes sense in general... sure

felipecrs: You did good, let's just wait for @rarkins to confirm the way forward.

rarkins: Yes, we prefer to avoid platform-specific code in workers

felipecrs: @rarkins sorry, your answer isn't clear to me. Which approach should we take? The one I suggested [here](https://github.com/renovatebot/renovate/pull/36466#discussion_r2142722942)?

jutzig: @felipecrs  please see latest commit, is this what you had in mind?

zharinov (approved):

lib/workers/repository/update/pr/code-owners.ts
felipecrs: Yes, that's exactly what I meant. :)

lib/workers/repository/update/pr/code-owners.spec.ts
```diff
@@ -20,6 +21,7 @@ describe('workers/repository/update/pr/code-owners', () => {
 
     beforeEach(() => {
       pr = mock<Pr>();
+      pr.sha = undefined;
```
viceice, line 24:
```suggestion
       pr = mock<Pr>({sha: undefined});
```
should work

lib/workers/repository/update/pr/code-owners.spec.ts
```diff
@@ -29,6 +31,14 @@ describe('workers/repository/update/pr/code-owners', () => {
       expect(codeOwners).toEqual(['@jimmy']);
     });
 
+    it('returns global code owner for commit with sha set', async () => {
+      pr.sha = 'f7374c2de8a4c95a7fd7182ab24044e3896aac02' as LongCommitSha;
+      fs.readLocalFile.mockResolvedValueOnce(['* @jimmy'].join('\n'));
```
viceice:
```suggestion
      fs.readLocalFile.mockResolvedValueOnce('* @jimmy');
```
no join for a single value

lib/util/git/index.ts
```diff
@@ -1027,18 +1027,28 @@ export async function getBranchLastCommitTime(
...
+  return getBranchFilesFromRef(`origin/${branchName}`);
+}
+
+export function getBranchFilesFromCommit(
+  referenceCommit: LongCommitSha,
+): Promise<string[] | null> {
+  return getBranchFilesFromRef(referenceCommit);
+}
+
+export async function getBranchFilesFromRef(
```
viceice:
```suggestion
async function getBranchFilesFromRef(
```
not externally used

lib/workers/repository/update/pr/code-owners.spec.ts
jutzig: Thanks for the review!
I tried that, but didn't work for me. Commit that addresses the other issues is incoming

jutzig:
anything left open for me to change? Thanks!

RahulGautamSingh (approved):

felipecrs (approved):

rarkins (approved):

jutzig:
@viceice I think your review is the last thing missing. Is there anything you want changed?

viceice (approved):

PR #36468, opened 2025-06-11
fix(codecommit): reset cached pull request list when initializing repository

dhutchison:
## Changes

Updates the CodeCommit platform to re-initialize its configuration fully when calling `initRepo`. This fixes an issue where duplicate PRs would be created for any repository after the first being processed.

## Context

The background information, including supporting logs, is covered in discussion  #26357 .

I have been running a monkey patched version of the node module for a few months now with this change and have not seen the duplicate issue re-occur since. 

The existing unit tests continue to pass with some modifications were required to avoid direct access to a field which is no longer defined as a constant, but should have the same intended effect (although I'm a little sceptical that the tests I modified test what they say they test - could be reading it wrong though, my TS/JS is a little rusty).

### DIFF
--- a/lib/modules/platform/codecommit/index.spec.ts
+++ b/lib/modules/platform/codecommit/index.spec.ts
@@ -30,6 +30,19 @@ import { git, logger } from '~test/util';
 
 const codeCommitClient = mockClient(CodeCommitClient);
 
+async function reInitRepo(codeCommit: Platform): Promise<void> {
+  codeCommitClient.on(GetRepositoryCommand).resolvesOnce({
+    repositoryMetadata: {
+      defaultBranch: 'main',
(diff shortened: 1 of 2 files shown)


rarkins (dismissed):

viceice (commented):
there are some more tests to fix

dhutchison:
Ok, I see I need to run at least `pnpm vitest lib/modules/platform/codecommit` also. 

I'll take a look in the next few days see if I can work out what is going on with those tests.

dhutchison:
I've pushed a new commit that should clear the PR cache state between the test runs as a few tests had been directly modifying the cache (that seemed to be the main cause of the failures, as far as I can tell).

rarkins (approved):

Issue #26571, opened 2024-01-10
Improve config option metadata to define where options can be used and improve JSON schema


rarkins:
### Describe the proposed change(s).

Enhance this source file: https://github.com/renovatebot/renovate/blob/main/lib/config/options/index.ts

Right now we have some options with "parent" defined (such as `packageRules>matchPackageNames`) but it's not enough info to generate the JSON schema accurately when options such as `enabled` can be at root or within `packageRules`, or options such as `configMigration` must be at the root of config.

It's even more complicated when you consider that we support some dynamic config options such as `minor`, or manager names such as `gomod`.

We should brainstorm about all the possible objects which config can be in (root, hostRules, packageRules, managers, etc) and then define the syntax. Perhaps `parent` should just be an array instead of single string.

rarkins:
Ref: #26484

rarkins:
I split off https://github.com/renovatebot/renovate/issues/26599 @RahulGautamSingh

RahulGautamSingh:
Adding `root` as a valid parent will allow us to inject `enabled` in both the `packageRules` and in the root object in the schema. It will also be useful for `managerFilePatterns`.

viceice:
maybe use `$` instead of `root` ?

RahulGautamSingh:
> maybe use $ instead of root ?

Would `.` work? Using period simplifies validation logic a lot

PR #36471, opened 2025-06-12
feat(config/options): allow all managers as parents of `managerFilePatterns`

### LINKED ISSUE #7267: Schema.json doesn't have field .regexManagers[].fileMatch
There is no completion and documentation for the fileMatch field in a regexManager object.

```json
{
  "regexManagers": [{
    "fileMatch": []
  }]
}
```

RahulGautamSingh:
## Changes
This is an attempt to include all the managers as allowed parents, so that we can generate accurate schema for config options which can be present within manager names like `npm` and also within other config options.

One example would be `managerFilePatterns` which can be present within `npm` and `customManagers`

## Context
This PR is an attempt to move the discussion further on below issue
- Ref: https://github.com/renovatebot/renovate/issues/26571
- Closes: https://github.com/renovatebot/renovate/issues/7267

### DIFF
--- a/tools/generate-imports.mjs
+++ b/tools/generate-imports.mjs
@@ -141,6 +141,25 @@ async function generateData() {
   );
 }
 
+async function generateManagerList() {
+  // get managers list
+  const managers = (
+    await fs.readdir('lib/modules/manager', { withFileTypes: true })
+  )
+    .filter((file) => file.isDirectory())
+    .map((file) => file.name)
+    .filter((mgr) => mgr !== 'custom')
+    .sort()
+    .map((fname) => `"${fname}"`);
+
+  const content = `
+export const AllManagersListLiteral = [${managers}] as const;
+export type ManagerName = typeof AllManagersListLiteral[number];
+`;
+
+  await updateFile(`lib/manager-list.generated.ts`, content);
+}
+
 async function generateHash() {
   console.log('generating hashes');
   try {
@@ -191,6 +210,7 @@ await (async () => {
   try {
     // data-files
     await generateData();
+    await generateManagerList();
     await generateHash();
     await Promise.all(
       (await glob('lib/**/*.generated.ts'))
--- a/tools/generate-imports.mjs
+++ b/tools/generate-imports.mjs
@@ -153,7 +153,7 @@ async function generateManagerList() {
     .map((fname) => `"${fname}"`);
 
   const content = `
-export const AllManagersListLiteral = [${managers}] as const;
+export const AllManagersListLiteral = [${managers.join(',')}] as const;
 export type ManagerName = typeof AllManagersListLiteral[number];
 `;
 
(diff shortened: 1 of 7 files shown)


rarkins:
Here's the schemas I get in main vs this PR:

[renovate-schema-main.json](https://github.com/user-attachments/files/20702180/renovate-schema-main.json)
[renovate-schema.json](https://github.com/user-attachments/files/20702181/renovate-schema.json)

I'm not sure that it's a problem, but the new one is almost twice as large. Diffs to follow

rarkins:
Here's an example of how a manager schema looks now:

```json
"ansible": {
      "description": "Configuration object for the ansible manager",
      "type": "object",
      "default": {
        "managerFilePatterns": [
          "/(^|/)tasks/[^/]+\\.ya?ml$/"
        ]
      },
      "$ref": "#",
      "items": {
        "allOf": [
          {
            "type": "object",
            "properties": {
              "description": {
                "oneOf": [
                  {
                    "type": "array",
                    "items": {
                      "type": "string",
                      "description": "A custom description for this configuration object"
                    }
                  },
                  {
                    "type": "string",
                    "description": "A custom description for this configuration object"
                  }
                ]
              },
              "managerFilePatterns": {
                "description": "RegEx (`re2`) and glob patterns for matching manager files.",
                "oneOf": [
                  {
                    "type": "array",
                    "items": {
                      "type": "string"
                    }
                  },
                  {
                    "type": "string"
                  }
                ]
              }
            }
          }
        ]
      }
    },
```

It looks like the "description" field is also added as part of this PR, but not sure why.

tools/generate-imports.mjs
```diff
@@ -141,6 +141,25 @@ async function generateData() {
...
+    await fs.readdir('lib/modules/manager', { withFileTypes: true })
+  )
+    .filter((file) => file.isDirectory())
+    .map((file) => file.name)
+    .filter((mgr) => mgr !== 'custom')
+    .sort()
+    .map((fname) => `"${fname}"`);
+
+  const content = `
+export const AllManagersListLiteral = [${managers}] as const;
```
viceice: shouldn't we use an explicit join? just to be safe?

viceice (commented):
otherwise LGTM

RahulGautamSingh:
> It looks like the "description" field is also added as part of this PR

It is how we have designed the schema of any config option with children to be. 
https://github.com/renovatebot/renovate/blob/568898848d30ef968742dccb9f2b90b434298d79/tools/docs/schema.ts#L102
Before manager names couldn't be parents so the description wasn't added. But this PR changed that.

Btw since you are testing this locally, can you confirm whether we get suggestions and warnings when writing `managerFilePatterns` in a `customManagers` object. It's not working locally for me, but it should as per the generated schema

rarkins:
Which way are you testing it? e.g. within IDE? CLI validation?

RahulGautamSingh:
> Which way are you testing it? e.g. within IDE? CLI validation?

within IDE

rarkins:
Which IDE, and how do you instruct it to use your locally-generated schema instead of the hosted one?

RahulGautamSingh:
> Which IDE, and how do you instruct it to use your locally-generated schema instead of the hosted one?

Found the issue, I was still using the hosted schema.
It works now, after referencing the local schema

viceice (approved):

PR #36474, opened 2025-06-12
feat(presets/monorepo): group oxlint and eslint-plugin-oxlint

gwennlbh:
seems like both packages are always released together (which makes sense)


lib/data/monorepo.json
```diff
@@ -642,6 +642,7 @@
       "/^com.oracle.database.jdbc:/",
       "/^com.oracle.database.nls:/"
     ],
+    "oxlint-eslint": ["oxlint", "eslint-plugin-oxlint"],
```
rarkins: Maybe "oxlint" is better for this? If someone uses _only_ oxlint, they will get a strange "Update exlint-eslint to v1.1.0" PR due to this grouping rule.

RahulGautamSingh (approved):

rarkins (approved):

rarkins (approved):

gwennlbh:
thx! do you know when this will be available on mend.io renovate  (my repo is on github.com) ? so i can remove the manual group from my config

lib/data/monorepo.json
```diff
@@ -669,6 +669,7 @@
       "/^com.oracle.database.jdbc:/",
       "/^com.oracle.database.nls:/"
     ],
+    "oxlint": ["oxlint", "eslint-plugin-oxlint"],
```
viceice, line 672: it should have been grouped via source url instead

PR #36481, opened 2025-06-12
feat(datasource/docker): quay v2 tag list fallback

mjnagel:
## Changes

This PR adds a fallback to use the docker v2 api with Quay if 401s are hit.

## Context

Additional context is provided in this discussion: https://github.com/renovatebot/renovate/discussions/36292

In short, the current method of querying the Quay API for tags is great for free/public images. For private images however there are limitations and it requires more auth/different creds than just pulling the images. This would introduce a fallback approach to go through the v2 api if requests are unauthorized.

Test repository (see issues/PRs): https://github.com/BagelLab/renovate-quay-test

Debug log file: [renovate.log](https://github.com/user-attachments/files/20794717/renovate.log)


lib/modules/datasource/docker/index.ts
```diff
@@ -656,8 +657,14 @@ export class DockerDatasource extends Datasource {
...
         ? 1000
         : 10000;
+    // Add the last parameter if currentVersion is provided to optimize tag fetching for Quay
+    const isQuay = regEx(/^https:\/\/quay\.io(?::[1-9][0-9]{0,4})?$/i).test(
+      registryHost,
+    );
     let url: string | null =
-      `${registryHost}/${dockerRepository}/tags/list?n=${limit}`;
+      isQuay && currentVersion
+        ? `${registryHost}/${dockerRepository}/tags/list?n=${limit}&last=${currentVersion}`
```
mjnagel: This `last` behavior is maybe the most questionable change - it would rely on a specific query parameter to attempt to shorten the amount of queries needed to Quay. In essence it gets any tag since the current version. Questions here:
1. Is this desirable, or is it too finicky/clunky?
2. If we want to do this - are there nuances with `extractVersion` and `currentVersion` here? I'm not as familiar with this portion, so not sure if that would need to be accounted for here.

RahulGautamSingh: 2. Could you add the reference to to the api docs so we can see what exactly is needed for the `last` param?
 I just want to be sure whether we need to pass the `currentValue` as is or we need to extract the verison and use that.
 
 For eg. `currentValue` = `1.0.0-apline` here `currentVersion` = `1.0.0`, which do you think should be used as the `last` param
 
 If it's the `currentVersion` then we could use `extractVersion` if present or rely on [valueToVersion](https://github.com/renovatebot/renovate/blob/788e408cbe351020b52e20df292fb4f11f8b67d6/lib/modules/versioning/docker/index.ts#L91) from the docker versioning.

mjnagel: Sure yeah, so this comes from the distribution spec: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-tags
> The last query parameter provides further means for limiting the number of tags. It is usually used in combination with the n parameter: /v2/<name>/tags/list?n=<int>&last=<tagname> [end-8b](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints)

> <name> is the namespace of the repository, <int> is the number of tags requested, and <tagname> is the value of the last tag. <tagname> MUST NOT be a numerical index, but rather it MUST be a proper tag. A request of this sort will return up to <int> tags, beginning non-inclusively with <tagname>. That is to say, <tagname> will not be included in the results, but up to <int> tags after <tagname> will be returned. The tags MUST be in lexical order.

In the scenario you provided I think we would want `currentValue` assuming that is the actual tag being used for the image? By way of example with a public image:
```
curl "https://quay.io/v2/metallb/speaker/tags/list?n=100&last=v0.13.2"
```

This is the query we would want to do here, with the actual tag having a `v` in front of the version. By contrast, doing this same query with just `0.13.2` (no `v`) would return all tags from the start - losing any benefit of this param. So the param has to be the actual image tag. I would definitely benefit from any pointers on what this might mean with respect to my current change and whether we'd want to adjust.

RahulGautamSingh: We would want to keep the name of the param `currentValue` throughout then, instead of `currentVersion` to prevent confusion. Btw, from the docs, it seems the `last` tag will not be included in the fetched tags? Is that correct?

mjnagel: Just pushed up a change to switch to `currentValue` everywhere - my mistake on that. For your question, that is correct - the `last` value will not be included. For example:
```console
❯ curl "https://quay.io/v2/metallb/speaker/tags/list?n=100&last=v0.13.2"
{"name":"metallb/speaker","tags":["v0.13.3","v0.13.4","v0.13.5","v0.13.6","v0.13.7","v0.13.9","v0.14","v0.14.0","v0.14.1","v0.14.2","v0.14.3","v0.14.4","v0.14.5","v0.14.6","v0.14.7","v0.14.8","v0.14.9","v0.15", ...
```

Note that v0.13.2 is not included in the results. Will that cause any issues with this?

RahulGautamSingh: That might caught issues with our `getCurrentVersion` logic in the lookup stage. Could you please test this on a repo so we can be sure?

RahulGautamSingh (commented):
Added comment

lib/modules/datasource/docker/index.ts
```diff
@@ -656,8 +657,14 @@ export class DockerDatasource extends Datasource {
       ecrRegex.test(registryHost) || ecrPublicRegex.test(registryHost)
         ? 1000
         : 10000;
+    // Add the last parameter if currentValue is provided to optimize tag fetching for Quay
+    const isQuay = regEx(/^https:\/\/quay\.io(?::[1-9][0-9]{0,4})?$/i).test(
+      registryHost,
+    );
```
viceice: use startsWith which is much faster than regex (perf)

lib/modules/datasource/docker/index.ts
```diff
@@ -725,22 +733,49 @@ export class DockerDatasource extends Datasource {
...
   ): Promise<string[] | null> {
     try {
       const isQuay = regEx(/^https:\/\/quay\.io(?::[1-9][0-9]{0,4})?$/i).test(
         registryHost,
       );
       let tags: string[] | null;
       if (isQuay) {
-        tags = await this.getTagsQuayRegistry(registryHost, dockerRepository);
+        try {
+          // Due to pagination and sorting limits on Quay standard v2 API we try the v1 API first
```
viceice:
```suggestion
          // Due to pagination and sorting limits on Quay Docker v2 API implementation we try the Quay v1 API first
```

lib/modules/datasource/docker/index.ts
```diff
@@ -1066,6 +1101,7 @@ export class DockerDatasource extends Datasource {
   async getReleases({
     packageName,
     registryUrl,
+    currentValue,
```
viceice: add to cache key above

lib/modules/datasource/docker/index.ts
mjnagel: I can give this a shot - it may take me a bit to get a test repo up and running my branch but I'll see if I can do that. I may also try to get a test added for `getCurrentVersion` - do you think that would also help build confidence in this?

lib/modules/datasource/docker/index.ts
mjnagel: Batched this with the other changes - forgot to apply directly sorry!

lib/modules/datasource/docker/index.ts
RahulGautamSingh: Yes

mjnagel: I've got a test repo pushed up here that I ran my renovate branch against: https://github.com/BagelLab/renovate-quay-test/pulls

For the quay images, that repository is the only one I have access to test against that has private images, but it is working as expected from what I can tell. There's limited variety in tag names (basically all of them are `<semver>-<suffixes>`) so I can't try too many other things here but it looks good so far. Is there anything you'd like to see from the logs or otherwise?

I'll also push up a short test file for `getCurrentVersion` and re-request review when ready.

RahulGautamSingh: Could you please only keep the necessary private image in the test repo (there are 5 currently, I think only 1 is needed to test) and also share the debug logs, so we can be sure the `v1` api req fails and we fallback to `v2`.

You can drop the `getCurrentVersion` tests for this PR they should be added in a separate PR. I thought we already had a test file and you will only need to add a test case but I was mistaken, Sorry!

mjnagel: Updated to remove that test file, also updated the description of the PR to point to my fork + include the debug logs - [renovate.log](https://github.com/user-attachments/files/20794717/renovate.log). From that log I see:
```console
DEBUG: GET https://quay.io/api/v1/repository/rfcurated/istio/proxyv2/tag/?limit=100&page=1&onlyActiveTags=true = (code=ERR_NON_2XX_3XX_RESPONSE, statusCode=401 retryCount=0, duration=196) (repository=BagelLab/renovate-quay-test)
DEBUG: Quay v1 API unauthorized, falling back to Docker v2 API (repository=BagelLab/renovate-quay-test)
       "registryHost": "https://quay.io",
       "dockerRepository": "rfcurated/istio/proxyv2"
```

and then it proceeds to pick up the update to 1.26.0 via the `v2` api.

lib/modules/datasource/docker/index.ts
```diff
@@ -656,8 +656,12 @@
       ecrRegex.test(registryHost) || ecrPublicRegex.test(registryHost)
         ? 1000
         : 10000;
+    // Add the last parameter if currentValue is provided to optimize tag fetching for Quay
+    const isQuay = registryHost.startsWith('https://quay.io');
```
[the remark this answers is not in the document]
rarkins: can this be an equals instead of startsWith? I'm not sure that quay.io has any extra path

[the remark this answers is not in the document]
mjnagel: Yep, no issue with that. Let me make the changes.

mjnagel:
I'll look at the test coverage diff later today - I assume fixing that is necessary to get this across the line?

Edit: In reviewing the [codecov report](https://app.codecov.io/gh/renovatebot/renovate/pull/36481) it seems like I increased coverage on the file changed so I'm unsure what to do here?

mjnagel:
Merging main to see if this is the cause of the codecov failures.

lib/modules/datasource/docker/index.ts
mjnagel: After thinking about this more I'm wondering if this should be in the cache key? The only cacheable calls here are ones for `docker.io` images, and this `currentValue` is only relevant for `quay.io` images. I'm leaning towards removing from the cache key based on this, what do you think?

mjnagel: Discussed on Slack - cache decorator applies to all registries so still relevant even if `cacheable` "filters out" non-dockerhub.

RahulGautamSingh:
In the test repo's PR: https://github.com/BagelLab/renovate-quay-test/pull/8/files

The bot updates the version correctly and also attaches a digest which was not there previously. Is this expected behavior for you? @mjnagel

mjnagel:
@RahulGautamSingh yeah I believe that is expected since I was extending `config:best-practices` which includes `docker:pinDigests`. I was just trying to get a quick test up with a simple config.

RahulGautamSingh:
Could you do one without that preset, please? 😅

mjnagel:
@RahulGautamSingh just ran, PRs updated: https://github.com/BagelLab/renovate-quay-test/pull/11/files

RahulGautamSingh:
Great, thanks!

RahulGautamSingh (approved):

lib/modules/datasource/docker/index.ts
```diff
@@ -647,6 +646,7 @@ export class DockerDatasource extends Datasource {
   private async getDockerApiTags(
     registryHost: string,
     dockerRepository: string,
+    currentValue?: string,
```
rarkins: I don't like how a generic currentValue param has been added here but the only way it's used is if the host is quay.io? If you imagine a future developer looking at this code and wondering if they need to provide currentValue then they'd need to trace through the code, do a git blame, find this PR, etc. I'd prefer this is done more clearly

lib/modules/datasource/docker/index.ts
```diff
@@ -656,8 +656,12 @@ export class DockerDatasource extends Datasource {
       ecrRegex.test(registryHost) || ecrPublicRegex.test(registryHost)
         ? 1000
         : 10000;
+    // Add the last parameter if currentValue is provided to optimize tag fetching for Quay
+    const isQuay = registryHost === 'https://quay.io';
     let url: string | null =
-      `${registryHost}/${dockerRepository}/tags/list?n=${limit}`;
+      isQuay && currentValue
+        ? `${registryHost}/${dockerRepository}/tags/list?n=${limit}&last=${currentValue}`
```
rarkins: It would be simpler to have a if block to append this later, including the comments describing why

lib/modules/datasource/docker/index.ts
```diff
@@ -725,20 +730,41 @@ export class DockerDatasource extends Datasource {
 
   @cache({
     namespace: 'datasource-docker-tags',
-    key: (registryHost: string, dockerRepository: string) =>
-      `${registryHost}:${dockerRepository}`,
+    key: (
+      registryHost: string,
+      dockerRepository: string,
+      currentValue?: string,
```
rarkins: This change of caching could explode the number of cache entries

rarkins (changes requested):

lib/modules/datasource/docker/index.ts
mjnagel: I can go ahead and rename the param for more clarity here - or were you looking for more of a change than that?

lib/modules/datasource/docker/index.ts
mjnagel: This makes sense, I'll make that update.

lib/modules/datasource/docker/index.ts
mjnagel: I can see that - I think I lean towards just removing currentValue from the cache entirely since it will only be used for private quay images? But I'm not sure if that would cause other issues here...

cc @viceice also since you had requested adding to the cache key for `getReleases` in https://github.com/renovatebot/renovate/pull/36481#discussion_r2150577531.

lib/modules/datasource/docker/index.ts
RahulGautamSingh: Yes, and add comments explaining the reasoning behind it's use.

RahulGautamSingh (commented):
Added comment

lib/modules/datasource/docker/index.ts
mjnagel: @RahulGautamSingh were you looking for something different than what is in the latest commit?
```ts
  /**
   * Get Docker API tags from a registry
   * @param registryHost - The registry host URL
   * @param dockerRepository - The Docker repository name
   * @param quayLastTagValue - Only used for Quay.io to optimize tag fetching by providing the last tag value seen
   * @returns Array of tags or null if not found
   */
  private async getDockerApiTags(
    registryHost: string,
    dockerRepository: string,
    quayLastTagValue?: string,
  ): Promise<string[] | null> {
```

RahulGautamSingh: Yes, and also this maybe:

```ts
async getReleases(
    releaseConfig: GetReleasesConfig,
    const { packageName, registryUrl, currentValue } = releaseConfig;
```

This way, `currentValue` is not present in the main function definition explicitly, which should prevent confusion for future contributors, like Rhys mentioned

lib/modules/datasource/docker/index.ts
```diff
@@ -1104,11 +1104,11 @@ export class DockerDatasource extends Datasource {
...
-  async getReleases({
-    packageName,
-    registryUrl,
-    currentValue,
-  }: GetReleasesConfig): Promise<ReleaseResult | null> {
+  async getReleases(
+    releaseConfig: GetReleasesConfig,
+  ): Promise<ReleaseResult | null> {
+    const { packageName, registryUrl, currentValue } = releaseConfig;
+
```
RahulGautamSingh:
```suggestion
```

RahulGautamSingh (approved):

mjnagel:
PR has been updated to only add the fallback to v2 api on 401s - removing the `last` parameter approach.

RahulGautamSingh:
> only add the fallback to v2 api on 401s - removing the last parameter approach.

Why?

rarkins (approved):

mjnagel:
> > only add the fallback to v2 api on 401s - removing the last parameter approach.
> 
> Why?

@RahulGautamSingh I guess it just came down to a discussion in Slack with @rarkins. While it does work, my preference is to just get this smaller change across the line so that private images in Quay can be tracked/updated with Renovate. The `last` parameter was an optimization around that but I think because using `currentValue` is unique to the docker datasource, and even further unique to Quay private images - it just isn't desired.

RahulGautamSingh (approved):

PR #36506, opened 2025-06-14
fix(terragrunt): handle local includes and return empty deps array

arnaud-dezandee:
## Changes

  This fix enables lock file maintenance for Terragrunt modules that don't directly contain a `terraform {}` block but instead include one from another location on the local filesystem using `include` blocks.

  Previously, Terragrunt files containing only local includes (like `include "root" { path = find_in_parent_folders("root.hcl") }`) would return `null` from the package file extraction, preventing lock file maintenance. Now these
  files return an empty dependencies array (`{ deps: [] }`), allowing Renovate to properly track and maintain lock files for such configurations.

  ## Context

  In Terragrunt configurations, it's common to have modules that don't directly define a `terraform {}` block but inherit it from parent configurations using `include` blocks. These configurations should still be eligible for lock
  file maintenance even though they don't contain direct Terraform dependencies.

  The fix specifically handles the case where:
  - A Terragrunt file contains `include "` blocks (indicating local includes)
  - But does not contain `terraform {` blocks (the current dependency detection criteria)

  The fix includes a new unit test that verifies files with only local Terragrunt includes return `{ deps: [] }` instead of `null`, ensuring proper lock file maintenance support.

### DIFF
--- a/lib/modules/manager/terragrunt/extract.ts
+++ b/lib/modules/manager/terragrunt/extract.ts
@@ -10,6 +10,7 @@ import {
 
 const dependencyBlockExtractionRegex = regEx(/^\s*(?<type>[a-z_]+)\s+{\s*$/);
 const contentCheckList = ['terraform {'];
+const includeBlockCheck = 'include "';
 
 export function extractPackageFile(
   content: string,
@@ -17,6 +18,9 @@ export function extractPackageFile(
 ): PackageFileContent | null {
   logger.trace({ content }, `terragrunt.extractPackageFile(${packageFile!})`);
   if (!checkFileContainsDependency(content, contentCheckList)) {
+    if (content.includes(includeBlockCheck)) {
+      return { deps: [] };
+    }
     return null;
   }
   let deps: PackageDependency<TerraformManagerData>[] = [];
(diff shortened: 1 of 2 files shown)


lib/modules/manager/terragrunt/extract.ts
```diff
@@ -10,13 +10,17 @@ import {
...
 
 export function extractPackageFile(
   content: string,
   packageFile?: string,
 ): PackageFileContent | null {
   logger.trace({ content }, `terragrunt.extractPackageFile(${packageFile!})`);
   if (!checkFileContainsDependency(content, contentCheckList)) {
+    if (content.includes(includeBlockCheck)) {
+      return { deps: [] };
+    }
```
RahulGautamSingh: If the `include "` string will always be present at the beginning of the file. A regex check like `/^include\s"/` or `content.startsWith('include "')` is more suitable here.

RahulGautamSingh:
Thank you for you PR! 

Could you please create a two reproductions as well so we can check the current behavior and the new behavior this PR introduces?

arnaud-dezandee:
> Thank you for you PR!
> 
> Could you please create a two reproductions as well so we can check the current behavior and the new behavior this PR introduces?

Hi @RahulGautamSingh, here is the reproduction repository https://github.com/arnaud-dezandee/renovate-36506

lib/modules/manager/terragrunt/extract.ts
arnaud-dezandee: This is not always true, in the reproduction repository the terragrunt file starts with a `locals` block: https://github.com/arnaud-dezandee/renovate-36506/blob/main/include-module/terragrunt.hcl

arnaud-dezandee:
@RahulGautamSingh I've updated the include check with a Regex which is more robust to match all possible syntax

rarkins (approved):

PR #36511, opened 2025-06-14
fix(gitlab): approve PR after automerge is added

Nyholm:
## Changes

I would like to send "approve" for gitlab MRs to happen after we add "auto merge".

## Context

More often then not, I get into a scenario like this: 
Screenshot 2025-06-14 at 17 32 52

17:16:45 We approve the PR
17.16:53 The first commit is pushed (which according to my gitlab settings, will reset approvals)
17.16.54 We enable auto merge

This leaves my PRs open because they are never approved. 

Looking at the `tryPrAutomerge()`, I am fairly sure (please correct if I am wrong), that we will still call the API to merge because the PR will be in state `ci_still_running` or `not_approved`. 

### DIFF
--- a/lib/modules/platform/gitlab/index.ts
+++ b/lib/modules/platform/gitlab/index.ts
@@ -766,11 +766,11 @@ export async function createPr({
     !!config.ignorePrAuthor,
   );
 
+  await tryPrAutomerge(pr.number, platformPrOptions);
+
   if (platformPrOptions?.autoApprove) {
     await approveMr(pr.number);
   }
-
-  await tryPrAutomerge(pr.number, platformPrOptions);
 
   return pr;
 }


rarkins:
Renovate always does its commit before automerging, so I'm not sure that this is really the problem. 

Commit is done, branch pushed, MR created, MR approved, MR automerge.

Nyholm:
You are correct. But the actions in Gitlab is not registered in that order. That is why there is a problem. 

The logic of the auto-merge is waiting for Gitlab to get ready. That is why approving after auto-merging will work.

rarkins:
I'm telling you the order that Renovate does things. I'd like to understand why/how things are in a different order.

For example, is this a scenario where MRs are rebased? Does the GitLab system have a race condition where the branch push is not detected until _after_ the MR approval? Maybe the race condition would still exist after this change, just in a different order.

As you are the only one who's reported a problem with this that I'm aware of, I don't want to rush this before being sure about it.

PR #36522, opened 2025-06-15
feat(config/options): allow root(`.`) and `UpdateType` as parents

### LINKED ISSUE #26571: Improve config option metadata to define where options can be used and improve JSON schema

RahulGautamSingh:
## Changes
- Allow UpdateTypes (major minor, ...) and root ( `.` ) as parent of config option definition

## Context
- Closes: https://github.com/renovatebot/renovate/issues/26571

### DIFF
--- a/lib/config/defaults.ts
+++ b/lib/config/defaults.ts
@@ -23,7 +23,7 @@ export function getConfig(): AllConfig {
   const options = getOptions();
   const config: AllConfig = {};
   options.forEach((option) => {
-    if (!option.parents) {
+    if (!option.parents || option.parents.includes('.')) {
       config[option.name] = getDefault(option);
     }
   });
--- a/lib/config/types.ts
+++ b/lib/config/types.ts
@@ -359,7 +359,7 @@ export type UpdateType =
   | 'replacement';
 
 // These are the update types which can have configuration
-export const updateTypesOptions = [
+export const UpdateTypesOptions = [
   'major',
   'minor',
   'patch',
@@ -371,7 +371,7 @@ export const updateTypesOptions = [
   'replacement',
 ] as const;
 
-export type UpdateTypeOptions = (typeof updateTypesOptions)[number];
+export type UpdateTypeOptions = (typeof UpdateTypesOptions)[number];
 
 export type FetchChangeLogsOptions = 'off' | 'branch' | 'pr';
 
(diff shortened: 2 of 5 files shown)


lib/config/defaults.ts
```diff
@@ -23,7 +23,7 @@ export function getConfig(): AllConfig {
   const options = getOptions();
   const config: AllConfig = {};
   options.forEach((option) => {
-    if (!option.parents) {
+    if (!option.parents || option.parents.includes('.')) {
```
secustor, line 26: This would make `default` noop for all root options. That is if we add the parents as expected.

lib/config/types.ts
```diff
@@ -358,6 +358,21 @@ export type UpdateType =
   | 'bump'
   | 'replacement';
 
+// These are the update types which can have configuration
+export const updateTypesOptions = [
```
viceice: why not using same case for both? typescript should be smart enough to detect right reference

lib/config/defaults.ts
RahulGautamSingh: Sorry, I don't understand what you mean. 

Do you mean that `enabled` which will have `parents:['.', 'hostRules', ...]` will have no default value after this change OR that options like `binarySource` which can only in root config and have no `parents` array will not get a default value?

secustor: Sry misread here...

secustor (approved):

rarkins:
This produces a ~1000 line diff between the main branch JSON schema and this PR. Has it been reviewed for correctness?

RahulGautamSingh:
> This produces a ~1000 line diff between the main branch JSON schema and this PR. Has it been reviewed for correctness?

Yes. `enabled` will now be added as a children to all `npm` like manager objects, `minor` like update type objects so tis is expected.

PR #36532, opened 2025-06-16
feat(presets): Add officeHours

martincostello:
## Changes

Add an `officeHours` schedule preset that runs whenever `nonOfficeHours` doesn't.

## Context

I wanted to add a schedule that was more-or-less _"while I'm awake during the week"_, and found there wasn't an opposite schedule to `nonOfficeHours`, so this change adds it.


rarkins (commented):
Although we went with a conservative "hours when someone is unlikely to be working" for nonOfficeHours, I don't think that "office hours" is a good term for 5am-10pm

martincostello:
I'm open to renaming it if the overall concept is welcome. I just went with "office hours" as it's the natural linguistic opposite of "non office hours" (i.e. `!nonOfficeHours == officeHours` due to the double negative).

Or I could leave the name as-is and change the hours to something like `8-17`/`9-16`/`10-15`.

rarkins:
I think 8am-6pm is reasonable for "office hours" if you'd like to keep the name.

rarkins (approved):

PR #36588, opened 2025-06-18
feat(presets/internal): group drupal packages

beto-aveiga:


lib/config/presets/internal/drupal.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/* eslint sort-keys: ["error", "asc", {caseSensitive: false, natural: true}] */
+
+export const presets: Record<string, Preset> = {
+  core: {
+    description: 'Group Drupal core updates together.',
+    packageRules: [
+      {
+        groupName: 'Drupal Core',
+        labels: ['drupal core'],
```
RahulGautamSingh: We don't add labels in presets as it is something that can very on a user basis

lib/config/presets/internal/drupal.ts
RahulGautamSingh, line 1: No need to create a new preset file for this.

You want to group presets by packageName so this will go in the group presets file.

RahulGautamSingh (commented):
Thank you for your PR!

Could you please make the following changes:

1. Sign the CLA
2. Change the PR title to: `feat(presets/internal)`: group `drupal` packages

beto-aveiga:
I will update this PR today or tomorrow. Thanks.

lib/config/presets/internal/group.ts
```diff
@@ -111,6 +111,15 @@ const staticGroups = {
       },
     ],
   },
+  drupalCore: {
+    description: 'Group Drupal core updates.',
+    packageRules: [
+      {
+        groupName: 'Drupal core',
+        matchPackageNames: ['drupal/core', '/^drupal/core-.*/'],
```
rarkins:
```suggestion
        matchPackageNames: ['drupal/core', 'drupal/core-*'],
```

I think this is the same and simpler?

rarkins (dismissed):

lib/config/presets/internal/group.ts
beto-aveiga: Yes! Thanks!

lib/config/presets/internal/group.ts
```diff
@@ -111,6 +111,15 @@ const staticGroups = {
       },
     ],
   },
+  drupalCore: {
```
viceice:
```suggestion
  'drupal-core': {
```
I would prefer this instead of camlcase

beto-aveiga: Sure!

RahulGautamSingh (approved):

rarkins (approved):

PR #36373, opened 2025-06-07
feat(nuget): add deprecation message for NuGet packages

FantasyTeddy:
## Changes

This change reads the deprecation message from NuGet packages and therefore allows the deprecated packages to be displayed in the dashboard.

## Context

If a NuGet package is deprecated, there is no hint from renovate in the dashboard, even though NuGet provides that information via its API.

## Documentation

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

PR #36589, opened 2025-06-18
fix(nuget): only deprecate entire package if all releases are deprecated

FantasyTeddy:
## Changes

This fixes the issue in discussion #36584 where the previous implementation (#36373) marked packages as deprecated that only had _some_ deprecated versions.

## Context

This was a misunderstanding from my side, how `isDeprecated` and `deprecationMessage` works.
- `isDeprecated`: marks one release/version of a package as deprecated
- `deprecationMessage`: marks the entire package as deprecated

The later is only the case if _all_ versions of a package are deprecated.

### DIFF
--- a/lib/modules/datasource/nuget/__fixtures__/nunit/v3_registration.json
+++ b/lib/modules/datasource/nuget/__fixtures__/nunit/v3_registration.json
@@ -118,6 +118,13 @@
             "@id": "https://api.nuget.org/v3/catalog0/data/2018.10.16.13.04.06/nunit.2.6.0.12051.json",
             "@type": "PackageDetails",
             "authors": "Charlie Poole",
+            "deprecation": {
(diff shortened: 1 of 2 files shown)


RahulGautamSingh:
Could you please, link to the source of truth you are referring to?

FantasyTeddy:
@RahulGautamSingh: What do you mean?

> The later is only the case if _all_ versions of a package are deprecated.

If this is what you mean, I don't have any source for that statement.
My reasoning is the following:
- If an intermediate version is being deprecated, there is no reason to deprecate the entire package.
- Even if the latest version is being deprecated, this does not necessarily mean that the entire package is deprecated, if there are earlier versions that aren't.
- Only when all versions of a package are deprecated, there is no doubt that the entire package is deprecated.

rarkins (approved):

Issue #36503, opened 2025-06-13
Bitbucket Cloud API token support


rarkins:
### Discussed in https://github.com/renovatebot/renovate/discussions/36499 (further details inside)

<div type='discussions-op-text'>

<sup>Originally posted by **szhang1200** June 13, 2025</sup>
### Tell us more.

Bitbucket Cloud has announced they are deprecating app passwords. The Renovate bot uses Bitbucket app passwords to call Bitbucket APIs and perform git operations.

https://community.atlassian.com/forums/Bitbucket-articles/Deprecation-notice-Bitbucket-Cloud-shifts-to-API-tokens-from-app/ba-p/3040975#M2462

Bitbucket API tokens are personal access tokens designed to be long term replacement for app passwords.

There will be a 12 month deprecation period. On June 9, 2026 app passwords will stop working. **I propose that Renovate adds the option to integrate via Bitbucket API tokens.** 

There are some differences in how API tokens will work compared to app passwords. For Bitbucket APIs, a basic auth with Atlassian email + API token is required. For git operations they require a static username `x-bitbucket-api-token-auth` + API token. See https://support.atlassian.com/bitbucket-cloud/docs/using-api-tokens/

The proposal is to implement support for API tokens now. In 12 months time deprecate app passwords as an integration option. Would be great to get some ideas on this :)

</div>

nastoychev:
Guys would be possible to update the documentation page as well?

szhang1200:
@rarkins I'm thinking we go with a solution like the following. Happy to hear your suggestion too.

1) Add an `email` option for the bot account, which is the bitbucket email
2) Add option for `api_token` in the `config.js` file, add option to set API token in env var `API_TOKEN`, and add option in CLI `--api-token=`
3) Have both app password and API token supported simultaneously for Bitbucket
4) When using API tokens, for calling Bitbucket APIs we basic auth with email and the API token (note API token is not a bearer token auth)
5) When using git operations use the static username `x-bitbucket-api-token-auth` + API token

rarkins:
I don't think there's a need to add any new fields

szhang1200:
Reuse existing fields? ie `username` will be bitbucket email and then `password` is the API token?

rarkins:
Where do you need to use the email? We already have a gitAuthor field if you're talking about for git committing

rarkins:
[...]

PR #36614, opened 2025-06-20
feat: add support for Bitbucket API tokens

szhang1200:
## Changes

When users set up Renovate on Bitbucket with API tokens, they would:

- set the `username` field with Atlassian email
- set the `password` field with the API token (prefix of ATAT)

The docs should be updated to direct users to authenticate with the method above.

Please note for git operations if we detect the password to be an API token, we can authenticate with a static username `x-bitbucket-api-token-auth`.

This code change is designed to be backwards compatible so it does not break existing integrations with app passwords.

## Context

Renovate to support Bitbucket API tokens since app passwords are being deprecated.

https://github.com/renovatebot/renovate/issues/36503

### DIFF
--- a/lib/modules/platform/bitbucket/index.ts
+++ b/lib/modules/platform/bitbucket/index.ts
@@ -73,12 +73,13 @@ let renovateUserUuid: string | null = null;
 export async function initPlatform({
   endpoint,
   username,
+  email,
   password,
   token,
 }: PlatformParams): Promise<PlatformResult> {
-  if (!(username && password) && !token) {
+  if (!(username && password) && !(email && password) && !token) {
     throw new Error(
-      'Init: You must configure either a Bitbucket token or username and password',
+      'Init: You must configure either a Bitbucket access token or email and password',
     );
   }
   if (endpoint && endpoint !== BITBUCKET_PROD_ENDPOINT) {
@@ -92,6 +93,9 @@ export async function initPlatform({
   const options: HttpOptions = { memCache: false };
   if (token) {
     options.token = token;
+  } else if (email) {
+    options.username = email;
+    options.password = password;
   } else {
     options.username = username;
     options.password = password;
@@ -253,9 +257,15 @@ export async function initRepo({
   // TODO #22198
   const hostnameWithoutApiPrefix = regEx(/api[.|-](.+)/).exec(hostname!)?.[1];
 
-  const auth = opts.token
-    ? `x-token-auth:${opts.token}`
-    : `${opts.username!}:${opts.password!}`;
+  let auth = '';
+  if (opts.token) {
+    auth = `x-token-auth:${opts.token}`;
+  } else if (opts.password?.startsWith('ATAT')) {
+    auth = `x-bitbucket-api-token-auth:${opts.password}`;
+  } else {
+    auth = `${opts.username!}:${opts.password!}`;
+  }
+
   const url = git.getUrl({
     protocol: 'https',
     auth,
--- a/lib/modules/platform/bitbucket/index.ts
+++ b/lib/modules/platform/bitbucket/index.ts
@@ -78,7 +78,7 @@ export async function initPlatform({
 }: PlatformParams): Promise<PlatformResult> {
   if (!(username && password) && !token) {
     throw new Error(
-      'Init: You must configure either a Bitbucket access token or email and password',
+      'Init: You must configure either a Bitbucket token or username and password',
     );
   }
   if (endpoint && endpoint !== BITBUCKET_PROD_ENDPOINT) {
--- a/lib/modules/platform/bitbucket/index.ts
+++ b/lib/modules/platform/bitbucket/index.ts
@@ -73,11 +73,10 @@ let renovateUserUuid: string | null = null;
 export async function initPlatform({
   endpoint,
   username,
-  email,
   password,
   token,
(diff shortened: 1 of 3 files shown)


lib/modules/platform/bitbucket/index.ts
```diff
@@ -73,12 +73,13 @@ let renovateUserUuid: string | null = null;
...
   username,
+  email,
   password,
   token,
 }: PlatformParams): Promise<PlatformResult> {
-  if (!(username && password) && !token) {
+  if (!(username && password) && !(email && password) && !token) {
     throw new Error(
-      'Init: You must configure either a Bitbucket token or username and password',
+      'Init: You must configure either a Bitbucket access token or email and password',
```
rarkins: This seems wrong for now because username/password is still supported

rarkins (changes requested):
Can you retry by reusing "username" instead of adding "email"? I think the logic in your code would be essentially identical (in fact, removing a few lines)

lib/modules/platform/bitbucket/index.ts
```diff
@@ -73,12 +73,13 @@ let renovateUserUuid: string | null = null;
 export async function initPlatform({
   endpoint,
   username,
+  email,
   password,
   token,
 }: PlatformParams): Promise<PlatformResult> {
-  if (!(username && password) && !token) {
+  if (!(username && password) && !(email && password) && !token) {
```
rarkins: If using username, this could be reverted

lib/modules/platform/bitbucket/index.ts
```diff
@@ -92,6 +93,9 @@ export async function initPlatform({
   const options: HttpOptions = { memCache: false };
   if (token) {
     options.token = token;
+  } else if (email) {
```
rarkins: This block could also be reverted

lib/modules/platform/bitbucket/index.ts
```diff
@@ -253,9 +257,15 @@ export async function initRepo({
   // TODO #22198
   const hostnameWithoutApiPrefix = regEx(/api[.|-](.+)/).exec(hostname!)?.[1];
 
-  const auth = opts.token
-    ? `x-token-auth:${opts.token}`
-    : `${opts.username!}:${opts.password!}`;
+  let auth = '';
+  if (opts.token) {
+    auth = `x-token-auth:${opts.token}`;
+  } else if (opts.password?.startsWith('ATAT')) {
```
rarkins, line 259: This would stay the same if using username/password

lib/modules/platform/types.ts
```diff
@@ -14,6 +14,7 @@ export interface PlatformParams {
   endpoint?: string;
   token?: string;
   username?: string;
+  email?: string;
```
rarkins: Remove this. But if it were to stay, you'd need to add it to config options as it must be exposed to users for configuration

szhang1200: If we don't add this field then I don't think any changes are required? Only the docs need to be changed

szhang1200: Sorry i realised we still need that static username since `username` would become an email

szhang1200:
> Can you retry by reusing "username" instead of adding "email"? I think the logic in your code would be essentially identical (in fact, removing a few lines)

lib/modules/platform/bitbucket/index.ts
szhang1200: Here we still need to detect an API token, because otherwise the username could be an email address

lib/modules/platform/bitbucket/index.ts
szhang1200: I'll revert this back

lib/modules/platform/bitbucket/index.ts
rarkins: Yes, sorry that's what I meant. Your _changes_ would remain

rarkins (approved):
Nice and clean! Have you been able to verify that it _works_ ?

szhang1200:
> Nice and clean! Have you been able to verify that it _works_ ?

@rarkins I'm verifying these changes right now, I'me encountering errors related to `punycode` when I `pnpm start`, not sure if you have some workaround here.

szhang1200:
> Nice and clean! Have you been able to verify that it _works_ ?

@rarkins I commented out the stuff related to `punycode` and I can verify renovate works with both API tokens and app passwords. I was able to verify with my test repo, when I point renovate to my repo I can see PR created with both API token and app password.

rarkins (changes requested):
Just needs docs updated now

szhang1200:
> Just needs docs updated now

I updated the tests so now it should cover the new code path. Docs are in another place I assume?

rarkins (approved):

lib/modules/platform/bitbucket/index.ts
```diff
@@ -253,9 +253,15 @@ export async function initRepo({
...
```
viceice, line 259: hopefully no user is using that prefix for their real password. 🫣

viceice:
the docs should also have a hint about the API token auth.

lib/modules/platform/bitbucket/index.ts
szhang1200: @viceice This should be fine the password here should be an app password from Bitbucket, they all have ATBB prefix

PR #36618, opened 2025-06-20
fix: don't reuse branch if results change

### LINKED ISSUE #36615: Renovate looping bug for grouped dependencies over multiple files
### Discussed in https://github.com/renovatebot/renovate/discussions/35524

<sup>Originally posted by **samgiz** April 24, 2025</sup>

None

GitHub, Renovate version 39.238.1

Renovate does not handle updates for grouped packages correctly when they appear in multiple files, and a new package update appears.

I've observed this bug specifically when updating `tauri-monorepo` packages, which groups together cargo and pnpm packages. Seemingly there is a race condition when different packages get added to npm / cargo repositories, and then Renovate ends up:

1. Creating a PR with some of the updates
2. Finds an update for another package in the group at a later time
3. Gets in an infinite loop of either only updating cargo or pnpm packages in the PR.

I've done a small investigation into what causes this behaviour and it's approximately:

1. See that there's a new package update so the PR needs updating
2. Find all the package files that need updating
3. Check out to the base branch
4. Only apply the update to files that need updating, make the commit and force-push the PR

This seems like a fundamentally wrong approach to me. Renovate should be including files that are already part of the PR but don't need to be changed, otherwise you end up with the looping behaviour as observed.

I've created a [reproducer](https://github.com/samgiz/renovate-35524-reproducer).

### Logs (if relevant)

_No response_</div>

rarkins:
## Context

- Closes #36615


viceice (commented):
needs test for coverage

viceice (approved):

Issue #29536, opened 2024-06-08
Allow configurable repo config file names


rarkins:
### Describe the proposed change(s).

The list of repo config fileNames is currently hardcoded in https://github.com/renovatebot/renovate/blob/main/lib/config/app-strings.ts

We should allow for this list to be configurable, including which file name to use as _default_.

Blocked by #29535

RahulGautamSingh:
We could expose a file config option `configFileNames` which will be an array of names the user can configure for the bot to use.

After detecting this from file config, we can prepend this list to the internal `configFileNames` array.

For default, we can use the first element from the user configured array(if present)

rarkins:
We also need to think about how sometimes we rely on renovate.json specially. Perhaps we should instead take the first element in the array, which would allow the user to reorder it (doesn't matter if a file name appears twice)

PR #36630, opened 2025-06-20
feat(config): add support for JSONC

Erwan-loot:
## Changes

Add support of JSONC files discussed here https://github.com/renovatebot/renovate/pull/36141

## Context

As discussed [here](https://github.com/renovatebot/renovate/pull/36141) the support on the syntax JSONC is now deployed but the files JSONC itself not yet, and when we want to using JSONC files to permit the IDE  ( like VSCODE) to validate the syntax if we put some comments

### DIFF
--- a/lib/config/app-strings.ts
+++ b/lib/config/app-strings.ts
@@ -1,12 +1,16 @@
 export const configFileNames = [
   'renovate.json',
   'renovate.json5',
+  'renovate.jsonc',
   '.github/renovate.json',
   '.github/renovate.json5',
+  '.github/renovate.jsonc',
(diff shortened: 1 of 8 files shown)


viceice (changes requested):
this needs a refactor first. we don't want this massive extend of the file list. 

first the platform specific folder (like `.github/`) should only be searched on the current platform.
so it should not check the `.gitlab/` folder on GitHub or Gitea.

viceice:
https://github.com/renovatebot/renovate/issues/29536

rarkins:
We will implement this a different way

PR #36635, opened 2025-06-20
feat(datasource/unity3d): add pagination

bdovaz:
## Changes

Add pagination to Unity releases API:

https://services.docs.unity.com/release/v1/#tag/Release/operation/getUnityReleases

## Context

- https://github.com/renovatebot/renovate/discussions/36612


lib/modules/datasource/unity3d/index.ts
```diff
@@ -59,27 +60,36 @@ export class Unity3dDatasource extends Datasource {
...
+      );
+
+      for (const release of response.body.results) {
+        result.releases.push({
+          version: withHash
+            ? `${release.version} (${release.shortRevision})`
+            : release.version,
+          releaseTimestamp: asTimestamp(release.releaseDate),
+          changelogUrl: release.releaseNotes.url,
+          isStable: translatedRegistryUrl === Unity3dDatasource.streams.lts,
```
astellingwerf: There's not need to recalculate this value every time. If you declare a `const isStable: boolean = translatedRegistryUrl === Unity3dDatasource.streams.lts;` after line 61, you can simplify this line to just:
```suggestion
          isStable,
```

astellingwerf:
I'd say it's a good idea to have at least one test case that reads more than just one page.

bdovaz:
> I'd say it's a good idea to have at least one test case that reads more than just one page.

Done, I created it programmatically without new fixtures as I know the convention is to try to avoid them.

RahulGautamSingh:
> Done, I created it programmatically without new fixtures as I know the convention is to try to avoid them.

There are some fixtures changes in this PR, what would they be?

bdovaz:
> > Done, I created it programmatically without new fixtures as I know the convention is to try to avoid them.
> 
> There are some fixtures changes in this PR, what would they be?
> 

I have only changed the value of the “total” field because it had the original value when I created it at the time and it was a value that did not represent the amount of results that the file actually has.

This was not important until this PR because this field was not used but now when it is used it has to represent the data correctly.

viceice (changes requested):
please reduce the huge fixtures to only contain required stuff

lib/modules/datasource/unity3d/index.ts
```diff
@@ -20,6 +20,7 @@ export class Unity3dDatasource extends Datasource {
     stable: `${Unity3dDatasource.homepage}releases/editor/releases.xml`,
     beta: `${Unity3dDatasource.homepage}releases/editor/beta/latest.xml`,
   };
+  static readonly limit: number = 25;
```
viceice, line 23: use a higher limit if possible (like 100) to have less API requests

lib/modules/datasource/unity3d/index.ts
```diff
@@ -59,27 +60,39 @@ export class Unity3dDatasource extends Datasource {
...
     const translatedRegistryUrl = this.translateStream(registryUrl!);
 
-    const response = await this.http.getJson(
-      translatedRegistryUrl,
-      UnityReleasesJSON,
-    );
+    const isStable: boolean =
+      translatedRegistryUrl === Unity3dDatasource.streams.lts;
+
+    let total: number | undefined = undefined;
```
viceice:
```suggestion
    let total: number | null = null;
```
we prefer null over undefined

lib/modules/datasource/unity3d/index.ts
```diff
@@ -59,27 +60,39 @@ export class Unity3dDatasource extends Datasource {
...
-      result.releases.push({
-        version: withHash
-          ? `${release.version} (${release.shortRevision})`
-          : release.version,
-        releaseTimestamp: asTimestamp(release.releaseDate),
-        changelogUrl: release.releaseNotes.url,
-        isStable: translatedRegistryUrl === Unity3dDatasource.streams.lts,
-      });
-    }
+    do {
```
viceice: use a for loop

lib/modules/datasource/unity3d/index.ts
bdovaz: In the discussion from which this PR originated I commented, the maximum is 25:

image

bdovaz:
> please reduce the huge fixtures to only contain required stuff

I have left empty the `downloads` array, which is the one that occupied the most space and was not used.

lib/modules/datasource/unity3d/index.spec.ts
```diff
@@ -23,12 +24,38 @@ describe('modules/datasource/unity3d/index', () => {
     }
   };
 
+  const createUnityReleases = (
```
viceice: use normal function declaration instead of lambda

lib/modules/datasource/unity3d/index.spec.ts
```diff
@@ -260,4 +301,31 @@ describe('modules/datasource/unity3d/index', () => {
...
+    const uriPageTwo = new URL(
+      `${Unity3dDatasource.streams.lts}&limit=25&offset=25`,
+    );
+
+    const total = 30;
+
+    httpMock
+      .scope(uriPageOne.origin)
+      .get(`${uriPageOne.pathname}${uriPageOne.search}`)
+      .reply(200, JSON.stringify(createUnityReleases(total, 0, 25)));
```
viceice:
```suggestion
      .reply(200, createUnityReleases(total, 0, 25));
```
nock will handle it, you don't need to serialize here.

lib/modules/datasource/unity3d/index.spec.ts
```diff
@@ -260,4 +301,31 @@ describe('modules/datasource/unity3d/index', () => {
...
+    httpMock
+      .scope(uriPageTwo.origin)
+      .get(`${uriPageTwo.pathname}${uriPageTwo.search}`)
+      .reply(200, JSON.stringify(createUnityReleases(total, 25, 5)));
```
viceice:
```suggestion
      .reply(200, createUnityReleases(total, 25, 5));
```

viceice (commented):
otherwise LGTM

lib/modules/datasource/unity3d/index.spec.ts
```diff
@@ -23,12 +24,38 @@ describe('modules/datasource/unity3d/index', () => {
...
+  const createUnityReleases = (
+    total: number,
+    offset: number,
+    resultCount: number,
+  ) {
+    const results = [];
+
+    for (let i = 0; i < resultCount; i++) {
+      results.push({
+        version: `6000.0.${offset + i + 1}f1`,
```
astellingwerf: If you loop from 1 up to and including the `resultCount`, you don't need to increment every single loop:
```suggestion
    for (let i = 1; i <= resultCount; i++) {
      results.push({
        version: `6000.0.${offset + i}f1`,
```

bdovaz:
Ready to merge

astellingwerf (approved):

RahulGautamSingh (approved):

lib/modules/datasource/unity3d/index.spec.ts
```diff
@@ -23,12 +24,38 @@ describe('modules/datasource/unity3d/index', () => {
     }
   };
 
+  function createUnityReleases(
+    total: number,
+    offset: number,
+    resultCount: number,
+  ) {
```
viceice:
```suggestion
  ): UnityReleasesJSON {
```
add return type

bdovaz: Compilation errors with this suggestion

viceice (dismissed):

lib/modules/datasource/unity3d/schema.ts
```diff
@@ -12,5 +12,6 @@ const UnityRelease = z.object({
 });
 
 export const UnityReleasesJSON = z.object({
+  total: z.number(),
   results: UnityRelease.array(),
 });
```
viceice, line 17:
```suggestion
});
export type UnityReleasesJSON = z.infer<typeof UnityReleasesJSON>;
```

https://github.com/renovatebot/renovate/blob/a930a363a536a378ca7eb9f7d23dd92b511043b7/lib/modules/datasource/aws-eks-addon/schema.ts#L18

viceice (approved):

astellingwerf (approved):

viceice (approved):

PR #36644, opened 2025-06-22
feat: configurable repo config filenames

### LINKED ISSUE #29536: Allow configurable repo config file names

RahulGautamSingh:
## Changes
- Add a new file config option named `configFileNames`, which will allow users to use a custom filename for the repository config

## Context
- Closes: #29536 

Started this so we can move forward on the above issue

Onboarding: https://github.com/Rahul-renovate-testing/onboarding-with-custom-file/pull/3
Normal Run: https://github.com/Rahul-renovate-testing/onboarding-with-custom-file


viceice (commented):
conflicted

lib/config/app-strings.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+describe('config/app-strings', () => {
+  beforeEach(() => {
+    GlobalConfig.reset();
+  });
+
+  it('returns platform specific config filenames if platform detected', () => {
+    GlobalConfig.set({ platform: 'github' });
+    const filenames = getConfigFileNames();
+    expect(filenames.includes('.gitlab/renovate.json')).toBeFalse();
```
viceice: technically this is a breaking change, not sure if people depend on this 🤔

@rarkins @secustor wdyt?

lib/config/validation.ts
```diff
@@ -44,6 +44,7 @@ import {
 } from './validation-helpers/utils';
 
 const options = getOptions();
+const configFileNames = getConfigFileNames();
```
viceice: they custom filenames are not configured at this stage 🤔

lib/modules/manager/renovate-config-presets/index.ts
```diff
@@ -8,7 +8,7 @@ export { extractPackageFile } from './extract';
 export const url = '../../../config-presets.md';
 
 export const defaultConfig = {
-  managerFilePatterns: configFileNames.filter(
+  managerFilePatterns: getConfigFileNames().filter(
```
viceice, line 11: same here global config isn't set when this module scope code is executed

lib/workers/repository/init/apis.ts
```diff
@@ -13,6 +14,14 @@ export type WorkerPlatformConfig = RepoResult &
   RenovateConfig &
   Record<string, any>;
 
+const configFileNames = getConfigFileNames();
+
+// TODO #22198
+const getDefaultConfigFile = (config: RenovateConfig): string =>
```
viceice: use normal function declaration , no lambda

lib/workers/repository/init/merge.spec.ts
```diff
@@ -209,9 +211,11 @@ describe('workers/repository/init/merge', () => {
         configFileName: '.github/renovate.json',
         configFileParsed: {},
       });
+      GlobalConfig.reset();
```
viceice: do it in after or before each, otherwise it's bubbling to other tests on failure

lib/workers/repository/onboarding/branch/create.ts
```diff
@@ -7,6 +8,17 @@ import { getDefaultConfigFileName } from '../common';
 import { OnboardingCommitMessageFactory } from './commit-message';
 import { getOnboardingConfigContents } from './config';
 
+const defaultConfigFile = getConfigFileNames()[0];
```
viceice: inline to take user config into account

lib/workers/repository/onboarding/common.ts
```diff
@@ -7,6 +7,7 @@ export function getSemanticCommitPrTitle(config: RenovateConfig): string {
   return `${config.semanticCommitType ?? 'chore'}: ${config.onboardingPrTitle}`;
 }
 
+const configFileNames = getConfigFileNames();
```
viceice: same as other places, user config isn't available at this stage

viceice (changes requested):

lib/config/app-strings.spec.ts
secustor: hmm I agree with you that this is a breaking. 

I guess nobody will rely this as I can not see someone using `.gitlab` on GitHub. 
IMO we should keep backwards compatibility and create a separate PR to only support the host specific dot files.

lib/config/app-strings.ts
```diff
@@ -10,3 +13,32 @@ export const configFileNames = [
   '.renovaterc.json5',
   'package.json',
 ];
+
+let userAddedConfigFileNames: string[] = [];
+
+export function addConfigFileNames(fileNames: string[]): void {
```
secustor:
```suggestion
export function setUserConfigFileNames(fileNames: string[]): void {
```
nitpick you are overwriting and not adding stuff.

lib/config/app-strings.ts
```diff
@@ -10,3 +13,32 @@ export const configFileNames = [
...
+  userAddedConfigFileNames = fileNames;
+}
+
+export function getConfigFileNames(): string[] {
+  let allFileNames = [...userAddedConfigFileNames, ...configFileNames];
+  const platform = GlobalConfig.get('platform');
+  // should not happen
+  if (!platform) {
+    return allFileNames;
```
secustor: We should log a warning in that case

docs/usage/configuration-options.md
secustor, line 1: We should add also a reference in the on boarding guide https://github.com/renovatebot/renovate/blob/main/docs/usage/getting-started/installing-onboarding.md#configuration-location

lib/workers/repository/onboarding/branch/create.ts
```diff
@@ -7,6 +8,17 @@ import { getDefaultConfigFileName } from '../common';
...
+
+export function getDefaultConfigFileName(
```
secustor: You add two functions with the same purpose

lib/util/git/index.ts
```diff
@@ -1176,7 +1176,7 @@ export async function prepareCommit({
         try {
           /* v8 ignore next 2 -- TODO: add test */
           const addParams =
-            fileName === configFileNames[0] ? ['-f', fileName] : fileName;
+            fileName === getConfigFileNames()[0] ? ['-f', fileName] : fileName;
```
secustor, line 1192: This seems wrong to me. 

Anyone aware why this does only take `renovate.json` into consideration?

docs/usage/self-hosted-configuration.md
```diff
@@ -398,6 +398,25 @@ It is highly unlikely that you should ever need to add this to your permanent gl
...
+```json
+{
+  "configFileNames": ["myrenovate.json"]
+}
+```
+
+<!-- prettier-ignore -->
+!!! note
+    If you want renovate to use a custom filename for the onboarding branch.
+    You also need to change the [`onboardingConfigFileName`](#onboardingconfigfilename).
```
astellingwerf: Isn't this supposed to be a single sentence?

docs/usage/self-hosted-configuration.md
```diff
@@ -906,6 +925,11 @@ If `commitMessagePrefix` or `semanticCommits` values are set then they will be p
 If set to one of the valid [config file names](./configuration-options.md), the onboarding PR will create a configuration file with the provided name instead of `renovate.json`.
 Falls back to `renovate.json` if the name provided is not valid.
 
+<!-- prettier-ignore -->
+!!! note
+    If you want renovate to use a custom filename for the onboarding branch.
+    You need add allow that filename using the [`configFileNames`](#configfilenames) option.
```
astellingwerf: Isn't this supposed to be a single sentence?

lib/config/app-strings.ts
```diff
@@ -10,3 +13,32 @@ export const configFileNames = [
...
+  }
+
+  allFileNames = allFileNames.filter((filename) => {
```
astellingwerf: Please include a notice of this filtering in the configuration documentation.

astellingwerf (changes requested):

lib/workers/repository/onboarding/common.ts
RahulGautamSingh: `configFileNames` are present in the file config.

lib/config/app-strings.ts
RahulGautamSingh: I messed up the conflict merge :)

lib/config/validation.ts
RahulGautamSingh: inlined it

`config-validator` will not be able to get the custom filenames still. Will need to mention this in the docs.

lib/modules/manager/renovate-config-presets/index.ts
RahulGautamSingh: What should we do here, then?

Loosen the patterns or add a comment in the docs asking users to add filepatterns to this manager if they want to use it with custom config filenames.

RahulGautamSingh: Yes 

We could ask users to include a package rule in their config that extends the manageFilePatterns of `renovate-config-preset` manager if they are using custom filenames?

lib/util/git/index.ts
RahulGautamSingh: @rarkins Do you recall why this was added?

lib/config-validator.ts
RahulGautamSingh, line 1: Need to mention in the docs, that if custom filename is used and one wants run config-validator on it. The command needs to include the custom filename too. As the command `pnpm config-validator` still checks the default list.

viceice (dismissed):
docs could be improved later

lib/workers/global/config/parse/file.ts
rarkins, line 1: Please ensure that this new feature works with inherited config too. You may need to move some logic out of this file here

rarkins (changes requested):

lib/workers/global/config/parse/file.ts
rarkins: See https://docs.renovatebot.com/self-hosted-configuration/#inheritconfig and https://docs.renovatebot.com/mend-hosted/hosted-apps-config/#inherited-config

secustor (commented):
conflicted, else LGTM

lib/workers/global/config/parse/file.ts
RahulGautamSingh: Instead of moving the code section to where we merge inherited config. I think we need to call it at both places using a utility function. Else, it will not work if inherited config is not in use.

secustor (commented):
has conflicts

lib/workers/repository/onboarding/branch/index.spec.ts
viceice:
```suggestion
      ).toHaveBeenCalledExactlyOnceWith(expectConfig, getConfigFileNames()[0]);
```

lib/util/git/index.ts
secustor: I think we should change to include all possible and configured file names

secustor (commented):
has conflicts and see open conversation

lib/util/git/index.ts
RahulGautamSingh: I want to do it in another PR, so it's easier to revert in case something breaks, however unlikely that is.

secustor (approved):

jamietanna:
(This was incorrectly tagged as the cause of an issue)

PR #36659, opened 2025-06-23
docs(composer): private repositories must include the package name

junowilderness:
## Changes

Update the private package documentation for Composer to indicate that the `name` key is required.

## Context

This requirement was noticed in https://github.com/renovatebot/renovate/discussions/35609#discussioncomment-13036012 but is otherwise undocumented. It is true.

### DIFF
--- a/docs/usage/getting-started/private-packages.md
+++ b/docs/usage/getting-started/private-packages.md
@@ -187,13 +187,15 @@ Avoid adding a `hostRule` with `hostType=github` because:
 - it overrides the default Renovate application token for everything else
 - it causes unwanted side effects
 
-The repository in `composer.json` should have the `vcs` type with a `https` URL.
+The repository in `composer.json` should have the `vcs` type with a `https` URL
+and must include the package's name.
 For example:
 
 ```json
 {
   "repositories": [
     {
+      "name": "organization/private",
       "type": "vcs",
       "url": "https://github.com/organization/private-repository"
     }


rarkins:
Should be confirmed in source code first to be sure this statement is correct

junowilderness:
It doesn't work without it. If that's a bug—that's another matter. I would gladly set up a public reproduction if there was a way to do so without leaking a secret token.

rarkins:
That's your opinion/experience, not source code.

junowilderness:
It's that of at least two others, not only me. That said, it may be that authentication is uninvolved and a public reproduction of https://github.com/renovatebot/renovate/discussions/35609 may be trivial for me to set up.

rarkins:
By source code I meant within renovate, eg https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/composer/schema.ts renovate/lib/modules/manager/composer at main · renovatebot/renovate

junowilderness:
Thank you for pointing to the precise file. I think private Composer repos are discovered as `NamedRepo`, and for that to work [NamedRepo makes the `name` key required](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/composer/schema.ts#L64).

RahulGautamSingh:
The `ComposerRepo` and `PackageRepo` do not require the `name` field to be mandatory as per their respective schemas.
https://github.com/renovatebot/renovate/blob/76771a9991f922c0680be974a55bc407945851c6/lib/modules/manager/composer/schema.ts#L19

junowilderness:
Indeed, `name` is unnecessary _if the forge is Packagist_. In practice, for some reason, `name` is required for private repos on Gitlab. 

**I understand the only way forward is to create a public reproduction case.**

PR #36691, opened 2025-06-25
feat: Change junit5 monorepo to junit-framework

marcphilipp:
## Changes

Change junit5 monorepo to junit-framework since the junit5 GitHub repo has been renamed accordingly.

IIUC the documentation will be updated automatically.

### DIFF
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -382,7 +382,10 @@
     "jna": "https://github.com/java-native-access/jna",
     "json-smart-v2": "https://github.com/netplex/json-smart-v2",
     "jsplumb": "https://github.com/jsplumb/jsplumb",
-    "junit-framework": "https://github.com/junit-team/junit-framework",
+    "junit-framework": [
+      "https://github.com/junit-team/junit5",
+      "https://github.com/junit-team/junit-framework"
+    ],
     "kernelmemory": "https://github.com/microsoft/kernel-memory",
     "kiota": "https://github.com/microsoft/kiota",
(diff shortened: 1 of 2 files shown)


lib/data/monorepo.json
```diff
@@ -382,7 +382,7 @@
     "jna": "https://github.com/java-native-access/jna",
     "json-smart-v2": "https://github.com/netplex/json-smart-v2",
     "jsplumb": "https://github.com/jsplumb/jsplumb",
-    "junit5": "https://github.com/junit-team/junit5",
+    "junit-framework": "https://github.com/junit-team/junit-framework",
```
astellingwerf: This is technically a breaking change. Users that reference the `monorepo:junit5` preset will run into issues.

astellingwerf (changes requested):

lib/data/monorepo.json
marcphilipp: Should we add it as a new repo instead?

viceice: we need a migration. we've already some preset renamings. please check them.

viceice: add also the old url, so old versions are still properly grouped.

marcphilipp: Done in 75f3f8894dd881647c681c04db1bdfe2f68fb114

marcphilipp: Done in 3063e593d0df545bbe3f4e7afa8f213d4801e258

rarkins (approved):

viceice (approved):

astellingwerf (approved):

PR #36693, opened 2025-06-25
feat(manager/pre-commit): add support digest updating

### LINKED ISSUE #22567: Support digest updating in pre-commit manager

Support updating digests in pre-commit manager. For example this:

```
  - repo: https://github.com/python-jsonschema/check-jsonschema
    rev: a00caac4f0cec045f7f67d222c3fcd0744285c51 # frozen: 0.23.1
    hooks:
      - id: check-renovate
```

### Describe why we need/want these change(s).

See #22488

https://github.com/renovate-reproductions/22488

### LINKED ISSUE #22568: Support digest pinning in pre-commit manager

Support digest pinning for pre-commit manager

### Describe why we need/want these change(s).

See #22488

https://github.com/renovate-reproductions/22488

lobbin:
## Changes

Add support to work with pinned, "frozen", versions of pre-commit repo components.

## Context

- Closes #22567

Potentially also close #22568, but I know too little of renovate for that.


lobbin:
I'm not fully happy with the part that looks up and finds the comment with the frozen version

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -79,15 +81,20 @@ function determineDatasource(
...
-  currentValue?: string;
-} {
-  logger.debug(`Found version ${tag}`);
+  comment?: string,
+): PackageDependency {
+  let currentValue = tag;
+  let currentDigest: string | undefined;
+
+  if (tag.length === 40 && comment) {
+    const match = /.*frozen: (.+)$/.exec(comment);
```
RahulGautamSingh: import and use `regEx` wrapper here
```suggestion
    const match = regEx(/.*frozen: (.+)$/).exec(comment);
```

RahulGautamSingh (commented):
Could please, link the repo you tested it on?

lobbin:
> Could please, link the repo you tested it on?

This is the repo, all the history intact
https://github.com/lobbin/renovate-testing/tree/main

RahulGautamSingh:
In the reproduction you linked.

Could you show an example of where only the digest is updated while the frozen version remains the same?

It's possible incase of docker tags to have multiple digests for a single tag/version. I am not sure if the same applies to pre-commit.

lobbin:
> In the reproduction you linked.
> 
> Could you show an example of where only the digest is updated while the frozen version remains the same?
> 
> It's possible incase of docker tags to have multiple digests for a single tag/version. I am not sure if the same applies to pre-commit.

Here is an example where the version is the same, but the digest is updated:
https://github.com/lobbin/renovate-testing/pull/10

RahulGautamSingh (approved):

rarkins (dismissed):

RahulGautamSingh:
@lobbin, Please update this branch once, so the workflow is triggered and the checks pass.

lib/modules/manager/pre-commit/__fixtures__/pinned.pre-commit-config.yaml
```diff
@@ -0,0 +1,28 @@
+failfast: true
```
viceice: please inline it and use `codeBlock` also reduce to minimum

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -125,22 +137,51 @@ function extractDependency(
...
+            typeof revPair.value.comment === 'string'
+          ) {
+            repoComments[idx] = revPair.value.comment;
+          }
+        }
+      });
+    }
+  }
+
+  precommitFile.repos.forEach((item, idx) => {
```
viceice: convert to `for-of` loop

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -125,22 +137,51 @@ function extractDependency(
...
-  precommitFile.repos.forEach((item) => {
+
+  // Map for quick lookup of comments by repo index
+  const repoComments: Record<number, string | undefined> = {};
+  if (doc?.contents && isMap(doc.contents)) {
+    const reposPair = doc.contents.items.find(
+      (pair: any) => pair.key && pair.key.value === 'repos',
+    );
+    if (reposPair?.value && isSeq(reposPair.value)) {
+      reposPair.value.items.forEach((repoItem: any, idx: number) => {
```
viceice: use `for-of` loop

rarkins (commented):
Can the refactoring be separated?

lobbin:
Updated, fixed conflicts and lint issues

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -131,25 +143,54 @@ function extractDependency(
...
     return [];
   }
   const packageDependencies: PackageDependency[] = [];
-  precommitFile.repos.forEach((item) => {
+
+  // Map for quick lookup of comments by repo index
+  const repoComments: Record<number, string | undefined> = {};
+  if (doc?.contents && isMap(doc.contents)) {
+    const reposPair = doc.contents.items.find(
+      (pair: any) => pair.key?.value === 'repos',
```
RahulGautamSingh: don't use any

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -131,25 +143,54 @@ function extractDependency(
...
+    );
+    if (reposPair?.value && isSeq(reposPair.value)) {
+      for (const [idx, repoItem] of reposPair.value.items.entries()) {
+        if (repoItem && isMap(repoItem)) {
+          const revPair = repoItem.items.find(
+            (pair: any) => pair.key?.value === 'rev',
```
RahulGautamSingh: no any

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -131,25 +143,54 @@ function extractDependency(
...
+          );
+          if (
+            revPair?.value &&
+            isScalar(revPair.value) &&
+            typeof revPair.value.comment === 'string'
```
RahulGautamSingh:
```suggestion
            isString(revPair.value.comment)
```

import isString from @sindresorhus/is;

RahulGautamSingh (changes requested):
Else, LGTM

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -86,15 +89,20 @@ const gitUrlRegex = regEx(/\.git$/i);
...
-} {
-  logger.debug(`Found version ${tag}`);
+  comment?: string,
+): PackageDependency {
+  let currentValue = tag;
+  let currentDigest: string | undefined;
+
+  if (tag.length === 40 && comment) {
+    const match = regEx(/.*frozen: (.+)$/).exec(comment);
+    if (match?.[1]) {
```
viceice: use named capture group

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -112,13 +120,18 @@ function extractDependency(
...
         ...sourceDef,
         depName,
         depType: 'repository',
         packageName: depName,
-        currentValue: tag,
+        currentValue,
       };
+      if (currentDigest) {
+        dep.currentDigest = currentDigest;
+        dep.replaceString = `${currentDigest} # frozen: ${currentValue}`;
```
viceice: you should store the match to not changed any other stuff or tighten the regex

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -131,25 +144,56 @@ function extractDependency(
     datasource: undefined,
     packageName: undefined,
     skipReason: 'invalid-url',
-    currentValue: tag,
-  };
+    currentValue,
+  } as PackageDependency;
```
viceice:
```suggestion
  };
```

i don't think this is needed, typescript should be smart enough

viceice (commented):
otherwise lgtm

lib/modules/manager/pre-commit/extract.ts
lobbin: Not sure I fully understand. You mean the wildcard before "frozen"?

viceice: yes, there are users who prefer two spaces before the comment. see GitHub action manager , it's using regex to extract and reuses the match as replacement string

lobbin: Yeah, this is a miss from my side, when using `pre-commit autoupdate --freeze` you actually get two spaces before the comment. However, with the approach of extracting the comment via the yaml parser, it's not possible to do that. So either that we always use two spaces, or I need to do something similar to the github manager to use both a regexp and yaml parser.

viceice: using regex in addition to the yaml parser sounds good

lobbin:
> Hi there,
> 
> This PR appears to have been inactive for a while. Please let us know if you are still working on it, or if we can close it for now.
> 
> Thanks, the Renovate team

Yes, new update pending

lobbin:
Ok, I change a bit so it works more like the github-actions. Comments/digests/versions are now extracted using regex and then applied. Test was updated to make sure it works with different spaces

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -84,17 +84,50 @@ function determineDatasource(
...
+      continue;
+    }
+
+    const match = revLineWithFrozenCommentRegex.exec(line);
+    if (match?.groups) {
+      const {
+        currentDigest,
+        currentValue,
+        replaceString,
+        commentWhiteSpaces = ' ',
```
viceice:
```suggestion
        commentWhiteSpaces,
```
no need, will always match because at lease on single whitespace is required

viceice (dismissed):

lib/modules/manager/pre-commit/extract.ts
```diff
@@ -84,17 +84,50 @@ function determineDatasource(
...
+    }
+
+    const match = revLineWithFrozenCommentRegex.exec(line);
+    if (match?.groups) {
+      const {
+        currentDigest,
+        currentValue,
+        replaceString,
+        commentWhiteSpaces,
+      } = match.groups;
```
viceice:
```suggestion
      const { currentDigest, currentValue, replaceString, commentWhiteSpaces } =
        match.groups;
```

viceice (approved):

PR #36697, opened 2025-06-25
fix(galaxy-collection): suppress incorrect warning if collection is not found in registry

dmosen:
## Changes

Prevent logging of warnings if a collection is not found in a registry

## Context

If a collection is not found in one or more of the defined galaxy-collection registries, a warning is logged. This shows up in the logs and in the Dependency Dashboard. This is misleading and should be considered normal behavior.

### DIFF
--- a/lib/modules/datasource/galaxy-collection/index.ts
+++ b/lib/modules/datasource/galaxy-collection/index.ts
@@ -1,6 +1,7 @@
 import is from '@sindresorhus/is';
 import { logger } from '../../../logger';
 import { cache } from '../../../util/cache/package/decorator';
+import { HttpError } from '../../../util/http';
 import * as p from '../../../util/promises';
 import { regEx } from '../../../util/regex';
 import { ensureTrailingSlash, joinUrlParts } from '../../../util/url';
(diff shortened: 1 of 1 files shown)


lib/modules/datasource/galaxy-collection/index.ts
```diff
@@ -51,10 +52,12 @@ export class GalaxyCollectionDatasource extends Datasource {
     const { val: baseProject, err: baseErr } = await this.http
       .getJsonSafe(baseUrl, GalaxyV3)
       .onError((err) => {
-        logger.warn(
-          { url: baseUrl, datasource: this.id, packageName, err },
-          'Error fetching from url',
-        );
+        if (!(err instanceof HttpError && err.response?.statusCode === 404)) {
```
RahulGautamSingh, line 55:
```suggestion
        if (err instanceof HttpError && err.response?.statusCode !== 404) {
```
Does this achieve the same goal?

dmosen: I don't think so. Applying logical rules we could transform the condition to:
``!err instanceof HttpError || err.response?.statusCode !== 404``

But I see that it's both is not very understandable. Would the following be better?

```suggestion
        const isHttp404Error = err instanceof HttpError && err.response?.statusCode === 404;
        if (!isHttp404Error) {
```

Edit: I just realized that your solution solves the problem as well and just skips remaining error types. So you can decide which solution fits better.

RahulGautamSingh: The original is fine

RahulGautamSingh (approved):

rarkins (approved):

PR #36771, opened 2025-07-01
feat(presets): substitute entire parameter string for `{{args}}`

amezin:
## Changes

Allow passing values containing commas to presets. Substitute `{{args}}` in the preset with the entire string between parentheses.

## Context

I use presets for custom regex managers. I pass `managerFilePatterns` as a parameter. Glob syntax is much more convenient than regex, but glob syntax often needs commas.

### DIFF
--- a/lib/config/presets/types.ts
+++ b/lib/config/presets/types.ts
@@ -23,6 +23,7 @@ export interface ParsedPreset {
   presetName: string;
   tag?: string | undefined;
   params?: string[] | undefined;
+  paramString?: string | undefined;
 }
 
 export type PresetFetcher = (
--- a/lib/config/presets/types.ts
+++ b/lib/config/presets/types.ts
@@ -23,7 +23,7 @@ export interface ParsedPreset {
   presetName: string;
   tag?: string | undefined;
   params?: string[] | undefined;
-  paramString?: string | undefined;
+  rawParams?: string | undefined;
 }
 
 export type PresetFetcher = (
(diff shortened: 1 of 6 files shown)


rarkins (dismissed):

astellingwerf (changes requested):
This seems like it is the wrong solution for this problem. If the comma confuses the preset string parser, shouldn't we instead add support for escaping commas?

`{{args}}` may still be valuable for varargs kinda references, but that's an entirely different usecase.

amezin:
> This seems like it is the wrong solution for this problem. If the comma confuses the preset string parser, shouldn't we instead add support for escaping commas?

I use globs instead of regexes exactly to avoid double-escaping (one `\` for escaping in regex, one `\` for JSON, and you end up with something that's hard to read - and hard to write correctly for other team members).

Also, strictly speaking, such change won't be backwards-compatible.

> `{{args}}` may still be valuable for varargs kinda references, but that's an entirely different usecase.

There's (currently) no way to generate a json array, not even talking about anything more complex. So varargs are pretty much useless. You can't even use `arg*` as a variable in handlebars.

RahulGautamSingh (approved):

lib/config/presets/types.ts
```diff
@@ -23,6 +23,7 @@ export interface ParsedPreset {
   presetName: string;
   tag?: string | undefined;
   params?: string[] | undefined;
+  paramString?: string | undefined;
```
secustor:
```suggestion
  rawParams?: string | undefined;
```

I prefer this naming as it makes clear that this is raw value before parsing

amezin: I think variable/property naming in the code is less important (can be changed at any time) than the user-visible name. Should it be `{{rawArgs}}` instead of `{{args}}` then?

secustor: `{{args}}` as userfacing is preferable for me.

secustor (approved):

viceice (approved):

PR #36779, opened 2025-07-01
fix(gitea): prefer default merge method

viceice:
## Changes
Currently renovate just tries to guess the enabled merge methods and ignores the configured default merge method on repo.
This is probably partially breaking, as it's now following the default merge method instead of trying the first working method.

## Context

The `default_merge_style` is always filled but it seems you can set a merge method wich is disabled.
I consider that a platform bug and we don't need a fallback.

### DIFF
--- a/lib/modules/platform/gitea/index.ts
+++ b/lib/modules/platform/gitea/index.ts
@@ -320,15 +320,27 @@ const platform: Platform = {
       throw new Error(REPOSITORY_BLOCKED);
     }
 
-    if (repo.allow_rebase) {
+    if (repo.allow_rebase && repo.default_merge_style === 'rebase') {
       config.mergeMethod = 'rebase';
-    } else if (repo.allow_rebase_explicit) {
+    } else if (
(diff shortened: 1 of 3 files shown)


lib/modules/platform/gitea/index.ts
```diff
@@ -320,15 +320,27 @@ const platform: Platform = {
       throw new Error(REPOSITORY_BLOCKED);
     }
 
-    if (repo.allow_rebase) {
+    if (repo.allow_rebase && repo.default_merge_style === 'rebase') {
```
secustor, line 323: I think this is a good point to extract this to a function. 

Would it be even possible that `default_merge_style` and the `allow_` settings are conflicting? If no, we could simplify this.

viceice: yes, they can conflict if a user disables the default merge method. I consider that a user / platform error and renovate should stop processing that repo.

secustor (approved):

RahulGautamSingh (approved):

PR #36781, opened 2025-07-01
feat(presets): added fumadocs to mono repo groups

gperdomor:
## Changes

Add fumadocs to monorepo config presets

## Context

With this, fumadocs updates will be grouped

### DIFF
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -323,6 +323,7 @@
     ],
     "formatjs": "https://github.com/formatjs/formatjs",
     "framework7": "https://github.com/framework7io/framework7",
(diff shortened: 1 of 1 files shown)


RahulGautamSingh (approved):

rarkins (approved):

viceice:
@rarkins all packages of that repo have different versions, so the PRs are now always immortal.

I would revert this if no plausible reason to group them by default.

rarkins:
Can the packages be upgraded independently, or do they frequently break if upgraded one by one?

It would be best to group immortality of they frequently break, otherwise revert.

viceice:
no answer yet, let's revert

PR #36804, opened 2025-07-02
feat: Introduce support for minimumMinorAge configuration option

tomkerkhove:
## Changes

Introduce support for minimumMinorAge configuration option as discussed in https://github.com/renovatebot/renovate/discussions/35691

## Context

I'm not a TypeScript developer so I've worked with GitHub Copilot to implement this.


rarkins (changes requested):
You don't need a new feature to get this functionality. You could do matchUpdateTypes=minor and then set minimumReleaseAge. Unfortunately I don't think any of this code is useful.

I thought you wanted something different to this. Eg checking the age of the minor stream by looking at the age of x.y.0

tomkerkhove:
> I thought you wanted something different to this. Eg checking the age of the minor stream by looking at the age of x.y.0

@rarkins The goal we have is to auto-upgrade to `x.y.z` where we want to wait for `x.y` to be x days old. So that means if `x.y.0` is stable we'd upgrade to that, but if they pushed `0->4` we'd prefer to upgrade to `x.y.4` instead.

Are you saying matchUpdateTypes + `minimumReleaseAge` already handles that? Something similar to this?

```diff
{
  "packageRules": [
    {
      "groupName": "Azure SDK (Gateway)",
-     "matchUpdateTypes": ["minor", "patch"],
+     "matchUpdateTypes": ["minor"],
+     "minimumReleaseAge": "7 days",
      "matchPackageNames": ["Azure{.,}**", "Microsoft.Azure{.,}**", "Microsoft.ApplicationInsights{.,}**", "WindowsAzure{.,}**"],
      "excludePackagePrefixes": ["Azure.Identity","Microsoft.Azure.Api"],
      "additionalBranchPrefix": "gateway/",
      "matchFileNames": ["App/**/**"]
    },
  ]
}
```

rarkins:
@tomkerkhove in that case:
- I understood you correctly
- What you want it not possible today
- This PR was completely wrong (does not do what you want)

tomkerkhove:
I know I should not have trusted robots =)

Let me try to learn TypeScript and write it myself.

PR #36839, opened 2025-07-04
feat: Glob patterns for `cacheTtlOverride` keys

### LINKED ISSUE #36836: cacheTtlOverride: allow glob matching

Allow glob matching in `cacheTtlOverride`, including `*`, `datasource-*`, etc.

Retain the existing exact match and give those highest priority, otherwise if no exact match found then iterate through any values with a `*` and apply them in order (last wins).

zharinov:
## Context

- Closes: #36836


rarkins (changes requested):
Needs docs update

docs/usage/self-hosted-configuration.md
```diff
@@ -295,20 +295,46 @@ In the self-hosted setup, use option to enable caching of private packages to im
...
   }
 }
 ```
 
+In this example:
+
+- `datasource-docker-tags` gets 120 minutes (exact match)
+- `datasource-npm` gets 90 minutes (exact match)
+- Other datasources like `datasource-maven` get 15 minutes (matches `*` which is evaluated last)
+- Changelog namespaces like `changelog-github` get 30 minutes (regex match)
```
rarkins: Why is a regex match higher priority? Also, you didn't mention it in the "Priority order" earlier

zharinov: It's wrong, actually `*` wins for all non-exact matches. BTW, why not use "first match wins" strategy?

rarkins: I guess it's a possibility

zharinov: Let's do the "first match wins" because I think it's the behavior people expect by default.

rarkins: I spoke with @nabeelsaabna and we prefer longest match wins

zharinov: You mean the pattern with longest length, as the easy way to obtain "specificity" semantics?

rarkins: Yes

docs/usage/self-hosted-configuration.md
```diff
@@ -295,20 +295,46 @@ In the self-hosted setup, use option to enable caching of private packages to im
...
+You can use:
+
+- **Exact matches**: Direct namespace names
+- **Glob patterns**: Wildcards like `datasource-*` or `*`
+- **Regex patterns**: Regular expressions like `/^datasource-/`
+
+Priority order:
+
+1. Exact namespace matches take highest priority
+2. If no exact match, patterns are evaluated in the order they appear in the object, and the first matching pattern wins
```
rarkins: Should be longest match first

docs/usage/self-hosted-configuration.md
```diff
@@ -367,14 +396,15 @@ Valid codes for namespaces are as follows:
 - `datasource-maven:cache-provider`
 - `datasource-maven:postprocess-reject`
 - `datasource-node-version`
-- `datasource-npm:data`
+- `datasource-npm:cache-provider`
```
rarkins: Was this wrong? Or did it change?

zharinov: Yes, the list was a bit outdated

zharinov: Maybe we need to point at the source code instead

rarkins: No, it needs to be docs. Maybe we need a test to ensure it matches. When did this change happen?

lib/util/cache/package/ttl.spec.ts
```diff
@@ -0,0 +1,380 @@
...
+        });
+      });
+
+      it('uses fallback when override is not a number', () => {
+        GlobalConfig.set({
+          cacheTtlOverride: {
+            'datasource-npm': 'invalid',
+          },
+        });
+        const result = resolveTtlValues('datasource-npm' as never, 60);
```
viceice, line 356:
```suggestion
        const result = resolveTtlValues('datasource-npm' as const, 60);
```
doesn't this work better? 🤔

docs/usage/self-hosted-configuration.md
rarkins: Is adding a test to ensure it's in-sync feasible?

lib/util/cache/package/ttl.spec.ts
zharinov: Not assignable, I just wanted to use whatever strings because it doesn't important to use only existing namespaces here

docs/usage/self-hosted-configuration.md
zharinov: https://github.com/renovatebot/renovate/pull/36909

rarkins (approved):

PR #36842, opened 2025-07-04
feat: Implement `cacheTtlOverride` for datasource index cache

zharinov:
## Changes

- Ref: #36836


zharinov:
It leads to cache resets. It breaks overrides for those who overrided `datasource-releases` namespace.

docs/usage/self-hosted-configuration.md
```diff
@@ -373,7 +373,7 @@ Valid codes for namespaces are as follows:
 - `datasource-packagist`
 - `datasource-pod`
 - `datasource-python-version`
-- `datasource-releases`
+- `datasource-releases-{{datasource}}`: special short-lived cache for final results of the particular datasource fetches
```
rarkins: This is not clear to me. How/where/why is it used instead of other ones?

rarkins (changes requested):

docs/usage/self-hosted-configuration.md
zharinov: The full key is being formed like `{namespace}-{key}`. This cause confusion when the namespace is `datasource-releases` and key is `pypi {registry} {package}`, because the full key would be `datasource-releases-pypi {registry} {package}`. Since override functionality only works with namespace part and ignores the key part, trying to override `datasource-releases-pypi` leads to confusion because there is no such a namespace.

rarkins: So users should still use datasource-releases?

zharinov: No, they should use `datasource-releases-*` if they want to override all such namespaces, or they could selectively do `datasource-releases-{npm,maven}`, or they could directly match against `datasource-releases-pypi`. I just moved the datasource part from the key part to the namespace part in this PR.

zharinov: Let's merge this one first, so that I could reflect this in documentation:
- #36839

lib/util/cache/package/types.ts
```diff
@@ -94,7 +94,7 @@ export type PackageCacheNamespace =
   | 'datasource-packagist'
   | 'datasource-pod'
   | 'datasource-python-version'
-  | 'datasource-releases'
+  | `datasource-releases-${string}`
```
rarkins: Please see lines 19-23 and tell me if this is a "breaking" change?

zharinov: It resets the caches and it breaks overrides for those who uses `datasource-releases` as the key to match against.

rarkins (dismissed):

rarkins (approved):

viceice (approved):

PR #36843, opened 2025-07-04
fix(gradle): update dependency verification metadata also without `verify-metadata` enabled

Churro:
## Changes

Enables refreshing dependency checksums during gradle artifact updates when `<verify-metadata>false</verify-metadata>`. Until now, this was only done if metadata verification was also enabled. This covers cases, such as seen with [groovy](https://github.com/apache/groovy/blob/master/gradle/verification-metadata.xml#L42) or [elasticsearch](https://github.com/elastic/elasticsearch/blob/5e16bc3fa615d76a5f188e0b722691da2981e633/gradle/verification-metadata.xml#L4-L5), where checkums are present for JARs (that gradle is verifying), but where no additional metadata verification is enabled.

## Context

- https://github.com/renovatebot/renovate/discussions/32747
- gradle documentation on [disabling metadata verification](https://docs.gradle.org/current/userguide/dependency_verification.html#sec:disabling-metadata-verification)
- [gradle source code](https://github.com/gradle/gradle/blob/b06b7de4572c102a02ed831ae1f37e2da8255c35/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/verification/verifier/DependencyVerifier.java#L87) that shows that dependency verification is still run on deps that are not of type `ArtifactVerificationOperation.ArtifactKind.METADATA` if metadata verification is disabled

### DIFF
--- a/lib/modules/manager/gradle/artifacts.ts
+++ b/lib/modules/manager/gradle/artifacts.ts
@@ -96,38 +96,41 @@ async function buildUpdateVerificationMetadataCmd(
   if (!verificationMetadataFile) {
     return null;
   }
-  const hashTypes: string[] = [];
+
   const verificationMetadata = await readLocalFile(verificationMetadataFile);
-  if (
-    verificationMetadata?.includes('<verify-metadata>true</verify-metadata>')
-  ) {
-    logger.debug('Dependency verification enabled - generating checksums');
-    for (const hashType of ['sha256', 'sha512']) {
-      if (verificationMetadata?.includes(`<${hashType}`)) {
-        hashTypes.push(hashType);
-      }
-    }
-    if (!hashTypes.length) {
-      hashTypes.push('sha256');
-    }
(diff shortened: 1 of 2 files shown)


lib/modules/manager/gradle/artifacts.ts
```diff
@@ -96,38 +96,41 @@ async function buildUpdateVerificationMetadataCmd(
...
-    }
+  const verifiesChecksums = verificationMetadata?.includes(
+    '<verify-metadata>true</verify-metadata>',
+  );
+  const verifiesSignatures = verificationMetadata?.includes(
+    '<verify-signatures>true</verify-signatures>',
+  );
+
+  const hashTypes = ['sha256', 'sha512'].filter((type) =>
+    verificationMetadata?.includes(`<${type} `),
```
Churro, line 109: Note that this extra whitespace here is intentional

rarkins (approved):

0785386021:
How to TU SDK

PR #36849, opened 2025-07-05
refactor: Improve HTTP cache stat reporting

zharinov:
## Changes

- Keep `localHit`/`localMiss` counter names
- Rename `hit`/`miss` to `etagHit`/`etagMiss` to avoid confusion
  - `httpHit`/`httpMiss` is also possible and more precise since the caching also works with `Last-Modified` value, but `etagHit`/`etagMiss` looks better the same way we prefer to say "Linux" instead of "GNU/Linux"
- Rename the increment methods to exactly match counter name: `localHit()`, `etagMiss()`, etc
- Don't report zero values

### DIFF
--- a/lib/util/http/cache/package-http-cache-provider.ts
+++ b/lib/util/http/cache/package-http-cache-provider.ts
@@ -55,6 +55,7 @@ export class PackageHttpCacheProvider extends AbstractHttpCacheProvider {
   ): Promise<HttpResponse<T> | null> {
     const cached = await this.get(url);
     if (!cached) {
+      HttpCacheStats.localMiss(url);
       return null;
     }
 
@@ -66,11 +67,11 @@ export class PackageHttpCacheProvider extends AbstractHttpCacheProvider {
     const deadline = cachedAt.plus({ minutes: this.softTtlMinutes });
     const now = DateTime.now();
     if (now >= deadline) {
-      HttpCacheStats.incLocalMisses(url);
+      HttpCacheStats.localMiss(url);
       return null;
     }
 
-    HttpCacheStats.incLocalHits(url);
+    HttpCacheStats.localHit(url);
     return cached.httpResponse as HttpResponse<T>;
   }
 
--- a/lib/util/http/cache/package-http-cache-provider.ts
+++ b/lib/util/http/cache/package-http-cache-provider.ts
@@ -55,7 +55,6 @@ export class PackageHttpCacheProvider extends AbstractHttpCacheProvider {
   ): Promise<HttpResponse<T> | null> {
     const cached = await this.get(url);
     if (!cached) {
-      HttpCacheStats.localMiss(url);
       return null;
     }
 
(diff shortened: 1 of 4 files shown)


lib/util/http/cache/package-http-cache-provider.ts
```diff
@@ -55,6 +55,7 @@ export class PackageHttpCacheProvider extends AbstractHttpCacheProvider {
   ): Promise<HttpResponse<T> | null> {
     const cached = await this.get(url);
     if (!cached) {
+      HttpCacheStats.localMiss(url);
```
zharinov:
```suggestion
```

zharinov:
It should not be just a rename, for each URL it should reflect the full cache story:
- local cache miss, item populated
- local cache hit, item returned
- local cache stale, etag is the same, the TTL updated
- local cache stale, etag is new, the item fully updated

zharinov:
Okay, let's please merge this PR, then I'll figure out how to make reports more meaningful

rarkins:
So what does it mean now? e.g.

```
 "https://registry.npmjs.org": {
         "/@aws-sdk%2Fclient-codecommit": {"etagMiss": 1, "localHit": 1},
         "/@aws-sdk%2Fclient-ec2": {"etagMiss": 1, "localHit": 1},
         "/@aws-sdk%2Fclient-ecr": {"etagMiss": 1, "localHit": 1},
         "/@aws-sdk%2Fclient-eks": {"etagMiss": 1, "localHit": 1},
         "/@aws-sdk%2Fclient-rds": {"etagMiss": 1, "localHit": 1},
         "/@aws-sdk%2Fclient-s3": {"etagMiss": 1, "localHit": 1},
         "/@aws-sdk%2Fcredential-providers": {"etagMiss": 1, "localHit": 1},
         "/@baszalmstra%2Frattler": {"etagMiss": 1},
         "/@breejs%2Flater": {"etagMiss": 1},
```

Will it show etagMiss even if we had no cached etag? What does localHit then mean for the ones with localHit?

zharinov:
`localHit` is when the cached value is so fresh we even don't bother to go and check upstream. `localMiss` means the cache entry is found but needs revalidate. `etagHit` means 304, `etagMiss` means 200.

This all is confusing, and I started to think this should instead somehow reflect the full path of what happened instead of "hit this" and "miss that" which requires mental deciphering.

I think I'll better draw some explainer first, which we could use in docs, and then rewrite stats accordingly.

lib/util/stats.ts
```diff
@@ -379,10 +379,10 @@ export class HttpStats {
 }
 
 interface HttpCacheHostStatsData {
-  hit: number;
-  miss: number;
   localHit?: number;
   localMiss?: number;
+  etagHit?: number;
+  etagMiss?: number;
```
viceice, line 385: should be sorted alphabetically?

viceice (commented):
otherwise LGTM

rarkins (changes requested):
Needs deconflicting

PR #36874, opened 2025-07-07
fix(platform/azure): support getRawFile from tags if branch lookup fails

tonur:
## Changes

Added functionality to platform/azure git getRawFile azure-devops-node-api usage to try to find file with both git branch and git tag and not fail until both options are tried.
Added test that returns null first on git branch lookup as it fails to find file, to simulate git tag lookup second call.

## Context
- #36835

### DIFF
--- a/lib/modules/platform/azure/index.ts
+++ b/lib/modules/platform/azure/index.ts
@@ -127,7 +127,6 @@ export async function getRawFile(
   repoName?: string,
   branchOrTag?: string,
 ): Promise<string | null> {
-  try {
     const azureApiGit = await azureApi.gitApi();
 
     let repoId: string | undefined;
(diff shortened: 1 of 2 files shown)


tonur:
image
I have successfully verified that it finds the file as intended now, just with a branch lookup first and a tag lookup afterwards.
I made this decision due to that the variable is called "branchOrTag" and just looked up in that order.

lib/modules/platform/azure/index.ts
```diff
@@ -127,6 +129,8 @@ export async function getRawFile(
   repoName?: string,
   branchOrTag?: string,
 ): Promise<string | null> {
+  let item: GitItem | undefined;
+
```
RahulGautamSingh:
```suggestion
```

lib/modules/platform/azure/index.ts
```diff
@@ -144,23 +148,34 @@ export async function getRawFile(
...
-      undefined, // scopePath
-      undefined, // recursionLevel
-      undefined, // includeContentMetadata
-      undefined, // latestProcessedChange
-      undefined, // download
-      branchOrTag ? versionDescriptor : undefined, // versionDescriptor
-      true, // includeContent
-    );
-
+    // Try to get file from repo with branch, if not found, then try with tag #36835
```
RahulGautamSingh:
```suggestion
        let item: GitItem | undefined;
        const versionDescriptor: GitVersionDescriptor = {
      version: branchOrTag,
    } satisfies GitVersionDescriptor;
    // Try to get file from repo with branch, if not found, then try with tag #36835
```

lib/modules/platform/azure/index.ts
```diff
@@ -144,23 +148,34 @@ export async function getRawFile(
...
+    for (const versionType of [GitVersionType.Branch, GitVersionType.Tag]) {
+      const versionDescriptor: GitVersionDescriptor = {
+        version: branchOrTag,
+        versionType,
+      };
```
RahulGautamSingh:
```suggestion
      versionDescriptor.versionType = versionType;
```

lib/modules/platform/azure/index.ts
```diff
@@ -144,23 +148,34 @@ export async function getRawFile(
...
+        branchOrTag ? versionDescriptor : undefined, // versionDescriptor
+        true, // includeContent
+      );
+      if (item) {
+        break; // exit loop if item is found
+      } else {
+        logger.debug(
+          `File: ${fileName} not found in ${repoName} with ${versionType}: ${branchOrTag} - trying next versionType`,
+        );
+        continue;
```
RahulGautamSingh:
```suggestion
```
no remaining code to skip

lib/modules/platform/azure/index.ts
```diff
@@ -144,23 +148,34 @@ export async function getRawFile(
...
+        undefined, // latestProcessedChange
+        undefined, // download
+        branchOrTag ? versionDescriptor : undefined, // versionDescriptor
+        true, // includeContent
+      );
+      if (item) {
+        break; // exit loop if item is found
+      } else {
+        logger.debug(
+          `File: ${fileName} not found in ${repoName} with ${versionType}: ${branchOrTag} - trying next versionType`,
```
RahulGautamSingh:
```suggestion
          `File: ${fileName} not found in ${repoName} with ${versionType}: ${branchOrTag}`
```
On the second iteration, we don't need to say we are looking again. Better to remove

RahulGautamSingh:
Also, could you change the PR Title to: `fix(platform/azure): support getRawFile from tags if branch lookup fails`

tonur:
I will fix the errors in a moment, need to be on a better PC to do this with Prettier installed, I will request a new review afterwards.

EDIT: Sorry that it took longer than expected. Feel free to review when you get time, no need to rush, but I really appreciate the help, thank you so much!

lib/modules/platform/azure/index.ts
```diff
@@ -144,23 +146,34 @@ export async function getRawFile(
...
-      undefined, // download
-      branchOrTag ? versionDescriptor : undefined, // versionDescriptor
-      true, // includeContent
-    );
-
+    // Try to get file from repo with branch, if not found, then try with tag #36835
+    for (const versionType of [GitVersionType.Branch, GitVersionType.Tag]) {
+      versionDescriptor.versionType = versionType;
+
+      item = await azureApiGit.getItem(
```
rarkins, line 157: What value does this return if the branch is missing? And no throw?

tonur: If the branch is missing, it will default to the default branch of the repo, either "master", "main" or such.
image

Seen here:
image
image

rarkins: I mean does this function call throw an error if it can't find the result? This code assumes that it returns a falsey result.

tonur: Oh, sorry I misunderstood.
I tested it multiple times, as I thought it would throw an exception, but it doesn't.
You can see that I changed away from using a try catch to using just the return value (null if not found) in this commit: https://github.com/renovatebot/renovate/pull/36874/commits/5cd3ca7945c89be25cb9fa0059efeefcb400e9e9

To be clear, I have tested multiple times, and the function `azureApiGit.getItem` will return null if the file, the branch, the tag, or any other issue happens.

Sorry if I did not answer your question, please let me know if I still am misunderstanding something!

lib/modules/platform/azure/index.ts
```diff
@@ -144,23 +146,34 @@ export async function getRawFile(
...
-      undefined, // recursionLevel
-      undefined, // includeContentMetadata
-      undefined, // latestProcessedChange
-      undefined, // download
-      branchOrTag ? versionDescriptor : undefined, // versionDescriptor
-      true, // includeContent
-    );
-
+    // Try to get file from repo with branch, if not found, then try with tag #36835
+    for (const versionType of [GitVersionType.Branch, GitVersionType.Tag]) {
```
rarkins:
```suggestion
    for (const versionType of [GitVersionType.Tag, GitVersionType.Branch]) {
```

Let's reverse the order to try the tag first, as that was what was intended in the docs and how it works in other platforms

tonur: I reversed the order in this commit, and also updated the tests and comments:
https://github.com/renovatebot/renovate/pull/36874/commits/7adb1a908dbdd0d29d800a9135758fe3d98e58ea
Let me know if this is sufficient.

rarkins (approved):

RahulGautamSingh (approved):

viceice (approved):

PR #36876, opened 2025-07-07
docs: updated docs for secrets API token permissions

justo-mend:
## Changes

Updated docs for Mend-hosted Apps Credentials section to include details about the scopes and permissions for the API token when calling private repos.

## Context

User discovered that secrets APIs don't work on private repos or when API auth tokens don't have the correct resource or permissions/scope.


docs/usage/mend-hosted/credentials.md
```diff
@@ -111,7 +111,14 @@ Calls to the APIs require authentication with a personal access token from the G
 The token must be included in the request header as the `authorization` property.
 
 You can use a [classic token](https://github.com/settings/tokens) or a [fine-grained token](https://github.com/settings/personal-access-tokens).
-No additional scopes or permissions are required for the token.
+Mend recommends the use of fine-grained tokens for enhanced security.
+
+#### Scopes and permissions for tokens
+
+When calling APIs on _private_ repositories, the following scopes and permissions are required:
```
rarkins, line 119:
```suggestion
To be able to update secrets on both public and private repositories, the following scopes and permissions are required:
```

docs/usage/mend-hosted/credentials.md
```diff
@@ -111,7 +111,14 @@ Calls to the APIs require authentication with a personal access token from the G
...
+
+- Classic tokens require `repo` scope (Full control of private repositories)
+- Fine-grained tokens must be for the `resource` of the org that the API is being called on, and must have a minimum of `metadata:read-only` permission on the selected repositories.
```
rarkins:
```suggestion
- Fine-grained tokens must set the "Resource owner" as the org that the API is being called on, and GitHub's default (minimal) permissions.
```

I do not see metadata:read-only as an option when creating a fine-grained token for Orgs

rarkins: image

And which of these?

rarkins (changes requested):

docs/usage/mend-hosted/credentials.md
justo-mend: The access settings for Fine-grained tokens relate only to Repositories. There are no settings for Org access.

docs/usage/mend-hosted/credentials.md
justo-mend: No scopes or permissions are required for the token when updating secrets on public repos. The permissions are only required when updating secrets on private repos.

docs/usage/mend-hosted/credentials.md
justo-mend: Perhaps:
> Fine-grained tokens must set the "Resource owner" as the org that the API is being called on. When the API is for a private repo, "Repository access" must be set to "All repositories" or "Only select repositories" and include Repository permissions of "Metadata:Read-only".

docs/usage/mend-hosted/credentials.md
justo-mend: Suggesting:

> To update secrets on public repositories, no scopes or repository permissions are required.
> To update secrets on private repositories, the following scopes and permissions are required:

rarkins: I think that secrets in public repos will be rare, and using an API for updating secrets in public repos even rarer. We should aim to keep things simple for the 99% of people who are doing normal behavior - private repos.

I would prefer to not even mention public or private - it's complicating things unnecessarily. In other words, I suggest documentation should be like "The following works in all circumstances" and not playing trivia with one or more ways you can optimize it for edge cases.

The original wording was undesirable because it essentially says "To update secrets in private repos.." without clarifying what about public.

rarkins: In fact I think you should consider simplifying it further by describing the one, universal way to get all scenarios working with fine-grained tokens.

rarkins: And link to https://github.com/settings/personal-access-tokens/new

docs/usage/mend-hosted/credentials.md
rarkins: Let's back up a step. Make sure you describe, in order of appearance what _each_ field should be set to. Here's what I see when setting up a fine-grained token:

image
image

This is what I see when selecting "Resource owner":

image

Here's what I see when I expand Permissions (I don't see metadata as an option):

image
image
image

rarkins (changes requested):
Change it to be fine-grained tokens only, with every field described.

docs/usage/mend-hosted/credentials.md
rarkins: Note: changing from Public repositories to All repositories or Select repositories results in a "Repository permissions" section being added, and inside it contains the "Metadata" setting I wasn't previously able to find

docs/usage/mend-hosted/credentials.md
justo-mend: I have added clarification about requirements for calling APIs on public repos.
I don't feel right declaring that the only way to configure the token is by providing permissions and scopes when this is not true for public APIs.

docs/usage/mend-hosted/credentials.md
justo-mend: Updated docs to include screenshots of the fine-grained token experience.

rarkins (approved):

PR #36883, opened 2025-07-07
feat: Add Options on Http Call made on terraform-provider

ncharaf:
## Changes

Be able to specify Option on Http Call made on terraform-provider

## Context

In my company, I had to use a private Terraform provider that hosted in our internal Artifactory (Jfrog). I manage to make renovate fetch the file from Jfrog thanks to this [comment](https://github.com/renovatebot/renovate/issues/21486#issuecomment-2558094090) but I am getting an error related to the fact that the provider isn't available in public TF registry which make renovate think that it's a failure & send a failure  build notif to BitBucket.

Bitbucket build failure

The idea is that in my case to update renovate config in order to allow failure if we got a 404. I took [this](https://github.com/renovatebot/renovate/blob/ef2e45314fa6c2b4b58fceb1d8e93d447954f80e/lib/modules/datasource/docker/common.ts#L60) as an example.

Example of config
```json
{
  "hostRules": [
    {
      "abortOnError": false,
      "hostType": "terraform-provider",
      "matchHost": "registry.terraform.io"
    }
  ]
}
```

I had tested my modification locally & it's working. I don't get any more failure after that

image

As said in previous issue #21486 , I think will help a lot of people/company to use mirror but also to be able to fetch private provider

- [X] No unit tests but ran on a real repository, or

### DIFF
--- a/lib/modules/datasource/terraform-provider/index.ts
+++ b/lib/modules/datasource/terraform-provider/index.ts
@@ -1,8 +1,11 @@
 // TODO: types (#22198)
 import is from '@sindresorhus/is';
 import { logger } from '../../../logger';
+import type { HostRule } from '../../../types';
 import { ExternalHostError } from '../../../types/errors/external-host-error';
 import { cache } from '../../../util/cache/package/decorator';
+import * as hostRules from '../../../util/host-rules';
+import type { HttpOptions } from '../../../util/http/types';
 import * as p from '../../../util/promises';
 import { regEx } from '../../../util/regex';
 import { asTimestamp } from '../../../util/timestamp';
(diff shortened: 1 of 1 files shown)


ncharaf:
Hello @rarkins , I just created [this discussion](https://github.com/renovatebot/renovate/discussions/36891)

lib/modules/datasource/terraform-provider/index.ts
```diff
@@ -244,9 +247,19 @@ export class TerraformProviderDatasource extends TerraformDatasource {
       serviceDiscovery,
       repository,
     );
+
+    const opts: HostRule & HttpOptions = hostRules.find({
+      hostType: TerraformProviderDatasource.id,
+      url: registryURL,
+    });
+    const options: HttpOptions = {
+      throwHttpErrors: opts.abortOnError,
```
viceice, line 256: those options are already applied on  lower http layer, so your code doesn't change anything.

viceice (changes requested):

lib/modules/datasource/terraform-provider/index.ts
ncharaf: For the abortOnError yes but not sure for throwHttpErrors. If it's the case as well for throwHttpErrors can you point me that ?

Thanks

ncharaf: I found a workaround

PR #36900, opened 2025-07-08
docs(docker): more specific dockerhub registry auth

muellerst-hg:
## Changes

Adds hostRule to the dockerhub authentication example. Without hostrule, the config is applied to all registries, which probably fails if you have any dependency from a different registry e.g. AWS ECR.

## Context

Closes #36479

### DIFF
--- a/docs/usage/docker.md
+++ b/docs/usage/docker.md
@@ -199,7 +199,7 @@ We will explain how to authenticate for the most common registries.
 
 #### DockerHub
 
-Here is an example of configuring a default Docker username/password in `config.js`.
+Here is an example of configuring a Docker username/password for DockerHub in `config.js`.
 The Docker Hub password is stored in a process environment variable.
 
 ```js title="config.js"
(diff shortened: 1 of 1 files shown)


muellerst-hg:
I am not going to sign CLAs. Take it or leave it.

rarkins (dismissed):

docs/usage/docker.md
```diff
@@ -199,14 +199,15 @@ We will explain how to authenticate for the most common registries.
 
 #### DockerHub
```
rarkins:
```suggestion
#### Docker Hub
```

docs/usage/docker.md
```diff
@@ -199,14 +199,15 @@ We will explain how to authenticate for the most common registries.
...
 
-Here is an example of configuring a default Docker username/password in `config.js`.
+Here is an example of configuring a Docker username/password for DockerHub in `config.js`.
```
rarkins:
```suggestion
Here is an example of configuring a Docker username/password for Docker Hub in `config.js`.
```

docs/usage/docker.md
rarkins: Let's fix this while we're editing it now

rarkins (approved):

PR #36903, opened 2025-07-08
fix(cache): Make TTL overrides work for all namespaces

zharinov:
## Changes

- rename the current implementation of `set(...)` to `setWithRawTtl(...)`
- expose for external use another `set(...)` implementation which resolves TTL overrides and calls `setWithRawTtl(...)`
- cleanup external `getTtlOverrides(...)` usage


lib/util/http/cache/package-http-cache-provider.ts
```diff
@@ -1,7 +1,7 @@
 import is from '@sindresorhus/is';
 import { DateTime } from 'luxon';
 import { GlobalConfig } from '../../../config/global';
-import { get, set } from '../../cache/package'; // Import the package cache functions
+import * as packageCache from '../../cache/package'; // Import the package cache functions
```
viceice: I really prefer named imports

viceice (dismissed):

rarkins (dismissed):

rarkins (approved):

PR #36916, opened 2025-07-09
fix(http): improve URL validation logging for non-HTTP URLs

falconfs:
## Changes

Changed the log level from `logger.error` to `logger.warn` when encountering an invalid URL in the `resolveUrl` method in `lib/util/http/http.ts`. Also updated the log message to reflect that it's a warning instead of an error.

## Context

Related to discussion: https://github.com/renovatebot/renovate/discussions/36734

My first try is to change the log level from `error` to `warn` in the `resolveUrl` method in `lib/util/http/http.ts`, if the URL parseable but not a HTTP(S) URL.
Problem now is that Got cannot Handle URLs that are not HTTP(S) URLs, so it throws an error.

So I need a little bit help here to find a way to handle this case properly.

- [x] Code inspection only

### DIFF
--- a/lib/util/http/http.ts
+++ b/lib/util/http/http.ts
@@ -282,7 +282,12 @@ export abstract class HttpBase<
     }
 
     const parsedUrl = parseUrl(url);
-    if (!parsedUrl || !isHttpUrl(parsedUrl)) {
+    if (parsedUrl && !isHttpUrl(parsedUrl)) {
+      logger.warn(
+        { url: requestUrl, baseUrl, resolvedUrl: url },
+        'Request Warning: is not an HTTP URL',
+      );
+    } else if (!parsedUrl) {
       logger.error(
         { url: requestUrl, baseUrl, resolvedUrl: url },
         'Request Error: cannot parse url',
--- a/lib/util/http/http.ts
+++ b/lib/util/http/http.ts
@@ -282,11 +282,7 @@ export abstract class HttpBase<
     }
 
     const parsedUrl = parseUrl(url);
-    if (parsedUrl && !isHttpUrl(parsedUrl)) {
-      logger.warn(
-        { url: requestUrl, baseUrl, resolvedUrl: url },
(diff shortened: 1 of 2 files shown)


lib/util/http/http.ts
```diff
@@ -282,7 +282,12 @@ export abstract class HttpBase<
     }
 
     const parsedUrl = parseUrl(url);
-    if (!parsedUrl || !isHttpUrl(parsedUrl)) {
+    if (parsedUrl && !isHttpUrl(parsedUrl)) {
+      logger.warn(
+        { url: requestUrl, baseUrl, resolvedUrl: url },
+        'Request Warning: is not an HTTP URL',
+      );
```
rarkins: So this will still return the parsedUrl instead of throwing an Error? If so then what happens next?

falconfs: The background is that an error code 1 causes alerts for us because Renovate has crashed.
I think it would be better if this type of error should be treated as a soft error, similar to an error in the Package Lookup Failure that is displayed in the dependency dashboard, and not as a serious error that causes Renovate to terminate with code 1. 
Now I'm trying to solve this, but I need some help.

rarkins: My concern is that this code change leads to a different type of error. You are not just lowering the log level, you are changing the execution logic

falconfs (commented):
fix(datasource): validate helmRepository URL format before processing

falconfs:
Okay, I'll change my part a little bit: If the Helm repository URL is not an HTTP URL, a warning message is issued and Renovate does nothing further for this “release.”

lib/util/http/http.ts
```diff
@@ -283,6 +283,7 @@ export abstract class HttpBase<
 
     const parsedUrl = parseUrl(url);
     if (!parsedUrl || !isHttpUrl(parsedUrl)) {
+    } else if (!parsedUrl) {
```
RahulGautamSingh, line 286: This doesnt look right

PR #36929, opened 2025-07-09
fix(lookup): add generation of pseudo-version for commit digests on Go updates

chrisnavar:
## Changes

This change attempts to fix the bug explained on [discussion 36798](https://github.com/renovatebot/renovate/discussions/36798), related with the generation of a pseudo-version for the latest digest, as it's not correctly generated at the moment, only setting the newDigest but not the newValue field correctly for the version.


rarkins (changes requested):
Sorry, but we want to avoid having manager-specific code in lib/workers - it's meant to be a generic layer

viceice (commented):
this should probably be moved to `getNewValue` of go versioning? 🤔

chrisnavar:
> this should probably be moved to `getNewValue` of go versioning? 🤔

Hi Michael, sorry for my delay in the response as I haven't had time to properly look into this until now. I've checked but looking at the code I don't know if we could move it to the getNewValue of go versioning. I've thought of adding a helper function in the gomod manager that is called in lib/workers, leaving all the Go logic inside gomod, but it would still require a small manager check in lib/workers. 
If not, I think it could work too adding a new go versioning module that would completely remove the manager-specific code in lib/workers, but it would require extending the `NewValueConfig` interface as it doesn't include `digest/timestamp`, and change the lineParser from `loose` to `go`, which would then be taken by the new versioning module. 
Let me know what you think, or if you think there's another better alternative to solve this.

viceice (commented):
i think this can be closed in favor of ?
- #41588

PR #36960, opened 2025-07-11
fix(gitlab): consider the merge train setting in getBranchForceRebase

cjpearson:
## Changes

This changes the logic in the `getBranchForceRebase` function for GitLab to consider the merge train setting, as enabling merge trains means that merge requests do not need to be up to date in order to merge. This should match the behavior in the documentation: "It will use behind-base-branch if configured to automerge or repository has been set to require PRs to be up to date. Otherwise, conflicted will be used instead"

As part of this change, `mergeTrainsEnabled` is added to the config object.

## Context

- https://github.com/renovatebot/renovate/discussions/36774


lib/modules/platform/gitlab/types.ts
```diff
@@ -68,6 +68,7 @@ export interface RepoResponse {
   repository_access_level: 'disabled' | 'private' | 'enabled';
   merge_requests_access_level: 'disabled' | 'private' | 'enabled';
   merge_method: MergeMethod;
+  merge_trains_enabled?: boolean;
```
viceice, line 75:
```suggestion
  /**
   * only available with paid plans
   * https://docs.gitlab.com/ci/pipelines/merge_trains
   */
  merge_trains_enabled?: boolean;
```

maybe needs prettier fix

viceice (approved):

PR #36995, opened 2025-07-15
fix(local-scm): Increase default execSync buffer

canon-cmre-liam-lamb:
## Changes

This changes explicitly sets the max buffer to 10MB size provided to execSync when reading the files list in local mode. 
Using a larger default limits the risk of failing to load the files in larger projects when the output size of the child process exceeds the default buffer size of 1MB. 

## Context
- See discussion #34715 


canon-cmre-liam-lamb:
Is it expected for some tests to take as long as 8/16 did? I'm assuming this was manually cancelled?

rarkins:
There's auto cancellation if tests fail

canon-cmre-liam-lamb:
It looks like the tests passed, but the step failed as it timed out after 10 minutes (from what I can see in the actions logs)

BoomchainLabs (approved):
Looks good

lib/modules/platform/local/scm.ts
```diff
@@ -33,7 +33,8 @@ export class LocalFs implements PlatformScm {
   async getFileList(): Promise<string[]> {
     try {
       // fetch file list using git
-      const stdout = execSync('git ls-files', { encoding: 'utf-8' });
+      const maxBuffer = 1024 * 1024 * 10; // 10 MB
```
astellingwerf: Can I suggest?:
```suggestion
      const maxBuffer = 10 * 1024 * 1024; // 10 MiB in bytes
```

rarkins (approved):

PR #37004, opened 2025-07-15
fix(github): Fix PR cache fetch loop

zharinov:


lib/modules/platform/github/api-cache.ts
```diff
@@ -32,39 +31,35 @@ export class ApiCache<T extends ApiPageItem> {
...
 
-      const itemNewTime = DateTime.fromISO(newItem.updated_at);
-      const itemOldTime = oldItem?.updated_at
-        ? DateTime.fromISO(oldItem.updated_at)
-        : null;
-
-      if (!dequal(oldItem, newItem)) {
-        logger.trace(`PR cache: updating item ${number}`);
-        items[number] = newItem;
+      if (dequal(oldItem, newItem)) {
```
viceice, line 45: why we do this expensive deep equal when we've an updated at for comparison 🤔

viceice (commented):
failing tests

lib/modules/platform/github/api-cache.ts
viceice: ☝️⁉️

jamietanna:
As part of https://github.com/renovatebot/renovate/issues/40426, but replaced by https://github.com/renovatebot/renovate/pull/41462

PR #37052, opened 2025-07-17
fix(gitlab): Prevent early pagination termination in PR cache sync

mikethecalamity:
Fixes a pagination bug in GitLab PR cache that caused Renovate to attempt reusing merged PRs when searching for open ones. The cache sync would stop fetching pages prematurely when any PR matched cached data, leaving older PRs with stale state information.

- https://github.com/renovatebot/renovate/discussions/37010

## Changes

* Remove early pagination termination logic in `GitlabPrCache.reconcile()` method
* Continue fetching all pages until GitLab returns empty results instead of stopping when unchanged PRs are found
* Add check to properly terminate pagination only when no more results are available (`rawItems.length === 0`)

## Context

The `reconcile()` method incorrectly assumed that if any PR in the current page matched cached data (`dequal(oldItem, newItem) === true`), then all older PRs must also be unchanged, so it would stop pagination with `needNextPage = false`. This flawed logic caused issues when:

1. Recent PRs hadn't changed (triggering early termination)
2. Older PRs had state changes (e.g., merged) but weren't reached due to stopped pagination
3. Renovate would find these stale "open" PRs in cache when searching for reusable branches

The bug became apparent when PR volume increased enough to push older PRs to later pages that weren't being fetched due to the premature termination.

### DIFF
--- a/lib/modules/platform/gitlab/pr-cache.ts
+++ b/lib/modules/platform/gitlab/pr-cache.ts
@@ -105,7 +105,9 @@ export class GitlabPrCache {
     const { items: oldItems } = this.cache;
     let { updated_at } = this.cache;
 
-    let needNextPage = true;
+    if (rawItems.length === 0) {
+      return false;
+    }
 
     for (const rawItem of rawItems) {
       const id = rawItem.iid;
@@ -116,7 +118,6 @@ export class GitlabPrCache {
       const itemNewTime = DateTime.fromISO(rawItem.updated_at);
 
       if (dequal(oldItem, newItem)) {
-        needNextPage = false;
         continue;
       }
 
@@ -129,7 +130,8 @@ export class GitlabPrCache {
     }
 
     this.cache.updated_at = updated_at;
-    return needNextPage;
+    
+    return true;
   }
 
   private async sync(http: GitlabHttp): Promise<GitlabPrCache> {


mikethecalamity:
@viceice  @rarkins

viceice (changes requested):
this makes PR cache useless because we always fetch all PRs in each run.

the root cause seems to be a wrong order of returned PRs. don't they get a new updated at when merged?

viceice:
@RahulGautamSingh it seems we've same issue like azure here. can you check if there's an order by update at param?

https://github.com/renovatebot/renovate/blob/879f2ae7d3ef31fd71643ff3ec31a81efad8eb9a/lib/modules/platform/gitlab/pr-cache.ts#L138

viceice:
`order_by=updated_at`

https://docs.gitlab.com/api/merge_requests/

viceice:
we could also leverage the `updated_after` to only get updated PRs.

`order=desc` to  last updated first 🤔

RahulGautamSingh (commented):
As mentioned in previous comments. Please, modify the logic to use `update_at` option for fetching PRs.

zharinov:
`updated_after` parameter is godsent, we have to use this and simplify the overall fetch logic

zharinov:
- Here we go: https://github.com/renovatebot/renovate/pull/37076

PR #37053, opened 2025-07-17
fix: Include architecture when resolving latest JVM release version

markelliot:
## Changes

When resolving JVM release version pass the current system architecture.

## Context

On July 17, 2025, my Renovate jobs started to fail because the Adoptium JDK was releasing a new version, and, when querying release_versions, the new version shows up even though specific architectures have lower availabile versions. By passing the current system architecture, we ensure that the version lookup is bounded by what's actually available.

See also https://github.com/renovatebot/renovate/discussions/37046

### DIFF
--- a/lib/modules/datasource/java-version/index.ts
+++ b/lib/modules/datasource/java-version/index.ts
@@ -1,3 +1,4 @@
+import { arch } from 'node:os';
 import { logger } from '../../../logger';
 import { ExternalHostError } from '../../../types/errors/external-host-error';
 import { cache } from '../../../util/cache/package/decorator';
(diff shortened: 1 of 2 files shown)


lib/modules/datasource/java-version/index.ts
```diff
@@ -96,4 +100,16 @@ export class JavaVersionDatasource extends Datasource {
...
+    switch (arch()) {
+      case 'arm64':
+        return 'aarch64';
+      case 'x64':
+        return 'x64';
+      default:
+        // should never happen
+        throw new Error('Unsupported architecture: ' + arch());
+    }
+  }
```
markelliot, line 114: similar to https://github.com/containerbase/base/blob/6267a2079e21c1b603becf11346b111c5e128bcf/src/cli/services/env.service.ts#L24-L37

viceice (changes requested):
I'll work on a different solution today. this solution will break existing users

markelliot:
Ok, thanks

Issue #33874, opened 2025-01-27
Support `uv` workspaces


rarkins:
### Discussed in https://github.com/renovatebot/renovate/discussions/33275

<div type='discussions-op-text'>

<sup>Originally posted by **kkom** December 25, 2024</sup>
### How are you running Renovate?

A Mend.io-hosted app

### Please tell us more about your question or problem

When using the `uv` [workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/) feature, Renovate-generated PRs do not update the workspace's `uv.lock` file.

Here's the layout of a typical `uv` project using workspaces:

```
albatross
├── packages
│   ├── bird-feeder
│   │   ├── pyproject.toml
│   │   └── src
│   │       └── bird_feeder
│   │           ├── __init__.py
│   │           └── foo.py
│   └── seeds
│       ├── pyproject.toml
│       └── src
│           └── seeds
│               ├── __init__.py
│               └── bar.py
├── pyproject.toml
├── README.md
├── uv.lock
└── src
    └── albatross
        └── main.py
```

Note how there is a single `uv.lock` file, but multiple `pyproject.toml`. Dependencies of member packages in a workspace are locked in the top-level `uv.lock` file, but Renovate fails to update it when bumping versions.

Perhaps Renovate isn't workspace-aware and assumes that each `pyproject.toml` should have a corresponding `uv.lock` file in the same directory?

AFAIK, the right solution is to run `uv lock` in the workspace root.

### Logs (if relevant)

<details><summary>Logs</summary>

```

Replace this text with your logs, between the starting and ending triple backticks

```

</details>
</div>

Reproduction forked to https://github.com/renovate-reproductions/33275

latk:
I'm struggling with a similar scenario using the Poetry manager, where I have path-dependencies from the top-level `pyproject.toml` file to each workspace member package, and only a top-level lockfile exists. Poetry and uv are very similar here as far as Renovate is concerned, so I'm referring to both as `{manager}` in the following.

As I'm working with a self-hosted system, I unsuccessfully tried a couple of workarounds:

I can create symlinks from `package/*/{manager}.lock` to the top-level `{manager}.lock` file, in which case Renovate is able to at least resolve the currently locked versions correctly, but these symlinked lockfiles cannot be updated as they contain now-invalid relative paths.

[...]

PR #37064, opened 2025-07-18
feat(pep621): Add support for uv workspaces

akern40:
## Changes

Use `findLocalSiblingOrParent` for lockfile resolution instead of `getSiblingFileName`.

## Context

Supports #33874, for uv. I'm not familiar enough with Poetry to provide the same kind of PR.

In uv, [workspace repos](https://docs.astral.sh/uv/concepts/projects/workspaces/) consist of a "root package" along with a set of other packages that are all managed with a single environment, and therefore a single lockfile. Currently, renovate locates the `uv.lock` file for a given project by concatenating the directory of the `pyproject.toml` file with the filename `uv.lock`. This is accurate for non-workspaces.

This PR takes a relatively simple approach, using an existing utility function to search for a `uv.lock` file up the file tree, stopping at the first one found. Missing lockfiles are treated identically to missing lockfiles in the "sibling-only" / single-project case.

For re-locking, no change is needed; uv itself will handle the lockfile resolution via `uv lock`, even if it is run in a subdirectory.

- [x] Both unit tests + ran on a real repository (see https://github.com/akern40/renovate-uv-workspace-test/pull/4)

### DIFF
--- a/lib/modules/manager/pep621/processors/uv.ts
+++ b/lib/modules/manager/pep621/processors/uv.ts
@@ -5,7 +5,7 @@ import { logger } from '../../../../logger';
 import type { HostRule } from '../../../../types';
 import { exec } from '../../../../util/exec';
 import type { ExecOptions, ToolConstraint } from '../../../../util/exec/types';
-import { getSiblingFileName, readLocalFile } from '../../../../util/fs';
+import { findLocalSiblingOrParent, readLocalFile } from '../../../../util/fs';
 import { getGitEnvironmentVariables } from '../../../../util/git/auth';
 import { find } from '../../../../util/host-rules';
 import { Result } from '../../../../util/result';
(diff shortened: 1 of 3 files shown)


lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -122,21 +122,25 @@ export class UvProcessor implements PyProjectProcessor {
...
-      if (err) {
-        logger.debug({ packageFile, err }, `Error parsing uv lock file`);
-      } else {
-        for (const dep of deps) {
-          const packageName = dep.packageName;
-          if (packageName && packageName in lockFileMapping) {
-            dep.lockedVersion = lockFileMapping[packageName];
+    const lockFileName = await findLocalSiblingOrParent(packageFile, 'uv.lock');
+    if (lockFileName === null) {
+      logger.debug({ packageFile }, `No uv lock file found`);
```
rarkins, line 127: Should this return, or is it valid for it to continue without a lock file name?

akern40: I'm basing this on the existing behavior for when the file is found but its contents can't be read - it logs the error and returns the dependency list without adding any locked versions.

secustor (commented):
Could workspaces define different versions of UV and have own lockfiles?

akern40:
> Could workspaces define different versions of UV and have own lockfiles? 

I'm not positive what you're asking, could you give an example of your concern? A single workspace cannot have multiple lockfiles, and the lockfiles of multiple workspaces (which may exist in the same repo) do not interact with each other. uv workspaces also cannot be nested, so you will never have a lockfile within the tree of a workspace, only adjacent to it. 

The version of uv you use is also not configurable from within a Python project or workspace.

The lockfile schema is versioned, though. Backwards compatibility is not guaranteed, only compatibility within a single minor version of uv (so, compatible across patch versions). It's possible that the uv binary used by renovate is different from the one used by the original repo owner, and that could cause problems. However, that issue is extant in the current implementation and not affected by this PR, I don't think.

secustor:
> I'm not positive what you're asking, could you give an example of your concern? A single workspace cannot have multiple lockfiles, and the lockfiles of multiple workspaces (which may exist in the same repo) do not interact with each other. uv workspaces also cannot be nested, so you will never have a lockfile within the tree of a workspace, only adjacent to it.
> 
> The version of uv you use is also not configurable from within a Python project or workspace.
> 
> The lockfile schema is versioned, though. Backwards compatibility is not guaranteed, only compatibility within a single minor version of uv (so, compatible across patch versions). It's possible that the uv binary used by renovate is different from the one used by the original repo owner, and that could cause problems. However, that issue is extant in the current implementation and not affected by this PR, I don't think.

K thx, that answers partly my question. 

We are using the [required version of UV](https://github.com/renovatebot/renovate/blob/aefbbc17222661b0391325ce6e9838a0bce8f8a9/lib/modules/manager/pep621/processors/uv.ts#L173) for picking the UV version to run.   
So my question has been if UV ignores that value from workspace `pyproject.toml` files or how it resolves conflicts. 
As you described, if we are not following the version of UV it would change the lock file in an unexpected manor.

akern40:
Ah got it! Sorry, I had no idea you could require a uv version in your configuration, that's on me. Should have fact-checked myself 😅

But yes, if there's a version mismatch it might alter the lockfile schema.

viceice (approved):

secustor (approved):

PR #37066, opened 2025-07-18
feat: make dashboard branch/manager output in stable order

trim21:
## Changes

sort deps before we render dashboard, so we will have a stable dashboard output.

### DIFF
--- a/lib/workers/repository/package-files.ts
+++ b/lib/workers/repository/package-files.ts
@@ -85,7 +85,9 @@ export class PackageFiles {
     const pad = data.size > 1; // padding condition for a multi base branch repo
     let deps = '';
 
-    for (const [branch, packageFiles] of data) {
+    for (const [branch, packageFiles] of Array.from(data).toSorted(([a], [b]) =>
+      a.localeCompare(b),
+    )) {
       deps += pad
         ? `<details><summary>Branch ${branch}</summary>\n<blockquote>\n\n`
         : '';
@@ -95,12 +97,14 @@ export class PackageFiles {
         continue;
       }
 
-      const managers = Object.keys(packageFiles);
+      const managers = Array.from(Object.keys(packageFiles));
       if (managers.length === 0) {
         deps += none;
         deps += pad ? '</blockquote>\n</details>\n\n' : '';
         continue;
       }
+
+      managers.sort();
 
       for (const manager of managers) {
         deps += `<details><summary>${manager}</summary>\n<blockquote>\n\n`;
--- a/lib/workers/repository/package-files.ts
+++ b/lib/workers/repository/package-files.ts
@@ -108,7 +108,9 @@ export class PackageFiles {
 
       for (const manager of managers) {
         deps += `<details><summary>${manager}</summary>\n<blockquote>\n\n`;
-        for (const packageFile of packageFiles[manager]) {
+        for (const packageFile of Array.from(packageFiles[manager]).toSorted(
+          (a, b) => a.packageFile.localeCompare(b.packageFile),
+        )) {
           deps += `<details><summary>${packageFile.packageFile}</summary>\n\n`;
           for (const dep of packageFile.deps) {
             const ver = dep.currentValue;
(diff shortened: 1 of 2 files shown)


rarkins (commented):
Why? In which scenarios is there "instability"?

trim21:
> Why? In which scenarios is there "instability"?

if you click "Check this box to trigger a request for Renovate to run again on this repository" you may find the order of managers change on every times it generate new dashboard.

rarkins:
What is the ordering which Renovate uses today? It's not random

trim21:
We use whatever `Object.keys(packageFiles)` gives us, which has order in only certain cases.

trim21:
correction: I observe this when I newlly merge a renovatebot PR or some deps changed,  click "Check this box to trigger a request for Renovate to run again on this repository" without deps change doesn't always trigger this.

rarkins:
Do you have a public example of this order changing? e.g. a public github.com issue where the issue edit history shows this?

trim21:
yeah

https://github.com/trim21/qbittorrent-feedstock/issues/4

image


image

rarkins:
That's strange. I expected them to be in alphabetical order of manager name. I wonder if the fact that jsonata is here has an impact (it's technically the "custom" manager)

trim21:
I would suggest let's do not rely on the order of `Object.keys` and manually sort to make sure of it.

lib/workers/repository/package-files.ts
```diff
@@ -85,7 +85,9 @@ export class PackageFiles {
     const pad = data.size > 1; // padding condition for a multi base branch repo
     let deps = '';
 
-    for (const [branch, packageFiles] of data) {
+    for (const [branch, packageFiles] of Array.from(data).toSorted(([a], [b]) =>
```
viceice:
```suggestion
    for (const [branch, packageFiles] of Array.from(data).sort(([a], [b]) =>
```
no need to create a new array (perf)

lib/workers/repository/package-files.ts
```diff
@@ -85,7 +85,9 @@ export class PackageFiles {
...
+      a.localeCompare(b),
```
viceice: add the numeric parameter (third argument), so versions are correctly sorted

lib/workers/repository/package-files.ts
```diff
@@ -95,16 +97,20 @@ export class PackageFiles {
         continue;
       }
 
-      const managers = Object.keys(packageFiles);
+      const managers = Array.from(Object.keys(packageFiles));
```
viceice:
```suggestion
      const managers = Object.keys(packageFiles).sort();
```
no need to clone, it's s new array

lib/workers/repository/package-files.ts
```diff
@@ -95,16 +97,20 @@ export class PackageFiles {
...
       if (managers.length === 0) {
         deps += none;
         deps += pad ? '</blockquote>\n</details>\n\n' : '';
         continue;
       }
 
+      managers.sort();
+
```
viceice:
```suggestion
```

lib/workers/repository/package-files.ts
```diff
@@ -95,16 +97,20 @@ export class PackageFiles {
...
       for (const manager of managers) {
         deps += `<details><summary>${manager}</summary>\n<blockquote>\n\n`;
-        for (const packageFile of packageFiles[manager]) {
+        for (const packageFile of Array.from(packageFiles[manager]).toSorted(
```
viceice:
```suggestion
        for (const packageFile of Array.from(packageFiles[manager]).sort(
```

lib/workers/repository/package-files.ts
trim21: sorry I didn't get this

trim21: OK I know. But I don't think it will help sorting versions, it only help sorting string contains numbers.

viceice (approved):

rarkins (approved):

PR #37073, opened 2025-07-19
feat(matchManagers): support `regexOrGlob` matching

RahulGautamSingh:
## Changes
Remove validation logic which matches exact manager name and replace it with regexOrGlob based validation (i.e `patternMatch:true` in the `option/index.ts`)

## Context
`matchManagers` pkg rule matches supports regex+glob matching but we still have old validation logic for the `matchManagers` in the codebase which errors when using regex/glob


viceice (approved):

rarkins:
Does this change the json schema? Is all validation logic now removed?

RahulGautamSingh:
> Is all validation logic now removed?

No, it use [existing validation](https://github.com/renovatebot/renovate/blob/main/lib/config/validation-helpers/regex-glob-matchers.ts) for `regedOrGlob` options. 

 And, no the json schema remains same(an array of strings).

rarkins:
I meant validating it against a list of valid managers. Eg fail "foo" or "!foo" when it's an exact match. I assume it's a no now? I think it's ok

RahulGautamSingh:
Yes, we lose that ability.

PR #37076, opened 2025-07-19
feat(gitlab): Improved PR cache

zharinov:
## Changes

- First fetch: use simple fetch with pagination
- Subsequent fetches: set `updated_after` and enable ETag-based caching


lib/util/http/gitlab.ts
```diff
@@ -39,7 +39,6 @@ export class GitlabHttp extends HttpBase<GitlabHttpOptions> {
 
     const result = await super.requestJsonUnsafe<T>(method, opts);
     if (opts.httpOptions.paginate && is.array(result.body)) {
-      delete opts.httpOptions.cacheProvider;
```
zharinov, line 42: I tested, it seems to work well with pagination

RahulGautamSingh (approved):

lib/modules/platform/gitlab/pr-cache.ts
```diff
@@ -37,6 +36,13 @@ export class GitlabPrCache {
         updated_at: null,
         author,
       };
+    } else if (pullRequestCache.updated_at?.match(/\.\d\d\dZ$/)) {
```
RahulGautamSingh:
```suggestion
    } else if (pullRequestCache.updated_at?.match(regEx(/\.\d\d\dZ$/))) {
```

And it's import.

lib/modules/platform/gitlab/pr-cache.ts
```diff
@@ -101,62 +107,41 @@ export class GitlabPrCache {
...
+    if (items.length) {
+      for (const item of items) {
+        const id = item.iid;
+        this.cache.items[id] = prInfo(item);
       }
 
-      const uri = parseUrl(parseLinkHeader(res.headers.link)?.next?.url);
-      query = uri ? uri.search : null;
+      const [{ updated_at }] = items;
+      this.cache.updated_at = updated_at.replace(/\.\d\d\dZ$/, 'Z');
```
RahulGautamSingh:
```suggestion
      this.cache.updated_at = updated_at.replace(regEx(/\.\d\d\dZ$/), 'Z');
```

RahulGautamSingh (changes requested):

rarkins:
@zharinov how will this handle existing GitLab MR caches? (Which seem faulty)

E.g. will it cause a one-time full refetch, or will it somehow "self-heal" the cache without fetching all?

We need to make sure that users don't need to _manually_ clear faulty caches after this

zharinov:
We infer the older version of cache from having milliseconds suffix of the timestamp. The new cache strips them away before saving because it'll need to query PRs with timestamp without milliseconds. So, it's supposed to be one-time reset and re-fetch.

lib/modules/platform/gitlab/pr-cache.ts
viceice: ☝️

viceice (commented):
otherwise lgtm

viceice (dismissed):

rarkins (changes requested):
Needs test fixes

rarkins (approved):

PR #37077, opened 2025-07-20
fix(log): sanitize rawExec errors

Gabriel-Ladzaretti:
## Changes
Raw exec errors `ExecError` currently include the env vars that are sent to the child process, which are printed as is in case of an error.
In this pr, we sanitize these errors by stripping the env values and printing only the keys.


lib/logger/utils.ts
```diff
@@ -132,6 +133,12 @@ export default function prepareError(err: Error): Record<string, unknown> {
     response.errors = err.errors.map((error) => prepareError(error));
   }
 
+  // handle rawExec error
+  if (err instanceof ExecError && is.nonEmptyObject(err.options?.env)) {
+    const env = Object.keys(err.options.env);
+    response.options = { ...err.options, env };
```
viceice, line 139: why not calling sanitizer on values only 🤔

viceice (commented):
otherwise LGTM

lib/logger/utils.ts
Gabriel-Ladzaretti: i dont think the value sanitize is generic enough, we dont really know in advance what env is being passed nor the hardcoded keys seems env specific. it looks like the sanitize values func is more renovate config oriented ?

wdyt ?

Gabriel-Ladzaretti: for example :man_shrugging: 
image

Gabriel-Ladzaretti: in the long living containers solutions we inherit the entire env from the parent, hence renovate is not able to santize renovate unrelated secrets.

rarkins: I think let's just keep with the env keys _until_ we find some case where knowing the values are essential.

rarkins (approved):

viceice (approved):

PR #37089, opened 2025-07-21
chore: add Repository Result metadata to "Repository Finished" log

jamietanna:
## Changes

Add log metadata to the log line (so it becomes part of i.e. the JSON object that is then logged)

## Context

The `Repository result` log message is useful for understanding
at-a-glance why a Renovate run may have not worked.

Adding the metadata that currently exists in the log line as Meta keys
would allow for logging tools to be able to more meaningfully
interpolate the data, without requiring parsing of the underlying
string.

Co-authored-by: Jamie Tanna <jamie.tanna@elastic.co>

### DIFF
--- a/lib/workers/repository/index.ts
+++ b/lib/workers/repository/index.ts
@@ -156,7 +156,18 @@ export async function renovateRepository(
   ObsoleteCacheHitLogger.report();
   AbandonedPackageStats.report();
   const cloned = isCloned();
-  logger.info({ cloned, durationMs: splits.total }, 'Repository finished');
+  /* v8 ignore next 11 -- coverage not required of these `undefined` checks, as we're happy receiving an `undefined` in the logs */
+  logger.info(
+    {
+      cloned,
+      durationMs: splits.total,
+      result: repoResult?.res,
+      status: repoResult?.status,
+      enabled: repoResult?.enabled,
+      onboarded: repoResult?.onboarded,
+    },
+    'Repository finished',
+  );
   resetRepositoryLogLevelRemaps();
   return repoResult;
 }


lib/workers/repository/result.ts
```diff
@@ -86,6 +86,12 @@ export function processResult(
...
   }
   logger.debug(
+    {
+      result: res,
+      status,
+      enabled: enabled!,
+      onboarded: onboarded!,
+    },
     // TODO: types (#22198)
     `Repository result: ${res}, status: ${status}, enabled: ${enabled!}, onboarded: ${onboarded!}`,
```
RahulGautamSingh, line 90:
```suggestion
    'Repository result',
```

RahulGautamSingh (changes requested):

lib/workers/repository/result.ts
jamietanna: My intent here was to not change this, which some folks may be relying upon right now

rarkins: @nabeelsaabna @Gabriel-Ladzaretti are you parsing this log message anywhere?

rarkins: They definitely are

rarkins (commented):
Is this better done as part of the "Repository finished" INFO log message?

jamietanna:
Happy to move it over, that makes sense - wasn't sure how much movement we wanted here

rarkins:
@Gabriel-Ladzaretti wdyt?

lib/workers/repository/result.ts
nabeelsaabna: Yes! we are parsing this in a couple of projects

viceice (dismissed):

jamietanna:
To confirm, if we were looking to move it, would we want to duplicate [these lines](https://github.com/jamietanna/renovate/blob/c705b577c16b87b61bc53e2f8adba56760115c36/lib/workers/repository/result.ts#L65-L93) to also apply [here](https://github.com/jamietanna/renovate/blob/c705b577c16b87b61bc53e2f8adba56760115c36/lib/workers/repository/index.ts#L159)?

(I can't remember if we generally follow a "rule of 3" to only create shared functions when we're at the 3rd repetition, or if we follow "Do Not Repeat Yourself (DRY)")

jamietanna:
(Ignore me, I misread `RepoResult`'s contents)

lib/workers/repository/index.ts
```diff
@@ -156,7 +156,18 @@ export async function renovateRepository(
   ObsoleteCacheHitLogger.report();
   AbandonedPackageStats.report();
   const cloned = isCloned();
-  logger.info({ cloned, durationMs: splits.total }, 'Repository finished');
+  // istanbul ignore next
```
RahulGautamSingh: Use v8 ignore statements. Sometimes the istanbul ones don't work

RahulGautamSingh (approved):

lib/workers/repository/index.ts
jamietanna: Thanks, wasn't aware there was a difference here!

lib/workers/repository/index.ts
```diff
@@ -156,7 +156,18 @@ export async function renovateRepository(
   ObsoleteCacheHitLogger.report();
   AbandonedPackageStats.report();
   const cloned = isCloned();
-  logger.info({ cloned, durationMs: splits.total }, 'Repository finished');
+  // v8 ignore next -- coverage not required of `undefined` checks
```
jamietanna: are we happy with this message?

lib/workers/repository/index.ts
```diff
@@ -156,7 +156,18 @@ export async function renovateRepository(
   ObsoleteCacheHitLogger.report();
   AbandonedPackageStats.report();
   const cloned = isCloned();
-  logger.info({ cloned, durationMs: splits.total }, 'Repository finished');
+  // v8 ignore next -- coverage not required of these `undefined` checks, as we're happy receiving an `undefined` in the logs
```
jamietanna: are we happy with this message?

jamietanna: I'm not sure this is quite true here - as the `undefined` isn't really the thing here we're worried about - it's that it's not worth the time to add the tests (IMO, and based on where this comes from)

Maybe:

```suggestion
  // v8 ignore next -- test coverage of these checks are unnecessary, as ...
```

RahulGautamSingh: Try this: `/* v8 ignore next */`

jamietanna:
OK, CLA Checker is failing as I signed it when I was at my previous company, but as I've deregistered the email address with my account, it's now failing as there's no GitHub user that's associated with the email

When I come back to this PR, I'll rebase it and force push (even though we prefer not to) as I'll need to rewrite the committee and author on the commit(s) that are on my account, and add a `Co-authored-by` Git trailer

I'll mark this as Draft in the meantime

viceice (approved):

PR #37095, opened 2025-07-21
feat(presets/internal): Add pantheon-systems/drupal-integrations to Drupal core group

davereid:
## Changes

This adds the https://github.com/pantheon-systems/drupal-integrations package to the `drupal-core` group.

## Context

This is a follow-up to https://github.com/renovatebot/renovate/pull/36588. Many Drupal projects using Pantheon hosting use this package. It uses the same versioning and needs to be kept in sync. Major updates to Drupal core require also major updates to this Pantheon package if it is being used.


lib/config/presets/internal/group.ts
```diff
@@ -116,7 +116,11 @@ const staticGroups = {
     packageRules: [
       {
         groupName: 'Drupal core',
-        matchPackageNames: ['drupal/core', 'drupal/core-*'],
+        matchPackageNames: [
+          'drupal/core',
+          'drupal/core-*',
+          'pantheon-systems/drupal-integrations'
```
rarkins, line 122: Doesn't seem to have the _exact_ same release cadence, e.g. its latest is 11.2.0 now and not 11.2.2. This could result in "immortal" PRs from Renovate due to the mismatch of to-versions, but maybe ok

rarkins (approved):

viceice (changes requested):
conflicted and lint failures

deviantintegral:
@davereid the note about exact version matching is interesting, as it doesn't look like the pantheon library has had an update since February 2025. The current 11.2.0 release looks to allow the upcoming Drupal 12 release too. I wonder if they've relaxed their requirements since you looked at it?

https://github.com/pantheon-systems/drupal-integrations/blob/11.x/composer.json#L7

PR #37137, opened 2025-07-23
refactor: fix preset api types

viceice:
## Changes
fix some type issues with nullables

## Context

- needs #37136

### DIFF
--- a/lib/types/index.ts
+++ b/lib/types/index.ts
@@ -13,3 +13,9 @@ export type { PrState } from './pr-state';
 export type { ModuleApi, RenovatePackageJson } from './base';
 
 export type AutoMergeType = 'branch' | 'pr' | 'pr-comment';
+
+type Val = NonNullable<unknown>;
+
+export type Nullable<T extends Val> = T | null | undefined;
+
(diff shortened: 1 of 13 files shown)


lib/types/index.ts
```diff
@@ -13,3 +13,9 @@ export type { PrState } from './pr-state';
 export type { ModuleApi, RenovatePackageJson } from './base';
 
 export type AutoMergeType = 'branch' | 'pr' | 'pr-comment';
+
+type Val = NonNullable<unknown>;
+
+export type Nullable<T extends Val> = T | null | undefined;
```
astellingwerf: It's somewhat counterintuitive that besides null, this type may also be undefined. There are plenty of cases we're changing here, where we'll now allow extra values that we didn't allow before.

viceice: Do you have a better name for this type? Maybe `Nullish` ?

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

viceice:
I'll change more places to use this later. it would make this pr to large.

secustor (approved):

PR #37147, opened 2025-07-24
feat(config)!: replace `kbpgp` with `gpg` CLI

### LINKED ISSUE #30011: Replace PGP with `pgpainless-sop` / `bouncycastle`

Use the `org.pgpainless:pgpainless-sop` java lib compiled to wasm (TeaVM) as a PGP replacement, so we can fully drop `openpgp` because of licensing issues.

- https://github.com/pgpainless/pgpainless
- https://teavm.org/docs/intro/getting-started.html

viceice:
## Changes
Replace the `kbpgp` with the `gnupg` cli.
It's a little slower than the js variant.

## Context
- closes #30011


lib/config/decrypt/gnupg.ts
```diff
@@ -0,0 +1,71 @@
...
+      armoredMessage = `${startBlock}${armoredMessage}`;
+    }
+    if (!armoredMessage.endsWith(endBlock)) {
+      armoredMessage = `${armoredMessage}${endBlock}`;
+    }
+    const encryptedFilePath = upath.join(tmpDir, 'msg.pem');
+    await outputFile(encryptedFilePath, armoredMessage);
+
+    const { stdout, stderr } = await exec(
+      `gpg --batch --no-tty --yes --decrypt ${encryptedFilePath}`,
```
astellingwerf: Please shlex:quote the value of encryptedFilePath to prevent arbitrary command injection

lib/config/decrypt/gnupg.ts
```diff
@@ -0,0 +1,71 @@
...
+  encryptedStr: string,
+): Promise<string | null> {
+  const tmpDir = await mkdtemp(upath.join(os.tmpdir(), 'renovate-gpg-'));
+
+  if (!keyImported.has(privateKey)) {
+    try {
+      const keyFilePath = upath.join(tmpDir, 'key.pem');
+      await outputFile(keyFilePath, privateKey);
+      const { stdout, stderr } = await exec(
+        `gpg --batch --no-tty --yes --import ${keyFilePath}`,
```
astellingwerf: Please shlex:quote the value of keyFilePath to prevent arbitrary command injection

viceice: How should that happen? `keyFilePath` is not dependend to userinput. it's a random file.

I'll quote it to be save against whitespaces on windows.

secustor (dismissed):

rarkins (changes requested):
We need to decide if this is breaking or not

rarkins (changes requested):
Can we make this a breaking change and bump a major release?

rarkins (approved):

package.json
```diff
@@ -173,7 +173,6 @@
     "@pnpm/parse-overrides": "1001.0.2",
     "@qnighy/marshal": "0.1.3",
     "@renovatebot/detect-tools": "1.1.0",
-    "@renovatebot/kbpgp": "4.0.1",
```
jamietanna, line 176: Is it worth then deprecating this library? Looks like it's only Renovate using it: https://deps.dev/npm/%40renovatebot%2Fkbpgp/4.0.3/dependents

viceice: sure, but i'm thinking of using https://github.com/renovatebot/pgp (dotnet wasm variant) instead of the cli 🤔

jamietanna: I've raised https://github.com/renovatebot/kbpgp/issues/216 for the deprecation

jamietanna: > I'm thinking of using https://github.com/renovatebot/pgp (dotnet wasm variant) instead of the cli

Sounds like a plan - is that something that needs much validation ahead of next week's major release?

viceice: no, it's only used by self-hosted users and they can switch to openpgp if there is any isssue

jamietanna:
Replaced by https://github.com/renovatebot/renovate/pull/39032

PR #37159, opened 2025-07-25
feat(branch): allow branch automerging behind main branch

dandandy:
## Changes

This change allows automerging when rebaseWhen=conflict. Previously there was no way to allow not behind branches to auto merge.

It does this by updating the branch check logic to allow auto merge when rebaseWhen=conflict, and use git merge flag `--ff` instead of `--ff-only`. `--ff` was chosen since it will fast forward when the branch is up to date (consistent with previous behaviour), but create a merge commit if it is behind. 

## Context

- From discussion https://github.com/renovatebot/renovate/discussions/37155

#### When rebaseWhen=conflicted and automergeType=branch

1. Renovate creates branch, and a new commit is on main branch. Renovate branch is 1 commit behind
image
2. Renovate is run again, Renovate creates a merge commit (since it uses merge flag `--ff` instead of `--ff-only`) 
image

#### When rebaseWhen=conflicted and automergeType=pr

1. Renovate creates branch and raises PR, and a new commit is on main branch. Renovate branch is 1 commit behind 
image
2. Renovate is run again, Renovate auto merges PR 
image


docs/usage/configuration-options.md
```diff
@@ -345,6 +345,8 @@ If you prefer that Renovate more silently automerge _without_ Pull Requests at a
 - Automerge the branch commit if it's: (a) up-to-date with the base branch, and (b) passing all tests
 - As a backup, raise a PR only if either: (a) tests fail, or (b) tests remain pending for too long (default: 24 hours)
 
+To allow automerges on branches that are behind and green, then add `rebaseWhen="conflicted"`. This will not merge a conflicted branch.
```
```suggestion
To allow automerges on branches that are behind and green, add `rebaseWhen="conflicted"`. 
Note that this will not merge a conflicted branch.
```

RahulGautamSingh (commented):
Else, LGTM

RahulGautamSingh:
To reduce some changes we could do:

```ts
export async function mergeBranch(
  branchName: string,
  mergeFlag: MergeFlag = '--ff-only',
): Promise<void> {
```

lib/workers/repository/update/branch/index.ts
```diff
@@ -673,7 +674,9 @@ export async function processBranch(
     // skip if we have a non-immediate pr and there is an existing PR,
     // we want to update the PR and skip the Auto merge since status checks aren't done yet
     if (!config.artifactErrors?.length && (!commitSha || config.ignoreTests)) {
-      const mergeStatus = await tryBranchAutomerge(config);
+      const mergeFlag: MergeFlag =
+        config.rebaseWhen === 'conflicted' ? '--ff' : '--ff-only';
```
rarkins: We don't want git awareness (flags) at the worker layer here. Instead the param should be more generic like "allowBehindBase" true/false

rarkins (changes requested):

lib/workers/repository/update/branch/index.ts
```diff
@@ -673,7 +674,8 @@ export async function processBranch(
     // skip if we have a non-immediate pr and there is an existing PR,
     // we want to update the PR and skip the Auto merge since status checks aren't done yet
     if (!config.artifactErrors?.length && (!commitSha || config.ignoreTests)) {
-      const mergeStatus = await tryBranchAutomerge(config);
+      const allowBehindBase = config.rebaseWhen === 'conflicted';
```
rarkins: What about the case where rebaseWhen=never ?

dandandy: I thought that we wanted it to allow behind base just when rebaseWhen=conflicted  ( https://github.com/renovatebot/renovate/discussions/37155#discussioncomment-13886560 ), but happy to allow it for rebaseWhen=never too

rarkins: I think when I suggested `conflicted`, I simply wasn't thinking about `never`. Seems to make sense to have both?

dandandy: Done, please take another look

dandandy:
Hey just following up on this as I'm still waiting for a review. Thank you!

lib/workers/repository/update/branch/index.ts
```diff
@@ -458,12 +459,8 @@ export async function processBranch(
       !(keepUpdatedLabel && branchPr?.labels?.includes(keepUpdatedLabel)) &&
       !dependencyDashboardCheck
     ) {
+      config.reuseExistingBranch = true;
```
rarkins: Is there a need to explicitly set it to true here?

lib/workers/repository/update/branch/automerge.ts
```diff
@@ -3,6 +3,7 @@ import type { RenovateConfig } from '../../../../config/types';
 import { logger } from '../../../../logger';
 import { platform } from '../../../../modules/platform';
 import { scm } from '../../../../modules/platform/scm';
+
```
rarkins:
```suggestion
```

lib/workers/repository/update/branch/index.ts
```diff
@@ -26,6 +26,7 @@ import { scm } from '../../../../modules/platform/scm';
 import { ExternalHostError } from '../../../../types/errors/external-host-error';
 import { getElapsedMs } from '../../../../util/date';
 import { emojify } from '../../../../util/emoji';
+
```
rarkins:
```suggestion
```

rarkins (changes requested):
Does this impact when automergeType=pr too?

lib/workers/repository/update/branch/index.ts
dandandy: No looks like this was a mistake, I'll remove it.

dandandy:
> Does this impact when automergeType=pr too?

My changes don't impact when automergeType=pr because of the condition in tryBranchAutomerge.

lib/workers/repository/update/branch/automerge.ts
```diff
@@ -46,7 +47,8 @@ export async function tryBranchAutomerge(
         logger.info(`DRY-RUN: Would automerge branch ${config.branchName!}`);
       } else {
         await scm.checkoutBranch(config.baseBranch!);
-        await scm.mergeAndPush(config.branchName!);
+        const mergeFlag = allowBehindBase ? '--ff' : '--ff-only';
```
rarkins: This is a git-specific concept which is now misplaced in the worker layer. The abstract concept of allowBehindBranch should be converted to the git concept of fast-forward down at the git layer

rarkins (dismissed):

lib/util/git/index.spec.ts
RahulGautamSingh, line 1: Add one test for mergeBranch with allowBehindBase=true

RahulGautamSingh (changes requested):

lib/util/git/index.spec.ts
RahulGautamSingh, line 1: Please change the test descriptions to:

1. merges branch when branch is behind base
2. merges branch when branch is upto date with base

lib/workers/repository/update/branch/index.spec.ts
RahulGautamSingh, line 1: what happens when branch is set for automerge and is behind base but conflicted?

dandandy: I've added a unit test for this, it should throw an exception because it's conflicted.

RahulGautamSingh: If branch is conflicted and rebaseWhen=conflicted it should rebase the branch and then automerge. If rebaseWhen=never then we can error/skip.

dandandy: I see, I've added some code and tests to address this.

lib/workers/repository/update/branch/index.ts
```diff
@@ -724,6 +724,24 @@ export async function processBranch(
           };
         }
       }
+      if (
+        mergeStatus === 'stale' &&
+        ['never'].includes(config.rebaseWhen!) &&
```
RahulGautamSingh:
```suggestion
        config.rebaseWhen === 'never' &&
```

RahulGautamSingh (commented):
LGTM

lib/workers/repository/update/branch/index.ts
```diff
@@ -702,7 +697,9 @@ export async function processBranch(
     // skip if we have a non-immediate pr and there is an existing PR,
     // we want to update the PR and skip the Auto merge since status checks aren't done yet
     if (!config.artifactErrors?.length && (!commitSha || config.ignoreTests)) {
-      const mergeStatus = await tryBranchAutomerge(config);
+      const allowBehindBase =
+        config.rebaseWhen === 'conflicted' || config.rebaseWhen === 'never';
+      const mergeStatus = await tryBranchAutomerge(config, allowBehindBase);
```
viceice: I don't like to change the default behavior without a conflig option. this is a breaking change.

viceice (commented):
marked as breaking because it's a behavior change and maybe unexpected to some users.

@rarkins suggested to implement it this way, so will accept it when all open conversations are resolved

lib/workers/repository/update/branch/index.ts
dandandy: I don't mind either way. Maybe you two can discuss and let me know?

rarkins:
I'd suggested merging it after the next major release at the time because it had a higher than normal chance that it might break something and that way we avoid having to go back to the old major release to fix

lib/workers/repository/update/branch/index.ts
secustor: Why adding this new parameter and not instead having this logic inside of `tryBranchAutomerge`?  

I would prefer an experimental feature flag here too and in the next major it can be enabled by default.

lib/workers/repository/update/branch/index.spec.ts
```diff
@@ -708,6 +708,88 @@ describe('workers/repository/update/branch/index', () => {
...
+      );
+      npmPostExtract.getAdditionalFiles.mockResolvedValueOnce({
+        artifactErrors: [],
+        updatedArtifacts: [],
+      });
+      scm.branchExists.mockResolvedValue(true);
+      scm.isBranchConflicted.mockResolvedValue(false);
+      platform.getBranchPr.mockResolvedValueOnce(null as any);
+      checkExisting.prAlreadyExisted.mockResolvedValueOnce(undefined as any);
+      commit.commitFilesToBranch.mockResolvedValue(null as any);
```
secustor: Why the `any`s?

lib/workers/repository/update/branch/index.spec.ts
```diff
@@ -708,6 +708,88 @@ describe('workers/repository/update/branch/index', () => {
...
+      config.ignoreTests = true;
+      config.prCreation = 'immediate';
+      config.reuseExistingBranch = true;
+      config.cacheFingerprintMatch = 'matched' as CacheFingerprintMatchResult;
+      await branchWorker.processBranch(config);
+      expect(automerge.tryBranchAutomerge).toHaveBeenCalledTimes(1);
+      expect(automerge.tryBranchAutomerge).toHaveBeenCalledWith(
+        expect.any(Object),
+        true,
+      );
```
secustor: Can be replaced with `toHaveBeenCalledExactlyOnceWith`

lib/workers/repository/update/branch/index.spec.ts
```diff
@@ -708,6 +708,88 @@ describe('workers/repository/update/branch/index', () => {
...
+      automerge.tryBranchAutomerge.mockResolvedValueOnce('automerged');
+      config.automerge = true;
+      config.automergeType = 'branch';
+      config.rebaseWhen = 'auto';
+      await branchWorker.processBranch(config);
+      expect(automerge.tryBranchAutomerge).toHaveBeenCalledTimes(1);
+      expect(automerge.tryBranchAutomerge).toHaveBeenCalledWith(
+        expect.any(Object),
+        false,
+      );
```
secustor, line 1018: Same here.

docs/usage/configuration-options.md
```diff
@@ -345,6 +345,9 @@ If you prefer that Renovate more silently automerge _without_ Pull Requests at a
 - Automerge the branch commit if it's: (a) up-to-date with the base branch, and (b) passing all tests
 - As a backup, raise a PR only if either: (a) tests fail, or (b) tests remain pending for too long (default: 24 hours)
 
+To allow automerges on branches that are behind and green, add `rebaseWhen="conflicted"` or `rebaseWhen="never"`.
+Note that this will not merge a conflicted branch.
```
secustor: IMO this should have another semantic as from the users perspective it is confusing that users should 
```suggestion
Automerges will only be triggered if a fast forward merges are possible. The only exceptions are: 
- `rebaseWhen="conflicted"`
- `rebaseWhen="never"`
in these cases merge commits will be created if necessary. 
```

lib/workers/repository/update/branch/index.ts
dandandy: Yes, I can fix this and add an experimental feature flag. 

By experimental feature flag, do you mean "Experimental variables" environment variables that start with `RENOVATE_X_`?

Or is an experimental feature flag something else?

docs/usage/configuration-options.md
dandandy: Since we are now talking about putting this feature behind an experimental flag, should we revert the docs?

lib/workers/repository/update/branch/index.spec.ts
dandandy: I was having type issues, but it's not happening now so I've removed them.

RahulGautamSingh:
Lint failing

lib/workers/repository/update/branch/index.ts
```diff
@@ -536,11 +536,6 @@ export async function processBranch(
       !dependencyDashboardCheck
     ) {
       logger.debug('rebaseWhen=never so skipping branch update check');
```
viceice: is this log messages still valid after removing the return below?

viceice (commented):
sorry for late review, LGTM beside this: 

please rebase/ merge main and fix lint issues

lib/workers/repository/update/branch/index.ts
dandandy: Not necessary, I have removed it.

lib/workers/repository/update/branch/index.ts
dandandy: I have added the experimental flag

RahulGautamSingh (approved):
LGTM

lib/workers/repository/update/branch/index.ts
```diff
@@ -778,6 +766,9 @@ export async function processBranch(
     // skip if we have a non-immediate pr and there is an existing PR,
     // we want to update the PR and skip the Auto merge since status checks aren't done yet
     if (!config.artifactErrors?.length && (!commitSha || config.ignoreTests)) {
+      const allowBehindBase =
+        config.allowBranchAutomergeBehindBase === true &&
+        (config.rebaseWhen === 'conflicted' || config.rebaseWhen === 'never');
```
viceice: use array style like the removed code below

lib/workers/repository/update/branch/automerge.ts
```diff
@@ -46,7 +46,10 @@ export async function tryBranchAutomerge(
         logger.info(`DRY-RUN: Would automerge branch ${config.branchName!}`);
       } else {
         await scm.checkoutBranch(config.baseBranch!);
-        await scm.mergeAndPush(config.branchName!);
+        const allowBehindBase =
+          config.allowBranchAutomergeBehindBase === true &&
+          (config.rebaseWhen === 'conflicted' || config.rebaseWhen === 'never');
```
viceice: use array style, see below

dandandy:
Hey could I please get another review? Thank you

viceice (dismissed):

viceice:
removed breaking label because it#s now opt-in

@secustor can you review to have some more eyes?

secustor (dismissed):

RahulGautamSingh (commented):
Could you resolve the conflicts @dandandy

dandandy:
> Could you resolve the conflicts @dandandy

That is fixed now @RahulGautamSingh

lib/util/git/index.spec.ts
```diff
@@ -779,7 +822,7 @@ describe('util/git/index', { timeout: 30000 }, () => {
         isSymlink: true,
       };
       const commit = await git.commitFiles({
-        branchName: 'renovate/future_branch',
+        branchName: 'renovate/symlink_branch',
```
secustor: Why this change?

secustor (commented):
all committers have to sign the CLA

lib/util/git/index.spec.ts
dandandy: for some reason this test case behaves differently on my local than in your pipeline. I've since restored it.

secustor (approved):
LGTM, but waiting for a second pair of eyes

PR #37161, opened 2025-07-25
feat(bitbucket-server): extend `CODEOWNER` support

portly-halicore-76:
## Changes

As described in the discussion here: https://github.com/renovatebot/renovate/discussions/36996 I've implemented the BitBucket Server CODEOWNER logic. This includes:

* Implements expandGroupsMembers for BitBucket Server as codeOwners can contain reviewer-groups
* Implements extractRulesFromCodeOwnersLines for BitBucket Server 

## Context

See related Discussion. More info about CODEOWNERS for BitBucket server can be found [here](https://confluence.atlassian.com/bitbucketserver/code-owners-1296171116.html) 

I believe I can't update that the expandUserGroups configuration option is implemented for BitBucket server as well, Is that generated documentation?


RahulGautamSingh:
Thanks for the PR!

Do you think that some parts can be separated into smaller PRs so it's easier to review?

For eg. the `addReviewers` refactor I think, can be done separately?

portly-halicore-76:
> Thanks for the PR!
> 
> Do you think that some parts can be separated into smaller PRs so it's easier to review?
> 
> For eg. the `addReviewers` refactor I think, can be done separately?

Thanks for your early feedback. The refactor in addReviewers is strictly related to the CODEOWNERS implementation. However, it can be regarded as a seperate change. I will split the PR!

portly-halicore-76:
@viceice why is this marked as draft? Because of the testCoverage not met?

viceice:
> @viceice why is this marked as draft? Because of the testCoverage not met?

because it depends on 
- #37199

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1154,6 +1156,152 @@ export async function mergePr({
...
+  cleanedLines: string[],
+): FileOwnerRule[] {
+  const results: FileOwnerRule[] = [];
+
+  const reversedLines = cleanedLines
+    .filter((line) => line.trim() !== '' && !line.trim().startsWith('#'))
+    .reverse();
+
+  for (const line of reversedLines) {
+    const safeLine = line.replace(/\\\\ /g, '<<SPACE>>'); // double-backslash before space
```
astellingwerf: Are you sure that `<<SPACE>>` is never part of the valid input? Would it not be safer to use something like this:
```javascript
function splitEscapedSpaces(str) {
  return str.split(' ').reduce((acc, part) => {
    const prev = acc[acc.length - 1];

    if (prev && prev.endsWith('\\')) {
      acc[acc.length - 1] = prev.slice(0, -1) + ' ' + part;
    } else {
      acc.push(part);
    }

    return acc;
  }, []);
}
```

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1154,6 +1156,152 @@ export async function mergePr({
...
+
+  return null; // Not a random selector
+}
+
+function resolveRandom(users: string[], count?: number): string[] {
+  if (!count || users.length <= count) {
+    return users;
+  }
+  const shuffled = [...users].sort(() => 0.5 - Math.random());
+  return shuffled.slice(0, count);
```
astellingwerf: A constant minus a random number is just as random as the random number on its own.

It's highly inefficient to first shuffle the entire array, and then only pick `count` elements. Can you generate `count` unique indices and return the matching values from the array of users?

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1154,6 +1156,152 @@ export async function mergePr({
...
+      return parseInt(match[1], 10); // e.g., random(2) → 2
+    } else {
+      return 1; // plain "random"
+    }
+  }
+
+  return null; // Not a random selector
+}
+
+function resolveRandom(users: string[], count?: number): string[] {
```
astellingwerf: We know that `count` won't be null at this stage

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1154,6 +1156,152 @@ export async function mergePr({
...
+  if (value.startsWith(randomPrefix)) {
+    const match = /^random\((\d+)\)$/.exec(value);
+    if (match) {
+      return parseInt(match[1], 10); // e.g., random(2) → 2
+    } else {
+      return 1; // plain "random"
+    }
+  }
+
+  return null; // Not a random selector
```
astellingwerf: This could be simplified to:
```suggestion
  const match = /^random(?:\((\d+)\))?$/.exec(value);
  if (match) {
    return parseInt(match[1] ?? "1", 10); // e.g., random(2) → 2
  } else {
    return null;
  }
```

There should be a test case for the `return null` scenario too

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1154,6 +1156,152 @@ export async function mergePr({
...
+}
+
+function resolveRandom(users: string[], count?: number): string[] {
+  if (!count || users.length <= count) {
+    return users;
```
astellingwerf: This case should be covered by a test

lib/modules/platform/bitbucket-server/index.ts
Churro: If I may chime in here, there is a Fisher-Yates shuffle implementation that could be reused:
https://github.com/renovatebot/renovate/blob/da20597a6369b3ec132ca2ee176b5df750c2104b/lib/util/sample.ts#L1

astellingwerf: Even better 👍

astellingwerf (approved):

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1221,18 +1221,12 @@ function isReviewerGroup(entry: string): boolean {
...
-  const randomPrefix = 'random';
-
-  if (value.startsWith(randomPrefix)) {
-    const match = /^random\((\d+)\)$/.exec(value);
-    if (match) {
-      return parseInt(match[1], 10); // e.g., random(2) → 2
-    } else {
-      return 1; // plain "random"
-    }
+  const match = /^random(?:\((\d+)\))?$/.exec(value);
```
RahulGautamSingh:
```suggestion
  const match = regEx(/^random(?:\((\d+)\))?$/).exec(value);
```

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1221,18 +1221,12 @@ function isReviewerGroup(entry: string): boolean {
...
+  if (match) {
+    return parseInt(match[1] ?? "1", 10); // e.g., random(2) → 2
```
RahulGautamSingh:
```suggestion
    return parseInt(match[1] ?? "1", 10);
```

RahulGautamSingh (commented):
.

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: Already committed this.

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1214,6 +1217,152 @@ export async function mergePr({
...
+  reviewers: string[],
+): Promise<string[]> {
+  logger.debug(`expandGroupMembers(${reviewers.join(', ')})`);
+  const expandedUsers: string[] = [];
+
+  for (const reviewer of reviewers) {
+    const [baseEntry, modifier] = reviewer.split(':');
+
+    if (isReviewerGroup(baseEntry)) {
+      const groupName = baseEntry.replace(regEx('^@reviewer-group/'), '');
```
Churro:
```suggestion
    if (baseEntry.startsWith('@reviewer-group/')) {
      const groupName = baseEntry.substring(16);
```

* IMHO indirection for a one-line check that's used only once is too much of abstraction
* RegEx match-replace is unnecessary, as it's already known here that it's a reviewer group

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1214,6 +1217,152 @@ export async function mergePr({
...
+      const groupUsers = await getUsersFromReviewerGroup(groupName);
+      if (groupUsers.length === 0) {
```
Churro:
```suggestion
      if (!groupUsers.length) {
```

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1214,6 +1217,152 @@ export async function mergePr({
...
+      if (modifier === undefined) {
+        expandedUsers.push(...groupUsers); // Add all users from the group
+        continue;
+      }
+      const randomCount = parseModifier(modifier);
+      if (randomCount === null) {
+        expandedUsers.push(...groupUsers);
+        continue;
+      }
+      expandedUsers.push(...sampleSize(groupUsers, randomCount));
```
Churro:
```suggestion
      if (modifier) {
        const randomCount = parseModifier(modifier);
        if (randomCount) {
          expandedUsers.push(...sampleSize(groupUsers, randomCount));
          continue;
        }
      }

      expandedUsers.push(...groupUsers);
```

Inverting the logic makes it more explicit what is going on and also removes yet another `expandedUsers.push(...groupUsers); continue;`

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1214,6 +1217,152 @@ export async function mergePr({
...
+      result.push(part);
+    }
+  }
+
+  return result;
+}
+
+function isReviewerGroup(entry: string): boolean {
+  return entry.startsWith('@reviewer-group/');
+}
```
Churro:
```suggestion

```

Can be removed if inlined

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1214,6 +1217,152 @@ export async function mergePr({
...
+  interface reviewerGroup {
+    name: string;
+    users: { emailAddress: string; active: boolean }[];
+    scope: { type: 'REPOSITORY' | 'PROJECT' };
+  }
+
+  const allGroups: reviewerGroup[] = [];
+
+  try {
+    const { body } = await bitbucketServerHttp.getJsonUnchecked<
```
Churro: Please use `getJson` with `reviewerGroup` in schema.ts

Churro (changes requested):

lib/modules/platform/bitbucket-server/index.ts
astellingwerf: I dislike the magic number in the suggested approach

Churro: Given that the preceding line includes `@reviewer-group/`, I'd say that the number's origin is very obvious. Defining `@reviewer-group/` as a const string and using `.length` would be a more obvious alternative that's still cheaper than regex.

Churro (approved):

lib/workers/repository/update/pr/code-owners.spec.ts
```diff
@@ -285,6 +286,82 @@ describe('workers/repository/update/pr/code-owners', () => {
...
+      beforeAll(() => {
+        Object.defineProperty(platform, 'extractRulesFromCodeOwnersLines', {
+          value: bitbucketserver.extractRulesFromCodeOwnersLines,
+          writable: true,
+        });
+      });
+
+      it('returns code owners for matching file using escaped spaces', async () => {
+        fs.readLocalFile.mockResolvedValueOnce(
+          String.raw`src/** @Jane\\ Doe @john@example.com`,
```
RahulGautamSingh, line 299:
```suggestion
          'src/** @Jane\\ Doe @john@example.com',
```

lib/workers/repository/update/pr/code-owners.spec.ts
```diff
@@ -285,6 +286,82 @@ describe('workers/repository/update/pr/code-owners', () => {
...
+
+        const codeOwners = await codeOwnersForPr(pr);
+
+        // Since we don't simulate the actual group resolution, this will just include the literal string
+        expect(codeOwners).toEqual([
+          '@reviewer-group/content-designers:random',
+        ]);
+      });
+
+      it('ignores lines with no owners and uses fallback for empty rule', async () => {
```
RahulGautamSingh: What is the fallback in this test?

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1214,6 +1223,141 @@ export async function mergePr({
...
+}
+
+function parseModifier(value: string): number | null {
+  const match = regEx('^random(?:\\((\\d+)\\))?$').exec(value);
+  if (match) {
+    return parseInt(match[1] ?? '1', 10);
+  } else {
+    return null;
+  }
+}
```
RahulGautamSingh:
```suggestion
  if (!match) {
    return null;
  }

  return parseInt(match[1] ?? '1', 10);
}
```

We prefer this style

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1214,6 +1223,141 @@ export async function mergePr({
...
+      usernames: [...new Set(entries)],
+      score: pattern.length,
+      match: (path: string) => matcher.ignores(path),
+    });
+  }
+
+  return results;
+}
+
+function splitEscapedSpaces(str: string): string[] {
```
RahulGautamSingh: Move this to util file

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1214,6 +1223,141 @@ export async function mergePr({
...
+  const parts = str.split(' ');
+  const result: string[] = [];
+
+  for (const part of parts) {
+    const last = result[result.length - 1];
+
+    if (last?.endsWith(String.raw`\\`)) {
```
RahulGautamSingh:
```suggestion
    if (last?.endsWith('\\')) {
```

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1214,6 +1223,141 @@ export async function mergePr({
...
+
+  return results;
+}
+
+function splitEscapedSpaces(str: string): string[] {
+  const parts = str.split(' ');
+  const result: string[] = [];
+
+  for (const part of parts) {
+    const last = result[result.length - 1];
```
RahulGautamSingh: This can be moved outside the loop

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -1214,6 +1223,141 @@ export async function mergePr({
...
+
+    allGroups.push(...reviewerGroups.body);
+  } catch (err) {
+    logger.debug({ err, groupName }, 'Failed to get reviewer groups for repo');
+    return [];
+  }
+
+  // First, try to find a repo-scoped group with this name
+  const repoGroup = allGroups.find(
+    (group) => group.name === groupName && group.scope?.type === 'REPOSITORY',
```
RahulGautamSingh, line 1301:
```suggestion
    (group) => group.name === groupName && (group.scope?.type === 'REPOSITORY' || group.scope?.type === 'PROJECT'),
```
doing this should reduce number of lines

RahulGautamSingh (changes requested):

lib/workers/repository/update/pr/code-owners.spec.ts
portly-halicore-76: This is done on purpose to prevent escaping slashes etc. I have some issues getting exactly the same content as being produced by a real reposity in BitBucket Server. it must be a space escaped by two slashes in front. When commiting your suggestion, this would be one slash escaped, causing difficulties. Therefore, please leave this as is.

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: Same as before: This is done on purpose to prevent escaping slashes etc. I have some issues getting exactly the same content as being produced by a real reposity in BitBucket Server. it must be a space escaped by two slashes in front. When commiting your suggestion, this would be one slash escaped, causing difficulties. Therefore, please leave this as is.

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: Nope, we should implement the behaviour BitBucket Server implements; a group found on `REPOSITORY` level takes presedense over a group found on `PROJECT` level. Therefore, please leave as is.

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: Done!

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: Yes is can.

lib/workers/repository/update/pr/code-owners.spec.ts
portly-halicore-76: The test description is not right. Following bitbucket server behavior, a rule without owners can be used prevented to prevent users being added. I've updated the description. 

See: https://confluence.atlassian.com/bitbucketserver/code-owners-1296171116.html

portly-halicore-76:
@RahulGautamSingh can you give your final thoughts about this. Looking much forward to use this functionality.

lib/modules/platform/bitbucket-server/schema.ts
```diff
@@ -35,6 +35,18 @@ export const PullRequestActivity = z.union([
...
 
+export const ReviewerGroup = z.object({
+  name: z.string(),
+  users: z.array(UserSchema),
+  scope: z.object({
+    type: z.union([z.literal('REPOSITORY'), z.literal('PROJECT')]),
+  }),
+});
+
+export const ReviewerGroupsSchema = z.array(ReviewerGroup);
```
viceice:
```suggestion
export const ReviewerGroups = z.array(ReviewerGroup);
```

no trailing `Schema`

https://github.com/renovatebot/renovate/blob/main/docs/development/zod.md#use-schemats-files-for-zod-schemas

lib/modules/platform/bitbucket-server/utils.ts
```diff
@@ -148,3 +149,28 @@ export function getExtraCloneOpts(opts: HostRule): GitOptions {
...
   return {};
 }
+
+export function splitEscapedSpaces(str: string): string[] {
+  const parts = str.split(' ');
+  const result: string[] = [];
+  let last: string | undefined;
+
+  for (const part of parts) {
+    if (last?.endsWith(String.raw`\\`)) {
```
viceice: why this `String.raw` 🤔

rarkins (approved):

rarkins (dismissed):

lib/modules/platform/bitbucket-server/utils.ts
```diff
@@ -148,3 +149,28 @@ export function getExtraCloneOpts(opts: HostRule): GitOptions {
...
+
+  return result;
+}
+
+export function parseModifier(value: string): number | null {
+  const match = regEx('^random(?:\\((\\d+)\\))?$').exec(value);
+  if (!match) {
+    return null;
+  }
+  return parseInt(match[1] ?? '1', 10);
```
viceice:
```suggestion
  return parseInt(match[1] ?? '1');
```

`10` is the default radix since ecma 5

viceice (commented):
see comments

lib/modules/platform/bitbucket-server/utils.ts
portly-halicore-76: As explained to @RahulGautamSingh as I want to make sure we exactly parse two slashes. As we dont have to escape any, we are good Here.

portly-halicore-76: Let's run the tests without this... 🤞

portly-halicore-76: Did it!

lib/modules/platform/bitbucket-server/schema.ts
```diff
@@ -35,6 +35,18 @@ export const PullRequestActivity = z.union([
...
+  name: z.string(),
+  users: z.array(UserSchema),
+  scope: z.object({
+    type: z.union([z.literal('REPOSITORY'), z.literal('PROJECT')]),
+  }),
+});
+
+export const ReviewerGroups = z.array(ReviewerGroup);
+
+export type ReviewerGroups = z.infer<typeof ReviewerGroups>;
```
viceice:
```suggestion
```

not used

lib/modules/platform/bitbucket-server/utils.ts
```diff
@@ -148,3 +149,28 @@ export function getExtraCloneOpts(opts: HostRule): GitOptions {
...
+  const result: string[] = [];
+  let last: string | undefined;
+
+  for (const part of parts) {
+    if (last?.endsWith('\\\\')) {
+      result[result.length - 1] = last.slice(0, -2) + ' ' + part;
+    } else {
+      result.push(part);
+    }
+    last = result[result.length - 1];
```
viceice:
```suggestion
    last = result.at(-1);
```

viceice (commented):
otherwise LGTM 🤗

rarkins (approved):

PR #37186, opened 2025-07-28
fix(jsonata): Eval later queries despite non-match

jplitza:
## Changes

This commit lets the jsonata custom manager continue with the other queries in the same file if one doesn't match or produces errors.

## Context

The jsonata custom manager can be provided with multiple queries via the matchStrings configuration option.

Before this commit, if any of those doesn't match for a file, all later querites are skipped. This doesn't make sense, since the queries can be completely independent. As an example, one might want to look for image names in several places of every CI pipeline. Not matching in one place of one pipeline shouldn't cause the other places to be ignored in that
file.

### DIFF
--- a/lib/modules/manager/custom/jsonata/utils.ts
+++ b/lib/modules/manager/custom/jsonata/utils.ts
@@ -33,7 +33,7 @@ export async function handleMatching(
         },
         'The jsonata query returned no matches. Possible error, please check your query. Skipping',
       );
-      return [];
+      continue;
     }
 
     const parsed = QueryResultZodSchema.safeParse(queryResult);
@@ -44,7 +44,7 @@ export async function handleMatching(
         { err: parsed.error, jsonataQuery: query, packageFile, queryResult },
         'Query results failed schema validation',
       );
-      return [];
+      continue;
     }
   }
 
--- a/lib/modules/manager/custom/jsonata/index.spec.ts
+++ b/lib/modules/manager/custom/jsonata/index.spec.ts
@@ -332,6 +332,36 @@ describe('modules/manager/custom/jsonata/index', () => {
     });
   });
 
+  it('extracts other matchStrings if one finds no match', async () => {
+    const config = {
+      fileFormat: 'json',
+      matchStrings: [
+        `packages.{ "depName": package }`,
+        `{"depName": "bar"}`],
+      currentValueTemplate: '1.0.0',
+      datasourceTemplate: 'npm',
+    };
+    const res = await extractPackageFile('{}', 'unused', config);
+    expect(logger.logger.debug).toHaveBeenCalledWith(
+      {
+        packageFile: 'unused',
+        jsonataQuery:
+          'packages.{ "depName": package }',
+      },
+      'The jsonata query returned no matches. Possible error, please check your query. Skipping',
+    );
+    expect(res).toMatchObject({
+      ...config,
+      deps: [
+        {
+          depName: 'bar',
+          currentValue: '1.0.0',
+          datasource: 'npm',
+        },
+      ],
+    });
+  });
+
   it('populates manager config and jsonata manager template fields in extract result', async () => {
     const config = {
       fileFormat: 'json',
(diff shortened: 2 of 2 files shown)


RahulGautamSingh:
Thank you fro the PR! It can make sense in some cases to continue for the other queries.

Please, fix the tests so I can approve this

jplitza:
@RahulGautamSingh The failing tests are completely unrelated to the change and the same as for 8e2d402e9af00b381bb515e04b48865241cbb9a1, the latest commit on main. So I doubt I can do anything about them, except maybe rebasing to the older 46433cd51c4d34225b43ffe915e52650d3715742, the last commit on main where all was green.

rarkins:
We are fixing main. But this PR needs its own tests (an example which fails on main, works on this PR)

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -44,7 +44,7 @@ export async function handleMatching(
         { err: parsed.error, jsonataQuery: query, packageFile, queryResult },
         'Query results failed schema validation',
       );
-      return [];
+      continue;
```
rarkins, line 47: Is the warn above this still appropriate?

lib/modules/manager/custom/jsonata/utils.ts
```diff
@@ -33,7 +33,7 @@ export async function handleMatching(
         },
         'The jsonata query returned no matches. Possible error, please check your query. Skipping',
```
rarkins, line 34: Is this message still correct, if this is a "typical" use case?

lib/modules/manager/custom/jsonata/utils.ts
jplitza: Yes, since any individual query generating unexpected schema is probably faulty.

lib/modules/manager/custom/jsonata/utils.ts
jplitza: I guess that's a question of taste. AFAIK the regex custom manager doesn't warn if its regular expression doesn't match in a file, so I'd say the jsonata custom manager shouldn't either.

OTOH, if you defined the query in the repo's renovate.json, the `managerFilePatterns` only match this single file, and then the JSONata query doesn't, that probably isn't what the user wanted.

Then again, the user will see that the expected dependency was detected, which I would consider to be enough feedback.

So I'd slightly lean to removing the message. It's only debug though.

rarkins: Sounds like it could be downgraded to TRACE. @RahulGautamSingh wdyt?

viceice (commented):
I would prefer to add an option to configure this like in regex manager. otherwise we've a breaking behavior here.

https://docs.renovatebot.com/configuration-options/#matchstringsstrategy

jplitza:
@viceice I don't quite get the comparison. The way I understand it, this PR would make the jsonata custom manager behave like the "any" (default) strategy of the regex custom manager. The way it currently works isn't equivalent to any of those strategies.

And I cannot think of any example where the current behavior of skipping all queries if any query doesn't match is intended, since they cannot be combined in any way (like with the "combination" strategy of the regex custom manager). Every query has to produce all required attributes on its own. The fact that you can specify multiple matchStrings at all is basically just a shortcut to specifying multiple custom managers - with this undocumented exception that if one query doesn't match, all are skipped.

lib/modules/manager/custom/jsonata/utils.ts
RahulGautamSingh: We could move the message to after all queries have been processed and see still have no matches.

secustor:
I agree here with `jplitza`. An user that provides multiple queries will expect based on the docs and our `regex` manager this is the behaviour. 

In regard to this being a breaking change, as Renovate would only detect / update additional dependencies I would release this as a fix.

rarkins (approved):

RahulGautamSingh (approved):

PR #37199, opened 2025-07-28
feat(bitbucket-server): Resolves reviewer email-addresses to Users

portly-halicore-76:
## Changes

The `addReviewers` function in BitBucket Server, uses a helper to find the username by a REST call. If the user is not found, it is not added to the PR. This prevents dealing with the `invalidReviewersException` which might be thrown by BitBucket Server. 

## Context

The `addReviewers` function in BitBucket Server, swallowed the invalidReviewersException. In reality, this means, no reviewers were added to PR's where one reviewer was invalid. This in contrast to the `updatePR` func, in which the invalid reviewers were filtered and the request was sent again. However, when implementing https://github.com/renovatebot/renovate/discussions/36996, the code must deal with finding userNames based on emailaddresses which might be set in CODEOWNERS. That led to this implementation.


portly-halicore-76:
Split from https://github.com/renovatebot/renovate/pull/37161

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -691,6 +691,43 @@ export async function addReviewers(
...
+ *
+ * @param userinfo - A string that could be the user's email, display name, or username.
+ * @returns An object containing the user's `slugs` for active found users.
+ */
+async function getUserDetails(userinfo: string): Promise<string[]> {
+  try {
+    const url = `./rest/api/1.0/users?filter=${encodeURIComponent(userinfo)}&permission.1=REPO_READ&permission.1.projectKey=${encodeURIComponent(
+      config.projectKey,
+    )}&permission.1.repositorySlug=${encodeURIComponent(config.repositorySlug)}`;
+    const response = await bitbucketServerHttp.getJsonUnchecked<
```
Churro: Please use `getJson` with response validation instead.

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -691,6 +691,43 @@ export async function addReviewers(
...
+ * Behavior:
+ * - If no users match the filter, returns empty array.
+ * - Returns the `slug` of the matching active users.
+ *
+ * @param userinfo - A string that could be the user's email, display name, or username.
+ * @returns An object containing the user's `slugs` for active found users.
+ */
+async function getUserDetails(userinfo: string): Promise<string[]> {
+  try {
+    const url = `./rest/api/1.0/users?filter=${encodeURIComponent(userinfo)}&permission.1=REPO_READ&permission.1.projectKey=${encodeURIComponent(
```
Churro: `encodeURIComponent` is not needed, the request URI is automatically encoded as needed.

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -691,6 +691,43 @@ export async function addReviewers(
...
+      config.projectKey,
+    )}&permission.1.repositorySlug=${encodeURIComponent(config.repositorySlug)}`;
+    const response = await bitbucketServerHttp.getJsonUnchecked<
+      {
+        slug: string;
+        active: boolean;
+      }[]
+    >(url, {
+      paginate: true,
```
Churro: Since `userInfo` carries only one specific reviewer user or email address, wouldn't `paginate: false` make more sense? `filter` uses `contain()`, so if someone specifies a reviewer name by the name `a`, dozens of pages might be traversed.

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -703,22 +740,34 @@ async function updatePRAndAddReviewers(
 
     // TODO: can `reviewers` be undefined? (#22198)
     const reviewersSet = new Set([...pr.reviewers!, ...reviewers]);
+    const reviewerSlugs = new Set<string>();
+
+    // Find username, as reviever can be an emailaddress or a username, and the request only wants usernames
+    // If the user is not found, it will be removed from the reviewers list
+    for (const entry of reviewersSet) {
+      const users = await getUserDetails(entry);
+      if (users.length === 0) {
```
Churro: `!users.length`

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -703,22 +740,34 @@ async function updatePRAndAddReviewers(
 
     // TODO: can `reviewers` be undefined? (#22198)
     const reviewersSet = new Set([...pr.reviewers!, ...reviewers]);
+    const reviewerSlugs = new Set<string>();
+
+    // Find username, as reviever can be an emailaddress or a username, and the request only wants usernames
+    // If the user is not found, it will be removed from the reviewers list
+    for (const entry of reviewersSet) {
+      const users = await getUserDetails(entry);
```
Churro: If I'm not mistaken a reviewer entry is supposed to be a username or email address, so when would one except multiple user objects as a result here?

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -703,22 +740,34 @@ async function updatePRAndAddReviewers(
...
+          reviewers: Array.from(reviewerSlugs).map((name) => ({
             user: { name },
           })),
         },
       },
     );
     await getPr(prNo, true);
   } catch (err) {
-    logger.warn({ err, reviewers, prNo }, `Failed to add reviewers`);
+    logger.debug({ err, prNo }, `Failed to add reviewers`);
```
Churro: Why `logger.debug`? Even if inactive/invalid reviewers are filtered beforehand, there are other conditions why this query could fail, e.g. 404 or 409.

Churro (changes requested):
I'd suggest moving the resolving of reviewers right before the `await retry...` in `addReviewers` and pass it the filtered list. Otherwise, renovate would recurringly resolve all (invalid) reviewers up to 3 times if updating/adding them fails e.g. due to BB rate limiting:

https://github.com/renovatebot/renovate/blob/c7b653e33951db904d179f43196916be4c41d09e/lib/modules/platform/bitbucket-server/index.ts#L683-L692

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: I can understand where this comes from. Yes, a username should be only unique. However, I found that multiple users within BitBucket Server can use the same emailaddress. Therefore, multiple users can be returned. Therefore, this should be a set IMHO.

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: Will revert this.

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: Can you explain a bit more? The endpoint returns paged results. It will hopefully return just one user by username, but as there is no endpoint which returns users based on an emailaddress, this is IMHO the best we can do. See the spec: https://developer.atlassian.com/server/bitbucket/rest/v906/api-group-system-maintenance/#api-api-latest-users-get

What do you suggest?

Churro: As also stated in the linked REST API docs, `filter=...` leads to an imprecise search with `contains()` on username, name, or email address. If someone specifies just the letters `a`, `b`, `c `, `d`, etc. in a CODEOWNERS file or the renovate settings, every single of these entries would potentially lead to a very high number of result pages being traversed (+ uncached).

The endpoint returns paged results but specifying `paginate: false` just means "fetch the first page only". 

An even more restrictive version would be: `paginate: false, limit: 1` to fetch only the first page and return only one result. The problem then is "what if that first entry" isn't the one you were looking for but a "wildcard match" with another user, so `peginate: false` in combination with checking if any of the 25 users on page has exactly that username or email address might be a good trade-off.

lib/modules/platform/bitbucket-server/index.ts
Churro: In that case, additional safeguards seem reasonable to prevent "reviewer bombs": 
E.g., if I specify `a` as a reviewer, I'd have anyone whose user, name or email address contains the letter `a` added to the PR. Even with "regular use", there might be collisions, e.g., specifying the user `admin` would mean that users `admin1`, `admin2`, ... would also be added, analogous the email `test@example.com` could mean that `anothertest@example.com` gets also added.

Perhaps this can be mitigated by enforcing that user or email address must match the provided reviewer `entry`. 

I'm not convinced though that the handling of an edge case justifies crawling potentially dozens of pages with users recurringly for every reviewer and PR.

lib/modules/platform/bitbucket-server/index.ts
Churro: Perhaps a "cheaper" and more targeted approach could be to try resolving "non-email addresses" with _[Get User](https://developer.atlassian.com/server/bitbucket/rest/v906/api-group-system-maintenance/#api-api-latest-users-userslug-get)_ first and if this doesn't result in the expected user object, to then do a search over all users.

portly-halicore-76: WDYT about this approach? Tests will be updated when we agree on this.

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: WDYT about this approach? Tests will be updated when we agree on this.

portly-halicore-76: In this way, we don't check the users permissions for this repo, which are found based on their slug. For this, I will create a follow-up to deal with invalidReviewerException in the same fashion as the `updatePr` function does. I'll wait until we have a Go on this PR.

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -680,17 +681,91 @@ export function addAssignees(iid: number, assignees: string[]): Promise<void> {
...
+ *
+ * Lookup strategy:
+ *  1. Try exact slug match via `/users/{slug}`.
+ *  2. Fallback to `/users?filter=...` and return slugs of active users with exact match on
+ *     email.
+ *
+ * @param userinfo - A string that could be the user's slug or email-address.
+ * @returns List of user slugs for active, matched users.
+ */
+async function getUserDetails(userinfo: string): Promise<string[]> {
```
viceice: move to new `user.ts` file to reduce complexity of this file.

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -680,17 +681,91 @@ export function addAssignees(iid: number, assignees: string[]): Promise<void> {
...
+
+    if (user.active) {
+      logger.debug({ userinfo }, 'Resolved user via slug match');
+      return [user.slug];
+    }
+    logger.debug({ userinfo }, 'Resolved inactive user');
+    return [];
+  } catch (err: any) {
+    logger.debug({ err, userinfo }, 'User lookup failed');
+  }
```
viceice: Can we skip this and only use the search below to reduce requests?
At least check if `userinfo` is an email like string and then skip this.
You can use zod[^1] to check for valid email

How likely this is a username instead of a email?

[^1]: https://v3.zod.dev/?id=strings

portly-halicore-76: @viceice Thanks for the review. Nice, was looking for a way to check if string is email, didn't found it.

I based this approach on the one found in GitLab implementation. Check if entry is user by the request, if not check if is email. 

We can skip the request to check if is user is found, and rely on https://github.com/renovatebot/renovate/pull/37241 entirely. Logging will let Renovate user know that they've added a not found username. Agreed?

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: I believe this will just increase complexity as this is tightly coupled to `BitbucketServerHttp` and we must create / export & import that in a new file. All request related code is in this file. Agreed?

Churro: There seems to be quite some repetition in tests, where only the reviewer value is different. Moving this to `user.ts` could help simplifying the tests and make it easier to cover all banches of `getUserDetails()`. It will also come in handy for `getUsersFromReviewerGroup()`.

Not verified but looking at `user.ts` in the `github` platform, the overhead of adding `BitbucketServerHttp` should be limited. Perhaps you may even want to extract ``./rest/api/1.0/users?filter...` into a separate method, since it needs project and repo slugs.

lib/modules/platform/bitbucket-server/index.ts
Churro: I agree that it would make sense to skip the `./rest/api/1.0/users/${userinfo}`  request for email addresses in `userinfo`, as this would always fail.

I ran some tests comparing response times for exact user matches vs. search with `filter=...`. Even if only one entry is returned, the search operation always takes ~2.5x to 3x times longer, so I'd argue that it's worth checking usernames / slugs via exact match only.

Churro (changes requested):

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: @Churro what I mean is not quite how you interpreted the question. I mean, don't try to figure out `/users/` at all in case we deal with anything other than an email. Just rely on error handling as will be implemented in #37241, as it functionally has ever been. 

Only if we know a userInfo **is an email** as allowed following CODEOWNERS implementation from BitBucket server, we do the filtered lookup. Therefore the func getUserDetails will only cover the `filter=...` logic as you suggested in another thread. In addition I would also state that we follow-up on @viceice suggestion, skip pagination and assume an emailadres will have at most 100 results (first page, also alligned with CODEOWNERS implementation in which at least 100 users are added) to reduce requests. 

I believe this would be the cleanest approach Agreed?

Churro: Ah, so in other words:
* resolve found email addresses to user slugs
* by default, try adding reviewers with given user slugs assuming they are all valid
* if adding reviewers failed, try again without invalid ones

Assuming there won't be more user slugs pointing to an email address than listed on the first page seems practical. However, IIRC, the `users` API endpoint returns 25 entries, not 100 😕 

Overall, sounds good 👍

portly-halicore-76: @Churro great thanks for pointers. Will find out how to request 100 results on the first page, assuming that's already baked in somewhere.

Churro: Hint: `/rest/api/1.0/users?filter=....&limit=100`, so `paginate: false, limit: 100` should work.

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: agreed but when tried, ran into lots of difficulties getting tests setup correctly. Because of setting up `Config` & api request. I gave up. The complexity was way bigger than adding a few tests here. If not agreable, please provide a boilerplate testfile setup to get this going 🙏

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: Implemented, thanks for the hint. Only adding `&limit `was enough without explicitely stating `paginate: false`.

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -686,11 +688,65 @@ export async function addReviewers(
...
+        'No users found for email-address',
+      );
+      return [];
+    }
+    return users.body
+      .filter(
+        (u: { active: any; emailAddress: string }) =>
+          u.active && u.emailAddress === emailAddress,
+      )
+      .map((u: { slug: any }) => u.slug);
```
Churro:
```suggestion
      .filter((u) => u.active && u.emailAddress === emailAddress)
      .map((u) => u.slug);
```

Types are already clear thanks to `UsersSchema`

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -686,11 +688,65 @@ export async function addReviewers(
...
+      `&permission.1.projectKey=${config.projectKey}` +
+      `&permission.1.repositorySlug=${config.repositorySlug}`;
+
+    const users = await bitbucketServerHttp.getJson(
+      filterUrl,
+      { limit: 100 },
+      UsersSchema,
+    );
+
+    if (!users.body.length) {
```
Churro: Code style suggestion:
* Invert the logic here -> `if (users.body.length) { return users.body.filter(....).map(...); }`
* Make this "fall through", so right before the end of the method put the `logger.debug(...)` and finally `return [];`

This way, the method gets a default `return` that doesn't depend on a nested logic in between

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -686,11 +688,65 @@ export async function addReviewers(
...
+      return [];
+    }
+    return users.body
+      .filter(
+        (u: { active: any; emailAddress: string }) =>
+          u.active && u.emailAddress === emailAddress,
+      )
+      .map((u: { slug: any }) => u.slug);
+  } catch (err) {
+    logger.warn({ err, emailAddress }, `Failed to resolve email to user`);
```
Churro: Nitpick:
```suggestion
    logger.warn({ err, emailAddress }, `Failed to resolve email address to user slug`);
```

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -75,7 +76,7 @@ export const id = 'bitbucket-server';
 
 let config: BbsConfig = {} as any;
 
-const bitbucketServerHttp = new BitbucketServerHttp();
+export const bitbucketServerHttp = new BitbucketServerHttp();
```
Churro: Not needed here, correct?

lib/modules/platform/bitbucket-server/index.spec.ts
```diff
@@ -1002,6 +1055,138 @@ describe('modules/platform/bitbucket-server/index', () => {
...
+            ]);
+
+          const actual = await bitbucket.getUserSlugsByEmail('e-mail@test.com');
+          expect(actual).toBeArrayOfSize(2);
+          expect(actual).toEqual(
+            expect.arrayContaining([
+              'usernamefoundbyemail',
+              'usernamefoundbyemailtoo',
+            ]),
+          );
```
Churro:
```suggestion
          expect(actual).toStrictEqual([
            'usernamefoundbyemail',
            'usernamefoundbyemailtoo',
          ]);
```

lib/modules/platform/bitbucket-server/index.spec.ts
```diff
@@ -1002,6 +1055,138 @@ describe('modules/platform/bitbucket-server/index', () => {
...
+                q.filter === 'e-mail@test.com' &&
+                q['permission.1'] === 'REPO_READ' &&
+                q['permission.1.repositorySlug'] === 'repo' &&
+                q['permission.1.projectKey'] === 'SOME',
+            )
+            .reply(500, []);
+
+          await expect(
+            bitbucket.getUserSlugsByEmail('e-mail@test.com'),
+          ).rejects.toThrowErrorMatchingSnapshot();
```
Churro: Better maintainable than snapshots (so please remove the snapshot changes too):

```suggestion
          ).rejects.toThrow('Response code 500 (Internal Server Error)');
```

Churro (changes requested):
Otherwise LGTM

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -686,11 +688,65 @@ export async function addReviewers(
 ): Promise<void> {
   logger.debug(`Adding reviewers '${reviewers.join(', ')}' to #${prNo}`);
 
-  await retry(updatePRAndAddReviewers, [prNo, reviewers], 3, [
+  const reviewerSlugs = new Set<string>();
+
+  for (const entry of reviewers) {
+    // If entry is an email-address, resolve userslugs
+    if (z.string().email().safeParse(entry).success) {
```
viceice: please create a email schema and add it to the Schema utils

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -686,11 +688,65 @@ export async function addReviewers(
...
+      const slugs = await getUserSlugsByEmail(entry);
+      slugs.forEach((slug) => reviewerSlugs.add(slug));
```
viceice: use for-of loop instead

Churro (approved):

rarkins (approved):

PR #37241, opened 2025-07-31
fix(bitbucket-server): Handle invalidReviewers correctly

portly-halicore-76:
## Changes

Just as is implemented in the updatePR function, there is now handling of invalidReviewers correctly in the updatePRAndAddReviewers function by removing the invalid ones and try again.

## Context

Needed as there can be users which have not got repo access or are not known in BitBucket Server. This now is not handled correctly, causing no-one to be added as a reviewer to the PR. 

Follow-up from https://github.com/renovatebot/renovate/pull/37199


lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -721,14 +730,17 @@ async function updatePRAndAddReviewers(
...
-    ) {
-      logger.debug(
-        '409 response to adding reviewers - has repository changed?',
-      );
-      throw new Error(REPOSITORY_CHANGED);
+    } else if (err.statusCode === 409) {
+      if (utils.isInvalidReviewersResponse(err) && !bitbucketInvalidReviewers) {
+        // Retry again with invalid reviewers being removed
+        const invalidReviewers = utils.getInvalidReviewers(err);
+        await updatePRAndAddReviewers(prNo, reviewers, invalidReviewers);
```
Churro: I think going with an approach like this would be more concise:

```suggestion
        const filteredReviewers = reviewers.filter(
          (name) => !invalidReviewers.includes(name)
        );
        await updatePRAndAddReviewers(prNo, filteredReviewers);
```

Churro (changes requested):
IMHO `fix` would be a better prefix than `chore`, as this PR actually addresses an inconsistency with how `updatePr()` behaves.

lib/modules/platform/bitbucket-server/index.ts
Churro: Sry, there was a copy-paste error in the suggestion: `updatePRAndAddReviewers` obviously needs to be passed `filteredReviewers` and not `reviewers` 😊

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -332,6 +332,7 @@ export async function getBranchForceRebase(
     res.body?.mergeConfig?.defaultStrategy?.id.includes('ff-only'),
   );
 }
+
```
RahulGautamSingh: revert

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -704,6 +706,13 @@ async function updatePRAndAddReviewers(
     // TODO: can `reviewers` be undefined? (#22198)
     const reviewersSet = new Set([...pr.reviewers!, ...reviewers]);
 
+    // Remove invalidReviewers from set if any
+    if (bitbucketInvalidReviewers?.length) {
+      for (const invalid of bitbucketInvalidReviewers) {
```
RahulGautamSingh:
```suggestion
      for (const invalidReviewer of bitbucketInvalidReviewers) {
```

& corresponding code change

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -704,6 +706,13 @@ async function updatePRAndAddReviewers(
...
+        reviewersSet.delete(invalid);
+      }
+    }
```
RahulGautamSingh: When will this be called?

As per the implementation, we only check for invalid reviewer after request fails. And when we are calling the function again after error we are not passing the invalid revewers list instead we filter the reviewers list there only.

Isn't this redundant then

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: Solved

lib/modules/platform/bitbucket-server/index.ts
portly-halicore-76: fixed, was a wrong suggestion commited. we now pass the invalid ones

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -704,6 +705,13 @@ async function updatePRAndAddReviewers(
     // TODO: can `reviewers` be undefined? (#22198)
     const reviewersSet = new Set([...pr.reviewers!, ...reviewers]);
 
+    // Remove invalidReviewers from set if any
```
Churro: My suggestion was probably misunderstood: If invalid reviewers are filtered before retrying, this code and the optional `bitbucketInvalidReviewers` is not needed anymore.

portly-halicore-76: Yes I overlooked that detail a few minutes ago. Why defer from the exact implementation, which is proven to be usefull in the `updatePR` func?

portly-halicore-76: Implemented, this is indeed better

portly-halicore-76: If wanted, can be implemented in seperate pr in `updatePr` as well.

Churro: Thx, a refactor PR for `updatePr` would be great 😊

Churro (approved):

viceice (approved):

PR #37281, opened 2025-08-01
feat(preset): add Apache Lucene monorepo

tigsikram:
## Changes

Add [Apache Lucene](https://github.com/apache/lucene) as monorepo. Its packages are released together with the same version.

### DIFF
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -671,6 +671,7 @@
   "patternGroups": {
     "angularmaterial": ["/^@angular/material/", "/^@angular/cdk/"],
     "apache-camel": "/^org.apache.camel:/",
+    "apache-lucene": "/^org.apache.lucene:/",
     "apache-poi": "/^org.apache.poi:/",
     "aws-java-sdk": "/^com.amazonaws:aws-java-sdk-/",
     "aws-java-sdk-v2": "/^software.amazon.awssdk:/",
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -17,6 +17,7 @@
     "angular-eslint": "https://github.com/angular-eslint/angular-eslint",
     "angularfire": "https://github.com/angular/angularfire",
     "angularjs": "https://github.com/angular/angular.js",
+    "apache-lucene": "https://gitbox.apache.org/repos/asf?p=lucene.git",
     "apollo-server": "https://github.com/apollographql/apollo-server",
     "apolloclient": "https://github.com/apollographql/apollo-client",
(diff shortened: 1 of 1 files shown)


lib/data/monorepo.json
```diff
@@ -671,6 +671,7 @@
   "patternGroups": {
     "angularmaterial": ["/^@angular/material/", "/^@angular/cdk/"],
     "apache-camel": "/^org.apache.camel:/",
+    "apache-lucene": "/^org.apache.lucene:/",
```
viceice: can't they be grouped by source url?

tigsikram: they can! i found `"matchPackageNames": ["/^org.apache.lucene:/"]` to look cleaner than `"matchSourceUrls": ["https://gitbox.apache.org/repos/asf?p=lucene.git"]`, but i'm happy either way. 545d7733bef0cc5d042cea828e484235d4b59b42

viceice (approved):

lib/data/monorepo.json
viceice: we prefer `matchSourceUrls` if possible

lib/data/monorepo.json
```diff
@@ -17,6 +17,7 @@
     "angular-eslint": "https://github.com/angular-eslint/angular-eslint",
     "angularfire": "https://github.com/angular/angularfire",
     "angularjs": "https://github.com/angular/angular.js",
+    "apache-lucene": "https://gitbox.apache.org/repos/asf?p=lucene.git",
```
viceice:
```suggestion
    "apache-lucene": [
      "https://gitbox.apache.org/repos/asf?p=lucene.git",
      "https://github.com/apache/lucene"
    ],
```

If they change it some day. The first url is a redirect to the github repo.

viceice (changes requested):
- https://github.com/apache/lucene/issues/15067

lib/data/monorepo.json
tigsikram: Excellent, thank you! I'll remember this for future redirects.
I suspect this might be a case of Apache wanting to stay in control of their URLs, if they ever switch from GitHub to another host.

rmuir:
Sorry for the trouble with our source urls (I'm a happy renovate user too), IMO, the github sourceUrl may be the best to use for the apache-lucene updates. 

We put release notes there too [example](https://github.com/apache/lucene/releases/tag/releases%2Flucene%2F10.2.2), but I'm not sure if they'll work in renovate PRs due to tag naming conventions. A workaround for any trouble there might be to set `changelogUrl` to something like `https://github.com/apache/lucene/blob/main/lucene/CHANGES.txt`

viceice (approved):

PR #37287, opened 2025-08-02
feat(managers): add Paket manager support

### LINKED ISSUE #11211: Add support for Paket Nuget package management

We use paket to resolve our C# nuget dependencies as it provides predictable package resolution similar to that of npm or yarn. A lock file is generated which contains the exact packages to resolve.

The ability to update packages managed through paket would be hugely beneficial to us as we currently have to update packages manually.

https://fsprojects.github.io/Paket/

Paket has a command for determining which packages are out of date: 

`paket outdated`
https://fsprojects.github.io/Paket/paket-outdated.html

and a further command for updating specific dependencies.
https://fsprojects.github.io/Paket/paket-update.html

No

fpellet:
## Changes

First implementation to support paket (https://github.com/fsprojects/Paket), a package manager opensource for dotnet.   
Not all features are supported. Ready:

- lock file maintenance
- update nuget packages

## Context

- closes #11211 


lib/modules/manager/paket/extract.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+            currentVersion: '2.5.0',
+            datasource: NugetDatasource.id,
+            rangeStrategy: 'update-lockfile',
+            lockedVersion: '2.5.0',
+          },
+        ],
+        lockFiles: ['/app/test/paket.lock'],
+      });
+    });
+  });
```
RahulGautamSingh, line 346: Maintain  new line after each `it` and `describe` block
```suggestion
  });
  
```

lib/modules/manager/paket/index.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+            rangeStrategy: 'update-lockfile',
+            lockedVersion: '2.5.0',
+          },
+        ],
+        lockFiles: ['/app/test/paket.lock'],
+      });
+    });
+  });
+  describe('updateArtifacts()', () => {
+    const command: UpdateArtifact = {
```
RahulGautamSingh: Use a better variable name

lib/modules/manager/paket/readme.md
```diff
@@ -0,0 +1,11 @@
...
+
+Extracts dependencies from files:
+
+- `paket.dependencies`
+- `paket.lock`
+
+Restrictions:
+
+- Support only nuget source (not github, gist, http, git)
+- Not support constraint on paket.dependencies
```
RahulGautamSingh:
```suggestion
- Constraint on paket.dependencies not supported
```

lib/modules/manager/paket/readme.md
```diff
@@ -0,0 +1,11 @@
+Manage dependencies for [Paket](https://fsprojects.github.io/Paket/)
+
+Extracts dependencies from files:
+
+- `paket.dependencies`
+- `paket.lock`
+
+Restrictions:
+
+- Support only nuget source (not github, gist, http, git)
```
RahulGautamSingh:
```suggestion
- Only updates nuget source (github, gist, http, git not supported)
```

lib/modules/manager/paket/index.ts
```diff
@@ -0,0 +1,125 @@
...
+  UpdateLockedResult,
+} from '../types';
+import { getAllPackages, updateAllPackages, updatePackage } from './tool';
+
+export const displayName = 'Paket';
+export const url = 'https://fsprojects.github.io/Paket/';
+export const categories: Category[] = ['dotnet'];
+
+export const defaultConfig = {
+  managerFilePatterns: ['/(^|/)paket\\.dependencies$/'],
```
RahulGautamSingh, line 18:
```suggestion
  managerFilePatterns: ['**paket.dependencies'],
```

We prefer glob

lib/modules/manager/paket/index.ts
```diff
@@ -0,0 +1,125 @@
...
+};
+
+export const supportedDatasources = [NugetDatasource.id];
+
+function getSiblingFileName(fileName: string, siblingName: string): string {
+  const subDirectory = getParentDir(fileName);
+  return upath.join(subDirectory, siblingName);
+}
```
RahulGautamSingh: We have a similar function in the util folder or somewhere in the codebase. Please use that.

lib/modules/manager/paket/tool.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export async function getAllPackages(
+  filePath: string,
+): Promise<{ name: string; version: string; group: string }[]> {
+  const execOptions: ExecOptions = {
+    cwdFile: filePath,
+  };
+  const result = await exec(
+    `paket show-installed-packages --silent`,
```
RahulGautamSingh: This will mean that Paket Manager needs to be installed. But that should not be necessary when only updating packages in the `packet.dependencies` file without touching the lockfile. 

Or, does the lockfile always need to be updated along with the package file?

Also, was this discussed formally? From the issue comments it seems you asked a question but I coudn't find any discussions around this. If I missed it please reference the discussion here as well

fpellet: The lockfile must be updated. The package versions are only in this file.
I was told that the issue was not the right place to discuss the implementation. A discussion took place on [Slack](https://renovatebot.slack.com/archives/CAFH752JU/p1752936424911909?thread_ts=1752921704.249159&cid=CAFH752JU).

Can you tell me what I should do? Because I'm not sure what I should ask for and where to do it. Not updating the lockfile is equivalent to refusing support for this package manager.

lib/modules/manager/paket/extract.spec.ts
fpellet: done

lib/modules/manager/paket/index.spec.ts
fpellet: done

lib/modules/manager/paket/index.ts
fpellet: done

lib/modules/manager/paket/tool.ts
```diff
@@ -0,0 +1,55 @@
...
+    execOptions,
+  );
+  return result.stdout
+    .split('\n')
```
RahulGautamSingh: import `newlineRegex` from `util/regex`

```suggestion
    .split(newlineRegex)
```

lib/modules/manager/paket/tool.ts
```diff
@@ -0,0 +1,55 @@
+import { quote } from 'shlex';
+import { exec } from '../../../util/exec';
+import type { ExecOptions } from '../../../util/exec/types';
+
+export interface UpdatePackage {
+  filePath: string;
+  packageName?: string;
+  group?: string;
+  version?: string;
+}
```
RahulGautamSingh:
```suggestion
}

```

lib/modules/manager/paket/index.ts
```diff
@@ -0,0 +1,119 @@
...
+  });
+
+  const lockFileName = getSiblingFileName(packageFile, 'paket.lock');
+  return {
+    deps,
+    lockFiles: [lockFileName],
+  };
+}
+
+export const supportsLockFileMaintenance = true;
```
RahulGautamSingh: Move this to top

lib/modules/manager/paket/tool.ts
```diff
@@ -0,0 +1,55 @@
...
+export async function updatePackage(command: UpdatePackage): Promise<void> {
+  const execOptions: ExecOptions = {
+    cwdFile: command.filePath,
+  };
+  const groupFilter = command.group ? ` --group ${quote(command.group)}` : '';
```
RahulGautamSingh:
```suggestion
  const groupFilter = command.group ? ` --group ${quote(command.group)} ` : '';
```

Same whitespace at end for the package and version string too

RahulGautamSingh (changes requested):
Can you please also test this on a test_repo and link it here.

Cases to test:
1. single dep gets updated
2. multiple deps get updated

fpellet:
> Can you please also test this on a test_repo and link it here.
> 
> Cases to test:
> 
>     1. single dep gets updated
> 
>     2. multiple deps get updated

I tested both cases manually (you can see the MRs).
Here is the repo: https://github.com/fpellet/renovate-paket
There are two dependencies to update (Xunit and Fake).

I just modified it to create two directories, one with a single dependency and one with two dependencies.
It's ok for you?

lib/modules/manager/paket/tool.ts
fpellet: done

lib/modules/manager/paket/index.ts
fpellet: done

lib/modules/manager/paket/tool.ts
fpellet: done, but I don't understand why. This is a deviation from the normal use of the command line (several spaces between args).

lib/modules/manager/paket/extract.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+  updateArtifacts,
+  updateLockedDependency,
+} from './index';
+import { fs, git } from '~test/util';
+
+describe('modules/manager/paket/index', () => {
+  const packageFileName = '/app/test/paket.dependencies';
+
+  beforeEach(() => {
+    vi.mock('../../../util/fs');
```
viceice: wrong, move to module scope. vitest will hoist it and it's called only once

lib/modules/manager/paket/tool.ts
viceice: we don't accept managers which are using external tools to extract dependencies. so you need to extract those yourself.

@zharinov the lockfile seems simple (indention based like yaml), can you suggest a parser?

lib/modules/manager/paket/tool.ts
```diff
@@ -0,0 +1,59 @@
...
+    execOptions,
+  );
+}
+
+export async function updateAllPackages(
+  filePath: string,
+  group?: string,
+): Promise<void> {
+  await updatePackage({ filePath, group });
+}
```
viceice: inline the `updatePackage`, there is no reason to have such small wrapper functions which effectively only change flat args to an object arg

lib/modules/manager/paket/tool.ts
```diff
@@ -0,0 +1,59 @@
...
+  };
+  const groupFilter = command.group ? ` --group ${quote(command.group)} ` : '';
+  const packageFilter = command.packageName
+    ? ` ${quote(command.packageName)} `
+    : '';
+  const version = command.version
+    ? ` --version ${quote(command.version)} `
+    : '';
+  await exec(
+    `paket update${groupFilter}${version}${packageFilter}`,
```
viceice: multiple update calls should be done in one go to reduce execution time. `exec` can handle an array of string as commands executed sequentially

viceice (changes requested):
blocked by
- https://github.com/containerbase/base/issues/4938

lib/modules/manager/paket/tool.ts
fpellet: This adds semantics. Otherwise, you need to understand the implementation of updatePackage to know that you have to give a specific parameter to update all packages.
But I can adapt and remove it if you prefer.

lib/modules/manager/paket/tool.ts
fpellet: I don't understand. This method is called only once, not multiple times via index.ts. There is no loop

lib/modules/manager/paket/tool.ts
zharinov: > can you suggest a parser?

Yes, manually split by line and count indents

purkhusid:
What would it take to get this PR back on track?

RahulGautamSingh:
> What would it take to get this PR back on track?

> we don't accept managers which are using external tools to extract dependencies. so you need to extract those yourself.

This ☝🏻 and other suggested changes

lib/modules/manager/paket/extract.spec.ts
fpellet: done

fpellet:
https://github.com/renovatebot/renovate/pull/37287#discussion_r2318556021 Fixed

lib/modules/manager/paket/__fixtures__/.gitignore
RahulGautamSingh, line 1: Why is this needed?

lib/modules/manager/paket/readme.md
```diff
@@ -0,0 +1,11 @@
+Manage dependencies for [Paket](https://fsprojects.github.io/Paket/)
+
+Extracts dependencies from files:
+
+- `paket.dependencies`
+- `paket.lock`
+
+Restrictions:
+
+- Only updates nuget source (github, gist, http, git not supported)
```
RahulGautamSingh:
```suggestion
- Only updates nuget source (github, gist, http, and git not supported)
```

lib/modules/manager/paket/index.ts
```diff
@@ -0,0 +1,141 @@
...
+  packageFile: string,
+  _config: ExtractConfig,
+): Promise<PackageFileContent<PaketPackage>> {
+  logger.debug(`paket.extractPackageFile(${packageFile})`);
+
+  const lockFileName = getSiblingFileName(packageFile, 'paket.lock');
+  const lockFileContentMap = await getFiles([lockFileName]);
+  const lockFileContent = lockFileContentMap[lockFileName];
+  if (!lockFileContent) {
+    throw new Error(`Impossible to find paket lock file: ${lockFileName}`);
```
RahulGautamSingh:
```suggestion
    throw new Error(`Could not to paket lock file: ${lockFileName}`);
```

lib/modules/manager/paket/parsers/dependencies-file.ts
```diff
@@ -0,0 +1,52 @@
...
+  name: string;
+  options: string[];
+}
+interface ReduceState {
+  groups: DependenciesGroup[];
+  currentGroup: DependenciesGroup;
+}
+
+export function parse(content: string): DependenciesFile {
+  const result: ReduceState = content.split('\n').reduce(
```
RahulGautamSingh: Please convert this to a for...loop

lib/modules/manager/paket/parsers/dependencies-file.ts
```diff
@@ -0,0 +1,52 @@
...
+    (state: ReduceState, line: string) => {
+      const lineParts = line.split(/\s+/).filter((p) => p !== '');
```
RahulGautamSingh: And import `isNonEmptyString` from `@sindresorhus/is`
```suggestion
      const lineParts = line.split(/\s+/).filter(isNonEmptyString);
```

lib/modules/manager/paket/index.ts
```diff
@@ -0,0 +1,141 @@
...
+            d.groupName === group.groupName &&
+            d.packageName.toUpperCase() === p.name.toUpperCase(),
+        );
+
+        const version = lockVersion?.version;
+        const name = lockVersion?.packageName ?? p.name;
+        return {
+          depType: 'dependencies',
+          depName: name,
+          packageName: name,
```
RahulGautamSingh:
```suggestion
```

if they are the same no need to repeat, existing code handles it

RahulGautamSingh (changes requested):

lib/modules/manager/paket/__fixtures__/.gitignore
fpellet: To generate the .lock file from paket.dependencies, I run the paket command, which will download the packages and create two directories in addition to the .lock file.  
I could delete it now, but it seems useful for facilitating future updates.

lib/modules/manager/paket/index.ts
fpellet: done

lib/modules/manager/paket/parsers/dependencies-file.ts
fpellet: done

lib/modules/manager/paket/parsers/dependencies-file.ts
fpellet: done

lib/modules/manager/paket/parsers/dependencies-file.ts
```diff
@@ -0,0 +1,61 @@
+import { isNonEmptyString } from '@sindresorhus/is';
+import { newlineRegex } from '../../../../util/regex';
+
+interface DependenciesFile {
```
RahulGautamSingh: Move interfaces to a separate `types.ts` file

lib/modules/manager/paket/parsers/lock-file.ts
```diff
@@ -0,0 +1,99 @@
+import { newlineRegex } from '../../../../util/regex';
+
+export type SourceType = 'nuget';
+export interface Dependency {
```
RahulGautamSingh: move interfaces to a `types.ts` file

lib/modules/manager/paket/parsers/lock-file.ts
```diff
@@ -0,0 +1,99 @@
...
+  text: string;
+  indent: number;
+}
+
+function extractLines(content: string): Line[] {
+  return content
+    .split(newlineRegex)
+    .map((line) => ({
+      text: line.trim(),
+      indent: /^( *)/.exec(line)![1].length,
```
RahulGautamSingh:
```suggestion
      indent: regEx(/^(\s*)/).exec(line)![1].length,
```

lib/modules/manager/paket/parsers/lock-file.ts
```diff
@@ -0,0 +1,99 @@
...
+    }))
+    .filter((line) => line.text !== '');
```
RahulGautamSingh:
```suggestion
    .filter((line) => isNonEmptyString(line.text));
```

lib/modules/manager/paket/parsers/lock-file.ts
```diff
@@ -0,0 +1,99 @@
...
+    currentGroupName: string;
+    currentRemote: string;
+  }
+  const initialState: State = {
+    dependencies: [],
+    currentSource: null,
+    currentGroupName: 'Main',
+    currentRemote: '',
+  };
+  const finalState = lines.reduce((state, line): State => {
```
RahulGautamSingh: reduce -> for...of loops

RahulGautamSingh (changes requested):

fpellet:
https://github.com/containerbase/base/issues/4938 is merged. I rebased with the latest version. It's possible to merge now this PR ?

lib/modules/manager/paket/extract.spec.ts
viceice: it's still here, move to module scope, vitest will hoist it anyways, so this will confuse because it's not executed in this context

lib/modules/manager/paket/index.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+        expect(siblingName).equals('paket.lock');
+        return '/app/test/paket.lock';
+      },
+    );
+  }
+
+  describe('extractPackageFile()', () => {
+    beforeEach(() => {
+      initializeMock();
+    });
```
viceice: move on level up, so you only need one before each

lib/modules/manager/paket/index.ts
```diff
@@ -0,0 +1,139 @@
+import type { Category } from '../../../constants/index.ts';
```
viceice, line 1: split this file into multiple files: index, extract, artifacs

lib/modules/manager/paket/tool.ts
viceice: please remove, add proper tsdoc to describe what should be passed

viceice (commented):
does Paket executing any code from the repo?

eg like local plugin which could execute arbitrary code while restoring dependencies?

lib/modules/manager/paket/extract.spec.ts
fpellet: sorry, it's done now

lib/modules/manager/paket/tool.ts
fpellet: done

lib/modules/manager/paket/index.spec.ts
fpellet: done

lib/modules/manager/paket/index.ts
fpellet: done (index, extract, artifacs and update-lock)

fpellet:
> does Paket executing any code from the repo?
> 
> eg like local plugin which could execute arbitrary code while restoring dependencies?

Paket is only run to update the lock file. It does not modify other versioned files.
It isn't used in other situations, such as reading dependency files (including lock files).
However, it does download dependencies.
The operation is similar to npm/pnpm, but it does not affect dependency files other than the lock file (versions are only stored in the lock file).

docs/usage/configuration-options.md
RahulGautamSingh, line 1: this can be dropped, this is list os now auto-generated..

lib/modules/manager/paket/parsers/dependencies-file.ts
```diff
@@ -0,0 +1,48 @@
+import { isNonEmptyString } from '@sindresorhus/is';
+import { newlineRegex } from '../../../../util/regex.ts';
+import type { DependenciesFile, DependenciesFileGroup } from '../types.ts';
+
+interface ReduceState {
+  groups: DependenciesFileGroup[];
+  currentGroup: DependenciesFileGroup;
+}
+function analyzeLine(state: ReduceState, line: string): ReduceState {
+  const lineParts = line.split(/\s+/).filter(isNonEmptyString);
```
RahulGautamSingh:
```suggestion
  const lineParts = line.split(regEx(/\s+/)).filter(isNonEmptyString);
```

lib/modules/manager/paket/extract.ts
```diff
@@ -0,0 +1,58 @@
...
+  _config: ExtractConfig,
+): Promise<PackageFileContent<PaketPackage>> {
+  logger.debug(`paket.extractPackageFile(${packageFile})`);
+
+  const lockFileName = getSiblingFileName(packageFile, 'paket.lock');
+  const lockFileContentMap = await getFiles([lockFileName]);
+  const lockFileContent = lockFileContentMap[lockFileName];
+  if (!lockFileContent) {
+    throw new Error(`Could not to paket lock file: ${lockFileName}`);
+  }
```
viceice: why using `getFiles` instead of a simple `fs.eadLocalFile(lockfileName)`

lib/modules/manager/paket/extract.ts
```diff
@@ -0,0 +1,58 @@
...
+        const name = lockVersion?.packageName ?? p.name;
+        return {
+          depType: 'dependencies',
+          depName: name,
+          currentVersion: version,
+          datasource: NugetDatasource.id,
+          rangeStrategy: 'update-lockfile',
+          lockedVersion: version,
+        };
+      });
```
viceice: please extract a module function for better readabillity

lib/modules/manager/paket/tool.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+          '  - xunit.analyzers: 1.23.0 (added)\n' +
+          '  - xunit.assert: 2.9.3 (added)\n' +
+          '  - xunit.core: 2.9.3 (added)\n' +
+          '  - xunit.extensibility.core: 2.9.3 (added)\n' +
+          '  - xunit.extensibility.execution: 2.9.3 (added)\n' +
+          '  Installing into projects:\n' +
+          '    Created dependency graph (29 packages in total)\n' +
+          '  Total time taken: 6 seconds\n',
+      });
+      GlobalConfig.set(adminConfig);
```
viceice: move to `beforeAll` function, you never change the options anyways

lib/modules/manager/paket/tool.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+import type { ExecResult } from '~test/exec-util.ts';
+import { exec } from '~test/exec-util.ts';
+import { GlobalConfig } from '../../../config/global.ts';
+import type { RepoGlobalConfig } from '../../../config/types.ts';
+import type { CommandWithOptions } from '../../../util/exec/types.ts';
+import { runPaketUpdate } from './tool.ts';
+import type { UpdatePackage } from './types.ts';
+
+const defaultExecResult = { stdout: '', stderr: '' };
+export function mockExecAll(
```
viceice: we already have such a function, reuse

lib/modules/manager/paket/artifacts.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function updateArtifacts(
+  updateArtifact: UpdateArtifact,
+): Promise<UpdateArtifactsResult[] | null> {
+  logger.debug(`paket.updateArtifacts(${updateArtifact.packageFileName})`);
+
+  const lockFileName = getSiblingFileName(
+    updateArtifact.packageFileName,
+    'paket.lock',
+  );
+  const existingLockFileContentMap = await getFiles([lockFileName]);
```
viceice: why using `getFiles` from git instead of `readLocalFile` from fs?

lib/modules/manager/paket/artifacts.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  await runPaketUpdate({ filePath: lockFileName });
```
viceice: add a try-catch and return artifact error on error

lib/modules/manager/paket/extract.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+      (fileName: string, siblingName: string) => {
+        expect(fileName).equals(packageFileName);
+        expect(siblingName).equals('paket.lock');
+        return '/app/test/paket.lock';
+      },
+    );
+  }
+  beforeEach(() => {
+    initializeMock();
+  });
```
viceice:
```suggestion
  beforeEach(() => {
    fs.getSiblingFileName.mockImplementation(
      (fileName: string, siblingName: string) => {
        expect(fileName).equals(packageFileName);
        expect(siblingName).equals('paket.lock');
        return '/app/test/paket.lock';
      },
    );
  });
```

no need for a wrapper function if only used once

lib/modules/manager/paket/update-lock.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+      (fileName: string, siblingName: string) => {
+        expect(fileName).equals(packageFileName);
+        expect(siblingName).equals('paket.lock');
+        return '/app/test/paket.lock';
+      },
+    );
+  }
+  beforeEach(() => {
+    initializeMock();
+  });
```
viceice:
```suggestion
  beforeEach(() => {
    fs.getSiblingFileName.mockImplementation(
      (fileName: string, siblingName: string) => {
        expect(fileName).equals(packageFileName);
        expect(siblingName).equals('paket.lock');
        return '/app/test/paket.lock';
      },
    );
  });
```

lib/modules/manager/paket/extract.ts
fpellet: I had copied from Nuget. It has changed.

lib/modules/manager/paket/extract.ts
fpellet: done

lib/modules/manager/paket/extract.spec.ts
fpellet: ok, I have replace expect by a if. It's not allow in beforeEach

lib/modules/manager/paket/update-lock.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+import * as tool from './tool.ts';
+import { updateLockedDependency } from './update-lock.ts';
+
+vi.mock('../../../util/fs');
+
+describe('modules/manager/paket/update-lock', () => {
+  const packageFileName = '/app/test/paket.dependencies';
+
+  beforeEach(() => {
+    fs.getSiblingFileName.mockImplementation(
```
RahulGautamSingh: Is it necessary to do it `beforeEach`?

lib/modules/manager/paket/tool.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    ])(
+      'can specify parameters (%o)',
+      async (command: UpdatePackage, expected) => {
+        const execSnapshots = mockExecAll();
+
+        await runPaketUpdate(command);
+
+        expect(execSnapshots.map((s) => s.cmd)).toEqual([expected]);
+      },
+    );
```
RahulGautamSingh, line 54: maintain one line space between tests

lib/modules/manager/paket/tool.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+          '  Installing into projects:\n' +
+          '    Created dependency graph (29 packages in total)\n' +
+          '  Total time taken: 6 seconds\n',
+      });
+
+      await runPaketUpdate({ filePath: packageFilePath });
+
+      expect(execSnapshots.map((s) => s.cmd)).toEqual(['paket update']);
+    });
+    test.each([
```
RahulGautamSingh: use `it` or `test` don't mix both in single test file

lib/modules/manager/paket/extract.ts
```diff
@@ -0,0 +1,86 @@
...
+export async function extractPackageFile(
+  content: string,
+  packageFile: string,
+  _config: ExtractConfig,
+): Promise<PackageFileContent<PaketPackage>> {
+  logger.debug(`paket.extractPackageFile(${packageFile})`);
+
+  const lockFileName = getSiblingFileName(packageFile, 'paket.lock');
+  const lockFileContent = await readLocalFile(lockFileName, 'utf8');
+  if (!lockFileContent) {
```
RahulGautamSingh, line 72: Is lock file always necessary? Just confirming..

lib/modules/manager/paket/extract.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+          throw new Error(`Not expected siblingName: ${siblingName}`);
+        }
+        return '/app/test/paket.lock';
+      },
+    );
+  });
+
+  describe('extractPackageFile()', () => {
+    const config: ExtractConfig = {};
+    const packageFileContent = `
```
RahulGautamSingh: use `codeBlock` utility

RahulGautamSingh (changes requested):
Full review complete.

Please apply the change at all places where applicable.

Logically everything seems fine, the requested changes are mostly patterns we prefer for development.

lib/modules/manager/paket/extract.ts
fpellet: Yes, paket necessarily works with a lock file.

lib/modules/manager/paket/update-lock.spec.ts
fpellet: good catch. it's deleted

lib/modules/manager/paket/tool.spec.ts
fpellet: done

lib/modules/manager/paket/tool.spec.ts
fpellet: done

lib/modules/manager/paket/extract.spec.ts
fpellet: done

lib/modules/manager/paket/parsers/dependencies-file.spec.ts
```diff
@@ -4,14 +4,14 @@ import { parse } from './dependencies-file.ts';
 describe('modules/manager/paket/parsers/dependencies-file', () => {
   describe('parse()', () => {
     it('should return all packages on main group', () => {
-      const result = parse(codeBlock`
-source https://api.nuget.org/v3/index.json
+      const result = parse(`
```
RahulGautamSingh: codeBlock.

lib/modules/manager/paket/artifacts.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+import { updateArtifacts } from './artifacts.ts';
+import * as tool from './tool.ts';
+
+vi.mock('../../../util/fs');
+
+describe('modules/manager/paket/artifacts', () => {
+  const packageFileName = '/app/test/paket.dependencies';
+
+  beforeEach(() => {
+    fs.getSiblingFileName.mockImplementation(
```
RahulGautamSingh, line 14: same with this, not needed in beforeEach

lib/modules/manager/paket/extract.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+import type { ExtractConfig } from '../types.ts';
+import { extractPackageFile } from './extract.ts';
+
+vi.mock('../../../util/fs');
+
+describe('modules/manager/paket/extract', () => {
+  const packageFileName = '/app/test/paket.dependencies';
+
+  beforeEach(() => {
+    fs.getSiblingFileName.mockImplementation(
```
RahulGautamSingh, line 13: same

RahulGautamSingh (changes requested):

lib/modules/manager/paket/extract.spec.ts
fpellet: getSiblingFileName is used by the implementation. Why is it useless?

lib/modules/manager/paket/artifacts.spec.ts
fpellet: getSiblingFileName is used by the implementation. Why is it useless?

lib/modules/manager/paket/extract.spec.ts
RahulGautamSingh: useless to repeatedly call It before each test, which is what beforeEach does.

Just inititlaize this mock implementation once that should be okay

fpellet: Unless I am mistaken, mocks are reset with each test. I cannot replace them with a beforeAll

RahulGautamSingh: You are right. Keep it as is.

RahulGautamSingh (approved):
LGTM

lib/modules/manager/paket/parsers/dependencies-file.spec.ts
```diff
@@ -0,0 +1,69 @@
+import { codeBlock } from 'common-tags';
+import { parse } from './dependencies-file.ts';
+
+describe('modules/manager/paket/parsers/dependencies-file', () => {
+  describe('parse()', () => {
+    it('should return all packages on main group', () => {
+      const result = parse(codeBlock`
+      source https://api.nuget.org/v3/index.json
```
viceice: can you indent with two more spaces?

lib/modules/manager/paket/parsers/dependencies-file.ts
```diff
@@ -0,0 +1,48 @@
+import { isNonEmptyString } from '@sindresorhus/is';
+import { newlineRegex, regEx } from '../../../../util/regex.ts';
+import type { DependenciesFile, DependenciesFileGroup } from '../types.ts';
+
+interface ReduceState {
+  groups: DependenciesFileGroup[];
+  currentGroup: DependenciesFileGroup;
+}
+function analyzeLine(state: ReduceState, line: string): ReduceState {
```
viceice:
```suggestion

function analyzeLine(state: ReduceState, line: string): ReduceState {
```

lib/modules/manager/paket/parsers/dependencies-file.ts
```diff
@@ -0,0 +1,48 @@
...
+  const lineParts = line.split(regEx(/\s+/)).filter(isNonEmptyString);
```
viceice:
```suggestion
  const lineParts = line.split(regEx(/\s+/)).filter(isNonEmptyStringAndNotWhitespace);
```

lib/modules/manager/paket/parsers/lock-file.spec.ts
```diff
@@ -0,0 +1,138 @@
+import { codeBlock } from 'common-tags';
+import type { LockFileDependency } from '../types.ts';
+import { parse } from './lock-file.ts';
+
+describe('modules/manager/paket/parsers/lock-file', () => {
+  describe('parse()', () => {
+    it('should return all nuget package version on main group', () => {
+      const result = parse(codeBlock`
+      RESTRICTION: == net8.0
```
viceice: indent

lib/modules/manager/paket/artifacts.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+            return Promise.resolve('Old fake lock file content');
+          } else {
+            return Promise.resolve(newContentLockFile);
+          }
+        },
+      );
+
+      const result = await updateArtifacts(updateArtifact);
+
+      expect(toolSpy.mock.calls).toEqual([[{ filePath: lockFileName }]]);
```
viceice: use `toHaveBeenCalledWith` and alike

lib/modules/manager/paket/artifacts.ts
```diff
@@ -0,0 +1,47 @@
...
+  logger.debug(`paket.updateArtifacts(${updateArtifact.packageFileName})`);
+
+  const lockFileName = getSiblingFileName(
+    updateArtifact.packageFileName,
+    'paket.lock',
+  );
+  const existingLockFileContent = await readLocalFile(lockFileName);
+
+  try {
+    await runPaketUpdate({ filePath: lockFileName });
```
viceice: delete lockfile before when `updateArtifact.isLockfileMaintenance===true`

lib/modules/manager/paket/extract.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+        }
+        return '/app/test/paket.lock';
+      },
+    );
+  });
+
+  describe('extractPackageFile()', () => {
+    const config: ExtractConfig = {};
+    const packageFileContent = codeBlock`
+    source https://api.nuget.org/v3/index.json
```
viceice: indent with two more spaces

lib/modules/manager/paket/tool.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Call the paket cli to update dependencies by refreshing the lock file.
+ * The packages will be downloaded (normally not committed on git), and the lock file will be updated.
+ * Other versioned files are not affected.
+ * https://fsprojects.github.io/Paket/paket-update.html
+ *
+ * @param command - parameters of paket cli
+ */
+export async function runPaketUpdate(command: UpdatePackage): Promise<void> {
+  const execOptions: ExecOptions = {
+    cwdFile: command.filePath,
```
viceice: you need to add toolConstraints to install dotnet and paket. see other managers

https://github.com/renovatebot/renovate/blob/321d121e680e5e722e3139af8c99a0791ba302c5/lib/util/exec/containerbase.ts#L21

https://github.com/renovatebot/renovate/blob/321d121e680e5e722e3139af8c99a0791ba302c5/lib/modules/manager/bundler/artifacts.ts#L162-L172

lib/modules/manager/paket/artifacts.ts
romerod: As I'm looking forward to this feature:
From the documentation https://fsprojects.github.io/Paket/paket-update.html#Updating-all-packages "paket update" will delete the lock file before updating. So from my POV this handler updateArtifact.isLockfileMaintenance===true, what would be the expected behavior when IsLockfileMaintenance is false?
Wouldn't IsLockfileMaintenance===false not go thru updateLockedDependency?

fpellet: @viceice It is not necessary to delete the lock file beforehand. Doing so will not improve performance. On the contrary, it actually disables certain security measures, such as locking the package's data source.

lib/modules/manager/paket/tool.ts
fpellet: Is that okay with you? Did I forget anything?

purkhusid:
Any chance we could get this over the finish line? Really looking forward to this addition. Thanks for the hard work @fpellet

fpellet:
> Any chance we could get this over the finish line? Really looking forward to this addition. Thanks for the hard work @fpellet

For me, everything looks good. I'm waiting for a second contributor to approve it, specifically @viceice

purkhusid:
@viceice A friendly ping. Would be great to have Paket available in Renovate.

viceice (commented):
please fix ci

fpellet:
> please fix ci

@viceice done

lib/modules/manager/paket/index.ts
```diff
@@ -0,0 +1,18 @@
...
+export { updateLockedDependency } from './update-lock.ts';
+
+export const displayName = 'Paket';
+export const url = 'https://fsprojects.github.io/Paket/';
+export const categories: Category[] = ['dotnet'];
+export const supportsLockFileMaintenance = true;
+export const lockFileNames = ['paket.lock'];
+
+export const defaultConfig = {
+  managerFilePatterns: ['**paket.dependencies'],
```
RahulGautamSingh:
```suggestion
  managerFilePatterns: ['**/paket.dependencies'],
```

lib/modules/manager/paket/extract.ts
```diff
@@ -0,0 +1,86 @@
...
+} from './types.ts';
+
+const searchPackageVersion = (
+  dependencies: LockFileDependency[],
+  group: DependenciesFileGroup,
+  p: DependenciesFilePackage,
+): LockFileDependency | undefined =>
+  dependencies.find(
+    (d) =>
+      d.groupName === group.groupName &&
```
RahulGautamSingh: why no upperCase conversion here?

lib/modules/manager/paket/extract.ts
```diff
@@ -0,0 +1,86 @@
...
+  content: string,
+  packageFile: string,
+  _config: ExtractConfig,
+): Promise<PackageFileContent<PaketPackage>> {
+  logger.debug(`paket.extractPackageFile(${packageFile})`);
+
+  const lockFileName = getSiblingFileName(packageFile, 'paket.lock');
+  const lockFileContent = await readLocalFile(lockFileName, 'utf8');
+  if (!lockFileContent) {
+    throw new Error(`Could not to paket lock file: ${lockFileName}`);
```
RahulGautamSingh: typo in the err message

lib/modules/manager/paket/parsers/lock-file.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  if (line.indent === 2 && line.text.startsWith('remote:')) {
+    return {
+      ...state,
+      currentRemote: line.text.replace('remote: ', ''),
+    };
+  }
+
+  if (line.indent === 4) {
+    const packageMatch = /^(\S+)\s+\(([^)]+)\)/.exec(line.text);
```
RahulGautamSingh: use named groups not positional indexes.

lib/modules/manager/paket/extract.ts
```diff
@@ -0,0 +1,86 @@
...
+import { parse as parseLockFile } from './parsers/lock-file.ts';
+import type {
+  DependenciesFile,
+  DependenciesFileGroup,
+  DependenciesFilePackage,
+  LockFileDependency,
+  PaketPackage,
+} from './types.ts';
+
+const searchPackageVersion = (
```
RahulGautamSingh: no arrow const, use function declaration

lib/modules/manager/paket/update-lock.ts
```diff
@@ -0,0 +1,30 @@
+import { logger } from '../../../logger/index.ts';
+import { readLocalFile } from '../../../util/fs/index.ts';
+import type { UpdateLockedConfig, UpdateLockedResult } from '../types.ts';
+import { runPaketUpdate } from './tool.ts';
+
+export async function updateLockedDependency(
+  config: UpdateLockedConfig,
+): Promise<UpdateLockedResult> {
+  logger.debug(`paket.updateLockedDependency(${config.lockFile}})`);
```
RahulGautamSingh:
```suggestion
  logger.debug(`paket.updateLockedDependency(${config.lockFile})`);
```

lib/modules/manager/paket/update-lock.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    };
+
+    it('update all packages', async () => {
+      const toolSpy = vi.spyOn(tool, 'runPaketUpdate');
+      toolSpy.mockResolvedValue();
+
+      const newContentLockFile = 'New fake lock file content';
+      fs.readLocalFile.mockImplementation(
+        (filename: string, _encoding: 'utf8') => {
+          expect(filename).equals(lockFileName);
```
RahulGautamSingh: use toBe or toEqual. rest of the codebase uses these.

lib/modules/manager/paket/artifacts.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+          return Promise.resolve('Old fake lock file content');
+        },
+      );
+
+      const result = await updateArtifacts(updateArtifact);
+
+      expect(result).toBeNull();
+    });
+
+    it('return artefact error if cmd failed', async () => {
```
RahulGautamSingh:
```suggestion
    it('return artifact error if cmd failed', async () => {
```

lib/modules/manager/paket/types.ts
```diff
@@ -0,0 +1,34 @@
...
+export type LockFileSourceType = 'nuget';
+export interface LockFileDependency {
+  source: LockFileSourceType;
+  groupName: string;
+  remote: string;
+  packageName: string;
+  version: string;
+}
+
+export interface PaketPackage {
```
RahulGautamSingh: where is this type used?

RahulGautamSingh (changes requested):

lib/modules/manager/paket/types.ts
fpellet: Actually, it's for a future improvement, but that has nothing to do with this

lib/modules/manager/paket/artifacts.ts
```diff
@@ -0,0 +1,59 @@
...
+        {
+          toolName: 'paket',
+          constraint: updateArtifact.config.constraints?.paket,
+        },
+      ],
+    });
+
+    const newLockFileContent = await readLocalFile(lockFileName);
+
+    if (existingLockFileContent === newLockFileContent) {
```
RahulGautamSingh, line 93: two buffer comparisons will always be false, right? unless `readLocalFile` caches the result internally.

lib/modules/manager/paket/extract.ts
```diff
@@ -0,0 +1,86 @@
...
+  content: string,
+  packageFile: string,
+  _config: ExtractConfig,
+): Promise<PackageFileContent> {
+  logger.debug(`paket.extractPackageFile(${packageFile})`);
+
+  const lockFileName = getSiblingFileName(packageFile, 'paket.lock');
+  const lockFileContent = await readLocalFile(lockFileName, 'utf8');
+  if (!lockFileContent) {
+    throw new Error(`Could not find paket lock file: ${lockFileName}`);
```
RahulGautamSingh: return null and log a warning instead.

lib/modules/manager/paket/parsers/lock-file.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  if (line.indent === 2 && line.text.startsWith('remote:')) {
+    return {
+      ...state,
+      currentRemote: line.text.replace('remote: ', ''),
+    };
+  }
+
+  if (line.indent === 4) {
+    const packageMatch = /^(?<packageName>\S+)\s+\((?<version>[^)]+)\)/.exec(
```
RahulGautamSingh: regEx wrapper

RahulGautamSingh (changes requested):

RahulGautamSingh (commented):
lgtm

purkhusid:
Any chance that this could get into main before it lags too much behind again?

lib/util/exec/containerbase.ts
```diff
@@ -156,6 +156,11 @@ export const allToolConfig: Record<ToolName, ToolConfig> = {
     packageName: 'npm',
     versioning: 'npm',
   },
+  paket: {
+    datasource: 'nuget',
+    packageName: 'paket',
+    versioning: 'semver',
```
viceice, line 162: why not using `nuget` versioning here?

@secustor @jamietanna `nuget` versioning will not hit this block because it treats `1.2.3` as range.

https://github.com/renovatebot/renovate/blob/1540a34fcf45598ca2b9461e0e194f4259a9281b/lib/util/exec/containerbase.ts#L313-L325

lib/modules/manager/paket/update-lock.ts
```diff
@@ -0,0 +1,30 @@
...
+): Promise<UpdateLockedResult> {
+  logger.debug(`paket.updateLockedDependency(${config.lockFile})`);
+
+  const existingLockFileContent = await readLocalFile(config.lockFile, 'utf8');
+
+  await runPaketUpdate({
+    filePath: config.lockFile,
+    packageName: config.depName,
+    version: config.newVersion,
+    toolConstraints: [{ toolName: 'dotnet' }, { toolName: 'paket' }],
```
viceice: No contraints in `UpdateLockedConfig` ? we should pass it. I wonder if we call tools in other managers in `updateLockedDependency` 🤔 I thougth we only do that in `updateArtifacts`. 🤔 

please see cargo manager, we don't need to call any tool here. it can all be handled by `updateArtifacts`

lib/modules/manager/paket/parsers/dependencies-file.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+describe('modules/manager/paket/parsers/dependencies-file', () => {
+  describe('parse()', () => {
+    it('should return all packages on main group', () => {
+      const result = parse(codeBlock`
+        source https://api.nuget.org/v3/index.json
+
+        framework: net8.0
+
+        nuget Fsharp.Core
```
viceice, line 12: it's possible to pin a specific version or a range here  which should be currentValue.

https://fsprojects.github.io/Paket/dependencies-file.html

lib/modules/manager/paket/extract.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+                Fake (5.16)
+          `);
+        },
+      );
+      const packageFileContent = codeBlock`
+        source https://api.nuget.org/v3/index.json
+
+        group GroupA
+          source https://api.nuget.org/v3/index.json
+          nuget Fake
```
viceice, line 268: see other comment. add sample with pinned nuget packages

https://fsprojects.github.io/Paket/dependencies-file.html

lib/modules/manager/paket/extract.ts
```diff
@@ -0,0 +1,87 @@
...
+  p: DependenciesFilePackage,
+): PackageDependency {
+  const lockVersion = searchPackageVersion(dependencies, group, p);
+
+  const version = lockVersion?.version;
+  const name = lockVersion?.packageName ?? p.name;
+  return {
+    depType: 'dependencies',
+    depName: name,
+    currentVersion: version,
```
viceice: this should be the version / constraint from `paket.dependencies` if defined.

we probably need a new `paket` versioning with understods the paket range identifiers (`~>`, `>=`, ...).

if you don't want to implement that then at least extract them and set a proper skip reason, so they are visible in logs

lib/modules/manager/paket/extract.ts
```diff
@@ -0,0 +1,87 @@
...
+    datasource: NugetDatasource.id,
+    rangeStrategy: 'update-lockfile',
```
viceice: you should implement `getRangeStrategy` instead

https://github.com/renovatebot/renovate/blob/1540a34fcf45598ca2b9461e0e194f4259a9281b/lib/modules/manager/cargo/range.ts#L4-L15

viceice (changes requested):
there are still a lot open implementation details and probably a wrong understanding how renovate internals work.

viceice:
https://github.com/renovatebot/renovate/blob/main/docs/development/adding-a-package-manager.md#updateartifacts-async-optional

it says use `updateLockedDependency` to directly update lockfile. if you need to call a binary use `updateArtifacts`

https://github.com/renovatebot/renovate/blob/main/docs/development/adding-a-package-manager.md#updatelockeddependency

lib/modules/manager/paket/extract.ts
fpellet: This MR was solely intended to maintain the lock files. I have to re-validate it every three months, even though Rhys approved it over a year ago.
I know you are doing your best to maintain this repository and I appreciate your efforts, but I am starting to lose hope that this PR will ever be approved.
Perhaps I don't fully understand all your comments.
But receiving new comments every three months, along with new requirements, is a real nightmare. It feels like you don't want this contribution.

I think I'll take a look at your comments, but after that, I'll give up and let another volunteer handle it, or find a different solution for managing my versioning.

lib/util/exec/containerbase.ts
fpellet: Paket's published versions are plain SemVer, and as you noted, nuget versioning treats a bare 1.2.3 as a range, so isSingleVersion() would fail in resolveConstraint() and a user setting constraints: { paket: '8.0.3' } wouldn't get the pinned install. That's why semver is used here.

zharinov:
Additions to this PR will hopefully address the requested changes:

- https://github.com/fpellet/renovate/pull/1

Paket versioning:

- #44366 

Example repo:

- https://github.com/renovate-testing/paket-e2e/pulls

viceice (commented):
looks good so far. please merge https://github.com/fpellet/renovate/pull/1, `paket` versioning is on merge queue and should be merged in a few minutes

MichaelSiggelkowRnS:
Is there something really left to get this merged? I really hope to get this soon! Thanks for all your efforts to everyone!

fpellet:
I've realigned it with "main". What else can I do to hope it gets merged soon?

lib/util/exec/containerbase.ts
```diff
@@ -169,6 +169,11 @@ export const allToolConfig: Record<ToolName, ToolConfig> = {
...
```
RahulGautamSingh: Now that paket versioning is merged and treats `1.2.3` as a single version, use versioning: 'paket' here.

lib/modules/manager/paket/readme.md
```diff
@@ -0,0 +1,15 @@
...
+- `paket.dependencies`
+- `paket.lock`
+
+The `source` entries of each group are used as registry URLs for the dependencies of that group.
+
+Restrictions:
+
+- Only updates nuget source (github, gist, http, and git not supported)
+- Version constraints in `paket.dependencies` are updated with the [`paket` versioning](../../versioning/paket/index.md) module
+- If `paket.lock` is missing, dependencies are extracted but not updated
```
RahulGautamSingh: deps with a version constraint get no `skipReason` when `paket.lock` is missing, so they still get updated.

Set a skip reason, so they're skipped or update this statement.

lib/modules/manager/paket/tool.ts
```diff
@@ -0,0 +1,39 @@
...
+
+function buildUpdateCommand(command: UpdatePackage): string {
+  const groupFilter = command.group ? ` --group ${quote(command.group)} ` : '';
+  const packageFilter = command.packageName
+    ? ` ${quote(command.packageName)} `
+    : '';
+  const version = command.version
+    ? ` --version ${quote(command.version)} `
+    : '';
+  return `paket update${groupFilter}${version}${packageFilter}`;
```
RahulGautamSingh: `paket update` installs into projects and runs `dotnet restore` unless `--no-install` is passed. 

only the lock file is useful for us, so add `--no-install to every command to skip the project writes and the restore.

@fpellet is that okay or do you think we need the dotnet restore?

also to mention, we already do the dotnet restore in nuget manager unconditionally so it could be that it's not that necessary to remove, but I am not familar with the ecosystem to be sure

RahulGautamSingh (changes requested):

lib/modules/manager/paket/tool.ts
fpellet: Good point

fpellet:
Anything else?

RahulGautamSingh (approved):

lib/modules/manager/paket/artifacts.ts
```diff
@@ -0,0 +1,122 @@
...
+  const toolConstraints: ToolConstraint[] = [
+    {
+      toolName: 'dotnet',
+      constraint: updateArtifact.config.constraints?.dotnet,
+    },
+    {
+      toolName: 'paket',
+      constraint: updateArtifact.config.constraints?.paket,
+    },
+  ];
```
viceice, line 81: @secustor needs to be fixed later for #45826

lib/modules/manager/paket/extract.ts
```diff
@@ -0,0 +1,91 @@
...
+
+function findLockedDependency(
+  dependencies: LockFileDependency[],
+  group: DependenciesFileGroup,
+  pkg: DependenciesFilePackage,
+): LockFileDependency | undefined {
+  return dependencies.find(
+    (d) =>
+      d.groupName.toUpperCase() === group.groupName.toUpperCase() &&
+      d.packageName.toUpperCase() === pkg.name.toUpperCase(),
```
viceice: store `group.groupName.toUpperCase()` and `pkg.groupName.toUpperCase()` in variables before find call (perf)

lib/modules/manager/paket/update-locked.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  if (!lockFileContent) {
+    return { status: 'update-failed' };
+  }
+
+  const group = managerData?.group;
+  const lockedEntries = parseLockFile(lockFileContent).filter(
+    (dep) =>
+      dep.packageName.toUpperCase() === depName.toUpperCase() &&
+      (!group || dep.groupName.toUpperCase() === group.toUpperCase()),
```
viceice: same here cache uppercase of `depName` and `group` (perf)

viceice (commented):
otherwise lgtm

fpellet:
@viceice Anything is fine to merge

purkhusid:
Friendly ping. Would be absolutely awesome to see this merged!

PR #37315, opened 2025-08-04
feat(pip-compile): add --no-emit-package to uv pip compile allowed arguments

Zokormazo:
## Changes

This PR supports the optional `--no-emit-package=foo` argument for `uv pip compile` on the pip-compile manager

## Context

Using the argument was throwing an error on the manager due to an unknown argument. Adding it to the allowed arguments list avoid the error.


viceice:
what is that argumernt good for?

Zokormazo:
@viceice is helpful to ignore specific dependencies from the compiled requirements.txt file. Is the equivalent of `--unsafe-package` on `pip-compile`

As a packager sometimes I need to skip a given child dependency of one of my dependencies. This can be either because the package I'm using is declaring a dependency on a wrong way for my packagaing purpose (`psycopg[binary]` for example), or because I want to change an specific dependency to a fork, or because is already present on the base image I'm building on.

Using `--omit-package=[package]` I can ignore the given package from my requirements file and achieve that.

Fixed the missing equal on the tests.

viceice (approved):

PR #37369, opened 2025-08-07
feat(github-actions): Support actions/setup-dotnet with:dotnet-version

martincostello:
## Changes

Add support for extracting `dotnet-version` from `actions/setup-dotnet` in the GitHub Actions manager.

## Context

Renovate does not currently support updating the value specified by `dotnet-version` with [`actions/setup-dotnet`](https://github.com/actions/setup-dotnet) such as the below:

```yaml
- uses: actions/setup-dotnet@v4
  with:
    dotnet-version: 8.0.x
```

These changes should light-up support based on the [custom syntax for the .NET SDK version the action supports](https://github.com/actions/setup-dotnet?tab=readme-ov-file#supported-version-syntax) (which is not the same syntax `global.json` itself supports).

I'm not sure what the best name for the new versioning strategy should be as it seems specific to the `actions/setup-dotnet` action. For now I've called it `dotnet-sdk` and based it on `nuget`.

### Examples

Example updates this PR should enable depending on configuration.

#### Major update for A

```diff
- dotnet-version: 8
+ dotnet-version: 9
```

#### Major update for A.x

```diff
- dotnet-version: 8.x
+ dotnet-version: 9.x
```

#### Major update for A.B

```diff
- dotnet-version: 8.0
+ dotnet-version: 9.0
```

#### Major update for A.B.x

```diff
- dotnet-version: 8.0.x
+ dotnet-version: 9.0.x
```

#### Major update for A.B.Cxx

```diff
- dotnet-version: 8.0.1xx
+ dotnet-version: 9.0.1xx
```

#### Minor update for A.B.Cxx

```diff
- dotnet-version: 8.0.1xx
+ dotnet-version: 8.0.4xx
```

#### Minor update for A.B.C

```diff
- dotnet-version: 8.0.100
+ dotnet-version: 8.0.413
```

#### Major update for A.B.C

```diff
- dotnet-version: 8.0.100
+ dotnet-version: 9.0.304
```

#### Pinning for minor A.B.x

```diff
- dotnet-version: 8.0.x
+ dotnet-version: 8.0.413
```

#### Minor update for two A.B.C values

```diff
 dotnet-version: | 
-    8.0.100
-    9.0.100
+    8.0.413
+    9.0.304
```

I'm not sure if the last example will work or not - I've updated the manager so that it correctly extracts multiple values, but whether they would be successfully applied to the YAML file is a different question. I couldn't immediately see where to add a test for that scenario.

## Documentation

- [ ] ~~No documentation update is required~~

## How I've tested my work

- [ ] ~~Code inspection only, or~~
- [ ] ~~No unit tests but ran on a real repository, or~~
- [ ] ~~Both unit tests + ran on a real repository~~


lib/modules/manager/github-actions/extract.ts
```diff
@@ -188,20 +200,27 @@ function extractSteps(
...
-            currentValue,
-            depType: 'uses-with',
+          currentValue.split(newlineRegex).forEach((value) => {
+            if (value) {
+              deps.push({
+                datasource: datasource ?? GithubReleasesDatasource.id,
+                depName: action,
+                packageName: packageName ?? `actions/${action}-versions`,
+                versioning,
+                extractVersion: '^(?<version>\\d+\\.\\d+\\.\\d+)(-\\d+)?$', // Actions release tags are like 1.24.1-13667719799
```
martincostello, line 226: This might not be correct, as it doesn't look like it would work for values with an `x` in.

lib/modules/versioning/dotnet-sdk/types.ts
```diff
@@ -0,0 +1,29 @@
...
+ * - `8.0.x`
+ * - `8.0.4xx`
+ *
+ * Invalid range: `8.0.*-foo`
+ */
+export interface DotnetSdkFloatingRange {
+  type: 'dotnet-sdk-floating-range';
+  major: number;
+  minor?: number | 'x';
+  patch?: number | 'x';
```
astellingwerf, line 24: The example on line 16 does not match the type declared here

lib/modules/versioning/dotnet-sdk/version.ts
```diff
@@ -0,0 +1,77 @@
+import { regEx } from '../../../util/regex';
+import type { DotnetSdkVersion } from './types';
+
+// We can't just use `Number.parseInt()` beacause it parses `123abc` as `123`,
```
astellingwerf:
```suggestion
// We can't just use `Number.parseInt()` because it parses `123abc` as `123`,
```

lib/modules/versioning/dotnet-sdk/types.ts
martincostello: True - it's mainly being used as a way to round-trip the string into its parts and back again for the `A.B.x` case when `floating` is `minor`.

Line 16 will be `8` `0` `400` in memory for the `floating: 'patch'` case.

lib/modules/versioning/dotnet-sdk/version.ts
martincostello: Fixed. This is a copy-pasta typo from https://github.com/renovatebot/renovate/blob/b751d86212c087b8fc3539986e82d1a61d9bd25a/lib/modules/versioning/nuget/version.ts#L4.

viceice (commented):
we usually don't review draft PR's

martincostello:
I'm not 100% on exactly how to approach this, so feedback/guidance sought. Then I can deal with the Copilot review comments and any other feedback.

lib/modules/manager/github-actions/extract.ts
viceice: use for-of loop

viceice (commented):
please extract the versioning into a separate PR

martincostello:
It would be good to get some comments on the versioning part, as it's the bit I'm the least sure of doing in a Renovate-y way, then I can incorporate any feedback into a separate PR to add just that part.

RahulGautamSingh:
**Minor update for A.B.Cxx**

```diff
- dotnet-version: 8.0.1xx
+ dotnet-version: 8.0.4xx
```

**Minor update for A.B.C**
```diff
- dotnet-version: 8.0.100
+ dotnet-version: 8.0.413
```

These should be patch updates not minor?

RahulGautamSingh:
Thanks! for the PR @martincostello 

Shouldn't these be considered patch updates?

**Minor update for A.B.Cxx**
```diff
- dotnet-version: 8.0.1xx
+ dotnet-version: 8.0.4xx
```

**Minor update for A.B.C**
```diff
- dotnet-version: 8.0.100
+ dotnet-version: 8.0.413
```

lib/modules/versioning/dotnet-sdk/index.ts
```diff
@@ -0,0 +1,325 @@
...
+
+class DotnetSdkVersioningApi implements VersioningApi {
+  isCompatible(version: string, _current?: string): boolean {
+    return this.isValid(version);
+  }
+
+  isSingleVersion(version: string): boolean {
+    const r = parseRange(version);
+    if (!r) {
+      return parseVersion(version) !== null;
```
RahulGautamSingh, line 21: Why not just return false, as this should only happen incase of invalid version string?

lib/modules/versioning/dotnet-sdk/index.ts
```diff
@@ -0,0 +1,325 @@
...
+    if (v) {
+      return !v.prerelease;
+    }
+
+    const r = parseRange(version);
+    if (!r) {
+      return false;
+    }
+
+    return !r.prerelease;
```
RahulGautamSingh, line 38:
```suggestion
    const r = parseRange(version);
    if (r) {
      return !r.prerelease;
    }

    return false;
```

lib/modules/versioning/dotnet-sdk/index.ts
```diff
@@ -0,0 +1,325 @@
...
+    if (!input) {
+      return false;
+    }
+
+    const v = parseVersion(input);
+    if (!v) {
+      return false;
+    }
+
+    return true;
```
RahulGautamSingh, line 65:
```suggestion
    return !!v;
```

lib/modules/versioning/dotnet-sdk/index.ts
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  getMajor(version: string): number | null {
+    const v = parseVersion(version);
+    if (!v) {
+      return null;
+    }
+
+    return v.major;
```
RahulGautamSingh, line 74:
```suggestion
    return v?.major ?? null;
```

lib/modules/versioning/dotnet-sdk/index.ts
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  getMinor(version: string): number | null {
+    const v = parseVersion(version);
+    if (!v) {
+      return null;
+    }
+
+    return v.minor ?? null;
```
RahulGautamSingh, line 83: same as getMajor

lib/modules/versioning/dotnet-sdk/index.ts
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  getPatch(version: string): number | null {
+    const v = parseVersion(version);
+    if (!v) {
+      return null;
+    }
+
+    return v.patch ?? null;
```
RahulGautamSingh, line 92: same as getMajor

lib/modules/versioning/dotnet-sdk/version.ts
```diff
@@ -0,0 +1,77 @@
...
+  for (let i = 0; i < maxLen; i += 1) {
+    const xPart = xParts[i] ?? '';
+    const yPart = yParts[i] ?? '';
+    const xNum = ensureNumber(xPart);
+    const yNum = ensureNumber(yPart);
+
+    const res =
+      xNum !== null && yNum !== null
+        ? Math.sign(xNum - yNum)
+        : xPart.localeCompare(yPart, undefined, { sensitivity: 'base' });
```
RahulGautamSingh, line 28: Is this to compare `rc` and `preview` ? Is it okay to compare them lexogrpahically?

RahulGautamSingh (commented):
Initial Review

RahulGautamSingh (changes requested):
The test files have many redundant examples, please filter them out as well as it takes more focus to find the different test cases.

PR #37408, opened 2025-08-08
feat(manager/pub): extract git deps

wsn-dev:
## Changes

This changes allows the pub manager to extract git based pub packages.

## Context

Currently renovate is not able to extract git hosted pub packages.
With this pr i would like to introduce the extraction of such git sources.
Also i found the following discussion about it: https://github.com/renovatebot/renovate/discussions/15945
On the following Site you will find the offcial spec: https://dart.dev/tools/pub/dependencies#git-packages

- #15945
- #32357 


RahulGautamSingh:
Thank you for your PR! 

Could you please do a test run on a dummy repo and link it here?

And, modify the PR title to:

`feat(manager/pub):extract git deps`

wsn-dev:
> Thank you for your PR!
> 
> Could you please do a test run on a dummy repo and link it here?
> 
> And, modify the PR title to:
> 
> `feat(manager/pub):extract git deps`

Hi RahulGautamSingh,

thank you for your recommendations really appreciate it.

Here is a demo repo with a run of `renovate --platform=local` [renovate-pub-manager-demo](https://github.com/wsn-dev/renovate-pub-manager-demo/blob/main/debug.txt) using LOG_LEVEL=debug.

The image was built locally with: `docker build -t renovate-local --no-cache -f .\tools\docker\Dockerfile`.

Please let me know if there is more needed and if that was the correct way of demonstrate it.

lib/modules/manager/pub/extract.ts
```diff
@@ -32,36 +33,70 @@ function extractFromSection(
...
+
       if (version) {
         currentValue = version;
       } else if (path) {
         currentValue = '';
         skipReason = 'path-dependency';
+      } else if (is.object(git) && is.string(git?.ref)) {
+        currentValue = git?.ref;
+      } else if (is.object(git) && !is.string(git?.ref) && is.string(version)) {
+        currentValue = version;
```
viceice: you can never get here

lib/modules/manager/pub/extract.ts
```diff
@@ -32,36 +33,70 @@ function extractFromSection(
...
+        datasource: DartDatasource.id,
+        ...(registryUrls && { registryUrls }),
+        skipReason,
+      });
+    } else {
+      deps.push({
+        depName,
+        depType: sectionKey,
+        packageName: gitUrl,
+        datasource: GitRefsDatasource.id,
```
viceice, line 89: we need to drop this, otherwise all git deps get npm versioning assigned
https://github.com/renovatebot/renovate/blob/930fb7cd8ded2b092a1b4ad8cdcc57a5d492ac76/lib/modules/manager/pub/index.ts#L18

lib/modules/manager/pub/extract.ts
```diff
@@ -32,36 +33,70 @@ function extractFromSection(
...
 
       if (is.string(hosted)) {
         registryUrls = [hosted];
       } else if (is.string(hosted?.url)) {
         registryUrls = [hosted.url];
       }
 
+      if (is.object(git)) {
+        gitUrl = git?.url;
+      }
```
viceice, line 54:
```suggestion
      } else if (is.string(git)) {
        gitUrl = git;
      }
```

schema allows string, so i assume it's a full repo url only then?

wsn-dev: It is allowed to have both so it behaves like the path spec:
https://dart.dev/tools/pub/dependencies#git-packages
I hope it is correct to pass it trough directly as registryUrl.
It is also allowed to omit the .git part
So you can write it in the pubspec for example:
```
dependencies:
  kittens:
    # SSH URL:
    git: git@github.com:munificent/kittens.git
dependencies:
  kittens:
    # HTTPS URL:
    git: https://github.com/munificent/kittens.git
dependencies:
  kittens:
    git:
      url: git@github.com:munificent/kittens.git
      ref: some-branch-tag-or-ref
```

lib/modules/manager/pub/extract.ts
wsn-dev: dropped thank you

lib/modules/manager/pub/extract.ts
wsn-dev: Thank you
It took me a bit but now i get what you mean. 
Cause the version is already in the if in line 55.
So you probably mean i can not get to line 63.
I removed it

lib/modules/manager/pub/extract.ts
RahulGautamSingh: If both types are possible we need to handle both.

lib/modules/manager/pub/extract.ts
```diff
@@ -32,36 +33,68 @@ function extractFromSection(
...
+
       if (version) {
         currentValue = version;
       } else if (path) {
         currentValue = '';
         skipReason = 'path-dependency';
+      } else if (is.object(git) && is.string(git?.ref)) {
+        currentValue = git?.ref;
```
RahulGautamSingh:
```suggestion
        currentValue = git.ref;
```

lib/modules/manager/pub/extract.ts
```diff
@@ -32,36 +33,68 @@ function extractFromSection(
...
+        !is.string(version)
+      ) {
+        currentValue = '';
+        skipReason = 'unspecified-version';
       } else {
         currentValue = '';
       }
+
+      if (is.object(git) && is.string(git?.tag_pattern)) {
+        tagPattern = git?.tag_pattern;
```
RahulGautamSingh:
```suggestion
        tagPattern = git.tag_pattern;
```

RahulGautamSingh (changes requested):

lib/modules/manager/pub/extract.ts
wsn-dev: done thanks

lib/modules/manager/pub/extract.ts
wsn-dev: done thanks

lib/modules/manager/pub/extract.ts
RahulGautamSingh: @viceice Do we allow git urls as regsitryUrl?

wsn-dev: @RahulGautamSingh i think it was a typo from my side sorry  it is added as packagename Like described here https://docs.renovatebot.com/modules/datasource/git-refs/ but it would be still interesting if ssh is also accepted there

RahulGautamSingh: It should be okay then

wsn-dev:
Hi,
What’s the best way to proceed?
Should I update or change anything?

RahulGautamSingh (approved):
LGTM

wsn-dev:
Hi,
is there any Chance that the pr is getting merged?
Currently it is very exhausting to disable every git dependency since the lookup curently Maps everything to the pub Registry.
Cause otherwise Renovate will comment with an error of Failed to look up dart package
Thank you

lib/modules/manager/pub/extract.spec.ts
```diff
@@ -103,6 +123,48 @@ describe('modules/manager/pub/extract', () => {
...
+            skipReason,
+          },
+          {
+            currentValue: '^1.1.0',
+            depName: 'git_package_version_extract',
+            packageName:
+              'https://github.com/some-url/some-package-version-extract',
+            depType: dependenciesDepType,
+            datasource: gitRefDatasource,
+            extractVersion: 'v',
```
secustor: Have you tested this end to end?  

AFAIK this shouldn't work and should be removed as semver-coerced should be able to handle it automatically

wsn-dev: Hi,
sorry but i can not really follow here.
i use the extractVersion for the tagPattern you can specify in pub.

> The package you depend on has tagged the revision of each version of the package, you can use tag_pattern instead of ref, together with a version constraint.

https://dart.dev/tools/pub/dependencies#git-packages.

secustor: I'm refering to if you have tested this specific combination because [extractVersion doesn't work like that](
https://docs.renovatebot.com/configuration-options/#extractversion) and `semver-coerced` should handle this automatically. 

```suggestion
```

wsn-dev: Hi,
thank you.
I removed it then (If it works like the config option). 
So then the user can decide in rules to change it i guess.

secustor (approved):

PR #37434, opened 2025-08-11
feat(preset): add Gradle group

JuJup:
## Changes

Adds a preset for Gradle Version updates.

It is based on the nodeJs group.
Explicitly not using groupName, see discussion in this PR: https://github.com/renovatebot/renovate/pull/34379

### DIFF
--- a/lib/config/presets/internal/group.ts
+++ b/lib/config/presets/internal/group.ts
@@ -208,7 +208,7 @@ const staticGroups = {
         commitMessageTopic: 'Gradle',
         matchDatasources: ['docker', 'gradle-version'],
         matchPackageNames: [
-          '/(?:^|/)gradle/', // gradle or ends with "/gradle"
+          '/(?:^|/)gradle$/', // gradle or ends with "/gradle"
         ],
       },
     ],
--- a/lib/config/presets/internal/group.ts
+++ b/lib/config/presets/internal/group.ts
@@ -200,6 +200,19 @@ const staticGroups = {
       },
     ],
   },
+  gradle: {
+    description:
+      "Group anything that looks like Gradle together so that it's updated together.",
+    packageRules: [
+      {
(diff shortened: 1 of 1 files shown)


lib/config/presets/internal/group.ts
```diff
@@ -200,6 +200,19 @@ const staticGroups = {
       },
     ],
   },
+  gradle: {
+    description:
+      "Group anything that looks like Gradle together so that it's updated together.",
+    packageRules: [
+      {
+        commitMessageTopic: 'Gradle',
```
astellingwerf, line 208: This file is about group presets, not about commit messages

lib/config/presets/internal/group.ts
```diff
@@ -200,6 +200,19 @@ const staticGroups = {
...
+        matchDatasources: ['docker', 'gradle-version'],
+        matchPackageNames: [
+          '/(?:^|/)gradle/', // gradle or ends with "/gradle"
```
viceice:
```suggestion
          '/(?:^|/)gradle$/', // gradle or ends with "/gradle"
```
otherwise it will allow `something/gradle124node`

lib/config/presets/internal/group.ts
viceice: it's modeled fater this:
https://github.com/renovatebot/renovate/blob/930fb7cd8ded2b092a1b4ad8cdcc57a5d492ac76/lib/config/presets/internal/group.ts#L389-L405

JuJup: See this discussion: https://github.com/renovatebot/renovate/pull/34379

lib/config/presets/internal/group.ts
JuJup: Good catch, thank you 🙏🏼

viceice (approved):
failing ci

PR #37449, opened 2025-08-12
feat(nuget): allow restoring workloads before restoring packages

viceice:
## Changes
Some projects need to restore dotnet workloads before you can restore packages.
It happens for `browser-wasm` projects.

## Context
- https://github.com/renovatebot/pgp/pull/7#issuecomment-3146189879


viceice:
I'm not sure if we should simply always run it. it's mostly a no-op if no additional workloads are used 🤔

rarkins (approved):

PR #37457, opened 2025-08-12
feat(pip-compile): Support compiling dependencies from pyproject.toml

mbudnek:
## Changes

`pip-compile` (and `uv pip compile`) support compiling dependencies from pyproject.toml, but that functionality wasn't enabled in the pip-compile manager.  This uses the pep621 manager's `extractPackageFile` the same way we do with the pip_setup and pip_requirements managers to enable the pip-compile maanger to handle pyproject.toml files.

## Context

My org has started moving more of our package configuration into pyproject.toml files.  We still use setuptools for package builds though, and setuptools doesn't have a native lock file like other PEP-621 managers.  We use pip-compile to generate requirements.txt files that act as lock files for our CI tests and for application deployments.


lib/modules/manager/pip-compile/extract.spec.ts
```diff
@@ -56,15 +56,46 @@ describe('modules/manager/pip-compile/extract', () => {
       expect(packageFile?.deps[0]).toHaveProperty('depName', 'celery');
     });
 
+    it('returns object for pyproject.toml', async () => {
+      const pyproject = `[build-system]
```
viceice: use `codeBlock` template helper and indent properly

viceice (commented):
please provide a sample repo with lockfile

mbudnek:
@viceice The lock file is a normal pip requirements file just like it always is with pip-compile.  Here's an example repo I set up to test: https://github.com/mbudnek/renovate-pip-compile-pyproject-example

mbudnek:
@viceice Just a bump.  Sorry to bother you.  Is there anything else needed for this PR?

viceice (approved):

PR #37458, opened 2025-08-12
feat(poetry): supersede `pep621`

### LINKED ISSUE #33961: Suppress/skip pep621 manager detection of poetry 2.0 projects

Currently Poetry 2.0 projects are _not_ recognized by the "poetry" manager, but they _are_ recognized (partially) by the "pep621" manager. We should update the code in https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/pep621/extract.ts so that Poetry 2.0 projects are skipped by the pep621 manager.

This could potentially be done either by looking for signs of poetry in the pyproject.toml, or by looking for a poetry.lock. The pyproject.toml would seem to be more reliable if that is possible.

For now we can assume that "if it's as poetry pyproject.toml, it's not also another pep621 type of project at the same time (e.g. uv)". If someone has a use case where they have _both_ poetry and another pep621 at the same time, we can decide later whether to support that.

viceice:
## Changes
Supercede extractions from pep621 manager by poetry.
`project.dependencies` and `project.optional-dependencies` are no longer extracted by `pep621`.

## Context
- closes #33961

### DIFF
--- a/lib/modules/manager/poetry/index.ts
+++ b/lib/modules/manager/poetry/index.ts
@@ -11,6 +11,7 @@ export { extractPackageFile } from './extract';
 export { updateArtifacts } from './artifacts';
 export { updateLockedDependency } from './update-locked';
 
+export const supersedesManagers = ['pep621'];
 export const supportsLockFileMaintenance = true;
 
 export const url = 'https://python-poetry.org/docs';
--- a/lib/workers/repository/extract/supersedes.spec.ts
+++ b/lib/workers/repository/extract/supersedes.spec.ts
@@ -31,6 +31,23 @@ describe('workers/repository/extract/supersedes', () => {
             },
           ],
         },
+        {
+          manager: 'pep621',
+          packageFiles: [
+            { packageFile: 'pyproject.toml', deps: [] },
+            { packageFile: 'some/pyproject.toml', deps: [] },
+          ],
+        },
+        {
+          manager: 'poetry',
+          packageFiles: [
+            {
+              packageFile: 'pyproject.toml',
+              deps: [],
+              lockFiles: ['poetry.lock'],
+            },
+          ],
+        },
       ];
       processSupersedesManagers(extractResults);
       expect(extractResults).toMatchObject([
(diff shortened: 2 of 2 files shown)


RahulGautamSingh (approved):

rarkins (approved):

charlie-ang-collibra:
Hi @viceice !   If I may, I wanted to ask a question about this change.  We currently use `uv` with renovatebot and do have various optional `dependency-groups`.   After upgrading to this version, lock maintenance on `uv.lock` is no longer happening, although renovatebot continues to update the main dependencies section in `pyproject.toml`.   PRs that were opened against dependencies in the `groups` were also auto-closed.

Is this the intended effect of this on `uv` and `uv.lock`?    If it is and there is a migration guide, a quick link would be very appreciated!  Thank you.

jkkitakita:
@viceice 
me too. uv project is recognized as poetry even though there is no poetry.lock and there is a uv.lock.
I would love to have a guide on how to fix!

viceice:
will look at it as fast as possible

jonadaly:
As a temporary workaround, it looks like disabling the poetry manager entirely does solve this problem:

```json
"poetry": {
    "enabled": false
  }
```

charlie-ang-collibra:
> As a temporary workaround, it looks like disabling the poetry manager entirely does solve this problem:
> 
> ```json
> "poetry": {
>     "enabled": false
>   }
> ```

Much appreciated, @viceice , @jonadaly !   Yes, we applied the `enabled: false` solution yesterday in the repo config and it worked liked a charm!

Before that, I was looking at [supercedes.ts](https://github.com/renovatebot/renovate/blob/137d46d61ee7594c0a82fb042cb0745b372b1b29/lib/workers/repository/extract/supersedes.ts#L24-L29) and realized that pep621's `extractPackageFile` function never populates the `lockFiles` attribute.  Therefore, the logic in `processSupersedesManagers` will always make `poetry` win since it no longer (a) needs to be any poetry elements in pyproject.toml nor (b) checks for a presence of a `poetry.lock` file.  It was sufficient that pep621 did not populate the `lockFiles` attribute for it to be overridden by `poetry`.   I didn't look into how the `bun` / `npm` scenario was different.

Once I saw that, I arrived at the same conclusion @jonadaly did.  But it's very nice to hear that it's a valid solution.

viceice:
woking on a fix now, thanks @charlie-ang-collibra for the good analysation 🤗

viceice:
- #37633
this should fix it

vivodi:
> * [feat(pep621): return found lockfiles #37633](https://github.com/renovatebot/renovate/pull/37633)
>   this should fix it

The issue is not resolved. We are using uv.lock, but it is still being recognized as Poetry. https://github.com/Flexget/Flexget/issues/3566

viceice:
it's not yet deployed to the hosted app

PR #37522, opened 2025-08-14
feat: Add support for Base64-encoded keys

### LINKED ISSUE #37517: Support base64-encoded privateKey values

Support base64 encoding of the `privateKey` field.

Add a check in case the string value for `privateKey` config option is base64-encoded. If so, base64-decode it before attempting to use it.

Do not apply to (do not change) privateKeyOld, privateKeyPath, privateKeyPathOld.

zharinov:
## Context

- Closes: #37517

### DIFF
--- a/lib/util/git/private-key.ts
+++ b/lib/util/git/private-key.ts
@@ -7,6 +7,7 @@ import { logger } from '../../logger';
 import { exec } from '../exec';
 import { newlineRegex, regEx } from '../regex';
 import { addSecretForSanitizing } from '../sanitize';
+import { fromBase64, toBase64 } from '../string';
 
 type PrivateKeyFormat = 'gpg' | 'ssh';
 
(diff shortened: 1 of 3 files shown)


lib/util/git/private-key.ts
```diff
@@ -17,16 +18,38 @@ const sshKeyRegex = regEx(
 
 let gitPrivateKey: PrivateKey | undefined;
 
+/**
+ * Decodes Base64 string if roundtrip encoding matches
+ */
+function tryBase64(value: string): string | null {
```
rarkins, line 24: How/where does this return null?

lib/util/git/private-key.ts
```diff
@@ -17,16 +18,38 @@ const sshKeyRegex = regEx(
...
   protected abstract readonly gpgFormat: string;
 
   constructor(key: string, passphrase: string | undefined) {
-    this.key = key;
-    this.passphrase = passphrase;
+    const decodedKey = tryBase64(key);
+
+    if (decodedKey) {
+      this.key = decodedKey;
+      addSecretForSanitizing(key, 'global');
```
rarkins, line 45: This will add the secret regardless of whether it was base64-encoded or not. Is that intentional?

lib/util/git/private-key.ts
```diff
@@ -17,16 +18,38 @@ const sshKeyRegex = regEx(
...
   protected readonly passphrase: string | undefined;
   protected keyId: string | undefined;
   protected abstract readonly gpgFormat: string;
 
   constructor(key: string, passphrase: string | undefined) {
-    this.key = key;
-    this.passphrase = passphrase;
+    const decodedKey = tryBase64(key);
+
+    if (decodedKey) {
```
rarkins, line 43: I don't see how this is ever falsy unless it's `''`

zharinov: Yup, fixed this

rarkins (approved):

PR #37541, opened 2025-08-15
feat(datasource/maven): set latest tag ignored by default

rarkins:
Reintroduces latest/release tags for Maven datasource, but this time sets respectLatest to false if a latest tag is found. When this feature was released originally, there were too many situations where registry latest tags were unreliable (e.g. the most recent version pushed, not some "stable" latest). With this new approach, users have to explicitly opt in.


rarkins:
FYI @Shegox

Shegox:
I think as a optionally opt-in this is a great solution 👍 

(I still think that the official spec/implementation of the maven registry didn't intend for this use case)

rarkins:
Yes, not I think about it - using the term "latest" to implicitly mean "stable" is not a good idea in any ecosystem. We've spent much time in this project alone dealing with the nuances between latest, last, stable, more recent, etc.

viceice (approved):

Issue #37537, opened 2025-08-15
datasource/crate: preserve version metadata for postProcessResult lookup


rarkins:
### Discussed in https://github.com/renovatebot/renovate/discussions/37334

<div type='discussions-op-text'>

<sup>Originally posted by **DanielleHuisman** August  5, 2025</sup>
### How are you running Renovate?

A Mend.io-hosted app

### Please tell us more about your question or problem

Crates with version metadata, e.g. [`phonenumber@0.3.7+8.13.52`](https://crates.io/crates/phonenumber/0.3.7+8.13.52), are not supported by Renovate. Renovate fails to lookup crate information, because it strips the version metadata.

Current behaviour
```
GET https://crates.io/api/v1/crates/phonenumber/0.3.7 = (code=ERR_NON_2XX_3XX_RESPONSE, statusCode=404 retryCount=0, duration=18)
```

Expected behaviour
```
GET https://crates.io/api/v1/crates/phonenumber/0.3.7+8.13.52
```

See [this minimal reproduction](https://github.com/DanielleHuisman/renovate-crate-semver-metadata) for a more detailed explanation of the problem.

I think this problem has been introduced by https://github.com/renovatebot/renovate/pull/27642.

### Logs (if relevant)

<details><summary>Logs</summary>

```

DEBUG: GET https://crates.io/api/v1/crates/phonenumber/0.3.7 = (code=ERR_NON_2XX_3XX_RESPONSE, statusCode=404 retryCount=0, duration=18)
WARN: Release interceptor failed for "crate"
{
  "err": {
    "name": "HTTPError",
    "code": "ERR_NON_2XX_3XX_RESPONSE",
    "timings": {
      "start": 1754384095852,
      "socket": 1754384095853,
      "lookup": 1754384095853,
      "connect": 1754384095853,
      "secureConnect": 1754384095853,
      "upload": 1754384095853,
      "response": 1754384095869,
      "end": 1754384095870,
      "phases": {
        "wait": 1,
        "dns": 0,
        "tcp": 0,
        "tls": 0,
        "request": 0,
        "firstByte": 16,
        "download": 1,
        "total": 18
      }
    },
    "message": "Response code 404 (Not Found)",
[...]

PR #37548, opened 2025-08-15
fix(cargo): preserve version metadata

DanielleHuisman:
## Changes

Preserve version metadata in Cargo datasource, so the registry request doesn't fail. Instead, only strip version metadata from `newValue` in Cargo versioning.

I tried using `versionOrig` in the Cargo datasource, as suggested in the issue. This fixes the registry request issue, but instead it runs into a new issue when updating the Cargo lock file using `cargo update`. This command also required the version metadata.

## Context

#37537

## Documentation

## How I've tested my work

Example on a real repository: https://github.com/DanielleHuisman/renovate-issue-37537/pull/2


zharinov (approved):

rarkins (changes requested):
I'm not sure this is logically correct.

If you are adding logic so that currentVersion -> currentValue logic is changed, then it can create a problem when currentValue -> currentVersion is unchanged.

For example this PR sets `0.2.0+metadata` version to be `0.2.0` value, but if you were to run something like `matches('0.2.0+metadata', '0.2.0')` then it would return `false`.

I think we need more discussion about this before we accept a change from the solution proposed in the corresponding Issue

rarkins:
What exactly was the `cargo update` problem?

DanielleHuisman:
This is the error when using the `versionOrig` approach:

```
Command failed: cargo update --config net.git-fetch-with-cli=true --manifest-path Cargo.toml --package phonenumber@0.3.6+8.13.36 --precise 0.3.7
error: package ID specification `phonenumber@0.3.6+8.13.36` did not match any packages
help: there are similar package ID specifications:

  phonenumber@0.3.7+8.13.52
```

I see I misinterpreted this error yesterday. I thought the issue was the `--precise 0.3.7` part, but the `--package phonenumber@0.3.6+8.13.36` part is the actual issue.

Apparently, the first command in [`cargoUpdatePrecise`](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/cargo/artifacts.ts#L39) already updates the package, which causes the second command to fail.

rarkins:
Thanks. Can you either update this PR back to that approach or submit a separate draft PR so one of us can test it locally to see if that's fixable?

DanielleHuisman:
I tried commenting out the first Cargo command and that solves the issue for this particular situation. I'm unsure if that solution works for all situations. However, given the name of the function is `cargoUpdatePrecise` with a comment `// If all dependencies have locked versions then update them precisely.`, I'm wondering why the first step is to perform an "imprecise" update.

DanielleHuisman:
> Thanks. Can you either update this PR back to that approach or submit a separate draft PR so one of us can test it locally to see if that's fixable?

I reverted my commit and made a new one with the `versionOrig` approach. This should trigger the `cargo update` error.

DanielleHuisman:
Closing in favor of #37564.

PR #37562, opened 2025-08-16
feat(preset): add preset semverAllWeekly

cschritt:
## Changes

Added the preset `semverAllWeekly`


viceice (commented):
I don't see much value of those presets

secustor (commented):
This seems to be very specific tailored to your use case, why do you think this should be in the build in presets? 

If it is that about that you do not want to add those extends in every of your repositories, are you aware of self hosted presets?

cschritt:
Thanks for your feedback. I clearly should have explained my motivation for this change. 
I came accross the preset `config:semverAllMonthly` recently and thought that this could also make sense on a weekly schedule. We also already have `schedule:weekly` and `:maintainLockFilesWeekly`, so it might also make sense to be consistent with the 'monthly' presets we already have.

secustor (approved):
I will approve, as it makes sense from a consistency perspective.

PR #37633, opened 2025-08-20
feat(pep621): return found lockfiles

viceice:
## Changes
Managers should return lockfiles, so they're not superceded by other managers.

## Context
- #37458

### DIFF
--- a/lib/modules/manager/pep621/extract.ts
+++ b/lib/modules/manager/pep621/extract.ts
@@ -76,6 +76,7 @@ export async function extractPackageFile(
 
   // process specific tool sets
   let processedDeps = deps;
+  let lockFiles: string[] = [];
   for (const processor of processors) {
     processedDeps = processor.process(def, processedDeps);
     processedDeps = await processor.extractLockedVersions(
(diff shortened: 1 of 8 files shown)


rarkins (approved):

lib/modules/manager/pep621/extract.ts
```diff
@@ -76,17 +76,26 @@ export async function extractPackageFile(
...
+  let lockFiles: string[] = [];
   for (const processor of processors) {
     processedDeps = processor.process(def, processedDeps);
     processedDeps = await processor.extractLockedVersions(
       def,
       processedDeps,
       packageFile,
     );
+    if (processor.getLockfiles) {
+      lockFiles = await processor.getLockfiles(def, lockFiles, packageFile);
```
secustor, line 88: The logic of adding the lockfiles shouldn't be in the processors IMO, so we are not duplicating logic.

secustor (commented):
I haven't been aware that superseding managers are a thing and that the returned lockfiles are used in that context. 

We should add this to https://github.com/renovatebot/renovate/blob/main/docs%2Fdevelopment%2Fadding-a-package-manager.md

lib/modules/manager/pep621/extract.ts
viceice: I'm not sure what you mean here 🤔

secustor: I have been referring to something like this https://github.com/renovatebot/renovate/pull/37643

PR #37643, opened 2025-08-21
refactor(manager/pep621): use abstract class for processors

secustor:
## Changes

Refactoring processors to use abstract class for better re-usability of the `getLockFiles` method. 

### DIFF
--- a/lib/modules/manager/pep621/processors/abstract.ts
+++ b/lib/modules/manager/pep621/processors/abstract.ts
@@ -1,0 +1,54 @@
+import { logger } from '../../../../logger';
+import { getSiblingFileName, localPathExists } from '../../../../util/fs';
+import type {
+  PackageDependency,
+  UpdateArtifact,
+  UpdateArtifactsResult,
+} from '../../types';
+import type { PyProject } from '../schema';
+import type { PyProjectProcessor } from './types';
+
+export abstract class BasePyProjectProcessor implements PyProjectProcessor {
+  // The name of the lockfiles to be searched for.
+  // This should be defined in the concrete processor classes.
+  // No lockfiles will be forwarded outside the manager implementation.
+  lockfileName: string | undefined;
+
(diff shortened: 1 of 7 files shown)


lib/modules/manager/pep621/processors/abstract.ts
```diff
@@ -0,0 +1,54 @@
...
+  UpdateArtifactsResult,
+} from '../../types';
+import type { PyProject } from '../schema';
+import type { PyProjectProcessor } from './types';
+
+export abstract class BasePyProjectProcessor implements PyProjectProcessor {
+  // The name of the lockfiles to be searched for.
+  // This should be defined in the concrete processor classes.
+  // No lockfiles will be forwarded outside the manager implementation.
+  lockfileName: string | undefined;
```
viceice:
```suggestion
  protected lockfileName: string | undefined;
```

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -148,20 +154,6 @@ export class UvProcessor implements PyProjectProcessor {
     return Promise.resolve(deps);
   }
 
-  async getLockfiles(
-    _project: PyProject,
-    lockfiles: string[],
-    packageFile: string,
-  ): Promise<string[]> {
-    const lockFileName = await findLocalSiblingOrParent(packageFile, 'uv.lock');
```
viceice, line 156: we need to override because base class function isn't using the find local sibling or parent. UV supports workspaces with multiple projects and single lock file like pnpm...

secustor: hmm, I switched to use `findLocalSiblingOrParent` by default. I think this should be fine as this can only lead to issues if people would nest multiple projects inside each other and fail to commit the lockfile 🤔

viceice (approved):

PR #37651, opened 2025-08-21
fix(logging): redact JWT-like usernames in logs

IceHamster:
## Changes

- Add regex-based detection for JWT-like strings (header.payload.signature format).
- When sanitizing objects, redact the username field if it contains a JWT.
- Add a unit test to verify redaction works as expected.

## Context

Some registries (e.g., [NGINX Plus](https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-docker/#set-up-docker-for-the-f5-container-registry)) require using a JWT as the username when logging in:
```sh
docker login private-registry.nginx.com \
  --username=<JWT_FROM_MYF5> \
  --password=none
```
Our Renovate config mirrors this via hostRules:
```js
{
  "hostRules": [
    {
      "hostType": "docker",
      "matchHost": "private-registry.nginx.com",
      "username": "<output_of_jwt_token>",
      "password": "none"
    }
  ]
}
```
Because Renovate normally assumes username is a non-secret identifier, it was not sanitized by default.
This leads to JWT credentials being logged in plain text. This PR adds special handling to redact JWT-like values when they appear in username.


viceice (commented):
please describe why the username field isn't already sanitized. how do you configure the token?

IceHamster:
@viceice I have updated the _Context_ in the PR Description.

viceice (changes requested):
then this should be done somewhere here

https://github.com/renovatebot/renovate/blob/7cf05ac34dd63a970e0af11327d2fc44b2609e1d/lib/util/host-rules.ts#L66

so the username is added to sanitizer list

lib/logger/utils.ts
```diff
@@ -216,6 +216,9 @@ export function sanitizeValue(
     return arrayResult;
   }
 
+  // Define a regular expression to detect the JWT format (header.payload.signature)
+  const jwtRegex = /^[\w-]+\.[\w-]+\.[\w-]+$/;
```
looperunbalanced: This pattern would yield a significant amount of false positives, doesn't it?

IceHamster:
> then this should be done somewhere here
> 
> https://github.com/renovatebot/renovate/blob/7cf05ac34dd63a970e0af11327d2fc44b2609e1d/lib/util/host-rules.ts#L66
> 
> so the username is added to sanitizer list

Thanks for the suggestion! That's a good idea, but i think there's a timing issue here.

The raw configuration object gets logged at the debug level before the host-rules.ts logic has a chance to parse the rules and add the username to the sanitizer. So the JWT is already in the logs by the time that code runs.

Here's a snippet from the logs that shows this:

<details>
<summary>Click to expand debug log</summary>
<code>
DEBUG: Using RE2 regex engine
DEBUG: Parsing configs
# ...
DEBUG: File config
       "config": {
         "hostRules": [
           {
             "hostType": "docker",
             "matchHost": "private-registry.nginx.com",
             "username": "ey......",
             "password": "***********"
           },
         ]
       }
# ...
DEBUG: Found valid git version: 2.51.0
DEBUG: Setting global hostRules
DEBUG: Adding password authentication for private-registry.nginx.com (hostType=docker) to hostRules
</code>
</details>

lib/logger/utils.ts
IceHamster: Good point, thank you. The previous logic could cause false positives.

I've pushed a commit that tightens the validation. The new logic uses a more specific regex and adds a minimum length check, since JWTs are almost always significantly longer than usernames.

lib/logger/utils.ts
```diff
@@ -216,6 +216,9 @@ export function sanitizeValue(
     return arrayResult;
   }
 
+  // Define a regular expression to detect the JWT format (header.payload.signature)
+  const jwtRegex = /^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*$/;
```
looperunbalanced, line 220: Are you deliberately making the signature optional?

lib/logger/utils.ts
```diff
@@ -230,6 +233,13 @@ export function sanitizeValue(
         } else {
           curValue = '***********';
         }
+      } else if (
+        key === 'username' &&
+        val.length > 60 &&
+        is.string(val) &&
```
looperunbalanced: Shouldn't these conditions be in the opposite order?

lib/logger/utils.ts
IceHamster: Yes i am.
It would probably never happen, but if "alg" is set to none, we get an JWT token in this format `xxx.yyy.`

lib/logger/utils.ts
IceHamster: Yeah, you are probably right that it would be better to check that the value is an string, before checking
the length. Will push an fix.

viceice:
it needs to be done there too, otherwise it could be logged by other things when host rules are later used.

lib/logger/utils.ts
viceice, line 220:
```suggestion
  const jwtRegex = regEx(/^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*$/);
```

viceice (commented):
please also add it on host rules add function

RahulGautamSingh (approved):
LGTM except host rules comment

lib/util/host-rules.ts
```diff
@@ -38,6 +38,8 @@ export function migrateRule(rule: LegacyHostRule & HostRule): HostRule {
   return result;
 }
 
+const jwtRegex = /^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*$/;
```
astellingwerf: I imagine we ought to use re2 here instead.

astellingwerf (changes requested):

lib/util/host-rules.ts
```diff
@@ -63,6 +66,16 @@ export function add(params: HostRule): void {
       sanitize.addSecretForSanitizing(secret);
     }
   });
+
+  // Sanitize username if it looks like a JWT token
+  if (
+    is.string(rule.username) &&
+    rule.username.length > 60 &&
+    jwtRegex.test(rule.username)
```
viceice: I'm thinking if should always redact usernames,  just in case there are other formats

IceHamster: @viceice We could then add `username` to `confidentialFields`, so we do not need extra logic to mask the username.

Would it create other issues if we mask the username?

viceice: it wouldn't cause issues. please add to `confidentialFields`

IceHamster: @viceice I have added username to confidentialFields.

lib/logger/utils.ts
RahulGautamSingh, line 1: Is this special handling needed anymore? As the username is already confidential now

IceHamster: They are logged before they are set to "confidential"
But we can mask any username, and not only JWT tokens

RahulGautamSingh: Yes, let be consistent

RahulGautamSingh (approved):

lib/logger/utils.ts
astellingwerf: Michael's comment wasn't resolved yet...

astellingwerf (changes requested):

lib/logger/utils.ts
RahulGautamSingh: Weird, it showed resolved on my screen 🤨

lib/logger/utils.ts
viceice: add to `redactedFields` list instead

viceice (commented):
see open discussion

PR #37663, opened 2025-08-22
feat(copier): accept git URLs as source in copier answers files

johannesschillinger-dm:
## Changes

Enhances the copier manager to accept git ssh URLs in addition to the currently accepted "normal" URLs

## Context

- adresses discussion #33650


lib/modules/manager/copier/extract.spec.ts
```diff
@@ -20,6 +20,24 @@ describe('modules/manager/copier/extract', () => {
         ],
       });
     });
+    it('extracts repository and version from .copier-answers.yml with ssh URL', () => {
```
RahulGautamSingh, line 24:
```suggestion

    it('extracts repository and version from .copier-answers.yml with ssh URL', () => {
```

lib/modules/manager/copier/schema.ts
```diff
@@ -1,10 +1,13 @@
 import { z } from 'zod';
 import { Yaml } from '../../../util/schema-utils';
 
+const GitSshUrl = z.string().regex(/^git@[^:]*:(?<repository>.+)$/, {
```
RahulGautamSingh:
```suggestion
const GitSshUrl = z.string().regex(regEx(/^git@[^:]*:(?<repository>.+)$/), {
```

Should we not also check for `.git` at the end?

RahulGautamSingh (changes requested):

lib/modules/manager/copier/schema.ts
johannesschillinger-dm: good point @RahulGautamSingh , I added the .git$ part to the regex

Could you please explain the suggested addition of regEx() ?

lib/modules/manager/copier/schema.ts
```diff
@@ -1,10 +1,13 @@
 import { z } from 'zod';
 import { Yaml } from '../../../util/schema-utils';
 
+const GitSshUrl = z.string().regex(/^git@[^:]*:(?<repository>.+).git$/, {
```
viceice:
```suggestion
const GitSshUrl = z.string().regex(regEx(/^git@[^:]+:(?<repository>.+).git$/), {
```

lib/modules/manager/copier/extract.spec.ts
```diff
@@ -21,6 +21,25 @@ describe('modules/manager/copier/extract', () => {
...
+      const content = `
+        _commit: v1.0.0
+        _src_path: git@github.com:renovatebot/somedir/renovate.git
+      `;
+      const result = extractPackageFile(content);
+      expect(result).toEqual({
+        deps: [
+          {
+            depName: 'git@github.com:renovatebot/somedir/renovate.git',
```
viceice, line 33: I think this can be messaged further to use GitHub tags datasource.
we do such things in other managers already.
can of cause be done in a separate PR

johannesschillinger-dm: If I understand it correctly, that would be Github specific, while this feature aims to support all SSH based git Servers.
Therefore I do not intend to include it in this PR

lib/modules/manager/copier/schema.ts
RahulGautamSingh: we use the node/re2 library to safeguard regexes. the `regEx` is wrapper for it

johannesschillinger-dm: Thank you for the explanation - added the sanitation.

johannesschillinger-dm: A bit of research into the git ssh protocol led me to the following changes:
- remove hard-coded git username - all usernames are valid, git is just a popular default
- remove hard-coded .git repo ending - it is common and recommended to have a bare repo (just the .git folder) as a remote, but it is not required.

RahulGautamSingh:
Could you please test it on a test repo to be sure

johannesschillinger-dm:
> Could you please test it on a test repo to be sure

I ran this build on a company internal repo with a copier template specified by its SSH Git URL.
It worked as expected, detected the copier template and compared its version to what's available in our SCM.

I can provide highly redacted log files, or if you can envision an easily replicable public test case I'd be happy to run it.

RahulGautamSingh (approved):

johannesschillinger-dm:
@viceice I reckon this is ready to merge.

If you see anything that hinders it, please let me know

lib/modules/manager/copier/schema.ts
```diff
@@ -1,10 +1,16 @@
 import { z } from 'zod';
+import { regEx } from '../../../util/regex';
 import { Yaml } from '../../../util/schema-utils';
 
+const GitSshUrl = z
+  .string()
+  .regex(regEx(/^(?<username>[^@]+)@[^:]*:(?<repository>.+)$/), {
```
viceice:
```suggestion
  .regex(regEx(/^[^@]+@[^:]*:.+$/), {
```
i don't see a benefit to capture groups, they're not used later

johannesschillinger-dm: I'm a fan as they support readability later on, but won't argue about them. gone

RahulGautamSingh (approved):

viceice (approved):

PR #37678, opened 2025-08-24
refactor(gitlab): use `http.getJson`

RahulGautamSingh:
## Changes
- Use `http.getJson` instead of `http.getJsonUnchecked`
- For this I had to create schema for the gitlab pull request and update the tests with required field where needed

## Context
- `http.getJsonUnchecked` is not recommended
- In this PR I have moved the requests made in the `pr-cache` file to use `http.getJson` method and will soon do the same for the rest of the methods in the `index.ts` and `merge-requests.ts` file


lib/modules/platform/gitlab/schema.ts
```diff
@@ -7,3 +8,40 @@ export const LastPipelineId = z
...
   .transform(({ last_pipeline }) => last_pipeline.id);
+
+const GitlabUserSchema = z.object({
+  id: z.number(),
+  username: z.string(),
+});
+
+const LongCommitShaSchema = z
+  .string()
+  .length(40)
```
viceice: we should also support the new sha256 repos

RahulGautamSingh: You mean add support for char length `64`

viceice: yes, if they've that length 😅.

forgejo already support those repos

lib/modules/platform/gitlab/utils.ts
```diff
@@ -22,6 +23,7 @@ export const defaults = {
 export function prInfo(mr: GitLabMergeRequest): GitlabPr {
   const pr: GitlabPr = {
     sourceBranch: mr.source_branch,
+    targetBranch: mr.target_branch,
```
viceice, line 26: looks like a fixed bug to me 😅

lib/modules/platform/gitlab/schema.ts
```diff
@@ -7,3 +8,37 @@ export const LastPipelineId = z
...
+  merge_status: z.string().optional(),
+  assignee: GitlabUserSchema.optional(),
+  assignees: z.array(GitlabUserSchema).optional(),
+  reviewers: z.array(GitlabUserSchema).optional(),
+  labels: z.array(z.string()).optional(),
+  sha: LongCommitShaSchema.optional(),
+  head_pipeline: z
+    .object({
+      status: z.string(),
+      sha: z.string(),
```
viceice: shouldn't this also be long commit sha type?

lib/modules/platform/gitlab/utils.ts
RahulGautamSingh: Should I separate this out? I would like to create a new PR with this one and also removing `diverged_commits_count` field which seems unused

lib/modules/platform/gitlab/schema.ts
RahulGautamSingh: Yes

lib/modules/platform/gitlab/schema.ts
secustor: We should have this a shared schema as these should be the same on all platforms.

lib/modules/platform/gitlab/utils.ts
secustor: Yes, please separate it

viceice (commented):
separate PR for the fix, otherwise LGTM

lib/modules/platform/gitlab/utils.ts
RahulGautamSingh: https://github.com/renovatebot/renovate/pull/38385

lib/modules/platform/gitlab/schema.ts
RahulGautamSingh: I will also separate this in a separate PR then as it spans to all platforms and keep the schema flexible in this one as a string only.

RahulGautamSingh: image

pr gonna be big

viceice (approved):

lib/modules/platform/gitlab/schema.ts
```diff
@@ -7,3 +8,37 @@ export const LastPipelineId = z
...
+  title: z.string(),
+  description: z.string(),
+  state: z.string(),
+  source_branch: z.string(),
+  target_branch: z.string(),
+  created_at: z.string(),
+  updated_at: z.string(),
+  diverged_commits_count: z.number().optional(),
+  merge_status: z.string().optional(),
+  assignee: GitlabUserSchema.optional(),
```
Persi, line 30: Before adding this scheme the field `assignee` could be undefined/null. This seems to be changed here, as some of our projects fail with the following error now:

```
         "message": "Schema error",
         "stack": "ZodError: Schema error\n    at Object.get error [as error] (/usr/local/renovate/node_modules/.pnpm/zod@3.25.76/node_modules/zod/v3/types.cjs:45:31)\n    at ZodArray.parseAsync (/usr/local/renovate/node_modules/.pnpm/zod@3.25.76/node_modules/zod/v3/types.cjs:183:22)\n    at processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at GitlabHttp.requestJson (/usr/local/renovate/lib/util/http/http.ts:364:18)\n    at GitlabPrCache.sync (/usr/local/renovate/lib/modules/platform/gitlab/pr-cache.ts:134:29)\n    at Function.init (/usr/local/renovate/lib/modules/platform/gitlab/pr-cache.ts:64:7)\n    at Function.getPrs (/usr/local/renovate/lib/modules/platform/gitlab/pr-cache.ts:81:21)\n    at getPrList (/usr/local/renovate/lib/modules/platform/gitlab/index.ts:505:10)\n    at Proxy.findPr (/usr/local/renovate/lib/modules/platform/gitlab/index.ts:889:18)\n    at closedPrExists (/usr/local/renovate/lib/workers/repository/onboarding/branch/check.ts:48:6)\n    at isOnboarded (/usr/local/renovate/lib/workers/repository/onboarding/branch/check.ts:88:30)\n    at checkOnboardingBranch (/usr/local/renovate/lib/workers/repository/onboarding/branch/index.ts:36:27)\n    at getRepoConfig (/usr/local/renovate/lib/workers/repository/init/config.ts:13:12)\n    at initRepo (/usr/local/renovate/lib/workers/repository/init/index.ts:58:12)\n    at Object.renovateRepository (/usr/local/renovate/lib/workers/repository/index.ts:70:14)\n    at attributes.repository (/usr/local/renovate/lib/workers/global/index.ts:184:11)\n    at start (/usr/local/renovate/lib/workers/global/index.ts:169:7)\n    at /usr/local/renovate/lib/renovate.ts:19:22",
         "issues": {
           "0": {"assignee": "Expected object, received null"},
```

To be honest, i would expect to allow optional fields to be undefined/null, but our stacktraces looks different.

RahulGautamSingh: Need to add nullable here and this will be solved. Also this PR has been reverted and a new will PR will replace it

RahulGautamSingh: @Persi Do you know where I can find the accurate type definitions of these fields which are part of the merge request response from Gitlab? 

I looked at the https://docs.gitlab.com/api/merge_requests/#list-merge-requests api doc but this doesn't include that `assignee` can be `null`. And, this will become an issue as we try to enforce zod schemas.

jamietanna: https://github.com/jdalrymple/gitbeaker/blob/1724338a8b2bb195a0e90868fa35ef37d29f8ae3/packages/core/src/resources/MergeRequests.ts#L104 seems to indicate this may be expected - I guess we assume that an `array` could be either `[]` or `null`?

RahulGautamSingh: Thanks!

PR #37716, opened 2025-08-26
feat: Add datasource + manager for devbox itself

moimael:
## Changes

Adds a Devbox datasource + manager for Devbox itself. Currently, we only handle devbox dependencies in devbox.json, but not devbox itself.

## Context

## AI assistance disclosure

Claude 4 sonnet was used to generate most of the code. I double-checked the code, ran the tests, and tested renovate manually to make sure it worked as expected.

Result of running the dev version dry-run on our repo:
```
DEBUG: branches info extended (repository=alan-eu/alan-apps)
       "branchesInformation": [
         {
           "branchName": "renovate/devbox-0.x",
           "prNo": null,
           "prTitle": "chore(deps): update dependency devbox to v0.16.0",
           "result": "needs-approval",
           "upgrades": [
             {
               "datasource": "devbox-version",
               "depName": "devbox",
               "displayPending": "",
               "fixedVersion": "0.15.0",
               "currentVersion": "0.15.0",
               "currentValue": "0.15.0",
               "newValue": "0.16.0",
               "newVersion": "0.16.0",
               "packageFile": ".devbox-version",
               "updateType": "minor",
               "packageName": "devbox"
             }
           ]
         }
       ]
```


moimael:
@viceice is this feature something you'd consider merging?

lib/modules/datasource/devbox-version/index.ts
```diff
@@ -0,0 +1,63 @@
...
+        return null;
+      }
+
+      res.releases.push({ version });
+    } catch (err) {
+      if (err instanceof HttpError) {
+        if (err.response?.statusCode !== 404) {
+          throw new ExternalHostError(err);
+        }
+      }
```
RahulGautamSingh:
```suggestion
      if (err instanceof HttpError && err.response?.statusCode !== 404) {
          throw new ExternalHostError(err);
      }
```

RahulGautamSingh (changes requested):
Does the datasource provide all releases? If so we should use that endpoint cause users might want to upgrade to a new version which is not the latest?

moimael:
> Does the datasource provide all releases? If so we should use that endpoint cause users might want to upgrade to a new version which is not the latest?

I'm not aware of any other endpoint; this is the one used by the devbox updater itself. Maybe @burritobill could chime in here?

lib/modules/manager/devbox-version/extract.ts
```diff
@@ -0,0 +1,13 @@
+import { logger } from '../../../logger';
+import { DevboxVersionDatasource } from '../../datasource/devbox-version';
+import type { PackageDependency, PackageFileContent } from '../types';
+
+export function extractPackageFile(content: string): PackageFileContent {
+  logger.trace('devbox-version.extractPackageFile()');
+  const dep: PackageDependency = {
+    depName: 'devbox',
+    currentValue: content.trim(),
+    datasource: DevboxVersionDatasource.id,
```
viceice: why not simply using `github-releases` datasource?

https://github.com/jetify-com/devbox/releases

lib/modules/manager/devbox-version/index.ts
```diff
@@ -0,0 +1,15 @@
+import type { Category } from '../../../constants';
+import { DevboxVersionDatasource } from '../../datasource/devbox-version';
+import * as semverVersioning from '../../versioning/semver';
+
+export { extractPackageFile } from './extract';
+
+export const displayName = '.devbox-version';
```
viceice: that's a bad display name. shouldn't start with a dot. 🤔

lib/modules/manager/devbox-version/index.ts
```diff
@@ -0,0 +1,15 @@
...
+export const categories: Category[] = ['custom'];
```
viceice: drop this, `custom` is for custom managers only

lib/modules/manager/devbox-version/index.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const defaultConfig = {
+  managerFilePatterns: ['/(^|/)\\.devbox-version$/'],
+  versioning: semverVersioning.id,
```
viceice: should be on datasource if a new datasource is required

viceice (changes requested):

lib/modules/manager/devbox-version/extract.ts
moimael: Thanks, I'll have a look at using this!

lib/modules/manager/devbox-version/readme.md
```diff
@@ -0,0 +1,9 @@
+This manager maintains `.devbox-version` files. Available versions will be determined from the official Devbox releases endpoint.
+
+The `.devbox-version` file contains a single version string that specifies which version of Devbox to use for the project. This manager will check for newer versions and propose updates when available.
```
burritobill, line 3: Is this supported by devbox? Can you provide a link to where this is documented or implemented in the devbox source code?

I was only aware of the environment variable `DEVBOX_USE_VERSION`.

moimael: Well, you are completely right. I didn't realize we had a custom implementation for this (we actually load the file content in the ENV var using a custom direnv script). Would you be open to supporting some kind of file based versioning? It would make it a lot easier to be able to handle devbox updates.

burritobill: I don't work for Jetify so you would need to put it in as a feature request with them.

We only use devbox in local development and allow our engineers to upgrade ad hoc.

Other than the recent lockfile changes have you had any issues which have required your engineers to be on the same versions of devbox?

PR #37768, opened 2025-08-29
fix(gitea): allow fallbacks in case default merge style not allowed

dh-seanmurphy:
## Changes

Asked via discussion [here](https://github.com/renovatebot/renovate/discussions/37665).

As noted in #36779, the default merge style in Gitea is always set, but it's possible that it is set to a merge style which is not allowed for the repo. Currently, Renovate blocks the repo in this case.

Gitea's handing of this is less strict; if the default merge style is not an allowed merge style, Gitea suggests one of the allowed merge styles based on an ordered list.

This change mirrors the behaviour of Gitea by permitting fallbacks in case the default merge style is not an allowed merge style; the fallback is selected from an ordered list, using the same ordering as Gitea. If the default merge style is an allowed style, the behaviour is unchanged.

## Context

## AI assistance disclosure

I used ChatGPT to generate a first version of this and iterated on it. I obtained human feedback and then used ChatGPT to help with refactoring. Modifications to tests were performed manually.

### DIFF
--- a/lib/modules/platform/gitea/index.ts
+++ b/lib/modules/platform/gitea/index.ts
@@ -323,28 +323,35 @@ const platform: Platform = {
       throw new Error(REPOSITORY_BLOCKED);
     }
 
-    if (repo.allow_rebase && repo.default_merge_style === 'rebase') {
-      config.mergeMethod = 'rebase';
-    } else if (
-      repo.allow_rebase_explicit &&
-      repo.default_merge_style === 'rebase-merge'
-    ) {
-      config.mergeMethod = 'rebase-merge';
-    } else if (
-      repo.allow_squash_merge &&
-      repo.default_merge_style === 'squash'
-    ) {
-      config.mergeMethod = 'squash';
-    } else if (
-      repo.allow_merge_commits &&
-      repo.default_merge_style === 'merge'
-    ) {
-      config.mergeMethod = 'merge';
-    } else if (
-      repo.allow_fast_forward_only_merge &&
-      repo.default_merge_style === 'fast-forward-only'
-    ) {
-      config.mergeMethod = 'fast-forward-only';
+    const giteaOrdering: PRMergeMethod[] = [
+      'merge',
+      'rebase',
+      'rebase-merge',
+      'squash',
+      'fast-forward-only',
+    ];
+
+    const candidateMergeStyles = Array.from(
+      new Set<PRMergeMethod>([repo.default_merge_style, ...giteaOrdering]),
+    );
+
+    let selectedMergeStyle: PRMergeMethod | undefined;
+
+    for (const s of candidateMergeStyles) {
+      if (
+        (s === 'merge' && repo.allow_merge_commits) ||
+        (s === 'rebase' && repo.allow_rebase) ||
+        (s === 'rebase-merge' && repo.allow_rebase_explicit) ||
+        (s === 'squash' && repo.allow_squash_merge) ||
+        (s === 'fast-forward-only' && repo.allow_fast_forward_only_merge)
+      ) {
+        selectedMergeStyle = s;
+        break;
+      }
+    }
+
+    if (selectedMergeStyle) {
+      config.mergeMethod = selectedMergeStyle;
     } else {
       logger.debug(
         'Repository has no allowed merge methods - aborting renovation',
--- a/lib/modules/platform/gitea/index.ts
+++ b/lib/modules/platform/gitea/index.ts
@@ -57,6 +57,7 @@ import {
   DRAFT_PREFIX,
   getMergeMethod,
   getRepoUrl,
+  isAllowed,
   smartLinks,
   toRenovatePR,
   trimTrailingApiPath,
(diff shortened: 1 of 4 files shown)


lib/modules/platform/gitea/index.ts
```diff
@@ -323,28 +323,42 @@ const platform: Platform = {
...
-      repo.allow_fast_forward_only_merge &&
-      repo.default_merge_style === 'fast-forward-only'
-    ) {
-      config.mergeMethod = 'fast-forward-only';
+    function throwUnknownMergeStyle(_: never): never {
+      logger.debug('Repo has unknown merge style - aborting renovation');
+      throw new Error(REPOSITORY_BLOCKED);
+    }
+
+    function isAllowed(style: PRMergeMethod): boolean {
```
viceice: no nested functions, move to util or common file

lib/modules/platform/gitea/index.ts
```diff
@@ -323,28 +323,42 @@ const platform: Platform = {
...
+
+    // mirror behaviour of gitea - if default merge style is allowed, use this;
+    // else fall back to predefined order
+    const preferredOrder: PRMergeMethod[] = [
+      repo.default_merge_style,
+      'rebase',
+      'squash',
+      'merge',
+      'rebase-merge',
+      'fast-forward-only',
```
viceice: we should prefer ff before merge

viceice (changes requested):

lib/modules/platform/gitea/index.ts
dh-seanmurphy: Moved to `utils.ts`.

lib/modules/platform/gitea/index.ts
dh-seanmurphy: This is the order used by Gitea afaict:

image

We were just trying to replicate that behaviour - is this reasonable?

dh-seanmurphy:
Thanks for your speedy reaction here @viceice

lib/modules/platform/gitea/utils.ts
```diff
@@ -169,3 +172,24 @@ export function usableRepo(repo: Repo): boolean {
...
+    case 'rebase':
+      return repo.allow_rebase;
+    case 'rebase-merge':
+      return repo.allow_rebase_explicit;
+    case 'squash':
+      return repo.allow_squash_merge;
+    case 'fast-forward-only':
+      return repo.allow_fast_forward_only_merge;
+  }
+  return throwUnknownMergeStyle(style); // ensures changes to PRMergeMethod updated here
```
RahulGautamSingh:
```suggestion
  logger.debug('Repo has unknown merge style - aborting renovation');
  throw new Error(REPOSITORY_BLOCKED);
```

lib/modules/platform/gitea/utils.ts
```diff
@@ -169,3 +172,24 @@ export function usableRepo(repo: Repo): boolean {
   }
   return true;
 }
+
+function throwUnknownMergeStyle(_: never): never {
+  logger.debug('Repo has unknown merge style - aborting renovation');
+  throw new Error(REPOSITORY_BLOCKED);
+}
+
```
RahulGautamSingh:
```suggestion
```

RahulGautamSingh (changes requested):
This changes the behavior. Please also verify it works as expected by running against gitea repos

dh-seanmurphy:
> This changes the behavior. Please also verify it works as expected by running against gitea repos

@RahulGautamSingh - I'm not sure if you have specific testing in mind. I ran this against a test repo using the command:

```
LOG_LEVEL=debug pnpm start --platform gitea --endpoint $TEST_GITEA_INSTANCE --pr-concurrent-limit 4 --autodiscover false --pr-hourly-limit 5 $REPO 
```

I performed the following 4 tests:
- all merge methods allowed, default PR style=merge - PR created with a merge style (ie option to merge is Merge Commit)
- all merge methods allowed, default PR style=squash - PR created with a squash style (ie option to merge is Squash Commit)
- merge methods merge and squash disabled (others allowed), default PR style=squash - PR created with Rebase then Fast Forward style
- all merge methods disabled, default merge style=squash - repo is blocked

I did not perform any testing with automerging.

You noted that the behaviour is different; the current codebase would have differed only for the third of the above tests, blocking the repo due to the mismatch between default and allowed merge styles.

Probably worth noting that in gitea, the default merge method applies to existing open PRs on a repo - it seems to be independent of how they are created, ie if a PR is created with one default merge method and the default merge method is changed (to another allowed method), then gitea recommends using the new default merge method on open PRs.

Please let me know if you need me to perform further testing on this - thanks!

RahulGautamSingh:
Thanks for the tests they should be sufficient.

dh-seanmurphy:
> Thanks for the tests they should be sufficient.

If there is anything else you need me to do here, please let me know - thanks.

RahulGautamSingh (approved):

lib/modules/platform/gitea/index.ts
dh-seanmurphy: @viceice - is the above reasonable from your side? Can I mark this as resolved?

viceice: I still think we should try our own order to have as less as possible commits when auto merging.
so my preference would be:
- ff
- squash
- merge
- rebase
- rebase-merge

dh-seanmurphy: This is ok from my side; from my side, the main point is to have a fallback (as currently we are blocked on renovate upgrades due to this issue).

Should this be documented somewhere? If the behaviour is the same as gitea, then it's reasonably intuitive and prob not necessary to document; if there is an alternative behaviour which is non-obvious, it probably makes sense to have it documented imo...

dh-seanmurphy: I've modified the fallback order as suggested by @viceice (and updated the comment in the code); I've also added a commit which adds a note to the documentation.

Please let me know what you think - thanks.

dh-seanmurphy: @viceice - any chance you might have a few cycles to look through this?

viceice (approved):
can you please port it to the forgejo platform code too? can be a separate PR 🤗

viceice:
please resolve conversations to get it merged 😅

dh-seanmurphy:
> can you please port it to the forgejo platform code too? can be a separate PR 🤗

I can try - we don't use forgejo, so I don't know if these aspects are exactly the same but I'll look into it.

viceice:
> > can you please port it to the forgejo platform code too? can be a separate PR 🤗
> 
> I can try - we don't use forgejo, so I don't know if these aspects are exactly the same but I'll look into it.
> 

they are still same there, you can also disable the default merge style.

dh-seanmurphy:
> > > can you please port it to the forgejo platform code too? can be a separate PR 🤗
> > 
> > 
> > I can try - we don't use forgejo, so I don't know if these aspects are exactly the same but I'll look into it.
> 
> they are still same there, you can also disable the default merge style.

Will do it later this week.

PR #37779, opened 2025-08-31
feat(manager/maven-wrapper): extract wrapper version from mvnw script file

RahulGautamSingh:
## Changes
- Add one more regex to extract the rapper from the version string in the `mvwn(.cmd)` script file

## Context
- Ref: #37773 and https://github.com/renovatebot/renovate/discussions/28648#discussioncomment-10939255

- [x] This closes an existing Issue: #37773

## AI assistance disclosure

The public repository: <URL>


RahulGautamSingh:
No snapshots added still the ci check for added snaps fails 🧐

RahulGautamSingh:
@mvitz and @breun Could you please take a look?

breun:
@RahulGautamSingh Thanks for your PR. I'm not a TypeScript developer, but the change looks good to me!

lib/modules/manager/maven-wrapper/index.ts
```diff
@@ -9,7 +9,10 @@ export const url = 'https://maven.apache.org/wrapper';
 export const categories: Category[] = ['java'];
 
 export const defaultConfig = {
-  managerFilePatterns: ['/(^|\\/).mvn/wrapper/maven-wrapper.properties$/'],
+  managerFilePatterns: [
+    '/(^|\\/).mvn/wrapper/maven-wrapper.properties$/',
+    '/^mvnw(.cmd)?/',
```
viceice:
```suggestion
    '/^mvnw(.cmd)?$/',
```

lib/modules/manager/maven-wrapper/extract.spec.ts
```diff
@@ -63,6 +65,39 @@ describe('modules/manager/maven-wrapper/extract', () => {
       ]);
     });
 
+    it('extracts maven warapper version from mvwn file', () => {
```
astellingwerf:
```suggestion
    it('extracts maven wrapper version from mvnw file', () => {
```

lib/modules/manager/maven-wrapper/extract.spec.ts
```diff
@@ -63,6 +65,39 @@ describe('modules/manager/maven-wrapper/extract', () => {
...
+          replaceString: '3.3.0',
+          datasource: 'maven',
+          depName: 'maven-wrapper',
+          packageName: 'org.apache.maven.wrapper:maven-wrapper',
+          versioning: 'maven',
+        },
+      ]);
+    });
+
+    it('extracts maven warapper version from mvwn file - REM', () => {
```
astellingwerf:
```suggestion
    it('extracts maven wrapper version from mvnw file - Windows', () => {
```

lib/modules/manager/maven-wrapper/extract.spec.ts
```diff
@@ -63,6 +65,39 @@ describe('modules/manager/maven-wrapper/extract', () => {
...
+          replaceString: '3.3.0',
+          datasource: 'maven',
+          depName: 'maven-wrapper',
+          packageName: 'org.apache.maven.wrapper:maven-wrapper',
+          versioning: 'maven',
+        },
+      ]);
+    });
+
+    it('returns null for invalid wrapper version string in from mvwn file', () => {
```
astellingwerf:
```suggestion
    it('returns null for invalid wrapper version string in from mvnw file', () => {
```

lib/modules/manager/maven-wrapper/extract.spec.ts
```diff
@@ -63,6 +65,39 @@ describe('modules/manager/maven-wrapper/extract', () => {
...
+      const res = extractPackageFile(`invalid ${mvwnWrapperString}`);
```
astellingwerf:
```suggestion
      const res = extractPackageFile(`invalid ${mvnwWrapperString}`);
```

lib/modules/manager/maven-wrapper/extract.spec.ts
```diff
@@ -6,6 +6,8 @@ const onlyMavenProperties =
   'distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.5.4/apache-maven-3.5.4-bin.zip';
 
 const wrapperAndMavenProperties = `distributionUrl=https://internal.artifactory.acme.org/artifactory/maven-bol/org/apache/maven/apache-maven/3.8.4/apache-maven-3.8.4-bin.zip\nwrapperUrl=https://internal.artifactory.acme.org/artifactory/maven-bol/org/apache/maven/wrapper/maven-wrapper/3.1.0/maven-wrapper-3.1.0.jar`;
+const mvwnWrapperString =
```
astellingwerf:
```suggestion
const mvnwWrapperString =
```

astellingwerf (changes requested):
Please correct the PR title accordingly

lib/modules/manager/maven-wrapper/index.ts
astellingwerf:
```suggestion
    '/(^|\\/)mvnw(.cmd)?$/',
```

breun:
I noticed that Maven Wrapper dropped its `wrapperVersion` property when the `only-script` mode is used (https://github.com/apache/maven-wrapper/pull/145), which is now the default mode, so new installations and updates will no longer have the `wrapperVersion` property in their configuration. 🫤

I also often encounter repositories that have `wrapperVersion` set to some version in `.mvn/wrapper/maven-wrapper.properties`, but don’t actually have the `mvnw*` scripts that match that version, probably because someone (or some tool?) thought it was a good idea to bump `wrapperVersion`.

When the Maven Wrapper version can be parsed from the `mvnw*` scripts, then I think Renovate should consider this the most reliable actually used current version. So, more reliable than the value of `wrapperVersion`.

Also, there may also be repositories with outdated `mvnw*` scripts that don’t contain this version and with a `maven-wrapper.properties` file without a `wrapperVersion` property. Renovate won’t be able to determine the current version for those cases, but if possible it would be nice if Renovate would create a PR with an update to the latest version.

RahulGautamSingh:
So how do you think we should proceed?

We will currently update all versions we detect in all matching files.

breun:
> So how do you think we should proceed?
> 
> We will currently update all versions we detect in all matching files.

I’m not sure what this means exactly, and I’m currently on my phone (and I’m not a TypeScript developer), but I think the logic should be like this:

When an outdated Maven Wrapper version is found via any of the detection methods (which are `wrapperVersion` property and regex matches on `mvnw` and `mvnw.cmd`?), Renovate should run `mvn wrapper:${maven_wrapper_new_version}:wrapper` to update.

If the presence of Maven Wrapper is detected, but its version cannot be determined (e.g. old `mvnw*` scripts from before when the version comment line was added and no `wrapperVersion` property), then ideally Renovate should also run `mvn wrapper:${maven_wrapper_new_version}:wrapper` to update. There would be no ‘from’ version for that update, but if that’s a problem maybe Renovate can use a dummy placeholder version like `0.0.0`?

lib/modules/manager/maven-wrapper/extract.spec.ts
```diff
@@ -6,6 +6,8 @@ const onlyMavenProperties =
   'distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.5.4/apache-maven-3.5.4-bin.zip';
 
 const wrapperAndMavenProperties = `distributionUrl=https://internal.artifactory.acme.org/artifactory/maven-bol/org/apache/maven/apache-maven/3.8.4/apache-maven-3.8.4-bin.zip\nwrapperUrl=https://internal.artifactory.acme.org/artifactory/maven-bol/org/apache/maven/wrapper/maven-wrapper/3.1.0/maven-wrapper-3.1.0.jar`;
+const mvnwWrapperString =
+  'Apache Maven Wrapper startup batch script, version 3.3.0';
```
viceice, line 10: this is not the maven version, it's the version of the wrapper file. so we shouldn't extract it if we don't have a datasource for lookup.

https://github.com/apache/netbeans-nbpackage/pull/113/files

viceice (changes requested):
We'll wait for upstream issue decision. it makes no sense to use the version from the wrapper file as it's not the maven version.

- https://github.com/apache/maven-wrapper/issues/364

breun:
> We'll wait for upstream issue decision. it makes no sense to use the version from the wrapper file as it's not the maven version.

I'm not sure what you mean. It's indeed not the Maven version, but the Maven Wrapper version, which is also what this PR is about.

I submitted the PR to Maven Wrapper to put back the `wrapperVersion` configuration property, but when that gets released there will still be many repositories that either don't have that property, or have that property, but without the corresponding versions of the scripts, so I think it's a good idea to merge this change regardless of what happens in Maven Wrapper.

lib/modules/manager/maven-wrapper/extract.spec.ts
RahulGautamSingh: It's not related to the `maven` dep. The datasource is `maven` and dep is `maven-wrapper`

viceice (approved):

PR #37828, opened 2025-09-04
feat(managers): add Deno manager support

### LINKED ISSUE #6237: Support deno

Support Deno as a regular package manager. 

Provide upgrades if someone uses branch versions

**Additional context**

https://deno.land/x/

Hajime-san:
## Changes

Closes https://github.com/renovatebot/renovate/issues/6237

This PR adds deno manager.
Additionally, deno manager also supports `package.json` for  their [node-compat](https://docs.deno.com/runtime/fundamentals/node/).

The Deno CLI itself supports private registries that are compatible with npm registries by reading `.npmrc`.
However, this PR does not implement such support due to I was concerned that the PR would become too large, which is possible by referencing the code of existing npm managers.
I think that the current implementation covers most use cases, and that it will be sufficient to add support for private npm registries later, while addressing any issues that arise in real-world repositories after this PR is shipped.

needs JSR datasource support:
- https://github.com/renovatebot/renovate/pull/37881

needs Deno versioning module:
- https://github.com/renovatebot/renovate/pull/40968

add Deno to [Containerbase](https://github.com/containerbase/base): 
- https://github.com/containerbase/base/pull/4959

## Context

- [x] This closes an existing Issue: #6237

## AI assistance disclosure

- [x] Yes — other (please describe):

I received assistance from AI in various aspects of submitting this PR.
I found it particularly useful in being able to write complex regular expressions and what to do with markdown links in mkdoc.
It also provided advice and code generation on performance optimization, mainly after the code I wrote had passed unit tests.
Therefore, I think of it mainly as acting as a rubber duck and support for the code I wrote.

On the other hand, Renovate's code is quite complex, and I found it difficult to build features from scratch because it would be difficult to reduce the context size.

The tools I used were GitHub Copilot Agent, DeepWiki, and Google Gemini, and the models were Claude Sonnet4 and Gemini 2.5 Pro.

The public repository: https://github.com/Hajime-san/renovate-test3


lib/modules/manager/deno/schema.ts
```diff
@@ -0,0 +1,113 @@
...
+    workspace: z
+      .union([
+        z.array(z.string()),
+        z.object({
+          members: z.array(z.string()),
+        }),
+      ])
+      .optional(),
+  }),
+);
```
zharinov: That's not how we use Zod library. Take a look: https://github.com/renovatebot/renovate/blob/main/docs/development/zod.md

Basically, `extract.ts` must become a thin wrapper around `SomeSchema.parse(...)` call with 10–20 LOCs. Think of it not like a data validation, but data parsing.

zharinov (changes requested):
Thanks for your contribution. Let's have JSR datasource and Deno manager as separate PRs.

lib/modules/manager/deno/schema.ts
zharinov: Example for inspiration:
https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/pixi/extract.ts
https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/pixi/schema.ts

The rule of thumb: everything with side effects resides in `extract.ts`, data transformation lives in `schema.ts`. From what I see, the Deno extract code could be about 200–300 LOCs, not 600.

Hajime-san:
@zharinov 
Thank you.
I'm going to revert the JSR datasource commits from this PR and put them into a new one.
This PR can continue as the Deno datasource.

lib/modules/manager/deno/index.ts
```diff
@@ -0,0 +1,33 @@
...
+    '/(^|/)deno\\.(json|jsonc)$/',
+  ],
+  digest: {
+    prBodyDefinitions: {
+      Change:
+        '{{#if displayFrom}}`{{{displayFrom}}}` -> {{else}}{{#if currentValue}}`{{{currentValue}}}` -> {{/if}}{{/if}}{{#if displayTo}}`{{{displayTo}}}`{{else}}`{{{newValue}}}`{{/if}}',
+    },
+  },
+  prBodyDefinitions: {
+    Change:
```
viceice, line 36: why this?

lib/modules/manager/deno/schema.ts
```diff
@@ -0,0 +1,113 @@
...
+  z.object({
+    imports,
+    scopes,
+  }),
+);
+
+export type ImportMapJsonFile = z.infer<typeof ImportMapJsonFile>;
+
+// https://github.com/denoland/vscode_deno/blob/7e125c6ffcdcdebd587f97be5341d404f5335b87/schemas/lockfile.schema.json
+const DenoLockV5 = z.object({
```
viceice: is this referenced anywhere or can it joined to the Schema below?

lib/modules/manager/deno/update.ts
```diff
@@ -0,0 +1,363 @@
...
+          newFileContent = result;
+        }
+      }
+    }
+  }
+
+  return newFileContent;
+}
+
+export function updateDependency(
```
viceice, line 112: did you tried the default renovate auto replace? is this function really required?

lib/modules/manager/deno/index.ts
Hajime-san: I think the process branches that it is not possible to generate a diff URL, because the source URL is not included in the case of jsr datasource and deno datasource.

examples

npm datasource https://github.com/Hajime-san/renovate-test3/pull/3
jsr datasource https://github.com/Hajime-san/renovate-test3/pull/2

lib/modules/manager/deno/update.ts
Hajime-san: I think so, I got received errors like below when I commented out the `updateDependency`.

```bash
...
DEBUG: Could not extract deno.jsonc (manager=deno) after autoreplace. Did the autoreplace make the file unparseable? (repository=Hajime-san/renovate-test4, branch=renovate/chalk-5.x)
 WARN: Error updating branch: update failure (repository=Hajime-san/renovate-test4, branch=renovate/chalk-5.x)
...
DEBUG: Could not extract node/package.json (manager=deno) after autoreplace. Did the autoreplace make the file unparseable? (repository=Hajime-san/renovate-test4, branch=renovate/mongoose-8.x)
 WARN: Error updating branch: update failure (repository=Hajime-san/renovate-test4, branch=renovate/mongoose-8.x)
...
```

lib/modules/manager/deno/schema.ts
```diff
@@ -0,0 +1,429 @@
...
+    z.object({
+      members: z.array(z.string()),
+    }),
+  ])
+  .optional()
+  .transform((workspaces): string[] | undefined => {
+    if (!workspaces) {
+      return undefined;
+    }
+    return 'members' in workspaces ? workspaces.members : workspaces;
```
zharinov: Type checking shouldn't happen here, instead transform must be done at each branch of union, leading to the uniform result

lib/modules/manager/deno/schema.ts
```diff
@@ -0,0 +1,429 @@
...
+  })
+  .transform((workspaces): DenoManagerData => ({ workspaces }));
+
+export const Imports = LooseRecord(z.string())
+  .catch({})
+  .transform((imports): PackageDependency<DenoManagerData>[] => {
+    const deps = [];
+    for (const depValue of Object.values(imports)) {
+      const dep = DenoDependency.parse({ depValue, depType: 'imports' });
```
zharinov: This one, and any other `DenoDependency.parse()` shouldn't happen inside `.tranform()`, compose schema instead.

lib/modules/manager/deno/schema.ts
```diff
@@ -0,0 +1,429 @@
...
+    fileName: z.string(),
+  })
+  .transform(async (data) => {
+    // handle lock file
+    const { content, fileName } = data;
+
+    let lockFile: string | undefined;
+    if (content.lock) {
+      const lock = content.lock;
+      if (typeof lock === 'string' && (await localPathIsFile(lock))) {
```
zharinov: This schema must be just a function (because `async` transforms are potentially problematic). However, this `typeof` cascade indicates the lack of schema somewhere else.

lib/modules/manager/deno/schema.ts
```diff
@@ -0,0 +1,429 @@
...
+  compilerOptions?: {
+    types?: string[];
+    jsxImportSource?: string;
+    jsxImportSourceTypes?: string;
+  };
+  lint?: {
+    plugins?: string[];
+  };
+  workspace?: string[] | { members: string[] };
+}
```
zharinov: There is no point of using this, those fields must be parsed as schema.

lib/modules/manager/deno/schema.ts
```diff
@@ -0,0 +1,429 @@
...
+  .transform((imports): PackageDependency<DenoManagerData>[] => {
+    const deps = [];
+    for (const depValue of Object.values(imports)) {
+      const dep = DenoDependency.parse({ depValue, depType: 'imports' });
+      deps.push(dep);
+    }
+    return deps;
+  });
+
+export const Scopes = LooseRecord(LooseRecord(z.string()).catch({}))
```
zharinov: There is no point in using internal `.catch()`, because `LooseRecord` rejects wrong values

zharinov (changes requested):
Please implement `schema.ts` without resorting to vibe-coding

lib/modules/manager/deno/schema.ts
Hajime-san: It used in `update.ts`. Should I move the type definition to there?

zharinov: No, `update.ts` must use schema as well, instead of ad-hoc type checks and type asserts

zharinov: Interfaces with all-optionals are basically `any` type with nicer code completion support. While the codebase contains a lot of this due to pre-TypeScript stage of the project, it's discouraged to introduce new types like that.

Hajime-san: IMO, there is little benefit in parsing values ​​that have already been parsed by zod again since `update.ts` is executed after `extract.ts`. It would be better to align the style with the existing code.

zharinov: Ad-hoc nullity checks and typeof checks are parsing, and better done with zod. It's not the right target for perf optimization.

zharinov: It could be, though, the type inferred from schema with properly filled default values (via catch), such that update logic is minimally intertwined with such type/null checks. I.e., once we successfully parsed, we're sure about presence of whatever nested data.

lib/modules/manager/deno/artifacts.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+  newPackageFileContent: '',
+  packageFileName: '',
+  updatedDeps: [],
+};
+
+describe('modules/manager/deno/artifacts', () => {
+  describe('updateArtifacts()', () => {
+    beforeEach(() => {
+      GlobalConfig.set({
+        localDir: '',
```
viceice: use a tmp dir or a fixed path, See other tests with `tmp-promise` usage or fixed path

zharinov (approved):
LGTM, thanks for such a significant amount of work!

ihorkatkov:
Any plans to merge it @viceice? We made our own workarounds, but we would be happy to have an official way

viceice:
I'll try to do some reviews later today

lib/modules/manager/deno/artifacts.ts
```diff
@@ -0,0 +1,113 @@
...
+  logger.debug(`deno.updateArtifacts(${packageFileName})`);
+  const isLockFileMaintenance = config.updateType === 'lockFileMaintenance';
+
+  if (is.emptyArray(updatedDeps) && !isLockFileMaintenance) {
+    logger.debug('No updated deno deps - returning null');
+    return null;
+  }
+
+  // Find the first deno dependency in order to handle mixed manager updates
+  const lockFileName = updatedDeps.find((dep) => dep.manager === 'deno')
```
viceice: I'm pretty sure all updates are already filtered by manager 🤔

lib/modules/manager/deno/artifacts.ts
```diff
@@ -0,0 +1,113 @@
...
+    ?.lockFiles?.[0];
+
+  if (!lockFileName) {
+    logger.debug('No lock file found. Skipping artifact update.');
+    return null;
+  }
+
+  const oldLockFileContent = await readLocalFile(lockFileName);
+  if (!oldLockFileContent) {
+    logger.debug(`Failed to read ${lockFileName}. Skipping artifact update.`);
```
viceice, line 37: this should cause an artifact error return

lib/modules/manager/deno/artifacts.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  for (const updateDep of updatedDeps) {
+    if (
+      updateDep.depType === 'tasks' ||
+      updateDep.depType === 'tasks.command'
+    ) {
+      logger.warn(
+        `depType "${updateDep.depType}" can't be updated with a lock file: ${updateDep.depName}`,
+      );
+      return null;
```
viceice: artifact error return

lib/modules/manager/deno/artifacts.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  try {
+    await writeLocalFile(packageFileName, newPackageFileContent);
+    if (isLockFileMaintenance) {
+      await deleteLocalFile(lockFileName);
```
viceice, line 70: is this needed for deno? some managers don't need it to update the lockfile properly

lib/modules/manager/deno/artifacts.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    // run from its referred deno.json/deno.jsonc location if import map is used
+    const cwdFile =
+      updatedDeps.find((dep) => dep.managerData?.importMapReferrer)?.managerData
```
viceice: this looks strange. double manager data?

lib/modules/manager/deno/artifacts.ts
```diff
@@ -0,0 +1,113 @@
...
+        ?.importMapReferrer ?? packageFileName;
+
+    const execOptions: ExecOptions = {
+      cwdFile,
+      docker: {},
+      toolConstraints: [
+        {
+          toolName: 'deno',
+          constraint: updateArtifact?.config?.constraints?.deno,
```
viceice: is there a way to detect the required version from a repo file?

bartlomieju:
Hi 🖐️ Bartek from the Deno team here. Is there anything we can help to expedite this PR getting merged and released? We'll be happy to do a blog post if there's interest from your side. Thanks!

lib/modules/manager/deno/compat.ts
```diff
@@ -0,0 +1,133 @@
...
+  return extractDenoCompatiblePackageJson(packageFileContent, packageFile);
+}
+
+export function extractDenoCompatiblePackageJson(
+  content: string,
+  packageFile: string,
+): PackageFile<DenoManagerData> | null {
+  let packageJson: NpmPackage;
+  try {
+    packageJson = JSON.parse(content);
```
RahulGautamSingh: use parseJson from utils

lib/modules/manager/deno/compat.ts
```diff
@@ -0,0 +1,133 @@
...
+  packageFile: string,
+): Promise<PackageFile<DenoManagerData> | null> {
+  const packageFileContent = await readLocalFile(packageFile, 'utf8');
+  if (!packageFileContent) {
+    logger.debug({ packageFile }, 'Deno: No package.json found');
+    return null;
+  }
+
+  return extractDenoCompatiblePackageJson(packageFileContent, packageFile);
+}
```
RahulGautamSingh: can't this be done within `extractDenoCompatiblePackageJson` fn itself?

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,297 @@
...
+      );
+      return match?.groups?.currentValue;
+    }
+  }
+
+  // jsr and npm datasource
+  if (datasource === 'jsr' || datasource === 'npm') {
+    if (
+      !lockFileContent.lockedVersions ||
+      is.emptyObject(lockFileContent.lockedVersions)
```
RahulGautamSingh:
```suggestion
      is.nonEmptyObject(lockFileContent.lockedVersions)
```

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,297 @@
...
+    }
+  }
+}
+
+interface WorkspaceContext {
+  lockFiles?: string[];
+  rootDir: string;
+  packageFile: string;
+  matchers: Minimatch[];
+}
```
RahulGautamSingh: keep this in types.ts file

RahulGautamSingh (commented):
Will review again later this day

RahulGautamSingh:
@bartlomieju Thanks for reaching out! The best way to help would be with the review process. This PR introduces quite a few changes, and it’s easy for things to slip through. Having your expertise with Deno involved in the review would be very valuable and much appreciated.

lib/modules/manager/deno/artifacts.ts
Hajime-san: FYI: grep with `if (!oldLockFileContent) {` found that some managers returns `null` here.

lib/modules/manager/deno/post.ts
Hajime-san: I reverted this at d0c3cf2 due to seems to be wrong.
This `if` statement should be an early return.

lib/modules/manager/deno/artifacts.ts
Hajime-san: @bartlomieju Does the `deno install` command have any option to update only a lock file?
If it exists, there is no need to delete the file during lock file maintenance.

bartlomieju: @Hajime-san not really - it will always update lock file and `deno.json`/`package.json`

lib/modules/manager/deno/artifacts.ts
```diff
@@ -0,0 +1,127 @@
...
+      toolConstraints: [
+        {
+          toolName: 'deno',
+          constraint: config.constraints?.deno,
+        },
+      ],
+    };
+
+    // "deno install" don't execute lifecycle scripts of package.json by default
+    // https://docs.deno.com/runtime/reference/cli/install/#native-node.js-addons
```
bartlomieju, line 92: FYI - the behavior here will be the same as in `pnpm` - any package that has an install script needs to be explicitly approved using a dedicated subcommand (currently WIP at https://github.com/denoland/deno/pull/30966, shipping in the next couple weeks)

lib/modules/manager/deno/artifacts.ts
```diff
@@ -0,0 +1,127 @@
...
+    // NOTE: Specifying individual packages e.g. "deno install <datasource>:<package>@<version>"
+    // could be better to reduce registry queries, but does not support the http import module
```
bartlomieju: I'd say that handling http packages is out of scope here and the focus should be only on npm/JSR packages

lib/modules/manager/deno/compat.ts
```diff
@@ -0,0 +1,128 @@
...
+  managerData,
+  packageFile,
+}: Partial<PackageFile<DenoManagerData>>): Promise<{
+  workspaces?: string[];
+  packagePaths: string[];
+} | null> {
+  if (!packageFile) {
+    return null;
+  }
+  logger.debug(`Detecting deno's node compat Workspaces`);
```
bartlomieju, line 53: @Hajime-san could you explain a bit more here - does this work with Deno workspaces too?

lib/modules/manager/deno/extract.ts
```diff
@@ -0,0 +1,129 @@
...
+    if (upath.basename(matchedFile) === 'deno.lock') {
+      // node-compat
+      const extracted = await collectPackageJson(matchedFile);
+      if (extracted) {
+        packageFiles.push(...extracted);
+      }
+    }
+
+    // deno.json or deno.jsonc
+    if (upath.basename(matchedFile).startsWith('deno.json')) {
```
bartlomieju, line 35: @Hajime-san I don't understand this bit - why are you going into "node-compat" if there's `deno.lock` file? As it stands if there's `deno.lock` the current project can have `deno.json(c)` and/or `package.json`

lib/modules/manager/deno/index.ts
```diff
@@ -0,0 +1,41 @@
...
+export const supportsLockFileMaintenance = true;
+export const supportedDatasources = [
+  NpmDatasource.id,
+  JsrDatasource.id,
+  DenoDatasource.id,
+];
+
+export const defaultConfig = {
+  managerFilePatterns: [
+    '/(^|/)deno\\.lock$/', // for node-compat
```
bartlomieju: Ditto - I could use some pointers here

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,290 @@
...
+  // jsr and npm datasource
+  if (datasource === 'jsr' || datasource === 'npm') {
+    if (
+      !lockFileContent.lockedVersions ||
+      is.emptyObject(lockFileContent.lockedVersions)
+    ) {
+      return null;
+    }
+
+    // find "jsr:@scope/name@1.2.3" from "jsr:@scope/name@1.2.3"
```
bartlomieju, line 96: I believe this comment is not correct - what did you mean here?

bartlomieju:
Asked @dsherret for another review since he knows the most about Deno lockfile and dependency handling

lib/modules/manager/deno/artifacts.ts
RahulGautamSingh: Agreed. @Hajime-san We can keep basic/necessary support in this PR and extra/enhancements/complex ones for follow up PRs. This will make it easier to understand and safer to merge.

lib/modules/manager/deno/extract.ts
Hajime-san: This branch is the path when `deno.lock` and `package.json` exist.
If `package.json` is found, it may potentially be an npm workspace, so the path could solve `package.json` next to `deno.lock` as the workspace root.

Could you check the `npm workspace compatible` test case below?
https://github.com/renovatebot/renovate/pull/37828/files#diff-e42be453c207146a57f3f7820d3f2d2b8313e446a3a714e1a7992254d517cf0cR321-R421

lib/modules/manager/deno/compat.ts
Hajime-san: No. This `detectNodeCompatWorkspaces` is used only to find `package.json`.
The Deno workspace is solved by iterating the argument of `matchedFiles`(could be deno.json/deno.jsonc) of `extractAllPackageFiles`.

lib/modules/manager/deno/index.ts
Hajime-san: Same opinion above https://github.com/renovatebot/renovate/pull/37828#discussion_r2438052852

This is required for projects that consist only of `deno.lock` and `package.json`.
The existence of `deno.lock` suggests that the project has this configuration.
If only `package.json` is set, it may be a Node.js (npm) manager other than Deno.

lib/modules/manager/deno/post.ts
Hajime-san: This is a process that checks the lock file and extracts the version just in case the package version is fixed in the `packageFile` (deno.json/package.json).

Hajime-san:
@bartlomieju @dsherret
FYI: The key to the Deno manager implementation is that it does not completely cover the behavior of the deno cli, but rather provides a level of granularity that allows the Renovate team to maintain the code.
For example, deno cli`s package.json` compatibility does not support [these specifications](https://github.com/denoland/deno_npm/blob/0d2b66e68bfe92a050558febab3f59f4e5cc9c23/src/registry.rs#L391-L399), but the Renovate Deno manager implementation does not specifically handle this.
This means that these detailed behaviors are left to the user to resolve.

Hajime-san:
@viceice @RahulGautamSingh 
CI suddenly fails but I don't know what's going on.
The `pnpm eslint-ci` command works fine my local repository.

```bash
Run pnpm eslint-ci

> renovate@0.0.0-semantic-release eslint-ci /home/runner/work/renovate/renovate
> eslint . --cache --cache-strategy content --cache-location .cache/eslint --format gha

ESLint Annotations
  Error: Prefer toHaveBeenCalledWith (/* expected args */) (vitest/prefer-called-exactly-once-with)

 ELIFECYCLE  Command failed with exit code 1.
Error: Process completed with exit code 1.
```

felipecrs:
Double check if you have some toHaveBeenCalledWith and replace with toHaveBeenCalledExactlyOnceWith.

Hajime-san:
Thank you! @felipecrs

lib/modules/manager/deno/artifacts.ts
```diff
@@ -0,0 +1,117 @@
...
+    updateArtifact;
+  logger.debug(`deno.updateArtifacts(${packageFileName})`);
+  const isLockFileMaintenance = config.updateType === 'lockFileMaintenance';
+
+  if (is.emptyArray(updatedDeps) && !isLockFileMaintenance) {
+    logger.debug('No updated deno deps - returning null');
+    return null;
+  }
+
+  const lockFileName = updatedDeps[0].lockFiles?.[0];
```
RahulGautamSingh: Incase of `lockFileMaintenance` the `updatesDeps` array can be empty as per the if-check before, so we should have a fallback here.

lib/modules/manager/deno/artifacts.ts
```diff
@@ -0,0 +1,117 @@
...
+    await exec('deno install', execOptions);
+
+    const newLockFileContent = await readLocalFile(lockFileName);
+    if (
+      !newLockFileContent ||
+      Buffer.compare(oldLockFileContent, newLockFileContent) === 0
+    ) {
+      return null;
+    }
+    return [
```
RahulGautamSingh, line 103:
```suggestion
    
    return [
```

lib/modules/manager/deno/compat.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+vi.mock('../../../util/fs');
+// used in detectNodeCompatWorkspaces()
+vi.mock('find-packages', () => ({
+  findPackages: vi.fn(),
+}));
+
+describe('modules/manager/deno/compat', () => {
+  describe('extractDenoCompatiblePackageJson()', () => {
+    it('invalid package.json', async () => {
```
RahulGautamSingh:
```suggestion
    it('returns null if invalid package.json', async () => {
```

lib/modules/manager/deno/compat.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+          _from: 1,
+        }),
+      );
+      const result = await extractDenoCompatiblePackageJson('package.json');
+      expect(result).toBeNull();
+    });
+  });
+
+  describe('detectNodeCompatWorkspaces()', () => {
+    it('handle no packageFile', async () => {
```
RahulGautamSingh:
```suggestion
    it('returns null if no packageFile', async () => {
```

lib/modules/manager/deno/compat.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+        packageFile: undefined,
+      });
+      expect(result).toBeNull();
+    });
+  });
+
+  describe('collectPackageJson()', () => {
+    it('node-compat package.json', async () => {
+      GlobalConfig.set({ localDir: '' });
+      const { findPackages } = await import('find-packages');
```
RahulGautamSingh: Does this disturb the tests? @viceice

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,290 @@
...
+export function getLockedVersion(
+  deps: PackageDependency<DenoManagerData>,
+  lockFileContent: LockFile,
+): string | undefined | null {
+  if (is.emptyObject(lockFileContent)) {
+    return null;
+  }
+  const { datasource, currentRawValue, currentValue, depName } = deps;
+
+  // deno datasource
```
RahulGautamSingh:
```suggestion
```

RahulGautamSingh (changes requested):
I have reviewed all files now and after these changes, it's a go from my side.

Hajime-san:
Sorry for splitting the suggestion commit. The GitHub's of `Add suggestion to batch` is hell experience.

lib/modules/manager/deno/artifacts.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+import type { DirectoryResult } from 'tmp-promise';
+import tmp from 'tmp-promise';
+import { GlobalConfig } from '../../../config/global';
+import { TEMPORARY_ERROR } from '../../../constants/error-messages';
+import { exec as _exec } from '../../../util/exec';
+import { ExecError } from '../../../util/exec/exec-error';
+import type { UpdateArtifact } from '../types';
+import { updateArtifacts } from './artifacts';
+
+vi.mock('../../../util/exec');
```
viceice: we've specifically exec mocking feature which should be used instead, so you can assert exact executed commands 

https://github.com/renovatebot/renovate/blob/ed5ad43e2bd6c14bdef497491c58972f6794c5e2/lib/modules/manager/mix/artifacts.spec.ts#L83

lib/modules/manager/deno/artifacts.ts
viceice: isn't the lockfile name also on `updateArtifact`?

lib/modules/manager/deno/artifacts.ts
```diff
@@ -0,0 +1,118 @@
...
+        {
+          toolName: 'deno',
+          constraint: config.constraints?.deno,
+        },
+      ],
+    };
+
+    // "deno install" don't execute lifecycle scripts of package.json by default
+    // https://docs.deno.com/runtime/reference/cli/install/#native-node.js-addons
+    await exec('deno install', execOptions);
```
viceice, line 93: we usually delete the lockfile before this on lockfile maintenance

lib/modules/manager/deno/compat.spec.ts
viceice: why dynamic import? it's mocked statically

lib/modules/manager/deno/compat.ts
```diff
@@ -0,0 +1,128 @@
...
+    logger.error({ err, packageFile }, 'Error parsing package.json');
+    return null;
+  }
+
+  const extracted = extractPackageJson(packageJson, packageFile);
+  if (!extracted) {
+    return null;
+  }
+
+  const res: PackageFile<DenoManagerData> = {
```
viceice: why cloning the object instead of just assigning the package file?

lib/modules/manager/deno/compat.ts
```diff
@@ -0,0 +1,128 @@
...
+    const result = await detectNodeCompatWorkspaces(pkg);
+    /* v8 ignore next 3: hard to test */
+    if (!result) {
+      return null;
+    }
+    const { workspaces, packagePaths } = result;
+    pkg.managerData = {
+      ...pkg.managerData,
+      // override workspace
+      workspaces,
```
viceice, line 100: just assign instead of cloning (perf)

lib/modules/manager/deno/artifacts.ts
Hajime-san: https://github.com/renovatebot/renovate/pull/37828#discussion_r2465092684

lib/modules/manager/deno/compat.spec.ts
RahulGautamSingh: Yeah that's what I was thinking too @Hajime-san please fix this

ihorkatkov:
Is there anything left to push it forward?

RahulGautamSingh (approved):

felipecrs:
The new `deno install --lockfile-only` sounds very useful for Renovate!

https://deno.com/blog/v2.6#improved-lockfile-and-install-workflows

Hajime-san:
@felipecrs 
Yeah, pnpm integration use the flag exactly.

https://github.com/renovatebot/renovate/blob/2ee1c2bf4295f6d7798c527bb19592b29a1ca92e/lib/modules/manager/npm/post-update/pnpm.ts#L94-L127

@viceice 
Should I add the flag to the command, in this PR? Or would it be better to cool down until the feature is mature?

The binary of containerbase was up to date.

- https://github.com/containerbase/base/pull/5586

hckhanh:
Is there any new progress on this amazing PR?

Hajime-san:
There was a conflict with the main branch, which was resolved. In addition, there was a change in the coverage measurement method, which caused the coverage to drop, so some code was modified and test code was added.

Probably related:

- https://github.com/renovatebot/renovate/issues/40625

viceice (commented):
needs a rebase to drop versioning and datasource

Hajime-san:
@viceice
> needs a rebase to drop versioning and datasource

I think the datasource changes are not included in this PR.

Should I make a separate PR for versioning and rebase the implementation to this branch later?

lib/modules/manager/deno/artifacts.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+import type { DirectoryResult } from 'tmp-promise';
+import tmp from 'tmp-promise';
+import { mockExecAll } from '../../../../test/exec-util.ts';
+import { GlobalConfig } from '../../../config/global.ts';
+import { TEMPORARY_ERROR } from '../../../constants/error-messages.ts';
+import { ExecError } from '../../../util/exec/exec-error.ts';
+import type { UpdateArtifact } from '../types.ts';
+import { updateArtifacts } from './artifacts.ts';
+
+vi.mock('fs-extra');
```
viceice:
```suggestion
vi.mock('../../../util/fs/index.ts');
```

mock our fs utils instead of `fs-extra`

lib/modules/manager/deno/compat.ts
```diff
@@ -0,0 +1,120 @@
+import is from '@sindresorhus/is';
+import { findPackages } from 'find-packages';
+import upath from 'upath';
+import { GlobalConfig } from '../../../config/global.ts';
+import { logger } from '../../../logger/index.ts';
+import { parseJson } from '../../../util/common.ts';
+import { getSiblingFileName, readLocalFile } from '../../../util/fs/index.ts';
```
viceice, line 7:
```suggestion
import { ensureLocalPath, getSiblingFileName, readLocalFile } from '../../../util/fs/index.ts';
```

lib/modules/manager/deno/compat.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  // npm workspace
+  if (is.nonEmptyArray(managerData?.workspaces)) {
+    filters = managerData?.workspaces;
+  }
+
+  // SAFETY: localDir should always be defined
+  const localDir = GlobalConfig.get('localDir')!;
+  const packages = await findPackages(
+    upath.dirname(upath.join(localDir, packageFile)),
```
viceice:
```suggestion
    upath.dirname(ensureLocalPath(packageFile)),
```

`ensureLocalPath` simplifies it and throws on errors

lib/modules/manager/deno/extract.ts
```diff
@@ -0,0 +1,133 @@
...
+      lockFiles,
+    );
+  }
+
+  const packageFile: PackageFile<DenoManagerData> = {
+    deps: content.dependencies,
+    packageFile: fileName,
+    managerData: content.managerData,
+    lockFiles,
+  } satisfies PackageFile<DenoManagerData>;
```
viceice:
```suggestion
  };
```

no need for duplicate type assignment

lib/modules/manager/deno/extract.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  return [packageFile, importMapPackageFile].filter(
+    Boolean,
```
viceice:
```suggestion
    isObject,
```

import from `@sindresorhus/is`

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  return parsedLockfile.data;
+}
+
+export function getLockedVersion(
+  deps: PackageDependency<DenoManagerData>,
+  lockFileContent: LockFile,
+): string | undefined | null {
+  if (is.emptyObject(lockFileContent)) {
```
viceice:
```suggestion
  if (isEmptyObject(lockFileContent)) {
```

use named imports

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    // "redirects": {
+    //   "https://deno.land/std": "https://deno.land/std@0.224.0"
+    // },
+    const key =
+      currentValue && depName ? `${depName}@${currentValue}` : depName;
+    if (
+      lockFileContent.redirectVersions &&
+      is.nonEmptyObject(lockFileContent.redirectVersions) &&
```
viceice:
```suggestion
      isNonEmptyObject(lockFileContent.redirectVersions) &&
```

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,291 @@
...
+      );
+      return match?.groups?.currentValue;
+    }
+  }
+
+  // jsr and npm datasource
+  if (datasource === 'jsr' || datasource === 'npm') {
+    if (
+      !lockFileContent.lockedVersions ||
+      is.emptyObject(lockFileContent.lockedVersions)
```
viceice:
```suggestion
      isEmptyObject(lockFileContent.lockedVersions)
```

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,291 @@
...
+  return null;
+}
+
+export async function collectPackageJsonAsWorkspaceMember(
+  packageFiles: PackageFile<DenoManagerData>[],
+): Promise<void> {
+  // detect package.json as members of a deno workspace
+  const workspaceRoots = packageFiles.filter(
+    (pkg) =>
+      is.nonEmptyArray(pkg.managerData?.workspaces) &&
```
viceice:
```suggestion
      isNonEmptyArray(pkg.managerData?.workspaces) &&
```

Please fix all other occurences

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  for (const pkg of packageFiles) {
+    const workspaces = pkg.managerData?.workspaces;
+    if (is.nonEmptyArray(workspaces)) {
+      const rootDir = upath.dirname(pkg.packageFile);
+      const matchers = workspaces.map(
+        (pattern) =>
+          // allow ./sub/* to match sub
+          new Minimatch(upath.normalize(pattern), {
```
viceice:
```suggestion
          minimatch(upath.normalize(pattern), {
```

use our minimatch utils
https://github.com/renovatebot/renovate/blob/321d121e680e5e722e3139af8c99a0791ba302c5/lib/util/minimatch.ts#L6

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,291 @@
...
+  for (const packageFile of invalidPackageFiles) {
+    const pkg = packageMap.get(packageFile);
+    // packageFile always exists in packageMap since invalidPackageFiles comes from workspaceContexts
+    // which is built from packageFiles. This check is defensive programming.
+    /* v8 ignore next 3: hard to test - packageFile always exists in packageMap */
+    if (!pkg) {
+      continue;
+    }
+    // remove invalid workspace
+    delete pkg.managerData?.workspaces;
```
viceice:
```suggestion
    delete pkg?.managerData?.workspaces;
```

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,291 @@
...
+    }
+  }
+  for (const packageFile of invalidPackageFiles) {
+    const pkg = packageMap.get(packageFile);
+    // packageFile always exists in packageMap since invalidPackageFiles comes from workspaceContexts
+    // which is built from packageFiles. This check is defensive programming.
+    /* v8 ignore next 3: hard to test - packageFile always exists in packageMap */
+    if (!pkg) {
+      continue;
+    }
```
viceice:
```suggestion
```

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,291 @@
...
+}
+
+async function applyLockedVersion(
+  packageFiles: PackageFile<DenoManagerData>[],
+): Promise<void> {
+  // apply locked versions from lock files
+  // use cache to avoid reading the same lock file multiple times
+  const lockFileCache = new Map<string, LockFile>();
+  for (const pkg of packageFiles) {
+    if (is.nonEmptyArray(pkg.lockFiles)) {
```
viceice: invert condition to reduce nesting

lib/modules/manager/deno/schema.ts
```diff
@@ -0,0 +1,351 @@
...
+
+  const remoteVersions = new Set<string>();
+  for (const key of Object.keys(remote)) {
+    remoteVersions.add(key);
+  }
+  return {
+    lockedVersions,
+    redirectVersions,
+    remoteVersions,
+    lockfileVersion: Number(version),
```
viceice:
```suggestion
    lockfileVersion: parseInt(version, 10),
```

be explicit

lib/modules/manager/deno/update.ts
```diff
@@ -0,0 +1,357 @@
...
+        searchString,
+        newString,
+      );
+      // Compare the parsed JSON/JSONC structure of old and new
+      if (dequal(parsedContents, parseJsonc(testContent))) {
+        return testContent;
+      }
+    }
+  }
+  /* v8 ignore next -- needs test */
```
viceice: please add a test

lib/modules/manager/deno/update.ts
```diff
@@ -0,0 +1,357 @@
...
+      const matches = Object.entries(parsedContents.tasks).filter(
+        ([, value]) => {
+          if (typeof value === 'string') {
+            return value.includes(searchCurrentValue);
+          }
+          return false;
+        },
+      );
+      for (const [key] of matches) {
+        // prettier-ignore
```
viceice: why?

lib/modules/manager/deno/update.ts
```diff
@@ -0,0 +1,357 @@
...
+        parsedContents.tasks[key] = (parsedContents.tasks[key] as string).replace(searchCurrentValue, newString);
```
viceice: why do we need the `as string`

lib/modules/manager/deno/update.ts
```diff
@@ -0,0 +1,357 @@
...
+      const matches = Object.entries(parsedContents.tasks).filter(
+        ([, value]) => {
+          if (value && typeof value === 'object' && 'command' in value) {
+            return value.command?.includes(searchCurrentValue);
+          }
+          return false;
+        },
+      );
+      for (const [key] of matches) {
+        // prettier-ignore
```
viceice: same here. the code looks pretty ugly now

lib/modules/manager/deno/compat.ts
```diff
@@ -0,0 +1,120 @@
...
+export async function detectNodeCompatWorkspaces({
+  managerData,
+  packageFile,
+}: Partial<PackageFile<DenoManagerData>>): Promise<{
+  workspaces?: string[];
+  packagePaths: string[];
+} | null> {
+  if (!packageFile) {
+    return null;
+  }
```
viceice:
```suggestion
```

lib/modules/manager/deno/compat.ts
```diff
@@ -0,0 +1,120 @@
...
+  };
+  result.packageFile = packageFile;
+  return result;
+}
+
+// referring to lib/modules/manager/npm/extract/pnpm.ts detectPnpmWorkspaces()
+export async function detectNodeCompatWorkspaces({
+  managerData,
+  packageFile,
+}: Partial<PackageFile<DenoManagerData>>): Promise<{
```
viceice:
```suggestion
}: Pick<PackageFile<DenoManagerData>, 'managerData' | 'packageFile'>): Promise<{
```

lib/modules/manager/deno/compat.ts
```diff
@@ -0,0 +1,120 @@
...
+    await extractDenoCompatiblePackageJson(rootPackageJson);
+  if (rootPackageFile) {
+    rootPackageFile.lockFiles = lockFiles;
+
+    // detect node compat workspaces
+    const result = await detectNodeCompatWorkspaces(rootPackageFile);
+    /* v8 ignore next 3: hard to test */
+    if (!result) {
+      return null;
+    }
```
viceice:
```suggestion
```

lib/modules/manager/deno/schema.ts
```diff
@@ -0,0 +1,352 @@
...
+  const remoteVersions = new Set<string>();
+  for (const key of Object.keys(remote)) {
+    remoteVersions.add(key);
+  }
+  return {
+    lockedVersions,
+    redirectVersions,
+    remoteVersions,
+    // oxlint-disable-next-line Redundant radix parameter
+    lockfileVersion: parseInt(version, 10),
```
Hajime-san: seems false positive by oxlint

lib/modules/manager/deno/post.spec.ts
```diff
@@ -0,0 +1,706 @@
...
+          },
+        },
+      );
+      expect(result).toBeNull();
+    });
+  });
+
+  describe('collectPackageJsonAsWorkspaceMember()', () => {
+    it('should collect package.json files as deno workspace members', async () => {
+      GlobalConfig.set({ localDir: '' });
```
viceice, line 208: should be moved to a `beforeAll` otherwise it can cause bad side effects when other users add more tests later.

lib/modules/manager/deno/schema.ts
viceice: @zharinov Do you know more?

viceice (commented):
otherwise LGTM

viceice (dismissed):

Hajime-san:
@zharinov 
Sorry for pinning, could you take a look again? I'm happy to move forward this PR.

lib/modules/manager/deno/schema.ts
```diff
@@ -0,0 +1,352 @@
...
+
+  const remoteVersions = new Set<string>();
+  for (const key of Object.keys(remote)) {
+    remoteVersions.add(key);
+  }
+  return {
+    lockedVersions,
+    redirectVersions,
+    remoteVersions,
+    // oxlint-disable-next-line Redundant radix parameter
```
viceice: I think this can now be removed 
```suggestion
```

lib/modules/manager/deno/update.ts
```diff
@@ -0,0 +1,346 @@
...
+  if (datasource === 'jsr' || datasource === 'npm') {
+    return currentValue
+      ? `${datasource}:${depName}@${currentValue}`
+      : `${datasource}:${depName}`;
+  }
+  return null;
+}
+
+// ported from lib/modules/manager/npm/update/dependency/index.ts
+function replaceAsString(
```
viceice, line 29: this can probably moved to `lib/utils` to be usable by others who change json files. 🤔

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,286 @@
...
+    }
+  }
+}
+
+async function applyLockedVersion(
+  packageFiles: PackageFile<DenoManagerData>[],
+): Promise<void> {
+  // apply locked versions from lock files
+  // use cache to avoid reading the same lock file multiple times
+  const lockFileCache = new Map<string, LockFile>();
```
viceice:
```suggestion
  const lockFileCache: Record<string, LockFile> = {};
```

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,286 @@
...
+    }
+    const lockFile = pkg.lockFiles[0];
+    let lockFileContent: LockFile;
+
+    if (lockFileCache.has(lockFile)) {
+      lockFileContent = lockFileCache.get(lockFile)!;
+    } else {
+      lockFileContent = await getDenoLock(lockFile);
+      lockFileCache.set(lockFile, lockFileContent);
+    }
```
viceice:
```suggestion
    let lockFileContent = lockFileCache[lockFile];

    if (!lockFileContent) {
      lockFileContent = await getDenoLock(lockFile);
      lockFileCache[lockFile] = lockFileContent;
    }
```

lib/modules/manager/deno/post.ts
```diff
@@ -0,0 +1,286 @@
...
+}
+
+export function normalizeWorkspace(
+  packageFiles: PackageFile<DenoManagerData>[],
+): void {
+  // determine workspace root to collect lock files of workspace members
+  const workspaceContexts: WorkspaceContext[] = [];
+
+  // create reference map for packageFile object
+  const packageMap = new Map<string, PackageFile<DenoManagerData>>();
```
viceice: use `Record<string, PackageFile<DenoManagerData>`  for simpler null checks, see other suggestions

lib/modules/manager/deno/index.ts
```diff
@@ -0,0 +1,40 @@
...
+// See: normalizeWorkspace in lib/modules/manager/deno/post.ts
+export { updateDependency } from './update.ts';
+
+export const url =
+  'https://docs.deno.com/runtime/getting_started/installation/';
+export const categories: Category[] = ['js'];
+
+export const supersedesManagers = ['npm'];
+export const supportsLockFileMaintenance = true;
+// NOTE: Deno could support another lock filename that specified in deno.json
```
viceice: should be added to readme

lib/modules/manager/deno/update.spec.ts
```diff
@@ -0,0 +1,1016 @@
...
+          {
+            imports: {
+              fs: 'https://deno.land/std@0.223.0/fs/mod.ts',
+            },
+          },
+          null,
+          2,
+        );
+
+        const upgrade: UpdateDependencyConfig['upgrade'] = {
```
viceice:
```suggestion
        const upgrade: Upgrade = {
```

also fix others here

viceice (commented):
conflicted

lib/modules/manager/deno/update.ts
Hajime-san: The `npm` implementation is specific to processing package.json. Is it worth to unify them since this one processes deno.json(c)?

https://github.com/renovatebot/renovate/blob/5ab19552c165d11b6481861847b51dd25b6017e4/lib/modules/manager/npm/update/dependency/index.ts#L35-L118

viceice (approved):

PR #37861, opened 2025-09-05
fix(presets): use semantic commit type `chore` for lockfile updates with `:semanticPrefixFixDepsChoreOthers` preset

### LINKED ISSUE #6791: `chore` commit type when only the lockfile is updated
### Which Renovate are you using?

WhiteSource Renovate App

### Which platform are you using?

GitHub.com

### Have you checked the logs? Don't forget to include them if relevant

No. I think what is happening is probably expected, but i'm hoping it can be configured to behave differently.

### What would you like to do?

For a CLI tool, I think it mostly makes sense to pin versions because because it is essentially an "app". However, since it is also published as a package, there are some dependencies that i want to avoid resulting in new versions. for context, i publish using `semantic-release`, which heavily relies on commit type to determine whether to publish as a major, minor, or patch version. i consider [snyk](https://www.npmjs.com/package/snyk) to be in this category since it isn't really a true dependency, but must be installed as a prod dependency to work fully.

I [configured one of my projects](https://github.com/travi/cli/blob/d000e3878a44b7a889d08037ddae49a6862b7de3/.renovaterc.json#L3-L6) to prefer updating the lockfile rather than pinning `snyk`, which seems to have accomplished that piece. However, the lockfile updates still use the `fix` commit type rather than `chore`. Since the lockfile is not included when a package is installed, I don't think it is valuable to publish a new version of the package.

Is it possible to configure lockfile-only updates to use the `chore` commit type, even when the update is for a prod dependency?

sisp:
## Changes

I've extended the `:semanticPrefixFixDepsChoreOthers` preset to use the semantic commit type `chore` for lockfile updates (`rangeStrategy: 'update-lockfile'`). The rationale for this setting is as follows: Runtime dependencies of libraries typically have loose constraints (e.g., lower bounds). Yet, it's useful to lock also runtime dependencies for offering a reproducible development environment and deterministic/reproducible testing. Renovate's `rangeStrategy: 'update-lockfile'` setting allows updating only the locked version of a direct dependency without changing the version constraint in the manifest – independent of overall lockfile maintenance. With appropriate package rules, this range strategy can be applied only to runtime dependencies while, e.g., pinned versions of dev dependencies are bumped both in the manifest and lockfile. Now, the semantic commit type `fix` should only be used for changes that affect the consumers of a library, but updating only the locked version of a direct runtime dependency is a dev-only change, hence the semantic commit type should be `chore`. To achieve this behavior, I've added a package rule to the `:semanticPrefixFixDepsChoreOthers` preset at the end of the rules array to set the semantic commit type `chore` whenever the `isLockfileUpdate` attribute is `true`. This rule must come last to override any preceding rule that sets the semantic commit type to `fix`. Alternatively, we could add `matchJsonata: ['isLockfileUpdate != true']` to all package rules in this preset that set the semantic commit type to `fix`, but I felt that's too repetitive. I'm open to different opinions though. :slightly_smiling_face:

## Context

- [x] This closes an existing Issue: #6791

## AI assistance disclosure

The public repository: https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers

See the `main` branch for the current behavior of the `:semanticPrefixFixDepsChoreOthers` preset, the `fix` branch contains the package rule I've added in this PR, which is effectively equivalent to adding it to the preset. As you can see in this test project, the Renovate PR that targets the `main` branch uses the semantic commit type `fix` whereas the Renovate PR that targets the `fix` branch uses the semantic commit type `chore`.

### DIFF
--- a/lib/config/presets/__snapshots__/index.spec.ts.snap
+++ b/lib/config/presets/__snapshots__/index.spec.ts.snap
@@ -103,6 +103,12 @@ exports[`config/presets/index > resolvePreset > migrates automerge in presets 1`
       "semanticCommitType": "fix",
     },
     {
+      "matchJsonata": [
+        "isLockfileUpdate = true",
+      ],
+      "semanticCommitType": "chore",
+    },
+    {
       "matchPackageNames": [
         "*",
       ],
--- a/lib/config/presets/internal/default.ts
+++ b/lib/config/presets/internal/default.ts
@@ -583,6 +583,10 @@ export const presets: Record<string, Preset> = {
         matchManagers: ['poetry'],
         semanticCommitType: 'fix',
       },
+      {
+        matchJsonata: ['isLockfileUpdate = true'],
+        semanticCommitType: 'chore',
+      },
     ],
   },
   separateMajorReleases: {


viceice (commented):
I'm not sure about this, requesting more reviews

secustor (commented):
I agree with the notion, that it should be `chore`, but think this change would apply for ANY lock file update.

Have you tested this against a repo?

sisp:
@secustor What do you mean by ANY lockfile update? Which additional scenarios would you like to check? I can extend the test repo accordingly.

secustor:
I'm referring to updates where versions in package files have been updated and then lockfiles too.

sisp:
@secustor Ah, got it, thanks for clarifying. I've updated the [test repo](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers) to test both scenarios: (a) lockfile-only update and (b) manifest / package file update + lockfile update. The additional package rule that I've added to the preset in this PR only applies to lockfile-only updates, i.e. an update of the runtime dependencies in the manifest / package file and the lockfile will continue using the semantic commit type `fix`. Please check the [test repo's PRs](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/pulls) to confirm this behavior.

secustor (approved):

RahulGautamSingh:
> Please check the [test repo's PRs](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/pulls) to confirm this behavior.

PR 1 and 2 in the linked test repo..both change the lockfile but have different commit type. Why?

RahulGautamSingh (commented):
> Please check the [test repo's PRs](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/pulls) to confirm this behavior.

PR 1 and 2 in the linked test repo..both change the lockfile but have different commit type. Why?

sisp:
Let me try to clarify:

| Scenario | Test PR (current behavior) | Test PR (new behavior of this PR) |
| - | - | - |
| Lockfile-only update | [#1](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/pull/1) – `fix` ❌️ | [#2](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/pull/2) – `chore` ✅️ |
| Package file + lockfile update | [#4](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/pull/4) – `fix` ✅️ | [#5](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/pull/5) – `fix` ✅️ |

So, [#1](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/pull/1) demonstrates the current behavior – lockfile-only update with commit type `fix`, which I consider incorrect – and [#2](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/pull/2) demonstrates the behavior after the fix of this PR – lockfile-only update with commit type `chore`, which I consider correct.

In the test repo, I configure Renovate to consider the [default branch and the `fix` branch as base branches](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/blob/8b699f93d47c53c31f69481ae81c3710cae162c5/renovate.json#L4) and to [merge configs](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/blob/8b699f93d47c53c31f69481ae81c3710cae162c5/renovate.json#L5), so it submits PRs for both. In the `fix` branch, I add the new [package rule](https://github.com/sisp/test-renovate-lockfile-update-semanticprefixfixdepschoreothers/blob/ee124257667a6d69a61ca5fe5e681ed5098db980/renovate.json#L17-L20), so all PRs that target the `fix` branch are created with this additional rule in place. This way, we can compare behavior without the new rule and with the new rule in a single test repo.

RahulGautamSingh (approved):

PR #37864, opened 2025-09-05
fix(nvm): handle code comments

breadadams:
## Changes

Update the `nvm` manager to filter out code comments before extracting the Node version.

## Context

- [x] This closes an existing Issue: https://github.com/renovatebot/renovate/discussions/37850

## AI assistance disclosure

The public repository: https://github.com/breadadams/renovate-nvmrc-comment-bug-testing

### DIFF
--- a/lib/modules/manager/nvm/extract.spec.ts
+++ b/lib/modules/manager/nvm/extract.spec.ts
@@ -37,7 +37,7 @@ describe('modules/manager/nvm/extract', () => {
 
     it('supports code comments', () => {
       const res = extractPackageFile(
-        '# This is a comment\nv20.19.3\n# This is another comment',
+        '# This is a comment\nv20.19.3 # This is an inline comment\n# This is another comment',
       );
       expect(res.deps).toEqual([
         {
--- a/lib/modules/manager/nvm/extract.spec.ts
+++ b/lib/modules/manager/nvm/extract.spec.ts
@@ -25,10 +25,23 @@ describe('modules/manager/nvm/extract', () => {
     });
 
     it('skips non ranges', () => {
-      const res = extractPackageFile('latestn');
+      const res = extractPackageFile('latest\n');
       expect(res.deps).toEqual([
         {
-          currentValue: 'latestn',
+          currentValue: 'latest',
+          datasource: 'node-version',
+          depName: 'node',
+        },
+      ]);
+    });
+
+    it('supports code comments', () => {
+      const res = extractPackageFile(
+        '# This is a comment\nv20.19.3\n# This is another comment',
+      );
+      expect(res.deps).toEqual([
+        {
+          currentValue: 'v20.19.3',
           datasource: 'node-version',
           depName: 'node',
         },
--- a/lib/modules/manager/nvm/extract.ts
+++ b/lib/modules/manager/nvm/extract.ts
@@ -1,3 +1,4 @@
+import is from '@sindresorhus/is';
 import { NodeVersionDatasource } from '../../datasource/node-version';
 import type { PackageDependency, PackageFileContent } from '../types';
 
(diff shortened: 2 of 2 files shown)


lib/modules/manager/nvm/extract.spec.ts
```diff
@@ -25,10 +25,23 @@ describe('modules/manager/nvm/extract', () => {
...
+      const res = extractPackageFile('latest\n');
       expect(res.deps).toEqual([
         {
-          currentValue: 'latestn',
+          currentValue: 'latest',
+          datasource: 'node-version',
+          depName: 'node',
+        },
+      ]);
+    });
```
breadadams, line 36: ℹ️ Git diff being a little funky here, I've just fixed the `latestn` typo.

lib/modules/manager/nvm/extract.ts
```diff
@@ -4,7 +4,12 @@ import type { PackageDependency, PackageFileContent } from '../types';
 export function extractPackageFile(content: string): PackageFileContent {
   const dep: PackageDependency = {
     depName: 'node',
-    currentValue: content.trim(),
+    currentValue: content
+      .split('\n')
+      // Exclude code comments
+      .filter((str) => !str.startsWith('#'))
+      .join('\n')
+      .trim(),
```
astellingwerf: Would it make sense to do all the transformations like they do in nvm:
https://github.com/nvm-sh/nvm/blob/69075e99b0e4b43d31c2fe5360089f308d9cb6b0/nvm.sh#L539

RahulGautamSingh: We could add logic to filter inline comments and trim the lines before `.join('\n')` is called @breadadams could you add these?

breadadams: Thanks for the feedback!

> Would it make sense to do all the transformations like they do in nvm:

Good point @astellingwerf, I wasn't considering the inline comments as I've never used them.

> We could add logic to filter inline comments and trim the lines before `.join('\n')` is called

@RahulGautamSingh I don't think that's necessary, I've gone with a regex approach which handles regular comments _and_ inline ones.

RahulGautamSingh (approved):

astellingwerf (approved):

lib/modules/manager/nvm/extract.ts
```diff
@@ -4,7 +4,14 @@ import type { PackageDependency, PackageFileContent } from '../types';
 export function extractPackageFile(content: string): PackageFileContent {
   const dep: PackageDependency = {
     depName: 'node',
-    currentValue: content.trim(),
+    currentValue: content
+      .split('\n')
+      // Remove code comments
+      .map((line) => line.replace(/#.*$/, '').trim())
+      // Filter empty lines
+      .filter(Boolean)
```
viceice:
```suggestion
      .filter(isNotEmptyAndNotWhitespace)
```

breadadams: Not sure I follow @viceice, that doesn't appear to be a function in the codebase – but that's what's being achieved here by using `.filter(Boolean)`.

breadadams: Ah, are you referring to the util from the `@sindresorhus/is` package?

breadadams: Applied in c4d82184099266220b2f94660f9f376a55429408

astellingwerf (approved):

lib/modules/manager/nvm/extract.ts
```diff
@@ -1,10 +1,17 @@
...
 
 export function extractPackageFile(content: string): PackageFileContent {
   const dep: PackageDependency = {
     depName: 'node',
-    currentValue: content.trim(),
+    currentValue: content
+      .split('\n')
+      // Remove code comments
+      .map((line) => line.replace(/#.*$/, '').trim())
+      .filter(is.nonEmptyStringAndNotWhitespace)
```
viceice:
```suggestion
      .filter(isNonEmptyStringAndNotWhitespace)
```

please use the named export

viceice (commented):
see comment, otherwise lgtm

lib/modules/manager/nvm/extract.ts
breadadams: Done, however there are 359 cases of `import is from '@sindresorhus/is';` across the codebase, that's why I went with that – although I can imagine the tree-shaking benefits of using the named export.

Would a PR switching them all over to named exports be welcome?

viceice (approved):

PR #37997, opened 2025-09-13
feat(datasource/rpm): support zstd-compressed data

amezin:
## Changes

Add support for zstd-compressed `primary.xml` in rpm datasource.

## Context

`primary.xml` in Fedora 42 repositories is compressed with zstd: https://download.fedoraproject.org/pub/fedora/linux/releases/42/Everything/x86_64/os/repodata/

Also, Fedora 41 "updates" repository is compressed with zstd too: https://download.fedoraproject.org/pub/fedora/linux/updates/41/Everything/x86_64/repodata/

So, to get package versions from Fedora, the datasource must support zstd decompression in addition to gzip.

Please select one of the below:

- [ ] This closes an existing Issue: #
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

The public repository: https://github.com/amezin/renovate-rpm-zstd

PR #38000, opened 2025-09-13
fix(datasource/rpm): use zlib in stream mode to handle large files

amezin:
## Changes

Use Gunzip stream, instead of decompressing to a buffer.

## Context

Zlib splits the data into reasonably sized chunks during decompression.

Before this change, for `registryUrl=http://download.opensuse.org/update/leap/15.6/sle/repodata/`, Renovate terminates with the error:

```
Error: Cannot create a string longer than 0x1fffffe8 characters
    at StringDecoder.write (node:string_decoder:104:10)
    at SAXStream.write (/home/amezin/renovate/node_modules/.pnpm/sax@1.4.1/node_modules/sax/lib/sax.js:243:28)
    at Readable.ondata (node:internal/streams/readable:1009:22)
    at Readable.emit (node:events:519:28)
    at Readable.emit (node:domain:489:12)
    at Readable.read (node:internal/streams/readable:782:10)
    at flow (node:internal/streams/readable:1283:53)
    at resume_ (node:internal/streams/readable:1262:3)
    at processTicksAndRejections (node:internal/process/task_queues:90:21)
```

After this change, the same repository is processed successfully.

## AI assistance disclosure

The public repository: https://github.com/amezin/renovate-rpm-stream

### DIFF
--- a/lib/modules/datasource/rpm/index.spec.ts
+++ b/lib/modules/datasource/rpm/index.spec.ts
@@ -1,4 +1,5 @@
 import { gzipSync } from 'node:zlib';
+import { setTimeout } from 'timers/promises';
 import { RpmDatasource } from '.';
 import * as httpMock from '~test/http-mock';
 import { logger } from '~test/util';
@@ -184,6 +185,7 @@ describe('modules/datasource/rpm/index', () => {
         ],
       });
 
+      await setTimeout();
       expect(logger.logger.debug).not.toHaveBeenCalledWith(
         expect.stringMatching(/^Empty response body/),
       );
@@ -208,6 +210,7 @@ describe('modules/datasource/rpm/index', () => {
         rpmDatasource.getReleasesByPackageName(primaryXmlUrl, packageName),
       ).rejects.toThrow(`Response code 404 (Not Found)`);
 
+      await setTimeout();
       expect(logger.logger.debug).not.toHaveBeenCalledWith(
         expect.stringMatching(/^Empty response body/),
       );
@@ -234,6 +237,7 @@ describe('modules/datasource/rpm/index', () => {
         'Empty response body from getting ' + primaryXmlUrl + '.',
       );
 
+      await setTimeout();
       expect(logger.logger.debug).toHaveBeenLastCalledWith(
         `Empty response body from getting ${primaryXmlUrl}.`,
       );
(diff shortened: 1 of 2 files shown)


viceice (approved):

lib/modules/datasource/rpm/index.ts
```diff
@@ -118,7 +115,7 @@ export class RpmDatasource extends Datasource {
     packageName: string,
   ): Promise<ReleaseResult | null> {
     let response: HttpResponse<Buffer>;
-    let decompressedBuffer: Buffer;
+    let decompressStream: Readable;
     try {
       // primaryGzipUrl is a .gz file, need to extract it before parsing
       response = await this.http.getBuffer(primaryGzipUrl);
```
viceice: we should use `got.stream` here, then you can pipe it directly without a memory buffer

viceice (changes requested):

lib/modules/datasource/rpm/index.ts
amezin: Likely yes, but:

1. There is a length check a few lines below: https://github.com/renovatebot/renovate/blob/096e7bcfd31bc9299269ce6ecb0c8df34ef5cbb1/lib/modules/datasource/rpm/index.ts#L126-L127 - how should it be implemented then? Checking `content-length` isn't exactly the same thing, as far as I understand.

2. It'll be harder to check the magic number in https://github.com/renovatebot/renovate/pull/37997

amezin: Done

amezin: After looking a bit closer at `deb` datasource, I've noticed that it saves both compressed and uncompressed files into cache dir: https://github.com/renovatebot/renovate/blob/71ad15d975ff7208f36497d84611ea739fcf3790/lib/modules/datasource/deb/index.ts#L66-L112

Should `rpm` do the same thing? If so, should the code be moved to utils?

RahulGautamSingh: Yes, yes

lib/modules/datasource/rpm/index.spec.ts
```diff
@@ -182,6 +184,20 @@ describe('modules/datasource/rpm/index', () => {
           },
         ],
       });
+
+      await setTimeout();
```
viceice, line 188: why this?

lib/modules/datasource/rpm/index.ts
```diff
@@ -187,22 +166,37 @@ export class RpmDatasource extends Datasource {
     });
 
     await new Promise<void>((resolve, reject) => {
-      let settled = false;
+      const emptyCheck = (): void => {
```
viceice, line 169: no inline functions

PR #38066, opened 2025-09-17
fix(jsonata): updates when version and digest changed (#36461)

jalil-salame:
## Changes

- added a regression test for the issue discussed in #36461 (taken from #37976)
- changed the logic in `auto-replace` to consider the number of changes to be made.
    If more than one change needs to be made and `replaceString` is not set then also set `replaceWithoutReplaceString`.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/workers/repository/update/branch/auto-replace.ts
+++ b/lib/workers/repository/update/branch/auto-replace.ts
@@ -196,6 +196,18 @@ async function checkExistingBranch(
   return existingContent;
 }
 
+// check if current !== newString and return 1 if it does or 0 if it doesn't
+function updatedToInt(
+  current: string | undefined,
+  newString: string | undefined,
+): number {
+  if (current && newString && newString !== current) {
+    return 1;
+  } else {
+    return 0;
+  }
+}
+
 export async function doAutoReplace(
   upgrade: BranchUpgradeConfig,
   existingContent: string,
@@ -225,11 +237,6 @@ export async function doAutoReplace(
     return await checkExistingBranch(upgrade, existingContent);
   }
 
-  // check if current !== newString and return 1 if it does or 0 if it doesn't
-  const updatedToInt = (
-    current: string | undefined,
-    newString: string | undefined,
-  ): number => (current && newString && newString !== current ? 1 : 0);
   // count how many strings need to be updated
   const changedCount =
     updatedToInt(depName, newName) +
--- a/lib/workers/repository/update/branch/auto-replace.ts
+++ b/lib/workers/repository/update/branch/auto-replace.ts
@@ -202,8 +202,8 @@ async function checkExistingBranch(
  * @remarks
  * Useful for counting the number of updates to do.
  *
- * @param {string|undefined} current The current value (if undefined then no update is required)
- * @param {string|undefined} newString The new value (if undefined then no update is required)
+ * @param current The current value (if undefined then no update is required)
+ * @param newString The new value (if undefined then no update is required)
  *
  * @returns 1 if `current !== newString` and 0 if they are equal or at least one is undefined.
  */
--- a/lib/workers/repository/update/branch/auto-replace.ts
+++ b/lib/workers/repository/update/branch/auto-replace.ts
@@ -224,11 +224,29 @@ export async function doAutoReplace(
   if (reuseExistingBranch) {
     return await checkExistingBranch(upgrade, existingContent);
   }
+
+  // check if current !== newString and return 1 if it does or 0 if it doesn't
+  const updatedToInt = (
+    current: string | undefined,
+    newString: string | undefined,
+  ): number => (current && newString && newString !== current ? 1 : 0);
+  // count how many strings need to be updated
+  const changedCount =
+    updatedToInt(depName, newName) +
+    updatedToInt(currentValue, newValue) +
+    updatedToInt(currentDigest, newDigest) +
+    updatedToInt(currentDigestShort, newDigest);
+  logger.debug(
+    { packageFile, depName, changedCount },
+    'N values changed and need to be updated',
+  );
+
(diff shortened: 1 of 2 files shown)


lib/workers/repository/update/branch/auto-replace.ts
```diff
@@ -224,11 +224,29 @@ export async function doAutoReplace(
   if (reuseExistingBranch) {
     return await checkExistingBranch(upgrade, existingContent);
   }
+
+  // check if current !== newString and return 1 if it does or 0 if it doesn't
+  const updatedToInt = (
```
viceice: use normal function declaration

jalil-salame: Fixed!

lib/workers/repository/update/branch/auto-replace.ts
```diff
@@ -196,6 +196,18 @@ async function checkExistingBranch(
   return existingContent;
 }
 
+// check if current !== newString and return 1 if it does or 0 if it doesn't
+function updatedToInt(
+  current: string | undefined,
+  newString: string | undefined,
+): number {
+  if (current && newString && newString !== current) {
```
viceice, line 214: this returns 0 if one is undefined and the other not. is this expected? 🤔

lib/workers/repository/update/branch/auto-replace.ts
```diff
@@ -196,6 +196,18 @@ async function checkExistingBranch(
   return existingContent;
 }
 
+// check if current !== newString and return 1 if it does or 0 if it doesn't
```
viceice: use tsdoc comment style

lib/workers/repository/update/branch/auto-replace.ts
jalil-salame: Yeah, since if either `current` or `new` is missing we cannot search for `current` and replace it with `new`.

E.g. unpinned docker image update: has a new version & digest, but doesn't have a current digest so we only update one value; the version.

lib/workers/repository/update/branch/auto-replace.ts
```diff
@@ -196,6 +196,28 @@ async function checkExistingBranch(
...
 }
 
+/**
+ * Check if an update from `current` to `newString` should be performed and return 1 if so.
+ *
+ * @remarks
+ * Useful for counting the number of updates to do.
+ *
+ * @param {string|undefined} current The current value (if undefined then no update is required)
+ * @param {string|undefined} newString The new value (if undefined then no update is required)
```
viceice:
```suggestion
 * @param current The current value (if undefined then no update is required)
 * @param newString The new value (if undefined then no update is required)
```
types not required for tsdoc only for jsdoc style 🤗

viceice (approved):

RahulGautamSingh:
We are basing it on the fact that when only `currentDigest` is getting updated `currentDigestShort` will also be defined, which makes the change count > 1, right? @jalil-salame

jalil-salame:
Did it cause a regression? I don't know if both `currentDigest` and `currentDigestShort` are defined

jalil-salame:
In my tests only `currentDigest` was defined

RahulGautamSingh:
> Did it cause a regression?

No such cases as of now. I am just trying to understand the thinking behind the change

jalil-salame:
Originally, if `replaceString` was `undefined` the code would set it to `currentValue` (version) or `currentDigest` (the first that was not `undefined`). Then it would try to replace the value in `replaceString` with the `newValue` and `newDigest`.

E.g `regex` manager:

```
# From this data
replaceString = "python:3.12@sha256:badbeef"
currentValue = "3.12"
currentDigest = "badbeef"
newValue = "3.13"
newDigest = "abcdef"
# It generates this string
replaceWith = "python:3.13@sha256:abcdef"
```

This doesn't work for the `jsonata` manager, since it doesn't provide a `replaceString`:

```
# From this data
replaceString = undefined (thus defaulted to currentValue)
currentValue = "3.12"
currentDigest = "badbeef"
newValue = "3.13"
newDigest = "abcdef"
# It generates this string
replaceWith = "3.13" # notice the digest is missing
```

It will then perform the update with `replaceString === currentValue` thus missing the `currentDigest` and failing to update.

The code changes the logic so it doesn't set the `replaceString` if there is more than one update:

```
# From this data
replaceString = undefined (thus defaulted to currentValue)
currentValue = "3.12"
currentDigest = "badbeef"
newValue = "3.13"
newDigest = "abcdef"
# Don't use replaceString since it was undefined and we have a newValue and newDigest
```

This reuses the codepath that was used to update the `depName`.

RahulGautamSingh:
Sounds good

RahulGautamSingh:
Regression: https://github.com/renovatebot/renovate/discussions/38252

RahulGautamSingh:
I don't think we should count both `currentDigest` and `currentDigestShort` as 2 changes...cause the short is derived from the digest itself.

RahulGautamSingh:
@jalil-salame We will revert the PR for now, you can submit another one which prevents the error mentioned in the link discussion.

jalil-salame:
@RahulGautamSingh  I can adjust the patch, but I would like to have a regression test for #38252 before making any changes

RahulGautamSingh:
> @RahulGautamSingh I can adjust the patch, but I would like to have a regression test for #38252 before making any changes

Done ✅
https://github.com/renovatebot/renovate/pull/38298

jalil-salame:
Thanks a bunch c:

jalil-salame:
Follow-up which fixes the regression: https://github.com/RahulGautamSingh/renovate/pull/32 (on Rahul's fork so I can base it on their branch)

PR #38085, opened 2025-09-17
feat(bitbucket): main branch status for mend app dashboard

zharinov:
## Changes

- Awaits: #38084
- Extends: #32165 (with unit tests)

## Context

## AI assistance disclosure

The public repository: <URL>


lib/modules/platform/bitbucket/index.spec.ts
```diff
@@ -258,6 +258,244 @@ describe('modules/platform/bitbucket/index', () => {
...
+        repoFingerprint: expect.any(String),
+      });
+
+      // Verify no POST request was made to create status (since it already exists)
+      const trace = httpMock.getTrace();
+      const postRequests = trace.filter((req) => req.method === 'POST');
+      expect(postRequests).toHaveLength(0);
+    });
+
+    it('enabled: should log debug message when creating new status', async () => {
```
RahulGautamSingh: Can't this log message be tested in the second test itself?

RahulGautamSingh (commented):
LGTM

lib/modules/platform/bitbucket/index.ts
```diff
@@ -230,6 +231,27 @@ export async function initRepo({
 
     config.defaultBranch = mainBranch;
 
+    if (bbMendAppDashboardStatus) {
```
RahulGautamSingh, line 237: I don't think this be available here, if set in repo config? The option should be self-hosted

RahulGautamSingh (changes requested):

lib/modules/platform/bitbucket/index.ts
zharinov: It's similar to `bbUseDevelopmentBranch` config option, if you search both of them, they appear in same places of codebase. So if we suppose `bbUseDevelopmentBranch` is being used in production, this one also will be available.

RahulGautamSingh: Yes. And `bbUseDevelopmentBranch` is a self hosted option :)

https://docs.renovatebot.com/self-hosted-configuration/#bbusedevelopmentbranch

zharinov: Hm, then I'm not sure being only available for self-hosted setup is what's intended. 

@setchy Do you want it to work in the Bitbucket Cloud, right?

setchy: Our current state is self-hosted renovate, running against Bitbucket Cloud.  The motivation for raising this enhancement is to simply leverage the Dependency Dashboard via Mend Developer Portal.

If/when we do migrate to using the mend-hosted bitbucket app, we would definitely need `bbUseDevelopmentBranch` and `bbMendAppDashboardStatus` to work with the hosted-app

RahulGautamSingh (approved):

zharinov:
@viceice Let's merge please, if you don't mind

lib/modules/platform/bitbucket/__snapshots__/index.spec.ts.snap
```diff
@@ -147,3 +147,5 @@ exports[`modules/platform/bitbucket/index > massageMarkdown() > returns diff fil
 exports[`modules/platform/bitbucket/index > updatePr() > rethrows exception when PR update error due to unknown reviewers error 1`] = `[HTTPError: Response code 400 (Bad Request)]`;
 
 exports[`modules/platform/bitbucket/index > updatePr() > throws an error on failure to get current list of reviewers 1`] = `[HTTPError: Response code 500 (Internal Server Error)]`;
+
+exports[`modules/platform/bitbucket/index ensureComment() does not throw 1`] = `false`;
```
viceice: where does this come from ?

viceice (commented):
otherwise LGTM

lib/modules/platform/bitbucket/__snapshots__/index.spec.ts.snap
zharinov: No idea, deleted it

lib/config/options/index.ts
```diff
@@ -2116,6 +2116,15 @@ const options: RenovateOptions[] = [
     default: false,
     supportedPlatforms: ['bitbucket'],
   },
+  {
+    name: 'bbMendAppDashboardStatus',
+    description: `Creates a main branch status linked to the repository's Mend App Dashboard.`,
+    type: 'boolean',
+    default: false,
+    supportedPlatforms: ['bitbucket'],
+    globalOnly: true,
```
viceice, line 2125: we decided that all global only options need to be read via `GlobalConfig` object, so please refactor to use that instead of passing via normal repo config

lib/modules/platform/bitbucket/index.spec.ts
```diff
@@ -260,8 +262,225 @@ describe('modules/platform/bitbucket/index', () => {
...
+        isFork: false,
+        repoFingerprint: expect.any(String),
+      });
+
+      expect(logger.logger.debug).toHaveBeenCalledWith(
+        'Creating branch status for Mend.io Dashboard',
+      );
+    });
+
+    it('enabled: should not create status when getBranchStatusCheck returns existing status', async () => {
```
RahulGautamSingh, line 434: Isn't this same as the test above+above it?

RahulGautamSingh (commented):
Else, lgtm

setchy:
Similar to the above, this week I created a custom solution within my org for main branch status which links to the Bitbucket Cloud Dependency Dashboard Issue

Screenshot 2025-12-12 at 6 38 04 AM

I'll look to add it into the renovate-cli as a config option after this version is landed.  Might call the flag `bbIssueDependencyDashboardStatus` or similar

viceice (commented):
conflicted

PR #38145, opened 2025-09-21
docs: introduce JSON schema validation and migration checks

astellingwerf:
## Changes

Introducing JSON schema validation as well as migration checks, to ensure the configuration snippets in documentation always represent valid and current settings.

## Context

## AI assistance disclosure

The public repository: <URL>


docs/usage/config-presets.md
RahulGautamSingh, line 1: Are you sure about this one?

tools/check-fenced-code.mjs
```diff
@@ -12,28 +14,91 @@ let issues = 0;
...
+/**
+ *
+ * @param {reporterCallback} reporter
+ * @param {import("../dist/config/types.js").RenovateConfig} original
+ */
+function checkMigrationStatus(reporter, original) {
+  const migrated = MigrationsService.run(original);
+  if (MigrationsService.isMigrated(original, migrated)) {
+    reporter(
+      'The JSON is contains unmigrated configuration. Migrated JSON: ' +
```
RahulGautamSingh:
```suggestion
      'The JSON contains unmigrated configuration. Migrated JSON: ' +
```

docs/usage/config-presets.md
RahulGautamSingh: I think using label will only add one label not both

astellingwerf: Well, it would, right, because we're creating that preset right here. The problem is that we're creating a preset conflicting with existing properties.

RahulGautamSingh: I meant if we change it here in the docs but the implementation is not synced the users will be confused. So we should do it in another PR

astellingwerf: To the best of my understanding it's merely an example of how user can create their own presets. The implementation only exists in this documentation.

RahulGautamSingh: There is an existing defailt `:labels` and `:label` preset. So, it will conflict I mean. Let's keep it as is

RahulGautamSingh (approved):
LGTM

docs/usage/config-presets.md
astellingwerf: I'm starting to get confused, sorry: You want me to revert this particular change, right?

.github/workflows/build.yml
```diff
@@ -213,6 +213,9 @@ jobs:
           node-version: ${{ needs.setup-build.outputs.node-version }}
           os: ${{ runner.os }}
 
+      - name: Build
+        run: pnpm build
+
```
viceice: why those changes, shouldn't be necessary

astellingwerf: Because `tools/check-fenced-code.mjs` now needs a compiled version of the migration service to be present. I'm open to suggestions of other ways to attain the same functionality.

docs/usage/config-presets.md
RahulGautamSingh: Yes

.github/workflows/build.yml
astellingwerf: @viceice, would it make sense for me to convert this file to typescript, so that we can run it with `tsx` without compiling first? I don't know the background behind this not being TS, do you know maybe?

viceice: feel free to convert to typescript and call via tsx. it was written as mjs to save time for compilation step.

astellingwerf: I split it off into #38190 for a cleaner commit history. I'll merge/rebase after that's been merged.

lib/modules/datasource/custom/readme.md
```diff
@@ -360,6 +360,8 @@ To work around a missing API, you can create dependency "files".
 These files are served via HTTP(S), so that Renovate can access them.
 For example, imagine the following file `versiontracker.json` for the software `something`:
 
+<!-- schema-validation-ignore -->
```
viceice: does it only ignore next json block?

lib/modules/manager/custom/jsonata/readme.md
```diff
@@ -67,9 +67,9 @@ When you configure a JSONata manager, use the following syntax:
 ```
 
 Overwrite the `<query>` placeholder text with your [JSONata](https://docs.jsonata.org/overview.html) query.
-The JSONata query transforms the content to a JSON object, similar to the this:
+The JSONata query transforms the content to a JSON object, similar to this:
 
-```javascript title="dependencies information extracted usig jsonata query"
+```javascript title="dependencies information extracted using a JSONata query"
```
viceice: should be json instead of JavaScript

tools/check-fenced-code.ts
```diff
@@ -1,50 +1,110 @@
+import { format } from 'node:util';
+import type { ValidateFunction } from 'ajv';
+// eslint-disable-next-line import-x/no-named-as-default
```
viceice: false positive? or better use named import?

tools/check-fenced-code.ts
```diff
@@ -1,50 +1,110 @@
...
+import Ajv from 'ajv';
+import * as draft04 from 'ajv/lib/refs/json-schema-draft-04.json';
```
viceice:
```suggestion
import draft04 from 'ajv/lib/refs/json-schema-draft-04.json';
```
this should work 

https://github.com/renovatebot/renovate/blob/af378c601a572a5dbff486388b0b1486b1c3a74c/lib/config/presets/internal/replacements.ts#L1

tools/check-fenced-code.ts
```diff
@@ -1,50 +1,110 @@
...
+
+function checkSchemaCompliantJson(
+  reporter: ReporterCallback,
+  value: object,
+): RenovateConfig | undefined {
+  const isValid = validate(value);
+  if (isValid) {
+    return value as RenovateConfig;
+  }
+  validate.errors?.forEach((error) => {
```
viceice: use for-of loop

tools/check-fenced-code.ts
```diff
@@ -1,50 +1,110 @@
...
+    reporter(error.dataPath + ' ' + error.message);
```
viceice: use string template

tools/check-fenced-code.ts
```diff
@@ -1,50 +1,110 @@
...
+  for (const [index, token] of tokens.entries()) {
+    if (
+      token.type !== 'fence' ||
+      !(token.info === 'json' || token.info.startsWith('json '))
+    ) {
+      continue;
     }
-  });
+
+    const reporter: ReporterCallback = (message) => {
```
viceice: please no inline functions and use normal function declaration

lib/modules/datasource/custom/readme.md
astellingwerf: You're correct, I updated the name accordingly

tools/check-fenced-code.ts
```diff
@@ -1,50 +1,109 @@
...
+      token.type !== 'fence' ||
+      !(token.info === 'json' || token.info.startsWith('json '))
+    ) {
+      continue;
+    }
+
+    const validJson = checkValidJson(file, token);
+    if (
+      validJson === undefined ||
+      tokens[index - 2]?.content ===
```
viceice: use `Array.at()`

tools/check-fenced-code.ts
```diff
@@ -1,50 +1,109 @@
...
 const errorBody =
   'Fix this manually by ensuring each block is a valid, complete JSON document.';
+const errorLogFormat = process.env.CI
+  ? `::error file=%s,line=%d,endLine=%d,title=${errorTitle}::%s. ${errorBody}`
+  : `${errorTitle} (%s lines %d-%d): %s`;
+
+function reportIssue(file: string, token: Token, message: string): void {
+  const [start, end] = token.map ?? [-1, -1];
+  issues += 1;
+  console.log(format(errorLogFormat, file, start + 1, end + 1, message));
```
jamietanna:
```suggestion
  console.error(format(errorLogFormat, file, start + 1, end + 1, message));
```

jamietanna (commented):
Thanks very much for this @astellingwerf, this looks great!

Looks like there a coupel of last minor comments, then I'd say let's get this merged 🚀

tools/check-fenced-code.ts
```diff
@@ -1,50 +1,109 @@
...
+  }
+}
+
+function checkMigrationStatus(
+  file: string,
+  token: Token,
+  original: RenovateConfig,
+): void {
+  const migrated = MigrationsService.run(original);
+  if (MigrationsService.isMigrated(original, migrated)) {
```
jamietanna, line 62: This is a question about the existing method - is this right? I'd expect - from the name - that this would return `true` if it's _already_ migrated, not if it needs to be migrated

astellingwerf: We provide the original and the migrated instance of the configuration. If they differ, we must assume that the `MigrationService.run` did something to migrate some legacy configuration.

jamietanna:
I see that https://github.com/renovatebot/renovate/pull/38419 could've been caught by this, but wouldn't be as it's marked as a `javascript` code block, not `json`

Not something I want us to solve as part of either PR, but something we should look at whether we can make our Renovate config snippets consistent

tools/check-fenced-code.ts
jamietanna: Ta - looks like this was my incorrect assumption. I've raised https://github.com/renovatebot/renovate/pull/38421 for future reference

lib/modules/manager/custom/jsonata/readme.md
jamietanna, line 1: Were some of the wording changes things we'd added while we were in this file? Just tryign to confirm where they came from

jamietanna (approved):
Awesome stuff 🚀 

Will see if the merge commit coming in from `main` works OK, and that last outstanding question just to confirm, then this looks good to go in - thanks very much!

lib/modules/manager/custom/jsonata/readme.md
astellingwerf: While correcting some of the code blocks, I also came across some typos or grammatical issues. I thought I'd update it while we're working on the file already.

PR #38243, opened 2025-09-26
fix(migrations): correct expectation in line with implementation

jamietanna:
## Changes

While working on #38240, I noticed that the failing tests I was writing
didn't trigger.

To fix this, I've raised #38242. Before this, however, we should fix the
currently failing test on main, by correcting the expected migration.

This is the existing behaviour of the implementation, which may not be
expected.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/config/migrations/custom/file-match-migration.spec.ts
+++ b/lib/config/migrations/custom/file-match-migration.spec.ts
@@ -30,7 +30,7 @@ describe('config/migrations/custom/file-match-migration', () => {
         managerFilePatterns: ['filename3'],
       },
       {
-        managerFilePatterns: ['/filename1/', '/filename2/', 'filename3'],
+        managerFilePatterns: ['filename3', '/filename1/', '/filename2/'],
       },
     );
   });


lib/config/migrations/custom/file-match-migration.spec.ts
```diff
@@ -30,7 +30,7 @@ describe('config/migrations/custom/file-match-migration', () => {
         managerFilePatterns: ['filename3'],
       },
       {
-        managerFilePatterns: ['/filename1/', '/filename2/', 'filename3'],
+        managerFilePatterns: ['filename3', '/filename1/', '/filename2/'],
```
jamietanna, line 33: This is what the implementation says it should be - is that correct? Or have we maybe broken something?

RahulGautamSingh: Good catch! Yes, your changed version is correct

RahulGautamSingh (approved):

viceice (approved):

PR #38267, opened 2025-09-28
feat(allowedMinimumReleaseAge): Add new allowedMinimumReleaseAge option

onigoetz:
## Changes

This PR introduces a new option `allowedMinimumReleaseAge` that's similar to `minimumReleaseAge` but behaves like `allowedVersions`: it filters out versions that are too young and only sends PRs for versions older than `allowedMinimumReleaseAge`.

The reason for this PR is that behaviour of Renovate's `minimumReleaseAge` clashes with Yarn's `npmMinimalAgeGate`.
If you set both `minimumReleaseAge` and `npmMinimalAgeGate` to 3 days; Renovate will create a PR immediately, and Yarn will refuse to install the package as it's too young.

`allowedMinimumReleaseAge` works with the same heuristic as `npmMinimalAgeGate`: only install/create PRs for packages older than X.

This is a complement to the comment I added to the discussion here: https://github.com/renovatebot/renovate/discussions/38247#discussioncomment-14530157

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2720,7 +2720,7 @@ The datasource that Renovate uses must have a release timestamp for the `minimum
 Some datasources may have a release timestamp, but in a format Renovate does not support.
 In those cases a feature request needs to be implemented.
 
-You might alsow want to look into the [`allowedMinimumReleaseAge`](#allowedminimumeeleaseage) packageRule that will only open with versions that are older than the specified duration.
+You might alsow want to look into the [`allowedMinimumReleaseAge`](#allowedminimumreleaseage) packageRule that will only open with versions that are older than the specified duration.
 
 <!-- prettier-ignore -->
 !!! warning "Warning for Maven users"
@@ -2943,6 +2943,34 @@ For example you have multiple `package.json` and want to use `dependencyDashboar
 !!! warning
     Avoid nesting any `object`-type configuration in a `packageRules` array, such as a `major` or `minor` block.
 
+### allowedMinimumReleaseAge
+
+You can use `allowedMinimumReleaseAge` - usually within a `packageRules` entry - to limit the allowed versions.
+
+For example if you only want NPM versions that are at least three days old at the moment of the Pull Request:
+
+```json
+{
+  "packageRules": [
+    {
+      "matchDatasources": ["npm"],
+      "allowedMinimumReleaseAge": "3 days"
+    }
+  ]
+}
+```
+
+This option is similar to `minimumReleaseAge`. Here is an example to illustrate the difference.
+
+Your application currently has a dependency on `@ctrl/tinycolor` at version `4.1.0`
+On the 15th of september 2025, version `4.1.1` was released.
+We're the 16th of september.
+
+Here's what happens with each setting;
+
+- `allowedMinimumReleaseAge: "3 days"` Will wait two more days to open a Pull Request.
+- `minimumReleaseAge: "3 days"` Will open a Pull Request as soon as it sees the new version but adds a "renovate/stability-days" pending status check.
+
 ### allowedVersions
 
 You can use `allowedVersions` - usually within a `packageRules` entry - to limit how far to upgrade a dependency.
@@ -2965,33 +2993,6 @@ Renovate calculates the valid syntax for this at runtime, because it depends on 
 <!-- prettier-ignore -->
 !!! warning
     `allowedVersions` and `matchUpdateTypes` cannot be used in the same package rule.
-
-### allowedMinimumReleaseAge
-
-You can use `allowedMinimumReleaseAge` - usually within a `packageRules` entry - to limit the allowed versions.
-
-For example if you only want NPM versions that are at least three days old at the moment of the Pull Request: 
-
-```json
-{
-  "packageRules": [
-    {
-      "matchDatasources": ["npm"],
-      "allowedMinimumReleaseAge": "3 days"
-    }
-  ]
-}
-```
-
-This option is similar to `minimumReleaseAge`. Here is an example to illustrate the difference.
-
-Your application currently has a dependency on `@ctrl/tinycolor` at version `4.1.0`
-On the 15th of september 2025, version `4.1.1` was released.
-We're the 16th of september. 
-
-Here's what happens with each setting;
-* `allowedMinimumReleaseAge: "3 days"` Will wait two more days to open a Pull Request.
-* `minimumReleaseAge: "3 days"` Will open a Pull Request as soon as it sees the new version but adds a "renovate/stability-days" pending status check.
 
 #### Using regular expressions
 
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2720,7 +2720,7 @@ The datasource that Renovate uses must have a release timestamp for the `minimum
 Some datasources may have a release timestamp, but in a format Renovate does not support.
 In those cases a feature request needs to be implemented.
 
-You might also want to look into the [`allowedMinimumReleaseAge`](#allowedminimumreleaseage) packageRule that will only create updates with versions that are older than the specified duration.
+You might also want to look into the [`allowedMinimumReleaseAge`](#allowedminimumreleaseage) configuration that will only create updates with versions that are older than the specified duration.
 
 <!-- prettier-ignore -->
 !!! warning "Warning for Maven users"
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2720,6 +2720,8 @@ The datasource that Renovate uses must have a release timestamp for the `minimum
 Some datasources may have a release timestamp, but in a format Renovate does not support.
 In those cases a feature request needs to be implemented.
 
+You might alsow want to look into the [`allowedMinimumReleaseAge`](#allowedminimumeeleaseage) packageRule that will only open with versions that are older than the specified duration.
+
 <!-- prettier-ignore -->
 !!! warning "Warning for Maven users"
     For `minimumReleaseAge` to work, the Maven source must return reliable `last-modified` headers.
@@ -2963,6 +2965,33 @@ Renovate calculates the valid syntax for this at runtime, because it depends on 
 <!-- prettier-ignore -->
 !!! warning
     `allowedVersions` and `matchUpdateTypes` cannot be used in the same package rule.
+
+### allowedMinimumReleaseAge
+
+You can use `allowedMinimumReleaseAge` - usually within a `packageRules` entry - to limit the allowed versions.
+
+For example if you only want NPM versions that are at least three days old at the moment of the Pull Request: 
+
+```json
+{
+  "packageRules": [
+    {
+      "matchDatasources": ["npm"],
+      "allowedMinimumReleaseAge": "3 days"
+    }
+  ]
+}
+```
+
+This option is similar to `minimumReleaseAge`. Here is an example to illustrate the difference.
+
+Your application currently has a dependency on `@ctrl/tinycolor` at version `4.1.0`
+On the 15th of september 2025, version `4.1.1` was released.
+We're the 16th of september. 
+
+Here's what happens with each setting;
+* `allowedMinimumReleaseAge: "3 days"` Will wait two more days to open a Pull Request.
+* `minimumReleaseAge: "3 days"` Will open a Pull Request as soon as it sees the new version but adds a "renovate/stability-days" pending status check.
 
 #### Using regular expressions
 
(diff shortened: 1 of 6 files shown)


onigoetz:
Oh I missed the https://docs.renovatebot.com/configuration-options/#internalchecksfilter option in the documentation 🤯 I will try that

onigoetz:
> Oh I missed the https://docs.renovatebot.com/configuration-options/#internalchecksfilter option in the documentation 🤯 I will try that

Oh I see but "strict" is already the default, so my original point still stands; the feature works in a confusing way.
* It's useful as a user to see that a new version will be available, but in the present case (combine the feature with Yarn) Yarn will still fail to update the package.
* How does that work with grouped PRs? let's say we have a project with many dependencies, and we regularly get updates; will the PR never be opened because there is always a new dependency waiting for its `minimumReleaseAge` ?

lib/config/validation.ts
```diff
@@ -423,6 +423,7 @@ export async function validateConfig(
                   // It's too late to apply any of these options once you already have updates determined
                   const preLookupOptions = [
                     'allowedVersions',
+                    'allowedMinimumReleaseAge',
```
astellingwerf:
```suggestion
                    'allowedMinimumReleaseAge',
                    'allowedVersions',
```

lib/workers/repository/process/lookup/types.ts
```diff
@@ -12,6 +12,7 @@ import type { Timestamp } from '../../../../util/timestamp';
 
 export interface FilterConfig {
   allowedVersions?: string;
+  allowedMinimumReleaseAge?: string;
```
astellingwerf: The addition alphabetically comes first, so please insert it as the first item.

docs/usage/configuration-options.md
```diff
@@ -2941,6 +2943,34 @@ For example you have multiple `package.json` and want to use `dependencyDashboar
...
+}
+```
+
+This option is similar to `minimumReleaseAge`. Here is an example to illustrate the difference.
+
+Your application currently has a dependency on `@ctrl/tinycolor` at version `4.1.0`
+On the 15th of september 2025, version `4.1.1` was released.
+We're the 16th of september.
+
+Here's what happens with each setting;
```
```suggestion
Here's what happens with each setting:
```

docs/usage/configuration-options.md
```diff
@@ -2941,6 +2943,34 @@ For example you have multiple `package.json` and want to use `dependencyDashboar
...
+
+- `allowedMinimumReleaseAge: "3 days"` Will wait two more days to open a Pull Request.
+- `minimumReleaseAge: "3 days"` Will open a Pull Request as soon as it sees the new version but adds a "renovate/stability-days" pending status check.
```
```suggestion
- `allowedMinimumReleaseAge: "3 days"` will wait two more days to open a Pull Request.
- `minimumReleaseAge: "3 days"` will open a Pull Request as soon as it sees the new version, but adds a "renovate/stability-days" pending status check.
```

docs/usage/configuration-options.md
```diff
@@ -2941,6 +2943,34 @@ For example you have multiple `package.json` and want to use `dependencyDashboar
...
+    }
+  ]
+}
+```
+
+This option is similar to `minimumReleaseAge`. Here is an example to illustrate the difference.
+
+Your application currently has a dependency on `@ctrl/tinycolor` at version `4.1.0`
+On the 15th of september 2025, version `4.1.1` was released.
+We're the 16th of september.
```
```suggestion
On the 15th of September 2025, version `4.1.1` was released.
We're the 16th of September.
```

docs/usage/configuration-options.md
```diff
@@ -2720,6 +2720,8 @@ The datasource that Renovate uses must have a release timestamp for the `minimum
 Some datasources may have a release timestamp, but in a format Renovate does not support.
 In those cases a feature request needs to be implemented.
 
+You might alsow want to look into the [`allowedMinimumReleaseAge`](#allowedminimumreleaseage) packageRule that will only open with versions that are older than the specified duration.
```
```suggestion
You might also want to look into the [`allowedMinimumReleaseAge`](#allowedminimumreleaseage) packageRule that will only create updates with versions that are older than the specified duration.
```

docs/usage/configuration-options.md
```diff
@@ -2941,6 +2943,34 @@ For example you have multiple `package.json` and want to use `dependencyDashboar
 !!! warning
     Avoid nesting any `object`-type configuration in a `packageRules` array, such as a `major` or `minor` block.
 
+### allowedMinimumReleaseAge
+
+You can use `allowedMinimumReleaseAge` - usually within a `packageRules` entry - to limit the allowed versions.
```
astellingwerf, line 2948:
```suggestion
You can use `allowedMinimumReleaseAge`, usually within a `packageRules` entry, to limit the allowed versions.
```

astellingwerf (changes requested):

docs/usage/configuration-options.md
onigoetz: I can change it but I took the exact phrasing and formatting of `allowedVersions`:
https://github.com/onigoetz/renovate/blob/73cd355167eb1b50dd265acefafabb9afbc1c2c8/docs/usage/configuration-options.md?plain=1#L2976

astellingwerf: OK, fair enough, let's keep it as is then

onigoetz:
@astellingwerf I addressed the requested changes

docs/usage/configuration-options.md
```diff
@@ -2720,6 +2720,8 @@ The datasource that Renovate uses must have a release timestamp for the `minimum
 Some datasources may have a release timestamp, but in a format Renovate does not support.
 In those cases a feature request needs to be implemented.
 
+You might also want to look into the [`allowedMinimumReleaseAge`](#allowedminimumreleaseage) packageRule that will only create updates with versions that are older than the specified duration.
```
```suggestion
You might also want to look into the [`allowedMinimumReleaseAge`](#allowedminimumreleaseage) configuration that will only create updates with versions that are older than the specified duration.
```

astellingwerf (changes requested):

astellingwerf (approved):

onigoetz:
Hi @astellingwerf I see the PR is approved. What are the next steps ?

astellingwerf:
> Hi @astellingwerf I see the PR is approved. What are the next steps ?

Besides my approval, we'll need the approval of someone who's authorized to merge.

Churro:
>If you set both minimumReleaseAge and npmMinimalAgeGate to 3 days; Renovate will create a PR immediately

Did you take a deeper look why a PR is created immediately in your case? This should actually not happen with `internalChecksFilter="strict"`, see https://docs.renovatebot.com/configuration-options/#suppress-branchpr-creation-for-x-days

I'd find it very confusing to have two options named almost identical, apart from the `allowed` prefix. `minimumReleaseAge` was previously called `stabilityDays` and aimed for exactly the same purpose. It's likely that an end-user, unfamiliar with the intrinsics of renovate, might configure the values differently, raising even more confusion about PRs being or not being created.

onigoetz:
> Did you take a deeper look why a PR is created immediately in your case? This should actually not happen with internalChecksFilter="strict", see https://docs.renovatebot.com/configuration-options/#suppress-branchpr-creation-for-x-days

`internalChecksFilter="strict"` is the default and that's what I had so I was still able to get a PR for a too-fresh package

It is possible that I clicked in the Dependency Dashboard and triggered the PR myself, but that's still an issue for two reasons
1. We could still end up installing packages that are vulnerable because they are too fresh; I would expect the strict setting to completely hide packages that are too fresh, not installable by accident
2. When the repositories uses Yarn with `npmMinimalAgeGate` it strictly respects the set limit and thus fails the installation (both when creating the PR in Renovate and when running in CI)

> I'd find it very confusing to have two options named almost identical, apart from the allowed prefix. minimumReleaseAge was previously called stabilityDays and aimed for exactly the same purpose. It's likely that an end-user, unfamiliar with the intrinsics of renovate, might configure the values differently, raising even more confusion about PRs being or not being created.

I totally agree with your statement, I would find it confusing as well, I tried to make it as clear as possible in the documentation on the difference on the two.

My goal is to make sure that `minimumReleaseAge` + `internalChecksFilter="strict"` _never_ opens a PR for a package younger than `minimumReleaseAge`; this PR achieves this goal. but there might be other ways?

I also have the impression that when grouping PRs (either the PR is held indefinitely or the PR is open with too young versions). But I didn't test this hypothesis.

I understand that `minimumReleaseAge` tries to give as much information as possible to user that an update exists but will not be ready for some time. But I also like the simple elegance of `allowedVersions` where it just excludes versions that you tell it to exclude without possible recourse

RahulGautamSingh:
> My goal is to make sure that minimumReleaseAge + internalChecksFilter="strict" never opens a PR for a package younger than minimumReleaseAge; this PR achieves this goal. but there might be other ways?


https://docs.renovatebot.com/configuration-options/#prcreation

This will do the job of not creating new PRs if branch has pending/non-green status checks.

Sorry for the late response

onigoetz:
> This will do the job of not creating new PRs if branch has pending/non-green status checks.

Okay, might be because of poor wording on my side; let me rephrase; I want to guarantee that we NEVER open a PR nor even a branch on our codebase with a dependency that is younger than X days.
In the case of a breach like shai hulud or the previous breach with the crypto wallet thingy, opening a PR or even a branch will immediately run in our CI and might thus infect the machines on which it runs. Our systems work in a way that these breaches had zero impact due to other security measures, but we wish to protect ourselves from future breaches.

For now all options that were proposed (minimumReleaseAge + internalChecksFilter="strict" + prCreation) still have the following issues:

1. We could still end up installing packages that are vulnerable because they are too fresh; I would expect the strict setting to completely hide packages that are too fresh, and not installable by accident
2. When the repositories uses Yarn with `npmMinimalAgeGate` it strictly respects the set limit and thus fails the installation (both when creating the branch through Renovate and when running in CI)

RahulGautamSingh:
Can you reproduce a case where the branch get created when minimum release age is not met?

The discussion you have linked, specifically talks about the case where branch already existed before the `minimumReleaseAge` was configured.

alan-agius4:
@RahulGautamSingh, in the issue linked, which I am the author of, the branch was created after `minimumReleaseAge` was created. This can for example happen when a PR gets merged, and the branch is not deleted and in which case renovate will open a PR.

RahulGautamSingh:
> @RahulGautamSingh, in the issue linked, which I am the author of, the branch was created after `minimumReleaseAge` was created. This can for example happen when a PR gets merged, and the branch is not deleted and in which case renovate will open a PR.

1. In your case that is expected behaviour
2. You can use `prCreation` option to prevent that

RahulGautamSingh:
This PR is getting confusing with 2 topics being discussed so let's move to discussions   

@alan-agius4 Please update your discussion to make it clear what you need fixed? Cause if you want renovate to not create PRs from old branches when updates have not met the release age you can use `prCreation` option for that.

@onigoetz Please create a new discussion with your requirements so that we can discuss if it's a feature we'd like to support or not.

onigoetz:
Yes sure I'll open a discussion and yes, I have an example I can provide

onigoetz:
@RahulGautamSingh here is the discussion with all relevant details: https://github.com/renovatebot/renovate/discussions/38324

jamietanna:
Moving this into draft temporarily - we're discussing in https://github.com/renovatebot/renovate/discussions/38324 and it may be that there's a bug that we need to fix in a different way to this PR 👍

jamietanna:
@onigoetz mind seeing if the proposal in https://github.com/renovatebot/renovate/pull/38363 would solve your ask here?

onigoetz:
> @onigoetz mind seeing if the proposal in #38363 would solve your ask here?
 
Hi @jamietanna, it's an interesting proposal but it doesn't solve the same as what this PR solves and what's being discussed in #38324

jamietanna:
This should be solved by https://github.com/renovatebot/renovate/pull/38720 🤞

onigoetz:
Closing this PR as you've implemented something else.

I might reopen an issue as I still see the same issues on my open source projects ( PRs created even if packages haven't reached the age threshold or updates to add packages that haven't reached the age threshold )

jamietanna:
Thanks - I'd meant to come back to do the same. Thanks for the feedback + suggestion here, and helping us work on improving the `minimumReleaseAge` support! 

> PRs are created even if ...

https://docs.renovatebot.com/key-concepts/minimum-release-age/#recommended-settings should catch this - if that's not happening, that's more of a bug, so let's investigate and fix 👍

onigoetz:
Thanks @jamietanna, I'll dig through my logs and provide a solid case if that's indeed happening :)

PR #38352, opened 2025-10-02
chore: log when pending upgrades are filtered 

RahulGautamSingh:
## Changes

- Add a log for when pending updates are filtered from a branch with multiple updates

## Context

## AI assistance disclosure

The public repository: <URL>


lib/workers/repository/updates/generate.ts
```diff
@@ -191,6 +191,7 @@ export function generateBranchConfig(
   let branchUpgrades = upgrades;
   if (!branchUpgrades.every((upgrade) => upgrade.pendingChecks)) {
     // If the branch isn't pending, then remove any upgrades within which *are*
+    logger.debug('Branch is not pending, removing pending upgrades');
```
jamietanna: Worth also logging which branch we're operating on, which is being skipped?

lib/workers/repository/update/pr/pr-reuse.ts
```diff
@@ -4,11 +4,10 @@ import { logger } from '../../../../logger';
 import { platform } from '../../../../modules/platform';
 import type { Pr } from '../../../../modules/platform/types';
 
-const REOPEN_THRESHOLD_MILLIS = 1000 * 60 * 60 * 24 * 7;
+const REOPEN_THRESHOLD_MILLIS = 1000 * 60 * 60 * 24 * 7; // 7 days
```
jamietanna: Would a change like:


```suggestion
const ONE_DAYS_MILLIS = 1000 * 60 * 60 * 24;
const REOPEN_THRESHOLD_MILLIS = 7 * ONE_DAYS_MILLIS;
```

be reasonable in the style we follow in the Renovate codebase, as well as what could be seen as more idiomatic Typescript?

(genuinely unsure - looking for your experience and input!)

erikjuhani: > [!note]
> Edit: Wanted to come back to this as perhaps a better term would be `TimeUnitMs` or `DurationMs`, which includes the base unit in the name.

Hey, I randomly ended up here, but I've thought about this exact same problem and this solution feels readable and nice to use, particularly the lsp hints.

> ```ts
> enum Duration {
>     Millisecond = 1,
>     Second = Duration.Millisecond * 1000,
>     Minute = Duration.Second * 60,
>     Hour = Duration.Minute * 60,
>     Day = Duration.Hour * 24,
>     Week = Duration.Day * 7
> }
> 
> const REOPEN_THRESHOLD_MILLIS = Duration.Week;
> // ^ 604800000
> ```
> — [ts-playground-link](https://www.typescriptlang.org/play/?#code/KYOwrgtgBAImBOBDALgSwPYigbwFBQKgFlUAbU1AZ2AGNMATKAXigEYAafQgZVoedgIUGEADoS5KnxCMAVGwAMSzoWKpwyYALhI0mUbzoyo8gGwKVhABLoE2oXrEkNWsxa4EYiAJ73dI0Rs7eQAmABZLAgB1YGAAaz9hfS9feQB2XABfXFwjSmQoACUAUQB5AAVigDkAfQAVKxLuK1KAGRgaogBJVtau7kTHURj4gG5cAHoJqAA9KHMwgA4lFaA)
> 


Screenshot 2025-10-05 at 14 45 59
Screenshot 2025-10-05 at 14 45 49

RahulGautamSingh: I like the above idea by @erikjuhani. But since this change is unrelated to this PR I will create a separate one for it

jamietanna:
@RahulGautamSingh mind looking at the failing test?

RahulGautamSingh:
> @RahulGautamSingh mind looking at the failing test?

That's actually a lucky test fail.

We were filtering out non-pending for both when all updates were all non-pending or updates mixed non-pending + pending.

But we only want to filter if upgrades are mixed ones. I will fix the code

jamietanna (commented):
Could we look at whether we need to add some additional tests to cover this behaviour?

viceice (commented):
shouldn't there a new test to verify?

lib/workers/repository/updates/generate.spec.ts
```diff
@@ -1846,5 +1846,72 @@ describe('workers/repository/updates/generate', () => {
...
+      ] satisfies BranchUpgradeConfig[];
+      let res = generateBranchConfig(branch);
+      expect(res.pendingChecks).toBeFalse();
+      expect(logger.logger.debug).toHaveBeenCalledWith(
+        { branch: branch[0].branchName },
+        'Branch is not pending, removing pending upgrades',
+      );
+
+      // set all upgrades to not-pending
+      branch[0].pendingChecks = false;
```
jamietanna: Is this comment right? We're only setting one upgrade

jamietanna (changes requested):
If I remove the implementation changes in `generate.ts`, the test still passes - can we double check we're testing the right things?

(I'll look more in more detail tomorrow/Thursday - wanted to pop this comment on before I left for the day!)

lib/workers/repository/updates/generate.spec.ts
RahulGautamSingh: others are already set to `false`

RahulGautamSingh:
I had a typo ie. `toHaveBeenCalledWith` instead of `toHaveBeenCalled` + I needed to reset the mocks.

Thanks for going the effort to test on old code and catching this, while the code is correct the test was flaky.

lib/workers/repository/updates/generate.spec.ts
```diff
@@ -1846,5 +1846,74 @@ describe('workers/repository/updates/generate', () => {
         );
       },
     );
+
+    it('only filter pending upgrades, when both pending and non-pending upgrades are present', () => {
```
secustor: IMO this should be two different test cases. Then you do not have to reset the mocks

lib/workers/repository/updates/generate.ts
```diff
@@ -189,10 +189,22 @@ export function generateBranchConfig(
...
+  );
+  if (hasPendingUpgrades && hasNonPendingUpgrades) {
     // If the branch isn't pending, then remove any upgrades within which *are*
+    logger.debug(
+      { branch: branchUpgrades[0].branchName },
+      'Branch is not pending, removing pending upgrades',
+    );
     branchUpgrades = branchUpgrades.filter((upgrade) => !upgrade.pendingChecks);
   }
+
```
jamietanna: Is this extra newline needed?

```suggestion
```

lib/workers/repository/updates/generate.ts
```diff
@@ -189,10 +189,22 @@ export function generateBranchConfig(
...
+  const hasPendingUpgrades = branchUpgrades.some(
+    (upgrade) => upgrade.pendingChecks,
+  );
+  const hasNonPendingUpgrades = branchUpgrades.some(
+    (upgrade) => !upgrade.pendingChecks,
+  );
+  if (hasPendingUpgrades && hasNonPendingUpgrades) {
     // If the branch isn't pending, then remove any upgrades within which *are*
+    logger.debug(
+      { branch: branchUpgrades[0].branchName },
```
jamietanna, line 195: Will the `branchUpgrades[*].branchName` always be the same?

lib/workers/repository/updates/generate.ts
```diff
@@ -189,10 +189,22 @@ export function generateBranchConfig(
...
 ): BranchConfig {
   let branchUpgrades = upgrades;
-  if (!branchUpgrades.every((upgrade) => upgrade.pendingChecks)) {
+
+  const hasPendingUpgrades = branchUpgrades.some(
+    (upgrade) => upgrade.pendingChecks,
+  );
+  const hasNonPendingUpgrades = branchUpgrades.some(
+    (upgrade) => !upgrade.pendingChecks,
+  );
```
jamietanna: Alternatively, this could be a single loop with two booleans stored outside of the loop - could be more efficient in some cases (but not sure we need to micro-optimise here)

lib/workers/repository/updates/generate.ts
RahulGautamSingh: Yes

lib/workers/repository/updates/generate.ts
RahulGautamSingh: Yes, readability here is better

jamietanna (changes requested):
I can only see an assertion on the log message - can we validate that the filtering has actually been applied? Otherwise I can revert the functionality, keep a log message, and the tests still pass

RahulGautamSingh:
I have reverted my changes to only add a log msg. Now, we will see some tests fail. This happens because the log msg gets triggered even when there are no pending releases and we want to prevent that.

secustor (commented):
Needs a new PR title, else LGTM

secustor (approved):

jamietanna (approved):

PR #37967, opened 2025-09-11
feat(best-practices)!: provide default `minimumReleaseAge` for npm

jamietanna:
## Changes

As noted in #37952, we've seen a significant increase in recent years on
a number of package ecosystems being targeted with supply chain security
attacks.

We're in a particularly strong position to improve the level of security
across the world by setting a default for `minimumReleaseAge` in the
`best-practices` config preset.

To do this, we can introduce a new preset,
`security:minimum-release-age` to house these recommended options.

We use new presets `security:minimum-release-age` and
`security:minimum-release-age-npm`, to allow consumers to utilise
`ignorePresets` to undo this configuration, if they so wish.

As a first step towards securing our users, we can set a default for
npm, which as a large ecosystem with an increased number of packages
that depend on other packages (compared to other ecosystems), has been a
more targeted area.

In the future, we can expand this for other managers, as it makes sense,
and as support is added for them to use `minimumReleaseAge`.

- #37952.

## Context

Please select one of the below:

- [x] This closes an existing Issue: #37952
- [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [ ] No documentation update is required

We've not yet decided if there's anything else (via https://github.com/renovatebot/renovate/discussions/37952) that needs updating

## How I've tested my work (please select one)

I have verified these changes via:

- [x] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: N/A

PR #38363, opened 2025-10-03
feat(minimumReleaseAge): optionally enforce presence of `releaseTimestamp`

jamietanna:
## Changes

As noted in #38290, #38348, a "rough edge" of the current implementation
of `minimumReleaseAge` is that:

- if we have set `minimumReleaseAge` to apply to a given dependency
- and we have 3 dependencies that need an update
  - 2 of which have a `releaseTimestamp` that is not yet "stable"
  - 1 of which does not have a `releaseTimestamp`

The dependency update that does not have a `releaseTimestamp` will be
treated as "stable", and will receive an update immediately.

To make this a little more predictable, and explicit, we should add a
new configuration item, `minimumReleaseAgeBehaviour`, to control this
behaviour.

This adds the new configuration field, as well as making it available as
a strongly typed field in the codebase.

In the case that `minimumReleaseAgeBehaviour=timestamp-optional` is
being used, we will WARN on any packages that are being updated, but
maybe shouldn't be, as a way to proactively inform the user of possible
misconfiguration.

- #38290
- #38348.

## Context

- [ ] This closes an existing Issue: #38290, #38348

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2707,7 +2707,7 @@ Example:
 
 This feature used to be called `stabilityDays`.
 
-If `minimumReleaseAge` is set to a time duration _and_ the update has a release timestamp header, then Renovate will check if the set duration has passed.
+If `minimumReleaseAge` is set to a time duration _and_ the update has a release timestamp header, then Renovate will check if the set duration has passed. This behaviour can be changed using [`minimumReleaseAgeBehaviour`](#minimumreleaseagebehaviour).
 
 Note: Renovate will wait for the set duration to pass for each **separate** version.
 Renovate does not wait until the package has seen no releases for x time-duration(`minimumReleaseAge`).
@@ -2774,6 +2774,16 @@ Renovate adds a "renovate/stability-days" pending status check to each branch/PR
 This pending check prevents the branch going green to automerge before the time has passed.
 
 <!-- markdownlint-enable MD001 -->
+
+## minimumReleaseAgeBehaviour
+
+When `minimumReleaseAge` is set to a time duration, the `minimumReleaseAgeBehaviour` will be used to control whether the release timestamp is required.
+
+When set to `timestamp-required`, this version is not treated stable unless there is release timestamp, and that release timestamp is past the [`minimumReleaseAge`](#minimumreleaseage).
+
+When set to `timestamp-optional`, Renovate will treat a release without a releaseTimestamp as stable.
+
+This only applies when used with [`minimumReleaseAge`](#minimumreleaseage).
 
 ## minor
 
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -1967,6 +1967,14 @@ const options: RenovateOptions[] = [
     default: null,
   },
   {
+    name: 'minimumReleaseAgeBehaviour',
+    description:
+      'When set in conjunction with `minimumReleaseAge`, controls whether the `releaseTimestamp` for a dependency update is required.',
+    type: 'string',
+    default: 'timestamp-optional',
+    allowedValues: ['timestamp-required', 'timestamp-optional'],
+  },
+  {
     name: 'abandonmentThreshold',
     description:
       'Flags packages that have not been updated within this period as abandoned.',
--- a/lib/config/types.ts
+++ b/lib/config/types.ts
@@ -219,6 +219,9 @@ export type RenovateRepository =
 
 export type UseBaseBranchConfigType = 'merge' | 'none';
 export type ConstraintsFilter = 'strict' | 'none';
+export type MinimumReleaseAgeBehaviour =
+  | 'timestamp-optional'
+  | 'timestamp-required';
 
 export const allowedStatusCheckStrings = [
   'minimumReleaseAge',
@@ -326,6 +329,8 @@ export interface RenovateConfig
   additionalBranchPrefix?: string;
   sharedVariableName?: string;
   minimumGroupSize?: number;
+
+  minimumReleaseAgeBehaviour?: MinimumReleaseAgeBehaviour;
 }
 
 const CustomDatasourceFormats = [
--- a/lib/workers/repository/process/lookup/filter-checks.ts
+++ b/lib/workers/repository/process/lookup/filter-checks.ts
@@ -1,5 +1,6 @@
 import is from '@sindresorhus/is';
 import { mergeChildConfig } from '../../../../config';
+import type { MinimumReleaseAgeBehaviour } from '../../../../config/types';
 import { logger } from '../../../../logger';
 import type { Release } from '../../../../modules/datasource';
 import { postprocessRelease } from '../../../../modules/datasource/postprocess-release';
@@ -42,6 +43,14 @@ export async function filterInternalChecks(
     // Don't care if minimumReleaseAge or minimumConfidence are unmet
     release = sortedReleases.pop();
   } else {
+    const candidateVersionsWithoutReleaseTimestamp: Record<
+      MinimumReleaseAgeBehaviour,
+      string[]
+    > = {
+      'timestamp-required': [],
+      'timestamp-optional': [],
+    };
+
     // iterate through releases from highest to lowest, looking for the first which will pass checks if present
     for (let candidateRelease of sortedReleases.reverse()) {
       // merge the release data into dependency config
@@ -73,21 +82,44 @@ export async function filterInternalChecks(
       // Now check for a minimumReleaseAge config
       const { minimumConfidence, minimumReleaseAge, updateType } =
         releaseConfig;
-      if (
-        is.nonEmptyString(minimumReleaseAge) &&
-        candidateRelease.releaseTimestamp
-      ) {
-        if (
-          getElapsedMs(candidateRelease.releaseTimestamp) <
-          coerceNumber(toMs(minimumReleaseAge), 0)
+      if (is.nonEmptyString(minimumReleaseAge)) {
+        const minimumReleaseAgeBehaviour =
+          releaseConfig.minimumReleaseAgeBehaviour;
+
+        // if there is a releaseTimestamp, regardless of `minimumReleaseAgeBehaviour`, we should process it
+        if (candidateRelease.releaseTimestamp) {
+          // we should skip this if we have a timestamp that isn't passing checks:
+          if (
+            getElapsedMs(candidateRelease.releaseTimestamp) <
+            coerceNumber(toMs(minimumReleaseAge), 0)
+          ) {
+            // Skip it if it doesn't pass checks
+            logger.trace(
+              { depName, check: 'minimumReleaseAge' },
+              `Release ${candidateRelease.version} is pending status checks`,
+            );
+            pendingReleases.unshift(candidateRelease);
+            continue;
+          }
+        } // or if there is no timestamp, and we're running in `minimumReleaseAgeBehaviour=timestamp-required`
+        else if (
+          is.nullOrUndefined(candidateRelease.releaseTimestamp) &&
+          minimumReleaseAgeBehaviour === 'timestamp-required'
         ) {
-          // Skip it if it doesn't pass checks
-          logger.trace(
-            { depName, check: 'minimumReleaseAge' },
-            `Release ${candidateRelease.version} is pending status checks`,
-          );
+          // Skip it, as we require a timestamp
+          candidateVersionsWithoutReleaseTimestamp[
+            minimumReleaseAgeBehaviour
+          ].push(candidateRelease.version);
           pendingReleases.unshift(candidateRelease);
           continue;
+        } // if there is no timestamp, and we're running in `optional` mode, we can allow it
+        else if (
+          is.nullOrUndefined(candidateRelease.releaseTimestamp) &&
+          minimumReleaseAgeBehaviour === 'timestamp-optional'
+        ) {
+          candidateVersionsWithoutReleaseTimestamp[
+            minimumReleaseAgeBehaviour
+          ].push(candidateRelease.version);
         }
       }
 
@@ -115,6 +147,31 @@ export async function filterInternalChecks(
       release = candidateRelease;
       break;
     }
+
+    if (candidateVersionsWithoutReleaseTimestamp['timestamp-required'].length) {
+      logger.debug(
+        {
+          depName,
+          versions:
+            candidateVersionsWithoutReleaseTimestamp['timestamp-required'],
+          check: 'minimumReleaseAge',
+        },
+        `Marking ${candidateVersionsWithoutReleaseTimestamp['timestamp-required'].length} release(s) as pending, as they not have a releaseTimestamp and we're running with minimumReleaseAgeBehaviour=require-timestamp`,
+      );
+    }
+
+    if (candidateVersionsWithoutReleaseTimestamp['timestamp-optional'].length) {
+      logger.warn(
+        {
+          depName,
+          versions:
+            candidateVersionsWithoutReleaseTimestamp['timestamp-optional'],
+          check: 'minimumReleaseAge',
+        },
+        `${candidateVersionsWithoutReleaseTimestamp['timestamp-optional'].length} release(s) did not have a releaseTimestamp, but as we're running with minimumReleaseAgeBehaviour=optional-timestamp, proceeding`,
+      );
+    }
+
     if (!release) {
       if (pendingReleases.length) {
         // If all releases were pending then just take the highest
--- a/lib/workers/repository/update/branch/index.spec.ts
+++ b/lib/workers/repository/update/branch/index.spec.ts
@@ -238,19 +238,40 @@ describe('workers/repository/update/branch/index', () => {
       });
     });
 
-    // TODO: will be fixed in https://github.com/renovatebot/renovate/discussions/38290 / https://github.com/renovatebot/renovate/discussions/38348
-    it('does not skip branch if release is missing releaseTimestamp with minimumReleaseAge set', async () => {
-      schedule.isScheduledNow.mockReturnValueOnce(true);
-      config.prCreation = 'not-pending';
-      config.upgrades = partial<BranchUpgradeConfig>([
-        {
-          // no releaseTimestamp
-          minimumReleaseAge: '100 days',
-        },
-      ]);
-      scm.isBranchModified.mockResolvedValueOnce(false);
-      await branchWorker.processBranch(config);
-      expect(reuse.shouldReuseExistingBranch).toHaveBeenCalled();
+    describe('if release is missing releaseTimestamp with minimumReleaseAge set', () => {
+      it('skips branch if minimumReleaseAgeBehaviour=timestamp-required', async () => {
+        schedule.isScheduledNow.mockReturnValueOnce(true);
+        config.prCreation = 'not-pending';
+        config.upgrades = partial<BranchUpgradeConfig>([
+          {
+            releaseTimestamp: undefined,
+            minimumReleaseAge: '100 days',
+            minimumReleaseAgeBehaviour: 'timestamp-required',
+          },
+        ]);
+
+        const res = await branchWorker.processBranch(config);
+        expect(res).toEqual({
+          branchExists: false,
+          prNo: undefined,
+          result: 'pending',
+        });
+      });
+
+      it('does not skip branch if minimumReleaseAgeBehaviour=timestamp-optional', async () => {
+        schedule.isScheduledNow.mockReturnValueOnce(true);
+        config.prCreation = 'not-pending';
+        config.upgrades = partial<BranchUpgradeConfig>([
+          {
+            releaseTimestamp: undefined,
+            minimumReleaseAge: '100 days',
+            minimumReleaseAgeBehaviour: 'timestamp-optional',
+          },
+        ]);
+        scm.isBranchModified.mockResolvedValueOnce(false);
+        await branchWorker.processBranch(config);
+        expect(reuse.shouldReuseExistingBranch).toHaveBeenCalled();
+      });
     });
 
     it('skips branch if minimumConfidence not met', async () => {
--- a/lib/workers/repository/update/branch/index.ts
+++ b/lib/workers/repository/update/branch/index.ts
@@ -1,7 +1,10 @@
 import is from '@sindresorhus/is';
 import { DateTime } from 'luxon';
 import { GlobalConfig } from '../../../../config/global';
-import type { RenovateConfig } from '../../../../config/types';
+import {
+  type MinimumReleaseAgeBehaviour,
+  type RenovateConfig,
+} from '../../../../config/types';
 import {
   CONFIG_VALIDATION,
   MANAGER_LOCKFILE_ERROR,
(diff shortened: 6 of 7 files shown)


jamietanna:
Still needs `lib/workers/repository/update/branch/index.ts` updating

lib/workers/repository/update/branch/index.ts
```diff
@@ -381,22 +385,65 @@ export async function processBranch(
...
+                  depName: upgrade.depName,
+                  minimumReleaseAge: upgrade.minimumReleaseAge,
+                },
+                `Update does not have releaseTimestamp, and as we're running with minimumReleaseAgeTimestamp=required, this release will be marked as pending status checks`,
+              );
+              config.stabilityStatus = 'yellow';
+              continue;
+            } else {
+              // TODO
+              // config.warnings?.push({
```
jamietanna: Is using `warnings` the right thing here? 

Or via https://github.com/renovatebot/renovate/discussions/38348#discussion-8974527 could we add something to the PR body?

lib/workers/repository/process/lookup/filter-checks.ts
```diff
@@ -67,22 +71,73 @@ export async function filterInternalChecks(
...
+          minimumReleaseAgeTimestamp === 'required' &&
+          is.nullOrUndefined(candidateRelease.releaseTimestamp)
         ) {
-          // Skip it if it doesn't pass checks
+          // Skip it, as we require a timestamp
+          console.log(
+            { depName, check: 'minimumReleaseAge' },
+            `Release ${candidateRelease.version} did not have a releaseTimestamp, and as we're running with minimumReleaseAgeTimestamp=required, this release will be marked as pending status checks`,
+          );
           logger.trace(
```
jamietanna: I wonder if this should be bumped up a log level?

lib/workers/repository/update/branch/index.ts
jamietanna: @viceice @RahulGautamSingh any thoughts, here?

lib/config/options/index.ts
```diff
@@ -1966,6 +1966,14 @@ const options: RenovateOptions[] = [
     type: 'string',
     default: null,
   },
+  {
+    name: 'minimumReleaseAgeTimestamp',
+    description:
+      'When set in conjunction with `minimumReleaseAge`, controls whether the releaseTimestamp from a package lookup is needed to enforce the `minimumReleaseAge`. When set to `required`, this version is not treated stable unless there is releaseTimestamp, and that releaseTimestamp is past the `minimumReleaseAge`. When set to`optional`, Renovate will treat a release without a releaseTimestamp as stable',
+    type: 'string',
+    default: 'required',
```
viceice: we should set it as optional in first version to not break yet and switch it in next major which we should plan very soon.

lib/workers/repository/process/lookup/filter-checks.ts
viceice: maybe just cound them up and log once to not flod the logs with too many version loggings

lib/workers/repository/update/branch/index.spec.ts
```diff
@@ -238,19 +238,40 @@ describe('workers/repository/update/branch/index', () => {
...
-      config.upgrades = partial<BranchUpgradeConfig>([
-        {
-          // no releaseTimestamp
-          minimumReleaseAge: '100 days',
-        },
-      ]);
-      scm.isBranchModified.mockResolvedValueOnce(false);
-      await branchWorker.processBranch(config);
-      expect(reuse.shouldReuseExistingBranch).toHaveBeenCalled();
+    describe('if release is missing releaseTimestamp with minimumReleaseAge set', async () => {
```
viceice:
```suggestion
    describe('if release is missing releaseTimestamp with minimumReleaseAge set', () => {
```

lib/workers/repository/update/branch/index.ts
viceice: a warning seems good to me. not sure if `config.warnings` is added to the pr 🤷‍♂️

lib/config/options/index.ts
jamietanna: Very fair shout - I think I'd considered it, but this makes sense.

I'll make it `required` in `config:best-practices`, and `optional` here 👍 

Edit: captured https://github.com/renovatebot/renovate/issues/38457 to switch it over

lib/workers/repository/update/branch/index.ts
```diff
@@ -371,34 +374,69 @@ export async function processBranch(
...
+            } else {
+              // if there is no timestamp, and we're running in `optional` mode, we can allow it, but make sure to warn the user
+              depNamesWithoutReleaseTimestamp.optional.push(upgrade.depName!)
+            }
           }
         }
+
+        if (depNamesWithoutReleaseTimestamp.required) {
+          logger.debug(
+            { depNames: depNamesWithoutReleaseTimestamp.required, },
```
jamietanna: Happy with `depNames` as a log key here?

lib/workers/repository/update/branch/index.ts
```diff
@@ -371,34 +374,69 @@ export async function processBranch(
...
+            `${depNamesWithoutReleaseTimestamp.required.length} upgrades did not have a releaseTimestamp, and as we're running with minimumReleaseAgeTimestamp=required, these upgrades(s) will be marked as pending status checks`,
+          );
+        } else if (depNamesWithoutReleaseTimestamp.optional) {
+          logger.warn(
+            { depNames: depNamesWithoutReleaseTimestamp.optional, },
+            `${depNamesWithoutReleaseTimestamp.optional.length} upgrades did not have a releaseTimestamp, but as we're running with minimumReleaseAgeTimestamp=optional, proceeding`,
+          );
+        }
+
```
jamietanna: Should we find the unique `depName`s here, or are we happy that we may be logging the same `depName` multiple times?

lib/workers/repository/process/lookup/filter-checks.ts
jamietanna: I've amended this to log a group of these together, mind taking a look?

lib/workers/repository/update/branch/index.ts
```diff
@@ -371,34 +375,72 @@ export async function processBranch(
...
+          logger.debug(
+            { depNames: depNamesWithoutReleaseTimestamp.required },
+            `${depNamesWithoutReleaseTimestamp.required.length} upgrades did not have a releaseTimestamp, and as we're running with minimumReleaseAgeTimestamp=required, these upgrades(s) will be marked as pending status checks`,
+          );
+        } else if (depNamesWithoutReleaseTimestamp.optional) {
+          logger.warn(
+            { depNames: depNamesWithoutReleaseTimestamp.optional },
+            `${depNamesWithoutReleaseTimestamp.optional.length} upgrades did not have a releaseTimestamp, but as we're running with minimumReleaseAgeTimestamp=optional, proceeding`,
+          );
+        }
```
jamietanna: These are firing incorrectly - should only fire if there are elements in the array

lib/workers/repository/process/lookup/filter-checks.ts
```diff
@@ -64,27 +66,82 @@ export async function filterInternalChecks(
...
+      } else if (candidateVersionsWithoutReleaseTimestamp.optional) {
+        logger.warn(
+          {
+            depName,
+            versions: candidateVersionsWithoutReleaseTimestamp.optional,
+            check: 'minimumReleaseAge',
+          },
+          `${candidateVersionsWithoutReleaseTimestamp.optional.length} release(s) did not have a releaseTimestamp, but as we're running with minimumReleaseAgeTimestamp=optional, proceeding`,
+        );
+      }
```
jamietanna: These are firing incorrectly - should only fire if there are elements in the array

lib/workers/repository/process/lookup/filter-checks.ts
```diff
@@ -64,27 +66,82 @@ export async function filterInternalChecks(
...
+
       // Now check for a minimumReleaseAge config
       const { minimumConfidence, minimumReleaseAge, updateType } =
         releaseConfig;
-      if (
-        is.nonEmptyString(minimumReleaseAge) &&
-        candidateRelease.releaseTimestamp
-      ) {
+      if (is.nonEmptyString(minimumReleaseAge)) {
+        let minimumReleaseAgeTimestamp: MinimumReleaseAgeTimestamp = 'optional';
```
jamietanna: What's the best way of getting this to stay in sync with the config value's `default`?

lib/workers/repository/update/branch/index.ts
```diff
@@ -371,34 +375,72 @@ export async function processBranch(
...
-                timeElapsed,
-                minimumReleaseAge: upgrade.minimumReleaseAge,
-              },
-              'Update has not passed minimum release age',
-            );
-            config.stabilityStatus = 'yellow';
-            continue;
+        if (is.nonEmptyString(upgrade.minimumReleaseAge)) {
+          let minimumReleaseAgeTimestamp: MinimumReleaseAgeTimestamp =
+            'optional';
```
jamietanna: What's the best way of getting this to stay in sync with the config value's `default`?

jamietanna:
Testing with https://github.com/jamietanna/renovate-testing-minimumReleaseAge-38324-38290-38348, with the following diff:

```diff
diff --git renovate.json renovate.json
index f42a00d..e56c288 100644
--- renovate.json
+++ renovate.json
@@ -1,5 +1,9 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "separateMajorMinor": true,
+  "separateMultipleMajor": true,
+  "separateMultipleMinor": true,
+  "separateMinorPatch": true,
   "minimumReleaseAge": "1 day",
   "customManagers": [
     {
diff --git server.js server.js
index 36a5f1a..c19c451 100644
--- server.js
+++ server.js
@@ -92,6 +92,18 @@ const server = http.createServer((req, res) => {
     v010,
     v011,
     v012,
+    {
+      version: '0.1.3',
+    },
+    {
+      version: '2.1.4',
+    },
+    {
+      version: '3.1.5',
+    },
+    {
+      version: '4.1.6',
+    },
   ].filter(v => v !== undefined);
```

This results in the following logging:

```
DEBUG: Using queue: host=localhost:12345, concurrency=16 (repository=local)
 WARN: 1 release(s) did not have a releaseTimestamp, but as we're running with minimumReleaseAgeTimestamp=optional, proceeding (repository=local)
       "depName": "some-package",
       "versions": ["0.1.3"],
       "check": "minimumReleaseAge"
 WARN: 1 release(s) did not have a releaseTimestamp, but as we're running with minimumReleaseAgeTimestamp=optional, proceeding (repository=local)
       "depName": "some-package",
       "versions": ["2.1.4"],
       "check": "minimumReleaseAge"
 WARN: 1 release(s) did not have a releaseTimestamp, but as we're running with minimumReleaseAgeTimestamp=optional, proceeding (repository=local)
       "depName": "some-package",
       "versions": ["3.1.5"],
       "check": "minimumReleaseAge"
 WARN: 1 release(s) did not have a releaseTimestamp, but as we're running with minimumReleaseAgeTimestamp=optional, proceeding (repository=local)
       "depName": "some-package",
       "versions": ["4.1.6"],
       "check": "minimumReleaseAge"
DEBUG: PackageFiles.add() - Package file saved for base branch (repository=local)
DEBUG: Package releases lookups complete (repository=local)
DEBUG: No releaseTimestamp for some-package update to 0.1.3 (repository=local)
DEBUG: Repository libYears (repository=local)
       "libYears": {"managers": {"regex": 0}, "total": 0},
       "dependencyStatus": {"outdated": 1, "total": 1}
DEBUG: branchifyUpgrades (repository=local)
DEBUG: detectSemanticCommits() (repository=local)
DEBUG: getCommitMessages (repository=local)
DEBUG: semanticCommits: detected "unknown" (repository=local)
DEBUG: semanticCommits: disabled (repository=local)
DEBUG: 4 flattened updates found: some-package, some-package, some-package, some-package (repository=local)
DEBUG: Returning 4 branch(es) (repository=local)
DEBUG: config.repoIsOnboarded=true (repository=local)
DEBUG: packageFiles with updates (repository=local)
       "config": {
         "regex": [
           {
             "deps": [
               {
                 "depName": "some-package",
                 "currentValue": "v0.1.0",
                 "datasource": "custom.localExample",
                 "versioning": "semver-coerced",
                 "replaceString": "v0.1.0",
                 "updates": [
                   {
                     "bucket": "patch",
                     "newVersion": "0.1.3",
                     "newValue": "0.1.3",
                     "newMajor": 0,
                     "newMinor": 1,
                     "newPatch": 3,
                     "updateType": "patch",
                     "isBreaking": true,
                     "branchName": "renovate/some-package-0.1.x"
                   },
                   {
                     "bucket": "v2",
                     "newVersion": "2.1.4",
                     "newValue": "2.1.4",
                     "newMajor": 2,
                     "newMinor": 1,
                     "newPatch": 4,
                     "updateType": "major",
                     "isBreaking": true,
                     "branchName": "renovate/some-package-2.x"
                   },
                   {
                     "bucket": "v3",
                     "newVersion": "3.1.5",
                     "newValue": "3.1.5",
                     "newMajor": 3,
                     "newMinor": 1,
                     "newPatch": 5,
                     "updateType": "major",
                     "isBreaking": true,
                     "branchName": "renovate/some-package-3.x"
                   },
                   {
                     "bucket": "v4",
                     "newVersion": "4.1.6",
                     "newValue": "4.1.6",
                     "newMajor": 4,
                     "newMinor": 1,
                     "newPatch": 6,
                     "updateType": "major",
                     "isBreaking": true,
                     "branchName": "renovate/some-package-4.x"
                   }
                 ],
                 "packageName": "some-package",
                 "warnings": [],
                 "currentVersion": "0.1.0",
                 "currentVersionTimestamp": "2025-09-28T10:10:22.532Z",
                 "currentVersionAgeInDays": 10,
                 "isSingleVersion": true,
                 "fixedVersion": "v0.1.0"
               }
             ],
             "matchStrings": ["(?<currentValue>\\S+)"],
             "depNameTemplate": "some-package",
             "datasourceTemplate": "custom.localExample",
             "versioningTemplate": "semver-coerced",
             "packageFile": "packagefile.txt"
           }
         ]
       }
DEBUG: detectSemanticCommits() (repository=local)
DEBUG: semanticCommits: returning "disabled" from cache (repository=local)
DEBUG: repository problems (repository=local)
       "repoProblems": [
         "WARN: 1 release(s) did not have a releaseTimestamp, but as we're running with minimumReleaseAgeTimestamp=optional, proceeding"
       ]
```

Need to confirm why these aren't grouped

Edit: It's because this is called on i.e. [here](https://github.com/renovatebot/renovate/blob/a4e275b1cd2ff8e06cff557897450cef8c6f5808/lib/workers/repository/process/lookup/index.ts#L484-L494) where this is inside a loop of various buckets of updates.

So we're unfortunately going to get duplicates of this log, but some may be over a group of dependencies 🤞

lib/workers/repository/process/lookup/filter-checks.ts
```diff
@@ -115,6 +152,29 @@ export async function filterInternalChecks(
...
+    }
+
+    if (candidateVersionsWithoutReleaseTimestamp.optional.length) {
+      logger.warn(
+        {
+          depName,
+          versions: candidateVersionsWithoutReleaseTimestamp.optional,
+          check: 'minimumReleaseAge',
+        },
+        `${candidateVersionsWithoutReleaseTimestamp.optional.length} release(s) did not have a releaseTimestamp, but as we're running with minimumReleaseAgeTimestamp=optional, proceeding`,
```
jamietanna: do we have a helper to auto-pluralise a word?

lib/config/types.ts
```diff
@@ -219,6 +219,13 @@ export type RenovateRepository =
 
 export type UseBaseBranchConfigType = 'merge' | 'none';
 export type ConstraintsFilter = 'strict' | 'none';
+export type MinimumReleaseAgeTimestamp = 'required' | 'optional';
```
jamietanna: Should these be the other way around, given `optional` is the current default?

docs/usage/configuration-options.md
```diff
@@ -2707,7 +2707,7 @@ Example:
 
 This feature used to be called `stabilityDays`.
 
-If `minimumReleaseAge` is set to a time duration _and_ the update has a release timestamp header, then Renovate will check if the set duration has passed.
+If `minimumReleaseAge` is set to a time duration _and_ the update has a release timestamp header, then Renovate will check if the set duration has passed. This behaviour can be changed using `minimumReleaseAgeTimestamp`.
```
RahulGautamSingh: Link the option here

lib/config/options/index.ts
```diff
@@ -1966,6 +1966,14 @@ const options: RenovateOptions[] = [
     type: 'string',
     default: null,
   },
+  {
+    name: 'minimumReleaseAgeTimestamp',
+    description:
+      'When set in conjunction with `minimumReleaseAge`, controls whether the releaseTimestamp from a package lookup is needed to enforce the `minimumReleaseAge`. When set to `required`, this version is not treated stable unless there is releaseTimestamp, and that releaseTimestamp is past the `minimumReleaseAge`. When set to `optional`, Renovate will treat a release without a releaseTimestamp as stable',
```
RahulGautamSingh: We can keep this short. Something like, 

"Used to mark release without releaseTimestamp as non-optional"

This also gives me an idea to improve our json schema by adding a new field called, requires or dependsOn which can be used to link options to each other if they need to be used together. wdyt??

lib/config/presets/internal/config.ts
```diff
@@ -14,6 +14,7 @@ export const presets: Record<string, Preset> = {
       ':pinDevDependencies',
       'abandonments:recommended',
     ],
+    minimumReleaseAgeTimestamp: 'required',
```
RahulGautamSingh: This makes this PR a breaking change, right? As many users would still be expecting PRs if no release timestamp is available

lib/workers/repository/process/lookup/filter-checks.ts
RahulGautamSingh: I think default value will be passed itself no need to initalize

lib/workers/repository/process/lookup/filter-checks.ts
```diff
@@ -73,21 +83,48 @@ export async function filterInternalChecks(
...
-          coerceNumber(toMs(minimumReleaseAge), 0)
+          isMinimumReleaseAgeTimestamp(releaseConfig.minimumReleaseAgeTimestamp)
         ) {
-          // Skip it if it doesn't pass checks
-          logger.trace(
-            { depName, check: 'minimumReleaseAge' },
-            `Release ${candidateRelease.version} is pending status checks`,
-          );
+          minimumReleaseAgeTimestamp = releaseConfig.minimumReleaseAgeTimestamp;
+        }
```
RahulGautamSingh: not needed,

lib/workers/repository/process/lookup/filter-checks.ts
```diff
@@ -115,6 +152,29 @@ export async function filterInternalChecks(
...
     }
+
+    if (candidateVersionsWithoutReleaseTimestamp.required.length) {
+      logger.debug(
+        {
+          depName,
+          versions: candidateVersionsWithoutReleaseTimestamp.required,
+          check: 'minimumReleaseAge',
+        },
+        `${candidateVersionsWithoutReleaseTimestamp.required.length} release(s) did not have a releaseTimestamp, and as we're running with minimumReleaseAgeTimestamp=required, these release(s) will be marked as pending status checks`,
```
RahulGautamSingh:
```suggestion
        `Marking the following ${candidateVersionsWithoutReleaseTimestamp.required.length} release(s) pending as they did not have a releaseTimestamp, and minimumReleaseAgeTimestamp=required`,
```

How's this?

RahulGautamSingh (changes requested):

lib/config/presets/internal/config.ts
jamietanna: I think for `config:best-practices`, we can change things a little bit more without adhering to SemVer of "breaking"?

I would expect users of `config:best-practices` to be aware that there may be changes introduced that change the way things behave

cc @secustor @viceice in case I'm misrepresenting things

lib/config/options/index.ts
jamietanna: Out of interest, why do we want to keep this short?

jamietanna: > This also gives me an idea to improve our json schema by adding a new field called, requires or dependsOn which can be used to link options to each other if they need to be used together. wdyt??

If we're using a version of JSON Schema that supports indicating this, then yep, that sounds like a plan

lib/workers/repository/process/lookup/filter-checks.ts
jamietanna: Via https://github.com/renovatebot/renovate/pull/38363/files#r2420592017 the default is `required`

lib/workers/repository/process/lookup/filter-checks.ts
jamietanna: What's not needed?

lib/workers/repository/update/branch/index.ts
```diff
@@ -371,34 +375,60 @@ export async function processBranch(
...
+              depNamesWithoutReleaseTimestamp.required.push(upgrade.depName!);
+              config.stabilityStatus = 'yellow';
+              continue;
+            } else {
+              // if there is no timestamp, and we're running in `optional` mode, we can allow it, but make sure to warn the user
+              depNamesWithoutReleaseTimestamp.optional.push(upgrade.depName!);
+            }
           }
         }
+
```
jamietanna:
```suggestion
```

lib/config/options/index.ts
viceice: > Out of interest, why do we want to keep this short?

This is copied before the other text into the docs. So should not duplicate it.
It's also copied to the json schema.

lib/config/presets/internal/config.ts
viceice: I'm also in favor to not do that in a minor release and plan a major release soon.

lib/workers/repository/process/lookup/filter-checks.ts
viceice: no

lib/workers/repository/update/branch/index.ts
viceice: usually it automatically passes the default value from the config. 🤔

viceice (commented):
see comments

lib/config/presets/internal/config.ts
jamietanna: OK - so even for `config:best-practices`, we shouldn't default `minimumReleaseAgeTimestamp=required`?

Would you then say that https://github.com/renovatebot/renovate/issues/38457 would be a future breaking change? Or enabled at the same time

docs/usage/configuration-options.md
```diff
@@ -2775,6 +2775,16 @@ This pending check prevents the branch going green to automerge before the time
 
 <!-- markdownlint-enable MD001 -->
 
+## minimumReleaseAgeTimestamp
```
secustor: I find the naming confusing, based on the naming I would expect this to be an option to set the timestamp used for `minimumReleaseAge`. 

I would suggest to use `minimumReleaseAgeBehaviour`, which is open enough for future additional options

lib/config/presets/internal/config.ts
secustor: If we would default to `optional` for now a feature release would be sufficient as we give only the option to use it. 

The linked on the other hand would be breaking, as we introduce behavioral changes outside of bug fixes. Also a major release gives the platform ( release notes ) on the docs page to explain the reasoning further.

lib/config/types.ts
secustor: AFAIK we don't have a strict rule which defines the order, though it make sense to have the default first

lib/workers/repository/process/lookup/filter-checks.ts
secustor: You don't need to set this. 

https://github.com/renovatebot/renovate/pull/38363/files#diff-0053f214ec552bf58e21e52b0af7fcfe1699c7be2989cee50c52d44c7046ed2fR1974

Will ensure that the value is `optional` if the user does not supply a value

lib/workers/repository/update/branch/index.ts
secustor: See this thread: https://github.com/renovatebot/renovate/pull/38363/files#r2413325481

docs/usage/configuration-options.md
jamietanna: Makes sense - so i.e.

```
minimumReleaseAgeBehaviour=require-timestamp
# or
minimumReleaseAgeBehaviour=optional-timestamp
```

?

lib/config/presets/internal/config.ts
jamietanna: > If we would default to `optional` for now a feature release would be sufficient as we give only the option to use it.

To confirm, do you mean that if the default (for non `config:best-practices` users) would be `optional`, it'd be fine in a feature release to set it to `required` for `config:best-practices`?

lib/workers/repository/process/lookup/filter-checks.ts
jamietanna: If I don't set this, then I get:

```
lib/workers/repository/process/lookup/filter-checks.ts:112:11 - error TS2454: Variable 'minimumReleaseAgeTimestamp' is used before being assigned.

112           minimumReleaseAgeTimestamp === 'required'
              ~~~~~~~~~~~~~~~~~~~~~~~~~~

lib/workers/repository/process/lookup/filter-checks.ts:123:11 - error TS2454: Variable 'minimumReleaseAgeTimestamp' is used before being assigned.

123           minimumReleaseAgeTimestamp === 'optional'
              ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

jamietanna:
Once we've finalised conversations + are ready to approve, I'll rebase + correct the history + commit message, as we've changed some of the planned implementation

lib/config/presets/internal/config.ts
jamietanna: Also, if we're saying we can't change this behaviour in a feature release, could https://github.com/renovatebot/renovate/pull/37967 ever get merged, unless it's in a major version?

docs/usage/configuration-options.md
viceice: sounds good

viceice (commented):
lgtm beside teh renaming

lib/workers/repository/process/lookup/filter-checks.ts
RahulGautamSingh: 1. no need to set the value of `minimumReleaseAgeTimestamp` here as `required/optional`. If we have a defauilt value set in `options/index.ts` it will be passed till here in the `releaseConfig`.

Just do `minimumReleaseAgeTimestamp = releaseConfig.minimumReleaseAgeTimestamp`

2. Since the value is passed/present already we won't be needing the `isMinimumReleaseAgeTimestamp` fn

lib/config/presets/internal/config.ts
jamietanna: Discussed offline - the key concern here is that it's quite a "high profile" change, so we _would_ want to make sure that we're calling this out as part of the next breaking change

jamietanna:
Currently validating against a real repository - will move to Draft until finished

lib/config/types.ts
```diff
@@ -326,6 +329,8 @@ export interface RenovateConfig
   additionalBranchPrefix?: string;
   sharedVariableName?: string;
   minimumGroupSize?: number;
+
+  minimumReleaseAgeBehaviour?: MinimumReleaseAgeBehaviour;
```
jamietanna, line 333: Note that we've not done this (yet) for `minimumReleaseAge` - I'll follow up

lib/workers/repository/update/branch/index.ts
```diff
@@ -371,32 +374,56 @@ export async function processBranch(
     if (
       config.upgrades.some(
         (upgrade) =>
-          (is.nonEmptyString(upgrade.minimumReleaseAge) &&
-            is.nonEmptyString(upgrade.releaseTimestamp)) ||
```
jamietanna, line 375: This has moved to inside the loop, as we care whether `releaseTimestamp` is there or not

lib/workers/repository/process/lookup/filter-checks.ts
```diff
@@ -115,6 +147,31 @@ export async function filterInternalChecks(
...
+        {
+          depName,
+          versions:
+            candidateVersionsWithoutReleaseTimestamp['timestamp-required'],
+          check: 'minimumReleaseAge',
+        },
+        `Marking ${candidateVersionsWithoutReleaseTimestamp['timestamp-required'].length} release(s) as pending, as they not have a releaseTimestamp and we're running with minimumReleaseAgeBehaviour=require-timestamp`,
+      );
+    }
+
```
jamietanna, line 162: To be consistent with the other one?
```suggestion
```

viceice (approved):

PR #38396, opened 2025-10-06
fix(copier): Use correct value for vcs-ref

kabooboo:
## Changes

Today, the `copier` manager uses a coalesce of `newVersion` then `newValue` for the next version to which update. This behaviour breaks for repositories that version with tags that are not the exact version, e.g.: `foobar-v1.2.3` instead of `1.2.3`.

This happens because:
1. Renovate detects the version
2. Renovate extracts the "useful" part of the version with `extractVersion` or `versionCompatibility`
3. Renovate passes newVersion and newValue to the manager
4. Since we coalesce first to newVersion, we're unable to pass the full version string to the copier command.

If we're going the other way around and simply use  `newValue`, the case for `foobar-v1.2.3` is covered.

## Context

## AI assistance disclosure

The public repository: <URL>


RahulGautamSingh (changes requested):
Thanks for the PR!

LGTM
Can you please also add a new test for this specific case?

kabooboo:
@RahulGautamSingh it should be all good, the test was added in commit 765d3bd4bbac2f06755b7c23dd3f85ee810ff419

RahulGautamSingh (approved):

viceice (approved):

PR #38423, opened 2025-10-07
feat(presets): add schedule presets for weekdays except Friday

luciomartinez:
## Changes

Add schedule presets to include weekdays except Fridays.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/config/presets/internal/schedule.ts
+++ b/lib/config/presets/internal/schedule.ts
@@ -39,7 +39,7 @@ export const presets: Record<string, Preset> = {
       'Schedule automerge during typical office hours on weekdays (i.e., 8 AM - 6 PM).',
   },
   automergeOfficeHoursExceptFriday: {
-    automergeSchedule: officeHours,
+    automergeSchedule: officeHoursExceptFriday,
     description:
       'Schedule automerge during typical office hours on weekdays except friday (i.e., 8 AM - 6 PM).',
   },
--- a/lib/config/presets/internal/schedule.ts
+++ b/lib/config/presets/internal/schedule.ts
@@ -7,8 +7,10 @@ const earlyMondays = ['* 0-3 * * 1'];
 const monthly = ['* 0-3 1 * *'];
 const nonOfficeHours = ['* 0-4,22-23 * * 1-5', '* * * * 0,6'];
 const officeHours = ['* 8-17 * * 1-5'];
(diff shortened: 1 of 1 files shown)


lib/config/presets/internal/schedule.ts
```diff
@@ -36,6 +38,11 @@ export const presets: Record<string, Preset> = {
     description:
       'Schedule automerge during typical office hours on weekdays (i.e., 8 AM - 6 PM).',
   },
+  automergeOfficeHoursExceptFriday: {
+    automergeSchedule: officeHours,
```
astellingwerf:
```suggestion
    automergeSchedule: officeHoursExceptFriday,
```

astellingwerf (changes requested):

lib/config/presets/internal/schedule.ts
```diff
@@ -90,6 +97,10 @@ export const presets: Record<string, Preset> = {
     description: 'Schedule for weekdays.',
     schedule: weekdays,
   },
+  weekdays: {
```
astellingwerf:
```suggestion
  weekdaysExceptFriday: {
```

astellingwerf (changes requested):
These changes can do with a little more tlc.

jamietanna:
I'll let @astellingwerf re-review before we merge (or if you're happy me doing final review, LMK)

Not sure if it's the intent here to do "no deploy Fridays" - I'm not _personally_ a fan of this, but makes sense that some folks do, and so I agree that adding this in for others to use without having to personally hand-craft makes sense

astellingwerf:
Hi @jamietanna, I can re-review for the functional and grammatical change, but I'm not sure I'm a fan of this change. Like, where does it stop? Should we have a "Weekdays except Mondays" preset too? Except Wednesdays, except Tuesdays on even days?

Renovate is powerful enough for users to define their own scheduling patterns to cover whatever business logic is appropriate for them, and it's well documented. To me, there's no obvious need to incorporate this particular preset in the product by default.

jamietanna:
A very fair point - looking at https://docs.renovatebot.com/presets-schedule/ I can see we don't have any similar ones here, so I'll agree that this is something we should be cautious in accepting, in case this "sets a precedent", or bloats our shared presets to maintain.

Thanks for raising this @luciomartinez - looking forward to hear if there are any other areas we can help improve!

PR #38424, opened 2025-10-07
feat!: Decrease office hours for schedule presets from 10 to 8 hours

luciomartinez:
## Changes

Office hours meaning is by its definition:

> Office hours are the times when an office is open for business. For example, office hours in Britain are usually between 9 o'clock and 5 o'clock from Monday to Friday.

However, the current configuration, covers from 8 AM to 6 PM. Hopefully, this is no ones office hours, as it'd entail someone working 10 hours per day. 🙂 

With this change, I propose to fix the configuration to go along with the general definition from 9 AM to 5 PM.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/config/presets/internal/schedule.ts
+++ b/lib/config/presets/internal/schedule.ts
@@ -6,7 +6,7 @@ const daily = ['* 0-3 * * *'];
 const earlyMondays = ['* 0-3 * * 1'];
 const monthly = ['* 0-3 1 * *'];
 const nonOfficeHours = ['* 0-4,22-23 * * 1-5', '* * * * 0,6'];
-const officeHours = ['* 8-16 * * 1-5'];
+const officeHours = ['* 9-16 * * 1-5'];
 const quarterly = ['* * 1 */3 *'];
 const weekdays = ['* * * * 1-5'];
 const weekends = ['* * * * 0,6'];
@@ -34,7 +34,7 @@ export const presets: Record<string, Preset> = {
   automergeOfficeHours: {
     automergeSchedule: officeHours,
     description:
-      'Schedule automerge during typical office hours on weekdays (i.e., 8 AM - 5 PM).',
+      'Schedule automerge during typical office hours on weekdays (i.e., 9 AM - 5 PM).',
   },
   automergeQuarterly: {
     automergeSchedule: quarterly,
@@ -78,7 +78,7 @@ export const presets: Record<string, Preset> = {
   },
   officeHours: {
     description:
-      'Schedule during typical office hours on weekdays (i.e., 8 AM - 5 PM).',
+      'Schedule during typical office hours on weekdays (i.e., 9 AM - 5 PM).',
     schedule: officeHours,
   },
   quarterly: {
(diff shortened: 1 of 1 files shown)


astellingwerf:
I'm not sure that it's Renovate's responsibility to prevent your app's deployment right as people rush to the door.

luciomartinez:
Thanks @astellingwerf for looking into it! 

I agree with your point of view, and that's not the purpose of this PR. 

Let me rephrase the description with a more assertive goal instead. Hope it makes more sense now 🙏

lib/config/presets/internal/schedule.ts
```diff
@@ -6,7 +6,7 @@ const daily = ['* 0-3 * * *'];
 const earlyMondays = ['* 0-3 * * 1'];
 const monthly = ['* 0-3 1 * *'];
 const nonOfficeHours = ['* 0-4,22-23 * * 1-5', '* * * * 0,6'];
-const officeHours = ['* 8-17 * * 1-5'];
+const officeHours = ['* 8-16 * * 1-5'];
```
luciomartinez:
```suggestion
const officeHours = ['* 9-16 * * 1-5'];
```

lib/config/presets/internal/schedule.ts
```diff
@@ -34,7 +34,7 @@ export const presets: Record<string, Preset> = {
   automergeOfficeHours: {
     automergeSchedule: officeHours,
     description:
-      'Schedule automerge during typical office hours on weekdays (i.e., 8 AM - 6 PM).',
+      'Schedule automerge during typical office hours on weekdays (i.e., 8 AM - 5 PM).',
```
luciomartinez:
```suggestion
      'Schedule automerge during typical office hours on weekdays (i.e., 9 AM - 5 PM).',
```

lib/config/presets/internal/schedule.ts
```diff
@@ -78,7 +78,7 @@ export const presets: Record<string, Preset> = {
   },
   officeHours: {
     description:
-      'Schedule during typical office hours on weekdays (i.e., 8 AM - 6 PM).',
+      'Schedule during typical office hours on weekdays (i.e., 8 AM - 5 PM).',
```
luciomartinez:
```suggestion
      'Schedule during typical office hours on weekdays (i.e., 9 AM - 5 PM).',
```

RahulGautamSingh:
Isn't on the user to merge the PR/review when they want. The bot running 1hr outside office hours doesn't seem like a problem.

I also think the preset has 10hrs so it might give the bot more time to run on the repo (especially for users using the Github app).

Also, everyone can use their own presets so this change seems unnecessary.

jamietanna:
If I'm honest, I'm not quite sure this is the right thing to be changing here

I _believe_ the reason we went for 0800-1800 was not because "10 hour working days are expected", but i.e. to allow for Renovate to get a PR raised + CI running, so that when someone gets to the office i.e. at 0900, they'd be able to action the PR

Given different companies, countries, and people have different "working days", I think trying to be perfectly prescriptive is a hard one - as it's a little bit subjective

Additionally, given this is a breaking change, I'm not sure if this is something we'd want to do (as I don't believe we have a

RahulGautamSingh:
We will leave it to the users

PR #38474, opened 2025-10-08
docs(aws-machine-image): clarify that currentImageName is updated by Renovate

josiblair:
## Description

The example regex pattern incorrectly captures `currentImageName` into the `currentDigest` group, causing Renovate to update this comment field. The `currentImageName` field should remain static (typically "unknown") and not be modified during updates.

Remove the named capture group `<currentDigest>` from the `currentImageName` pattern so the comment line is preserved as-is while only the AMI ID is updated.

## Changes

Updated the example regex pattern in the AWS Machine Image datasource documentation to remove the `<currentDigest>` named capture group from the `currentImageName` line. This prevents Renovate from updating the `# currentImageName=unknown` comment, which should remain static.

Before: `(.*currentImageName=(?<currentDigest>.*?)\n)?`
After: `(.*currentImageName=.*?\n)?`

## Context

## AI assistance disclosure

AI was used to help identify the issue in the regex pattern.

### DIFF
--- a/lib/modules/datasource/aws-machine-image/readme.md
+++ b/lib/modules/datasource/aws-machine-image/readme.md
@@ -116,6 +116,10 @@ my_ami2: ami-0083e9407e275acf2
 # amiFilter=[{"Name":"owner-id","Values":["602401143452"]},{"Name":"name","Values":["amazon-eks-node-1.20-*"]},{"profile":"test","region":"eu-central-1"}]
 # currentImageName=amazon-eks-node-1.20-v20240615
 ami = "ami-0083e9407e275acf2"
+
+# Without currentImageName comment (also works!)
+# amiFilter=[{"Name":"owner-id","Values":["602401143452"]},{"Name":"name","Values":["amazon-eks-node-1.19-*"]}]
+my_ami3: ami-0a1b2c3d4e5f6g7h8
 ```
 
 ```typescript
@@ -127,7 +131,7 @@ const myConfigObject = {
 };
 
 /**
- * Only AMI, no AMI name mentioned
+ * AMI with name tracked in comment
  * amiFilter=[{"Name":"owner-id","Values":["602401143452"]},{"Name":"name","Values":["amazon-eks-node-1.20-*"]}]
  * currentImageName=amazon-eks-node-1.20-v20240615
  */
@@ -148,5 +152,13 @@ resource "aws_instance" "web" {
     connection {
         user = "root"
     }
+}
+
+resource "aws_instance" "app" {
+    # Without currentImageName comment
+    # amiFilter=[{"Name":"owner-id","Values":["602401143452"]},{"Name":"name","Values":["amazon-eks-node-1.19-*"]}]
+    ami = "ami-0a1b2c3d4e5f6g7h8"
+
+    count = 1
 }
 ```
--- a/lib/modules/datasource/aws-machine-image/readme.md
+++ b/lib/modules/datasource/aws-machine-image/readme.md
@@ -64,7 +64,7 @@ module.exports = {
       customType: 'regex',
       managerFilePatterns: ['/.*/'],
       matchStrings: [
-        '.*amiFilter=(?<packageName>.*?)\n(.*currentImageName=.*?\n)?(.*\n)?.*?(?<depName>[a-zA-Z0-9-_:]*)[ ]*?[:|=][ ]*?["|\']?(?<currentValue>ami-[a-z0-9]{17})["|\']?.*',
+        '.*amiFilter=(?<packageName>.*?)\n(.*currentImageName=(?<currentDigest>.*?)\n)?(.*\n)?.*?(?<depName>[a-zA-Z0-9-_:]*)[ ]*?[:|=][ ]*?["|\']?(?<currentValue>ami-[a-z0-9]{17})["|\']?.*',
       ],
       datasourceTemplate: 'aws-machine-image',
       versioningTemplate: 'aws-machine-image',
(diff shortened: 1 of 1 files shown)


jamietanna:
Is the repo you tested it on public? (NP if not / not trivial to share, but wanted to confirm)

josiblair:
It's not public. It's on a self-hosted instance of GitLab. Here are some screenshots of before/after the regex change. 

Before: 
Screenshot 2025-10-08 at 9 43 15 AM

Change to regex:
Screenshot 2025-10-08 at 10 37 29 AM

After:
Screenshot 2025-10-08 at 10 38 43 AM

lib/modules/datasource/aws-machine-image/readme.md
```diff
@@ -84,7 +84,7 @@ Or as JSON:
         'managerFilePatterns': ['/.*/'],
         'matchStrings':
           [
-            ".*amiFilter=(?<packageName>.*?)\n(.*currentImageName=(?<currentDigest>.*?)\n)?(.*\n)?.*?(?<depName>[a-zA-Z0-9-_:]*)[ ]*?[:|=][ ]*?[\"|']?(?<currentValue>ami-[a-z0-9]{17})[\"|']?.*",
+            ".*amiFilter=(?<packageName>.*?)\n(.*currentImageName=.*?\n)?(.*\n)?.*?(?<depName>[a-zA-Z0-9-_:]*)[ ]*?[:|=][ ]*?["|\']?(?<currentValue>ami-[a-z0-9]{17})["|\']?.*",
```
jamietanna: why were these quotes changed? Was that to keep them in sync with the above regex?

(I'm going to follow-up with this PR and remove the duplication, as it doesn't provide much value)

jamietanna (commented):
So one thing this will lead to is the removal of the `currentDigest` - is that intended? I notice that in the "before" photo, we're updating the `currentImageName` to what looks like a useful new value?

josiblair:
Hmm, you're right to question this. I was interpreting `currentImageName=unknown` in the examples as meaning it should remain static, but I can see how updating it to the actual AMI name would be useful for tracking purposes.

However, I'm confused about the intended behavior because:
1. All the examples in the docs show `currentImageName=unknown`
2. There's no explanation of whether this field should be updated or remain static
3. The "unknown" value suggests it's a placeholder that shouldn't change

Could you clarify the intended behavior? Should `currentImageName` be:
- A static placeholder (always "unknown")
-OR-
- Updated to track the actual AMI name (as the current regex does)

If it's meant to be updated, perhaps I could update the docs to show examples with 'actual' AMI names instead of "unknown" to make this more clear? I'm not sure how to make it 'less confusing' 😅

lib/modules/datasource/aws-machine-image/readme.md
josiblair: Apologies - this quote change was unintentional - I was trying to keep both versions consistent but didn't realize the JSON version needs escaped quotes (\") while the JavaScript version uses raw quotes (").

I'll revert that change. The only intended change was removing the `currentDigest` capture group, but based on your feedback, I'm now questioning whether that's the right fix at all.

RahulGautamSingh (commented):
It is supposed to track the digest. But in the examples we haven't used any digest. So, if you could update one of them to digest that'd be great, given that digests for AMI images exist like for other manager packages

jamietanna:
OK, having tested this documentation, against https://github.com/JamieTanna-Mend-testing/docs-ami-datasource I can see that https://github.com/JamieTanna-Mend-testing/docs-ami-datasource/pull/2 would be generated.

Note that `currentImageName=amazon-eks-node-1.21-v20240703`, the _actual_ Amazon Machine Image name.

With this in mind, it's worth us correcting the docs as they currently are, removing the `=unknown`, and clarifying that it's an actual AMI name.

@josiblair let me know if you're happy using this PR to update it, or if you'd prefer us to do so - happy either way!

josiblair:
Done!
I've updated all the examples to show actual AMI names instead of `unknown` and added an explanatory note about how Renovate updates this field. The regex patterns are back to their original form (including the `currentDigest` capture).

Let me know if you'd like any other changes!

lib/modules/datasource/aws-machine-image/readme.md
```diff
@@ -93,46 +93,51 @@ Or as JSON:
 }
 ```
 
+**Note about `currentImageName`:**
+
+The optional `currentImageName` comment is automatically updated by Renovate to track the actual AMI name corresponding to the AMI ID. This provides human-readable context about which image version is being used. When Renovate finds a newer AMI, it will update both the AMI ID and the `currentImageName` comment.
```
RahulGautamSingh: One sentence per line

```suggestion
The optional `currentImageName` comment is automatically updated by Renovate to track the actual AMI name corresponding to the AMI ID.
This provides human-readable context about which image version is being used.
When Renovate finds a newer AMI, it will update both the AMI ID and the `currentImageName` comment.
```

[Via](https://github.com/renovatebot/renovate/blob/main/docs/development/style-guide.md#one-sentence-per-line)

RahulGautamSingh (commented):
Else, LGTM

jamietanna (commented):
One comment from Rahul, and the other change I'd like to see is an example _without_ the `currentImageName` - I see from the regex (and our improved docs) that it shouldn't be required, but it'd be good to have that confirmed, as part of the docs 🙏  

Otherwise LGTM, thanks for the back-and-forth!

josiblair:
added those examples. Let me know if there's anything else - thanks all!

jamietanna (approved):
Thank you!

PR #38566, opened 2025-10-13
chore(github): warn users without "Automatically delete head branches"

jamietanna:
## Changes

As noted in #38324, there could be a bug occur with branches not being
automagically deleted, where:

1. repository does not "Automatically delete head branches"
1. `minimumReleaseAge` is used to gate a release update, as the release
   is not passing that check
1. a group that includes that release is set up
1. a branch is created for all updates except the `minimumReleaseAge`
   (as expected)
1. the PR is merged
1. the branch is not deleted
1. next time Renovate runs, it sees the branch created - albeit out of
   date - so it treats this as permission to update the package gated by
   `minimumReleaseAge`

As a step towards fixing this long-term, we can introduce a log warning
to inform repo owners that they may be susceptible to issues.

## Context

Relates to #38325.

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/platform/github/index.spec.ts
+++ b/lib/modules/platform/github/index.spec.ts
@@ -749,6 +749,39 @@ describe('modules/platform/github/index', () => {
       expect(config).toMatchSnapshot();
     });
 
+    // https://github.com/renovatebot/renovate/discussions/38325
+    it('should warn if deleteBranchOnMerge is not set', async () => {
+      httpMock
+        .scope(githubApiHost)
+        .post(`/graphql`)
+        .reply(200, {
+          data: {
+            repository: {
+              deleteBranchOnMerge: false,
+              nameWithOwner: 'some/repo',
+              hasIssuesEnabled: true,
+              defaultBranchRef: {
+                name: 'master',
+                target: {
+                  oid: '1234',
+                },
+              },
+            },
+          },
+        });
+
+      await github.initRepo({ repository: 'some/repo' });
+
+      expect(logger.logger.warn).toHaveBeenCalledExactlyOnceWith(
+        {
+          repo: 'some/repo',
+          documentationUrl:
+            'https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-the-automatic-deletion-of-branches',
+        },
+        'Repository does not have "Automatically delete head branches" set, which could cause Renovate to re-use a repository in some cases',
+      );
+    });
+
     it('should throw error if archived', async () => {
       httpMock
         .scope(githubApiHost)
--- a/lib/modules/platform/github/index.ts
+++ b/lib/modules/platform/github/index.ts
@@ -582,6 +582,19 @@ export async function initRepo({
       );
       throw new Error(REPOSITORY_ARCHIVED);
     }
+
+    // https://github.com/renovatebot/renovate/discussions/38325
+    if (!repo.deleteBranchOnMerge) {
+      logger.warn(
+        {
+          repo: repository,
+          documentationUrl:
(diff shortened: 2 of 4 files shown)


lib/modules/platform/github/index.ts
```diff
@@ -582,6 +582,19 @@ export async function initRepo({
...
       throw new Error(REPOSITORY_ARCHIVED);
     }
+
+    // https://github.com/renovatebot/renovate/discussions/38325
+    if (!repo.deleteBranchOnMerge) {
+      logger.warn(
+        {
+          repo: repository,
+          documentationUrl:
+            'https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-the-automatic-deletion-of-branches',
```
jamietanna, line 592: I couldn't find us using `documentationUrl` before - is there any prior art?

lib/modules/platform/github/index.spec.ts
```diff
@@ -749,6 +749,39 @@ describe('modules/platform/github/index', () => {
...
+
+      await github.initRepo({ repository: 'some/repo' });
+
+      expect(logger.logger.warn).toHaveBeenCalledExactlyOnceWith(
+        {
+          repo: 'some/repo',
+          documentationUrl:
+            'https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-the-automatic-deletion-of-branches',
+        },
+        'Repository does not have "Automatically delete head branches" set, which could cause Renovate to re-use a repository in some cases',
```
RahulGautamSingh, line 781:
```suggestion
        'Repository does not have "Automatically delete head branches" set, which could cause Renovate to re-use a branch in some cases',
```

RahulGautamSingh (commented):
I am not sure if we should do this.

1. It is not necessary that users only use the github functionality to delete branches, as some can use GH actions to do so.

2. In some cases we were updating existing branches with PRs, not exactly the same as orphan branches.

lib/modules/platform/github/index.ts
secustor: I don't think so, but we should IMO link only our documentation

secustor (commented):
see conversation else LGTM

lib/modules/platform/github/index.ts
jamietanna: Do you think linking to GitHub's official docs will be an issue? Are you concerned about URLs breaking in the future? Or perhaps more that we should "only" try and point to things that we own? (And avoiding a precedent being set, where someone may then think this can be any arbitrary docs?

RahulGautamSingh: Agreed, `documentationUrl` sounds like we point to renovate's doc.

Changing it to `refer` or `referenceUrl` could work, or if not that then we can put the link in a comment in the code.

Also, we should check if we have ability to delete the branches on other platforms or not and make it a generic message if possible.

secustor: External docs lack Renovate context, so it will be incomplete or maybe too verbose. 
Also, as you say, also we would have to hunt down external documentation changes all the time. 

Further links are very long for being logged directly away, maybe rather a reference ID would more sense. That ID could be then looked up in the docs

lib/modules/platform/github/index.ts
```diff
@@ -582,6 +582,19 @@ export async function initRepo({
       );
       throw new Error(REPOSITORY_ARCHIVED);
     }
+
+    // https://github.com/renovatebot/renovate/discussions/38325
+    if (!repo.deleteBranchOnMerge) {
```
viceice, line 587: I think we should add an option to suppress the warning for specific cases? I'm not sure if they really exist 🤷‍♂️

PR #38598, opened 2025-10-13
fix(npm): avoid updating lock files for `engines`, `packageManager`, and `volta` deps

felipecrs:
## Changes

This fixes the issue where Renovate would attempt to update the lock files for `volta` and `packageManager` dependencies in `package.json`.

In most cases, this would be harmless, but in case there is a dependency with the same name as the package manager (e.g., `npm` or `pnpm`), Renovate would try to update the lock file with the new version of that dependency, which is incorrect.

## Context

- [x] This closes an existing Issue: https://github.com/renovatebot/renovate/discussions/35792

## AI assistance disclosure

The public repository: https://github.com/felipecrs/renovate-repro-infinite-loop-npm

### DIFF
--- a/lib/modules/manager/npm/extract/post/locked-versions.ts
+++ b/lib/modules/manager/npm/extract/post/locked-versions.ts
@@ -109,6 +109,11 @@ export async function getLockedVersions(
       const relativeDir = upath.relative(npmRootDir, packageDir);
 
       for (const dep of packageFile.deps) {
+        // Skip dependencies that do not alter package-lock.json
+        if (dep.depType === 'volta' || dep.depType === 'packageManager') {
+          continue;
+        }
+
         // TODO: types (#22198)
         let lockedDepName = dep.depName!;
         if (relativeDir && relativeDir !== '.') {
--- a/lib/modules/manager/npm/extract/post/locked-versions.ts
+++ b/lib/modules/manager/npm/extract/post/locked-versions.ts
@@ -109,8 +109,12 @@ export async function getLockedVersions(
       const relativeDir = upath.relative(npmRootDir, packageDir);
 
       for (const dep of packageFile.deps) {
-        // Skip dependencies that do not alter package-lock.json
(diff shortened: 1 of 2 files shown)


RahulGautamSingh (commented):
Should we do this for `depType=engines` as well?

lib/modules/manager/npm/extract/post/locked-versions.ts
```diff
@@ -109,6 +109,11 @@ export async function getLockedVersions(
       const relativeDir = upath.relative(npmRootDir, packageDir);
 
       for (const dep of packageFile.deps) {
+        // Skip dependencies that do not alter package-lock.json
+        if (dep.depType === 'volta' || dep.depType === 'packageManager') {
```
viceice: what about `engines` and `devEngines.runtime`?

felipecrs: That's a bit more complicated. `engines` is replicated to `package-lock.json` by npm. But let me do some tests while I'm on it.

RahulGautamSingh (approved):
LGTM

lib/modules/manager/npm/extract/post/locked-versions.ts
felipecrs: After testing, I found out that:

1. This fix can be extended to `engines`, i.e. locked version should not be set for `engines` deps
2. There are no regressions in doing so.
3. However, I was not able to reproduce the original issue when replacing `volta` with `engines`. This is probably already handled elsewhere in the code, and if I could figure it out, I could probably remove that piece of code because this solution is sharper.
4. Renovate does not support updating `devEngines`, but otherwise it would surely be applicable

/cc @RahulGautamSingh

felipecrs: Nevermind number 3, I missed a step. I can confirm I was able to reproduce the original issue except changing `volta` with `engines` and confirmed the fix too.

This PR is ready to go from my side.

viceice (approved):

PR #38630, opened 2025-10-15
feat(git): make recursive cloning of submodules optional

tamcore:
## Changes

This PR introduces a small flag `cloneSubmodulesRecursive`, which defaults to `true`, to not alter existing behaviour, that allows the user to turn off the recursive checkout of submodules.

With https://github.com/renovatebot/renovate/commit/0e9e938112b9d6f411b2a3893a184e1be2937cde the `--recursive` flag for `git submodule` was hardcoded. That causes issues for us, where a submodule (`https://github.com/gravitational/teleport`) references another submodule (`https://github.com/gravitational/teleport.e` via `e`), we don't have access to.

## Context

## AI assistance disclosure

- [X] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).

The public repository: <URL>

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -661,7 +661,7 @@ This name will appear in Renovate logs, making it easier to debug or trace speci
 
 Enabling this option will mean that detected Git submodules will be cloned at time of repository clone.
 By default all will be cloned, but this can be customized by configuring `cloneSubmodulesFilter` too.
-Submodules are always cloned recursively.
+Submodules are always cloned recursively by default (see [`cloneSubmodulesRecursive`](#clonesubmodulesrecursive)).
 
 Important: private submodules aren't supported by Renovate, unless the underlying `ssh` layer already has the correct permissions.
 
@@ -671,6 +671,10 @@ Use this option together with `cloneSubmodules` if you wish to clone only a subs
 
 This config option supports regex and glob filters, including negative matches.
 For more details on this syntax see Renovate's [string pattern matching documentation](./string-pattern-matching.md).
+
+## cloneSubmodulesRecursive
+
+Defaults to `true`. Setting this to `false` will disable recursive cloning of submodules.
 
 ## commitBody
 
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -714,7 +714,7 @@ This name will appear in Renovate logs, making it easier to debug or trace speci
 
 Enabling this option will mean that detected Git submodules will be cloned at time of repository clone.
 By default all will be cloned, but this can be customized by configuring `cloneSubmodulesFilter` too.
-Submodules are always cloned recursively.
+Submodules are always cloned recursively by default (see [`cloneSubmodulesRecursive`](#clonesubmodulesrecursive)).
 
 Important: private submodules aren't supported by Renovate, unless the underlying `ssh` layer already has the correct permissions.
 
@@ -724,6 +724,10 @@ Use this option together with `cloneSubmodules` if you wish to clone only a subs
 
 This config option supports regex and glob filters, including negative matches.
 For more details on this syntax see Renovate's [string pattern matching documentation](./string-pattern-matching.md).
+
+## `cloneSubmodulesRecursive`
+
+Defaults to `true`. Setting this to `false` will disable recursive cloning of submodules.
 
 ## `commitBody`
 
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -3167,6 +3167,12 @@ const options: Readonly<RenovateOptions>[] = [
     default: false,
   },
   {
+    name: 'cloneSubmodulesRecursive',
+    description: 'Set to `false` to disable recursive cloning of submodules.',
+    type: 'boolean',
+    default: true,
+  },
+  {
     name: 'cloneSubmodulesFilter',
     description:
       'List of submodules names or patterns to clone when cloneSubmodules=true.',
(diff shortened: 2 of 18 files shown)


docs/usage/configuration-options.md
```diff
@@ -651,7 +651,7 @@ This name will appear in Renovate logs, making it easier to debug or trace speci
 
 Enabling this option will mean that detected Git submodules will be cloned at time of repository clone.
 By default all will be cloned, but this can be customized by configuring `cloneSubmodulesFilter` too.
-Submodules are always cloned recursively.
+Submodules are always cloned recursively by default.
```
```suggestion
Submodules are always cloned recursively by default (see [`cloneSubmodulesRecursive`](#clonesubmodulesrecursive)).
```

jamietanna:
Interesting - cloning of a private repository in Git submodules, sounds like a bit of a painful experience for all contributors, too :/

jamietanna (commented):
Have you seen if the option of [`cloneSubmodulesFilter`](https://docs.renovatebot.com/configuration-options/#clonesubmodules) would be sufficient here?

tamcore:
> Have you seen if the option of [`cloneSubmodulesFilter`](https://docs.renovatebot.com/configuration-options/#clonesubmodules) would be sufficient here?

Yes. Unfortunately that only affects the direct submodules within the renovated repository. After that `Submodules are always cloned recursively`.

RahulGautamSingh:
@tamcore Could you please, fix the coverage issue?

tamcore:
Sorry for the delay. I hope the coverage is fine now. The changes have also been cleanly reapplied onto latest main, since just a rebase wasn't possible.

lib/util/git/index.spec.ts
```diff
@@ -263,6 +263,18 @@ describe('util/git/index', { timeout: 10000 }, () => {
...
+        expect((await git.getRepoStatus()).isClean()).toBeTrue();
+      });
+
+      it('clones submodules non-recursively when cloneSubmodulesRecursive is false', async () => {
+        await git.initRepo({
+          cloneSubmodules: true,
+          cloneSubmodulesRecursive: false,
           url: base.path,
         });
         expect((await git.getRepoStatus()).isClean()).toBeTrue();
```
RahulGautamSingh, line 350: This test passes whether `cloneSubmodulesRecursive` is `true` /`false` (verified locally). 

The fixture needs a nested submodule and the assertion needs to check that the inner submodule is absent when `false`.

lib/config/options/index.ts
```diff
@@ -3286,6 +3286,12 @@ const options: Readonly<RenovateOptions>[] = [
     type: 'boolean',
     default: false,
   },
+  {
+    name: 'cloneSubmodulesRecursive',
+    description: 'Set to `false` to disable recursive cloning of submodules.',
+    type: 'boolean',
```
jamietanna, line 3311: When thinking about this, I'd wondered if there may be cases where we'd want to have this as a numerical config option, so it's possible to tune it i.e. if you're doing `postUpgradeTasks` inside the submodule - it's _unlikely_ so maybe we leave it off for now.

lib/util/git/index.spec.ts
tamcore: Kindly check again :)

lib/config/options/index.ts
jamietanna: To rephrase - I think this is slightly different to https://github.com/renovatebot/renovate/issues/43592 which is more about - when we do clone submodules - whether to use a specific submodule depth

RahulGautamSingh (changes requested):
The new signature is `cloneSubmodulesRecursive: boolean | undefined`

But the two call sites disagree on how they pass it:

- `lib/workers/repository/init/index.ts`: !!config.cloneSubmodulesRecursive (resolved config, default true → recursive)
- syncGit in lib/util/git/index.ts: config.cloneSubmodulesRecursive raw from StorageConfig (can be undefined → non-recursive)

And Gerrit is missed: `lib/modules/platform/gerrit/index.ts` forwards `cloneSubmodules` and `cloneSubmodulesFilter` to `git.initRepo` but not `cloneSubmodulesRecursive`, so its `StorageConfig.cloneSubmodulesRecursive` stays undefined. 

Through the syncGit path that becomes non-recursive. It will be a behavior change for Gerrit users who enabled `cloneSubmodules` (previously always recursive).

Fix once, at the boundary, instead of relying on every platform to remember the passthrough:

```js
// in cloneSubmodules()
const recursive = cloneSubmodulesRecursive ?? true;
```

Then undefined preserves the old always-recursive behavior everywhere, the `!!` vs raw inconsistency stops mattering, and forgetting a platform (Gerrit, or the next one added) can't flip the default.

lib/util/git/index.ts
```diff
@@ -445,9 +447,18 @@ export async function cloneSubmodules(
...
+      }
+
       await gitRetry(() =>
-        git.env(gitEnv).submoduleUpdate(['--init', '--recursive', submodule]),
+        recursive
+          ? git
+              .env(gitEnv)
+              .submoduleUpdate(['--init', '--recursive', submodule])
+          : git.env(gitEnv).submoduleUpdate(['--init', submodule]),
       );
```
RahulGautamSingh, line 462: the ternary duplicates the `git.env(gitEnv).submoduleUpdate(...)` call. 

Build the args instead:
```ts
const args = recursive ? ['--init', '--recursive', submodule] : ['--init', submodule];
await gitRetry(() => git.env(gitEnv).submoduleUpdate(args));
```

RahulGautamSingh (approved):
nit

PR #38720, opened 2025-10-20
fix(workers/branch): skip branch update if `pendingChecks=true`

### LINKED ISSUE #38719: Prevent branch updates when minimumReleaseAge is not met

When a new version is released before an existing renovate branch/pr is merged. Then, in some cases the branch gets updated with the newer version even if the version does not meet the `minimumReleaseAge` criteria.

Some examples:

1. https://github.com/renovatebot/renovate/discussions/38324
2. https://github.com/renovatebot/renovate/discussions/38247

## Fix

If a branch exists and `pendingChecks=true` do not update the branch.

RahulGautamSingh:
## Changes
- Skip updating an existing branch if `pendingChecks=true`

## Context

Closes: #38719 

- [x] This closes an existing Issue: #38719

## AI assistance disclosure

The public repository: https://github.com/Rahul-renovate-testing/min-release-age-branch-update-old-behavior

### DIFF
--- a/lib/workers/repository/update/branch/index.ts
+++ b/lib/workers/repository/update/branch/index.ts
@@ -282,7 +282,7 @@ export async function processBranch(
 
         if (config.pendingChecks) {
           logger.info(
-            'Branch updating is skipped because it contains updates with pendingChecks=true',
+            'Branch updating is skipped because internalChecksFilter was not met',
           );
           return {
             branchExists: true,
--- a/lib/workers/repository/update/branch/index.ts
+++ b/lib/workers/repository/update/branch/index.ts
@@ -268,8 +268,8 @@ export async function processBranch(
 
       const prRebaseChecked = !!branchPr?.bodyStruct?.rebaseRequested;
 
-      if (branchExists && !dependencyDashboardCheck) {
-        if (config.stopUpdating && !prRebaseChecked) {
+      if (branchExists && !dependencyDashboardCheck && !prRebaseChecked) {
+        if (config.stopUpdating) {
           logger.info(
             'Branch updating is skipped because stopUpdatingLabel is present in config',
           );
@@ -282,7 +282,7 @@ export async function processBranch(
 
         if (config.pendingChecks) {
           logger.info(
-            'Branch updating is skipped because it contains updates with unmet release age',
+            'Branch updating is skipped because it contains updates with pendingChecks=true',
           );
           return {
             branchExists: true,
--- a/lib/workers/repository/update/branch/index.ts
+++ b/lib/workers/repository/update/branch/index.ts
@@ -287,7 +287,7 @@ export async function processBranch(
           return {
             branchExists: true,
             prNo: branchPr?.number,
-            result: 'no-work',
+            result: 'pending',
           };
         }
       }
--- a/lib/workers/repository/update/branch/index.ts
+++ b/lib/workers/repository/update/branch/index.ts
@@ -268,10 +268,21 @@ export async function processBranch(
 
       const prRebaseChecked = !!branchPr?.bodyStruct?.rebaseRequested;
 
-      if (branchExists && !dependencyDashboardCheck && config.stopUpdating) {
-        if (!prRebaseChecked) {
+      if (branchExists && !dependencyDashboardCheck) {
+        if (config.stopUpdating && !prRebaseChecked) {
           logger.info(
             'Branch updating is skipped because stopUpdatingLabel is present in config',
+          );
+          return {
+            branchExists: true,
+            prNo: branchPr?.number,
+            result: 'no-work',
+          };
+        }
+
+        if (config.pendingChecks) {
+          logger.info(
+            'Branch updating is skipped because it contains updates with unmet release age',
           );
           return {
             branchExists: true,
--- a/lib/workers/repository/update/branch/index.spec.ts
+++ b/lib/workers/repository/update/branch/index.spec.ts
@@ -723,6 +723,29 @@ describe('workers/repository/update/branch/index', () => {
       });
     });
 
+    it('returns if pending checks - but branch exists', async () => {
+      getUpdated.getUpdatedPackageFiles.mockResolvedValueOnce({
+        ...updatedPackageFiles,
+      });
+      npmPostExtract.getAdditionalFiles.mockResolvedValueOnce({
+        artifactErrors: [],
+        updatedArtifacts: [],
+      });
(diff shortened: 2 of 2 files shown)


jamietanna (commented):
Can we please give this a go with a real repo? I'd like us to get a bit more confidence that this does solve the reported issue, given it's fairly significant

Also, whether we can add any other test(s) as necessary

RahulGautamSingh:
Reproduction:

Current Behaviour: https://github.com/Rahul-renovate-testing/min-release-age-branch-update-old-behavior/pull/2
Behaviour with changes from this PR: https://github.com/Rahul-renovate-testing/min-release-age-branch-update/pulls

In both cases, an update branch exists which can be used to create a PR namely, `renovate/node-22.x`

lib/workers/repository/update/branch/index.ts
```diff
@@ -279,6 +279,17 @@ export async function processBranch(
...
         }
+
+        if (config.pendingChecks) {
+          logger.info(
+            'Branch updating is skipped because it contains updates with unmet release age',
+          );
+          return {
+            branchExists: true,
+            prNo: branchPr?.number,
+            result: 'no-work',
```
RahulGautamSingh: Should we make it `pending`?

jamietanna: I think `pending` would be more self-explanatory? It clarifies that this branch can't be created because it's pending checks, rather than "we didn't have anything to do" i.e. because things are up-to-date

RahulGautamSingh: The branch is present and maybe even PR. We are just not going to update it with the latest version cause that version is pending ie. no work to be done here ... since the user themselves chose the release age

secustor: `no-work` is semantically correct, though a debugging user would wonder why it is `no-work`. 

Yes, they could cross checks logs, so I`m in favour of `pending` and maybe a `reasoning` field or similar in the future.

lib/workers/repository/update/branch/index.ts
```diff
@@ -279,6 +279,17 @@ export async function processBranch(
             result: 'no-work',
           };
         }
+
+        if (config.pendingChecks) {
+          logger.info(
+            'Branch updating is skipped because it contains updates with unmet release age',
```
jamietanna: Is `pendingChecks` _only_ ever for Minimum Release Age? If not, we should amend the wording here

lib/workers/repository/update/branch/index.ts
```diff
@@ -268,8 +268,8 @@ export async function processBranch(
 
       const prRebaseChecked = !!branchPr?.bodyStruct?.rebaseRequested;
 
-      if (branchExists && !dependencyDashboardCheck && config.stopUpdating) {
-        if (!prRebaseChecked) {
+      if (branchExists && !dependencyDashboardCheck) {
+        if (config.stopUpdating && !prRebaseChecked) {
```
jamietanna: Did this need to change?

jamietanna (commented):
I wonder if https://github.com/zharinov/renovate/blob/1cbfe082d3646e3b6ef721462e682b494a0cafb9/lib/workers/repository/update/branch/index.ts#L201-L213 could be amended / added to instead?

jamietanna (commented):
Validating this against another real repo, https://github.com/JamieTanna-Mend-testing/renovate-iss-38324, I can see that this works as expected

lib/workers/repository/update/branch/index.ts
jamietanna: It looks like yes, right now - but I'd recommend consistency with other logs in this file:


```suggestion
        if (config.pendingChecks) {
          logger.info(
            `Branch ${config.branchName} updating is disabled because internalChecksFilter was not met`,
```

jamietanna:
Interestingly, this shows as ([via](https://github.com/JamieTanna-Mend-testing/renovate-iss-38324/issues/2)) (I believe due to the `pending` result:

Screenshot 2025-10-27 at 13 52 24

Is that what we're expecting?

jamietanna (dismissed):
Few comments, otherwise LGTM - thanks!

onigoetz:
Hi, that looks like a nice improvement.

I wonder what happens in the case of grouped updates?

__Given__
- default branch contains
  - a: 1.0.0
  - b: 1.0.0
- renovate branch contains
  - a: 1.0.1 (matching minimumReleaseAge)
  - b: 1.0.0

__When__
renovate finds two new updates
  - a: 1.0.2 (matching minimumReleaseAge)
  - b: 1.0.1 (NOT matching minimumReleaseAge)

__Then__
I would expect the branch to be updated in the following way:
- a: 1.0.2 (matching minimumReleaseAge)
- b: 1.0.0

Is it what happens with this change?

RahulGautamSingh:
> Interestingly, this shows as ([via](https://github.com/JamieTanna-Mend-testing/renovate-iss-38324/issues/2)) (I believe due to the `pending` result:
> 
> Screenshot 2025-10-27 at 13 52 24
> Is that what we're expecting?

Yes

lib/workers/repository/update/branch/index.ts
RahulGautamSingh: Yes, as it checks whether the stopUpdating lavels is present or not, which is what the check is for. Also, present in the info log below it.

lib/workers/repository/update/branch/index.ts
jamietanna: @RahulGautamSingh can we please make this change?

lib/workers/repository/update/branch/index.ts
```diff
@@ -279,6 +279,17 @@ export async function processBranch(
             result: 'no-work',
           };
         }
+
+        if (config.pendingChecks) {
+          logger.info(
+            'Branch updating is skipped because it contains updates with pendingChecks=true',
```
jamietanna: Do you not think it's worth logging the branch name, here?

lib/workers/repository/update/branch/index.ts
RahulGautamSingh: Branch name will be present in the metadata so no need to add.

lib/workers/repository/update/branch/index.ts
RahulGautamSingh: No, present in metadata, like this:
image

jamietanna (approved):

PR #38797, opened 2025-10-22
refactor(tools/schema): use an `options` object

jamietanna:
## Changes

So we can, in the future, modify other configuration more easily.

We should also rename `configurationOptions` to clarify they're Renovate
configuration, not options for this function.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/tools/docs/schema.ts
+++ b/tools/docs/schema.ts
@@ -167,10 +167,17 @@ function createSchemaForChildConfigs(
   }
 }
 
+interface GenerateSchemaOpts {
+  filename?: string;
+  version?: string;
+}
+
 export async function generateSchema(
   dist: string,
-  filename = 'renovate-schema.json',
-  version: string = pkg.version,
+  {
+    filename = 'renovate-schema.json',
+    version = pkg.version,
+  }: GenerateSchemaOpts = {},
 ): Promise<void> {
   const schema = {
     title: `JSON schema for Renovate ${version} config files (https://renovatebot.com/)`,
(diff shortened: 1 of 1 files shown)


tools/docs/schema.ts
```diff
@@ -167,21 +167,28 @@ function createSchemaForChildConfigs(
...
+}
+
 export async function generateSchema(
   dist: string,
-  filename = 'renovate-schema.json',
-  version: string = pkg.version,
+  options: GenerateSchemaOpts = {
+    filename: 'renovate-schema.json',
+    version: pkg.version,
+  },
```
jamietanna: Is this the right syntax for making it possible to provide a default, but allow overriding?

jamietanna: Or would we instead want to do something like:

```typescript
interface GenerateSchemaOpts {
  filename?: string;
  version?: string;
}

const DefaultGenerateSchemaOpts: GenerateSchemaOpts = {
  filename: 'renovate-schema.json',
  version: pkg.version,
}

export async function generateSchema(
  dist: string,
  options?: GenerateSchemaOpts,
): Promise<void> {
  const opts: GenerateSchemaOpts = { ...DefaultGenerateSchemaOpts, ...options }

```

viceice: maybe something like this:
```ts
export async function generateSchema(
  dist: string,
  options?: GenerateSchemaOpts,
): Promise<void> {
  const { version = pkg.version, filename = 'renovate-schema.json' } = options;
```
can also be combined into the function declaration.

viceice: eg;
https://github.com/renovatebot/renovate/blob/f1e38a1f156c11f1afe4ea1dc3a23385516cff5b/lib/config/presets/forgejo/index.ts#L52-L59

jamietanna: Ah nice, thank you - https://github.com/renovatebot/renovate/pull/38800 is where this pattern gets really used, and I think it looks good?

viceice (dismissed):

viceice (approved):

PR #38854, opened 2025-10-24
feat(lookup): add `maxMajorIncrement`

### LINKED ISSUE #20772: Add "max major increment" option with sensible default

Avoid upgrading from normal-looking versions to high versions like daily releases

Example: repository currently uses 19.2. There exists a new version 20.0, but also 2023.03.03 or something like that.

Today, Renovate will take the 2023 one. Instead, we should have a new configuration option like `maxMajorIncrement` defaulting to perhaps `50`. which filters out any releases which are >50 major releases higher than the current.

I think this would produce less problems by default than having no limit does today.

No

felipecrs:
## Changes

This PR adds a new configuration option `maxMajorIncrement` that limits the maximum allowed major version increment for dependency upgrades. This helps avoid upgrading from normal-looking versions to high version numbers like date-based releases.

The default value is set to 500 and setting it to 0 disables this filtering entirely.

Note: in some cases, bumping from 1.0.0 to 2025.0.0 may actually be wanted. But I think it's such a corner case that would be better handled by a human. I propose we let this be, and reconsider based on user feedback. Maybe some new preset for unrestricting major increment for certain known dependencies.

## Context

- [x] This closes an existing Issue: https://github.com/renovatebot/renovate/issues/20772

## AI assistance disclosure

The public repository:

- https://github.com/felipecrs/renovate-repro-high-major-bump/pull/7


felipecrs:
Looks like codecov doesn't take into account `// istanbul ignore if`.

docs/usage/configuration-options.md
```diff
@@ -2630,6 +2630,13 @@ Be careful with remapping `warn` or `error` messages to lower log levels, as it
 
 Add to this object if you wish to define rules that apply only to major updates.
 
+## maxMajorIncrement
+
+By default, Renovate limits major version increments to 50 to avoid accidentally upgrading to date-based versioning or other high version numbers.
```
viceice: attention, the text from option in typescript is added before this text, so please deduplicate it.

lib/config/options/index.ts
```diff
@@ -1720,6 +1720,16 @@ const options: RenovateOptions[] = [
     type: 'boolean',
     default: true,
   },
+  {
```
viceice, line 1729: please move after follow tags for better alphabetical order

lib/workers/repository/process/lookup/filter.ts
```diff
@@ -66,6 +71,27 @@ export function filterVersions(
     });
   }
 
+  if (maxMajorIncrement && maxMajorIncrement > 0) {
+    const currentMajor = versioningApi.getMajor(currentVersion);
+    if (currentMajor !== null) {
+      filteredReleases = filteredReleases.filter((r) => {
+        const releaseMajor = versioningApi.getMajor(r.version);
```
viceice: please extract to a module scoped function for better maintainability

docs/usage/configuration-options.md
felipecrs: Oh. I didn't realize that. Thank you.

felipecrs: I rewrote documentation in https://github.com/renovatebot/renovate/pull/38854/commits/149acff5ac3c28510d1c849c013488a103e75d48, I hope you like it better.

lib/workers/repository/process/lookup/filter.ts
```diff
@@ -24,15 +24,49 @@ function isReleaseStable(
...
+  depName: string,
+): Release[] {
+  const currentMajor = versioningApi.getMajor(currentVersion);
+  if (currentMajor === null) {
+    return releases;
+  }
+
+  return releases.filter((r) => {
+    const releaseMajor = versioningApi.getMajor(r.version);
+    // istanbul ignore if: shouldn't happen
```
viceice:
```suggestion
    /* v8 ignore next 3 -- shouldn't happen */
```

astellingwerf:
It might be of interest to look at #38922, as they seem to aim/support the same intent.

docs/usage/configuration-options.md
```diff
@@ -2630,6 +2630,12 @@ Be careful with remapping `warn` or `error` messages to lower log levels, as it
 
 Add to this object if you wish to define rules that apply only to major updates.
 
+## maxMajorIncrement
+
+This is particularly useful for preventing upgrades from [SemVer](https://semver.org/)-based versions to [CalVer](https://calver.org/)-based ones.
+
+For example, if you're on version `19.2.0`, Renovate will upgrade to `21.0.0` but filter out `2023.3.3`.
+
```
RahulGautamSingh, line 2654: Please also add an example.

RahulGautamSingh (commented):
LGTM

docs/usage/configuration-options.md
felipecrs: @RahulGautamSingh, I actually had included an example initially, but then reconsidered.

I don’t think this option needs an example, since it’s not something users are expected to tweak often. Including one might unintentionally suggest it’s common or recommended to change it.

For the few advanced users who will need to adjust this, the current explanation in the documentation should be clear enough.

However, if you still think I should add an example, please let me know and I'll readily do it. :)

RahulGautamSingh: I think an example of using it with `packageRules` to tweak behavior for a dep will be the use-case for most adv. users. So, that can be added.

felipecrs: Ok. I added it. I'm still not convinced this is a good thing.

The only real example I can think of (Home Assistant, as mentioned in the PR body) is a weird one. Allowing Renovate to upgrade `homeassistant` from 0.x to 2025.x would probably be unwanted anyway, because the dependency has been completely changed from one version to another. Ideally such a big bump should be done by a human.

RahulGautamSingh: You could modify to rule to restrict such changes, instead of allowing as that makes more sense.

RahulGautamSingh (commented):
Added comment.

docs/usage/configuration-options.md
felipecrs: > You could modify to rule to restrict such changes, instead of allowing as that makes more sense.

Well, then I don't need to write any example because that is what will happen by default. :)

RahulGautamSingh: Good to go in that case.

felipecrs: Alright, but do you mean I should keep or remove the example?

docs/usage/configuration-options.md
```diff
@@ -2630,6 +2630,26 @@ Be careful with remapping `warn` or `error` messages to lower log levels, as it
...
+{
+  "packageRules": [
+    {
+      "description": "Allow to upgrade Home Assistant from 0.x to 2025.x",
+      "matchPackageNames": ["homeassistant"],
+      "maxMajorIncrement": 0
+    }
+  ]
+}
+```
```
RahulGautamSingh, line 2667:
```suggestion
Limits how many major versions Renovate can upgrade to.

By default, Renovate allows a maximum jump of 50 major versions.
For example, if you’re currently on `1.0.0`, the highest upgrade offered will be `51.0.0`, even if `52.0.0` is available.

This is useful for preventing upgrades from [SemVer](https://semver.org/)-based versions to [CalVer](https://calver.org/)-based ones.

For example, if you're on version `19.2.0`, Renovate will upgrade to `21.0.0` but filter out `2023.3.3`.

You can combine this setting with packageRules to fine-tune it for specific dependencies.

Examples:

1. Allow unlimited upgrades (set to `0`):

```json
{
  "packageRules": [
    {
      "description": "Allow to upgrade Home Assistant from 0.x to 2025.x",
      "matchPackageNames": ["homeassistant"],
      "maxMajorIncrement": 0
    }
  ]
}
```

2. Restrict the increment further (e.g. limit to `10` major versions):

```json
{
  "packageRules": [
    {
      "description": "Limit xmldoc major upgrades to within 10 versions",
      "matchPackageNames": ["homeassistant"],
      "maxMajorIncrement": 10
    }
  ]
}
```


```

**Review by RahulGautamSingh — CHANGES_REQUESTED** [review]

File: docs/usage/configuration-options.md (hunk 6 above)
Comment (RahulGautamSingh) [reply docs/usage/configuration-options.md:2667]: How about this? Adding a second example,covers both the use-cases. 

Not sure why Github formatting breaks the suggested changes apart though :)

Comment (felipecrs) [reply docs/usage/configuration-options.md:2667]: Thank you, I like your suggestion.

Comment (felipecrs) [reply docs/usage/configuration-options.md:2667]: Done in 129754c51, I hope you like it. I tried keeping the existing style of other examples.

**Comment by felipecrs** [comment]
@RahulGautamSingh, @viceice, I also had to squash commits for this PR to update my git identity.

No changes since the last change (da1d41f17).

File: docs/usage/configuration-options.md (hunk 5 above)
Comment (RahulGautamSingh) [reply docs/usage/configuration-options.md:2654]: I like how it is now, so no further change

**Review by RahulGautamSingh — APPROVED** [review]
LGTM

File: lib/config/options/index.ts [hunk 7]
```diff
@@ -1735,6 +1735,16 @@ const options: RenovateOptions[] = [
...
+    name: 'maxMajorIncrement',
+    description:
+      'Limit the maximum major version increment allowed. Set to 0 to disable.',
+    stage: 'package',
+    type: 'integer',
+    default: 500,
+    cli: false,
+    env: false,
```
jamietanna, line 1746: Makes sense 👍

lib/workers/repository/process/lookup/filter.ts
```diff
@@ -24,17 +24,51 @@ function isReleaseStable(
...
+    const releaseMajor = versioningApi.getMajor(r.version);
+    /* v8 ignore next 3 -- shouldn't happen */
+    if (releaseMajor === null) {
+      return true;
+    }
+    const majorIncrement = releaseMajor - currentMajor;
+    if (majorIncrement > maxMajorIncrement) {
+      logger.trace(
+        `Skipping ${depName}@${r.version} because major increment ${majorIncrement} exceeds maxMajorIncrement ${maxMajorIncrement}`,
+      );
```
jamietanna: IMO:

```suggestion
      logger.once.debug(
        `Skipping ${depName}@${r.version} because major increment ${majorIncrement} exceeds maxMajorIncrement ${maxMajorIncrement}`,
      );
```

The `once` removes the risk of it getting too noisy, but given more folks are used to reading their DEBUG logs, it should help anyone catch the change, especially as we're setting a default on this for all users

docs/usage/configuration-options.md
```diff
@@ -2646,6 +2646,42 @@ Be careful with remapping `warn` or `error` messages to lower log levels, as it
...
+
+You can use `packageRules` to customize this behavior for specific packages:
+
+```json
+{
+  "packageRules": [
+    {
+      "description": "Allow to upgrade Home Assistant from 0.x to 2025.x",
+      "matchPackageNames": ["homeassistant"],
+      "maxMajorIncrement": 0
```
jamietanna, line 2663: Would it also be possible to set it to `null`? What happens then?

jamietanna: That may actually be more predictable - `0` _could_ mean "never upgrade" 🤔

jamietanna (changes requested):
Logging change + outstanding question, then LGTM!

docs/usage/configuration-options.md
felipecrs: @jamietanna, it could be. But maybe it's better to keep consistency with other numeric limits like [prConcurrentLimit](https://docs.renovatebot.com/configuration-options/#prconcurrentlimit), [branchConcurrentLimit](https://docs.renovatebot.com/configuration-options/#branchconcurrentlimit), [prHourlyLimit](https://docs.renovatebot.com/configuration-options/#prhourlylimit), [prCommitsPerRunLimit](https://docs.renovatebot.com/self-hosted-configuration/#prcommitsperrunlimit) where `0` means no limit.

What do you think?

> That may actually be more predictable - 0 could mean "never upgrade" 🤔

Yeah, I see. Note however disabling major updates should be done through:

```json
{
  "packageRules": [
    {
      "matchUpdateTypes": ["major"],
      "enabled": false
    }
  ]
}
```

Maybe I should add a note to documentation about it?

jamietanna: No I think that's a fair call - happy to keep it in-sync with others 👍

jamietanna (approved):

PR #38892, opened 2025-10-28
fix(maven): determine properly closed ranges

Vampire:
## Changes

A Maven range like [1,2) contains various (usually pre-release) versions of the 2.x line.
By Maven definition the smallest possible 2.x version is 2-alpha, so a properly closed range that only allows 1.x versions
would be [1,2-alpha). This is now detected and properly advanced.

## Context

## AI assistance disclosure


lib/modules/versioning/maven/compare.ts
```diff
@@ -524,6 +524,16 @@ function autoExtendMavenRange(
       interval.leftValue = coerceRangeValue(leftValue, newValue);
       interval.rightValue = incrementRangeValue(interval.leftValue);
     }
+  } else if (
```
RahulGautamSingh, line 545: The versioning code and this addition is a bit hard to understand on first try.

Please, add a comment here to explain what is happening & why so it's more readable?

RahulGautamSingh (changes requested):

lib/modules/versioning/maven/compare.ts
```diff
@@ -520,10 +520,34 @@ function autoExtendMavenRange(
...
+    // 4-alpha is smaller than 4.0.0 for example
+    // while often people that intend to only match version 1.x use [1,2)
+    // this is not correct as it matches various 2.x versions, mostly pre-versions
+    // so the correct range definition would be [1,2-alpha) as nothing
+    // starting with 2 can be lower than 2-alpha in Maven version ordering logic
+    //
+    // so this if handles the conceptually same case as the last if
+    // just for the case when proper Maven ranges are used
+    // [1,2-alpha) / 4.3.2 => [4,5-alpha)
+    if (compare(newValue, leftValue) !== -1) {
```
zharinov: If compare result is other than `-1`, is it okay to skip the following if-s? (i.e. checking `rightValue` and `leftValue` for `null`)

Looks like it's safer to merge nested if-s here, isn't it?

Vampire:
This now has a failing test that will be fixed once #38895 is merged first

lib/modules/versioning/maven/compare.ts
Vampire: It is the same logic as in the pre-existing first case above it.
And yes, it is safe.
If you land in that "if", `leftValue` and `rightValue` are guaranteed to be non-null, the other two `else if`s are for the cases where only one of the values is non-null.
So to get the same result you would then need to check additionally that the other value is not set.

What we could do is to drag the both are set case up from the first two, that might indeed be nicer.

Vampire: Ah, no, that would also change the logic, I think it needs to stay as it is

Vampire: If I'm not mistaken that is when you have a range from 1.2 to 1.3 and the new value is either 1.2 or less than 1.2 in which case the range is not modified.

But this is independent from the change here if the behavior is wrong I'd say.

RahulGautamSingh (changes requested):
> This now has a failing test that will be fixed once #38895 is merged first

@Vampire Please request a re-review once #38895 is merged

Vampire:
@RahulGautamSingh you finally merged #38895, so I know merged from main and pushed.
If I diff main against my branch there are 2 files changed.
Do you have any idea why GitHub is saying there are 1767 commits and 2468 changed files? o_O

Vampire:
Ah, wrong parent order.
I did merge my branch into `main` instead of `main` into my branch.
Now it looks sane again.

RahulGautamSingh (changes requested):
LGTM but missing coverage

Vampire:
What do you mean @RahulGautamSingh?
That the `if (compare(newValue, leftValue) !== -1) {` line is not fully covered?
I don't think that this can ever happen or be covered, but I copied the line from the existing case above.
Imho `compare(newValue, leftValue)` can only be `-1` if `leftValue` is greater than `newValue`.
But `const interval` only delivers the biggest interval where `newValue` is to the right of the right boundary,
and the enclosing `if` of the line in question only let this line be reached if rightvalue is leftvalue + 1.

So if I got the code right, you can only get `false` for that `if` condition if the new value is at the same time on the right of the right value and on the left of the left value while it is guaranteed that the right value is the left value + 1.

So if I'm right, the `if` is always `true` and coverage for that line cannot be completed.
It could maybe be deleted if I am right, but as I said, I just copied that from the case above.

RahulGautamSingh (commented):
I meant the CI coverage failure. If you think it's a mishap from the codecov then please update the branch so the CI check runs again.

Or, if that path cannot be tested add a ignore statement.

lib/modules/versioning/maven/compare.ts
```diff
@@ -534,10 +534,34 @@ function autoExtendMavenRange(
...
+    // and the new version is outside the range, construct a new range that follows the same semantic
+    // [1,2) / 4.3.2 => [4,5)
     if (compare(newValue, leftValue) !== -1) {
       interval.leftValue = coerceRangeValue(leftValue, newValue);
       interval.rightValue = incrementRangeValue(interval.leftValue);
     }
+  } else if (
+    leftValue !== null &&
+    rightValue !== null &&
+    incrementRangeValue(leftValue) + '-alpha' === rightValue
```
RahulGautamSingh: `-alpha` is hardcoded twice (here and L563). extract to a const at the top?

lib/modules/versioning/maven/compare.ts
```diff
@@ -534,10 +534,34 @@ function autoExtendMavenRange(
...
+    // https://maven.apache.org/pom.html#Version_Order_Specification
+    // 4-alpha is smaller than 4.0.0 for example
+    // while often people that intend to only match version 1.x use [1,2)
+    // this is not correct as it matches various 2.x versions, mostly pre-versions
+    // so the correct range definition would be [1,2-alpha) as nothing
+    // starting with 2 can be lower than 2-alpha in Maven version ordering logic
+    //
+    // so this if handles the conceptually same case as the last if
+    // just for the case when proper Maven ranges are used
+    // [1,2-alpha) / 4.3.2 => [4,5-alpha)
```
RahulGautamSingh, line 560: this only matches the exact `N-alpha` form..something like `[1,2.0.0-alpha)` is the same thing in Maven ordering but won't be picked up here. worth a line in the comment so users with non-canonical forms know?

RahulGautamSingh:
small thing: typo in the title, `determing` → `determining`

lib/modules/versioning/maven/compare.ts
Vampire: I'm not sure what you mean.
The existing code handles `[1,2)`, `[1.1,1.2)`, `[1.1.1,1.1.2)`.
The new code handles the same cases just with the prefix  `[1,2-alpha)`, `[1.1,1.2-alpha)`, `[1.1.1,1.1.2-alpha)`.

And the comment says that it only handles cases where increasing the left by one results in the right.
This does not match on something like `[1,2.0.0-alpha)`.

RahulGautamSingh (approved):

lib/modules/versioning/maven/compare.ts
```diff
@@ -534,10 +535,35 @@ function autoExtendMavenRange(
...
+    /* v8 ignore next */
```
viceice: please add a test instead

Vampire: @viceice Again, as far as I see a test is technically impossible.
Please read my comment https://github.com/renovatebot/renovate/pull/38892#issuecomment-4149085598.
In my opinion this case cannot happen.
I just copied it from the case above where it was already there too and is also not covered.
Imo those `if`s can be removed except if they should safe-guard against future changes when it would become relevant.

viceice: then it should be removed. dead code without explanation is bad. nobody knows it after some time

Vampire: I agree, but I'm not confident enough to remove it and am not really a TypeScript developer.
If you as maintainer and TypeScript expert analyze it and say it is really dead code and should be removed,
I suggest you remove it from the upper case on `main` and then I also remove it here,
or you can tell me to remove both, then I do it, but I don't give any guarantees in that case. :-)

Vampire:
So, how do we proceed here?

lib/modules/versioning/maven/compare.ts
viceice: just remove it here, I don't have time to check the existing code above soon. we are now more strict with ignore code coverage for new code. so when a case can't happen, don't handle it at all.

viceice (commented):
see comment

RahulGautamSingh (approved):

viceice (approved):

RahulGautamSingh (approved):

PR #38893, opened 2025-10-28
fix(maven): normalize both sides of version comparison

Vampire:
> [!WARNING]  
> This PR is a follow-up to #38892, the first commit here is to be ignored

## Changes

Without this a leftValue of v3 is incremented to 4 and then compared to v4.

## Context

## AI assistance disclosure

### DIFF
--- a/lib/modules/versioning/maven/compare.ts
+++ b/lib/modules/versioning/maven/compare.ts
@@ -518,7 +518,7 @@ function autoExtendMavenRange(
   if (
     leftValue !== null &&
     rightValue !== null &&
-    incrementRangeValue(leftValue) === rightValue
+    incrementRangeValue(leftValue) === coerceRangeValue(rightValue, rightValue)
   ) {
     if (compare(newValue, leftValue) !== -1) {
       interval.leftValue = coerceRangeValue(leftValue, newValue);
--- a/lib/modules/versioning/maven/compare.ts
+++ b/lib/modules/versioning/maven/compare.ts
@@ -524,6 +524,16 @@ function autoExtendMavenRange(
       interval.leftValue = coerceRangeValue(leftValue, newValue);
       interval.rightValue = incrementRangeValue(interval.leftValue);
     }
+  } else if (
+    leftValue !== null &&
+    rightValue !== null &&
+    incrementRangeValue(leftValue) + '-alpha' ===
+      coerceRangeValue(rightValue, rightValue)
+  ) {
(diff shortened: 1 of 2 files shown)


lib/modules/versioning/maven/compare.ts
```diff
@@ -518,12 +518,22 @@ function autoExtendMavenRange(
   if (
     leftValue !== null &&
     rightValue !== null &&
-    incrementRangeValue(leftValue) === rightValue
+    incrementRangeValue(leftValue) === coerceRangeValue(rightValue, rightValue)
```
RahulGautamSingh, line 521: This change is here to make sure we do not compare left value which will have no `v` prefix to a right value with it?

Vampire: Exactly.
https://github.com/renovatebot/renovate/blob/main/lib/modules/versioning/maven/compare.ts#L118 which is used by  `incrementRangeValue` and `coerceRangeValue` does silently drop the `v` prefix.
So by only using it on the left side of the comparison there will never be a match if the versions are `v`-prefixed.

lib/modules/versioning/maven/compare.ts
```diff
@@ -518,12 +518,22 @@ function autoExtendMavenRange(
...
   ) {
     if (compare(newValue, leftValue) !== -1) {
       interval.leftValue = coerceRangeValue(leftValue, newValue);
       interval.rightValue = incrementRangeValue(interval.leftValue);
     }
+  } else if (
+    leftValue !== null &&
+    rightValue !== null &&
+    incrementRangeValue(leftValue) + '-alpha' ===
```
zharinov, line 530: This `-alpha` looks too specific to me. Any idea for more generic solution?

Vampire: Again, this is not part this PR, but in the previous #38892 as noted in the warning in the start of the description.
And yes, it is quite specific as it is explicitly documented as the lowest possible thing in the Maven specification. ;-)

lib/modules/versioning/maven/compare.ts
Vampire: Ah, I guess using `compare` would be the proper thing to do

Vampire: No, it is not, that works too semantically and breaks tests

Vampire:
Obsoleted by changed #38895

PR #38895, opened 2025-10-28
fix(maven): do not loose the v-prefix of versions

Vampire:
## Changes

If versions start with a v prefix, this was lost when Maven ranges were involved.
The coerce and increment methods stripped the v away and it was never added back in.
Also version comparisons misbehaved in some situations.

## Context

## AI assistance disclosure


lib/modules/versioning/maven/compare.ts
```diff
@@ -515,31 +515,42 @@ function autoExtendMavenRange(
   }
 
   const { leftValue, rightValue } = interval;
+  const vPrefix = (leftValue ?? rightValue)?.startsWith('v') ? 'v' : '';
   if (
     leftValue !== null &&
     rightValue !== null &&
-    incrementRangeValue(leftValue) === rightValue
+    incrementRangeValue(leftValue) === coerceRangeValue(rightValue, rightValue)
```
RahulGautamSingh: What is the logic behind `coerceRangeValue(rightValue, rightValue)`? Won't it return `rightValue` as is.

Vampire: This change belongs to #38893, not this one.
But no, it would not, due to this nasty little line which removes the leading `v` which is then never brought back: https://github.com/renovatebot/renovate/blob/main/lib/modules/versioning/maven/compare.ts#L118

RahulGautamSingh: In which order should we review your PRs?

Vampire: Just read the warning at the top of the description :-)

Vampire: First #38892, then #38893, then this.
The other two are independent.

RahulGautamSingh (commented):
Would it not be bettter  to fix the `incrementRangeValue` and `coerceRangeValue` functions instead?

Vampire:
Maybe.
And if so, this might then also fix #38893 at the same time without needing that PR.
The question is where to properly target the fix.
The `v` is silently dropped in `tokenize` which is used in `coerceRangeValue` and `incrementRangeValue`  and other places.
Should this already not strip it away but deliver it as qualifier token, or other token?
Or should `coerceRangeValue` and `incrementRangeValue` see that there was the leading `v` and add it back?
...

I actually just did the minimal changes to fix my use-case. :-D

Vampire:
Ok, new try.
Now the underlying tokenizing is adapted.
The PR is now stand-alone and can be reviewed in total, #38893 is now obsolete

lib/modules/versioning/maven/compare.spec.ts
```diff
@@ -276,11 +276,57 @@ describe('modules/versioning/maven/compare', () => {
...
+        expected: [
+          {
+            isTransition: true,
+            prefix: '',
+            type: 'TYPE_QUALIFIER',
+            val: 'version',
+          },
           {
             isTransition: false,
             prefix: 'PREFIX_HYPHEN',
```
RahulGautamSingh, line 312: won't it be no prefix, before 2?

RahulGautamSingh:
Also, could you please update the PR description to reflect the new changes.

lib/modules/versioning/maven/compare.spec.ts
Vampire: I didn't touch the logic for tokens beside the first token. So it is hyphen prefix on 2 with isTransition true on the previous token. Probably because for the ordering `a1` and `a-1` are identical. Probably according to Maven specs the `v` should be treated as an own token, not a prefix. But the code just like before treats the leading v specially and assumes that both sides have it or not have the v prefix. If you have one version with and one without v, the new code misbehaves the same as the old one. But treating it as token would need even deeper changes, as the major version would not any longer be the first token and so on. Thus I decided to just treat it as prefix like before just to not loose it on reconstruction.

Vampire:
> Also, could you please update the PR description to reflect the new changes.

What do you want me to change concretely? From my PoV the description is still correct, just the approach changed.

Vampire:
Ah no, the comparison also works better now

Vampire:
Fixed

RahulGautamSingh (commented):
LGTM


@zharinov Can you review this please

viceice (approved):

PR #38922, opened 2025-10-29
feat(config): allow template fields in `allowedVersions`

RahulGautamSingh:
## Changes
- template.compile `allowedVersions` before using it for filtering versions
- allowed template fields are `currentVersion`, `patch`, `minor`, `major`

## Context

- This will make some frequently asked about behavior like: https://github.com/renovatebot/renovate/discussions/38528 possible

## AI assistance disclosure

The public repository: <URL>


lib/util/template/index.spec.ts
```diff
@@ -214,6 +214,12 @@ describe('util/template/index', () => {
     expect(output).toBe('ghc/dep');
   });
 
+  it('add', () => {
+    const userTemplate = '{{add 1 2}}';
+    const output = template.compile(userTemplate, undefined as never);
```
viceice:
```suggestion
    const output = template.compile(userTemplate, {});
```
does this suppress the error? I like to avoid such never cast if possible

lib/util/template/index.ts
```diff
@@ -13,6 +13,9 @@ type Options = HelperOptions & {
 
 handlebars.registerHelper('encodeURIComponent', encodeURIComponent);
 handlebars.registerHelper('decodeURIComponent', decodeURIComponent);
+handlebars.registerHelper('add', function (a, b) {
+  return a + b;
```
viceice: what if a is somehow a number string. I think users expect conversation to number ? I think this will currently do a string concat.

also what about null and undefined handling?

lib/workers/repository/process/lookup/filter.ts
```diff
@@ -66,23 +67,43 @@ export function filterVersions(
...
+    // allowedVersions tempalte support
+    const computedAllowedVersions = template.compile(allowedVersions, {
+      currentVersion,
+      major: currentMajor,
+      minor: currentMinor,
+      patch: currentPatch,
+    });
+
+    // eslint-disable-next-line
+    console.log({
```
viceice: ?

lib/workers/repository/process/lookup/filter.ts
```diff
@@ -66,23 +67,43 @@ export function filterVersions(
     });
   }
 
+  const currentMajor = versioningApi.getMajor(currentVersion);
+  const currentMinor = versioningApi.getMinor(currentVersion);
+  const currentPatch = versioningApi.getPatch(currentVersion);
+
   if (allowedVersions) {
-    const isAllowedPred = getRegexPredicate(allowedVersions);
+    // allowedVersions tempalte support
```
viceice:
```suggestion
    // allowedVersions template support
```

lib/workers/repository/process/lookup/filter.ts
```diff
@@ -66,23 +67,43 @@ export function filterVersions(
...
+    const computedAllowedVersions = template.compile(allowedVersions, {
+      currentVersion,
+      major: currentMajor,
+      minor: currentMinor,
+      patch: currentPatch,
```
viceice, line 118: those can be undefined or null

docs/usage/configuration-options.md
```diff
@@ -2985,6 +2977,21 @@ For example, if you want to upgrade to Angular v1.5 but _not_ to `angular` v1.6
 
 Renovate calculates the valid syntax for this at runtime, because it depends on the dynamic versioning scheme.
 
+You can also use the template fields `currentVersion`, `major`, `minor` and `patch` with `allowedVersions`.
+
+For example, if you want to upgrade to Angular v1.5.0 but _not_ to `angular` v1.7.0 or higher, you could set `allowedVersions` as such:
```
```suggestion
For example, if you want to upgrade to Angular v1.5.0 but _not_ to `angular` v1.7.0 or higher, you could set `allowedVersions` like this:
```

This isn't strictly true, is it? To the best of my understanding, this particular templating defines the maximum increment. Once it's on 1.5.x or 1.6.x, Renovate would still propose an update to 1.7 and beyond.

lib/workers/repository/process/lookup/filter.ts
```diff
@@ -91,27 +112,27 @@ export function filterVersions(
...
       const error = new Error(CONFIG_VALIDATION);
       error.validationSource = 'config';
-      error.validationError = 'Invalid `allowedVersions`';
+      error.validationError = 'Invalid `computedAllowedVersions`';
       error.validationMessage =
-        'The following allowedVersions does not parse as a valid version or range: ' +
-        JSON.stringify(allowedVersions);
+        'The following computedAllowedVersions does not parse as a valid version or range: ' +
+        JSON.stringify(computedAllowedVersions);
       throw error;
```
astellingwerf: This would be confusing, since the configuration doesn't contain a `computedAllowedVersions` property.

lib/util/template/index.ts
```diff
@@ -13,6 +13,9 @@ type Options = HelperOptions & {
 
 handlebars.registerHelper('encodeURIComponent', encodeURIComponent);
 handlebars.registerHelper('decodeURIComponent', decodeURIComponent);
+handlebars.registerHelper('add', function (a, b) {
```
astellingwerf: We'll wanna mention this on `docs/usage/templates.md`.

astellingwerf (changes requested):

docs/usage/configuration-options.md
RahulGautamSingh: Yes it depends on the current version

lib/util/template/index.ts
astellingwerf: I created #38932 to prevent omission.

lib/workers/repository/process/lookup/filter.ts
RahulGautamSingh: I think a warning & not filtering based on `allowedVersions` will be the way to go. As handlebars will replace the `null` fields with empty string which will break the filtering and might throw errors.

What do you think, throw error or no filtering?

lib/util/template/index.ts
secustor, line 1: `allowedVersions` need to be added to `exposedConfigOptions`, as it doesn't has the `Template` suffix and is templated

lib/workers/repository/process/lookup/filter.ts
secustor: From an UX perspective, an error is preferable rather than unexpected behavior which is hard to debug.

lib/util/template/index.ts
```diff
@@ -13,6 +13,9 @@ type Options = HelperOptions & {
...
+  return parseInt(a) + parseInt(b);
```
astellingwerf:
```suggestion
  return parseInt(a, 10) + parseInt(b, 10);
```

lib/util/template/index.ts
RahulGautamSingh: How should we handle undefined/null? Return 0?

viceice: I'm not sure, probably return zero

viceice: I'm not sure, probably return zero

docs/usage/configuration-options.md
RahulGautamSingh: Updated the description

lib/util/template/index.ts
RahulGautamSingh: I think we should leave it unhandled, except for adding a `parseInt` so that it thorws when null/undefiend is passed...which will result in a good error message in the debuh logs like this:

```log
DEBUG: Aggregate error is thrown (repository=Rahul-renovate-testing/allowed-versions-template-error)
       "err": {
         "stack": "AggregateError: \n    at next (/home/rahulgautamsingh/renovate/node_modules/.pnpm/p-map@6.0.0/node_modules/p-map/index.js:2:2821)\n    at processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at /home/rahulgautamsingh/renovate/node_modules/.pnpm/p-map@6.0.0/node_modules/p-map/index.js:2:3417",
         "errors": [
           {
             "validationSource": "config",
             "validationError": "Invalid `allowedVersions`",
             "validationMessage": "The following allowedVersions does not parse as a valid version or range: \"<22.0.NaN\"",
             "message": "config-validation",
             "stack": "Error: config-validation\n    at filterVersions (/home/rahulgautamsingh/renovate/lib/workers/repository/process/lookup/filter.ts:2:3989)\n    at lookupUpdates (/home/rahulgautamsingh/renovate/lib/workers/repository/process/lookup/index.ts:2:10018)\n    at processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at /home/rahulgautamsingh/renovate/lib/workers/repository/process/fetch.ts:2:3336\n    at Function.wrap (/home/rahulgautamsingh/renovate/lib/util/stats.ts:2:2180)\n    at /home/rahulgautamsingh/renovate/lib/workers/repository/process/fetch.ts:2:4367\n    at /home/rahulgautamsingh/renovate/node_modules/.pnpm/p-map@6.0.0/node_modules/p-map/index.js:2:3195"
           }
         ]
       }
```

RahulGautamSingh: Instead if we return 0 this kind of error will be swallowed

lib/util/template/index.spec.ts
```diff
@@ -214,6 +214,18 @@ describe('util/template/index', () => {
...
+    const userTemplate = '{{add 1 2}}';
+    const output = template.compile(userTemplate, {});
+    expect(output).toBe('3');
+  });
+
+  it('add - returns 0 if any pf the inputs undefined', () => {
+    const userTemplate = '{{add 1 undefined}}';
+    const output = template.compile(userTemplate, {});
+    expect(output).toBe('0');
+  });
```
RahulGautamSingh:
```suggestion
```

lib/util/template/index.spec.ts
```diff
@@ -214,6 +214,13 @@ describe('util/template/index', () => {
     expect(output).toBe('ghc/dep');
   });
 
+  it('add', () => {
+    const userTemplate = '{{add 1 2}}';
+    const output = template.compile(userTemplate, {});
+    expect(output).toBe('3');
+  });
+
+
```
RahulGautamSingh, line 227:
```suggestion
```

lib/util/template/index.ts
```diff
@@ -96,6 +96,7 @@ const helpers: Record<string, handlebars.HelperDelegate> = {
       return true;
     });
   },
+  add: (a, b) => parseInt(a) + parseInt(b),
```
viceice: does this throw?

viceice (commented):
it should error out if add failed because of undefined or null

docs/usage/configuration-options.md
```diff
@@ -2981,6 +2981,21 @@ For example, if you want to upgrade to Angular v1.5 but _not_ to `angular` v1.6
 
 Renovate calculates the valid syntax for this at runtime, because it depends on the dynamic versioning scheme.
 
+You can also use the template fields: `currentVersion`, `major`, `minor` and `patch` with `allowedVersions`.
```
```suggestion
You can also use the following template fields with `allowedVersions`:

- `currentVersion`
- `major`
- `minor`
- `patch` 
```

lib/util/template/index.ts
RahulGautamSingh: No, it will return `NaN`.

Should we throw an explicit error here, you think? Example:

```ts
 add: (a, b) => {
    if (isNumber(a) && isNumber(b)) {
      return a + b;
    }

    throw new Error('add: inputs are not valid');
  },
```

viceice: yes. we should be more explicit here

viceice (commented):
see comment

viceice (dismissed):

viceice (approved):

PR #38961, opened 2025-10-30
test(hostrules/github): add clearer precedence examples

jamietanna:
## Changes

As part of #38953, we are changing the way that host rules work for
GitHub.com, and automagically wiring them in for datasources that
interact with `github.com`.

To do this, we should first provide a clearer definition of how host
rules should work, especially when there may be conflicting sources of
tokens, or where we may be on GitHub Enterprise Server vs GitHub.com.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/util/http/host-rules.spec.ts
+++ b/lib/util/http/host-rules.spec.ts
@@ -24,6 +24,11 @@ describe('util/http/host-rules', () => {
       token: 'token',
     });
     hostRules.add({
+      hostType: 'github',
+      matchHost: 'api.github.com',
+      token: 'dotcom-api-token',
+    });
+    hostRules.add({
       hostType: 'gitea',
       password: 'password',
     });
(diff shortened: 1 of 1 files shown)


lib/util/http/host-rules.spec.ts
```diff
@@ -23,6 +23,11 @@ describe('util/http/host-rules', () => {
       hostType: 'github',
       token: 'token',
     });
+    hostRules.add({
+      hostType: 'github',
```
jamietanna, line 27: I wonder if maybe this should be removed 🤔

jamietanna: It's more likely that folks will use `matchHost` rather than `hostType` _and_ `matchHost`

I.e. in https://docs.renovatebot.com/presets-default/#githubcomtokenarg0 that's what we do

viceice (approved):

lib/util/http/host-rules.spec.ts
viceice: it's in use to configure host type detection from managers.

https://github.com/visualon/renovate-config/blob/de0d85d5455b64d695ccca92b6169eb61386ca45/shared.json#L272

jamietanna: Gotcha - I guess would we expect folks to set both `hostType` and `matchHost` when setting custom host rules i.e. when using GitHub Enterprise Server?

If so, we can leave this in, and then should correct the preset ☝️

viceice: yes, it won't otherwise make sense.

jamietanna: Raised https://github.com/renovatebot/renovate/pull/38975

PR #38987, opened 2025-10-31
feat: Cleanup mutexes before each run

zharinov:
## Changes

I'm planning to use it for locking the same HTTP requests.
I want to avoid growing the size of mutex storage for the long-running processes.

Also, this PR disables locks from being used in tests.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/test/setup.ts
+++ b/test/setup.ts
@@ -31,3 +31,7 @@ Object.defineProperty(global, 'fixtures', { value: _fixtures });
 declare global {
   const fixtures: typeof _fixtures;
 }
+
+vi.mock('../lib/util/mutex', () => ({
+  acquireLock: () => Promise.resolve(() => undefined),
+}));
--- a/test/setup.ts
+++ b/test/setup.ts
@@ -33,5 +33,6 @@ declare global {
 }
 
 vi.mock('../lib/util/mutex', () => ({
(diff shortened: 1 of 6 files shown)


viceice (commented):
failing tests, otherwise LGTM

test/setup.ts
```diff
@@ -31,3 +31,8 @@ Object.defineProperty(global, 'fixtures', { value: _fixtures });
 declare global {
   const fixtures: typeof _fixtures;
 }
+
+vi.mock('../lib/util/mutex', () => ({
+  initMutexes: () => vi.fn(),
+  acquireLock: () => vi.fn().mockImplementation(() => () => undefined),
```
viceice, line 37: isn't undefined the default return value from vi.fn() mock?

zharinov: Yeah it's better to always return some dummy function, because we're acquiring locks here and there. It just maybe we don't want to have real mutexes in most tests.

RahulGautamSingh (approved):

test/setup.ts
viceice: 🤦‍♂️right , you're returning a callback. isn't that reset before each test, so it returns undefined again instead of the callback?

viceice (approved):

PR #39001, opened 2025-10-31
chore(minimumReleaseAgeBehaviour): reduce number of `WARN` log lines

breuerfelix:
## Changes

When setting minimumReleaseAgeBehaviour to timestamp-optional it is intended that packages that does not have a release timestamp will get updated. Writing a warning for every package to the console is spam for an intentional setting.

We try to keep warnings to a minimum. When this change was introduced, we had over 400 warnings while before we were below 10 per run. When using minimumReleaseAge, this log messages makes it unable to "control" your warnings since you are not able to "fix" this warning unless you disable minimumReleaseAge completely or use timestamp-required, which would result in not updating these packages at all.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/workers/repository/process/lookup/filter-checks.ts
+++ b/lib/workers/repository/process/lookup/filter-checks.ts
@@ -161,7 +161,7 @@ export async function filterInternalChecks(
     }
 
     if (candidateVersionsWithoutReleaseTimestamp['timestamp-optional'].length) {
-      logger.warn(
+      logger.debug(
         {
           depName,
           versions:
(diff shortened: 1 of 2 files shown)


jamietanna:
Interesting, and appreciated feedback - I think when we added this, we weren't aware exactly how many logs this may lead to

I think I'd prefer to keep a `WARN` log in here, but only once - then these can remain at `DEBUG`.

https://docs.renovatebot.com/configuration-options/#loglevelremap is an alternative option here, that would allow for user-controlled means to tune this and allow you to downgrade in the meantime

> Writing a warning for every package to the console is spam for an intentional setting.

IMO, it's not necessarily an "intentional" setting - right now, it's the default, and this behaviour is not something everyone is aware of

In #38841 (v42) it'll no longer be the default, so will be more exceptional that someone intentionally sets this

jamietanna:
@breuerfelix any concerns with the re-addition of a single `WARN` log line?

jamietanna (approved):
Let's try this and see how it goes 👍

viceice (approved):

PR #39065, opened 2025-11-04
docs(prCreation): clarify `not-pending` does not create a branch

jamietanna:
## Context

Please select one of the below:

- [ ] This closes an existing Issue: #
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [x] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #39066, opened 2025-11-04
fix(presets): branches shouldn't be created until `minimumReleaseAge` passes

jamietanna:
## Changes

As found while writing the documentation for the functionality, it's
possible for a compromised package to have (branches and) PRs created,
before the `minimumReleaseAge` has been met, as the default
`prCreation=immediate` would raise it immediately.

Given we're attempting to avoid malicious dependency updates, we should
not create a branch until the window has passed (managed by
`internalChecksFilter`), to avoid any possible i.e. automated build/CI
infrastructure compromise.

> [!NOTE]
> I could see this being a breaking change - fine if so!

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/config/presets/internal/security.ts
+++ b/lib/config/presets/internal/security.ts
@@ -41,6 +41,8 @@ export const presets: Record<string, Preset> = {
       'Wait until the npm package is three days old before raising the update, this prevents npm unpublishing a package you already upgraded to.',
     npm: {
       minimumReleaseAge: '3 days',
+      internalChecksFilter: 'strict',
+      prCreation: 'not-pending',
     },
   },
 };


jamietanna (commented):
Need to validate with https://github.com/renovatebot/renovate/pull/39065 further

lib/config/presets/internal/security.ts
```diff
@@ -41,6 +41,8 @@ export const presets: Record<string, Preset> = {
       'Wait until the npm package is three days old before raising the update, this prevents npm unpublishing a package you already upgraded to.',
     npm: {
       minimumReleaseAge: '3 days',
+      internalChecksFilter: 'strict',
+      prCreation: 'not-pending',
```
jamietanna, line 45: I'm not actually sure if this is needed here - given `strict`, we shouldn't get any branches created if they're pending?

jamietanna: It is - otherwise they'll get raised when `pending`

viceice (approved):
`internalChecksFilter=strict` is the default

jamietanna:
Correct - I wanted to be super explicit here to make sure it's set to a reasonable value (and be self-documenting for folks who want to follow similar for themselves)

RahulGautamSingh:
This might be wrong. Instead of, achieving the intended goal of making `prCreation=not-pending` till 3 days after release, it will mark all `npm` manager extracted packages as `prCreation=not-pending` regardless of release age.

jamietanna:
This is related to https://github.com/renovatebot/renovate/pull/39242 right?

> as `prCreation=not-pending` regardless of release age

Wouldn't that mean that if a release is still `pending`, then a PR wouldn't be created (as expected)

However, I'd expect that once it is no longer pending, Renovate would be able to raise it 🤔

RahulGautamSingh:
Yes, I think it doesn't effect in the way I was thinking. Thanks

PR #39177, opened 2025-11-08
feat: add configValidationError option to log config validation errors as errors instead of warnings

reduckted:
## Changes

Added an option called `configValidationError`. When this is set to true, config validation errors will be logged as errors instead of warnings. As a result, if there are config validation errors, Renovate will exit with a non-zero exit code.

## Context

- [x] This closes an existing Issue, Closes: #24891

## AI assistance disclosure

The public repository: N/A (repository was private)

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -798,6 +798,10 @@ After we changed the [`baseBranchPatterns`](#basebranchpatterns) feature, the Re
 
 For more details, read the [config migration documentation](./config-migration.md).
 
+## configValidationError
+
+If enabled, config validation errors will be reported as errors instead of warnings, and Renovate will exit with a non-zero exit code.
+
 ## configWarningReuseIssue
 
 If no existing issue is found, Renovate's default behavior is to create a new Config Warning issue.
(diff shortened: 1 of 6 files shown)


jamietanna (approved):
This makes a lot of sense to do - thanks! I'm already going to start using it 🚀

joshuaspence:
I turned this on and it worked well except it would be great if it only affected global config and not repo config. Config validation errors for repos would be better suited on the dependency dashboard, I think

jamietanna:
> if it only affected global config and not repo config. Config validation errors for repos would be better suited on the dependency dashboard, I think

What do you mean, sorry? What's the behaviour change you're finding doesn't work best here?

joshuaspence:
> > if it only affected global config and not repo config. Config validation errors for repos would be better suited on the dependency dashboard, I think
> 
> What do you mean, sorry? What's the behaviour change you're finding doesn't work best here?

So with this change Renovate would fail (exit with an error) if it processed a repo containing this, for example:

```json
{
  "extends": ["local>foo/bar"]
}
```

Where the `foo/bar` repo doesn't exist.

joshuaspence:
Actually looking at the actual code, I think what I observed is the expected behavior. I would advise updating the documentation to make this more clear.

jamietanna:
@joshuaspence (not being facetious) - we welcome a contribution to the docs! Or if you're able to share where you're thinking we could improve, we can get a PR raised too

PR #39230, opened 2025-11-11
fix(copier): broken log message 

RahulGautamSingh:
## Changes
- Use `isNullOrUndefined` check instead of `?`(optional chaining) which can let some edge cases in

## Context

- [x] This closes an existing Issue: #39228 

## AI assistance disclosure


astellingwerf:
I can't help but feel icky that we can change this code back and forth without breaking any test. That also makes it hard for me to grasp why out of the 8 cases in #38897 seemingly matching the pattern, only 6 need a fix.

RahulGautamSingh:
> I can't help but feel icky that we can change this code back and forth without breaking any test. That also makes it hard for me to grasp why out of the 8 cases in #38897 seemingly matching the pattern, only 6 need a fix.

Agreed. I will add tests for these cases as well. 

Regarding 6/8 cases needing the change. The other 2 cases are `&&` expressions and not `||` and I verified that they pass the checks.

jamietanna:
Can we add misisng tests? Then we can check that things are broken before this PR's changes, then when applying this change, they should be fixed?

lib/modules/manager/copier/artifacts.ts
```diff
@@ -56,7 +57,7 @@ export async function updateArtifacts({
   updatedDeps,
   config,
 }: UpdateArtifact): Promise<UpdateArtifactsResult[] | null> {
-  if (updatedDeps?.length !== 1) {
+  if (isNullOrUndefined(updatedDeps) || updatedDeps.length !== 1) {
```
RahulGautamSingh:
```suggestion
  if (updatedDeps?.length !== 1) {
```

This is fine as the logic behaves correctly when `updateDeps` is `null | undefined`. But, the debug message inside the if-check breaks, so need to fix that.

lib/modules/manager/homebrew/extract.ts
RahulGautamSingh, line 1: This can be reverted. If the `sha256=null|undefined` the logic is working fine.

There's also a test to confirm this already present.

lib/modules/platform/forgejo/pr-cache.ts
RahulGautamSingh, line 1: Without a defined author, `forgejo` & `gitea` platform will throw error in the very beginning so this change cannot be tested for changes in those platform's code.

lib/modules/versioning/nuget/index.ts
RahulGautamSingh, line 1: The existing expression is not broken as the value will be compared to a constant value.

RahulGautamSingh:
@astellingwerf Could you please review again. I have added tests where possible and have added a comment on the other file where the existing logic is not breaking or a test cannot be added.

lib/modules/manager/copier/artifacts.ts
viceice: agreed, use the optional chaining here

viceice (commented):
otherwise LGTM

lib/modules/versioning/nuget/index.ts
viceice: so this change isn't required? please leave as is then

lib/modules/manager/homebrew/extract.ts
viceice: please revert

viceice (commented):
see comments

astellingwerf (approved):
Thanks, @RahulGautamSingh!

viceice (approved):

PR #39264, opened 2025-11-12
test(allowedEnv): document use of regex

jamietanna:
## Changes

It looks like we may have copied this from the `allowedHeaders` docs,
where `allowedHeaders` does support a regex, but we do not with
`allowedEnv`.

We can also make this clear with a test to capture this behaviour.
## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/config/validation.spec.ts
+++ b/lib/config/validation.spec.ts
@@ -1604,6 +1604,21 @@ describe('config/validation', () => {
       ]);
     });
 
+    it('validates env against the allowedEnv regex', async () => {
+      const config = {
+        env: {
+          SOME_VAR: 'SOME_VALUE',
+        },
+        allowedEnv: ['/^SOME.*/'],
+      };
+      const { warnings, errors } = await configValidation.validateConfig(
+        'global',
+        config,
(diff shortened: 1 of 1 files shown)


jamietanna:
In the future, we could/should add support - but for now, let's get it documented that it's unsupported

RahulGautamSingh:
But, in the definition in `config/options/index.ts` ... `patternMatch` is `true`.  And, config validation seems to allow it too 🤔

jamietanna:
Correct - this was me not correctly using the regex syntax 🫣

lib/config/validation.spec.ts
```diff
@@ -1604,6 +1604,21 @@ describe('config/validation', () => {
       ]);
     });
 
+    it('validates env when regex', async () => {
```
RahulGautamSingh:
```suggestion
    it('validates env against allowedEnv regex', async () => {
```

RahulGautamSingh (changes requested):

viceice (approved):

PR #39346, opened 2025-11-17
fix(manager/kustomize): prevent setting replaceString for OCI chart extraction with registry aliases

tiagodcc:
## Changes

When using the kustomize manager, Helm charts may be referenced using OCI registries. However, if registry aliases are used, `replaceString` is set by the `dockerfile` manager, leading to a failure while replacing the version key of the existing context. In this PR, the object key `replaceString` is deleted using a destruction.

The issue has already been described in the following discussion: https://github.com/renovatebot/renovate/discussions/37638

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/manager/kustomize/extract.ts
+++ b/lib/modules/manager/kustomize/extract.ts
@@ -164,9 +164,9 @@ export function extractHelmChart(
       false,
       aliases,
     );
-    const { replaceString, ...rest } = dep;
+    delete dep.replaceString;
     return {
-      ...rest,
+      ...dep,
       depName: helmChart.name,
       // https://github.com/helm/helm/issues/10312
       // https://github.com/helm/helm/issues/10678
--- a/lib/modules/manager/kustomize/extract.ts
+++ b/lib/modules/manager/kustomize/extract.ts
@@ -164,8 +164,9 @@ export function extractHelmChart(
       false,
       aliases,
     );
+    const { replaceString, ...rest } = dep;
     return {
-      ...dep,
+      ...rest,
       depName: helmChart.name,
       // https://github.com/helm/helm/issues/10312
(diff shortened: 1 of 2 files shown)


tiagodcc:
I had to force push to correct the author information of the commit.

This change only addresses the issue in the `kustomize` manager. However, I suspect a more central fix might be appropriate in the dockerfile manager, specifically here: https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/dockerfile/extract.ts#L186

It appears the specifyReplaceString parameter is being ignored in that logic. I was hesitant to make the change there as I wasn't completely sure about potential side effects for other managers.

What are your thoughts? If you agree the dockerfile manager is the better approach, I'm happy to update this PR.

RahulGautamSingh:
I get the issue, but this does not seems to be the correct solution. Can you please create a minimal reproduction so we can discuss the best solution?

tiagodcc:
Thanks for your fast reply! I created a minimal reproduction here: https://github.com/tiagodcc/kustomize-oci-helm

lib/modules/manager/kustomize/extract.ts
```diff
@@ -164,8 +164,9 @@ export function extractHelmChart(
       false,
       aliases,
     );
+    const { replaceString, ...rest } = dep;
```
viceice:
```suggestion
    delete dep.replaceString;
```
simplify

lib/modules/manager/kustomize/extract.ts
```diff
@@ -164,8 +164,9 @@ export function extractHelmChart(
...
     return {
-      ...dep,
+      ...rest,
```
viceice:
```suggestion
      ...dep,
```

viceice (approved):

PR #39417, opened 2025-11-19
fix(github/vulnerabilityAlerts): don't error on unknown ecosystems

### LINKED ISSUE #39270: fix(github/vulnerabilityAlerts): don't error on unknown ecosystems

As noted in #38503, when receiving a vulnerability alert for the GitHub Actions ecosystem (`actions`), Renovate encounters a Zod error.

We should correctly handle unknown ecosystem(s) by reporting via a debug log (including what ecosystem we've detected), but not failing to parse alerts.

mskanth972:
## Changes

This PR fixes the schema parsing error that occurs when repositories contain GitHub Actions-related advisories.
Renovate previously failed to parse alerts because the upstream schema did not recognize the new ecosystem type returned by the GitHub API.
This change updates the parsing code accordingly.

## Context

- [x] This closes an existing Issue, Closes: #39270 

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/platform/github/schema.spec.ts
+++ b/lib/modules/platform/github/schema.spec.ts
@@ -1,5 +1,4 @@
-import { GithubContentResponse } from './schema';
-
+import { GithubContentResponse, GithubVulnerabilityAlert } from './schema';
 describe('modules/platform/github/schema', () => {
   it('should be parse directory response', () => {
     const { error } = GithubContentResponse.safeParse([
@@ -105,5 +104,38 @@ describe('modules/platform/github/schema', () => {
       },
     });
     expect(error).toBeUndefined();
+  });
+
+  it('should skip vulnerability alerts with unsupported ecosystems', () => {
+    const result = GithubVulnerabilityAlert.parse([
+      {
+        dismissed_reason: null,
+        security_advisory: {
+          description: 'Test advisory',
+          identifiers: [{ type: 'CVE', value: 'CVE-2024-1234' }],
+        },
+        security_vulnerability: {
+          first_patched_version: { identifier: '1.0.0' },
+          package: { ecosystem: 'dotnet', name: 'test-package' },
+          vulnerable_version_range: '< 1.0.0',
+        },
+        dependency: { manifest_path: 'package.json' },
+      },
+      {
+        dismissed_reason: null,
+        security_advisory: {
+          description: 'Test advisory',
+          identifiers: [{ type: 'CVE', value: 'CVE-2024-5678' }],
+        },
+        security_vulnerability: {
+          first_patched_version: { identifier: '2.0.0' },
+          package: { ecosystem: 'npm', name: 'valid-package' },
+          vulnerable_version_range: '< 2.0.0',
+        },
+        dependency: { manifest_path: 'package.json' },
+      },
+    ]);
+    expect(result).toHaveLength(1);
+    expect(result[0].security_vulnerability?.package.ecosystem).toBe('npm');
   });
 });
--- a/lib/modules/platform/github/schema.ts
+++ b/lib/modules/platform/github/schema.ts
@@ -3,16 +3,24 @@ import { logger } from '../../../logger';
 import { LooseArray } from '../../../util/schema-utils';
 
 const Package = z.object({
-  ecosystem: z.union([
-    z.literal('maven'),
-    z.literal('npm'),
-    z.literal('nuget'),
-    z.literal('pip'),
-    z.literal('rubygems'),
-    z.literal('rust'),
-    z.literal('composer'),
-    z.literal('go'),
-  ]),
+  ecosystem: z
+    .union([
+      z.literal('maven'),
+      z.literal('npm'),
+      z.literal('nuget'),
+      z.literal('pip'),
+      z.literal('rubygems'),
+      z.literal('rust'),
+      z.literal('composer'),
+      z.literal('go'),
+    ])
+    .catch((ctx) => {
+      logger.debug(
+        { ecosystem: ctx.input },
+        'Skipping vulnerability alert with unsupported ecosystem',
+      );
+      return undefined as any;
+    }),
   name: z.string(),
 });
 
@@ -53,6 +61,8 @@ export const GithubVulnerabilityAlert = LooseArray(
       );
     },
   },
+).transform((alerts) =>
+  alerts.filter((alert) => alert.security_vulnerability?.package?.ecosystem),
 );
 export type GithubVulnerabilityAlert = z.infer<typeof GithubVulnerabilityAlert>;
 


jamietanna:
Thanks very much!

Before we merge this, I'd love for us to tackle https://github.com/renovatebot/renovate/issues/39270 first, so we can make sure we don't run into this issue again in the future

Are you able to take a look at that one?

mskanth972:
Hi @jamietanna, surely I can work on that. Looks like a valid point to avoid issues in future.

mskanth972:
Tested with the new code and now it just logs the unknown ecosystem instead of throwing error

```
$ LOG_LEVEL=debug pnpm start | grep "Unknown vulnerability ecosystem detected" -A 20 -B 20

DEBUG: migrateAndValidate()
DEBUG: No config migration necessary
DEBUG: http cache: saving config file (etag=...)
DEBUG: http cache: saving default config (etag=...)
DEBUG: Flattening nested packageRules
DEBUG: Resolved config needs migrating
DEBUG: Found repo ignorePaths
       "ignorePaths": [
         "**/node_modules/**",
         "**/bower_components/**",
         "**/vendor/**",
         "**/examples/**",
         "**/__tests__/**",
         "**/test/**",
         "**/tests/**",
         "**/__fixtures__/**"
       ]
DEBUG: http cache: saving vulnerability alerts response (etag=...)
DEBUG: Unknown vulnerability ecosystem detected
       "ecosystem": "actions"
DEBUG: Unknown vulnerability ecosystem detected
       "ecosystem": "actions"
DEBUG: GitHub vulnerability details
       "alerts": {
         "actions/SonarSource/sonarqube-scan-action": {
           ">= 4.0.0, <= 5.3.0": "5.3.1",
           ">= 4.0.0, < 6.0.0": "6.0.0"
         },
         "npm/happy-dom": {"< 20.0.0": "20.0.0", "< 20.0.2": "20.0.2"},
         "npm/js-yaml": {">= 4.0.0, < 4.1.1": "4.1.1"},
         "npm/glob": {">= 11.0.0, < 11.1.0": "11.1.0", ">= 10.2.0, < 10.5.0": "10.5.0"}
       }
DEBUG: alert package rules
       "alertPackageRules": [
         {
           "matchDatasources": ["undefined"],
           "matchPackageNames": ["SonarSource/sonarqube-scan-action"],
           "matchFileNames": [".github/workflows/main.yml"],
           "matchCurrentVersion": "< 6.0.0",
           "vulnerabilityFixVersion": "6.0.0",
           "prBodyNotes": [
...

```

lib/types/vulnerability-alert.ts
```diff
@@ -1,13 +1,5 @@
...
-  ecosystem:
-    | 'maven'
-    | 'npm'
-    | 'nuget'
-    | 'pip'
-    | 'rubygems'
-    | 'rust'
-    | 'composer'
-    | 'go';
+  ecosystem: string;
```
jamietanna: Please revert this, we want to retain strong types

jamietanna (changes requested):
Thanks for the change - does the PR title/description need to change here? 

We should have one PR to fix it (to allow unknown types) and one to add Actions as a follow-up

lib/types/vulnerability-alert.ts
mskanth972: Done

mskanth972:
> Thanks for the change - does the PR title/description need to change here?
> 
> We should have one PR to fix it (to allow unknown types) and one to add Actions as a follow-up

Yep, fixed it and also will raise the other PR once this is merged.

lib/modules/platform/github/index.ts
```diff
@@ -2104,6 +2104,22 @@ export async function getVulnerabilityAlerts(): Promise<VulnerabilityAlert[]> {
           vulnerable_version_range: vulnerableVersionRange,
           first_patched_version: firstPatchedVersion,
         } = alert.security_vulnerability;
+        const knownEcosystems = new Set([
```
jamietanna: I'd recommend pulling this out into a `const` outside of this

jamietanna (commented):
Looking good - mind addressing PR checks?

Also, can we try and add a test to validate that this works?

mskanth972:
> Looking good - mind addressing PR checks?
> 
> Also, can we try and add a test to validate that this works?

Sure, I can add a UT

Churro:
I wonder if this approach is really best suited for the actual problem. To my understanding the Zod parsing error happens because renovate encounters an unknown ecosystem, e.g., `actions` that is not part of this list: https://github.com/renovatebot/renovate/blob/295c22ddb5f6fd2f4b69dd938e87a77019940b8e/lib/modules/platform/github/schema.ts#L5-L15

IMHO it would be smarter to loosen the schema, i.e., not hard-code every supported vulnerability ecosystem, instead of adding yet another check for known ecosystems.

Unrelated to this PR, upon further processing, found GH vulnerabilties are then again translated into custom interfaces with a list of hardcoded ecosystem types: https://github.com/renovatebot/renovate/blob/295c22ddb5f6fd2f4b69dd938e87a77019940b8e/lib/types/vulnerability-alert.ts#L2-L11. 

I'd advocate to hardcode the list of ecosystems only once and when actually parsing the alerts ([here](https://github.com/renovatebot/renovate/blob/295c22ddb5f6fd2f4b69dd938e87a77019940b8e/lib/workers/repository/init/vulnerability.ts#L55)) ensure that renovate can successfully map the ecosystem to a versioning scheme, otherwise drop it.

mskanth972:
> I wonder if this approach is really best suited for the actual problem. To my understanding the Zod parsing error happens because renovate encounters an unknown ecosystem, e.g., `actions` that is not part of this list:
> 
> https://github.com/renovatebot/renovate/blob/295c22ddb5f6fd2f4b69dd938e87a77019940b8e/lib/modules/platform/github/schema.ts#L5-L15
> 
> IMHO it would be smarter to loosen the schema, i.e., not hard-code every supported vulnerability ecosystem, instead of adding yet another check for known ecosystems.
> 
> Unrelated to this PR, upon further processing, found GH vulnerabilties are then again translated into custom interfaces with a list of hardcoded ecosystem types:
> 
> https://github.com/renovatebot/renovate/blob/295c22ddb5f6fd2f4b69dd938e87a77019940b8e/lib/types/vulnerability-alert.ts#L2-L11
> 
> .
> I'd advocate to hardcode the list of ecosystems only once and when actually parsing the alerts ([here](https://github.com/renovatebot/renovate/blob/295c22ddb5f6fd2f4b69dd938e87a77019940b8e/lib/workers/repository/init/vulnerability.ts#L55)) ensure that renovate can successfully map the ecosystem to a versioning scheme, otherwise drop it.

Thanks for the suggestion, your approach makes a lot of sense. Centralizing the ecosystem list and validating unsupported ecosystems earlier keeps things cleaner than adding more checks later in the flow.

@jamietanna Please let me know if you're aligned with this direction, I already have the new commit drafted, and I can add test once we agree this is the preferred approach.

lib/modules/platform/github/index.ts
```diff
@@ -2104,6 +2105,13 @@ export async function getVulnerabilityAlerts(): Promise<VulnerabilityAlert[]> {
           vulnerable_version_range: vulnerableVersionRange,
           first_patched_version: firstPatchedVersion,
         } = alert.security_vulnerability;
+        if (!supportedEcosystems.has(ecosystem)) {
```
Churro: I don't think it makes sense to check the ecosystem twice. IMHO this check here is superfluous. The critical part for renovate begins when mapping it to a (supported) versioning scheme

lib/constants/supported-ecosystems.ts
```diff
@@ -0,0 +1,10 @@
+export const supportedEcosystems = new Set([
```
Churro: The scope of `supportedEcosystems` is limited to vulnerabilities only. IMHO defining this `const` in the top section of `vulnerability.ts` is more reasonable than decoupling it entirely into a separate file. There's neither practical benefit, nor is it needed elsewhere.

Churro (changes requested):

lib/modules/platform/github/index.ts
mskanth972: Removed this second check

lib/constants/supported-ecosystems.ts
mskanth972: Updated.

Churro:
Apologies for the confusion but when looking at the diff with a bit more focus now, I realized that @jamietanna asked for `ecosystem` to be strongly-typed, and I've inadvertently suggested quite the opposite.

I believe there's a tidier approach without extra `supportedEcosystems.has(ecosystem)` check that catches unsupported ecosystems gracefully already during input validation in [schema.ts](https://github.com/renovatebot/renovate/blob/5a212d215d45feca1888b02d743dc2acc26e3f8d/lib/modules/platform/github/schema.ts). E.g., by using zod-native capabilties this could look like this:
```ts
const Package = z.object({
  ecosystem: z
    .union([
      z.literal('maven'),
      z.literal('npm'),
      z.literal('nuget'),
      z.literal('pip'),
      z.literal('rubygems'),
      z.literal('rust'),
      z.literal('composer'),
      z.literal('go'),
    ])
    .catch((ctx) => {
      logger.debug(
        { ecosystem: ctx.input },
        'Skipping vulnerability alert with unsupported ecosystem',
      );
      return undefined as any;
    }),
  name: z.string(),
});
```

and later on just filter it:

```ts
export const GithubVulnerabilityAlert = LooseArray(
  z.object({
    dismissed_reason: z.string().nullish(),
    security_advisory: SecurityAdvisory,
    security_vulnerability: SecurityVulnerability,
    dependency: z.object({
      manifest_path: z.string(),
    }),
  }),
  {
    /* v8 ignore next */
    onError: ({ error }) => {
      logger.debug(
        { error },
        'Vulnerability Alert: Failed to parse some alerts',
      );
    },
  },
).transform((alerts) =>
  alerts.filter((alert) => alert.security_vulnerability?.package?.ecosystem),
);
```

Doing refactoring to avoid redefining ecosystems multiple times still makes sense IMHO but this is of course not the purpose of this PR.

viceice:
@Churro i'm in favor of your idea 👍

jamietanna:
Apologies for the delay, but I agree that your suggested approach is much nicer, thanks for weighing in!

mskanth972:
Thanks everyone for the detailed reviews and the thoughtful discussion. @Churro, @jamietanna I’ve updated the branch accordingly and added tests.

lib/modules/platform/github/schema.spec.ts
```diff
@@ -106,4 +107,91 @@ describe('modules/platform/github/schema', () => {
     });
     expect(error).toBeUndefined();
   });
+
```
Churro: AI generated? The tests replicate the same situation redundantly.

One test is enough if it shows that unknown ecosystems are successfully filtered, e.g
```ts
it('should skip vulnerability alerts with unsupported ecosystems', () => {
    const result = GithubVulnerabilityAlert.parse([
      {
        dismissed_reason: null,
        security_advisory: {
          description: 'Test advisory',
          identifiers: [{ type: 'CVE', value: 'CVE-2024-1234' }],
        },
        security_vulnerability: {
          first_patched_version: { identifier: '1.0.0' },
          package: { ecosystem: 'dotnet', name: 'test-package' },
          vulnerable_version_range: '< 1.0.0',
        },
        dependency: { manifest_path: 'package.json' },
      },
      {
        dismissed_reason: null,
        security_advisory: {
          description: 'Test advisory',
          identifiers: [{ type: 'CVE', value: 'CVE-2024-5678' }],
        },
        security_vulnerability: {
          first_patched_version: { identifier: '2.0.0' },
          package: { ecosystem: 'npm', name: 'valid-package' },
          vulnerable_version_range: '< 2.0.0',
        },
        dependency: { manifest_path: 'package.json' },
      },
    ]);

    expect(result).toHaveLength(1);
    expect(result[0].security_vulnerability?.package.ecosystem).toBe('npm');
  });
```

Churro (changes requested):

lib/modules/platform/github/schema.spec.ts
mskanth972: Yeah I used AI assistance to scaffold the initial tests, and that ended up producing more cases which I thought what was actually needed. But anyways let me add a single focused test is sufficient to validate that unsupported ecosystems are filtered out.

mskanth972: Fixed it.

lib/modules/platform/github/schema.spec.ts
```diff
@@ -82,7 +81,6 @@ describe('modules/platform/github/schema', () => {
     ]);
     expect(error).toBeUndefined();
   });
-
```
Churro, line 85: revert

lib/modules/platform/github/schema.spec.ts
```diff
@@ -106,4 +104,42 @@ describe('modules/platform/github/schema', () => {
     });
     expect(error).toBeUndefined();
   });
+
+  describe('GithubVulnerabilityAlert schema', () => {
+    beforeEach(() => {
+      vi.clearAllMocks();
```
Churro: Which mocks need clearing here?

Churro (changes requested):

lib/modules/platform/github/schema.spec.ts
mskanth972: I’ve updated the test so it now lives under a single test suite. Since the test already clearly states its purpose, it didn’t seem necessary to nest it under an additional module-level describe block.

lib/modules/platform/github/schema.spec.ts
mskanth972: Cleaned it.

lib/modules/platform/github/schema.spec.ts
Churro: That's fine but there's no need to delete lines that are completely unrelated to these efforts, including this.

It might be your personal taste not to use newlines but if you look around the source code of renovate, `it(...)` are usually separated by empty lines, so please follow this stylistic concept for code homogeneity.

Churro (changes requested):

lib/modules/platform/github/schema.spec.ts
mskanth972: Thanks for pointing that out. I didn’t intentionally remove the blank line, the file originally didn’t show any blanks and followed the same.
But you’re right that other schema test files in Renovate include a blank line after describe(...), so will updated this file to follow the same convention for consistency.

mskanth972: Oops, sorry thats my VS Code issue which removed all the existing blanks. But anyways fixed it.

Churro (approved):

jamietanna (commented):
Thanks for the work on the refactor here

I'm noticing that reverting the changes to `lib/modules/platform/github/schema.ts` doesn't make the tests fail - trying to see what we can do to tweak that

Edit: on second look, this is not failing because it's the `logger.debug` that shouldn't be happening after this change

jamietanna (approved):
I can confirm that given the tests in this PR and https://github.com/renovatebot/renovate/discussions/38503#discussioncomment-14949008 we no longer see a Zod error - we now correctly handle this (and return `undefined`)

PR #37956, opened 2025-09-11
fix(config/options): convert `ignorePrAuthor` to global only

RahulGautamSingh:
## Context

Please select one of the below:

- [x] This closes an existing Issue: #37947 
- [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [x] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #39436, opened 2025-11-20
fix: do not treat ignorePrAuthor as a globalOnly option 

Iridias:
## Changes

Remove `ignorePrAuthor` from the `GlobalConfig.OPTIONS` list and remove the `globalOnly` flag.

Otherwise, in `lib/workers/repository/index.ts:62` the `let config = GlobalConfig.set(` would result in a config without this setting.
And further down the line, in `initRepo`, `initApis` and finally `platform.initRepo(config)`, the `ignorePrAuthor` would always be `undefined`! 
And this would result in `GitlabPrCache` (and probably the other platform pr-caches as well) to always ignore this setting, rending it utterly useless - as it's that exact place, where it's needed.

Related discussion: https://github.com/renovatebot/renovate/discussions/38076 

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2424,8 +2424,6 @@ You should only want to enable this if you are changing the bot account (e.g. fr
 
 Setting this field to `true` in Github or GitLab will also mean that all Issues will be fetched instead of only those by the bot itself.
 
-
-
 ## ignorePresets
 
 Use this if you are extending a complex preset but don't want to use every "sub preset" that it includes.
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -2414,6 +2414,18 @@ If you are extending from the popular `config:recommended` preset then it adds i
 Set this to `true` if running plugins causes problems.
 Applicable for Composer only for now.
 
+## ignorePrAuthor
+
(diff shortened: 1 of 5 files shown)


jamietanna:
Apologies, I think a recent change we've made to the CLA Assistant configuration requires re-signing - the contents is the same, but I'll let you read it and re-sign when you get a chance

viceice (commented):
this is not the right fix. `ignorePrAuthor` is not intended to be overridable on repo config.
so the other places need to be fixed to use right global option.

viceice (changes requested):
forgot to request changes

jamietanna:
Yes that's correct - we recently made a change (https://github.com/renovatebot/renovate/pull/37956) to correct the fact that it should have only been a global option, but was mis-documented as such - good point Michael :)

viceice:
> Yes that's correct - we recently made a change (#37956) to correct the fact that it should have only been a global option, but was mis-documented as such - good point Michael :)

Sadly that fix introduced a new bug, which i'm fixing now

jamietanna:
Thanks for the contribution - we're going with https://github.com/renovatebot/renovate/pull/39578 in this case!

PR #39489, opened 2025-11-24
feat: Efficient package cache implementation

zharinov:
## Changes

Refactored `packageCache` from functional exports to a class-based architecture with `PackageCache` class and pluggable backends.
- Replaced functional file cache implementation with `PackageCacheBase` abstraction and `File`, `Redis`, and `Sqlite` implementations in `lib/util/cache/package/impl/`.
- Updated `lib/util/cache/package/index.ts` to export a singleton `packageCache` instance.
- Updated the `cache` decorator to use the singleton instance and optimized the stale-while-revalidate logic.
- Migrated tests to use the new `packageCache` singleton, removing unnecessary mocks.

The important change is the usage of memory cache which is being used even if no backend was chosen. It helps to avoid excessive requests to servers from Renovate instances being run from CI. It's not as good as persistent cache, but at least we do our best to minimize requests during single run.

## Context

## AI assistance disclosure

The public repository: <URL>


lib/util/cache/memory/index.ts
```diff
@@ -21,10 +21,7 @@ export function set(key: string, value: unknown): void {
 export function cleanDatasourceKeys(): void {
   if (memCache) {
     for (const key of Object.keys(memCache)) {
-      if (
-        key.startsWith('datasource-mem:pkg-fetch:') ||
-        key.startsWith('datasource-mem:releases:')
-      ) {
+      if (key.startsWith('datasource-mem:releases:')) {
```
viceice, line 24: so `datasource-mem:pkg-fetch:` namespace is no longer used? 🤔

lib/util/cache/package/impl/file.ts
```diff
@@ -0,0 +1,120 @@
...
+        this.getKey(namespace, key),
+      );
+      const cachedValue = JSON.parse(res.data.toString());
+      if (!cachedValue) {
+        return undefined;
+      }
+
+      if (DateTime.local() < DateTime.fromISO(cachedValue.expiry)) {
+        logger.trace({ namespace, key }, 'Returning cached value');
+        // istanbul ignore if
```
viceice: doesn't work 😉

zharinov:
_(extracted from README.md for convenience)_

# Package Cache

Central caching mechanism for Renovate datasources and lookups. Implements a two-layer architecture:

1. **L1:** In-memory `Map` (per-process).
2. **L2:** Persistent storage (File, Redis, or SQLite).

## TTL Handling Scope

Soft/Hard TTL logic is not isolated to the `@cache` decorator.
The HTTP cache layer in `PackageHttpCacheProvider` also resolves and applies soft/hard TTL independently (Cache-Control headers, etc.).
In core caching, only TTL overrides are applied; use `setWithRawTtl` to bypass this logic.

## Flow Architecture

The `@cache` decorator orchestrates the retrieval flow.

### Scenario 1: L1 hit (memory)

The fastest path - data is already in memory.

```mermaid
sequenceDiagram
    participant C as Caller
    participant D as Decorated method <br/>(via @cache)
    participant L1 as Memory <br/>(L1)

    C->>D: Call decorated method
    D->>L1: Check key
    L1->>D: L1 hit
    D->>C: Return cached value
```

### Scenario 2: L2 hit (fresh in storage)

Data is not in memory but exists in backend storage and is still fresh (within Soft TTL).

```mermaid
sequenceDiagram
    participant C as Caller
    participant D as Decorated method <br/>(via @cache)
    participant X as Mutex
    participant L1 as Memory <br/>(L1)
    participant L2 as Backend <br/>(L2)

    C->>D: Call decorated method
    D->>L1: Check key
    L1->>D: L1 miss

    D-->>X: Request lock (based on namespace + key)
    rect rgba(150, 100, 200, 0.15)
        X-->>D: Receive lock
        D->>L1: Check key again
        L1->>D: Still L1 miss

        D->>L2: Get cached record
        L2->>D: L2 hit
        D->>L1: Store in L1
        D-->>X: Release lock
    end
    D->>C: Return value
```

### Scenario 3: L2 hit (stale in storage)

**NOTE:** This is only works for `getReleases`/`getDigest` methods.
Data has expired Soft TTL but is within Hard TTL. Attempts to refresh but falls back if callback fails.

```mermaid
sequenceDiagram
    participant C as Caller
    participant D as Decorated method <br/>(via @cache)
    participant X as Mutex
    participant L1 as Memory <br/>(L1)
    participant L2 as Backend <br/>(L2)
    participant U as Original method <br/>(callback)

    C->>D: Call getReleases/getDigest
    D->>L1: Check key
    L1->>D: L1 miss

    D-->>X: Request lock
    rect rgba(150, 100, 200, 0.15)
        X-->>D: Receive lock
        D->>L1: Check key again
        L1->>D: Still L1 miss

        D->>L2: Get cached record
        L2->>D: L2 hit

        D->>U: Try to fetch fresh data
        alt Upstream Success
            U->>D: New value
            D->>L2: Update cache with new TTL
            D->>L1: Store in L1
        else Upstream Error
            U->>D: Error
            Note over D: Fallback to <br/>stale value
        end
        D-->>X: Release lock
    end
    D->>C: Return value
```

### Scenario 4: L2 miss

No cached data exists or Hard TTL has expired.

```mermaid
sequenceDiagram
    participant C as Caller
    participant D as Decorated method <br/>(via @cache)
    participant X as Mutex
    participant L1 as Memory <br/>(L1)
    participant L2 as Backend <br/>(L2)
    participant U as Original method <br/>(callback)

    C->>D: Call decorated method
    D->>L1: Check key
    L1->>D: L1 miss

    D-->>X: Request lock
    rect rgba(150, 100, 200, 0.15)
        X-->>D: Receive lock
        D->>L1: Check key again
        L1->>D: Still L1 miss

        D->>L2: Get cached record
        L2->>D: L2 miss

        D->>U: Fetch from upstream
        U->>D: Return value

        D->>L2: Store with Hard TTL
        D->>L1: Store in L1
        D-->>X: Release lock
    end
    D->>C: Return value
```

### Scenario 5: Private packages

When `cacheable()` returns false, only memory caching is used (which is reset after Renovate run). Setting `cachePrivatePackages` forces persistence of private packages (available to self-hosted users).

```mermaid
sequenceDiagram
    participant C as Caller
    participant D as Decorated method <br/>(via @cache)
    participant X as Mutex
    participant L1 as Memory <br/>(L1)
    participant U as Original method <br/>(callback)

    C->>D: Call decorated method
    D->>L1: Check key
    L1->>D: L1 miss

    D-->>X: Request lock
    rect rgba(150, 100, 200, 0.15)
        X-->>D: Receive lock
        D->>L1: Check key again
        L1->>D: Still L1 miss

        Note over D: Check cacheable()<br/>cacheable() = false<br/>cachePrivatePackages = false

        D->>U: Fetch from upstream (skip backend)
        U->>D: Return value

        D->>L1: Store in L1 only
        D-->>X: Release lock
    end
    D->>C: Return value
```

### Scenario 6: Concurrent Access (Race Protection)

Multiple concurrent calls for the same key - mutex ensures only one fetch. In this context, upstream is represented by both L2 and the original method (calling the actual external service).

```mermaid
sequenceDiagram
    participant C1 as Caller 1
    participant C2 as Caller 2
    participant D as Decorated method <br/>(via @cache)
    participant X as Mutex
    participant L1 as Memory <br/>(L1)
    participant U as Upstream

    C1->>D: Call method
    C2->>D: Call method (concurrent)

    D->>L1: Check key (C1)
    D->>L1: Check key (C2)
    L1->>D: L1 miss (both)

    D-->>X: C1 requests lock
    D-->>X: C2 requests lock
    Note over C2: C2 waits

    rect rgba(150, 100, 200, 0.15)
        X-->>D: C1 receives lock
        D->>L1: C1 checks again
        L1->>D: Still L1 miss
        D->>U: C1 fetches from upstream
        U->>D: Return value
        D->>L1: C1 stores in L1
        D-->>X: C1 releases lock
    end
    D->>C1: Return value

    rect rgba(150, 100, 200, 0.15)
        X-->>D: C2 receives lock
        D->>L1: C2 checks again
        L1->>D: L1 hit
        D-->>X: C2 releases lock
    end
    D->>C2: Return cached value
```

### Scenario 7: Decorated method throws

When callback fails and no cached data exists to fall back on.

```mermaid
sequenceDiagram
    participant C as Caller
    participant D as Decorated method <br/>(via @cache)
    participant X as Mutex
    participant L1 as Memory <br/>(L1)
    participant L2 as Backend <br/>(L2)
    participant U as Original method <br/>(callback)

    C->>D: Call decorated method
    D->>L1: Check key
    L1->>D: L1 miss

    D-->>X: Request lock
    rect rgba(150, 100, 200, 0.15)
        X-->>D: Receive lock
        D->>L1: Check key again
        L1->>D: Still L1 miss

        D->>L2: Get cached record
        L2->>D: L2 miss

        D->>U: Fetch from upstream
        U->>D: Error

        Note over D: No fallback available
        D-->>X: Release lock
    end
    D->>C: Throw error
```

### Scenario 8: Decorated method returns special values (`undefined` or `null`)

- **`undefined`**: Cached in Memory (L1) but **never persisted** to Backend (L2). Treated as a transient failure that should be retried on the next run.
- **`null`**: Cached in both L1 and L2 (if `cacheable` returns true). Treated as a valid result (e.g., "package not found").

The diagram below illustrates the `undefined` flow:

```mermaid
sequenceDiagram
    participant C as Caller
    participant D as Decorated method <br/>(via @cache)
    participant X as Mutex
    participant L1 as Memory <br/>(L1)
    participant U as Original method <br/>(callback)

    C->>D: Call decorated method
    D->>L1: Check key
    L1->>D: Miss

    D-->>X: Request lock
    rect rgba(150, 100, 200, 0.15)
        X-->>D: Receive lock
        D->>L1: Check key again
        L1->>D: Still miss

        D->>U: Execute callback
        U->>D: Return undefined

        D->>L1: Store in L1 only
        Note over D: undefined not persisted to L2
        D-->>X: Release lock
    end
    D->>C: Return undefined

    C->>D: Call again
    D->>L1: Check key
    L1->>D: L1 hit
    D->>C: Return undefined (no upstream call)
```

## Usage

Apply the decorator to class methods.

```typescript
import { cache } from '../../../util/cache/package';

class MyDatasource {
  @cache({
    namespace: 'datasource-my-source', // or function(arg) => string
    key: 'some-key', // or function(arg) => string
    ttlMinutes: 15, // Soft TTL
    // Optional. Default: () => true. Controls L2 persistence. L1 (Memory) always caches.
    cacheable: (result) => result !== null,
  })
  async getTags(pkgName: string): Promise<string[]> {
    // Expensive call to external source
  }
}
```

**Note:** The decorator prefixes L2 cache keys with `cache-decorator:`. Actual L2 key example: `datasource-my-source:cache-decorator:some-key`

### Persistence vs. Memory (`cacheable`)

The `cacheable` parameter **only controls persistence (L2)**.

- If `cacheable` returns `false`, the result is still stored in **L1 Memory**.
- Private packages are treated as non-cacheable by default unless forced by config.

| `cacheable()` Result | `cachePrivatePackages` Config | L1 (Memory) | L2 (Disk/Redis) |
| :------------------- | :---------------------------- | :---------- | :-------------- |
| `true`               | `false`                       | ✅          | ✅              |
| `false`              | `false`                       | ✅          | ❌              |
| `false`              | `true`                        | ✅          | ✅              |

## TTL Strategy (Soft vs. Hard)

Renovate uses a dual-TTL system to handle external source instability.

1. **Soft TTL (`ttlMinutes`):** Period where data is considered "fresh". Returned immediately without network calls.
2. **Hard TTL:** Period where data is physically retained on disk. Used for fallback if callback fails.

**Important:** The Stale-While-Revalidate (Hard TTL) logic is **only active** for methods named:

- `getReleases`
- `getDigest`

For all other methods, `Hard TTL = Soft TTL`.

### TTL Resolution Logic

```mermaid
graph TD
    A[Start] --> B{overriden via config?}
    B -- Yes --> C[Soft TTL = override]
    B -- No --> D[Soft TTL = decorator param]
    C --> E
    D --> E{Method Name is<br/>getReleases/getDigest?}
    E -- Yes --> F["Hard TTL = Max(Soft TTL, cacheHardTtlMinutes)"]
    E -- No --> G[Hard TTL = Soft TTL]
```

See also: [cacheHardTtlMinutes](https://docs.renovatebot.com/self-hosted-configuration/#cachehardttlminutes)

## Configuration Overrides (`cacheTtlOverride`)

Users can override Soft TTLs via `config.js`. The resolution uses **Longest Matching Pattern**.

| Pattern Type | Example                   | Priority         |
| :----------- | :------------------------ | :--------------- |
| Exact Match  | `datasource-npm`          | 1                |
| Glob         | `datasource-{npm,docker}` | 2 (Length based) |
| Regex        | `/^datasource-/`          | 3 (Length based) |
| Wildcard     | `*`                       | 4                |

## Backends

Backend is selected at startup based on environment and config (checked in order):

1. **Redis:** Enabled if `redisUrl` is configured (via env or CLI, see [docs](https://docs.renovatebot.com/self-hosted-configuration/#redisurl)).
2. **SQLite:** Enabled if `RENOVATE_X_SQLITE_PACKAGE_CACHE=true` and `cacheDir` is available.
3. **File:** Default fallback. Uses `cacache` with `gzip` in `cacheDir` (via env or CLI, see [docs](https://docs.renovatebot.com/self-hosted-configuration/#cachedir)).

lib/modules/datasource/deb/index.spec.ts
```diff
@@ -289,18 +293,14 @@ describe('modules/datasource/deb/index', () => {
...
-        mockHttpCalls(
-          'stable',
-          'non-free',
-          'amd64',
-          !!i,
-          fixturePackagesArchivePath,
-          fixturePackagesArchiveHash,
-        );
-      }
+      mockHttpCalls(
```
secustor, line 296: Where is this change coming from 🤔

lib/util/cache/package/impl/factory.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+vi.mock('../../../env', () => ({
+  getEnv: vi.fn(() => ({})),
+}));
+vi.mock('./file');
+vi.mock('./redis');
+vi.mock('./sqlite');
+
+describe('util/cache/package/impl/factory', () => {
+  beforeEach(() => {
+    vi.resetAllMocks();
```
viceice, line 17:
```suggestion
```
https://github.com/renovatebot/renovate/blob/744270eab3b8a05d0d1ae4679ea2d79ed0ac2750/vitest.config.mts#L88

lib/util/cache/package/impl/factory.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+        const cache = await PackageCache.create({});
+
+        expect(cache).toBeInstanceOf(PackageCache);
+        expect(cache.getType()).toBeUndefined();
+      });
+    });
+
+    describe('Redis backend', () => {
+      it('instantiates redis backend when redisUrl is configured', async () => {
+        const backend = { destroy: vi.fn() } as any;
```
viceice: please use partial helper method for betty type consistency

lib/util/cache/package/impl/file.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  beforeEach(async () => {
+    tmpDir = await dir({ unsafeCleanup: true });
+    cacheDir = tmpDir.path;
+    cacheFileName = upath.join(cacheDir, '/renovate/renovate-cache-v1');
+    packageCache = PackageCacheFile.create(cacheDir);
+  });
+
+  afterEach(async () => {
+    await cacache.rm.all(cacheFileName);
```
viceice: is this really needed? isn't the cleanup on next line also doing that?

lib/util/cache/package/impl/index.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+        expect(secondResult).toBeUndefined();
+        expect(backend.get).not.toHaveBeenCalled();
+      });
+
+      it('evicts L1 memory cache on reset', async () => {
+        const backend = {
+          get: vi.fn().mockResolvedValue('value'),
+          set: vi.fn(),
+          destroy: vi.fn(),
+        } as any;
```
viceice: same, please no any

lib/util/cache/package/impl/index.ts
```diff
@@ -0,0 +1,132 @@
...
+      if (this.memory.has(combinedKey)) {
+        return this.memory.get(combinedKey) as T;
+      }
+
+      return this.getUnsynced<T>(namespace, key);
+    });
+  }
+
+  /**
+   * NOTE: This MUST NOT be used outside of cache implementation, use `get()` instead
```
viceice, line 66: Do we need this public? Can this be private or protected?
```suggestion
   * NOTE: This MUST NOT be used outside of cache implementation, use `get()` instead
   * @private
```

lib/util/cache/package/impl/redis.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+      get: vi.fn(),
+      set: vi.fn(),
+      del: vi.fn(),
+      destroy: vi.fn(),
+    };
+
+    beforeEach(() => {
+      vi.resetAllMocks();
+      vi.mocked(createClient).mockReturnValue(clientMock as any);
+      vi.mocked(createCluster).mockReturnValue(clientMock as any);
```
viceice: avoid any and use partial helper.

lib/util/cache/package/impl/sqlite.ts
```diff
@@ -95,26 +79,41 @@ export class SqlitePackageCache {
...
     `);
 
-    this.countStatement = client
+    this.countStatement = this.client
       .prepare('SELECT COUNT(*) FROM package_cache')
       .pluck(true);
   }
 
-  async set(
+  static async create(cacheDir: string): Promise<PackageCacheSqlite> {
```
viceice: can you move this to top like on redis, so we've less changes here?

lib/util/cache/package/decorator.ts
```diff
@@ -43,124 +47,129 @@ interface CacheParameters {
...
+
+          if (now < cachedAt.plus({ minutes: hardTtlMinutes })) {
+            fallbackValue = cachedRecord.value;
+          }
         }
-      }
 
-      let newData: unknown;
-      if (oldData) {
+        let newValue: unknown;
```
viceice, line 141: can we extract those renames / refactors to be merged first, so this pr get's smaller?

lib/util/cache/package/ttl.spec.ts
```diff
@@ -8,26 +8,34 @@ describe('util/cache/package/ttl', () => {
...
 
   describe('getTtlOverride', () => {
-    describe('basic functionality', () => {
-      it('returns undefined when no cacheTtlOverride config', () => {
+    describe('No configuration', () => {
+      it('returns undefined when no cacheTtlOverride config exists', () => {
         GlobalConfig.set({});
-        expect(getTtlOverride('datasource-npm' as never)).toBeUndefined();
+        const res = getTtlOverride('datasource-npm' as never);
+        expect(res).toBeUndefined();
```
viceice, line 15: can we avoid such unrelated changes? Or extract them to separate PR?

test/util.ts
```diff
@@ -37,6 +38,7 @@ export const scm = vi.mocked(_scm);
 export const env = vi.mocked(_env);
 export const hostRules = vi.mocked(_hostRules);
 export const logger = vi.mocked(_logger, true);
+export const packageCache = vi.mocked(_packageCache);
```
viceice: can we move this to a separate file, so transitive deps are not imported to all tests (perf)

lib/util/cache/memory/index.ts
zharinov: Yes, it was used for synchronization (storing Promise) + cache combo.
In this PR we synchronize concurrent access with mutex and do L1 (mem) cache explicitly.

lib/modules/datasource/deb/index.spec.ts
zharinov: I checked in debugger, in the `main` branch it hits the same link with `GET` and `HEAD` requests about 5 times. In my branch it just makes a single `GET` and then no HTTP requests are made.

Seems like it's now a wrong place to check for race-conditions: our cache layer just ensures it with mutexes and L1 mem cache layer.

lib/util/cache/package/impl/redis.spec.ts
zharinov: Having trouble to apply partial helper to this:

```ts
type RedisClientType<M extends RedisModules = RedisDefaultModules, F extends RedisFunctions = {}, S extends RedisScripts = {}, RESP extends RespVersions = 2, TYPE_MAPPING extends TypeMapping = {}> = GenericRedisClientType<M, F, S, RESP, TYPE_MAPPING>;
```

Sticking to `as never`.

lib/util/cache/package/impl/index.ts
zharinov: Those needs to be used from decorator and http cache layer. I don't know how to idiomatically make this scope right.

lib/util/cache/package/decorator.ts
zharinov: https://github.com/renovatebot/renovate/pull/40014

lib/util/cache/package/ttl.spec.ts
zharinov: https://github.com/renovatebot/renovate/pull/40015

lib/modules/datasource/deb/index.spec.ts
zharinov: Ah, remembered the exact reason: L1 memory cache (per-run) works even without backend being initialized, like in this test.

lib/util/cache/package/decorator.ts
```diff
@@ -43,115 +47,129 @@ interface CacheParameters {
   ttlMinutes?: number;
 }
 
-/**
```
viceice, line 46: why removed?

lib/util/cache/package/impl/index.ts
viceice: we cn improve in folow up pr

lib/util/cache/package/decorator.ts
zharinov: Because it doesn't have any value

lib/util/cache/package/impl/factory.spec.ts
viceice: 👆?

viceice: 👆?

viceice (commented):
otherwise LGTM

viceice:
conflicted

RahulGautamSingh (approved):

zharinov:
- Superseded by #41434

PR #39512, opened 2025-11-24
feat(replacements): Adds replacement for Docker grafana/grafana-oss to grafana/grafana

MalteJoe:
## Changes

Adds a replacement strategy for Docker `grafana/grafana-oss`.

The Docker Hub Repository `grafana/grafana-oss` is not updated anymore (see banner at https://grafana.com/docs/grafana/v12.3/setup-grafana/installation/docker/) and is replaced by `grafana/grafana`.

I checked the SHA256 of `grafana/grafana-oss:12.1.1` and `grafana/grafana:12.1.1` to be the same (which they are)

## Context

## AI assistance disclosure

### DIFF
--- a/lib/data/replacements.json
+++ b/lib/data/replacements.json
@@ -706,11 +706,12 @@
     ]
   },
   "grafana-oss-to-grafana": {
-    "description": "Starting with Grafana release `12.4.0`, the `grafana/grafana-oss` Docker Hub repository will no longer be updated. Instead, we encourage you to use the `grafana/grafana` Docker Hub repository. These two repositories have the same Grafana OSS docker images.",
+    "description": "`grafana/grafana-oss`` was replaced by `grafana/grafana` with the same images",
     "packageRules": [
       {
         "matchDatasources": ["docker"],
-        "matchPackageNames": ["grafana/grafana-oss"],
+        "matchPackageNames": ["/^(?:docker\\.io/)?grafana/grafana-oss$/"],
+        "matchCurrentVersion": ">=12.1.1",
         "replacementName": "grafana/grafana"
       }
     ]
--- a/lib/data/replacements.json
+++ b/lib/data/replacements.json
@@ -19,6 +19,7 @@
       "replacements:framer-motion-to-motion",
       "replacements:hapi-to-scoped",
       "replacements:google-github-action-release-please-to-googleapis",
+      "replacements:grafana-oss-to-grafana",
       "replacements:jade-to-pug",
       "replacements:joi-to-scoped",
       "replacements:joi-to-unscoped",
@@ -701,6 +702,16 @@
         "matchDatasources": ["github-tags"],
         "matchPackageNames": ["google-github-actions/release-please-action"],
         "replacementName": "googleapis/release-please-action"
+      }
+    ]
+  },
+  "grafana-oss-to-grafana": {
+    "description": "Starting with Grafana release `12.4.0`, the `grafana/grafana-oss` Docker Hub repository will no longer be updated. Instead, we encourage you to use the `grafana/grafana` Docker Hub repository. These two repositories have the same Grafana OSS docker images.",
+    "packageRules": [
+      {
+        "matchDatasources": ["docker"],
+        "matchPackageNames": ["grafana/grafana-oss"],
+        "replacementName": "grafana/grafana"
       }
     ]
   },


lib/data/replacements.json
```diff
@@ -704,6 +705,16 @@
       }
     ]
   },
+  "grafana-oss-to-grafana": {
+    "description": "Starting with Grafana release `12.4.0`, the `grafana/grafana-oss` Docker Hub repository will no longer be updated. Instead, we encourage you to use the `grafana/grafana` Docker Hub repository. These two repositories have the same Grafana OSS docker images.",
+    "packageRules": [
+      {
+        "matchDatasources": ["docker"],
+        "matchPackageNames": ["grafana/grafana-oss"],
+        "replacementName": "grafana/grafana"
```
viceice, line 715:
```suggestion
        "matchCurrentVersion": ">=12.4.0",
        "replacementName": "grafana/grafana"
```

viceice (commented):
this doesn't work well when there is a registry infront 🤔

jamietanna:
> this doesn't work well when there is a registry infront 🤔

Do any of our replacements work with a registry in front? Could it be an existing issue? 👀

RahulGautamSingh:
> this doesn't work well when there is a registry infront 🤔

Why? It matches datasource and pkg name ... regardless of which regsitry one pulls info from for a pakcage won't it be necessary the registry maintainers track and sync these changes.

lib/data/replacements.json
MalteJoe: In the Bannar grafana states:

> Starting with Grafana release 12.4.0 , the grafana/grafana-oss Docker Hub repository will no longer be updated.

It isn't clear to me, whether that actually includes `12.4.0` or not, **but** these tags already exist in `grafana/grafana`, but not in `grafana/grafana-oss`:
- `12.3.0`
- `12.2.1`
- `12.2.0`
- `12.1.4`
- `12.1.3`
- `12.1.2`

The latest version in `grafana-oss` I could find is [`12.2.0-17142428006`](https://hub.docker.com/layers/grafana/grafana-oss/12.2.0-17142428006/images/sha256-40d6822024fe40012461d0cbb290f91319ed390bc83ccf70fa38f630cc9d762d) which is not even tagged as `12.2.0`. Only the `latest` image has been updated more recently. 

We are currently having a Dockerfile with `FROM grafana/grafana-oss:12.1.1` which is the latest version following that schema. This should be easily replaceable with `grafana/grafana:12.1.1` as both [`grafana/grafana-oss:12.1.1`](https://hub.docker.com/layers/grafana/grafana-oss/12.1.1/images/sha256-5749a0e982878aedaa2d320ed14d3bdce7a040a938f482258670009d419a597f) and [`grafana/grafana:12.1.1`](https://hub.docker.com/layers/grafana/grafana/12.1.1/images/sha256-5749a0e982878aedaa2d320ed14d3bdce7a040a938f482258670009d419a597f) even share the image digest.

To me personally it looks like they abandoned that image for a while now and it has the same contents as `grafana/grafana`, so I would not include the `"matchCurrentVersion": ">=12.4.0"`, as `grafana/grafana-oss:12.4.0` is likely to not be tagged accordingly. No one would be able use this replacement.

We probably can do `"matchCurrentVersion": ">=12.1.1"` as we have verified that this tags exists for both images and has the same digest

RahulGautamSingh: Seems true. The last version on [Dockerhub](https://hub.docker.com/r/grafana/grafana-oss) is indeed `12.2.0`. We should then do a `matchCurrentVersion>=12.2.0` to be safe

MalteJoe:
> Do any of our replacements work with a registry in front? Could it be an existing issue? 👀

Some of the other docker replacements at least use an optional prefix `/^(?:docker\\.io/)?`. Still wouldn't help if you are using a proxy registry.

lib/data/replacements.json
MalteJoe: There is only [`12.2.0-17142428006`](https://hub.docker.com/layers/grafana/grafana-oss/12.2.0-17142428006/images/sha256-40d6822024fe40012461d0cbb290f91319ed390bc83ccf70fa38f630cc9d762d) but no `12.2.0`

viceice:
> > this doesn't work well when there is a registry infront 🤔
> 
> Why? It matches datasource and pkg name ... regardless of which regsitry one pulls info from for a pakcage won't it be necessary the registry maintainers track and sync these changes.

because the registry is added to the dep name for docker images most of the time.

lib/data/replacements.json
```diff
@@ -704,6 +705,17 @@
       }
     ]
   },
+  "grafana-oss-to-grafana": {
+    "description": "`grafana/grafana-oss`` was replaced by `grafana/grafana` with the same images",
+    "packageRules": [
+      {
+        "matchDatasources": ["docker"],
+        "matchPackageNames": ["/^(?:docker\\.io/)?grafana/grafana-oss$/"],
```
RahulGautamSingh, line 713: Should be non-optional IMO, since we are making thos only for docker atm

RahulGautamSingh (changes requested):

lib/data/replacements.json
MalteJoe: It is done with an optional prefix for all of the other `docker` replacements (except when a specific registry is used anyway).

Also making the prefix mandatory would not update usages when the registry is not specified, right?
Something like `FROM grafana/grafana-oss:12.1.1` (using the default registry  `docker.io`) in a `Dockerfile`. Am I overlooking something?

viceice (approved):

PR #39554, opened 2025-11-26
feat(bitbucket-server): get username from header

### LINKED ISSUE #39527: Bitbucket-Server: get username using header
### Discussed in https://github.com/renovatebot/renovate/discussions/39278

<sup>Originally posted by **RahulGautamSingh** November 13, 2025</sup>
### Tell us more.

Currently, when we configure a bitbucket-server using a token, we do fetch username attached to the token. I am not sure if a username is always present for a token or not .. I'll need to verify that.

We don't fetch the username like we do on other platforms because there is no dedicated api endpoint for it.

There seems to be a way to get the username thought by adding the header: `X-AUSERNAME` to any request that we make to the bitbucket-server API. [^1]

The suggested way is to call the `/rest/api/1.0/application-properties` endpoint which never fails and attach the header with it to get user details related to the token from the response.

In the bitbucket-server code we conditionally make the request to `application-properties` endpoint here:
https://github.com/renovatebot/renovate/blob/84633f1e560d6e0a787184c5e57e0857c8faece8/lib/modules/platform/bitbucket-server/index.ts#L129-L139

I think we should remove the condition and always make this call with the required header, so we can fetch `username`..except for the rare case when `username` is already configures and `env.RENOVATE_X_PLATFORM_VERSION` is also set.

 [^1]: https://community.developer.atlassian.com/t/obtain-authorised-users-username-from-api/24422</div>

RahulGautamSingh:
## Changes
- Pass `token`(if present) to the api call we use to fetch BBS version so that we can get `x-ausername` header 

## Context

- [x] This closes an existing Issue, Closes: #39527

## AI assistance disclosure

The public repository: Not possible but tested locally

<details>

<summary> Logs </summary>

```
DEBUG: Platform: No username configured using headers["x-ausername"]
       "x-username": "renovate-bot-rahtes"
DEBUG: Bitbucket Server version is: 8.9.6
```

</details>


RahulGautamSingh:
Needs tests

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -123,20 +123,23 @@ export async function initPlatform({
...
+    bitbucketServerVersion ??= body.version;
+    if (headers['x-ausername'] && !username) {
+      logger.debug(
+        { 'x-username': headers['x-ausername'] },
+        'Platform: No username configured using headers["x-ausername"]',
+      );
+      config.username = headers['x-ausername'];
     }
+    logger.debug('Bitbucket Server version is: ' + bitbucketServerVersion);
+    /* v8 ignore stop */
```
viceice:
```suggestion
```

lib/modules/platform/bitbucket-server/index.ts
```diff
@@ -123,20 +123,23 @@ export async function initPlatform({
...
-    } else {
-      const { version } = (
-        await bitbucketServerHttp.getJsonUnchecked<{ version: string }>(
-          `./rest/api/1.0/application-properties`,
-        )
-      ).body;
-      bitbucketServerVersion = version;
-      logger.debug('Bitbucket Server version is: ' + bitbucketServerVersion);
+    let bitbucketServerVersion = env.RENOVATE_X_PLATFORM_VERSION;
+    /* v8 ignore start */
```
viceice:
```suggestion
```
Does no longer work. isn't this code path tested already?

lib/util/http/types.ts
```diff
@@ -70,6 +70,7 @@ export interface HttpOptions {
 
 export interface HttpHeaders extends IncomingHttpHeaders {
   link?: string | undefined;
+  'x-ausername'?: string; // bitbucket-server
```
viceice: i don't like this addition. do we really need this? can't we use `isNotEmptyAndNotWhitespaceString` instead?

viceice (approved):

PR #39579, opened 2025-11-27
fix(platform): `endpoint` is a global only option

viceice:
## Changes
- remove `endpoint` from `RepoParams`, it's a global only
- fixes a regresion of #35161

## Context

## AI assistance disclosure

The public repository: <URL>


lib/modules/platform/github/index.ts
```diff
@@ -491,13 +490,6 @@ export async function initRepo({
     cloneSubmodulesFilter,
     ignorePrAuthor: GlobalConfig.get('ignorePrAuthor', false),
   } as any;
-  /* v8 ignore next */
-  if (endpoint) {
-    // Necessary for Renovate Pro - do not remove
-    logger.debug(`Overriding default GitHub endpoint with ${endpoint}`);
-    platformConfig.endpoint = endpoint;
-    githubHttp.setBaseUrl(endpoint);
-  }
```
jamietanna, line 500: Double checking this

jamietanna (commented):
Good investigating! Is there anything we can do to catch issues like this in an automated way? I think in this case, we've found it's a little less widespread than when we started looking into it, but wondering if we can craft linting rules / automation to spot + block things like this getting back in

lib/modules/platform/github/index.ts
jamietanna: Confirmed internally, this should be OK to remove due to age!

lib/config/global.ts
jamietanna, line 1: I wonder if we should try and force this to be sorted, through linting?

jamietanna (approved):

lib/config/global.ts
viceice: yes, working on a test

PR #39406, opened 2025-11-19
docs(self-hosting): make `RENOVATE_GITHUB_COM_TOKEN` requirement more clear

jamietanna:
## Changes

As seen in #37685, #39399 and https://github.com/renovatebot/renovate/discussions/39522, it appears that these users have run into
the non-GitHub.com rate limits.

We can make this clearer by putting this in a more prominent position
for self-hosting users.

## Context

Please select one of the below:

- [ ] This closes an existing Issue: #
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [x] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #39585, opened 2025-11-27
fix(github): warn on rate limit errors

jamietanna:
## Changes

As noted in #39406, we're seeing a number of self-hosted users who have
missed the onboarding requirement for a `GITHUB_COM_TOKEN` for their
authentication to `github.com`.

On top of the docs improvements in #39406, we can make this clearer by
adding a warning log when we're going to report a `rate-limit-exceeded`
error, noting which hostname is failing (in case of GitHub Enterprise).

This requires a slightly duplicated `fail` function re-definition, so we
can override the baseUrl.

In the case that the request fails for `github.com`, we can also point
to our docs for the configuration needed.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/util/http/github.ts
+++ b/lib/util/http/github.ts
@@ -15,6 +15,7 @@ import { logger } from '../../logger';
 import { ExternalHostError } from '../../types/errors/external-host-error';
 import { getCache } from '../cache/repository';
 import { getEnv } from '../env';
+import * as hostRules from '../host-rules';
 import { maskToken } from '../mask';
 import * as p from '../promises';
 import { range } from '../range';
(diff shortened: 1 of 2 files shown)


lib/util/http/github.ts
```diff
@@ -116,6 +117,30 @@ function handleGotError(
...
+
+    const parsed = parseUrl(baseUrl);
+
+    const rule = hostRules.find({ url: baseUrl });
+    if (rule.token || rule.password) {
+      logger.once.warn(
+        'Rate limit exceeded for api.github.com, even though we are authenticated',
+      );
+    } else {
+      if (parsed?.hostname === 'api.github.com') {
```
RahulGautamSingh, line 129: Why this if-check is needed?

jamietanna: I'm trying to provide a  clearer difference between `github.com` rate limit (you need to set a `GITHUB_COM_TOKEN` or similar) and a self-hosted GitHub Enterprise server, which have different needs

viceice (approved):

PR #39589, opened 2025-11-27
refactor(splits): add a type for repository splits

jamietanna:
## Changes

As part of future changes to add an OpenTelemetry attribute for the
splits, we should make sure that we have a common type for the splits,
which can be used with `addSplit`.

This refactors - with some help from GPT-4.1 - to use Vitest's timer
mocking.

This also requires a slight refactor of the implementation to test more
appropriately, without duplicate calls to `Date.now`.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/util/split.spec.ts
+++ b/lib/util/split.spec.ts
@@ -2,11 +2,18 @@ import { addSplit, getSplits, splitInit } from './split';
 
 describe('util/split', () => {
   it('adds splits and returns results', () => {
+    vi.setSystemTime(0);
     splitInit();
-    addSplit('one');
-    addSplit('two');
+
+    vi.setSystemTime(1000);
+    addSplit('init');
+
+    vi.setSystemTime(3000);
+    addSplit('lookup');
+
     const res = getSplits();
     expect(res.total).toBeDefined();
-    expect(Object.keys(res.splits)).toHaveLength(2);
+    expect(res.splits.init).toEqual(1000);
+    expect(res.splits.lookup).toEqual(2000);
   });
 });
--- a/lib/config/types.ts
+++ b/lib/config/types.ts
@@ -17,6 +17,13 @@ export type RenovateConfigStage =
   | 'branch'
   | 'pr';
 
+export type RenovateSplit =
+  | 'init'
+  | 'onboarding'
+  | 'extract'
+  | 'lookup'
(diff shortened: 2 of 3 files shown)


jamietanna:
Will then be used in https://github.com/renovatebot/renovate/pull/38633

lib/util/split.spec.ts
```diff
@@ -1,12 +1,30 @@
 import { addSplit, getSplits, splitInit } from './split';
 
 describe('util/split', () => {
+  afterAll(() => {
+    vi.unstubAllGlobals();
```
viceice: are you sure you need this? global stubs should not interfere with other test files 🤔

lib/util/split.spec.ts
```diff
@@ -1,12 +1,30 @@
...
+  });
+
   it('adds splits and returns results', () => {
+    const values = [0, 1000, 3000];
+    let call = 0;
+    const OriginalDate = Date;
+
+    vi.stubGlobal(
+      'Date',
+      class extends OriginalDate {
```
viceice: use fake timer instead of this

lib/util/split.spec.ts
jamietanna: Possibly not - thought it would be good practice to make sure I un-stub anything stubbed?

(And GPT-4.1 told me to do it!)

secustor: If you are using like simple chat than the models will not be aware of our config, but rather implement like what most people have.

secustor (commented):
LGTM, outside of open conversations

lib/util/split.spec.ts
jamietanna: TY, that was simpler - I should've read up how we were doing it elsewhere in more depth

lib/util/split.spec.ts
```diff
@@ -1,12 +1,23 @@
 import { addSplit, getSplits, splitInit } from './split';
 
 describe('util/split', () => {
+  afterAll(() => {
+    vi.useRealTimers();
+  });
```
viceice:
```suggestion
```
not needed

viceice (commented):
otherwise LGTM

viceice (approved):

PR #39594, opened 2025-11-27
fix: use stricter types for renovate config

viceice:
## Changes
- Remove `Record<string,any>` from all Renovate Config types
- #22198

## Context

## AI assistance disclosure

The public repository: <URL>


jamietanna (commented):
As an aside, do we have a type for only repo config? Is that `RenovateSharedConfig`?

(I can raise a follow-up PR to add some JSDoc for the various types and what they're for)

lib/config/types.ts
```diff
@@ -67,23 +85,35 @@ export interface RenovateSharedConfig {
...
+  internalChecksFilter?: 'strict' | 'flexible' | 'none';
   keepUpdatedLabel?: string;
   labels?: string[];
   manager?: string;
   milestone?: number;
+  minimumReleaseAge?: Nullish<string>;
   npmrc?: string;
   npmrcMerge?: boolean;
+  npmToken?: string;
+
```
jamietanna, line 100: intentional linebreak?

jamietanna (commented):
Validated there's no change to the JSON Schema 👍

lib/config/types.ts
```diff
@@ -234,32 +287,52 @@ export const allowedStatusCheckStrings = [
...
+  baseBranches?: string[];
+  currentCompatibility?: string;
+  datasource?: string;
+  hasBaseBranches?: boolean;
+  isFork?: boolean;
+  isVulnerabilityAlert?: boolean;
+
+  /** ??? */
+  remediations?: unknown;
+  /** ??? */
```
jamietanna: re we filling these in now, or should we mark as TODOs?

lib/config/types.ts
```diff
@@ -286,21 +357,33 @@ export interface RenovateConfig
...
+  printConfig?: boolean;
+
+  pruneStaleBranches?: boolean;
 
   defaultRegistryUrls?: string[];
   registryUrls?: string[] | null;
   registryAliases?: Record<string, string>;
 
+  /**
+   * What is this used for?
```
jamietanna, line 378: I can't see it used anywhere (in a couple of the Mend codebases, nor in Renovate) so wonder if it's no longer needed - makes sense to deprecate, and we can remove in next major

lib/config/types.ts
viceice: We can add todo, i just couldn't figure for what that is. I plan to work on this in follow up pr's

lib/config/types.ts
viceice: yes, i plan to add more for readybillity. we should also split to more interfaces to better group properties by usage, but that's aout of scope for this pr.

lib/config/types.ts
viceice: It's internal any ways, we can remove in a patch release. I traced that code to early days of renovate. Maybe there was an intention to use that later, but it isn't used. Removal is out of scop to this PR.

lib/config/types.ts
```diff
@@ -234,32 +287,52 @@ export const allowedStatusCheckStrings = [
...
+export interface RenovateInternalConfig {
+  /** computed base branch from patterns - for internal use only */
+  baseBranches?: string[];
+  currentCompatibility?: string;
+  datasource?: string;
+  hasBaseBranches?: boolean;
+  isFork?: boolean;
+  isVulnerabilityAlert?: boolean;
+
+  /** ??? */
```
viceice:
```suggestion
  /** What is this used for? */
```

lib/config/types.ts
viceice:
```suggestion
  /** What is this used for? */
```

jamietanna (approved):

PR #39635, opened 2025-12-01
fix(github-actions): handling of quoted Docker image references

djmattyg007:
## Changes

The problem with the use of `\s` is that it swallows the newline at the end of the line, meaning the $ anchor doesn't have anything to match against, which results in it failing to match quoted entries. By simply switching to a dot (which doesn't match newlines), the issue is resolved for all references.

## Context

## AI assistance disclosure

### DIFF
--- a/lib/modules/manager/github-actions/extract.ts
+++ b/lib/modules/manager/github-actions/extract.ts
@@ -20,7 +20,7 @@ import { CommunityActions } from './community';
 import type { Steps } from './schema';
 import { Workflow } from './schema';
 
-const dockerActionRe = regEx(/^\s+uses\s*: ['"]?docker:\/\/([^'"]+)\s*$/);
+const dockerActionRe = regEx(/^\s+uses\s*: ['"]?docker:\/\/([^'"]+).*$/);
 const actionRe = regEx(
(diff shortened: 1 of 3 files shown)


RahulGautamSingh (approved):

lib/modules/manager/github-actions/extract.ts
```diff
@@ -20,7 +20,7 @@ import { CommunityActions } from './community';
 import type { Steps } from './schema';
 import { Workflow } from './schema';
 
-const dockerActionRe = regEx(/^\s+uses\s*: ['"]?docker:\/\/([^'"]+)\s*$/);
+const dockerActionRe = regEx(/^\s+uses\s*: ['"]?docker:\/\/([^'"]+).*$/);
```
viceice: the dot is a problem for non quoted cases. so also wrong

viceice (changes requested):

lib/modules/manager/github-actions/extract.ts
djmattyg007: If that's the case, why are all the existing tests passing without modification? If this isn't the correct solution, do you have any tips on what the real problem is?

viceice:
```suggestion
const dockerActionRe = regEx(/^\s+uses\s*: ['"]?docker:\/\/([^'"\s]+)/);
```

maybe this?

djmattyg007: I've updated my branch with your suggestion. All of the tests are still passing.

viceice (approved):

PR #39638, opened 2025-12-01
feat(devbox): Add support for various versioning apis

burritobill:
## Changes

I have had some engineers at Culture Amp telling me that some versions are being missed for JDK and go.

I initially implemented the versioning very simply to only allow for versions like 1.2.3 but because Devbox installs many different tools from Nix there are a lot of different version schemes that need to be supported.

This PR allows us to define different versioning APIs for different package names.
It covers some I have seen issues with, but also allows us to add more in future.

## Context

## AI assistance disclosure

The public repository: <URL>


lib/modules/manager/devbox/artifacts.ts
```diff
@@ -8,7 +8,7 @@ import { getSiblingFileName, readLocalFile } from '../../../util/fs';
 import type { UpdateArtifact, UpdateArtifactsResult } from '../types';
 
 export async function updateArtifacts({
-  config: { constraints, isLockFileMaintenance, updateType },
+  config: { constraints, isLockFileMaintenance },
```
viceice, line 11: unrelated but ok

lib/modules/manager/devbox/schema.ts
```diff
@@ -7,8 +7,9 @@ import {
   withDebugMessage,
 } from '../../../util/schema-utils';
 import { DevboxDatasource } from '../../datasource/devbox';
-import { api as devboxVersioning } from '../../versioning/devbox';
+import * as devboxVersioning from '../../versioning/devbox';
```
viceice: revert, see below

lib/modules/manager/devbox/schema.ts
```diff
@@ -30,7 +31,12 @@ const DevboxEntry = z
 
     dep.currentValue = currentValue;
 
-    if (!devboxVersioning.isValid(currentValue)) {
+    let versioning = devboxToolVersioning[depName];
+    versioning ??= devboxVersioning;
+
+    dep.versioning = versioning.id;
+
+    if (!versioning.api.isValid(currentValue)) {
```
viceice:
```suggestion
    const versioning = devboxToolVersioning[depName];

    if (versioning) {
      dep.versioning = versioning.id;
    }

    if (!(versioning?.api ?? api).isValid(currentValue)) {
```
don't always override versioning

lib/modules/versioning/devbox/index.spec.ts
```diff
@@ -6,7 +6,6 @@ describe('modules/versioning/devbox/index', () => {
     ${'1'}            | ${false}
     ${'01'}           | ${false}
     ${'1.01'}         | ${false}
-    ${'1.1'}          | ${false}
```
viceice, line 9: why the versioning changes here?

lib/modules/manager/devbox/tool-versioning.ts
```diff
@@ -0,0 +1,38 @@
...
+  ruby: {
+    api: rubyVersioning.api,
+    id: rubyVersioning.id,
+  },
+  python: {
+    api: pythonVersioning.api,
+    id: pythonVersioning.id,
+  },
+  // Using loose versioning due to versions like 1.2.3+2
+  jdk: {
```
viceice, line 24: should be semver

viceice (changes requested):

burritobill:
@viceice Not sure why the undesirable code check is failing, I haven't added any snapshots

lib/modules/manager/devbox/extract.spec.ts
```diff
@@ -120,12 +100,13 @@ describe('modules/manager/devbox/extract', () => {
         `,
         'devbox.lock',
       );
-      expect(result).toEqual({
+      expect(result).toMatchObject({
```
viceice: why all those changes?

viceice (commented):
otherwise LGTM

lib/modules/manager/devbox/extract.spec.ts
burritobill: I thought this would fix that ci failure. Happy to revert those.

lib/modules/manager/devbox/extract.spec.ts
```diff
@@ -55,29 +56,7 @@ describe('modules/manager/devbox/extract', () => {
             depName: 'nodejs',
             currentValue: '20.1.8',
             datasource: 'devbox',
-          },
-        ],
-      });
-    });
-
-    it('returns invalid-version when the devbox JSON file has a single package with an invalid version', () => {
```
RahulGautamSingh, line 63: won't removing this case coverage issues? Better to use a invalid toolname instead of deleting the test.

RahulGautamSingh (changes requested):

lib/modules/manager/devbox/extract.spec.ts
burritobill: I have added an additional check in the schema to make sure this check still works and have added it back

viceice (approved):

PR #39653, opened 2025-12-01
fix(instrumentation): correctly `async` / `await` Git operations

jamietanna:
## Changes

A previous refactor had incorrectly made it so we were not correctly
`await`ing any calls to the SimpleGit method calls.

This has mostly led to the Git operations continuing to work, but the
instrumentation hasn't been correctly rendering.

This also corrects incorrect return types that were previously not
returned correctly.

## Screenshots

Before (`git clone` immediately returns):

Screenshot 2025-12-01 at 11 41 19

After (`git clone` returns correct timing)

Screenshot 2025-12-01 at 11 40 25

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/util/git/instrument.ts
+++ b/lib/util/git/instrument.ts
@@ -5,6 +5,7 @@ import type {
   DiffResult,
   LogOptions,
   LogResult,
+  MergeResult,
   Options,
   RemoteWithRefs,
   RemoteWithoutRefs,
@@ -97,10 +98,10 @@ export class InstrumentedSimpleGit {
     return this.git.version();
   }
 
-  raw(args: string[]): Response<string> {
+  raw(args: string[]): Promise<string> {
     const { spanName, options } = prepareInstrumentation(args[0]);
 
-    return instrument(spanName, () => this.git.raw(args), options);
+    return instrument(spanName, async () => await this.git.raw(args), options);
   }
 
   async checkout(whatOrOptions: string | TaskOptions): Promise<string> {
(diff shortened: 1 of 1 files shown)


lib/util/git/instrument.ts
```diff
@@ -119,71 +120,99 @@ export class InstrumentedSimpleGit {
...
   }
-  async clone(repo: string, dir: string, options: string[]): Promise<void> {
+  clone(repo: string, dir: string, options: string[]): Promise<string> {
     const { spanName, options: spanOptions } = prepareInstrumentation('clone');
 
-    await instrument(
+    return instrument(
       spanName,
-      () => this.git.clone(repo, dir, options),
+      async () => await this.git.clone(repo, dir, options),
```
viceice, line 215: i wonder why this maks a difference. 🤔 

https://github.com/renovatebot/renovate/blob/92e176a7c43fc6bf037871e1a177fcd1b3298934/lib/instrumentation/index.ts#L185-L196

this should handle the response properly

viceice (approved):
i don't understand why those changese makes a difference, see comment. 🤷‍♂️

lib/util/git/instrument.ts
viceice: wait, maybe simple git isn't returning a proper `Promise` so it doesn't get into the if condition.

jamietanna: Ah, so possibly because it returns:

```typescript
export type Response<T> = SimpleGit & Promise<T>;
```

So maybe if we refactored ☝️ to also look at `instanceof Response`, it'd work?

jamietanna: It looks like the ordering if the types (when the object is constructed) can catch us out here - I've got a reproducing unit test I'm trying to get passing then will see if that fixes it

jamietanna: When adding:

```
      console.log({ name, typF: typeof fn, typeR: typeof ret, fn, ret, })
```

I see i.e.

```
{
  name: 'syncGit',
  typF: 'function',
  typeR: 'object',
  fn: [Function (anonymous)],
  ret: Promise {
    undefined,
    [Symbol(async_id_symbol)]: 1857648,
    [Symbol(trigger_async_id_symbol)]: 1857588,
    [Symbol(kResourceStore)]: BaseContext {
      _currentContext: [Map],
      getValue: [Function (anonymous)],
      setValue: [Function (anonymous)],
      deleteValue: [Function (anonymous)]
    }
  }
}
{
  name: 'git show',
  typF: 'function',
  typeR: 'object',
  fn: [Function (anonymous)],
  ret: Git2 {}
}
```

Notice that SimpleGit's calls return a `Git2` (from the CommonJS code)

jamietanna: Alternative: https://github.com/renovatebot/renovate/pull/39661

jamietanna:
https://github.com/renovatebot/renovate/pull/39661 was the better fix

jamietanna:
Moved the types changes into https://github.com/renovatebot/renovate/pull/39664

PR #39655, opened 2025-12-01
fix(cache): instrument cache creation + destroy

jamietanna:
## Changes

As we're already calculating the duration of operations, we can make
sure that we add our instrumentation around it.

This will then provide instrumentation across each of our cache
implementations.

Note that this doesn't introduce a span for each cache key we're looking
to clean up, given a small repo can have hundreds of cached items, let
alone for large repositories.

## Context

## AI assistance disclosure

The public repository: <URL>

At the start:

Screenshot 2026-03-09 at 12 08 05

At the end:

Screenshot 2026-03-09 at 12 08 30

### DIFF
--- a/lib/util/cache/package/backend.ts
+++ b/lib/util/cache/package/backend.ts
@@ -1,4 +1,5 @@
 import type { AllConfig } from '../../../config/types.ts';
+import { instrument } from '../../../instrumentation/index.ts';
 import { getEnv } from '../../env.ts';
 import type { PackageCacheBase } from './impl/base.ts';
 import { PackageCacheFile } from './impl/file.ts';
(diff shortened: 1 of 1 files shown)


jamietanna:
Maybe this should cover caching all of them

lib/util/cache/package/backend.ts
```diff
@@ -14,28 +15,30 @@ export function getCacheType(): typeof cacheType {
 }
 
 export async function init(config: AllConfig): Promise<void> {
-  await destroy();
+  await instrument('Initialise cache', async () => {
+    await destroy();
```
viceice, line 19:
```suggestion
  await destroy();
  await instrument('Initialise cache', async () => {
```
destroy should be outside of the init span. otherwise we get an additional nested detroy span

viceice (commented):
otherwise LGTM

PR #39669, opened 2025-12-01
fix(global): ensure that all `globalOnly` options can be set with Global Config

### LINKED ISSUE #9603: Add tests for "config/admin.ts"

Check that `setAdminConfig` iterates over all fields with `admin: true` (in `definitions.ts`)

**Did you already have any implementation ideas?**

jamietanna:
## Changes

As part of #9603, we intended to come back and make sure that all of our
`globalOnly` configuration fields are set here, so they can be correctly
updated by configuration (in various forms).

As some configuration may not work if we do not set this, we should make
sure that this is applied, as well as ensuring that there is a test to
catch this requirement.

## Context

- [x] This closes an existing Issue, Closes: #9603, 

## AI assistance disclosure

GPT-4.1

The public repository: <URL>


lib/config/global.ts
```diff
@@ -1,45 +1,108 @@
 import type { RenovateConfig, RepoGlobalConfig } from './types';
 
 export class GlobalConfig {
-  // TODO: once global config work is complete, add a test to make sure this list includes all options with globalOnly=true (#9603)
   private static readonly OPTIONS: (keyof RepoGlobalConfig)[] = [
```
jamietanna: As a separate change (prefactor), I'll see if we can make sure these are sorted

lib/config/global.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+describe('config/global', () => {
+  it('all', () => {
+    const globalOnlyOptions = getOptions()
+      .filter((o) => o.globalOnly)
+      .map((o) => o.name);
+
+    // HACK
+    const defined = (GlobalConfig as any).OPTIONS;
+    console.log(globalOnlyOptions);
```
jamietanna:
```suggestion
```

zharinov (approved):

lib/config/global.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { GlobalConfig } from './global';
+import { getOptions } from './options';
+
+describe('config/global', () => {
+  it('all', () => {
```
jamietanna: TODO

lib/config/global.ts
jamietanna: https://github.com/renovatebot/renovate/pull/39681

viceice (changes requested):
This will break renovate, as all thos options are filtered here and are no longer on `RepoGlobalConfig`

https://github.com/renovatebot/renovate/blob/78e5977977d2f0529ee1c2c621060aefea180241/lib/config/global.ts#L70

jamietanna:
OK fair - so this is maybe the inverse? We should be making sure that `GlobalConfig.get` is used for each of these options, rather than them called as i.e. `config.dryRun`?

jamietanna:
We should re-review this (when we come round to it) to determine if https://github.com/renovatebot/renovate/pull/24115 has any pointers to things that should (not) be included

PR #39704, opened 2025-12-02
fix(allowedUnsafeExecutions): allow usagfe in the `repository` stage

mglazer:
## Changes

Found this during actual integration testing. Test coverage here was a little bit hard to come by and I didn't even notice the string list that does this filtering until it was too late.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -3099,7 +3099,7 @@ const options: RenovateOptions[] = [
     type: 'array',
     subType: 'string',
     default: [],
-    stage: 'global',
+    stage: 'repository',
     globalOnly: true,
   },
   {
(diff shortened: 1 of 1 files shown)


jamietanna (changes requested):
Hmm I don't think this is quite what we want to do here - `stage=repository` means that this is _repository-level_ config that can be changed. That isn't the case, as this should only be a global setting

mglazer:
that's what I thought too, but unfortunately this seems to pull out all non-global configs:

https://github.com/renovatebot/renovate/blob/f383805e81592d866b4b8e24a658e1f93b7d3bc6/lib/config/index.ts#L53-L76

Which makes it unavailable later on.

This is in line with `binarySource` which also is used as GlobalConfig.

jamietanna:
Thanks - also doing some digging I can confirm that we need to set `stage=repository`. Thanks for reverting the `globalOnly` change, as that was my main concern here 👍 

I can confirm that this change then means this allows #39163 to run

jamietanna (approved):

jamietanna:
Just noticed I'd typo'd in the PR title that's now been released 😅

PR #39769, opened 2025-12-03
feat: map emojis to repository problems

setchy:
## Changes

Jazz up the Repository Problems section with some colorful emojis

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/workers/repository/common.ts
+++ b/lib/workers/repository/common.ts
@@ -1,3 +1,4 @@
+import type { LogLevelString } from 'bunyan';
 import { nameFromLevel } from 'bunyan';
 import { getProblems } from '../../logger';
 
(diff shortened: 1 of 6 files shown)


lib/workers/repository/common.ts
```diff
@@ -12,7 +12,26 @@ export function extractRepoProblems(
       )
       .map(
         (problem) =>
-          `${nameFromLevel[problem.level].toUpperCase()}: ${problem.msg}`,
+          `${emojiFromLevel(problem.level)} ${nameFromLevel[problem.level].toUpperCase()}: ${problem.msg}`,
```
jamietanna: We should conditionally add the space, as if there isn't an emoji, it'll have a single space

lib/workers/repository/common.ts
```diff
@@ -12,7 +12,26 @@ export function extractRepoProblems(
...
       ),
   );
 }
+
+export function emojiFromLevel(level: number): string {
```
jamietanna: Can we make sure this works with https://docs.renovatebot.com/self-hosted-configuration/#unicodeemoji ? I.e. return a `:warning:` that'll then go through the emojify function

viceice: I think that function is making the opposite? 🤔

viceice (commented):
otherwise LGTM

lib/workers/repository/common.ts
jamietanna: Sorry, poorly worded

I meant that we should be calling https://github.com/renovatebot/renovate/blob/HEAD/lib/util/emoji.ts#L75 here with i.e. `:warning`, so then if a self-hosted user has set `unicodeEmoji: false`, they receive the text output accordingly

jamietanna: I also wonder if maybe this should go into `../../logger'`? No big deal if not at this point

setchy: Refactored to use the `emojify` helper.

lib/workers/repository/common.ts
```diff
@@ -10,9 +11,24 @@ export function extractRepoProblems(
...
+  [WARN]: ':warning:',
+  [ERROR]: ':x:',
+  [FATAL]: ':skull:',
+};
+
+export function formatProblemLevel(level: number): string {
+  const name = nameFromLevel[level].toUpperCase();
+  const unicode = logLevelEmojis[level];
+
+  return unicode ? `${emojify(unicode)} ${name}` : name;
```
jamietanna:
```suggestion
  const emojiName = logLevelEmojis[level];

  return emojiName ? `${emojify(emojiName)} ${name}` : name;
```

Or something similar, because it's now not actually the `unicode`

jamietanna (commented):
Coverage, comment, and otherwise LGTM!

jamietanna (approved):

PR #39798, opened 2025-12-05
feat(manager/gomod): inject credentials in GOPROXY environment variable

Shegox:
## Changes

This PR adds support for injecting credentils from the `hostType: go-proxy` into the `GOPROXY` enviroment variable when updating go modules. 

This allows one to use Renovate to update dependencies which only exist in an private, authenticated GOPROXY and not e.g. in the backing version control system e.g. in [Artifactory](https://jfrog.com/help/r/jfrog-artifactory-documentation/connect-your-go-client-to-artifactory). 

Lookups in the go datasource already work with authentication and `hostType: go-proxy`. 

Additionally this PR adds some documentation for this. 
In the tests I switched from mocking the `hostRule` resolutations to using the actual `hostRule` implementation allowing for easier to read and update tests.

## Context

## AI assistance disclosure

- [x] Yes — other (please describe): Docs have been refined via GPT-5

The public repository: n/a, run on a private repository on GitHub Enterprise Server with a private JFROG artifactory


lib/modules/manager/gomod/artifacts.ts
```diff
@@ -195,11 +197,15 @@ export async function updateArtifacts({
...
     const env = getEnv();
+    // add auth of 'go-proxy' urls to URLs in GOPROXY
+    const goproxy = getGoProxyWithAuth(env.GOPROXY);
+
     const execOptions: ExecOptions = {
       cwdFile: goModFileName,
       extraEnv: {
         GOPATH: await ensureCacheDir('go'),
-        GOPROXY: env.GOPROXY,
+        // only set extraEnv GOPROXY if it doesn't have injected credentials
```
secustor, line 212: Why is this split necessary?

Shegox: As far as I understood it `extraEnv` is only taken if there doesn't exist an environment variable yet. 

So if you have set `env` in your repository and set `extraEnv` here with the injected GOPROXY the uninjected one from `env` is taken. 
Therefore if there are injected GOPROXY I pass them in `env` rather than `extraEnv`. 
If there are no injected credentials the current behavior is preserved. 
I choose this to ensure that there is a slimmer chance of breaking as the current behavior without injected credentials will stay the same.

We can switch as well to alway setting it as `env.GOPROXY` and that should probably be fine. What do you think?

lib/modules/manager/gomod/artifacts.spec.ts
```diff
@@ -933,18 +932,17 @@ describe('modules/manager/gomod/artifacts', () => {
 
   it('supports docker mode with credentials', async () => {
     fs.findLocalSiblingOrParent.mockResolvedValueOnce('vendor');
-    GlobalConfig.set({ ...adminConfig, binarySource: 'docker' });
-    hostRules.find.mockReturnValueOnce({
+    hostRules.add({
```
secustor, line 935: Please move this test refactors to a separate PR

lib/modules/manager/gomod/readme.md
```diff
@@ -76,6 +76,40 @@ Rules from this list are converted to environment variable directives if they ma
...
+
+```json
+{
+  "env": {
+    "GOPROXY": "https://private-goproxy.example.com,https://proxy.golang.org,direct",
+    "GONOSUMDB": "my-private-go-packages/*"
+  },
+  "hostRules": [
+    {
+      "hostType": "go-proxy",
```
secustor, line 104:
```suggestion
      "hostType": "go",
```

We follow to have here datasources and platforms usually. 
Do you see a use case where users want to use different credentials for the same urls?

lib/modules/manager/gomod/artifacts.ts
secustor: `getEnv` should already return the user space variables 🤔 

So setting `GOPROXY:  goproxy` should be enough

viceice (commented):
conflicted

PR #39824, opened 2025-12-06
fix(composer): do not use `--minimal-changes` with `lockFileMaintenance` and add `postUpdateOption` for not using it

### LINKED ISSUE #39823: Make Composer CLI --minimal-changes flag a configuration option
_Discussed in originally posted discussion by @p-golovin in_ https://github.com/renovatebot/renovate/discussions/37767

The `--minimal-changes` flag in the Composer commands makes lock file maintenance miss a lot of updates. Additionally, for years this flag wasn't part of Renovate's behavior, and it is still not a default behavior for Composer versions >2.7; rather, it is an optional flag.

I suspect a lot of people do not realize that they are missing updates. We noticed this when Renovate suddenly stopped creating PRs across multiple projects, even though `composer outdated` was still yielding results.

Given that this is an optional flag and breaks Renovate's previous behavior, I propose making it a configurable option rather than a default behavior.

elaman:
## Changes

Currently `--minimal-changes` flag is being applied to `composer` commands. It was recently introduced in PR #34218. This has caused following issue #39823. As a solution, new `composerMinimalChanges` configuration is introduced, which allows `--minimal-changes` to be optionally applied instead of being a default behavior. 

## Context

- [x] This closes an existing Issue, Closes: #39823 

## AI assistance disclosure

The public repository: <URL>


bobvandevijver:
Thank you for opening this PR. I was just researching why renovate stopped creating the desired lock file maintenance and found this exact flag being used in the log. For completeness, composer even notifies about this being a mistake in this command.

```
The --minimal-changes option should be used with package arguments or after modifying composer.json requirements, otherwise it will likely not yield any dependency changes.
```

However, I do not think it should be made configurable in the way you suggest. It is great to use with the actual update commands, but it should never be used with the lock file maintance action.

elaman:
@bobvandevijver my initial thought was to just undo the flag introduction. But given than it was already in the codebase, I thought of easy way of undoing it without removing it completely for folks who think its useful. This PR could be a first step, after which it can be reviewed and polished out.

docs/usage/configuration-options.md
```diff
@@ -766,6 +766,12 @@ Set to `null` (not recommended) to fully omit `--ignore-platform-reqs/--ignore-p
 This requires the Renovate image to be fully compatible with your Composer platform requirements in order for the Composer invocation to succeed, otherwise Renovate will fail to create the updated lock file.
 The Composer output should inform you about the reasons the update failed.
 
+## composerMinimalChanges
+
+If set to `true`, it is the equivalent of passing the `--minimal-changes` option to `update`, `require` or `remove` Composer commands.
+
+With this flag Composer will only perform absolutely necessary changes to dependencies. If packages cannot be kept at their currently locked version they are updated. For partial updates the allow-listed packages are always updated fully.
```
```suggestion
With this flag Composer will only perform absolutely necessary changes to dependencies.
If packages cannot be kept at their currently locked version they are updated. For partial updates the allow-listed packages are always updated fully.
```

RahulGautamSingh (changes requested):
Add the type to `RenovateConfig` instead .. it should get passed down from there where needed.

elaman:
> Add the type to `RenovateConfig` instead .. it should get passed down from there where needed.

Can you please be more specific? Thank you.

RahulGautamSingh:
> Can you please be more specific? Thank you.

Find the `RenovateConfig` interface in the code and add the `composerMinimalChanges` field in it. After this the rest of the type changes won't be necessary

talkinnl:
> Thank you for opening this PR. I was just researching why renovate stopped creating the desired lock file maintenance and found this exact flag being used in the log. For completeness, composer even notifies about this being a mistake in this command.
> 
> ```
> The --minimal-changes option should be used with package arguments or after modifying composer.json requirements, otherwise it will likely not yield any dependency changes.
> ```
> 
> However, I do not think it should be made configurable in the way you suggest. It is great to use with the actual update commands, but it should never be used with the lock file maintance action.

I also don't think it should be a config option. But even if it was, I think the old behavior should be the default.

The way the flag was added recently just completely disables using 'Lock file maintenance' for security updates of indirect dependencies for PHP.

elaman:
@talkinnl indeed the goal of the PR is to restore pre-#39823 behavior without --minimal-changes.

docs/usage/configuration-options.md
```diff
@@ -766,6 +766,13 @@ Set to `null` (not recommended) to fully omit `--ignore-platform-reqs/--ignore-p
 This requires the Renovate image to be fully compatible with your Composer platform requirements in order for the Composer invocation to succeed, otherwise Renovate will fail to create the updated lock file.
 The Composer output should inform you about the reasons the update failed.
 
+## composerMinimalChanges
```
viceice: add a new post update option instead

viceice (changes requested):
I think the simpler solution is to delete the lockfile before running lockfile maintenance like we do on allmost all other managers. 🤔

@jamietanna @secustor wdyt?

elaman:
@viceice and someone else please correct me if I'm wrong, but removing lock file would change the "content hash" resulting in a new PR even if there are no updates. While perhaps not bad, this is different from the behavior Renovate had before this issue. My sole aim here is/was to run composer update without that flag, as it does very good job as is without removing a lock file. That being said I'm open to any quick solution that will restore lock file maintenances ability to apply updates (sometimes security related) of unspecified dependencies on schedule.

talkinnl:
> @viceice and someone else please correct me if I'm wrong, but removing lock file would change the "content hash" resulting in a new PR even if there are no updates. 

No, the content hash relates to composer.json contents, so that composer can detect the lock being stale. :)

--

When doing lock maintenance, the flag should just never be used as it just breaks the whole intention of the lock update, so if someone want's the flag/config there, the lock maintenance feature is not for them.
So the config option should only relate to commands during direct dependency Renovate PRs.

elaman:
@talkinnl good to know, thanks. @viceice I've moved the flag to be used only via postUpdateOptions, which applies for regular updates only. Also added the lock file removal.

docs/usage/configuration-options.md
```diff
@@ -3848,6 +3848,7 @@ Table with options:
 | ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `bundlerConservative`        | Enable conservative mode for `bundler` (Ruby dependencies). This will only update the immediate dependency in the lockfile instead of all subdependencies. |
 | `composerWithAll`            | Run `composer update` with `--with-all-dependencies` flag instead of the default `--with-dependencies`.                                                    |
+| `composerMinimalChanges`     | Run `composer update` with `--minimal-changes` flag for (except for lock file maintanance).                                                                |
```
viceice: this is a breaking change because it's now opt-in.

@jamietanna @secustor I think we should invert it to not be breaking?

eg: `composerNoMinimalChanges`

bobvandevijver: The argument addition was also not marked as a breaking change while forcing new (and changed) behaviour on users. For me this change should actually be considered a bugfix to restore the original behaviour.

talkinnl: Yes, it's a restore. The original PR was the BC change, and in hindsight that PR should really not have gotten away with 'no documentation needed' too.

elaman: Aim of PR is to restore the previous behavior and fix the issue with minimal changes flag. I don't think it should be inverted.

jamietanna: > The original PR

Just to confirm here, we're talking about https://github.com/renovatebot/renovate/pull/34218 from March?

docs/usage/configuration-options.md
```diff
@@ -3849,6 +3849,7 @@ Table with options:
...
```
```suggestion
| `composerMinimalChanges`     | Run `composer update` with `--minimal-changes` flag for (except for lock file maintenance).                                                                |
```

docs/usage/configuration-options.md
jamietanna: Understood that when it was introduced (in March) it resulted in a break in behaviour, and this is a revert back to the previous behaviour. However, given the age, I'd also prefer to make this a non-breaking change

I think let's make this an opt-in change for now, requiring repo config to use that behaviour. 

We've got a major release planned for January (I still need to write the announcement) so at that point we can flip it to be the default

So the defaults will be:

- Now: `composerNoMinimalChanges: false`
- January: `composerNoMinimalChanges: true`

elaman: @jamietanna understood. Can we at least keep the variable name `composerMinimalChanges` and reverse the default value to `true` at least?

jamietanna: Sounds good to me - @viceice would you be happy with the following:

So the defaults will be:

- Now: `composerMinimalChanges: true`
- January: `composerMinimalChanges: false`

(note we've removed the inversion, but are handling it with a change to the default in the next major)

elaman: @jamietanna actually I don't see anywhere where default postupdateoptions are set. So perhaps composerNoMinimalChanges variable is not a bad idea after all.

viceice (dismissed):

viceice:
missing code coverage

lib/modules/manager/composer/artifacts.ts
```diff
@@ -172,6 +173,21 @@ export async function updateArtifacts({
       commands.push('git stash pop || true');
     }
 
+    if (config.isLockFileMaintenance) {
```
jamietanna: should this be a separate change?

elaman: It was suggested at some point, but I don't mind putting it into a different PR if issue for it is created.

jamietanna: Yeah I'd prefer it to be separate, please - thanks!

elaman: @jamietanna done. Created the other PR here: #39920

docs/usage/configuration-options.md
```diff
@@ -3885,6 +3885,7 @@ Table with options:
 | ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `bundlerConservative`        | Enable conservative mode for `bundler` (Ruby dependencies). This will only update the immediate dependency in the lockfile instead of all subdependencies. |
 | `composerWithAll`            | Run `composer update` with `--with-all-dependencies` flag instead of the default `--with-dependencies`.                                                    |
+| `composerNoMinimalChanges`   | Run `composer update` with no `--minimal-changes` flag (except for lock file maintenance, where it is a default behavior).                                 |
```
```suggestion
| `composerNoMinimalChanges`   | Run `composer update` with no `--minimal-changes` flag (does not affect lock file maintenance, which will never use `--minimal-changes`).            |
```

docs/usage/configuration-options.md
```diff
@@ -3885,6 +3885,7 @@ Table with options:
 | ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `bundlerConservative`        | Enable conservative mode for `bundler` (Ruby dependencies). This will only update the immediate dependency in the lockfile instead of all subdependencies. |
 | `composerWithAll`            | Run `composer update` with `--with-all-dependencies` flag instead of the default `--with-dependencies`.                                                    |
+| `composerNoMinimalChanges`   | Run `composer update` with no `--minimal-changes` flag (does not affect lock file maintenance, which will never use `--minimal-changes`).            |
```
```suggestion
| `composerNoMinimalChanges`   | Run `composer update` with no `--minimal-changes` flag (does not affect lock file maintenance, which will never use `--minimal-changes`).                  |
```

jamietanna (approved):
I think given there are two things we're doing, we could've split this further into a `fix: don't use with lockFileMaintenance` and a `feat(...): add postupdateOption` but this is fine 👍

Issue #39994, opened 2025-12-15
Allow specifying "maximum" values for user-configurable values in self-hosting config


jamietanna:
There are a number of repo config options that can have significant impact to the runtime of the CLI, depending on how it's deployed.

For instance, running the CLI in autodiscovery mode, or using another "single-threaded" option can mean that the longer a single repository takes, the longer the whole execution of the CLI (against all repos) takes, overall.

For configuration such as [`prConcurrentLimit`](https://docs.renovatebot.com/configuration-options/#prconcurrentlimit) and [`branchConcurrentLimit`](https://docs.renovatebot.com/configuration-options/#branchconcurrentlimit), which are very useful for tuning how many updates a given repository can have.

However, when running self-hosted, this can lead to significant impact if you have a big repository with a lot of dependencies, resulting in lots of branches being created by Renovate, possibly taking up time that other repos will not be able to process for (i.e. if running Mend Renovate Community Self-Hosted, or based on Mend Renovate Cloud limits)

This feature would:

- allow setting global self-hosting configuration (i.e. `limits.branchConcurrentLimit` or `maximumLimits.branchConcurrentLimit`) to provide a _maximum_ limit that this value can be at
- a `DEBUG` (~~or a `WARN`?~~) would be logged when overriding this compared to a repo's config

This could also include [`hostRules.concurrentRequestLimit`](https://docs.renovatebot.com/configuration-options/#concurrentrequestlimit)

(As discussed in https://github.com/renovatebot/renovate/discussions/39433)

jamietanna:
Via https://github.com/renovatebot/renovate/discussions/44009 this would be useful as a `minimum` limit too

PR #40086, opened 2025-12-20
feat: inherit support for `onboardingAutoCloseAge`

### LINKED ISSUE #40011: Support `onboardingAutoCloseAge` in inherited config

Ref: https://github.com/renovatebot/renovate/pull/39876#issuecomment-3656686704

RahulGautamSingh:
## Changes
- Support using `onboardingAutoCloseAge` in inherited config.

## Context

- [x] This closes an existing Issue, Closes: #40011 

## AI assistance disclosure

The public repository: 
1. inherit: https://github.com/Rahul-renovate-testing/test-inherit/pull/5
2. normal: https://github.com/Rahul-renovate-testing/test-onb-aca/pull/1


lib/workers/repository/onboarding/branch/check.ts
```diff
@@ -160,10 +159,9 @@ export async function isOnboarded(config: RenovateConfig): Promise<boolean> {
   logger.debug('Repo is not onboarded and no merged PRs exist');
   if (!config.suppressNotifications!.includes('onboardingClose')) {
     const ageOfOnboardingPr = getElapsedDays(closedOnboardingPr.createdAt!);
-    const onboardingAutoCloseAge = GlobalConfig.get('onboardingAutoCloseAge');
```
viceice, line 166: why not still store as variable to have less changes?

lib/config/options/index.ts
```diff
@@ -243,6 +243,7 @@ const options: RenovateOptions[] = [
     type: 'integer',
     default: null,
     globalOnly: true,
+    inheritConfigSupport: true,
```
viceice, line 247: global only and inherited are somewhat mutually exclusive 🤔

@jamietanna I don't like to mix-up them more than necessary

jamietanna: Yes I think that's fair - in that case, we'd want only `inheritConfigSupport`?

jamietanna: Should we also have a test to stop them being set together? -> #40114

RahulGautamSingh: inherited config can contain all repo config options and only some global options, so it makes sense to use both `globalOnly` and `inheritConfigSupport` together.

jamietanna (commented):
I do wonder if we should make it so the inheritConfig version of the setting should be <= what the global settings are

I.e. so the admin can set `onboardingAutoCloseAge=366`, and then orgs can do `onboardingAutoCloseAge=7` but _not_ `onboardingAutoCloseAge=1000`

RahulGautamSingh:
> I do wonder if we should make it so the inheritConfig version of the setting should be <= what the global settings are
> 
> I.e. so the admin can set `onboardingAutoCloseAge=366`, and then orgs can do `onboardingAutoCloseAge=7` but _not_ `onboardingAutoCloseAge=1000`

A good to have, IMO. Can do in a follow-up PR.

jamietanna:
> A good to have, IMO. Can do in a follow-up PR.

I'd prefer we get it in as part of this - I don't want to then change the behaviour once it's released

lib/config/options/index.ts
jamietanna: Discussed offline that we'll work to determine a better solution here (medium term) to work with https://github.com/renovatebot/renovate/issues/39685's approach

jamietanna (changes requested):
Need to add the logic for "inherited config can't have a higher value than global value" and then this LGTM

jamietanna (changes requested):
Do we need to have `onboardingAutoCloseAgeLimit`? I'd prefer that we infer it from the `onboardingAutoCloseAge`

If it's down to the `config.onboardingAutoCloseAge` pulling the inherited config version, can we check `GlobalConfig.get('onboardingAutoCloseAge')`?

RahulGautamSingh:
> Do we need to have `onboardingAutoCloseAgeLimit`? I'd prefer that we infer it from the `onboardingAutoCloseAge`
> 
> If it's down to the `config.onboardingAutoCloseAge` pulling the inherited config version, can we check `GlobalConfig.get('onboardingAutoCloseAge')`?

Will need to add `onboardingAutoCloseAge` to `RepoGlobalConfig` which is not good .. as we now have to manage the type issue caused by this here and in future when we finally decide on a way to separate the global only and inherited config options

jamietanna:
Mind addressing failures?

viceice (commented):
What about adding a new `InheritedConfig` class similar to `GlobalConfig` and fallback to `InheritedConfig` is global isn't set for the options which can be inherited and global only.

In this case here we still need to compare the values and use the global is it's lower than inherited value

jamietanna:
I think that sounds reasonable - on Slack we've been discussing about whether https://github.com/renovatebot/renovate/issues/39994 would be a good option for this, but I think it's different to those limits, where we've decided that `onboardingAutoCloseAge` shouldn't ever be higher than the global config

RahulGautamSingh:
Blocked by: https://github.com/renovatebot/renovate/issues/40446

jamietanna:
Should now be unblocked 🚀

lib/workers/repository/onboarding/branch/check.spec.ts
```diff
@@ -170,6 +171,30 @@ describe('workers/repository/onboarding/branch/check', () => {
       );
       expect(platform.ensureComment).not.toHaveBeenCalled();
     });
+
+    it('prefers inherited onboardingAutoCloseAge over global config', async () => {
+      const now = DateTime.now();
+      vi.setSystemTime(now.toMillis());
```
jamietanna, line 182: do we need an `afterEach` to reset this? Or does Vitest do it for us?

jamietanna (dismissed):
LGTM - mind addressing conflicts, then we'll get this merged 👍

jamietanna (approved):

jamietanna:
FYI this doesn't quite restrict the configuration as we'd planned for: https://github.com/renovatebot/renovate/pull/40810

PR #40098, opened 2025-12-22
feat: Add `exact` versioning

zharinov:
## Changes

- Required for #40095

## Context

## AI assistance disclosure

The public repository: <URL>


viceice (dismissed):

RahulGautamSingh (approved):

jamietanna (commented):
I wonder if we can think of an alternate name here - I think `dummy` here could be confusing (I wasn't sure what it was looking at the naming alone, I needed to read the docs)

I don't have an immediate suggestion right now - will try and come back later to suggest

zharinov:
`samever` 😁

jamietanna:
A few other suggested names:

- `samever` (I like your suggestion, Sergei - good pun on SemVer)
- `string-equality` (to make it very explicit)

Co-authored-by: gemini-2.5-pro (GitHub Copilot):

- `exact`
- `literal`
- `verbatim` (might be a bit too "wordy")
- `identity` (might be a bit too "wordy")

Co-authored-by: gpt-4.1 (GitHub Copilot):

- `static`

jamietanna:
@viceice @secustor thoughts on naming, here?

viceice:
I don't like dummy, but didn't have a better idea

jamietanna:
Any preferences based on the suggested list above?

secustor:
I like `static`. 

Though I loled with `samever`, this will be misread and will confuse users.

jamietanna (changes requested):
++ I think `static` sounds good as the versioning name - mind amending?

zharinov:
Let's avoid language keywords. I'm renaming to `exact`.

lib/modules/versioning/exact/index.ts
```diff
@@ -0,0 +1,72 @@
...
+    return this.isValid(version);
+  }
+
+  isStable(_version: string): boolean {
+    return true;
+  }
+
+  isCompatible(version: string): boolean {
+    return this.isValid(version);
+  }
```
jamietanna: This seems a little surprising - should this instead be the case that no two versions are compatible with each other?

zharinov: It doesn't affect behavior, but you're right. Fixed it.

lib/modules/versioning/exact/index.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  describe('isCompatible', () => {
+    it('returns true when version equals current', () => {
+      expect(exact.isCompatible('v1', 'v1')).toBe(true);
+    });
+
+    it('returns false when version differs from current', () => {
+      expect(exact.isCompatible('v1', 'v2')).toBe(false);
+    });
```
jamietanna, line 52: Could we use two versions that would be SemVer compatible here? 

I.e.

```suggestion
    it('returns false when version differs from current', () => {
      expect(exact.isCompatible('v1.0.0', 'v1.1.0')).toBe(false);
    });
```

This makes it clear that, unlike SemVer, things aren't compatible even if they look like they should be

jamietanna (changes requested):
One last comment, otherwise LGTM!

jamietanna (approved):

viceice (approved):

PR #40149, opened 2025-12-24
fix: avoid proposing replacements when the new dependency already exists

felipecrs:
## Changes

When Renovate replaces a dependency but the new dependency already exists (for whatever reason), Renovate would fail with `update failure`.

This PR changes the behavior to simply avoid proposing the update.

**Before:**

image

**After:**

image

## Context

## AI assistance disclosure

The public repository: https://github.com/felipecrs/renovate-repro-npm-replacement/pulls


felipecrs:
I don't think these coverage issues are related to this PR:

- https://app.codecov.io/gh/renovatebot/renovate/pull/40149

Can it be bypassed or do I need to build tests for the existing code? Or maybe I'm missing something?

viceice (commented):
coverage errors are related because you changed the file. please add test cases

felipecrs:
@viceice it wasn't straightforward to create tests for these lines, so I added ignore comments instead.

I hope it's ok, but please let me know if not.

RahulGautamSingh (commented):
The reasoning seem un-intuitive for me. 

If new name & old name both exist we should fail to signal something is wrong with the current state of the package file, instead of removing, cause replacing & removing aren't the same.

felipecrs:
I agree this is un-intuitive at first. But here is why I think it's a better option:

1. The current error message is very generic and doesn't give any clue at all on why it failed (see this PR's body for the error message).
2. The current error message can only be seen by the person administering Renovate, which has access to the logs or the reports.
   - Unless the dependency dashboard is enabled, in which case it is more accessible (although still unclear):

   image

   - However, not all platforms have dependency dashboard, namely Gerrit.

> If new name & old name both exist we should fail to signal something is wrong with the current state of the package file

Just to clarify, there's nothing wrong with the state of the package file itself. An old and new dependency can coexist just fine. Imagine a big repository, for example. You could have refactored some parts of it to use the new dependency, while still depending on the old one for other parts that are still missing to refactor.

> cause replacing & removing aren't the same.

Note the PR is only removing the old dependency because the replacement is already in place. It's not just removing without context, and the PR body makes that clear enough.

Having that PR makes it much easier for the reviewer to understand Renovate's intention than with the failure, even if the repository has a dependency dashboard.

---

I'm not claiming this solution is perfect. I just think it's better than the current approach. Ideally Renovate would create a Removal PR, rather than a Replacement, but that is not a thing yet.

Please let me know your thoughts.

I will understand if you prefer not to merge this PR, though.

RahulGautamSingh (commented):
> Imagine a big repository, for example. You could have refactored some parts of it to use the new dependency, while still depending on the old one for other parts that are still missing to refactor.

In this case, Renovate should not remove the old dep then, right?

felipecrs:
> In this case, Renovate should not remove the old dep then, right?

I think it should, to point developers that the old (deprecated) dependency is still in their codebase and should be handled.

felipecrs:
This PR is good to go from my side.

RahulGautamSingh:
> I think it should, to point developers that the old (deprecated) dependency is still in their codebase and should be handled.

Sorry, but that seems out of Renovate's scope.

felipecrs:
Maybe Renovate should simply bail out then, rather than failing? In case the new replacement is already in place.

RahulGautamSingh:
> Maybe Renovate should simply bail out then, rather than failing? In case the new replacement is already in place.

Yes, that make more sense. Log that replacement already exists and skip the branch modification. I think this is what's happening, just the error might not be surfaced.

felipecrs:
I've reworked this PR completely. I kindly ask you to read the PR description again.

- It no longer removes the old dependency.
- It's no longer npm-specific.
- If the replacement is already in place, the branch will no longer show in the Dependency Dashboard.
  - But there's still a warning that you have deprecated dependencies, which is nice.

After iterating on it, I realized this would be the best implementation from a UX point of view.

felipecrs:
Oh, wait a sec. This looks misleading:

image

felipecrs:
That's fixed now. Updated PR description with new screenshots (before/after).

What do you think about this approach?

lib/workers/repository/updates/flatten.ts
```diff
@@ -112,6 +113,16 @@ export async function flattenUpdates(
           delete depConfig.deps;
           depConfig.depIndex = depIndex; // used for autoreplace
           for (const update of dep.updates!) {
+            if (
+              update.updateType === 'replacement' &&
+              update.newName &&
+              replacementAlreadyExists(packageFile.deps, dep, update.newName)
+            ) {
+              logger.debug(
```
secustor, line 121:
```suggestion
              logger.debug({ fileName: packageFile.packageFile },
```
Add the name as structured metadata to make it easier to filter

lib/workers/repository/dependency-dashboard.ts
```diff
@@ -391,14 +391,25 @@ export async function ensureDependencyDashboard(
...
+                dep,
+                replacementUpdate.newName,
+              )
+            );
+          if (name && (dep.deprecationMessage ?? hasAnyReplacement)) {
             hasDeprecationsOrReplacements = true;
             deprecatedPackages[manager] ??= {};
             deprecatedPackages[manager][name] ??= {
-              hasReplacement,
+              hasReplacement: hasActionableReplacement,
```
secustor:
```suggestion
              hasActionableReplacement,
```
I would prefer if this variable is renamed too as you are changing the condition in case this is true.

secustor (commented):
else LGTM

secustor (approved):

viceice (approved):

jamietanna:
Possibly related: https://github.com/renovatebot/renovate/discussions/44559

PR #40168, opened 2025-12-26
fix(cache): use `bitbucket-server` as platform cache key

### LINKED ISSUE #39895: cache: use `platform` as the cache keys
It was recently noticed that [we use a camelCase variant of the platform name in `RepoCacheData`](https://github.com/renovatebot/renovate/blob/db824b9c85d3d2ceb8ec06a226d8e50ed945a2ae/lib/util/cache/repository/types.ts#L159-L161) only for `bitbucketServer`.

`bitbucketServer` should be cased as `bitbucket-server` (https://github.com/renovatebot/renovate/blob/db824b9c85d3d2ceb8ec06a226d8e50ed945a2ae/lib/constants/platforms.ts#L1-L13) as that matches the platform name.

We should fix this, so it's a 1-1 mapping between a platform name and a cache key _or_ make sure that there's a way to pass a `platform` and receive the cache settings for that platform.

This can either be done as a breaking change (in the next major) or as a non-breaking change for now, and then made breaking in the future, for consistency.

zharinov:
## Changes

Change the Bitbucket Server platform cache key from `bitbucketServer` (camelCase) to `bitbucket-server` (kebab-case) to match the platform ID in `PLATFORM_HOST_TYPES`.

Adds migration logic to automatically convert existing caches using the old key to the new format, ensuring a non-breaking upgrade path.

## Context

- [x] This closes an existing Issue, Closes: #39895

## AI assistance disclosure

Claude Code was used to implement the fix, write tests, and create this PR.

### DIFF
--- a/lib/modules/platform/bitbucket-server/pr-cache.ts
+++ b/lib/modules/platform/bitbucket-server/pr-cache.ts
@@ -11,11 +11,7 @@ import { prInfo } from './utils';
 
 /* v8 ignore next */
 function migrateBitbucketServerCache(platform: unknown): void {
-  if (!isPlainObject(platform)) {
-    return;
-  }
-
-  if (!isPlainObject(platform.bitbucketServer)) {
+  if (!isPlainObject(platform?.bitbucketServer)) {
     return;
   }
 
--- a/lib/modules/platform/bitbucket-server/pr-cache.ts
+++ b/lib/modules/platform/bitbucket-server/pr-cache.ts
@@ -1,26 +1,27 @@
-import { isNullOrUndefined, isString } from '@sindresorhus/is';
+import { isNullOrUndefined, isPlainObject, isString } from '@sindresorhus/is';
 import { dequal } from 'dequal';
 import { DateTime } from 'luxon';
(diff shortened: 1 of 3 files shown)


lib/modules/platform/bitbucket-server/pr-cache.ts
```diff
@@ -9,6 +9,20 @@ import { getQueryString } from '../../../util/url';
...
 
+/* v8 ignore next */
+function migrateBitbucketServerCache(platform: unknown): void {
+  if (!isPlainObject(platform)) {
+    return;
+  }
+
+  if (!isPlainObject(platform.bitbucketServer)) {
+    return;
+  }
```
RahulGautamSingh, line 20:
```suggestion
  if (!isPlainObject(platform?.bitbucketServer)) {
    return;
  }
```

works?

RahulGautamSingh (changes requested):
Else, LGTM

RahulGautamSingh (commented):
lint issue 😖

viceice (dismissed):

viceice (approved):

PR #40170, opened 2025-12-26
fix(auto-replace): prevent digest corruption when using replacementNameTemplate

zharinov:
## Changes

When using `replacementNameTemplate` with Docker images that have pre-pinned digests, the auto-replace logic was incorrectly entering the `currentDigestShort` replacement branch even when the digest didn't need to be changed.

Since `currentDigest === newDigest`, the first digest replacement block was skipped. But the `else if` checked only `currentDigestShort !== newDigest`, which was true (e.g., `"5fa2edb" !== "sha256:5fa2edb..."`), causing it to replace the short string with the full digest, corrupting the result to `sha256:sha256:...` format.

The fix adds a guard condition to skip the `currentDigestShort` replacement when the full `currentDigest` already matches `newDigest`.

## Context

- [x] This closes an existing Issue, Closes: #40169

## AI assistance disclosure

Used Claude Code to investigate the bug, identify the root cause, implement the fix, and write the regression test.

The public repository: https://github.com/renovate-testing/renovate-dockerfile-replacement-reproduction

### DIFF
--- a/lib/workers/repository/update/branch/auto-replace.spec.ts
+++ b/lib/workers/repository/update/branch/auto-replace.spec.ts
@@ -1547,6 +1547,37 @@ describe('workers/repository/update/branch/auto-replace', () => {
       );
     });
 
+    it('docker: replacement with same digest should not corrupt digest via currentDigestShort', async () => {
+      // Regression test for https://github.com/renovatebot/renovate/discussions/38703
+      // When doing a replacement where currentDigest === newDigest, the currentDigestShort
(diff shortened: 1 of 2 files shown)


zharinov:
- https://github.com/renovate-testing/renovate-dockerfile-replacement-reproduction/pull/1
- https://github.com/renovate-testing/renovate-dockerfile-replacement-reproduction/pull/2

RahulGautamSingh (approved):

viceice (approved):
please don't remove PR template part's

PR #40186, opened 2025-12-28
refactor(tools): reduce noise in generate-imports script

zharinov:
## Changes

Remove verbose console.log statements from `generate-imports.mjs` to reduce noise during build process. Use `console.error` for actual error messages.

## Context

## AI assistance disclosure

- [x] Code inspection only

### DIFF
--- a/tools/generate-imports.mjs
+++ b/tools/generate-imports.mjs
@@ -4,16 +4,13 @@ import { minimatch } from 'minimatch';
 import upath from 'upath';
 import { hashFile, hashFromArray } from './utils/hash.mjs';
 
-console.log('generating imports');
 const newFiles = new Set();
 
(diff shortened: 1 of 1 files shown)


tools/generate-imports.mjs
```diff
@@ -201,7 +196,7 @@ async function generateHash() {
       [hashMap, hashStrings.join('\n')].join('\n\n'),
     );
   } catch (err) {
-    console.log('ERROR:', err.message);
+    console.error('ERROR:', err.message);
```
RahulGautamSingh: we should mention when the error occured ie. while generating hashes/data/import

RahulGautamSingh (changes requested):

viceice (approved):

PR #40208, opened 2025-12-30
fix(managers/pep621): use lowest locked version from uv

### LINKED ISSUE #40176: Unexpected differentiation behavior between minor/major with uv
### Discussed in https://github.com/renovatebot/renovate/discussions/40166

<sup>Originally posted by **jvllmr** December 26, 2025</sup>

A Mend.io-hosted app

GitHub.com

42.59.0

In my project I have a group that contains all pep621 minor and patch updates: https://github.com/jvllmr/flay/blob/917b2e2921b381c9eb340cb939f9ed1b59ae9c50/.github/renovate.json#L49

However, in the following PR `pytest` in the `pyproject.toml` receives a major update within a grouped PR: https://github.com/jvllmr/flay/pull/270 (pytest is upgraded from `>=8.4.2` to `>=9.0.2`).
I would expect that there is a separate PR for the pytest major update.

When looking at the log you can see that uv last locked pytest to version `9.0.1` (probably because I generated that lockfile locally). I think renovate considers this a minor update because of this. Could this be changed so that renovate can resolve such situation? That would be great.

Reproduction repo: https://github.com/jvllmr/renovate-discussion-40166

### Logs (if relevant)

```
{
            "packageName": "pytest",
            "depName": "pytest",
            "datasource": "pypi",
            "depType": "dependency-groups",
            "currentValue": ">=8.4.2",
            "managerData": {
              "depGroup": "test"
            },
            "lockedVersion": "9.0.1",
            "updates": [
              {
                "bucket": "major",
                "newVersion": "9.0.2",
                "newValue": ">=9.0.2",
                "releaseTimestamp": "2025-12-06T21:30:49.000Z",
                "newVersionAgeInDays": 19,
                "newMajor": 9,
                "newMinor": 0,
                "newPatch": 2,
                "updateType": "patch",
                "isBreaking": false,
                "isRange": true,
                "isBump": true,
                "libYears": 0.2555869165398275,
                "branchName": "renovate/python-minor-patch"
              }
            ],
            "versioning": "pep440",
            "warnings": [],
            "sourceUrl": "https://github.com/pytest-dev/pytest",
            "registryUrl": "https://pypi.org/pypi",
            "changelogUrl": "https://docs.pytest.org/en/stable/changelog.html",
            "mostRecentTimestamp": "2025-12-06T21:30:49.000Z",
            "currentVersion": "9.0.1",
            "currentVersionTimestamp": "2025-09-04T14:34:20.000Z",
            "currentVersionAgeInDays": 113,
            "isSingleVersion": false,
            "fixedVersion": "9.0.1"
          },
```

</details>
</div>

jvllmr:
## Changes

Unlike most other package managers, uv can lock multiple versions in its lockfile and renovate chose the last entry as the locked version. This PR makes sure that the lowest pep440 version found is used by renovate as locked version.

## Context

- [x] This closes an existing Issue, Closes: #40176 

## AI assistance disclosure

The public repository: <URL>


lib/modules/manager/pep621/processors/uv.spec.ts
```diff
@@ -356,6 +356,62 @@ describe('modules/manager/pep621/processors/uv', () => {
...
+
+    it('resolves to lowest version found', async () => {
+      fs.findLocalSiblingOrParent.mockResolvedValueOnce('uv.lock');
+      fs.readLocalFile.mockResolvedValueOnce(codeBlock`
+      [[package]]
+      name = "pytest"
+      version = "8.4.2"
+      source = { registry = "https://pypi.org/simple" }
+      resolution-markers = [
+          "python_full_version < '3.10'",
```
secustor, line 368: Is this double locking happening because of the python version? 

@viceice Do you have thoughts regarding using the lowest locked version? 
I'm thinking we should maybe using the Python constraint to decide which lockedVersion to use here.

jvllmr: My thoughts on your questions:

> Is this double locking happening because of the python version?

This double locking is happening because the project has `requires-python = ">=3.9"` defined. Since pytest v9 is not usable with python 3.9 uv locks two different versions for 3.9 and >=3.10. Other package managers like PDM try to lock a single version that is compatible with all python versions or error out if none is found.

> I'm thinking we should maybe using the Python constraint to decide which lockedVersion to use here.

I also thought about taking the resolution markers into consideration, but how do we know which python version to choose? In this test case users could use every version from python 3.9 to python 3.14 for their local dev environment. Therefore, I guess it is almost impossible for renovate to predict what python version to use for evaluating these constraints.

> Do you have thoughts regarding using the lowest locked version?

This PR's purpose is to solve the problem that renovate with rangeStrategy bump considers a v9 upgrade a minor upgrade because renovate uses the locked version, which is currently 9.0.2 in this case, to do the version bumping. To fix this issue the lowest version inside the range must be used, which is most likely the lowest version in the lockfile. Tbh this only repairs a symptom and not the core issue of renovate using the locked version instead of the range's version for bumping, but I think this would be a good workaround compared to tackling the bigger issue and potentially causing new issues with other package managers. There could be issues with uv and other range strategies or with lockfile maintenance though... but no concrete issue came to my mind yet. Just my thoughts on this... I'm not the maintainer here 😅

viceice (dismissed):
looks ok to me as intermediate solution. I think we do something similar in other managers

lib/modules/manager/pep621/schema.ts
```diff
@@ -6,6 +6,7 @@ import {
 } from '../../../util/schema-utils/index.ts';
 import { normalizePythonDepName } from '../../datasource/pypi/common.ts';
 import { PypiDatasource } from '../../datasource/pypi/index.ts';
+import { api as pep440 } from '../../versioning/pep440.ts';
```
viceice:
```suggestion
import { api as pep440 } from '../../versioning/pep440/index.ts';
```

viceice (approved):

lib/modules/manager/pep621/schema.ts
```diff
@@ -228,8 +229,14 @@ export const UvLockfile = Toml.pipe(
...
-).transform(({ package: pkg }) =>
-  Object.fromEntries(
-    pkg.map(({ name, version }): [string, string] => [name, version]),
-  ),
-);
+).transform(({ package: pkgs }) => {
+  const pkgMap: Record<string, string> = {};
+
+  for (const { name, version } of pkgs) {
+    if (!(name in pkgMap) || pep440.isGreaterThan(pkgMap[name], version)) {
```
viceice, line 236: could a version be invalid to our pep440 schema? then this will throw 🤔

jvllmr: When I created this PR I installed several packages via uv from git and every time uv created a pep440 compliant version even for repositories with dynamic versioning. Therefore I don't expect non-compliant versions in a uv lock.

jvllmr: To double down on this: The only `Version` struct in uv's codebase is from their uv-pep440 crate, which implies to me that uv genereates only pep440 compliant versions.

https://github.com/astral-sh/uv/blob/6458a2b8a5a75a97e17e08f59d0fcad39bc78032/crates/uv-pep440/src/version.rs#L277

RahulGautamSingh (approved):
lgtm

viceice (approved):

PR #40235, opened 2025-12-31
fix(util/exec): always pass `shell` when preparing `rawExec` arguments

jamietanna:
## Changes

Noticed in #40233 and #40242.

As noted in #40242, without `shell` being passed around, this won't
correctly apply for Post Upgrade Tasks, and might break other usages.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/util/exec/index.ts
+++ b/lib/util/exec/index.ts
@@ -57,6 +57,7 @@ function getRawExecOptions(opts: ExecOptions): RawExecOptions {
     env: childEnv,
     maxBuffer,
     timeout,
+    ...(opts.shell !== undefined && { shell: opts.shell }),
     stdin: 'pipe',
     stdout: opts.ignoreStdout ? 'ignore' : 'pipe',
     stderr: 'pipe',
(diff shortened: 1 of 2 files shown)


lib/util/exec/index.ts
```diff
@@ -57,6 +57,7 @@ function getRawExecOptions(opts: ExecOptions): RawExecOptions {
     env: childEnv,
     maxBuffer,
     timeout,
+    shell: opts.shell,
```
jamietanna: Ideally we'd do `...opts`, but that then pulls in i.e. `ignoreStdout` at runtime, which isn't a valid option

I've spent a bit of time trying to make it work nicely, but given the fact that we only have 3 properties right now, maybe this sin't necessary

viceice:
```suggestion
    ...((opts.shell != undefined)&& {shell: opts.shell}),
```
this should reduce the test change noise

viceice:
```suggestion
    ...((opts.shell != undefined)&& {shell: opts.shell}),
```
this should reduce the test change noise

viceice (commented):
see comment

lib/util/exec/index.ts
jamietanna: Ah amazing, thank you!

viceice (approved):

secustor (approved):

PR #38142, opened 2025-09-21
feat: add apk datasource and apko manager support

punsoong-canva:
## Changes

This PR adds the `apk` datasource and [`apko`](https://github.com/chainguard-dev/apko/tree/main) manager support, allowing Renovate to provide version updates to the package files in [apko files](https://github.com/chainguard-dev/apko/blob/main/docs/apko_file.md), e.g.

```yaml
contents:
  repositories:
    - https://dl-cdn.alpinelinux.org/alpine/edge/main
  packages:
    - bash=5.2.37-r0 # -> 5.2.37-r33
```
This PR also adds support to update the lock files (e.g. `apko.lock.json`) generated by `apko lock apko.yaml`

## Context

Please select one of the below:

- [x] This closes an existing Issue: #38087
- [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

Used Cusor to generate code to model off the rpm and deb datasources as well as generate manager code + test.
Docs and test-input data had to be rewritten to match apk and apko references from:
 - https://wiki.alpinelinux.org/wiki/Alpine_Package_Keeper
 - https://wiki.alpinelinux.org/wiki/Package_policies#Package_versions
 - https://github.com/chainguard-dev/apko/tree/main/docs

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

The public repository: https://github.com/punsoong-canva/apko-renovate-reproduction

PRs raised based on this branch:
- https://github.com/punsoong-canva/apko-renovate-reproduction/pull/1 dependency update
- https://github.com/punsoong-canva/apko-renovate-reproduction/pull/2 lock file maintenance

PR #40250, opened 2026-01-02
feat(datasource): add APK datasource support

punsoong-canva:
## Changes

This PR adds `apk` datasource support.
This PR is based off https://github.com/renovatebot/renovate/pull/38142

## Context

Part of https://github.com/renovatebot/renovate/discussions/38087 
After this PR is accepted and merged, the `datasource` followed by the `manager` support PRs will be raised

## AI assistance disclosure

Used Cursor to generate code to model off the rpm and deb datasources as well as generate manager code + test.
Docs and test-input data had to be rewritten to match apk and apko references from:
 - https://wiki.alpinelinux.org/wiki/Alpine_Package_Keeper
 - https://wiki.alpinelinux.org/wiki/Package_policies#Package_versions
 - https://github.com/chainguard-dev/apko/tree/main/docs

In the original PR this is based off https://github.com/renovatebot/renovate/pull/38142

The public repository: https://github.com/punsoong-canva/apko-renovate-reproduction

PRs raised based on this branch:
- https://github.com/punsoong-canva/apko-renovate-reproduction/pull/1 dependency update
- https://github.com/punsoong-canva/apko-renovate-reproduction/pull/2 lock file maintenance


lib/modules/datasource/apk/readme.md
```diff
@@ -0,0 +1,18 @@
...
+https://dl-cdn.alpinelinux.org/alpine/v3.19/community/x86_64/APKINDEX.tar.gz
+```
+
+Example APK repository URLs:
+
+- Official Alpine Linux repositories (e.g., `https://dl-cdn.alpinelinux.org/alpine/v3.19/main`)
+- Community repositories (e.g., `https://dl-cdn.alpinelinux.org/alpine/v3.19/community`)
+- Wolfi APK repositories (e.g., `https://packages.wolfi.dev/os`)
+
+This datasource is used by the [apko manager](../../manager/apko/index.md) to provide version information for packages defined in `apko.yaml` files.
```
punsoong-canva:
```suggestion
```
will re-add this in the next PR which will add the `apko` manager and the readme file.

jamietanna:
This should now be good to rebase off `main` and start on reviewing :)

lib/modules/datasource/apk/readme.md
```diff
@@ -0,0 +1,16 @@
+The APK datasource is designed to work with Alpine Linux package repositories. It can fetch package information from APK repositories and provide version updates for Alpine Linux packages.
+
+Alpine Linux uses the APK package manager, and packages are distributed through repositories. Each repository contains an `APKINDEX.tar.gz` file that contains metadata about all available packages.
```
jamietanna: (our preference is to break a paragraph over multiple lines - once line per sentence)

```suggestion
The APK datasource is designed to work with Alpine Linux package repositories.
It can fetch package information from APK repositories and provide version updates for Alpine Linux packages.

Alpine Linux uses the APK package manager, and packages are distributed through repositories.
Each repository contains an `APKINDEX.tar.gz` file that contains metadata about all available packages.
```

lib/modules/datasource/apk/parser.ts
```diff
@@ -0,0 +1,75 @@
...
+        const colonIndex = line.indexOf(':');
+        if (colonIndex === -1) {
+          continue;
+        }
+
+        // get fields based on the PKGINFO spec - https://wiki.alpinelinux.org/wiki/Apk_spec
+        const key = line.substring(0, colonIndex);
+        const value = line.substring(colonIndex + 1).trim();
+
+        switch (key) {
```
jamietanna: It might be worth adding a DEBUG/TRACE log here in case we find any fields that we don't support?

jamietanna (commented):
Couple of small comments - otherwise LGTM

jamietanna:
@punsoong-canva mind also resolving conflicts? Then CI should start running again

lib/modules/datasource/apk/parser.spec.ts
```diff
@@ -0,0 +1,181 @@
+import { vi } from 'vitest';
+import { logger } from '../../../logger/index.ts';
+import { parseApkIndex } from './parser.ts';
+
+describe('modules/datasource/apk/parser', () => {
+  describe('parseApkIndex', () => {
+    it('should parse valid APK index content', () => {
+      const indexContent = `P:bash
```
viceice:
```suggestion
      const indexContent = codeBlock`P:bash
```

use `codeBlock` to properly indent

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,249 @@
...
+    try {
+      logger.debug('Extracting APK index from tar.gz buffer');
+
+      const decompressedBuffer = await gunzip(buffer);
+
+      logger.debug(
+        `Decompressed buffer size: ${decompressedBuffer.length} bytes`,
+      );
+
+      // Use tar-stream to extract the APKINDEX file
```
viceice: can we use our exising `tar` dependency instead? it supports extraction out of the box.

lib/modules/datasource/apk/parser.ts
punsoong-canva: There are [17 fields ](https://wiki.alpinelinux.org/wiki/Apk_spec#APKINDEX_Format) in this spec, but only these 4 are relevant to our datasource.
Did you want me to catch and ignore the other fields, but add a trace for the `default` case?

lib/modules/datasource/apk/index.ts
punsoong-canva: changed to use tar

lib/modules/datasource/apk/parser.spec.ts
punsoong-canva: Thanks, much neater now!

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,212 @@
...
+      );
+
+      return packages;
+    } catch (err) {
+      logger.warn({ registryUrl, err }, 'Failed to fetch APK packages');
+      throw err;
+    }
+  }
+
+  getPackages(registryUrl: string): Promise<ApkPackage[]> {
```
viceice: shouldn't this private as it's not public datasource api?

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,212 @@
...
+            });
+          }
+        },
+      });
+
+      await new Promise<void>((resolve, reject) => {
+        parser.on('end', resolve);
+        parser.on('error', reject);
+        parser.end(buffer);
+      });
```
viceice: why not using async extract like we do in terraform?

https://github.com/renovatebot/renovate/blob/594a453e5f64a9b237a3989afb19896d3279f5b3/lib/modules/manager/terraform/lockfile/hash.ts#L109-L135

1. pipe the downloaded file to cache
1. use tar extract from file

- https://github.com/isaacs/node-tar?tab=readme-ov-file#tarxoptions-filelist-callback-alias-tarextract

punsoong-canva: Thanks, updated to use async extract

lib/modules/datasource/apk/index.ts
punsoong-canva: thanks updated

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  override readonly defaultVersioning = looseVersioning;
+
+  // Alpine APK repository URL structure:
+  // https://dl-cdn.alpinelinux.org/alpine/{version}/{repository}/{architecture}/
+  // - version: latest-stable, v3.19, etc.
+  // - repository: main, community, testing
+  // - architecture: x86_64, aarch64, armv7, etc.
+  override readonly defaultRegistryUrls = [
+    'https://dl-cdn.alpinelinux.org/alpine/latest-stable/main/x86_64',
```
viceice: should we use base url with parameters like we do in debian datasource? 🤔

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,205 @@
...
+   * Fetches the APK index file from the repository
+   * @param {string} registryUrl - The URL of the APK repository
+   * @returns {string} The uncompressed APKINDEX content of the APK index file
+   */
+  private async getApkIndex(registryUrl: string): Promise<string> {
+    const indexUrl = joinUrlParts(registryUrl, 'APKINDEX.tar.gz');
+
+    try {
+      logger.debug({ indexUrl }, 'Attempting to download APKINDEX.tar.gz');
+      const response = await this.http.getBuffer(indexUrl);
```
viceice: pipe to file and don't load the full buffer to memory

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,205 @@
...
+
+      await tarExtract({
+        file: tarFile,
+        cwd: cacheDir,
+        filter: (path) => path === 'APKINDEX',
+      });
+
+      if (await fs.cachePathExists(extractedFile)) {
+        logger.debug('Successfully extracted APKINDEX content');
+        return await fs.readCacheFile(extractedFile, 'utf8');
```
viceice: we should not read the full file into memory. instead we should use a line based reader.
I think we do something similar in debian datasource

lib/modules/datasource/apk/index.ts
punsoong-canva: I think I prefer it like this. I was thrown off with the deb defaultRegistryURL as its not a valid URL but is translated into one later.
I have updated the readme.md to include a similar usage example as the deb datasource does.
Happy to follow the deb datasource pattern if you think its better.

lib/modules/datasource/apk/index.ts
punsoong-canva: Switched to http.stream + fs.pipeline into cache (no getBuffer).

lib/modules/datasource/apk/index.ts
punsoong-canva: Parse APKINDEX with readline + createCacheReadStream, same idea as deb.

jamietanna:
@viceice I think this is looking good, but would appreciate your review to confirm

@punsoong-canva mind addressing the conflict? Thanks!

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,197 @@
...
+    'The source URL is determined from the `url` field in the results.';
+
+  constructor() {
+    super(ApkDatasource.id);
+  }
+
+  /**
+   * Parses an extracted APKINDEX file (line-by-line). Exposed for tests.
+   */
+  private async parseApkIndexFile(path: string): Promise<ApkPackage[]> {
```
RahulGautamSingh: this wrapper exists only so the spec can spy on it. drop it and mock the module export instead.

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,197 @@
...
+    return parseApkIndexFile(path);
+  }
+
+  /**
+   * Gets all available packages from an APK repository
+   */
+  private async _getPackages(registryUrl: string): Promise<ApkPackage[]> {
```
RahulGautamSingh: best-practices.md says no underscore prefixes on properties. rename to `getPackagesUncached` or similar, and rename the cached wrapper accordingly.

lib/modules/datasource/apk/parser.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/**
+ * Parses an APK index file content and extracts package information
+ *
+ * @param indexContent - The extracted text content from APKINDEX.tar.gz file
+ *                       (tar.gz extraction is handled upstream)
+ * @returns Array of parsed APK package objects
+ */
+export function parseApkIndex(indexContent: string): ApkPackage[] {
```
RahulGautamSingh: this duplicates the file-based parser and is only used by the spec. drop it and test `parseApkIndexFile` via `outputCacheFile`.

lib/modules/datasource/apk/parser.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+      const result = parseApkIndex(indexContent);
+
+      expect(result).toHaveLength(3);
+      expect(result[0].name).toBe('bash');
+      expect(result[1].name).toBe('another-incomplete');
+      expect(result[2].name).toBe('nginx');
+    });
+
+    it('should handle parsing errors gracefully', () => {
+      const indexContent = null as any;
```
RahulGautamSingh: avoid `as any`. if `parseApkIndex` goes away (see other comment) this case goes with it.

lib/modules/datasource/apk/readme.md
```diff
@@ -0,0 +1,65 @@
...
+
+`depName` must match the package name in `APKINDEX` (the `P:` field), e.g. `nginx` for the `nginx` package. You can omit `release=` in the comment when the template default (`v3.19` above) matches your Alpine line.
+
+### Multiple Dockerfiles or Alpine versions
+
+The datasource still only receives one directory `registryUrl` per lookup. Besides the optional `release=` pattern in the usage example, you can:
+
+1. **Several custom managers** with different `managerFilePatterns` / `matchFilePatterns` and a fixed `registryUrlTemplate` each (e.g. one for `docker/alpine-3.18/**`, another for `docker/alpine-3.19/**`).
+
+2. **`packageRules`** with `matchFileNames` / `matchPaths` and `registryUrls` to override the directory URL for specific paths or packages.
```
RahulGautamSingh: style-guide.md: use `1.` for every item in a numbered list.

lib/modules/datasource/apk/index.spec.ts
```diff
@@ -0,0 +1,525 @@
...
+      const tarGzBuffer = await createTarGz([
+        { name: 'APKINDEX', content: 'P:bash\nV:5.0.0\n' },
+      ]);
+
+      httpMock
+        .scope('https://packages.wolfi.dev')
+        .get('/os/x86_64/APKINDEX.tar.gz')
+        .reply(200, tarGzBuffer);
+
+      const parseSpy = vi.spyOn(apkDatasource as any, 'parseApkIndexFile');
```
RahulGautamSingh: the `as any` is here only because `parseApkIndexFile` is a private wrapper. dropping that wrapper lets you spy on the imported module function and removes the cast.

lib/modules/datasource/apk/parser.ts
```diff
@@ -0,0 +1,133 @@
...
+    default:
+      break;
+  }
+}
+
+function flushApkPackage(packageInfo: Partial<ApkPackage>): ApkPackage | null {
+  if (packageInfo.name && packageInfo.version) {
+    return packageInfo as ApkPackage;
+  }
+  if (Object.keys(packageInfo).length) {
```
RahulGautamSingh: prefer `is.nonEmptyObject(packageInfo)` from `@sindresorhus/is`.

lib/modules/datasource/apk/index.ts
punsoong-canva: Done — dropped the private wrapper. The class now calls `parseApkIndexFile` from the module directly.

lib/modules/datasource/apk/index.ts
punsoong-canva: Done — renamed to `getPackagesUncached`, cached wrapper stays as `getPackages`.

lib/modules/datasource/apk/parser.ts
punsoong-canva: Done — removed the string-based `parseApkIndex`. All tests now go through `parseApkIndexFile` via `outputCacheFile`.

lib/modules/datasource/apk/parser.spec.ts
punsoong-canva: Done — removed along with `parseApkIndex`.

lib/modules/datasource/apk/readme.md
punsoong-canva: Done — both items now use `1.`.

lib/modules/datasource/apk/index.spec.ts
punsoong-canva: Done — now spying on the module export (`vi.spyOn(parser, "parseApkIndexFile")`) instead. No more `as any`.

lib/modules/datasource/apk/parser.ts
punsoong-canva: Done — updated to `is.nonEmptyObject(packageInfo)`.

jamietanna (dismissed):
Would like a re-review from the others on this PR before we merge, but this is looking good from my POV!

lib/modules/datasource/apk/index.ts
viceice: yes, please follow dep datasource pattern

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,192 @@
...
+      if (err instanceof HttpError) {
+        const statusCode = err.response?.statusCode;
+        if (statusCode === 429 || (statusCode && statusCode >= 500)) {
+          throw new ExternalHostError(err);
+        }
+        logger.warn({ registryUrl, err }, 'Failed to fetch APK packages');
+        throw err;
+      }
+
+      /* v8 ignore next 2 -- hard to test tar parser errors */
```
viceice: nope, just spy on extract or throw a custom error from nock

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    const indexUrl = joinUrlParts(registryUrl, 'APKINDEX.tar.gz');
+    const extractId = randomUUID();
+    const cacheDir = await fs.ensureCacheDir(upath.join('apk', extractId));
+    const tarFile = upath.join(cacheDir, 'APKINDEX.tar.gz');
+    const extractedFile = upath.join(cacheDir, 'APKINDEX');
+
+    try {
+      logger.debug({ indexUrl }, 'Attempting to download APKINDEX.tar.gz');
+      const readStream = this.http.stream(indexUrl);
```
viceice, line 80: we should cache last modified header, so we don't need to fetch if not modified

jamietanna: Would passing a `cacheProvider` here be the best way of doing that automagically?

lib/modules/datasource/apk/index.ts
jamietanna: Debian's datasource uses i.e. https://github.com/punsoong-canva/renovate/blob/f2a254ebe8cbdc9ca6d4eaa55f397685184e80a7/lib/modules/datasource/deb/index.ts#L50-L52

Where we're not parameterising it - I can't see another Datasource where we're doing this?

lib/modules/datasource/apk/index.ts
viceice: I don't think it works with streaming. @zharinov do you know?

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /**
+   * Gets releases for a specific package from APK repositories
+   */
+  async getReleases({
+    packageName,
+    registryUrl,
+  }: GetReleasesConfig): Promise<ReleaseResult | null> {
+    /* v8 ignore next 3 -- should never happen */
```
viceice:
```suggestion
    /* v8 ignore if -- should never happen */
```

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  private getPackages(registryUrl: string): Promise<ApkPackage[]> {
+    return withCache(
+      {
+        namespace: `datasource-${ApkDatasource.id}`,
+        key: registryUrl,
+        ttlMinutes: 60,
+        fallback: true,
+      },
+      () => this.getPackagesUncached(registryUrl),
```
viceice:
```suggestion
      () => this.,_getPackages(registryUrl),
```
we usually use underscore method name for such cases

lib/modules/datasource/apk/index.ts
viceice: can maybe done in a follow up PR

lib/modules/datasource/apk/index.ts
viceice: I prefer Debian variant because there can be multiple components. I think alpine `main` can also have multiple other names and users like to configure one url 🤔

jamietanna: So to confirm, what's the change we need to make here?

punsoong-canva: I've pushed https://github.com/renovatebot/renovate/pull/40250/commits/bf74f3fe704ec2fee1495d950ccd86b0f227cf83 to follow deb pattern

jamietanna:
A few notes via Claude Sonnet 5:

````markdown
## Correctness bugs

**1. Extra query params on `registryUrl` corrupt the built component URL** — `lib/modules/datasource/apk/url.ts:19-42`
`constructComponentUrls` only deletes `arch`/`branch`/`components` from the URL before calling `joinUrlParts`. Any other query param (e.g. an auth token on a private mirror: `...&token=SECRET`) survives into `url.toString()`, and `url-join` appends the path segments *after* the `?`, producing garbage like:
```
https://internal-mirror.example.com/alpine?token=SECRET/v3.19/main/x86_64
```
I reproduced this directly with `url-join`. Any registry URL carrying a param beyond the three known ones silently breaks. **CONFIRMED**

**2. `getReleases` doesn't de-dup releases across components** — `lib/modules/datasource/apk/index.ts:196`
Unlike the core datasource dispatcher's `mergeRegistries` (which dedupes by version), this datasource concatenates releases from every component URL of a single registry with no de-dup. Overlapping/duplicated `components=` entries (or a package appearing in more than one component's APKINDEX) produce duplicate release entries in the result. **CONFIRMED**

**3. `getComponents` silently drops repeated `components` query params** — `lib/modules/datasource/apk/url.ts:74-79`
Uses `url.searchParams.get('components')`, which only returns the first match. If `components` is ever specified twice in a registry URL (e.g. via config merging), the second value is silently discarded — no error, no log, packages in the dropped component are just never found. **PLAUSIBLE**

## Cleanup / style

**4. Truthy check on `buildDate` masks `0` and `NaN`** — `lib/modules/datasource/apk/index.ts:179`
`pkg.buildDate ? asTimestamp(...) : undefined` treats a legitimate `buildDate: 0` the same as "absent," and silently swallows `NaN` from a malformed `t:` field in the parser instead of surfacing it. Project style guide asks for `is.*` helpers over truthy checks.

**5. Linear scan per package lookup** — `lib/modules/datasource/apk/index.ts:164`
Every `getReleases` call does `packages.filter()` over the full cached component package list (thousands of entries for main/community) instead of indexing by name once at parse/cache time.

**6. Sequential per-component fetches** — `lib/modules/datasource/apk/index.ts:159`
Components for a single registry URL are fetched one at a time via `await` in a loop rather than concurrently, multiplying latency by component count on a cold cache.

**7. `forEach` instead of `for...of`** — `lib/modules/datasource/apk/url.ts:33-35`
Direct violation of `docs/development/best-practices.md`, which explicitly says to avoid `forEach`.
````

I'll re-review shortly, but I would like @viceice's review before this merges. He's OOO this week, so it may not be until next week.

Appreciate this one has taken a bit to get in, but we're super close and I'm really looking forward to it 🤞🏼

**Comment by punsoong-canva** [comment]
Thanks — fixed 1 and 7 fixed

2: no user-visible effect — the dispatcher dedupes by version on every lookup under `registryStrategy: 'merge'` (`lib/modules/datasource/index.ts#L282-L289`).
3: needs `components` written twice in one URL, so left as-is.
4: intended — a malformed `t:` gives `NaN`, so `releaseTimestamp` is `undefined`.
5, 6: for these, I think its the same as the last-modified caching discussed https://github.com/renovatebot/renovate/pull/40250#discussion_r3334154581.

**Comment by jamietanna** [comment]
Not sure if you've forgotten to push any changes? 👀

File: lib/modules/datasource/apk/parser.ts [hunk 22]
```diff
@@ -0,0 +1,88 @@
...
+      packageInfo.version = value;
+      break;
+    case 'U':
+      packageInfo.url = value;
+      break;
+    case 't':
+      packageInfo.buildDate = parseInt(value, 10);
+      break;
+    default:
+      break;
```
jamietanna, line 34: Are there likely to be many keys that we may not handle? Should we add logging for this?

lib/modules/datasource/apk/readme.md
```diff
@@ -0,0 +1,105 @@
...
+
+```
+https://dl-cdn.alpinelinux.org/alpine?branch=v3.19&components=main,community&arch=x86_64
+https://packages.wolfi.dev/os?arch=x86_64
+```
+
+The first URL points to the `v3.19` branch of the Alpine repository for the `x86_64` architecture, including the `main` and `community` components.
+Renovate looks up the package in every component you list, and aggregates the releases it finds.
+
+The second URL points to the Wolfi repository, which has no branch or component in its path.
```
jamietanna, line 39: This is good

I wonder if we should show how a URL configured for Renovate is then constructed - i.e. show the Renovate JSON for how to specify the `registryUrls`, and then how that gets constructed into a URL that the Datasource then uses?

lib/modules/datasource/apk/readme.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Usage example
```
jamietanna, line 67: I'd add an HTML comment i.e.

```markdown
<!-- TODO #43711 -->
```

I'd quite like to make sure that we're doing this _by default_ in the Dockerfile manager - means folks don't need to necessarily work as hard with the regex manager

lib/modules/datasource/apk/index.ts
jamietanna: https://github.com/renovatebot/renovate/issues/45033

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  override readonly customRegistrySupport = true;
+
+  override readonly registryStrategy = 'merge';
+
+  override readonly releaseTimestampSupport = true;
+  override readonly releaseTimestampNote =
+    'The release timestamp is determined from the `buildDate` field in the results.';
+  override readonly sourceUrlSupport = 'package';
+  override readonly sourceUrlNote =
+    'The source URL is determined from the `url` field in the results.';
```
jamietanna: Does this get set anywhere? (Spotted by Claude Opus 4.8) Looks to only be the `homepage`

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  constructor() {
+    super(ApkDatasource.id);
+  }
+
+  /**
+   * Gets all available packages from a single APK component
+   */
+  private async _getPackages(componentUrl: string): Promise<ApkPackage[]> {
```
jamietanna, line 69: Via Claude Opus 4.8:

> _getPackages returns [] on corrupt-tar / missing-APKINDEX / parse errors, and that [] is cached for 60 min via withCache — a transient bad download poisons the cache for an hour. Acceptable, but worth being aware of.

I think this is OK, but we may want to tweak it in the future

lib/modules/datasource/apk/url.ts
```diff
@@ -0,0 +1,126 @@
...
+ * // Returns ['https://dl-cdn.alpinelinux.org/alpine/v3.19/main/x86_64',
+ * //          'https://dl-cdn.alpinelinux.org/alpine/v3.19/community/x86_64']
+ * constructComponentUrls('https://dl-cdn.alpinelinux.org/alpine?branch=v3.19&components=main,community&arch=x86_64');
+ */
+export function constructComponentUrls(registryUrl: string): string[] {
+  try {
+    const url = parseUrl(registryUrl);
+    if (!url) {
+      return [];
+    }
```
jamietanna, line 31: Via Claude Opus 4.8 - are we happy returning empty in this case? I know we're documenting it as part of the contract, but it may hide misconfiguration issues

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,213 @@
...
+   */
+  override readonly defaultRegistryUrls = [
+    'https://dl-cdn.alpinelinux.org/alpine?branch=latest-stable&components=main&arch=x86_64',
+  ];
+
+  override readonly defaultConfig = defaultConfig;
+
+  override readonly customRegistrySupport = true;
+
+  override readonly registryStrategy = 'merge';
```
jamietanna, line 56: Note that this can end up being a performance hit when using multiple registries - as we've [seen in Maven Central](https://github.com/renovatebot/renovate/pull/43548). It'll depend on how folks use this and are configuring multiple registries, but one to note

lib/modules/datasource/apk/parser.ts
punsoong-canva: The spec has 13 other fields we'd ignore for every package — it would be too noisy to log these.

jamietanna:
@punsoong-canva are you still able to pick up the remaining comments, or want one of us to?

punsoong-canva:
hey @jamietanna , I pushed changes in https://github.com/renovatebot/renovate/pull/40250/changes/21be03fe9ee342800d2935fcdbfb49cb5863a51a 
I did forget to push as you mentioned.
wasn't sure if I should reply to each comment or not when I was reading the contribution guidelines.
Also happy for anyone else to pick up work on this. I've found I haven't been able to put a lot of focus on this 😞

jamietanna:
All good - there are a few other comments from me that I don't think we've yet worked through. I can come back to this next week when I get some time - if you're able to carry on with it that works, but otherwise I can try and get it over the line

lib/modules/datasource/apk/index.ts
punsoong-canva: Already fixed — `shouldCacheResult: isNonEmptyArray`.

punsoong-canva:
🫡 . I've resolved/replied to the comments I believe have been resolved. Please take a look at let me know if I've missed any

lib/modules/datasource/apk/url.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    validateParams(url);
+
+    const arch = validateSegment('arch', url.searchParams.get('arch')!);
+    const branch = getBranch(url);
+    const components = getComponents(url);
+
+    // Only the parameters describe the path, so anything else in the URL is dropped
+    const baseUrl = `${url.origin}${url.pathname}`;
```
RahulGautamSingh: `origin` excludes userinfo, so `https://user:token@repo.example.com/alpine?arch=x86_64` loses the credentials. 

deb, the pattern this was asked to follow, deletes the known params off the URL and uses `url.toString()`. Do the same.

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,210 @@
...
+            componentUrl,
+            releaseCount: releases.length,
+            releases,
+          },
+          'Found APK releases',
+        );
+
+        result ??= { releases: [], registryUrl };
+        result.homepage ??= matchingPackages.find((pkg) => pkg.url)?.url;
+        result.releases.push(...releases);
```
RahulGautamSingh, line 227: This pushes releases from every component into one `ReleaseResult`. The dedupe at datasource module level only runs across registry URLs, so a package present in both `main` and `community` at the same version yields two identical releases. Dedupe by version in the component loop and add a test.

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,210 @@
...
+      );
+
+      return packages;
+    } catch (err) {
+      if (err instanceof HttpError) {
+        const statusCode = err.response?.statusCode;
+        if (statusCode === 429 || (statusCode && statusCode >= 500)) {
+          throw new ExternalHostError(err);
+        }
+        logger.warn({ componentUrl, err }, 'Failed to fetch APK packages');
```
RahulGautamSingh: This warns for every `HttpError` that is not 429/5xx.

Use debug for 4xx, and also it is also logged twice, since `getReleases:201` logs the same failure again.

lib/modules/datasource/apk/index.spec.ts
RahulGautamSingh, line 1: Add a test shaped like a real index and compare on `upath.basename(path)`.

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    const componentUrls = constructComponentUrls(registryUrl);
+    let result: ReleaseResult | null = null;
+
+    for (const componentUrl of componentUrls) {
+      try {
+        const packages = await this.getPackages(componentUrl);
+
+        // Find packages matching the requested package name
+        const matchingPackages = packages.filter(
```
RahulGautamSingh: this filters the full package array for every dep. 

`deb` datasource caches `Record<string, PackageDescription[]>` keyed by package name. Key by name here too.

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,210 @@
...
+import { Datasource } from '../datasource.ts';
+import type { GetReleasesConfig, ReleaseResult } from '../types.ts';
+import { parseApkIndexFile } from './parser.ts';
+import type { ApkPackage } from './types.ts';
+import { constructComponentUrls } from './url.ts';
+
+export const apkDatasourceId = 'apk';
+
+const defaultConfig = {
+  commitMessageTopic: '{{{depName}}} Alpine package',
```
RahulGautamSingh: The readme and half the tests target `Wolfi` and `Chainguard` repos, which are not `Alpine`. 

Use "APK package".

lib/modules/datasource/apk/types.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * P:bash
+ * V:5.3-r1
+ * t:1752770415
+ * ```
+ *
+ * Only fields that are actually used downstream are included in this interface.
+ * Many APK index fields (arch, size, description, license, maintainer, etc.)
+ * are parsed but not stored to reduce complexity and memory usage since they
+ * are not needed for Renovate's version checking functionality.
```
RahulGautamSingh: It says the other fields "are parsed but not stored". They are not parsed, the default branch in `applyApkIndexLine` drops them.

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,210 @@
...
+      await fs.pipeline(readStream, writeStream);
+
+      await tarExtract({
+        file: tarFile,
+        cwd: cacheDir,
+        filter: (path) => path === 'APKINDEX',
+      });
+
+      if (!(await fs.cachePathExists(extractedFile))) {
+        logger.warn('APKINDEX file not found in tar archive');
```
RahulGautamSingh:
```suggestion
        logger.warn({componentUrl}, 'APKINDEX file not found in tar archive');
```

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,210 @@
...
+    logger.debug(`Fetching APK packages from ${componentUrl}`);
+
+    const indexUrl = joinUrlParts(componentUrl, 'APKINDEX.tar.gz');
+    const extractId = randomUUID();
+    const cacheDir = await fs.ensureCacheDir(upath.join('apk', extractId));
+    const tarFile = upath.join(cacheDir, 'APKINDEX.tar.gz');
+    const extractedFile = upath.join(cacheDir, 'APKINDEX');
+
+    try {
+      logger.debug({ indexUrl }, 'Attempting to download APKINDEX.tar.gz');
```
RahulGautamSingh: inline the url

lib/modules/datasource/apk/parser.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Parses an APK index file line-by-line to avoid loading large indexes into memory.
+ *
+ * @param extractedFile - Path to the extracted `APKINDEX` file (relative to Renovate cache).
+ */
+export async function parseApkIndexFile(
+  extractedFile: string,
+): Promise<ApkPackage[]> {
+  logger.debug({ extractedFile }, 'Parsing APK index file');
```
RahulGautamSingh: inline the extractedFile

lib/modules/datasource/apk/readme.md
```diff
@@ -0,0 +1,132 @@
...
+- `arch`: Architecture of the binary packages (e.g. `x86_64`, `aarch64`, `armv7`).
+- `branch`: Alpine branch, either a rolling alias like `latest-stable` or `edge`, or a fixed release like `v3.19`.
+  Omit this parameter for repositories which do not have a branch in their path.
+- `components`: Comma-separated list of repository components (e.g. `main,community,testing`).
+  Omit this parameter for repositories which serve their index directly below the repository root.
+
+!!! note
+  Renovate does not fetch the `registryUrl` as-is.
+  It combines the base URL with these parameters to build the actual index URLs, for example `<base>/<branch>/<component>/<arch>/APKINDEX.tar.gz`.
+  Only the parameters above are accepted, so that a typo fails loudly instead of being ignored.
```
RahulGautamSingh, line 27: 4  indent spaces not 2

lib/modules/datasource/apk/readme.md
```diff
@@ -0,0 +1,132 @@
...
+
+<!-- TODO #43711 -->
+
+## Usage example
+
+Say you pin Alpine packages in a `Dockerfile` and want Renovate to bump the versions.
+Combine the `apk` datasource with a [regex manager](../../manager/regex/index.md).
+
+Add a custom manager in `renovate.json`.
+The optional `branch` capture group is filled from the renovate comment and interpolated into `registryUrlTemplate` by the regex manager.
```
RahulGautamSingh:
```suggestion
The optional `branch` capture group is filled from the Renovate comment and interpolated into `registryUrlTemplate` by the regex manager.
```

RahulGautamSingh (changes requested):

lib/modules/datasource/apk/readme.md
punsoong-canva: Prettier rewrites 4 spaces back to 2, so this fails lint-prettier. The `deb` README uses 2 spaces as well

viceice (approved):

jamietanna (approved):
Thanks very much for your patience, and your great work on this @punsoong-canva!!

lib/modules/datasource/apk/index.ts
```diff
@@ -0,0 +1,241 @@
...
+    packagesByName[pkg.name].push(pkg);
+  }
+
+  return packagesByName;
+}
+
+export class ApkDatasource extends Datasource {
+  static readonly id = apkDatasourceId;
+
+  override readonly defaultVersioning = looseVersioning;
```
jamietanna, line 49: @punsoong-canva out of interest, was this intentionally not `versioning: apk`?

AyushSehrawat:
Hi! This commit has caused issue with arm (only tried on arm). I was setting up renovate on gitea instance and when i ran it i get the following error:

`runner/config.yaml` (Gitea config)
```yaml
runner:
  capacity: 1
  timeout: 1h
  insecure: false
  labels:
    - "ubuntu-latest:docker://docker.gitea.com/runner-images:ubuntu-latest"
    - "ubuntu-24.04:docker://docker.gitea.com/runner-images:ubuntu-24.04"
  tool_cache_mode: none
```

Renovate config repo on gitea

`config.js`
```js
module.exports = {
	platform: 'gitea',
	endpoint: 'http://gitea:3000/api/v1/',
	gitAuthor: 'Renovate Bot <renovate@email.com>',
	username: 'renovate-bot',
	autodiscover: true,
	onboardingConfig: {
		$schema: 'https://docs.renovatebot.com/renovate-schema.json',
		extends: ['config:recommended'],
	},
	optimizeForDisabled: true,
	persistRepoData: true,
};
```

`.gitea/workflows/renovate.yml`
```yml
name: renovate

on:
  workflow_dispatch:
  schedule:
    - cron: "0 12 * * *"
  push:
    branches: [main]

jobs:
  renovate:
    runs-on: ubuntu-latest
    container: ghcr.io/renovatebot/renovate:latest
    steps:
      - uses: actions/checkout@v4
      - run: renovate
        env:
          RENOVATE_CONFIG_FILE: ${{ gitea.workspace }}/config.js
          LOG_LEVEL: "debug"
          RENOVATE_TOKEN: ${{ secrets.RENOVATE_TOKEN }}
          GITHUB_COM_TOKEN: ${{ secrets.RENOVATE_GITHUB_TOKEN }}
```

Using this config when the action ran, i get the error

```bash
ERROR: unhandledRejection
       "err": {
         "code": "ERR_MODULE_NOT_FOUND",
         "message": "Cannot find package 'tar' imported from /usr/local/renovate/dist/modules/datasource/apk/index.js"
       }
       Error: Cannot find package 'tar' imported from /usr/local/renovate/dist/modules/datasource/apk/index.js
           at Object.getPackageJSONURL (node:internal/modules/package_json_reader:301:9)
           at packageResolve (node:internal/modules/esm/resolve:784:25)
           at moduleResolve (node:internal/modules/esm/resolve:873:18)
           at defaultResolve (node:internal/modules/esm/resolve:1006:11)
           at ModuleLoader.#cachedDefaultResolve (node:internal/modules/esm/loader:705:20)
           at ModuleLoader.#resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:725:38)
           at ModuleLoader.resolveSync (node:internal/modules/esm/loader:763:56)
           at ModuleLoader.#resolve (node:internal/modules/esm/loader:687:17)
           at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:607:35)
           at ModuleJob.syncLink (node:internal/modules/esm/module_job:276:33)
           at ModuleJob.link (node:internal/modules/esm/module_job:381:17)
           at new ModuleJob (node:internal/modules/esm/module_job:360:26)
           at ModuleLoader.#getOrCreateModuleJobAfterResolve (node:internal/modules/esm/loader:576:11)
           at afterResolve (node:internal/modules/esm/loader:611:52)
           at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:617:12)
           at ModuleJob.syncLink (node:internal/modules/esm/module_job:276:33)
           at ModuleJob.link (node:internal/modules/esm/module_job:381:17)
           at new ModuleJob (node:internal/modules/esm/module_job:360:26)
           at ModuleLoader.#getOrCreateModuleJobAfterResolve (node:internal/modules/esm/loader:576:11)
           at afterResolve (node:internal/modules/esm/loader:611:52)
```

Pinning the ghcr.io/renovatebot/renovate version from latest (44.64.0) to 44.62.2 has fixed the issue for me. I think this commit caused the issue https://github.com/renovatebot/renovate/pull/40250/changes/7bea61ab4e3deb9818709f79361cfac80c095f0e

jamietanna:
Thanks, investigating in https://github.com/renovatebot/renovate/discussions/45697

lib/modules/datasource/apk/index.ts
punsoong-canva: ah my bad. I see you've already fix this! thanks!

PR #40263, opened 2026-01-04
fix(package-rules): compile templated groupName before generating groupSlug

joelanford:
## Changes

When a `packageRule` sets a templated `groupName` (e.g., `{{packageFileDir}}`) without an explicit `groupSlug`, and a previous rule/preset has already set `config.groupSlug`, the `groupSlug` was incorrectly generated from the raw template string instead of the compiled value.

This caused all updates matching the rule to incorrectly share the same `groupSlug`, grouping them together unexpectedly.

### The Fix

Compile the template before slugifying:

```typescript
if (config.groupSlug && packageRule.groupName && !packageRule.groupSlug) {
  const compiledGroupName = compile(packageRule.groupName, config);
  toApply.groupSlug = slugify(compiledGroupName, { lower: true });
}
```

### How to Reproduce

```json
{
  "extends": ["config:best-practices"],
  "packageRules": [
    {
      "matchFileNames": ["**/*"],
      "groupName": "{{packageFileDir}}"
    }
  ]
}
```

With `config:best-practices` → `docker:pinDigests`, the `pinDigest` update type sets `groupSlug: 'pin-dependencies'`. When the user's rule applies with a templated `groupName`, the raw template gets slugified instead of the compiled value.

This is a follow-up to #25490 which fixed template compilation for `groupName` in `generateBranchName`, but missed this earlier location where `groupSlug` can be set from the raw template.

## Context

## AI assistance disclosure


lib/util/package-rules/index.ts
```diff
@@ -44,7 +44,9 @@ export async function applyPackageRules<T extends PackageRuleInputConfig>(
       const toApply = removeMatchers({ ...packageRule });
       if (config.groupSlug && packageRule.groupName && !packageRule.groupSlug) {
         // Need to apply groupSlug otherwise the existing one will take precedence
-        toApply.groupSlug = slugify(packageRule.groupName, {
+        // Compile the template first to handle templated groupNames
+        const compiledGroupName = compile(packageRule.groupName, config);
+        toApply.groupSlug = slugify(compiledGroupName, {
```
viceice, line 49: maybe we should instead remove the group slug from toApply here instead? so it's compiled later again? 🤔

PR #40282, opened 2026-01-06
fix(nix): extract correct dep when inputs are duplicated

spotdemo4:
## Changes

Currently, renovate assumes that the *name* of a root input corresponds to a node name. This is incorrect, as it is actually the root input's *value* that maps an input to a node. If there are multiple dependencies that share the same name (for example, transitive dependencies nixpkgs, nixpkgs_2, nixpkgs_3...), renovate will extract information from the node `nixpkgs` even if the root input `nixpkgs` is actually pointing to `nixpkgs_2`.

While a relatively simple issue, some refactoring was needed as the dependency name must still be the root input's name, as that's what `nix flake update` expects during `updateArtifacts()`.

## Context

## AI assistance disclosure

### DIFF
--- a/lib/modules/manager/nix/extract.spec.ts
+++ b/lib/modules/manager/nix/extract.spec.ts
@@ -1450,4 +1450,20 @@ describe('modules/manager/nix/extract', () => {
       ],
     });
   });
+
+  it('returns null when node is missing', async () => {
+    const flakeLock = codeBlock`{
+      "nodes": {
+        "root": {
+          "inputs": {
+            "unknown-flake": "unknown-flake"
+          }
+        }
+      },
+      "root": "root",
+      "version": 7
+    }`;
+    fs.readLocalFile.mockResolvedValueOnce(flakeLock);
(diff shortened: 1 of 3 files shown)


marwing:
IIUC, the root node's name is still hardcoded to the name `root` instead of using the [`root` attribute](https://github.com/NixOS/nix/blob/91c706852bd643f659fda3b0b7cf65f8b5b3fdee/src/nix/flake.md?plain=1#L691-L692). Would you be willing to fix this here as well (it kind of fits)? Otherwise, I might open a separate PR once I have some spare time (life's busy).

spotdemo4:
> IIUC, the root node's name is still hardcoded to the name `root` instead of using the [`root` attribute](https://github.com/NixOS/nix/blob/91c706852bd643f659fda3b0b7cf65f8b5b3fdee/src/nix/flake.md?plain=1#L691-L692). Would you be willing to fix this here as well (it kind of fits)? Otherwise, I might open a separate PR once I have some spare time (life's busy).

fixed :+1:

Issue #40284, opened 2026-01-06
Composer update fails with `Command failed: git stash pop || true` since `42.68.5`


jamietanna:
### Describe the proposed change(s).

Per #40261 - we're seeing that all composer updates fail due to https://github.com/renovatebot/renovate/blob/0dfe0f94e45c28df45bada7eac2ef777d2480c41/lib/modules/manager/composer/artifacts.ts#L173 no longer having a shell

BacLuc:
We also see this:
- https://github.com/ecamp/ecamp3/pull/8833
- https://github.com/ecamp/ecamp3/pull/8827

PR #40228, opened 2025-12-31
fix(util/exec): don't use `shell` by default

jamietanna:
## Changes

This resolves https://github.com/renovatebot/renovate/security/advisories/GHSA-pfq2-hh62-7m96 by hardening execution of external commands to no longer run inside a shell.

## Context

Please select one of the below:

- [ ] This closes an existing Issue: Closes #
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

The public repository: <URL>

PR #40290, opened 2026-01-06
fix(util/exec): allow ignoring failures per-command

jamietanna:
## Changes

As part of future changes which will fix issues with Composer + Yarn 1
introduced by https://github.com/renovatebot/renovate/pull/40228, we need to find
a way to ignore command failure, on a per-command basis.

To do this, we can provide a refactor to introduce a new type,
`CommandWithOptions` which has the ability to specify whether to
`ignoreFailure`.

To make this less of a refactor than would otherwise require, we can
pass around a `(string | CommandWithOptions)[]`
(or a `string[]` or a `CommandWithOptions[]`) to be used, and use
helpers `asRawCommand` and `asRawCommands` to extract the underlying
command where necessary.

In the case a command's failure is being ignored, return the error code
in the `ExecResult`.

We also need to make sure that when executing in Docker, we still
provide the `|| true`s, otherwise we'll start to see commands failing
unexpectedly.

This can be seen in use in:

- #40291 
- https://github.com/renovatebot/renovate/pull/40292

## Context

Prerequisite for #40284 

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/util/exec/common.ts
+++ b/lib/util/exec/common.ts
@@ -2,14 +2,20 @@ import type { ChildProcess } from 'node:child_process';
 import type { Readable } from 'node:stream';
 import { isNullOrUndefined } from '@sindresorhus/is';
 import { execa } from 'execa';
-import { split } from 'shlex';
+import { join, split } from 'shlex';
 import { instrument } from '../../instrumentation';
+import { logger } from '../../logger';
 import { getEnv } from '../env';
 import { sanitize } from '../sanitize';
 import type { ExecErrorData } from './exec-error';
 import { ExecError } from './exec-error';
-import type { DataListener, ExecResult, RawExecOptions } from './types';
-
+import type {
+  CommandWithOptions,
+  DataListener,
+  ExecResult,
+  RawExecOptions,
+} from './types';
+import { asRawCommand, isCommandWithOptions } from './utils';
 // https://man7.org/linux/man-pages/man7/signal.7.html#NAME
 // Non TERM/CORE signals
 // The following is step 3. in https://github.com/renovatebot/renovate/issues/16197#issuecomment-1171423890
@@ -81,11 +87,20 @@ function registerDataListeners(
 }
 
 export function exec(
-  theCmd: string,
+  commandArgument: string | CommandWithOptions,
   opts: RawExecOptions,
 ): Promise<ExecResult> {
+  let theCmd = commandArgument;
+  let ignoreFailure = false;
+  if (isCommandWithOptions(commandArgument)) {
+    theCmd = join(commandArgument.command);
+    if (commandArgument.ignoreFailure !== undefined) {
+      ignoreFailure = commandArgument.ignoreFailure;
+    }
+  }
+
   return new Promise((resolve, reject) => {
-    let cmd = theCmd;
+    let cmd = asRawCommand(theCmd);
     let args: string[] = [];
     const maxBuffer = opts.maxBuffer ?? 10 * 1024 * 1024; // Set default max buffer size to 10MB
 
@@ -139,15 +154,34 @@ export function exec(
         return;
       }
       if (code !== 0) {
-        reject(
-          new ExecError(
-            `Command failed: ${cp.spawnargs.join(' ')}\n${stringify(stderr)}`,
-            {
-              ...rejectInfo(),
-              exitCode: code,
-            },
-          ),
+        if (ignoreFailure === undefined || ignoreFailure === false) {
+          reject(
+            new ExecError(
+              `Command failed: ${cp.spawnargs.join(' ')}\n${stringify(stderr)}`,
+              {
+                ...rejectInfo(),
+                exitCode: code,
+              },
+            ),
+          );
+          return;
+        }
+
+        logger.once.debug(
+          {
+            command: cp.spawnargs.join(' '),
+            stdout: stringify(stdout),
+            stderr: stringify(stderr),
+            exitCode: code,
+          },
+          `Ignoring failure to execute comamnd \`${cp.spawnargs.join(' ')}\`, as ignoreFailure=true is set`,
         );
+
+        resolve({
+          stderr: stringify(stderr),
+          stdout: stringify(stdout),
+          exitCode: code,
+        });
         return;
       }
       resolve({
@@ -193,7 +227,10 @@ function kill(cp: ChildProcess, signal: NodeJS.Signals): boolean {
 }
 
 export const rawExec: (
-  cmd: string,
+  cmd: string | CommandWithOptions,
   opts: RawExecOptions,
-) => Promise<ExecResult> = (cmd: string, opts: RawExecOptions) =>
-  instrument(`rawExec: ${sanitize(cmd)}`, () => exec(cmd, opts));
+) => Promise<ExecResult> = (
+  cmd: string | CommandWithOptions,
+  opts: RawExecOptions,
+) =>
+  instrument(`rawExec: ${sanitize(asRawCommand(cmd))}`, () => exec(cmd, opts));
--- a/lib/util/exec/docker/index.ts
+++ b/lib/util/exec/docker/index.ts
@@ -1,4 +1,5 @@
 import { isNonEmptyString, isString } from '@sindresorhus/is';
+import { join } from 'shlex';
 import { GlobalConfig } from '../../../config/global';
 import { SYSTEM_INSUFFICIENT_MEMORY } from '../../../constants/error-messages';
 import { logger } from '../../../logger';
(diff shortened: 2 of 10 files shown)


jamietanna (commented):
Given the failure on https://github.com/renovatebot/renovate/pull/40292 (for Docker) I wonder if we should inject the `|| true` in those cases?

lib/util/exec/types.ts
```diff
@@ -56,3 +64,42 @@ export interface ExecOptions {
...
+ * configuration that can be configured on a per-command basis, that doesn't make sense to be on the `RawExecOptions`
+ */
+export interface CommandWithOptions {
+  command: string;
+
+  /** do not throw errors when a command fails, but do log that an error occurred */
+  ignoreFailure?: boolean;
+}
+
+export function isCommandWithOptions(cmd: any): cmd is CommandWithOptions {
```
viceice: a `types.ts` file should only declare types not functions. please move funcitons to a common or util file

prefer `unknown` over `any`
```suggestion
export function isCommandWithOptions(cmd: unknown): cmd is CommandWithOptions {
```

lib/util/exec/containerbase.ts
```diff
@@ -16,7 +16,7 @@ import { id as rubyVersioningId } from '../../modules/versioning/ruby';
 import { id as semverVersioningId } from '../../modules/versioning/semver';
 import { id as semverCoercedVersioningId } from '../../modules/versioning/semver-coerced';
 import { getEnv } from '../env';
-import type { Opt, ToolConfig, ToolConstraint } from './types';
+import { type Opt, type ToolConfig, type ToolConstraint } from './types';
```
viceice: unrelated change?

lib/util/exec/docker/index.ts
```diff
@@ -267,9 +267,33 @@ export async function generateDockerCommand(
...
 
-  const bashCommand = [...prepareCommands(preCommands), ...commands].join(
-    ' && ',
-  );
+  const bashCommandParts = [];
+
+  for (const preCommand of preCommands) {
+    if (isCommandWithOptions(preCommand) && isString(preCommand.command)) {
+      if (preCommand.ignoreFailure) {
+        bashCommandParts.push(`${preCommand.command} || true`);
```
viceice: do we need braces because of `&&` / `||` precedence?
```suggestion
        bashCommandParts.push(`(${preCommand.command} || true)`);
```

lib/util/exec/docker/index.ts
```diff
@@ -267,9 +267,33 @@ export async function generateDockerCommand(
...
+      }
+    } else if (isString(preCommand)) {
+      bashCommandParts.push(preCommand);
+    }
+  }
+
+  for (const command of commands) {
+    if (isCommandWithOptions(command)) {
+      if (command.ignoreFailure) {
+        bashCommandParts.push(`${command.command} || true`);
```
viceice: same here

lib/util/exec/docker/index.ts
jamietanna: We likely do - I noticed this while looking at it, too

But this is existing code, so I'd prefer to follow-up with it. I assume it must be working for some people, as I've not seen reports of issues?

viceice: I think docker mode isn't widely used anymore

lib/util/exec/index.spec.ts
```diff
@@ -973,6 +980,95 @@ describe('util/exec/index', () => {
...
+          stdout,
+          stderr,
+          exitCode: 10,
+        }),
+    );
+    GlobalConfig.set({ ...globalConfig });
+    const promise = exec([
+      {
+        command: 'foobar',
+        ignoreFailure: false,
```
viceice: doesn't match test description

lib/util/exec/index.spec.ts
```diff
@@ -5,7 +5,14 @@ import { TEMPORARY_ERROR } from '../../constants/error-messages';
 import { setCustomEnv } from '../env';
 import * as dockerModule from './docker';
 import { getHermitEnvs } from './hermit';
-import type { ExecOptions, RawExecOptions, VolumeOption } from './types';
+import {
+  type CommandWithOptions,
```
viceice: please move type before brace again

lib/util/exec/index.ts
```diff
@@ -12,15 +12,16 @@ import {
...
 import { getHermitEnvs, isHermit } from './hermit';
-import type {
-  DockerOptions,
-  ExecOptions,
-  ExecResult,
-  ExtraEnv,
-  Opt,
-  RawExecOptions,
+import {
+  type CommandWithOptions,
```
viceice: please move type before brace again for less changes

lib/util/exec/types.ts
```diff
@@ -56,3 +63,13 @@ export interface ExecOptions {
   timeout?: number | undefined;
   shell?: boolean | string | undefined;
 }
+
+/**
+ * configuration that can be configured on a per-command basis, that doesn't make sense to be on the `RawExecOptions`
+ */
+export interface CommandWithOptions {
+  command: string;
```
viceice: should we take the opportunity to forge passing args as separate array, so we don't need to parse command?

lib/util/exec/index.spec.ts
```diff
@@ -1064,4 +1160,33 @@ describe('util/exec/index', () => {
     await expect(promise).rejects.toThrow(TEMPORARY_ERROR);
     expect(removeDockerContainerSpy).toHaveBeenCalledTimes(0);
   });
+
+  describe('asRawCommand', () => {
```
viceice: duplicate tests? they're already in utils.spec.ts

lib/util/exec/index.spec.ts
jamietanna: Sorry, not sure why this keeps happening 😓

lib/util/exec/types.ts
jamietanna: Taking a look now - I'd say maybe as a follow-up PR, but I'll see how much it changes things

lib/util/exec/common.ts
```diff
@@ -81,11 +88,20 @@ function registerDataListeners(
...
 export function exec(
-  theCmd: string,
+  commandArgument: string | CommandWithOptions,
   opts: RawExecOptions,
 ): Promise<ExecResult> {
+  let theCmd = commandArgument;
+  let ignoreFailure = false;
+  if (isCommandWithOptions(commandArgument)) {
+    theCmd = shlex.join(commandArgument.command);
+    if (commandArgument.ignoreFailure !== undefined) {
```
jamietanna, line 97: Can't seem to trigger the coverage for this 🤔

lib/util/exec/common.ts
```diff
@@ -3,13 +3,20 @@ import type { Readable } from 'node:stream';
 import { isNullOrUndefined } from '@sindresorhus/is';
 import { execa } from 'execa';
 import { split } from 'shlex';
```
viceice:
```suggestion
import { join, split } from 'shlex';
```

lib/util/exec/common.ts
```diff
@@ -3,13 +3,20 @@ import type { Readable } from 'node:stream';
...
+import * as shlex from 'shlex';
```
viceice:
```suggestion
```

lib/util/exec/common.ts
```diff
@@ -81,11 +88,20 @@ function registerDataListeners(
...
 
 export function exec(
-  theCmd: string,
+  commandArgument: string | CommandWithOptions,
   opts: RawExecOptions,
 ): Promise<ExecResult> {
+  let theCmd = commandArgument;
+  let ignoreFailure = false;
+  if (isCommandWithOptions(commandArgument)) {
+    theCmd = shlex.join(commandArgument.command);
```
viceice:
```suggestion
    theCmd = join(commandArgument.command);
```

lib/util/exec/utils.ts
```diff
@@ -35,3 +36,44 @@ export function getChildEnv({
...
+
+  if (!Array.isArray(cmd.command)) {
+    return false;
+  }
+
+  if (!cmd.command.length) {
+    return false;
+  }
+
+  if (cmd.command.some((v) => typeof v !== 'string')) {
```
viceice:
```suggestion
  if (cmd.command.some((v) => !isString(v))) {
```

viceice (commented):
just some small things

viceice (approved):
we should cleanup in a follow up, so we split / prepare commands and args before raw exec, so we don't need to join and split the command again and again

lib/util/exec/index.ts
```diff
@@ -108,7 +113,7 @@ async function prepareRawExec(
     const cwd = getCwd(opts);
     const dockerOptions: DockerOptions = { ...docker, cwd, envVars };
     const dockerCommand = await generateDockerCommand(
-      rawCommands,
+      asRawCommands(rawCommands),
```
jamietanna, line 116: Regression: https://github.com/renovatebot/renovate/pull/40517

PR #40312, opened 2026-01-07
fix(jsonata): ensure all evaluations use a new environment

markspolakovs:
## Changes

When evaluating a JSONata expression concurrently, specifically when reusing an expression instance (as we do since we cache them), it's possible that a second concurrent evaluate() call clobbers the `$` binding before the first one has finished, which will make it use incorrect values.

Passing in a non-undefined bindings value causes JSONata to instead create a new environment for each evaluate() call, preventing these conflicts.

## Context

- [x] This closes an existing Issue: #40311 (still a discussion, not moved to an issue yet)

## AI assistance disclosure

The public repository: https://github.com/markspolakovs/renovate-concurrency-repro (as well as an internal one where we initially found this issue)

### DIFF
--- a/lib/util/jsonata.ts
+++ b/lib/util/jsonata.ts
@@ -11,7 +11,18 @@ export function getExpression(input: string): jsonata.Expression | Error {
   }
   let result: jsonata.Expression | Error;
   try {
-    result = jsonata(input);
+    const expression = jsonata(input);
+    // Wrap the evaluate method to default bindings to {} for concurrent evaluation safety
+    // This prevents race conditions when the same cached expression is evaluated
+    // concurrently with different data
+    const originalEvaluate = expression.evaluate.bind(expression);
+    expression.evaluate = (
+      data: unknown,
+      bindings: Record<string, unknown> = {},
+    ) => {
+      return originalEvaluate(data, bindings);
+    };
+    result = expression;
   } catch (err) {
     // JSONata errors aren't detected as TypeOf Error
     result = new Error(err.message);
--- a/lib/util/jsonata.ts
+++ b/lib/util/jsonata.ts
@@ -2,7 +2,11 @@ import jsonata from 'jsonata';
 import * as memCache from './cache/memory';
 import { toSha256 } from './hash';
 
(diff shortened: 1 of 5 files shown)


jamietanna:
I wonder what we can do tests-wise to surface this / stop us from introducing a regression in the future 🤔

secustor (changes requested):
We should rather fix this centrally, so people can not forget it. 

See https://github.com/renovatebot/renovate/compare/main...secustor:renovate:fix/jsonata-centrally

Feel free to cherry pick

markspolakovs:
That's a good shout, thank you @secustor (and for writing the test too!). Picked into my branch (along with a tweak to make the return type more specific, as the replacement `evaluate` function doesn't support the `callback` arg), and undone my change as it's no longer necessary.

secustor (approved):

markspolakovs:
I think I broke the release bot with my commit message 🙊 https://github.com/renovatebot/renovate/actions/runs/20796244373/job/59731066362

`[9:12:31 PM] [semantic-release] › ✘  An error occurred while running semantic-release: Error: Could not resolve to an issue or pull request with the number of 40311.`

viceice:
@secustor please take care it the closing issue is a discussion, then remove the `closes` otherwise the release will fail on posting success messages.

jamietanna:
(FYI this was also down to the fact that the commit message had it in - https://github.com/renovatebot/renovate/issues/39397 would have picked it up)

lib/util/jsonata.ts
```diff
@@ -11,7 +15,18 @@ export function getExpression(input: string): jsonata.Expression | Error {
   }
   let result: jsonata.Expression | Error;
   try {
-    result = jsonata(input);
+    const expression = jsonata(input);
+    // Wrap the evaluate method to default bindings to {} for concurrent evaluation safety.
+    // This prevents race conditions when the same cached expression is evaluated
+    // concurrently with different data. See #40311 for background.
+    const originalEvaluate = expression.evaluate.bind(expression);
```
viceice, line 22: that's a very crude workaround. we need to find a better cleaner solution

PR #40318, opened 2026-01-07
refactor(manager/gradle): rewrite reorderFiles() for performance

Churro:
## Changes

* `version-catalogs.ts`
  * Remove `.handler((ctx) => storeInTokenMap(ctx, 'methodName'))` as it is not needed/used anywhere
  * Change `pluginName` matching to `qValueMatcher`, since this value can be interpolated. Also aligns with `plugins.ts`
  * Extract `qAlias` 

* `utils.ts`:
  * Improve time complexity of `reorderFiles()` from O(n² log n) to O(n log n) by pre-computing paths and doing file ranking only once. In practice, this means a ~100-200x speedup for ordering 1000 files. The underlying logic is unchanged, only the execution order was optimized.

## Context

This PR is a preparation to improve version catalog handling around cases like https://github.com/renovatebot/renovate/discussions/40147.
It should help to keep follow-up changesets smaller.

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/manager/gradle/utils.ts
+++ b/lib/modules/manager/gradle/utils.ts
@@ -127,36 +127,32 @@ function getFileRank(filename: string): number {
 }
 
 export function reorderFiles(packageFiles: string[]): string[] {
-  return packageFiles.sort((x, y) => {
-    const xAbs = toAbsolutePath(x);
-    const yAbs = toAbsolutePath(y);
-
-    const xDir = upath.dirname(xAbs);
-    const yDir = upath.dirname(yAbs);
-
-    if (xDir === yDir) {
-      const xRank = getFileRank(xAbs);
-      const yRank = getFileRank(yAbs);
-      if (xRank === yRank) {
-        if (xAbs > yAbs) {
+  return packageFiles
+    .map((path) => {
+      const absPath = toAbsolutePath(path);
+      return {
+        path,
+        absPath,
+        dir: upath.dirname(absPath),
+        rank: getFileRank(absPath),
+      };
+    })
+    .sort((a, b) => {
+      // Different directories: check parent-child relationship
+      if (a.dir !== b.dir) {
+        if (a.dir.startsWith(b.dir + '/')) {
           return 1;
         }
-        if (xAbs < yAbs) {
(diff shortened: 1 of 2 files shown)


lib/modules/manager/gradle/utils.ts
```diff
@@ -127,36 +127,32 @@ function getFileRank(filename: string): number {
...
+      };
+    })
+    .sort((a, b) => {
+      // Different directories: check parent-child relationship
+      if (a.dir !== b.dir) {
+        if (a.dir.startsWith(b.dir + '/')) {
           return 1;
         }
-        if (xAbs < yAbs) {
+        if (b.dir.startsWith(a.dir + '/')) {
```
astellingwerf, line 146: Does this work OK on filesystems with the inverse slashes?

lib/modules/manager/gradle/utils.ts
```diff
@@ -127,36 +127,32 @@ function getFileRank(filename: string): number {
...
-      if (xRank === yRank) {
-        if (xAbs > yAbs) {
+  return packageFiles
+    .map((path) => {
+      const absPath = toAbsolutePath(path);
+      return {
+        path,
+        absPath,
+        dir: upath.dirname(absPath),
+        rank: getFileRank(absPath),
```
astellingwerf, line 137: I don't fully understand the usecase, but it seems like there are plenty of cases where the rank is not involved in the sorting. Would it make sense to delay calculation of the rank until we need it?

lib/modules/manager/gradle/utils.ts
Churro: Yes, converting these to forward slashes is a core part of `upath`, see https://www.npmjs.com/package/upath

lib/modules/manager/gradle/utils.ts
Churro: `rank` is always needed if there are at least two files on the same directory hierarchy level. As you usually have multiple gradle files in one folder, I think this could be considered the standard case. Delaying it to the `sort()` operation would mean evaluating the `rank` always at least twice (`a`, `b`).

viceice (approved):

PR #40366, opened 2026-01-10
refactor: replace chalk with node util for styling in logger

Dueen:
## Changes

replace chalk with node builtin util for styling in logger

## Context

## AI assistance disclosure

## Documentation

## How I've tested my work

- [x] Code inspection only

## Why?
`chalk` was only used for `lib/logger/pretty-stdout`, so keeping it pulls an extra transitive dependency (plus its transitive dependencies) into the runtime just for ANSI styling; moving to the node built-in utility shrinks the dependency tree, speeds up installs, and removes a potential supply-chain surface area.

## Availability / Compatibility
[`util.styleText`](https://nodejs.org/docs/latest/api/util.html#utilstyletextformat-text-options) was added in v21.7.0, v20.12.0 which means the supported engines (`^22.13` and `^24.11`) all satisfy the requirement, so there should be no compatibility issues.

### DIFF
--- a/lib/logger/pretty-stdout.spec.ts
+++ b/lib/logger/pretty-stdout.spec.ts
@@ -1,13 +1,6 @@
-import chalk from 'chalk';
+import * as util from 'node:util';
 import * as prettyStdout from './pretty-stdout';
 import type { BunyanRecord } from './types';
-
-vi.mock('chalk', () => ({
-  default: ['bgRed', 'blue', 'gray', 'green', 'magenta', 'red'].reduce(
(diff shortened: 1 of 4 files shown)


viceice (commented):
please expand PR description why this is needed and since when this built-in is available? validate it against our supported node versions.

Dueen:
> please expand PR description why this is needed and since when this built-in is available? validate it against our supported node versions.

done

RahulGautamSingh (approved):
LGTM

viceice (approved):
chalk is still a transitive dependency

PR #39314, opened 2025-11-14
feat(presets): log the resolved configuration (without internal presets)

jamietanna:
## Changes

As noted in #15827, it would be useful to log the "shallow"
configuration for a given repository - that is, the resolved
configuration, but without any Renovate internal presets being included.

This will be very useful for debugging, as it shows the final resulting
configuration, but without the - often very - verbose internal presets.

In the case that only this log line is shared, we can make sure to
reference the version of the Renovate CLI in the log line.

This requires building on top of changes in
6a4fa3f245dd8976fc8477c4d72f7aae32adab34 to be able to return a `merged`
and `unmerged` list of presets, and uses helpers from
c8f42e86b4f182a6c4c530250b2b1d2a0fb20e22, and
66b84dd12bd9a638b914883fa2781e2def601e9a to process the internal presets
accordingly.

Since we first worked on this, we've worked on improvements to caching
which should provide a require no additional performance overhead
performing this resolution twice.

This can also lead to being able to understand more clearly how Renovate
configuration is being used[0], as we have the final resolved configuration.

[0]: https://www.jvt.me/posts/2024/04/14/renovate-config-sql/

Other notes:

- I've not validated what happens with environment config
- This will be super useful for "Request Help" questions
## Context

Please select one of the below:

- [x] This closes an existing Issue: Closes #15827
- [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

The public repository: <URL>

PR #40419, opened 2026-01-14
chore(presets): add a helper for internal preset names

jamietanna:
## Changes

As part of https://github.com/renovatebot/renovate/pull/39314, it would be
useful to have a consistent way of looking up the internal-only presets, by
name.

We can wrap this into a once-computed Set of names.

We need to take care to allow `default` presets to be referenced with
and without the `default` prefix.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/config/presets/internal/index.ts
+++ b/lib/config/presets/internal/index.ts
@@ -43,3 +43,23 @@ export function getPreset({
 }: PresetConfig): Preset | undefined {
   return groups[repo] && presetName ? groups[repo][presetName] : undefined;
 }
+
+function computeInternalPresets(): string[] {
+  const internalPresets: string[] = [];
+
+  for (const k in groups) {
+    const v = groups[k];
+    for (const kk in v) {
+      if (k === 'default') {
+        internalPresets.push(`:${kk}`);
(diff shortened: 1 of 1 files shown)


viceice (approved):

lib/config/presets/internal/index.ts
```diff
@@ -43,3 +43,23 @@ export function getPreset({
 }: PresetConfig): Preset | undefined {
   return groups[repo] && presetName ? groups[repo][presetName] : undefined;
 }
+
+function computeInternalPresets(): string[] {
+  const internalPresets: string[] = [];
+
+  for (const k in groups) {
```
RahulGautamSingh, line 50: not blocker: but in best-practices we have asked to use `for..of`

https://github.com/renovatebot/renovate/blob/main/docs/development/best-practices.md#iterating-objects--containers

jamietanna: Fair shout, but as this is a `Record`, it doesn't work:

```
Diagnostics:
1. typescript: Type 'Record<string, Preset>' must have a '[Symbol.iterator]()' method that returns an iterator. [2488]
```

PR #40449, opened 2026-01-15
refactor(config): add an `InheritConfig` class

### LINKED ISSUE #40447: create `InheritConfig` class

RahulGautamSingh:
## Changes

- Implement new `InheritConfig` class
- Add new `getInheritOrGlobal` fn which return inherited values and fallbacks to global/admin values
- Add tests
- As part of inital PR I am moving the harmless `configFileNames` option to the `InheritConfig.OPTIONS`.

## Context

- [x] This closes an existing Issue, Closes: #40447

## AI assistance disclosure

The public repository: <URL>


jamietanna (commented):
Think this LGTM, but will wait for @viceice's approval :)

lib/config/types.ts
```diff
@@ -204,7 +211,7 @@ export interface GlobalOnlyConfigLegacy {
  *
  * Should only contain config options where globalOnly=true.
  */
-export interface RepoGlobalConfig {
+export interface RepoGlobalConfig extends RepoInheritConfig {
```
jamietanna: I assume this is because config could be either inherit or global-only?

(thinking out loud: I wonder if we have alternatives)

jamietanna: if we are doing this, can we update the comment, as it's no longer global-only

RahulGautamSingh: The comment doesn't say "only globalOnly=true", so we could  keep it as is.

lib/util/common.ts
```diff
@@ -144,3 +147,15 @@ export function parseJsonc(content: string): JsonValue {
   }
   throw new Error('Invalid JSONC');
 }
+
+/**
+ * Use only if an option is inheirt + global
+ * for only global use GlobalConfig.class
```
jamietanna: Are there any issues using this if it's a global-only option?

lib/util/common.ts
```diff
@@ -144,3 +147,15 @@ export function parseJsonc(content: string): JsonValue {
...
+ */
+export function getInheritedOrGlobal<Key extends keyof RepoInheritConfig>(
```
jamietanna: let's get some unit tests for this, please - I can imagine breaking this in the future!

lib/util/common.ts
RahulGautamSingh: This seemed more cleaner + informative to me. 

Do you think using it for both makes sense?

lib/config/types.ts
RahulGautamSingh: It makes sense to change the `RepoInheritConfig` to `InheritAndGlobalConfig` (or similar), wdyt?

lib/config/types.ts
```diff
@@ -160,6 +160,13 @@ export interface RenovateSharedConfig {
   versionCompatibility?: string;
 }
 
+/**
+ * Contains all options with globalOnly=true && inheritConfigSupport=true
+ */
+export interface RepoInheritConfig {
```
viceice:
```suggestion
export interface GlobalInheritableConfigOptions {
```

??

viceice (commented):
otherwise lgtm

lib/config/global.ts
```diff
@@ -18,6 +18,7 @@ export class GlobalConfig {
     'cacheHardTtlMinutes',
     'cachePrivatePackages',
     'cacheTtlOverride',
+    'configFileNames',
```
viceice, line 21: we need to check if this calls are before setting global or inherited config set:

https://github.com/renovatebot/renovate/blob/82410bd59da18998b2f5aa24bf4ef3870782e2d9/lib/workers/global/config/parse/index.ts#L186-L193

https://github.com/renovatebot/renovate/blob/82410bd59da18998b2f5aa24bf4ef3870782e2d9/lib/workers/repository/init/inherited.ts#L103-L111

lib/config/types.ts
RahulGautamSingh: Yes, more accurate

lib/config/global.ts
RahulGautamSingh: They are:

1. `GlobalConfig.set()` is called within `renovateRepository` fn so the `configFileNames`  from global config file will be deleted before it can be set in `GlobalConfig`
2. The `InheritConfig.set()` is called in lines `113` & `171` of the `inherited.ts` file, bt which point `configFileNames` have been already removed

lib/util/common.spec.ts
```diff
@@ -186,4 +188,39 @@ describe('util/common', () => {
...
     });
   });
+
+  describe('getInheritedOrGlobal', () => {
+    beforeEach(() => {
+      GlobalConfig.reset();
+      InheritConfig.reset();
+    });
+
+    it('returns undefined', () => {
```
jamietanna:
```suggestion
    it('returns undefined if not set', () => {
```

lib/util/common.spec.ts
```diff
@@ -186,4 +188,39 @@ describe('util/common', () => {
...
+      expect(getInheritedOrGlobal('configFileNames')).toBeUndefined();
+    });
+
+    it('returns inherited value', () => {
```
jamietanna:
```suggestion
    it('returns inherited value if only inherited value is set', () => {
```

lib/util/common.spec.ts
```diff
@@ -186,4 +188,39 @@ describe('util/common', () => {
...
+      InheritConfig.set({
+        configFileNames: ['inherited'],
+      });
+      expect(getInheritedOrGlobal('configFileNames')).toEqual(['inherited']);
+    });
+
+    it('returns global value', () => {
```
jamietanna:
```suggestion
    it('returns global value if only global value is set', () => {
```

lib/util/common.spec.ts
```diff
@@ -186,4 +188,39 @@ describe('util/common', () => {
...
+      GlobalConfig.set({
+        configFileNames: ['global'],
+      });
+      expect(getInheritedOrGlobal('configFileNames')).toEqual(['global']);
+    });
+
+    it('returns inherited value - when both global + inherited are available', () => {
```
jamietanna:
```suggestion
    it('returns inherited value - when both global + inherited are set', () => {
```

lib/util/common.spec.ts
```diff
@@ -186,4 +188,39 @@ describe('util/common', () => {
...
+      expect(getInheritedOrGlobal('configFileNames')).toEqual(['inherited']);
+    });
+
+    it('returns global value', () => {
+      GlobalConfig.set({
+        configFileNames: ['global'],
+      });
+      expect(getInheritedOrGlobal('configFileNames')).toEqual(['global']);
+    });
+
```
jamietanna, line 215: do we have any nullable inherit-able options? What happens when inherited version is explicitly set to a `null`/`undefined`?

RahulGautamSingh: Good one, I'll check

lib/util/common.ts
jamietanna: I'm thinking about cases where contributors/maintainers may not know that the option is one/both so trying to make it as easy as possible to do

viceice: I would not allow use this if the option is not global and inheritable for cleaner code and easier maintenance

viceice (commented):
LGTM besides the open conversations

viceice (approved):

jamietanna (approved):

PR #40497, opened 2026-01-17
feat(datasource): add GraalVM version datasource 

zeitlinger:
Blocked by https://github.com/renovatebot/renovate/pull/42604

## Changes

                                                                                                                 
  Adds a new `graalvm-version` datasource to fix compatibility issues with Oracle GraalVM versions in the mise manager.                                                                                                                  
                                                                                                                                                                                                                                         
## Context
                                                                                                                                                                                                                                         
  The mise manager currently uses `JavaVersionDatasource` (Adoptium API) for all Java variants, including `oracle-graalvm-*` prefixes. However, Adoptium API versions are incompatible with GraalVM, causing failures like               
  https://github.com/prometheus/client_java/pull/1734.                   

- [ ] This closes an existing Issue, Closes: # <!-- NOTE that this should NOT be a Discussion -->

## AI assistance disclosure

<!-- We request this information to assist reviewers in identifying AI-generated errors and other issues specific to AI usage. While we typically permit the use of AI tools, we appreciate being notified when they are employed. -->

The public repository: <URL>                                                                                                                                                               
                                                                                                                                                                                                                                         
  ### Solution                                                                                                                                                                                                                           
                                                                                                                                                                                                                                         
  Created a new `graalvm-version` datasource that:                                                                                                                                                                                       
  - Queries the mise-java.jdx.dev API for Oracle GraalVM releases                                                                                                                                                                        
  - Filters results by vendor (`oracle-graalvm`) and image type (`jdk`/`jre`)                                                                                                                                                            
  - Uses system detection to map Node.js platform values to API parameters                                                                                                                                                               
  - Supports explicit platform specification via query parameters                                                                                                                                                                        
                                                                                                                                                                                                                                         
  **Package naming format:**                                                                                                                                                                                                             
  - `oracle-graalvm-jdk?system=true` - Auto-detect current system                                                                                                                                                                        
  - `oracle-graalvm-jre?os=macosx&architecture=aarch64` - Explicit platform                                                                                                                                                              
  - `oracle-graalvm-jdk?releaseType=ea` - Early access releases                                                                                                                                                                          
                                                                                                                                                                                                                                         
  **System detection mappings:**                                                                                                                                                                                                         
  - Architecture: `ia32`→`i686`, `x64`→`x86_64`, `arm64`→`aarch64`, `arm`→`arm32`                                                                                                                                                        
  - OS: `darwin`→`macosx`, `win32`→`windows`, `linux`→`linux`                                                                                                                                                                            
                                                                                                                                                                                                                                         
  ### Changes                                                                                                                                                                                                                            
                                                                                                                                                                                                                                         
  **New files:**                                                                                                                                                                                                                         
  - `lib/modules/datasource/graalvm-version/index.ts` - Main datasource implementation                                                                                                                                                   
  - `lib/modules/datasource/graalvm-version/common.ts` - System detection utilities                                                                                                                                                      
  - `lib/modules/datasource/graalvm-version/schema.ts` - Zod validation                                                                                                                                                                  
  - `lib/modules/datasource/graalvm-version/types.ts` - TypeScript types                                                                                                                                                                 
  - `lib/modules/datasource/graalvm-version/readme.md` - Documentation                                                                                                                                                                   
  - `lib/modules/datasource/graalvm-version/*.spec.ts` - Comprehensive tests                                                                                                                                                             
  - `lib/modules/datasource/graalvm-version/__fixtures__/` - Test fixtures                                                                                                                                                               
                                                                                                                                                                                                                                         
  **Modified files:**                                                                                                                                                                                                                    
  - `lib/modules/datasource/api.ts` - Registered new datasource                                                                                                                                                                          
  - `lib/modules/manager/mise/upgradeable-tooling.ts` - Updated oracle-graalvm to use new datasource                                                                                                                                     
  - `lib/modules/manager/mise/extract.spec.ts` - Updated test expectations                                                                                                                                                               
  - `lib/util/cache/package/namespaces.ts` - Added cache namespace                                                                                                                                                                       
                                                                                                                                                                                                                                         
  ### Test Plan                                                                                                                                                                                                                          
                                                                                                                                                                                                                                         
  - [x] All 20 graalvm-version datasource tests passing                                                                                                                                                                                  
  - [x] All 57 mise manager tests passing                                                                                                                                                                                                
  - [x] TypeScript type checking passing                                                                                                                                                                                                 
  - [x] Linting passing                                                                                                                                                                                                                  
  - [x] 100% test coverage for new datasource                                                                                                                                                                                            
                                                                                                                                                                                                                                         
  **Verification:**               
                                                                                                                                                                                                       
  ```bash                                                                                                                                                                                                                                
  pnpm vitest graalvm-version --run                                                                                                                                                                                                      
  pnpm vitest lib/modules/manager/mise --run                                                                                                                                                                                             
  pnpm type-check                                                                                                                                                                                                                        
  pnpm lint                                                                                                                                                                                                                              
```
                                                                                                                                                                                                                                         
 ### References                                                                                                                                                                                                                             
                                                                                                                                                                                                                                         
  - Fixes compatibility issue from https://github.com/prometheus/client_java/pull/1734                                                                                                                                                   
  - Uses mise-java.jdx.dev API: https://mise-java.jdx.dev/                                                                                                                                                                               
  - Related to jdx/mise#7229                                                                                                                                                                                                             
                                                                                                                                                                                                                                         
  ---                                                                                                                                                                                                                                    
  🤖 This PR was written primarily by Claude Code.

### CODE REVIEW

File: lib/modules/datasource/graalvm-version/schema.ts [hunk 1]
```diff
@@ -0,0 +1,26 @@
...
+  image_type: z.string(),
+  java_version: z.string(),
+  jvm_impl: z.string(),
+  url: z.string().url(),
+  vendor: z.string(),
+  version: z.string(),
+});
+
+// Transform to Renovate Release format
+export const GraalVmReleases = LooseArray(MiseJavaReleaseSchema).transform(
```
roele: should probably be named `MiseJavaReleases`?

lib/modules/datasource/graalvm-version/readme.md
```diff
@@ -0,0 +1,43 @@
+# graalvm-version
```
roele: references to specific GraalVM should be replaced with vendor neutral naming/examples

lib/modules/datasource/graalvm-version/common.ts
```diff
@@ -0,0 +1,86 @@
...
+import type { Nullish } from '../../../types';
+import { trimLeadingSlash } from '../../../util/url';
+import type { PackageConfig } from './types';
+
+export const defaultRegistryUrl = 'https://mise-java.jdx.dev/';
+
+export const datasource = 'graalvm-version';
+
+export function parsePackage(packageName: string): PackageConfig {
+  const u = new URL(packageName, defaultRegistryUrl);
```
roele: maybe would be easier to use the `packageName` as plain vendor and define the `imageType` in a separate query parameter (defaulting to `jdk`)?

jamietanna:
@zeitlinger once you've had a chance to accept the CLA and take a look at the code coverage, can you please also make sure you've followed our PR template? Thanks!

zeitlinger:
@jamietanna done

jamietanna:
PR template?

zeitlinger:
> PR template?

added

zeitlinger:
@jamietanna can you check again?

viceice (commented):
I would prefer to integrate it into the java-version datasource

zeitlinger:
> I would prefer to integrate it into the java-version datasource

done

lib/util/cache/package/namespaces.ts
```diff
@@ -69,6 +69,7 @@ export const packageCacheNamespaces = [
   'datasource-go-proxy',
   'datasource-go',
   'datasource-golang-version',
+  'datasource-graalvm-version',
```
viceice: is that still used?

lib/modules/datasource/java-version/types.ts
```diff
@@ -8,8 +8,23 @@ export interface AdoptiumJavaResponse {
   versions?: AdoptiumJavaVersion[];
 }
 
+export interface MiseJavaRelease {
```
viceice: create zod schema

lib/modules/datasource/java-version/index.ts
```diff
@@ -19,39 +13,12 @@ export class JavaVersionDatasource extends Datasource {
...
 
-  override readonly customRegistrySupport = false;
+  override readonly customRegistrySupport = true;
 
-  override readonly defaultRegistryUrls = [defaultRegistryUrl];
+  override readonly defaultRegistryUrls = [adoptiumRegistryUrl];
 
   override readonly caching = true;
 
-  private async getPageReleases(
```
viceice: please do this refactor in a separate PR to be merged before

lib/modules/datasource/java-version/graalvm.ts
```diff
@@ -0,0 +1,77 @@
+import { logger } from '../../../logger/index.ts';
+import { ExternalHostError } from '../../../types/errors/external-host-error.ts';
+import { type Http, HttpError } from '../../../util/http/index.ts';
+import { joinUrlParts } from '../../../util/url.ts';
+import type { ReleaseResult } from '../types.ts';
+import type { MiseJavaRelease, PackageConfig } from './types.ts';
+
+export const graalvmRegistryUrl = 'https://mise-java.jdx.dev/';
```
viceice, line 10: does this also work for adoptium java? then we should use this as new primary and fall back to adoptium on error. 🤔

zeitlinger: yes, mise-java.jdx.dev has temurin alongside 18 other vendors (corretto, zulu, liberica, oracle, etc.) - so it would be possible.

I'd suggest we do it step-by-step: this PR adds GraalVM via mise-java, follow-up PR migrates Adoptium (maybe others) to use it too

lib/modules/datasource/java-version/index.ts
zeitlinger: https://github.com/renovatebot/renovate/pull/42604

lib/util/cache/package/namespaces.ts
zeitlinger: no

lib/modules/datasource/java-version/types.ts
zeitlinger: done

lib/modules/datasource/java-version/graalvm.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const result: ReleaseResult = {
+    homepage: 'https://www.oracle.com/java/graalvm/',
+    sourceUrl: 'https://github.com/oracle/graal',
+    registryUrl,
+    releases: [],
+  };
+
+  try {
+    const response = await http.getJson(url, MiseJavaRelease.array().catch([]));
```
RahulGautamSingh: this drops all releases if one element fails validation. Use `LooseArray(MiseJavaRelease)` instead.

lib/modules/datasource/java-version/schema.ts
```diff
@@ -10,3 +10,18 @@ export const AdoptiumJavaResponse = z.object({
 });
 
 export type AdoptiumJavaResponse = z.infer<typeof AdoptiumJavaResponse>;
+
+export const MiseJavaRelease = z.object({
```
RahulGautamSingh, line 14: this requires several fields the code never reads; needlessly brittle

RahulGautamSingh (changes requested):

lib/modules/datasource/java-version/graalvm.ts
zeitlinger: Implemented in [ea6eb2b052](https://github.com/zeitlinger/renovate/commit/ea6eb2b052): the response now uses `LooseArray(MiseJavaRelease)`, so malformed entries are skipped while valid releases are retained. Added regression coverage.

lib/modules/datasource/java-version/schema.ts
zeitlinger: Implemented in [ea6eb2b052](https://github.com/zeitlinger/renovate/commit/ea6eb2b052): `MiseJavaRelease` now validates only the consumed fields (`image_type`, `vendor`, and `version`), avoiding unnecessary requirements on unused fields.

lib/modules/datasource/java-version/graalvm.ts
zeitlinger: Thanks for raising this. This PR intentionally keeps Adoptium behavior unchanged and limits mise integration to GraalVM. Migrating Adoptium to use mise as primary with Adoptium fallback would change existing behavior, so I am deferring it to a focused follow-up, consistent with the earlier step-by-step plan. Do you still consider that migration blocking for this PR?

lib/modules/datasource/java-version/index.ts
```diff
@@ -12,19 +13,33 @@ export class JavaVersionDatasource extends Datasource {
...
     const pkgConfig = parsePackage(packageName);
-    logger.trace({ packageName, pkgConfig }, 'fetching java release');
+    logger.trace(
+      { registryUrl, packageName, pkgConfig },
+      'fetching java release',
+    );
 
     try {
+      if (pkgConfig.vendor === 'oracle-graalvm') {
+        const effectiveRegistryUrl = registryUrl ?? graalvmRegistryUrl;
```
RahulGautamSingh, line 44: `registryUrl` is never nullish here, so this fallback never fires. 

`defaultRegistryUrls` is `[adoptiumRegistryUrl]` and `customRegistrySupport` is now true, so `resolveRegistryUrls` always passes `https://api.adoptium.net`. 

Every GraalVM lookup then requests `https://api.adoptium.net/jvm/ga/<os>/<arch>.json`, gets a 404 and returns `null`.

lib/modules/datasource/java-version/index.spec.ts
```diff
@@ -1,142 +1,617 @@
...
+          .reply(200, nonMatchingReleases);
+        const res = await getPkgReleases({
+          datasource,
+          packageName: 'oracle-graalvm-jdk?os=linux&architecture=x86_64',
+          registryUrls: [graalvmRegistryUrl],
+        });
+        expect(res).toBeNull();
+      });
+
+      it('falls back to default graalvm registry when registryUrl is undefined', async () => {
```
RahulGautamSingh, line 547: this calls the datasource directly and skips `getPkgReleases`, so it never exercises registry resolution. 

Every other GraalVM test passes `registryUrls: [graalvmRegistryUrl]` explicitly, which is why the default path is broken and still green. Go through `getPkgReleases` with no `registryUrls`.

lib/modules/datasource/java-version/index.ts
```diff
@@ -12,19 +13,33 @@ export class JavaVersionDatasource extends Datasource {
     super(datasource);
   }
 
-  override readonly customRegistrySupport = false;
+  override readonly customRegistrySupport = true;
```
RahulGautamSingh: flipping this to true changes Adoptium behavior. 

`massageRegistryUrls` now trims the trailing slash, which cause the snapshot changes in this PR, and the same value feeds the cache key `${registryUrl}:${packageName}`, so every existing java-version cache entry is invalidated. It also opens the datasource to arbitrary registries with nothing in the readme about it.

lib/modules/manager/mise/upgradeable-tooling.ts
```diff
@@ -144,7 +144,7 @@ const miseCoreTooling: Record<string, ToolingDefinition> = {
       if (oracleGraalvmJdkMatches) {
         return {
           datasource: JavaVersionDatasource.id,
-          packageName: 'java-jdk',
+          packageName: 'oracle-graalvm-jdk?system=true',
```
RahulGautamSingh: system=true ties the result to the machine running Renovate. 

java-jdk returned the same versions everywhere. On an unmapped platform such as s390x or riscv64 this now returns null and the dependency silently stops updating. 

Drop system=true so the lookup does not depend on the runner.

lib/modules/datasource/java-version/graalvm.ts
```diff
@@ -0,0 +1,78 @@
...
+    const response = await http.getJson(
+      url,
+      LooseArray(MiseJavaRelease).catch([]),
+    );
+
+    // Filter by vendor and image_type
+    const filteredReleases = response.body
+      .filter((release) => {
+        return (
+          release.vendor === 'oracle-graalvm' &&
```
RahulGautamSingh: use `pkgConfig.vendor` instead of the literal. The vendor is already parsed, and hardcoding it blocks the follow-up that adds other vendors.

lib/modules/datasource/java-version/graalvm.ts
```diff
@@ -0,0 +1,78 @@
...
+  // If system detection returns null, we can't fetch - return null
+  if (!pkgConfig.os || !pkgConfig.architecture) {
+    logger.debug(
+      { os: pkgConfig.os, architecture: pkgConfig.architecture },
+      'Cannot fetch GraalVM releases without OS and architecture',
+    );
+    return null;
+  }
+
+  const releaseType = pkgConfig.releaseType ?? 'ga';
```
RahulGautamSingh, line 29: `releaseType` goes from a user query parameter straight into the URL path. 

Validate it against ga and ea.

lib/modules/datasource/java-version/types.ts
```diff
@@ -1,7 +1,9 @@
 import type { Nullish } from '../../../types/index.ts';
 
 export interface PackageConfig {
+  vendor: 'adoptium' | 'oracle-graalvm';
```
RahulGautamSingh: export this union as a named type. It is repeated in three signatures in common.ts.

lib/modules/datasource/java-version/common.ts
```diff
@@ -12,52 +12,110 @@ export const datasource = 'java-version';
...
-    default:
-      return 'jdk';
+  if (name.includes('-jre') || name === 'java-jre') {
+    return 'jre';
   }
+  return 'jdk';
 }
 
-function getSystemArchitecture(useSystem: boolean): Nullish<string> {
+function getSystemArchitecture(
```
RahulGautamSingh, line 52: The vendor check wraps a full switch here and again in getSystemOs.

Split into getGraalvmArchitecture and getAdoptiumArchitecture to flatten it.

lib/modules/datasource/java-version/index.spec.ts
```diff
@@ -1,142 +1,617 @@
...
+          packageName:
+            'oracle-graalvm-jdk?os=linux&architecture=x86_64&release-type=ea',
+          registryUrls: [graalvmRegistryUrl],
+        });
+        expect(res).toBeNull();
+      });
+
+      it('returns null when system detection fails', async () => {
+        vi.spyOn(process, 'arch', 'get').mockReturnValueOnce(
+          'unsupported' as any,
```
RahulGautamSingh: drop the as any. 

common.spec.ts mocks the same getter with a plain string and no cast.

RahulGautamSingh (commented):
#42604 is merged, so this is no longer blocked. 

The mise-java primary with Adoptium fallback question is still open and needs maintainer input @viceice @secustor 


PS> claude used for review run

zeitlinger:
@RahulGautamSingh addressed the feedback

lib/modules/datasource/java-version/graalvm.ts
secustor: I think we should touch other datasources separately.

lib/modules/datasource/java-version/graalvm.ts
```diff
@@ -0,0 +1,78 @@
...
+    homepage: 'https://www.oracle.com/java/graalvm/',
+    sourceUrl: 'https://github.com/oracle/graal',
+    registryUrl,
+    releases: [],
+  };
+
+  try {
+    const response = await http.getJson(
+      url,
+      LooseArray(MiseJavaRelease).catch([]),
```
secustor:
```suggestion
      LooseArray(MiseJavaRelease),
```
This swallows transient errors in case another body is returned. Therefore it will close open PRs in that case

lib/modules/manager/mise/upgradeable-tooling.ts
```diff
@@ -142,12 +142,15 @@ const miseCoreTooling: Record<string, ToolingDefinition> = {
...
       if (oracleGraalvmJdkMatches) {
-        return {
+        const packageName = 'oracle-graalvm-jdk';
+        const result = {
           datasource: JavaVersionDatasource.id,
           packageName: 'java-jdk',
           currentValue: oracleGraalvmJdkMatches.version,
           ...shortJavaVersioning(oracleGraalvmJdkMatches.version),
         };
+        result.packageName = packageName;
```
secustor: This field should be simply set after the spread

lib/modules/datasource/java-version/graalvm.ts
```diff
@@ -0,0 +1,78 @@
...
+export async function getGraalvmReleases(
+  http: Http,
+  pkgConfig: PackageConfig,
+  registryUrl: string,
+): Promise<ReleaseResult | null> {
+  logger.trace({ pkgConfig, registryUrl }, 'fetching GraalVM releases');
+
+  // Build URL: /jvm/{releaseType}/{os}/{architecture}.json
+  // If system detection returns null, we can't fetch - return null
+  if (!pkgConfig.os || !pkgConfig.architecture) {
```
RahulGautamSingh, line 21: after dropping `system=true`, 

the datasource has no default to fall back on. `oracle-graalvm-jdk` carries no os or architecture, so both are null here and every `java = "oracle-graalvm-*"` entry resolves to no releases and silently stops updating. 

Default to a fixed platform such as linux and x86_64 when the package name does not specify one, and add a test that resolves a bare `oracle-graalvm-jdk` through `getPkgReleases`.

lib/modules/manager/mise/upgradeable-tooling.ts
RahulGautamSingh: +1

put packageName: 'oracle-graalvm-jdk' after the spread and drop the local, the two forms produce the same object. 

The GraalVM case in `extract.spec.ts` still asserts with toMatchObject and never checks packageName, so add that assertion.

RahulGautamSingh (changes requested):

zeitlinger:
@RahulGautamSingh @secustor addressed the feedback

lib/modules/datasource/java-version/graalvm.ts
zeitlinger: Agreed.

lib/modules/datasource/java-version/index.spec.ts
```diff
@@ -1,142 +1,647 @@
...
+      it('uses the default OS when it is omitted', async () => {
+        httpMock
+          .scope(graalvmRegistryUrl)
+          .get(graalvmBasePath)
+          .reply(200, oracleGraalvmJdkReleases);
+        const res = await getPkgReleases({
+          datasource,
+          packageName: 'oracle-graalvm-jdk?architecture=x86_64',
+        });
+        expect(res?.releases).toHaveLength(3);
```
secustor, line 471: These tests should actually test what is in the description

zeitlinger: Addressed in fce3ff66f9. The default-OS test now supplies non-default aarch64 and verifies /jvm/ga/linux/aarch64.json; the default-architecture test supplies non-default windows and verifies /jvm/ga/windows/x86_64.json. This makes each omitted/defaulted value explicit.

zeitlinger:
Updated and ready for re-review:

- merged the latest `main` and resolved the merge conflicts
- preserved the current non-snapshot Java datasource test structure
- strengthened the default-platform tests so each test distinguishes the supplied value from the omitted/defaulted value
- updated the datasource helper for the current lint rules
- resolved the remaining review threads

Validation:
- targeted tests: 172 passed
- `pnpm check --all` on all affected paths: passed
- GitHub CI, including coverage: passed

The PR is now mergeable and only requires review approval.

lib/modules/datasource/types.ts
```diff
@@ -162,6 +162,10 @@ export interface DatasourceApi extends ModuleApi {
   getDigest?(config: DigestConfig, newValue?: string): Promise<string | null>;
   getReleases(config: GetReleasesConfig): Promise<ReleaseResult | null>;
   defaultRegistryUrls?: string[] | (() => string[]);
+  /** Return registry URLs for package-specific datasource defaults. */
+  getDefaultRegistryUrls?(packageName: string): string[] | undefined;
+  /** Return whether custom registry URLs are supported for this package. */
+  supportsCustomRegistry?(packageName: string): boolean;
```
secustor: We should not have the functions and the values.

If the functions are really needed we should migrate to expose the function only.

zeitlinger:
Completed in [98a34637f1](https://github.com/renovatebot/renovate/commit/98a34637f1), following [d4c0bd2607](https://github.com/renovatebot/renovate/commit/d4c0bd2607): migrated the datasource API to function-only registry accessors across implementations, consumers, tests, and docs, with legitimate accessor coverage tests and no blanket coverage ignores. User configuration `defaultRegistryUrls` fields remain unchanged. Validation: 2,055 datasource tests passed with `GONOPROXY`/`GOPRIVATE` unset, 365 Mise/Terraform tests passed, type-check and Prettier passed.

Final follow-up in [9fed8f6553](https://github.com/renovatebot/renovate/commit/9fed8f6553b62f9e66490d0fa2d94decb94a1a26): restored only two pre-existing coverage annotations accidentally removed during migration cleanup (Docker and Go datasource paths), not new blanket ignores. CI run [35080047341](https://github.com/renovatebot/renovate/actions/runs/35080047341) is fully green, including all 16 test shards, coverage threshold, test-success, lint, build, docs, CodeQL, and codecov. Awaiting maintainer review.

lib/modules/datasource/types.ts
zeitlinger: Implemented in [02613a2e2d](https://github.com/zeitlinger/renovate/commit/02613a2e2d), migrating datasource registry configuration to function-only accessors across implementations, consumers, tests, and docs. `defaultRegistryUrls` user configuration remains unchanged. Type-check, Prettier, and affected tests pass.

secustor (approved):

PR #40548, opened 2026-01-20
docs(major-release): add new JSON schema requirements

jamietanna:
## Changes

As per #37994, following major versions we'll archive the old config version in Schema Store i.e. https://github.com/SchemaStore/schemastore/pull/5294

## Context

## AI assistance disclosure

The public repository: <URL>


viceice:
we should host them ourselves to update them more easily, that was the reason why we switched to the current deployment.

viceice (commented):
see other comment

jamietanna:
The intent here is that we'll only publish them after-the-fact to Schema Store once the major version is retired, so we shouldn't need to update them - the latest version (for the current major) will always only be accessible on `docs.renovatebot.com`

viceice (approved):
we should release the schema files as separate release assets, so users could directly use them as versioned source directly. 🤔

PR #40466, opened 2026-01-16
fix(mise)!: rename packageName from `nodejs` to `node`

steadexe:
## Changes

I changed the packageName extracted by mise manager from `nodejs` to `node` as it is called node inside the mise.toml file and every other manager extract it as node as well.

I discovered the issue while digging why preset `group:nodeJs` was ignoring the mise.toml update

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: # 
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

There is an existing discussion about this https://github.com/renovatebot/renovate/discussions/40420

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [x] Code inspection only, or
- [x] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #40567, opened 2026-01-21
chore: move from volta.sh to mise

secustor:
## Changes

Moving away from [Volta.sh as it is officially unmaintained](https://github.com/volta-cli/volta/issues/2080). 
They are recommending Mise and as we are supporting it with Renovate I think we should follow their advice. 

https://mise.jdx.dev/mise-cookbook/nodejs.html#example-with-pnpm

## Context

- [ ] This closes an existing Issue, 

## AI assistance disclosure

The public repository: <URL>


mise.toml
```diff
@@ -0,0 +1,21 @@
+# https://mise.jdx.dev/mise-cookbook/nodejs.html#example-with-pnpm
+[tools]
+# we could configure Mise to reuse `.nvmrc`, though that feature is opting in via `mise settings add idiomatic_version_file_enable_tools node`
+node = "24.13.0"
```
viceice: we should add a test to ensure they are in sync

mise.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[hooks]
+# Enabling corepack will install the `pnpm` package manager specified in package.json#packageManager
+postinstall = 'npx corepack enable'
```
viceice: doesn't it support install pnpm? pnpm will ensure proper version from package.json

secustor: Not directly, this is basically their cookbook see line 1 

Though corepack will use the `packageManager` field which we keep up to date

mise.toml
secustor: We are auto updating this via Renovate, any specific concerns so that we need additional tests? 

The depNames are `nodejs` ( mise manager) vs `node` ( nvm manager )
Though I agree we should broaden [group:nodejs](https://docs.renovatebot.com/presets-group/#groupnodejs) to cover `nodejs` so it is handled in a single PR

mise.toml
viceice: ok, so do not use mise for me as it will break my existing setup 😏

mise.toml
viceice: name will be changed soon
- #40466

mise.toml
secustor: Out of curiosity what do you use for version management of pnpm if not corepack?

viceice: install pnpm globally and let pnpm handle it similar to corepack

viceice (dismissed):

mise.toml
```diff
@@ -0,0 +1,3 @@
+[tools]
+node = "24.13.0"
```
viceice:
```suggestion
node = "24.15.0"
```

viceice (approved):

PR #40606, opened 2026-01-22
fix(onboardingAutoCloseAge): correctly handle partial days elapsed

jamietanna:
## Changes

As noted in #40604, we're seeing cases where we need to wait
`onboardingAutoCloseAge+1` days until the onboarding PR will be
auto-closed.

To fix this, we can refactor `getElapsedDays` to optionally return a
floating point number, which we can then perform more careful
calculations against.

As we're now being a little more careful with timings being used for our
date comparisons, we need to make sure that Vitest fakes our timer
during these tests.

Instead of long floating-point numbers, we should use a 2 decimal
place value for log messages, as they're more readable to humans.

## Context

- [x] This closes an existing Issue: Closes #40604

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/util/date.ts
+++ b/lib/util/date.ts
@@ -2,12 +2,15 @@ import { DateTime } from 'luxon';
 
 const ONE_MINUTE_MS = 60 * 1000;
 
-export function getElapsedDays(timestamp: string): number {
+export function getElapsedDays(timestamp: string, floor = true): number {
   const currentVersionTimestampDate = DateTime.fromISO(timestamp);
   const now = DateTime.now();
   const diffInDays = now.diff(currentVersionTimestampDate, 'days').as('days');
-  const ageInDays = Math.floor(diffInDays);
-  return ageInDays;
+  if (floor) {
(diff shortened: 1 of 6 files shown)


lib/util/date.ts
```diff
@@ -6,8 +6,7 @@ export function getElapsedDays(timestamp: string): number {
   const currentVersionTimestampDate = DateTime.fromISO(timestamp);
   const now = DateTime.now();
   const diffInDays = now.diff(currentVersionTimestampDate, 'days').as('days');
-  const ageInDays = Math.floor(diffInDays);
-  return ageInDays;
+  return diffInDays;
```
viceice, line 13: did you checked all usages? I don't like to break them somehow

viceice (dismissed):

lib/util/date.ts
jamietanna: I'd checked and seemed comfortable with it, but it's fair, let's make sure this is an opt-in 👍

viceice (approved):

PR #40619, opened 2026-01-22
fix(config/validation): handle platform-prefixed `onboardingConfigFileName`

RahulGautamSingh:
## Changes
-  Handle platform-prefixed `onboardingConfigFileName`

## Context

- [x] This closes an existing Issue, Closes: #40607

## AI assistance disclosure

- [x] Both unit tests + ran on a real repository (verified locally)

The public repository: <URL>

### DIFF
--- a/lib/config/validation.spec.ts
+++ b/lib/config/validation.spec.ts
@@ -1722,6 +1722,20 @@ describe('config/validation', () => {
       expect(errors).toHaveLength(0);
     });
 
+    it('handles prefixed onboardingConfigFileName', async () => {
+      const config = {
+        platform: 'forgejo',
+        onboardingConfigFileName: '.forgejo/renovate.json',
+      };
+      const { warnings, errors } = await configValidation.validateConfig(
+        'global',
+        // @ts-expect-error: not sure why
+        config,
+      );
+      expect(warnings).toHaveLength(0);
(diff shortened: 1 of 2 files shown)


viceice (dismissed):
it should also respect onboarding file names option, which is set in my sample

jamietanna (commented):
What happens when running `renovate-config-validator`? I think `config.platform` will be undefined in this case, and I'm not sure what'll now happen

RahulGautamSingh:
Platform should be set in global config. So it will be present

jamietanna:
I mean if you run `renovate-config-validator` as a standalone thing, without `platform` set (which is the default, folks wouldn't always be using it with a `config.js` next to it which sets `platform`)

viceice:
validator should respect `configFileNames` in addition to the global config

https://code.forgejo.org/renovate/renovate/src/branch/main/src/config.json

RahulGautamSingh:
@viceice @jamietanna I wonder if `onboardingConfigFileName` should only be verified as a string, cause `configFileNames` can anything. Plus, `configFileNames` can be present in inheritConfig too so it's not necessary that at time of validation it will be present in the same config as `onboardingConfigFileName`

viceice (approved):

PR #40630, opened 2026-01-23
ci: never cancel in-progress release jobs

jamietanna:
## Changes

I've noticed a couple of partial releases recently where we've cancelled
the release job while a release had already started, leading to a mix of
our artifacts pushed.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -745,6 +745,10 @@ jobs:
     if: github.repository == 'renovatebot/renovate' && github.event_name != 'pull_request'
     runs-on: ubuntu-latest
     timeout-minutes: 60
+    concurrency:
+      group: ${{ github.workflow }}-release-${{ github.ref }}
+      # never cancel a release that has already started, as it may have started pushing artifacts
+      cancel-in-progress: false
     permissions:
       contents: write
       issues: write


.github/workflows/build.yml
```diff
@@ -745,6 +745,10 @@ jobs:
     if: github.repository == 'renovatebot/renovate' && github.event_name != 'pull_request'
     runs-on: ubuntu-latest
     timeout-minutes: 60
+    concurrency:
+      group: ${{ github.workflow }}-release-${{ github.ref }}
+      # never cancel a release that has already started, as it may have started pushing artifacts
+      cancel-in-progress: false
```
jamietanna, line 751: I wonder if we should add some logic that runs - as the release job starts, and just before `semantic-release` - to check if we're currently the latest commit on `main`

If not, we cancel the job, and then cede to a newer one

That'll prevent some of the wasted CI time (and make sure that many merges to `main` at a similar time will still cancel for each other

viceice (approved):
that's new? i didn't notices this is already possible on job level. i  only know at workflow level.

what happens when we set cancel in workflow to true but to false in job?

let's try this
https://github.com/renovatebot/renovate/blob/8a9b119144fd6edf84d8ea28156be7e1828195d9/.github/workflows/build.yml#L29-L31

jamietanna:
Try what, sorry?

viceice:
> Try what, sorry?

combination of job and workflow in progress cancel

PR #40682, opened 2026-01-26
fix(config): ensure that config options are immutable

jamietanna:
## Changes

Noticed while making changes in the JSON Schema generation, but we're
currently returning a mutable reference to all of our options, which
means that it might be possible to (accidentally) change configuration
options that then affects all of the running Renovate process.

To handle this, we can make sure to `Object.freeze` the options, as well
as making sure our types clarify this is read-only.

This isn't something that should be user-facing, but makes sure that
we're keeping these immutable.

We also need to make sure any modifications now return a new config
option.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -1,3 +1,4 @@
+import { isArray, isObject } from '@sindresorhus/is';
 import { AllManagersListLiteral } from '../../manager-list.generated';
 import { getManagers } from '../../modules/manager';
 import { getCustomManagers } from '../../modules/manager/custom';
@@ -6,7 +7,7 @@ import { getVersioningList } from '../../modules/versioning';
 import { supportedDatasources } from '../presets/internal/merge-confidence';
 import { type RenovateOptions, UpdateTypesOptions } from '../types';
 
-const options: RenovateOptions[] = [
+const options: Readonly<RenovateOptions>[] = [
   {
     name: 'mode',
     description: 'Mode of operation.',
@@ -3335,7 +3336,7 @@ const options: RenovateOptions[] = [
   },
 ];
 
-export function getOptions(): RenovateOptions[] {
+export function getOptions(): Readonly<RenovateOptions>[] {
   return options;
 }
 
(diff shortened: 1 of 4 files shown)


viceice (commented):
i think this will be a performance regression because we call this function a lot of times.


We should freeze the objects instead

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

lib/config/options/index.ts
```diff
@@ -3359,4 +3359,9 @@ function loadManagerOptions(): void {
   }
 }
 
+function freezeConfigOptions(): void {
+  options.map((o) => Object.freeze(o));
```
viceice: i assume no-one will change the `default` , `additionalProperties` or `allowedValues` 

maybe better use a normal loop and freeze the sub-objects above if they are objects?

viceice (commented):
otherwise LGTM

viceice (dismissed):

viceice:
failing tests 😁

viceice (dismissed):

lib/config/validation.ts
```diff
@@ -260,7 +260,7 @@ export async function validateConfig(
         !optionParents[key].includes(parentName as AllowedParents)
       ) {
         // TODO: types (#22198)
-        const options = optionParents[key]?.sort().join(', ');
+        const options = [...optionParents[key]].sort().join(', ');
```
viceice: this looks nicer
```suggestion
        const options = optionParents[key]?.toSorted().join(', ');
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted

lib/config/options/index.ts
```diff
@@ -3359,4 +3360,27 @@ function loadManagerOptions(): void {
...
 }
 
+function freeze(value: unknown): void {
+  if (isObject(value)) {
+    for (const v of Object.values(value)) {
+      freeze(v);
+    }
+
+    Object.freeze(value);
+  } else if (isArray(value)) {
```
viceice: array check needs to be before isObject because arrays are also objects

viceice (approved):

PR #40683, opened 2026-01-26
fix(json-schema): add a separate schema for Inherit Config

jamietanna:
## Changes

In shared preset repositories, some users will use a mix of Inherit
Config and Repo Config options with the existing JSON Schema.

As part of Renovate 43, we're removing the ability for any global
options to be usable in Repo Config, which would also mean that it's now
either a case of "ignoring warnings in my editor" and "using Global
Config JSON Schema", which isn't a great option.

We can instead create a third JSON Schema, one which is focussed on
shared presets that can include Inherit Config options.

This makes it clear that the repo config's JSON Schema should not be
used for anything Inherit Config.

## Context

Alternative to https://github.com/renovatebot/renovate/pull/40681

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/tools/docs/schema.ts
+++ b/tools/docs/schema.ts
@@ -170,6 +170,7 @@ function createSchemaForChildConfigs(
 interface GenerateSchemaOpts {
   filename?: string;
   version?: string;
+  isInherit?: boolean;
   isGlobal?: boolean;
 }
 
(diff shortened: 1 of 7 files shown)


tools/docs/schema.ts
```diff
@@ -178,29 +179,39 @@ export async function generateSchema(
...
+
+  if (isGlobal) {
+    schema.title = `JSON schema for Renovate ${version} global self-hosting configuration (https://renovatebot.com/)`;
+  }
+
+  if (isInherit) {
+    schema.title = `JSON schema for Renovate ${version} config files (with Inherit Config options) (https://renovatebot.com/)`;
+  }
+
   const configurationOptions = getOptions();
```
viceice: we should clone here instead of inside `getOptions` for performance

jamietanna: We're doing this a little better in https://github.com/renovatebot/renovate/pull/40682 which will then make sure that this is safer for other usages 👍

tools/docs/schema.ts
```diff
@@ -178,32 +179,42 @@ export async function generateSchema(
...
     allowComments: true,
     type: 'object',
     properties: {},
   };
+
+  if (isGlobal) {
+    schema.title = `JSON schema for Renovate ${version} global self-hosting configuration (https://renovatebot.com/)`;
+  }
+
+  if (isInherit) {
```
viceice:
```suggestion
  else if (isInherit) {
```
?

viceice (dismissed):
only an optional cosmetic thing

tools/docs/schema.ts
viceice: never mind, it can be both

jamietanna: It shouldn't be, though 🤔

viceice (approved):

Issue #9890, opened 2021-05-06
Switch from cjs to esm


viceice:
**What would you like Renovate to be able to do?**
As more and more dependencies are only available as esm we should also start transitioning to esm.

Full esm is supported since node [v12.22](https://nodejs.org/dist./v12.22.0/docs/api/esm.html#esm_modules_ecmascript_modules), v14 has [no stable esm](https://nodejs.org/dist./v14.16.0/docs/api/esm.html#esm_modules_ecmascript_modules)

**Did you already have any implementation ideas?**

1. First start publishing renovate as cjs and esm, so downstream projects can start testing esm too.
2. Remove cjs from renovate

*Blocker**
- https://github.com/facebook/jest/pull/10976

viceice:
![image](https://user-images.githubusercontent.com/1798109/117301370-e1a71a00-ae7a-11eb-8b0a-e20fe9f817f8.png)
https://nodejs.org/api/esm.html#esm_modules_ecmascript_modules

viceice:
v14 is esm stable too, but docs update not yet released

ref: nodejs/node#38409 nodejs/node#38480

thernstig:
> Full esm is supported since node [v12.22](https://nodejs.org/dist./v12.22.0/docs/api/esm.html#esm_modules_ecmascript_modules), v14 has [no stable esm](https://nodejs.org/dist./v14.16.0/docs/api/esm.html#esm_modules_ecmascript_modules)

I just happened to stumble upon this, and wanted to be a nice citizen 😄 

v14 does have stable ESM support, that is just a bug in the docs. See https://github.com/nodejs/node/issues/38409

voxpelli:
Was this solved by #9887 or additional work needed? Found some comments in #11133 around ESM that confused me. Is ESM modules considered not usable by Renovate yet or?

viceice:
See description of issue, this is blocked by jest issue.

HonkingGoose:
@viceice would having this in Jest be enough to get us going?

https://github.com/facebook/jest/pull/11818

viceice:
> @viceice would having this in Jest be enough to get us going?
> 
> [facebook/jest#11818](https://github.com/facebook/jest/pull/11818)

I don't think so, it's very basic and experimental. Lets wait for more stabillity.

rarkins:
Blocked by #13181

PR #40476, opened 2026-01-16
chore: ensure that `pnpm` also enforces `minimumReleaseAge`

jamietanna:
## Changes

On top of our Renovate configuration[0] not allowing Renovate to prepare
updates until npm packages have been available for 7 days, we can
provide "defence in depth" by also wiring in pnpm's configuration.

This takes the rules from [0] and applies them to explicitly allowlist
our packages, as well as set the limit to 7 days.

[0]: https://github.com/renovatebot/.github/blob/095f454e8b1141edb1045dd389bee4c59c51613e/default.json

## Context

Please select one of the below:

- [ ] This closes an existing Issue: Closes #
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [x] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #40686, opened 2026-01-26
build: Switch to rolldown

zharinov:
## Changes

Switch the build tool from `tsc` to `rolldown` while keeping `"type": "commonjs"` in package.json:

- Add `rolldown.config.ts` with preserved module structure (`preserveModules: true`)
- Replace `tsc -p tsconfig.app.json` with `rolldown -c` for compilation
- Update `tsconfig.json`: replace `outDir` with `noEmit` (rolldown handles emit)
- Delete `tsconfig.app.json` (no longer needed)
- Bundle `@oxc-project/runtime` for decorator support

Output:
- Same file structure as before (`dist/` with modules preserved)
- CJS format with `.js` extensions (unchanged from before)
- Sourcemaps included

## Context

- Ref: #9890

## AI assistance disclosure

Claude Code assisted with implementing and testing the rolldown configuration.

Verified:
- `pnpm type-check` passes
- `pnpm build` produces correct output structure
- `node dist/renovate.js --version` works
- `node dist/config-validator.js --help` works
- Unit tests pass


jamietanna (commented):
Should we use `fix:` for the PR, so it releases and we can work out sooner than later if there are any related issues?

rolldown.config.ts
```diff
@@ -0,0 +1,23 @@
+import { defineConfig } from 'rolldown';
+
+export default defineConfig({
+  input: ['lib/renovate.ts', 'lib/config-validator.ts'],
+  output: {
+    dir: 'dist',
+    format: 'cjs',
+    entryFileNames: '[name].cjs',
+    chunkFileNames: '[name].cjs',
```
viceice: please retain `.js` extension for now, the extension change needs to be done in a separate PR because it a breaking change

rolldown.config.ts
```diff
@@ -0,0 +1,23 @@
...
+  },
+  moduleTypes: {
+    '.cjs': 'js',
+  },
+  platform: 'node',
+  external: (id) =>
+    !id.startsWith('.') &&
+    !id.startsWith('/') &&
+    !id.startsWith('lib/') &&
+    !id.startsWith('@oxc-project/runtime'),
```
viceice, line 26: can we change this to a regulat expression? (perf)
```suggestion
  external: /^(?![./]|lib\/|@oxc-project\/runtime)/,
```

- https://rolldown.rs/reference/InputOptions.external

rolldown.config.ts
viceice: @secustor @jamietanna This is technically breaking change, but this package is also not expected to be used as a library.

I know it will break some mend usages and our own internal tools (this can be easily fixed and will be catched on ci)

rolldown.config.ts
zharinov: Why regex doesn't work here:
- Rolldown tests regex against both unresolved ID (`@oxc-project/runtime`) AND resolved path (`@oxc-project+runtime@0.110.0/...`)
- The resolved pnpm path has a leading space and + instead of /
- Functions only receive the unresolved ID, making them reliable for this edge case

The function is needed to bundle the `@oxc-project/runtime` decorator helpers.

rolldown.config.ts
zharinov: Reverted this one to produce `.js`

package.json
```diff
@@ -147,7 +147,7 @@
   },
   "volta": {
     "node": "24.13.0",
-    "pnpm": "10.28.1"
+    "pnpm": "10.28.0"
```
viceice:
```suggestion
    "pnpm": "10.28.1"
```

package.json
```diff
@@ -366,7 +367,7 @@
     "vitest": "4.0.18",
     "vitest-mock-extended": "3.1.0"
   },
-  "packageManager": "pnpm@10.28.1",
+  "packageManager": "pnpm@10.28.0",
```
viceice:
```suggestion
  "packageManager": "pnpm@10.28.1",
```

viceice (commented):
just a fixup for wrong merge

viceice (dismissed):

jamietanna (commented):
What does this mean for local dev? Can I still use `npm exec tsc` or do I need to use `npm exec rolldown -c` or similar?

package.json
```diff
@@ -353,6 +353,7 @@
     "nyc": "17.1.0",
     "oxlint": "1.39.0",
     "rimraf": "6.1.2",
+    "rolldown": "1.0.0-rc.1",
```
jamietanna, line 356: ⚠️ released 4 days ago

https://github.com/renovatebot/renovate/pull/40476 and our general rules would block this

rolldown.config.ts
jamietanna: Worth adding a comment to that effect? (both for "can we speed this up in the future" and "why do we have this")

package.json
viceice: maybe excluded this package until it's final?

viceice:
> What does this mean for local dev? Can I still use `npm exec tsc` or do I need to use `npm exec rolldown -c` or similar?

for what to do? you should still be able to run tsc, but it'll not output any code. compile sould be done with `pnpm build` or `pnpm compile:ts`

viceice (dismissed):

package.json
zharinov: Probably, it's okay to make an exception for transpiler, as far as it builds and doesn't break downstream consumers

jamietanna (approved):
Let's do this 👍

viceice (approved):

jamietanna:
We appear to be missing `.d.ts` files - investigating

PR #40696, opened 2026-01-26
build: Switch to `tsdown` for `.d.ts` generation

zharinov:
## Summary

- Switch from `rolldown` + `rolldown-plugin-dts` to `tsdown` (which bundles both)
- Fixes missing `.d.ts` files from #40686
- Rename `rolldown.config.ts` → `tsdown.config.mts`

tsdown is the official library bundler built on rolldown, with built-in dts generation via `rolldown-plugin-dts`.

## Test plan

- [x] `pnpm build` produces `.js`, `.d.ts`, and `.js.map` files
- [x] `node dist/renovate.js --version` works


viceice (dismissed):

viceice (approved):
let's test it

PR #40700, opened 2026-01-26
refactor: Rewrite imports with `.ts` extensions

zharinov:
## Changes

Rewrite all internal imports to use explicit `.ts` extensions. This enables:
- `allowImportingTsExtensions` in tsconfig
- Future native TypeScript execution (Node.js type stripping, Bun)
- Stricter module resolution

Also adds `eslint-plugin-file-extension-in-import-ts` to enforce `.ts` extensions going forward.

## Context

## AI assistance disclosure

Used AI to perform the bulk rewrite of ~2100 files.

Build passes, type-check passes.


eslint.config.mjs
```diff
@@ -3,6 +3,7 @@ import js from '@eslint/js';
 import vitest from '@vitest/eslint-plugin';
 import eslintConfigPrettier from 'eslint-config-prettier';
 import { createTypeScriptImportResolver } from 'eslint-import-resolver-typescript';
+import fileExtensionInImportTs from 'eslint-plugin-file-extension-in-import-ts';
```
viceice: is there something similar in oxlint?

lib/config/decrypt.spec.ts
```diff
@@ -2,10 +2,10 @@ import {
...
   getAzureCollection,
   validateDecryptedValue,
-} from './decrypt';
-import { GlobalConfig } from './global';
-import type { RenovateConfig } from './types';
-import { logger } from '~test/util';
+} from './decrypt.ts';
+import { GlobalConfig } from './global.ts';
+import type { RenovateConfig } from './types.ts';
+import { logger } from '~test/util.js';
```
viceice: why js here?

viceice (commented):
first round

lib/instrumentation/instrumented.ts
viceice, line 1: please do the decorator changes in a separate PR

tools/validate-schema.ts
```diff
@@ -2,7 +2,10 @@ import { promises } from 'fs';
 import type { ErrorObject, ValidateFunction } from 'ajv';
 import { Ajv } from 'ajv';
 import draft7MetaSchema from 'ajv/lib/refs/json-schema-draft-07.json';
-import addFormats from 'ajv-formats';
+import _addFormats from 'ajv-formats';
+
+const addFormats =
+  _addFormats as unknown as typeof import('ajv-formats').default;
```
viceice: why this?

viceice (commented):
otherwise LGTM

eslint.config.mjs
zharinov: Good point, switched to oxlint. For some reason, I didn't find it for the first time.

viceice:
isn't the linter showing wrong imports which are causing the test failures?  we should use the dynamic import mock variants, that would be catched by typescript

viceice (approved):

PR #40703, opened 2026-01-26
chore: replace `URL.parse`

### LINKED ISSUE #7797: Replace url.parse
`url.parse` is deprecated, and is likely to be removed in a new version of Node.js

> Use of the legacy `url.parse()` method is discouraged. Users should use the WHATWG `URL` API. Because the `url.parse()` method uses a lenient, non-standard algorithm for parsing URL strings, security issues can be introduced. Specifically, issues with host name spoofing and incorrect handling of usernames and passwords have been identified.

[See here for more information][nodejs].

It looks like the replacement is to do something like

```ts
let valid = false
try {
    new URL(dep.registryUrl);
    valid = true;
} catch {}
```

[nodejs]: https://nodejs.org/docs/latest-v15.x/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost

_Originally posted by @JamieMagee in https://github.com/renovatebot/renovate/pull/7792#r528994628_

RahulGautamSingh:
## Changes
- Replace `URL.parse` with WHATWG API equiv.

## Context

- [x] This closes an existing Issue, Closes: #7797

## AI assistance disclosure

The public repository: <URL>


lib/modules/platform/github/index.ts
```diff
@@ -126,7 +125,7 @@ export function isGHApp(): boolean {
 
 export async function detectGhe(token: string): Promise<void> {
   platformConfig.isGhe =
-    URL.parse(platformConfig.endpoint).host !== 'api.github.com';
+    new URL(platformConfig.endpoint).host !== 'api.github.com';
```
viceice, line 131: why host here and hostname on bitbucket?

RahulGautamSingh: keeping them as they were.

viceice: we should use hostname here too like the same check below

viceice:
can you check why coverage fails?

lib/modules/platform/github/index.ts
RahulGautamSingh: I think I'll do that in another PR as it's not related to this one.

RahulGautamSingh:
codecov is using mixing code changes :) might be a caching issue on their side.

viceice (approved):

PR #33113, opened 2024-12-14
feat(manager): support mise backends

risu729:
## Changes

Supports mise backends.
https://mise.jdx.dev/dev-tools/backends/

- [x] `core`
- [x] `asdf`
- [x] `aqua`
- [x] `cargo`
- [x] `dotnet`
- [x] `gem`
- [x] `go`
- [x] `npm`
- [x] `pipx`
- [x] `spm`
- [x] `ubi`
- [x] `vfox`

### Default Registry

Since the default registry now includes several backends, we need to reconsider how to store upgradeable tooling information.

Previously, `mise` only supported `asdf` plugins (and core plugins), but it now supports multiple backends. The [default registry](https://github.com/jdx/mise/blob/main/registry.toml) includes various backends, and some tools have fallbacks.

For example:

```
act                          aqua:nektos/act ubi:nektos/act asdf:gr1m0h/asdf-act                                 
```

Currently, Renovate supports these short names if they are manually registered in `upgradeable-tooling.ts`. However, we could potentially support tools not registered there by reading the registry. This approach would allow us to infer package names for backends such as `aqua`, `ubi`, `pipx`, etc.

Could creating a new datasource solve this issue? I’m not very familiar with how datasources work, but it seems the `go` datasource functions similarly by supporting multiple backend datasources.

Additionally, tools in the registry can include options like:

```
ubi:bitwarden/sdk[tag_regex=^bws,exe=bws]
```

Since the only relevant option for us is `tag_regex`, and only two tools in the registry use `tag_regex`, we could ignore it.

### Aqua Backend

The `aqua` backend is based on [aqua-registry](https://github.com/aquaproj/aqua-registry), a registry for [aqua](https://aquaproj.github.io/), a CLI version manager. This registry defines how to download GitHub release binaries and specifies which ones to use.

It might be beneficial to set the `extractVersion` property for GitHub releases/tags using the [`version_filter`](https://aquaproj.github.io/docs/reference/registry-config/version-prefix) from the `aqua` registry. Currently, we define `extractVersion` manually in `upgradeable-tooling.ts`.

While Aqua has its own Renovate config preset, [aqua-renovate-config](https://aquaproj.github.io/docs/products/aqua-renovate-config/), it also sets `extractVersion` manually.

### `v` prefix
mise automatically strips `v` prefix. Even if the version is set to `1.7.5` and the github release is `v1.7.5`, it can be installed for `aqua` and `ubi` backend.
[...]

PR #40706, opened 2026-01-26
feat(mise): add support for github backend

oschwald:
## Changes

This pull request adds support for the [mise `github` backend](https://mise.jdx.dev/dev-tools/backends/github.html). This is similar to the `ubi` backend except the version constraints allowed for it seem to be more limited.

## Context

## AI assistance disclosure

I used Claude Code to help with this PR. In particular, it helped significantly with the tests and making sure everything outside of the core logic was updated appropriately.

The public repository:

I tested it against our internal GitHub Enterprise Server instance.

### DIFF
--- a/lib/modules/manager/mise/backends.spec.ts
+++ b/lib/modules/manager/mise/backends.spec.ts
@@ -133,7 +133,7 @@ describe('modules/manager/mise/backends', () => {
         packageName: 'some/repo',
         datasource: 'github-releases',
         currentValue: '1.0.0',
-        extractVersion: '^release-(?<version>.+)',
+        extractVersion: '^release\\-(?<version>.+)',
       });
     });
 
@@ -146,7 +146,7 @@ describe('modules/manager/mise/backends', () => {
         packageName: 'some/repo',
         datasource: 'github-releases',
         currentValue: 'v1.0.0',
-        extractVersion: '^version-(?<version>.+)',
+        extractVersion: '^version\\-(?<version>.+)',
       });
     });
 
--- a/lib/modules/manager/mise/backends.spec.ts
+++ b/lib/modules/manager/mise/backends.spec.ts
@@ -117,15 +117,6 @@ describe('modules/manager/mise/backends', () => {
       });
     });
 
-    it('should set extractVersion if the version does not have leading v', () => {
-      expect(createGithubToolConfig('cli/cli', '2.64.0', {})).toStrictEqual({
-        packageName: 'cli/cli',
-        datasource: 'github-releases',
-        currentValue: '2.64.0',
-        extractVersion: '^v?(?<version>.+)',
-      });
-    });
-
     it('should not set extractVersion if the version has leading v', () => {
       expect(createGithubToolConfig('cli/cli', 'v2.64.0', {})).toStrictEqual({
         packageName: 'cli/cli',
--- a/lib/modules/manager/mise/backends.spec.ts
+++ b/lib/modules/manager/mise/backends.spec.ts
@@ -113,7 +113,6 @@ describe('modules/manager/mise/backends', () => {
         packageName: 'BurntSushi/ripgrep',
         datasource: 'github-releases',
         currentValue: '14.1.1',
-        extractVersion: '^v?(?<version>.+)',
       });
     });
 
@@ -158,7 +157,6 @@ describe('modules/manager/mise/backends', () => {
         packageName: 'some/repo',
         datasource: 'github-releases',
         currentValue: '1.0.0',
-        extractVersion: '^(?<version>.+)',
       });
     });
 
--- a/lib/modules/manager/mise/backends.spec.ts
+++ b/lib/modules/manager/mise/backends.spec.ts
@@ -3,6 +3,7 @@ import {
   createCargoToolConfig,
   createDotnetToolConfig,
   createGemToolConfig,
+  createGithubToolConfig,
   createGoToolConfig,
   createNpmToolConfig,
   createPipxToolConfig,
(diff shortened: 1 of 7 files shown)


lib/modules/manager/mise/backends.ts
```diff
@@ -110,6 +110,43 @@ export function createGemToolConfig(name: string): BackendToolingConfig {
...
+        extractVersion = '^(?<version>.+)';
+      }
+    } else {
+      // Custom prefix - escape special regex chars
+      const escapedPrefix = prefix.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+      extractVersion = `^${escapedPrefix}(?<version>.+)`;
+    }
+  } else if (!hasVPrefix) {
+    // Default: strip 'v' prefix if current version doesn't have it
+    extractVersion = '^v?(?<version>.+)';
```
viceice: I don't like that, it causes performance loose when it's not needed. it should be configured explicit if needed. 🤔

viceice (commented):
otherwise LGTM

lib/modules/manager/mise/backends.ts
oschwald: This was inspired by the `ubi` backend. The `aqua` backend seems to _always_ use `extractVersion`. Could you expand a bit on what your preferred solution might look like?

lib/modules/manager/mise/backends.spec.ts
```diff
@@ -104,6 +105,109 @@ describe('modules/manager/mise/backends', () => {
...
+        createGithubToolConfig('BurntSushi/ripgrep', '14.1.1', {}),
+      ).toStrictEqual({
+        packageName: 'BurntSushi/ripgrep',
+        datasource: 'github-releases',
+        currentValue: '14.1.1',
+        extractVersion: '^v?(?<version>.+)',
+      });
+    });
+
+    it('should set extractVersion if the version does not have leading v', () => {
```
RahulGautamSingh: Isn't this already covered in the previous test?

lib/modules/manager/mise/backends.ts
```diff
@@ -110,6 +110,43 @@ export function createGemToolConfig(name: string): BackendToolingConfig {
...
+  version: string,
+  toolOptions: MiseToolOptions,
+): BackendToolingConfig {
+  let extractVersion: string | undefined = undefined;
+  const hasVPrefix = version.startsWith('v');
+
+  if (isString(toolOptions.version_prefix)) {
+    const prefix = toolOptions.version_prefix;
+    if (prefix === '') {
+      // Empty prefix - no extractVersion needed if version has 'v'
```
RahulGautamSingh:
```suggestion
      // Empty prefix - no extractVersion needed if version starts with 'v'
```

lib/modules/manager/mise/backends.ts
```diff
@@ -110,6 +110,43 @@ export function createGemToolConfig(name: string): BackendToolingConfig {
...
+      }
+    } else {
+      // Custom prefix - escape special regex chars
+      const escapedPrefix = prefix.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+      extractVersion = `^${escapedPrefix}(?<version>.+)`;
+    }
+  } else if (!hasVPrefix) {
+    // Default: strip 'v' prefix if current version doesn't have it
+    extractVersion = '^v?(?<version>.+)';
+  }
```
RahulGautamSingh: Less nested ifs 

```suggestion
 if (isNonEmptyString(prefix)) {
    // Custom prefix - escape special regex chars
    const escapedPrefix = prefix.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
    extractVersion = `^${escapedPrefix}(?<version>.+)`;
  }

  // Empty prefix - no extractVersion needed if version has 'v'
  if (isEmptyString(prefix) && !hasVPrefix) {
    extractVersion = '^(?<version>.+)';
  }

  if (!hasVPrefix) {
    // Default: strip 'v' prefix if current version doesn't have it
    extractVersion = '^v?(?<version>.+)';
  }

```

RahulGautamSingh (changes requested):

lib/modules/manager/mise/backends.ts
oschwald: Although this seems a bit more performance intensive, I modified it. I did need to switch to `else if` to prevent the last if from overwriting `extractVersion` if it was set earlier. I could have an explicit check instead though if you prefer that.

RahulGautamSingh (approved):
LGTM

oschwald:
Do you want me to rebase and force push? I see there is now a conflict.

oschwald:
I rebased this. Please let me know if there is anything else that needs to be changed.

lib/modules/manager/mise/backends.ts
viceice: I'm concerned about performance. that regex is applied to all versions which can be expensive if you do it on a lot versions multiple times

viceice (commented):
see open conversation

lib/modules/manager/mise/backends.ts
oschwald: Thanks for your feedback. This follows the same pattern as the `ubi` backend, which it is replacing. (`mise` deprecated `ubi` after introducing `github`.) If the version has a v prefix, it will not be used. We do need it when there isn't a `v` as `mise` may be matching a tag called `v<version>` or just `<version>`. That said, I am not intimately familiar with the Renovate infrastructure, and I am happy to handle this another way if there is one. I would appreciate any input on that you may have.

oschwald: To clarify the underlying issue: **mise doesn't require the version in the config to match the tag format**. If a user specifies `1.2.3` in their `mise.toml`, mise will successfully match against either a `v1.2.3` tag or a `1.2.3` tag at install time. This means Renovate can't know which tag format the repository uses just by looking at the config file.

Without `extractVersion`, Renovate would miss upgrades in one of these scenarios:
- User has `1.2.3` in config, but repo uses `v1.2.4` tags → Renovate wouldn't find the upgrade
- User has `v1.2.3` in config, but repo uses `1.2.4` tags → Renovate wouldn't find the upgrade

I looked for a "try these exact tag names" mechanism in Renovate but couldn't find one. The `extractVersion` regex appears to be the standard approach for this problem.

Note that this same pattern is already used by the existing [`aqua`](https://github.com/renovatebot/renovate/blob/e5c2caac6540e0d05fd13bd6006c89f2f7229da5/lib/modules/manager/mise/backends.ts#L39) and [`ubi`](https://github.com/renovatebot/renovate/blob/e5c2caac6540e0d05fd13bd6006c89f2f7229da5/lib/modules/manager/mise/backends.ts#L200-L237) backends for mise, introduced in [#33113](https://github.com/renovatebot/renovate/pull/33113) (which you reviewed). The `github` backend replaces the now-deprecated `ubi`. If there's a better approach, I'm happy to update all three for consistency.

oschwald:
I responded in the above conversation. Please let me know if there is anything more I can do to move this forward.

lib/modules/manager/mise/backends.ts
viceice: wrong, renovate uses semver-coerce versioning for GitHub releases by default, so for renovate `v1.2.3 == 1.2.3`

oschwald: Thank you. I did not realize that. I pushed a new commit that simplifies the code.

I believe the existing `ubi` backend has the same issue, but I did not modify it to keep the scope of this PR small.

lib/modules/manager/mise/backends.ts
```diff
@@ -110,6 +115,32 @@ export function createGemToolConfig(name: string): BackendToolingConfig {
...
+  name: string,
+  version: string,
+  toolOptions: MiseToolOptions,
+): BackendToolingConfig {
+  let extractVersion: string | undefined = undefined;
+  const prefix = toolOptions.version_prefix;
+
+  if (isNonEmptyString(prefix)) {
+    // Custom prefix - escape special regex chars
+    const escapedPrefix = prefix.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
viceice: use our regex escape method

https://github.com/renovatebot/renovate/blob/aeab85564c53f87b26d8ab3ba193f2956a08655c/lib/util/regex.ts#L71

lib/modules/manager/mise/backends.ts
```diff
@@ -110,6 +115,32 @@ export function createGemToolConfig(name: string): BackendToolingConfig {
...
+    extractVersion = `^${escapedPrefix}(?<version>.+)`;
+  }
+
+  return {
+    packageName: name,
+    datasource: GithubReleasesDatasource.id,
+    currentValue: version,
+    ...(isString(extractVersion) ? { extractVersion } : {}),
```
viceice:
```suggestion
    ...(extractVersion && { extractVersion }),
```
simplify

oschwald:
I pushed the requested changes.

viceice (approved):

viceice:
please resolve comments, so pr can automerge 😁

PR #40731, opened 2026-01-27
refactor: Migrate from ESLint to oxlint and biome

zharinov:
## Changes

Replaces ESLint with faster native linters:
- **oxlint** - Rust-based linter for JS/TS rules
- **biome** - Linter + import organizer

### Removed (~10 packages)
- `eslint` and all plugins (`@typescript-eslint/*`, `eslint-plugin-import-x`, `eslint-plugin-promise`, `@vitest/eslint-plugin`, `@containerbase/eslint-plugin`)
- `eslint.config.mjs`

### Added
- `.oxlintrc.json` with comprehensive rule configuration
- Custom JS plugin for `no-tools-import` and `test-root-describe` rules
- `biome.json` for `noUndeclaredDependencies`, `useNamingConvention`, `useConsistentObjectDefinitions`, and `organizeImports`
- CI step to check for uncommitted changes after biome fixes
- Pre-commit hook runs biome to organize imports

### Scripts
- `oxlint` / `oxlint-fix` / `oxlint-ci` - oxlint commands
- `biome` / `biome-fix` / `biome-ci` - biome commands

## Context

## AI assistance disclosure

- [x] Yes — substantive assistance


jamietanna:
Oof just a few changes 😅 

I'm gonna say let's hold off until after v43 (unless we need it for the ESM migration)?

zharinov:
Hey, I haven't done double-check yet, but feel free to comment anyways

zharinov:
> I'm gonna say let's hold off until after v43 (unless we need it for the ESM migration)?

Don't think we need for ESM.
We have time to think what we need from the new linters.
I think it's not necessarily the 1-to-1 migration, but a chance to review the checks we want.

docs/usage/assets/css/style.css
```diff
@@ -1,10 +1,10 @@
 .md-header .md-header__button.md-logo {
-  padding: 0;
+	padding: 0;
```
viceice, line 2: why this change, looks wrong to me

lib/expose.cjs
```diff
@@ -1,51 +1,52 @@
...
+// oxlint-disable-next-line import/no-dynamic-require -- intentional dynamic require
 const pkg = (() => require(filePath))();
 
 /**
  * return's re2
  * @returns {RegExpConstructor}
  */
 function re2() {
-  return require('re2');
+	return require("re2");
```
viceice, line 19: no indention change and single quote usage!

lib/instrumentation/utils.ts
```diff
@@ -20,5 +20,8 @@ export function massageThrowable(e: unknown): string | undefined {
   if (e instanceof Error) {
     return e.message;
   }
-  return String(e); // eslint-disable-line @typescript-eslint/no-base-to-string
+  if (typeof e === 'string') {
+    return e;
+  }
+  return JSON.stringify(e);
```
viceice, line 26: should be done separately

lib/logger/utils.ts
```diff
@@ -131,9 +131,10 @@ export default function prepareError(err: Error): Record<string, unknown> {
     return prepareZodError(err);
   }
 
-  const response: Record<string, unknown> = {
-    ...err,
-  };
+  const response: Record<string, unknown> = {};
+  for (const key of Object.keys(err)) {
```
viceice, line 135: separate PR

lib/config/options/index.spec.ts
RahulGautamSingh, line 1: why are these changes necessary?

zharinov: This file violates "no conditionals in tests", I just asked to reorganize it somehow.

lib/config/validation.spec.ts
```diff
@@ -1,9 +1,9 @@
 import type { HostRule } from '../types/index.ts';
```
viceice, line 1: needs fix 😁

tools/lint/rules.js
viceice, line 1: we should extract a package from it, like i've done for the eslint rules 😁

.oxlintrc.json
```diff
@@ -1,43 +1,213 @@
 {
   "$schema": "https://raw.githubusercontent.com/oxc-project/oxc/main/npm/oxlint/configuration_schema.json",
-  "plugins": ["import", "typescript"],
+  "plugins": ["import", "typescript", "promise", "vitest"],
+  "jsPlugins": ["./tools/lint/rules.js"],
   "categories": {},
   "rules": {
+    "constructor-super": "error",
```
viceice, line 7: isn't there some preset support? so we don't need to configure them all manually 🤔

viceice (commented):
quotes are wrongly changed from single to double 😕

marklai1998:
@zharinov 
biome also come with decent ts+js rules, any reason dual wielding oxlint and biome?

zharinov:
> biome also come with decent ts+js rules, any reason dual wielding oxlint and biome?

@marklai1998 We don't need just decent set of rules, we want to supersede all the existing eslint rules. If there is at least one rule covered by biome and not covered by oxlint, it's worth to have both. And we have such rule: it's `noParameterProperties`, which helps us to make sure Renovate could be run with Node without transpiling, or in the Bun runtime.

Another benefit is imports ordering which seems to be more stable with biome at the moment. I could be wrong on this one, though.

Anyways, compared to the `eslint` overhead, the 10 seconds to run another tool is nothing. This decision should be revised in the future as oxlint matures, but the current focus is having decent modern tooling which does everything eslint does, without requiring 8Gb memory and 3–5 minutes to finish.

connorshea:
FWIW Oxfmt may be able to be used as the formatter to replace Prettier (it should be nearly 100% compatible with the code output by Prettier 3.8) and also perform the import sorting? But I realize that may be too large of a change to include in this at the same time, and I haven't really checked your sorting config yet. https://oxc.rs/docs/guide/usage/formatter/sorting.html

As for the handful of missing rules: if you want to, you can potentially use them via JS Plugin aliases until they're ported to Oxlint (although you'd need to keep the extra ESLint dependencies around for it to work, which isn't ideal). See https://oxc.rs/docs/guide/usage/linter/js-plugins.html#plugin-aliases

PR #40736, opened 2026-01-27
refactor: Remove decorators for ESM compatibility

zharinov:
## Changes

Remove TypeScript decorators to unblock ESM migration. Decorators require transpilation which blocks native ESM.

- Migrate `@cache` decorator to `cached()` function across all datasources
- Remove `lib/util/cache/package/decorator.ts`
- Remove `lib/instrumentation/decorator.ts`
- Update `instrumented()` to return a wrapped function instead of executing immediately

## Context

- Ref: #9890 

## AI assistance disclosure


lib/modules/datasource/aws-rds/index.ts
```diff
@@ -40,4 +36,15 @@ export class AwsRdsDatasource extends Datasource {
         })),
     };
   }
+
+  getReleases(config: GetReleasesConfig): Promise<ReleaseResult | null> {
```
viceice, line 40: would be better to move this before the private function for less changes?

viceice (commented):
otherwise LGTM

viceice:
it's not required for esm, it's required to allow bare node on typescript sources. right?

we won't stop using a final compile step because we plan to do some bundling in future.

zharinov:
The TC39 approved standard for decorators is different from what we've been using in our code base. And it requires another royal pile of changes to support them. All my previous experience working with decorators in Renovate is screaming: no decorators — no problems.

zharinov:
But yes, transpiling is inevitable. It's nice to be able run the codebase locally without transpile, it gives very useful properties. But publishing typescript directly is discouraged by NPM even if modern Node allows running it.

viceice:
I agree, I just like the old decorations because they allow some syntactic sugar. 😅

I'm fine to drop them until there's a built-in replacement.

lib/modules/datasource/aws-rds/index.ts
viceice: ☝️

viceice (commented):
see open conversation

lib/modules/datasource/aws-rds/index.ts
zharinov: The indent levels are different, it won't result to less changes

viceice: it would without whitespace compare 😉

viceice (approved):

lib/modules/datasource/aws-rds/index.ts
zharinov: C'mon, it's 60 file changes, it takes a while even for AI agent

PR #40737, opened 2026-01-28
feat(renovate-config-validator): add --repository option

Zokormazo:
## Changes

Add --repository option for renovate-config-validator to simulate Renovate's repository configuration merging during validation. When specified, the validator applies the same config merging logic as runtime, using matching entries from the global repositories array.

## Context

This addresses a specific validation issue with `matchBaseBranches` rules when global configuration provides branch context. In some self-hosted setups, the global config contains a `repositories` array where each entry can specify `baseBranches` or `baseBranchPatterns` for that repository. When a repository config uses `matchBaseBranches` in its `packageRules`, the validator needs access to this global branch context to properly validate the rule. Without the `--repository` option, the validator only sees the repository config in isolation and incorrectly reports that `baseBranchPatterns` must be configured. With this option, the validator replicates Renovate's runtime behavior by merging the appropriate global repository entry with the repository config, providing the necessary branch context for validation.

## AI assistance disclosure

- [x] Yes — other (please describe):

This PR used Claude Code to assist identifying the runtime behaviour of renovate and the key function we could reuse on the validator to achieve the same merging logic

- [x] Tested with multiple config.js and renovate.json variants wih and without repositories block

### DIFF
--- a/lib/config-validator.ts
+++ b/lib/config-validator.ts
@@ -6,6 +6,7 @@ import { dequal } from 'dequal';
 import { pathExists, readFile } from 'fs-extra';
 import { getConfigFileNames } from './config/app-strings.ts';
 import { GlobalConfig } from './config/global.ts';
+import { filterConfig, mergeChildConfig } from './config/index.ts';
 import { massageConfig } from './config/massage.ts';
 import { migrateConfig } from './config/migration.ts';
 import type { RenovateConfig } from './config/types.ts';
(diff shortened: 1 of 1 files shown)


RahulGautamSingh (commented):
`baseBranchPatterns` can be in the repo config too.

If we want to add global context to the repo config for validation, this is not the correct approach. We should initialize the `GlobalConfig` class instead(not thought out)

Please, start a discussion we should first decide on the approach and then you can repo-open this PR.

PR #40756, opened 2026-01-28
feat!: Switch to ESM modules

zharinov:
## Changes

Enable native ESM syntax in source code while maintaining backward compatibility for all consumers.

**Core changes:**
- Set `package.json` type to `"module"` for ESM semantics in TypeScript
- Configure dual CJS/ESM build output (`dist/cjs/`, `dist/esm/`)
- Add `exports` field for transparent consumer path resolution
- Add `with { type: 'json' }` to JSON imports (required by ESM)
- Replace `__dirname` with `import.meta.dirname` in test files

**Backward compatibility:**
- Add CJS fallback in config loader using `createRequire()`
- Users' existing `config.js` files keep working with either ESM or CJS syntax
- Consumer import paths unchanged: `renovate/dist/foo/bar` resolves to correct format

## Context

## AI assistance disclosure


lib/modules/datasource/hex/v2/package.ts
```diff
@@ -6,7 +6,7 @@
 
 // istanbul ignore file
 /* eslint-disable */
-import * as _m0 from 'protobufjs/minimal';
+import * as _m0 from 'protobufjs/minimal.js';
```
viceice, line 9: did you do this manually? then we need to document, because it's a generated file.

lib/modules/datasource/hex/v2/signed.ts
```diff
@@ -6,7 +6,7 @@
 
 // istanbul ignore file
 /* eslint-disable */
-import * as _m0 from 'protobufjs/minimal';
+import * as _m0 from 'protobufjs/minimal.js';
```
viceice, line 9: sam as above

lib/modules/manager/git-submodules/extract.ts
```diff
@@ -1,5 +1,5 @@
 import type { SimpleGit } from 'simple-git';
-import Git from 'simple-git';
```
viceice: wonder why this is needed. the esm has a default export 🤔

viceice (commented):
first round

package.json
```diff
@@ -2,13 +2,25 @@
...
+  },
+  "exports": {
+    ".": {
+      "import": "./dist/esm/renovate.js",
+      "require": "./dist/cjs/renovate.js",
+      "types": "./dist/esm/renovate.d.ts"
+    },
+    "./dist/*": {
+      "import": "./dist/esm/*",
+      "require": "./dist/cjs/*"
```
viceice: also needs types?

tools/validate-schema.ts
```diff
@@ -1,10 +1,12 @@
 import { promises } from 'fs';
 import type { ErrorObject, ValidateFunction } from 'ajv';
 import { Ajv } from 'ajv';
-import draft7MetaSchema from 'ajv/lib/refs/json-schema-draft-07.json';
-import addFormats from 'ajv-formats';
+import draft7MetaSchema from 'ajv/lib/refs/json-schema-draft-07.json' with { type: 'json' };
+import _addFormats from 'ajv-formats';
 import { glob } from 'glob';
 
+const addFormats = _addFormats as unknown as typeof _addFormats.default;
```
viceice, line 7: wrong types?

lib/util/git/index.spec.ts
```diff
@@ -1,6 +1,6 @@
 import fs from 'fs-extra';
 import type { PushResult } from 'simple-git';
-import Git from 'simple-git';
+import { simpleGit } from 'simple-git';
```
viceice, line 3: can't we do
```suggestion
import { simpleGit as Git } from 'simple-git';
```
to have less changes? or will it result in another lint error?

lib/expose.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+export function openpgp(): typeof import('openpgp') {
+  return require('openpgp');
+}
+
+/**
+ * return's sqlite
+ */
+export function sqlite(): typeof import('better-sqlite3') {
+  return require('better-sqlite3');
```
viceice: we can just use dynamic import now

lib/config-validator.ts
```diff
@@ -1,16 +1,18 @@
 #!/usr/bin/env node
 import { Command, CommanderError } from 'commander';
-import 'source-map-support/register';
+import 'source-map-support/register.js';
 import './punycode.cjs';
 import { dequal } from 'dequal';
-import { pathExists, readFile } from 'fs-extra';
+import fs from 'fs-extra';
+
+const { pathExists, readFile } = fs;
```
viceice: move below imports

lib/util/git/index.spec.ts
```diff
@@ -1258,7 +1258,7 @@ describe('util/git/index', { timeout: 10000 }, () => {
     });
   });
 
-  describe('syncGit()', () => {
+  describe('syncsimpleGit()', () => {
```
viceice:
```suggestion
  describe('syncGit()', () => {
```

lib/util/git/index.spec.ts
```diff
@@ -1376,7 +1376,7 @@ describe('util/git/index', { timeout: 10000 }, () => {
       });
     });
 
-    describe('syncGit()', () => {
+    describe('syncsimpleGit()', () => {
```
viceice:
```suggestion
    describe('syncGit()', () => {
```

lib/workers/repository/process/vulnerabilities.ts
```diff
@@ -8,7 +8,9 @@ import {
   isTruthy,
 } from '@sindresorhus/is';
 import type { CvssVector } from 'ae-cvss-calculator';
-import { fromVector } from 'ae-cvss-calculator';
+import * as _aeCvss from 'ae-cvss-calculator';
+
+const { fromVector } = _aeCvss as unknown as typeof _aeCvss.default;
```
viceice: move below imports

viceice (commented):
looks great

lib/modules/datasource/hex/v2/package.ts
zharinov: It's not actually that important:
1. Generated once in a lifetime
2. Not being used anywhere in the code really (we merged, but didn't finish the original work)

tools/validate-schema.ts
zharinov: Default CJS exports 🤷‍♂️

lib/modules/datasource/hex/v2/package.ts
viceice: so candidate for removal in a follow up 😁

lib/util/cache/package/sqlite.ts
```diff
@@ -1,9 +1,11 @@
 import { promisify } from 'node:util';
 import zlib, { constants } from 'node:zlib';
 import type { Database, Statement } from 'better-sqlite3';
-import { exists } from 'fs-extra';
+import fs from 'fs-extra';
+
+const { exists } = fs;
```
viceice: move below imports

lib/expose.ts
```diff
@@ -0,0 +1,32 @@
+// need to use dynamic path so that typescript does not include package.json in dist folder after compilation
+const filePath = new URL('../package.json', import.meta.url).pathname;
+export const pkg = (await import(filePath, { with: { type: 'json' } }))
```
viceice: we should not use toplevel await for now, as this really breaks cjs consumers. i only ment to dynamic import openpgp and sqlite, because the callers are already in async scope

lib/expose.ts
```diff
@@ -0,0 +1,32 @@
+// need to use dynamic path so that typescript does not include package.json in dist folder after compilation
```
viceice: is this still an issue with tsdown?

tools/validate-schema.ts
viceice: we can fixup later

viceice (commented):
nearly done

viceice (dismissed):

viceice:
needs prettier fix

viceice (approved):

jamietanna (commented):
Should this not be against `next-major`?

viceice (approved):

PR #40798, opened 2026-01-29
feat(elm): add elm package manager support

Flo0807:
## Changes

depends on #41000

  - Adds Elm Manager (`lib/modules/manager/elm/`): Extracts dependencies from `elm.json` files (both application and package types)

## Context

see https://github.com/renovatebot/renovate/discussions/24739

## AI assistance disclosure

The public repository: https://github.com/rtfeldman/elm-spa-example (note that I performed a dry run only)


viceice (changes requested):
please split PR into three parts.

this PR can stay as is.

first new PR : versioning 
second new PR: datasource 

and finally rebase this so it only includes the manager 


inline fixtures if they are that small

Flo0807:
> please split PR into three parts.
> 
> this PR can stay as is.
> 
> first new PR : versioning second new PR: datasource
> 
> and finally rebase this so it only includes the manager
> 
> inline fixtures if they are that small

done: #40999 and #41000 

 > inline fixtures if they are that small

Which fixtures are you referring to? We have fixtures for the manager implementation as well as for versioning (now in #40999) @viceice

viceice:
>  > inline fixtures if they are that small
> 
> Which fixtures are you referring to? We have fixtures for the manager implementation as well as for versioning (now in #40999) @viceice 

all

Flo0807:
> > > inline fixtures if they are that small
> > 
> > 
> > Which fixtures are you referring to? We have fixtures for the manager implementation as well as for versioning (now in #40999) @viceice
> 
> all

Check, done.

lib/modules/manager/elm/extract.spec.ts
```diff
@@ -0,0 +1,267 @@
+import { codeBlock } from 'common-tags';
+import { extractPackageFile } from './index.ts';
+
+const applicationJson = codeBlock`
```
viceice: move to test if only used once

lib/modules/manager/elm/extract.ts
```diff
@@ -0,0 +1,125 @@
...
+  };
+}
+
+export function extractPackageFile(
+  content: string,
+  packageFile: string,
+): PackageFileContent | null {
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(content);
```
viceice: use our `Json.pipe(ElmSchema)` util

lib/modules/manager/elm/schema.ts
```diff
@@ -0,0 +1,52 @@
...
+  version: z.string().optional(),
+  'elm-version': z.string(),
+  dependencies: z.record(z.string(), z.string()).optional().default({}),
+  'test-dependencies': z.record(z.string(), z.string()).optional().default({}),
+});
+
+/**
+ * Union type for any valid elm.json
+ */
+export const ElmJson = z.discriminatedUnion('type', [
```
viceice:
```suggestion
export const ElmJson = Json.pipe(z.discriminatedUnion('type', [
```
see other comment

Flo0807:
Hey @viceice, would you mind giving this PR a final review? It's the last missing piece for Elm support in Renovate, and I'd really appreciate getting it merged soon.

Tranquility:
I would love use this. please have another look @viceice

lib/modules/manager/elm/index.ts
```diff
@@ -0,0 +1,18 @@
+import type { Category } from '../../../constants/index.ts';
+import { ElmPackageDatasource } from '../../datasource/elm-package/index.ts';
+import { GithubTagsDatasource } from '../../datasource/github-tags/index.ts';
+
+export { extractPackageFile } from './extract.ts';
+
+export const displayName = 'Elm';
+export const url = 'https://elm-lang.org';
+export const categories: Category[] = ['js'];
```
RahulGautamSingh: Elm is its own language, not JS, right?

Should we create a new category instead? @viceice @secustor

lib/modules/manager/elm/readme.md
```diff
@@ -0,0 +1,22 @@
...
+For **packages**:
+
+- `dependencies`
+- `test-dependencies`
+
+Additionally, the `elm-version` field is extracted as `depType: elm-version` using the `github-tags` datasource for `elm/compiler`.
+
+### Version formats
+
+Applications use exact versions (e.g., `1.0.0`), while packages use Elm's range constraint format (e.g., `1.0.0 <= v < 2.0.0`).
```
RahulGautamSingh:
```suggestion
Applications use exact versions (e.g., `1.0.0`), while packages use the Elm range constraint format (e.g., `1.0.0 <= v < 2.0.0`).
```

we prefer not to use possesives

RahulGautamSingh (commented):
LGTM otherwise

lib/modules/manager/elm/index.ts
viceice: I don't know elm 😇

RahulGautamSingh: Let's create then 😃 @Flo0807

Flo0807: > Elm is its own language, not JS, right?

It is its own language but compiles to JavaScript. I am going to create a dedicated category.

lib/constants/category.ts
```diff
@@ -9,6 +9,7 @@ export const Categories = [
   'dart',
   'docker',
   'dotnet',
+  'elm',
```
RahulGautamSingh: elm will come after elixir

tools/docs/manager.ts
```diff
@@ -46,6 +46,7 @@ export const CategoryNames: Record<Category, string> = {
   dart: 'Dart',
   docker: 'Docker',
   dotnet: '.NET',
+  elm: 'Elm',
```
RahulGautamSingh: elm after elixis

RahulGautamSingh (commented):
One design note:

Elm applications pin both direct and indirect deps at exact versions in `elm.json`. 

Your approach for the manager bumps a single dep's string in place with no `updateArtifacts` step to re-run elm and re-solve. Bumping one direct dep without updating its indirects can produce an `elm.json` that the Elm compiler rejects until the user re-solves. 

That's a real limitation of the incremental approach which is likely acceptable for initial support, but worth confirming with the maintainers and mentioning it in the readme.

cc @jamietanna @viceice @secustor

Flo0807:
> One design note:
> 
> Elm applications pin both direct and indirect deps at exact versions in `elm.json`.
> 
> Your approach for the manager bumps a single dep's string in place with no `updateArtifacts` step to re-run elm and re-solve. Bumping one direct dep without updating its indirects can produce an `elm.json` that the Elm compiler rejects until the user re-solves.
> 
> That's a real limitation of the incremental approach which is likely acceptable for initial support, but worth confirming with the maintainers and mentioning it in the readme.
> 
> cc @jamietanna @viceice @secustor

Good catch! Sadly, Elm always installs the latest compatible version, and there is no `elm install pkg@1.2.3`. The de-facto standard for dependency management in the Elm community is [zwilias/elm-json](https://github.com/zwilias/elm-json). It allows pinning versions exactly and rewriting direct and indirect deps: `elm-json install --yes elm/http@2.0.0`.

I guess having `elm-json` available here is not straightforward, as we have to add it to containerbase first. Its latest commit is 4 years old, so it's probably difficult to argue about...

At least, we should not treat indirect dependencies as *normal* dependencies that cause upgrade PRs (see https://github.com/renovatebot/renovate/pull/40798/changes#diff-d426918eae7516c1e5d49b83364375b75f63b260b6f1074a76d8ad2df86f1445R30).

In addition, we can improve the current implementation by preventing invalid version bumps and being more defensive. We can't recompute the graph without a solver, but we can determine whether the unchanged indirect graph is still valid. Elm packages publish metadata at `https://package.elm-lang.org/packages/{name}/{version}/elm.json` we can use to verify. If a direct bump requires no graph changes, Renovate accepts it. If a dependency needs to be added, removed, moved, or upgraded, Renovate reports an artifact error instead of silently producing an invalid `elm.json`. 

If we want to support graph-changing updates in the future without relying on `elm-json` in containerbase, we could also build a TypeScript solver for this:
- Fetch available versions from `releases.json`
- Fetch each candidate’s `elm.json`
- Accumulate version and compiler constraints
- Use backtracking to choose a compatible graph
- Prefer currently selected indirect versions, then newest compatible versions
- Reconstruct regular and test indirect sections

But I guess it is better to not have graph-changing updates than to maintain such solver code in Renovate.

secustor:
We should never produce invalid artifacts. 
I would rather not add the manager to the index so it is not being called. 

Full graph resolution is not something we usually do, so no need for `elm-json`

PR #40812, opened 2026-01-30
feat(replacements): Update grafana-oss replacement

MalteJoe:
## Changes

Updates the grafana replacement rule to a newer version (`12.3.2` instead of `12.1.1`) pushed to the [deprecated repository](https://hub.docker.com/r/grafana/grafana-oss/tags).

This is a follow up to #39512 and #40293, updating to a more recent version, in case users of the deprecated repository have already upgraded to a newer version. I also added the `-ubuntu` variant so that may be replaced too. I checked the digests for both image versions to match, so even digest-pinned versions can be safely replaced.

To avoid creating more PRs, we could leave this PR open until a `12.4.0` release on the [main repository](https://hub.docker.com/r/grafana/grafana/tags) exists, as that [should be the first version not being released on the deprecated repository](https://grafana.com/docs/grafana/v12.3/setup-grafana/installation/docker/) and update this PR to the latest `12.3.x` version.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/data/replacements.json
+++ b/lib/data/replacements.json
@@ -714,7 +714,13 @@
       {
         "matchDatasources": ["docker"],
         "matchPackageNames": ["/^(?:docker\\.io/)?grafana/grafana-oss$/"],
-        "matchCurrentVersion": "12.1.1",
+        "matchCurrentVersion": "12.3.2",
+        "replacementName": "grafana/grafana"
+      },
+      {
+        "matchDatasources": ["docker"],
+        "matchPackageNames": ["/^(?:docker\\.io/)?grafana/grafana-oss$/"],
+        "matchCurrentVersion": "12.3.2-ubuntu",
         "replacementName": "grafana/grafana"
       }
     ]


lib/data/replacements.json
```diff
@@ -714,7 +714,13 @@
       {
         "matchDatasources": ["docker"],
         "matchPackageNames": ["/^(?:docker\\.io/)?grafana/grafana-oss$/"],
-        "matchCurrentVersion": "12.1.1",
+        "matchCurrentVersion": "12.3.2",
```
viceice, line 717:
```suggestion
        "matchCurrentVersion": ">=12.1.1-0",
```

maybe just use this? rneovate will fallback to semver if this isn't parseable with docker versioning.

it should also match `12.1.1-ubuntu` and should retain current value

MalteJoe: We tried that in https://github.com/renovatebot/renovate/pull/39512#discussion_r2567539968, but Dockerfiles starting with `FROM grafana/grafana-oss:12.1.1@sha256:a1701c2180249361737a99a01bc770db39381640e4d631825d38ff4535efa47d` or `FROM grafana/grafana-oss:12.1.1` weren't replaced at all. 

While it looks like Renovate tries to use SemVer for upgrading dependencies, the [docs for the docker versioning explicitly state that there is no support for ranges](https://docs.renovatebot.com/modules/versioning/docker/#rangesconstraints), which looks like what we observed, because no MRs were created for the examples above, but only after https://github.com/renovatebot/renovate/pull/40293 was merged.

Also I tested running Renovate 43.2.0 (without these changes) with `FROM grafana/grafana-oss:12.1.1-ubuntu` which didn't create a replacement PR either, so having rules for both variants still seems necessary.

viceice:
```suggestion
        "matchCurrentVersion": "/^12\\./",
```
why about using a regex?

MalteJoe: I really like the idea of using a regex here.

The annoying thing about this is [that the Grafana Team announced](https://grafana.com/docs/grafana/latest/setup-grafana/installation/docker/) that the repository wouldn't be updated anymore starting with `12.4.0` but they still released up to `12.4.3` and even the new major version `13.0.1` (which means Renovate may have upgraded users to that version already)

If we go with the Regex approach and adjust it to something like `/^1[23]\\./`, I see these risks:
* If the versions in the repositories matching that regex are not in sync, Renovate might try to replace to a version that does not exist
* If images are pinned by digest, the digest is not updated by the replacement rule. The images of the two repositories are **mostly** in sync, but not completely (see #40293) which could result in broken replacements. This might be fine though, as Renovate will start updating the digests with the new repository then.

PR #40840, opened 2026-02-02
fix(extract-update): return done instead of undefined for non-onboarded repos

ZackKanter:
## Changes

For non-onboarded repositories, `update()` was returning undefined which propagated to `processResult()` and caused logs to show `"Repository result: undefined"`. Now returns `done` which is the semantically correct `RepositoryResult` value.

## Context

## AI assistance disclosure

### DIFF
--- a/lib/workers/repository/index.ts
+++ b/lib/workers/repository/index.ts
@@ -170,8 +170,7 @@ export async function renovateRepository(
         );
       }
       await finalizeRepo(config, branchList, repoConfig);
-      // TODO #22198
-      repoResult = processResult(config, res!);
+      repoResult = processResult(config, res ?? 'done');
     }
     printRepositoryProblems(config.repository);
   } catch (err) /* istanbul ignore next */ {
--- a/lib/workers/repository/process/extract-update.spec.ts
+++ b/lib/workers/repository/process/extract-update.spec.ts
@@ -161,12 +161,12 @@ describe('workers/repository/process/extract-update', () => {
   });
 
   describe('update()', () => {
-    it('returns done for non-onboarded repos', async () => {
+    it('returns undefined for non-onboarded repos', async () => {
       const config = {
         repoIsOnboarded: false,
       };
       const res = await update(config, []);
-      expect(res).toBe('done');
+      expect(res).toBeUndefined();
     });
   });
 
--- a/lib/workers/repository/process/extract-update.spec.ts
+++ b/lib/workers/repository/process/extract-update.spec.ts
@@ -160,6 +160,16 @@ describe('workers/repository/process/extract-update', () => {
     });
   });
 
+  describe('update()', () => {
+    it('returns done for non-onboarded repos', async () => {
+      const config = {
+        repoIsOnboarded: false,
+      };
+      const res = await update(config, []);
+      expect(res).toBe('done');
+    });
+  });
+
   describe('isCacheExtractValid()', () => {
     let cachedExtract: BaseBranchCache;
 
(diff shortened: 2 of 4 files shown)


viceice (commented):
I'm not sure if this is the correct value

ZackKanter:
> I'm not sure if this is the correct value

I'm not 100% sure either. Do you agree that the current behavior seems incorrect/suboptimal? If so, is there someone who can weigh in on the desired value to return here?

jamietanna:
I'm OOO tomorrow, but I'll take a look on Wednesday/Thursday

ZackKanter:
> I'm OOO tomorrow, but I'll take a look on Wednesday/Thursday

Just checking back on this. Appreciate you taking a look.

jamietanna (changes requested):
IMO, it may be better to bring in one of the values from `ProcessStatus` or `RepositoryResult` here 

Given [we use `done`](https://github.com/renovatebot/renovate/blob/79f2f4197ba685ded1937df029859601e1008def/lib/workers/repository/result.ts#L55-L56) "successfully processed" I'd say that this should either be `onboarding` (as it's not yet onboarded) or some other relevant option

ZackKanter:
> IMO, it may be better to bring in one of the values from `ProcessStatus` or `RepositoryResult` here
> 
> Given [we use `done`](https://github.com/renovatebot/renovate/blob/79f2f4197ba685ded1937df029859601e1008def/lib/workers/repository/result.ts#L55-L56) "successfully processed" I'd say that this should either be `onboarding` (as it's not yet onboarded) or some other relevant option

Thanks for the feedback @jamietanna – you're right that `'done'` from `update()` is misleading for a repo that hasn't onboarded yet.

Re: the proposal to add `'onboarding'` to `RepositoryResult` – I'd push back on that a bit. Right now these two types seem to have a clear separation of concerns: `RepositoryResult` is an outcome ("what happened"), `ProcessStatus` is a state ("what kind of repo is this"). Adding `'onboarding'` to `RepositoryResult` would seem to blur that line, since it's not really an outcome of processing – it's a property of the repo. And `processResult()` already derives `status: 'onboarding'` from `config.repoIsOnboarded === false`, so we'd end up encoding the same information in two places. I don't have a huge amount of familiarity with the inner workings of Renovate, so this may be wrong.

After digging into this more, I think the bug is actually at the **call site** in `index.ts`, not in `update()` itself. `update()` returning `undefined` when there's nothing to do is accurate – it genuinely didn't write any updates. The problem seems to be that `processResult(config, res!)` blindly asserts `res` is non-null, so `undefined` leaks into the log output.

I've reworked the fix to keep `update()` unchanged and instead handle the `undefined` where the type conversion happens:

```typescript
// before (unsafe assertion, produced "Repository result: undefined")
repoResult = processResult(config, res!);

// after (null coalescing, falls back to 'done' when res is undefined)
repoResult = processResult(config, res ?? 'done');
```

`'done'` here is in the context of `RepositoryResult` – "the repository was processed without error" – which is true even for non-onboarded repos. The `onboarding` semantics are already handled by `processResult()`, which checks `config.repoIsOnboarded === false` and derives `status: 'onboarding'` from that. So the log output becomes:

```
Repository result: done, status: onboarding, enabled: true, onboarded: false
```

This keeps the separation clean: `RepositoryResult` describes what happened during processing (`done`, `automerged`, or an error), and `ProcessStatus` describes the repo's state (`onboarding`, `onboarded`, `activated`, etc.).

I also reverted `updateRepo()`'s return type back to `Promise<WriteUpdateResult | undefined>` and updated the test to assert `undefined` for non-onboarded repos.

PR #40867, opened 2026-02-03
fix(nix): use alias to refer to node that will be updated

tebriel:
## Changes

Extracts the aliases into a set and if a node is a member of the set, then it is a root input node, otherwise ignore it as it is a secondary input/dependency.

## Context

Discussion: https://github.com/renovatebot/renovate/discussions/40866

## AI assistance disclosure

Copilot attempted to autocomplete my lines but suggestions were not accepted as they were generally wrong. It _was_ running, so disclosing that.

The public repository: https://github.com/tebriel/renovate-repro

### DIFF
--- a/lib/modules/manager/nix/extract.ts
+++ b/lib/modules/manager/nix/extract.ts
@@ -1,4 +1,3 @@
-import { InputSerialization$ } from '@aws-sdk/client-s3';
 import { logger } from '../../../logger/index.ts';
 import { getSiblingFileName, readLocalFile } from '../../../util/fs/index.ts';
 import { regEx } from '../../../util/regex.ts';
--- a/lib/modules/manager/nix/extract.ts
+++ b/lib/modules/manager/nix/extract.ts
@@ -9,7 +9,7 @@ import type {
   PackageDependency,
   PackageFileContent,
 } from '../types.ts';
-import { NixFlakeLock, type NixInput } from './schema.ts';
+import { NixFlakeLock } from './schema.ts';
 
 // as documented upstream
 // https://github.com/NixOS/nix/blob/master/doc/manual/source/protocols/tarball-fetcher.md#gitea-and-forgejo-support
@@ -57,22 +57,21 @@ export async function extractPackageFile(
   // update under nodes
   const rootAliases = new Set<string>();
   for (const input in rootInputs) {
-    rootAliases.add(rootInputs[input] as string);
+    rootAliases.add(rootInputs[input]);
   }
 
-  for (const [depName, flakeInput] of Object.entries(flakeLock.nodes) as [
-    string,
-    NixInput,
-  ][]) {
+  Object.keys(flakeLock.nodes).forEach((depName) => {
     // the root input is a magic string for the entrypoint and only references other flake inputs
     if (depName === 'root') {
-      continue;
+      return;
     }
 
     // skip all locked and transitivie nodes as they cannot be updated by regular means
     if (!rootAliases.has(depName)) {
-      continue;
+      return;
     }
+
+    const flakeInput = flakeLock.nodes[depName];
 
     // flakeLocked example: { rev: '56a49ffef2908dad1e9a8adef1f18802bc760962', type: 'github' }
     const flakeLocked = flakeInput.locked;
@@ -84,7 +83,7 @@ export async function extractPackageFile(
         { flakeLockFile, flakeInput },
         `input is missing locked, skipping`,
       );
-      continue;
+      return;
     }
 
     if (flakeOriginal === undefined) {
@@ -92,7 +91,7 @@ export async function extractPackageFile(
         { flakeLockFile, flakeInput },
         `input is missing original, skipping`,
       );
-      continue;
+      return;
     }
 
     // indirect inputs cannot be reliably updated because they depend on the flake registry
@@ -101,7 +100,7 @@ export async function extractPackageFile(
         { flakeLockFile, flakeInput },
         `input is type indirect, skipping`,
       );
-      continue;
+      return;
     }
 
     // cannot update local path inputs
@@ -110,7 +109,7 @@ export async function extractPackageFile(
         { flakeLockFile, flakeInput },
         `input is type path, skipping`,
       );
-      continue;
+      return;
     }
 
     // if no rev is being tracked, we cannot update this input
@@ -119,7 +118,7 @@ export async function extractPackageFile(
         { flakeLockFile, flakeInput },
         `locked input is not tracking a rev, skipping`,
       );
-      continue;
+      return;
     }
 
     // if there's a new digest, set the corresponding digest in the lockfile so confirmations pass
@@ -206,7 +205,7 @@ export async function extractPackageFile(
     }
 
     deps.push(dep);
-  }
+  });
 
   if (deps.length) {
     return { deps };
--- a/lib/modules/manager/nix/extract.ts
+++ b/lib/modules/manager/nix/extract.ts
@@ -55,11 +55,10 @@ export async function extractPackageFile(
 
   // collect all root input aliases, these are a map of input name -> alias, the alias is the item we want to
   // update under nodes
-  const aliasVals = [];
+  const rootAliases = new Set<string>();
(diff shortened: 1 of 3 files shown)


tebriel:
I do not have an edit on my PR title so I cannot update to match semantic requirements.

image

lib/modules/manager/nix/extract.ts
```diff
@@ -1,3 +1,4 @@
+import { InputSerialization$ } from '@aws-sdk/client-s3';
```
viceice:
```suggestion
```

tebriel: wow ok maybe copilot injected this while I wasn't looking, I should probably just turn that thing off. Thanks

viceice (commented):
needs more testing, missing code coverage

tebriel:
@viceice understood, will find uncovered lines and add cases.

tebriel:
@viceice cleaned that up. Hopefully my reading of the typescript (and zod) documentation led to correct implementation of the updated types. Zod4 would've maybe been easier (we're on zod3 here) with [looserecord](https://zod.dev/api?id=zlooserecord). Hopefully the `as` usage here is appropriate (as we know the exact type of the `root` object).

lib/modules/manager/nix/extract.ts
```diff
@@ -9,7 +9,7 @@ import type {
   PackageDependency,
   PackageFileContent,
 } from '../types.ts';
-import { NixFlakeLock } from './schema.ts';
+import { NixFlakeLock, type NixInput } from './schema.ts';
```
RahulGautamSingh: type import should be in a new line unless, lint is complaining

lib/modules/manager/nix/extract.ts
```diff
@@ -53,14 +53,24 @@ export async function extractPackageFile(
     return null;
   }
 
-  for (const [depName, flakeInput] of Object.entries(flakeLock.nodes)) {
+  // collect all root input aliases, these are a map of input name -> alias, the alias is the item we want to
+  // update under nodes
+  const rootAliases = new Set<string>();
+  for (const input in rootInputs) {
+    rootAliases.add(rootInputs[input] as string);
```
RahulGautamSingh: use an if check with `isNonEmptyString`, we try not to use `as`

lib/modules/manager/nix/extract.ts
```diff
@@ -53,14 +53,24 @@ export async function extractPackageFile(
...
+  }
+
+  for (const [depName, flakeInput] of Object.entries(flakeLock.nodes) as [
+    string,
+    NixInput,
+  ][]) {
```
RahulGautamSingh: same about the `as` assertion here.

if we have parsed it this type will come from the schema..you can update the schema

RahulGautamSingh (changes requested):

lib/modules/manager/nix/extract.ts
tebriel: You can see the partial record is OR'd with the full record which should mean, if I understand, that since the value wasn't root it should choose the other one (but obviously that's not what's happening). I can add safeguards and phony tests that generate data we should never have if that's what you'd like.

Image

lib/modules/manager/nix/extract.ts
tebriel: I did update the schema and my type hints with the LSP were still wonky. Can you review my schema changes?

Image

as an example, note that rootInputs is the `root` from the union type, but using zod this way maybe is too much for TypeScript's type system?

Image

lib/modules/manager/nix/schema.ts
```diff
@@ -29,15 +29,26 @@ const OriginalInput = z.object({
...
-    nodes: z.record(z.string(), NixInput),
+    nodes: z.union([
+      z.record(
+        z.literal('root'),
+        z.object({
+          inputs: z.record(z.string(), z.string()),
+        }),
+      ),
+      z.record(z.string(), NixInput),
+    ]),
```
tebriel:
```typescript
// https://github.com/NixOS/nix/blob/f7fc24c/src/libflake/include/nix/flake/lockfile.hh
export const NixFlakeLock = Json.pipe(
  z.object({
    nodes: z
      .object({
        root: z
          .object({
            inputs: z.record(z.string(), z.string()),
          })
          .optional(),
      })
      .catchall(NixInput),
    version: z.literal(7),
  }),
);
```

this is an alternative but the type system still can't figure out that when we bail out if the key is `root` then it's a `NixInput` type.

lib/modules/manager/nix/extract.ts
```diff
@@ -53,17 +53,26 @@ export async function extractPackageFile(
...
 
-  for (const [depName, flakeInput] of Object.entries(flakeLock.nodes)) {
+  // collect all root input aliases, these are a map of input name -> alias, the alias is the item we want to
+  // update under nodes
+  const rootAliases = new Set<string>();
+  for (const input in rootInputs) {
+    rootAliases.add(rootInputs[input]);
+  }
+
+  Object.keys(flakeLock.nodes).forEach((depName) => {
```
tebriel, line 63: this allows the typing system to detect when we're using a node with the key of `root` or not.

lib/modules/manager/nix/extract.ts
```diff
@@ -53,17 +53,26 @@ export async function extractPackageFile(
...
     }
 
     // skip all locked and transitivie nodes as they cannot be updated by regular means
-    if (!(depName in rootInputs)) {
-      continue;
+    if (!rootAliases.has(depName)) {
+      return;
     }
 
+    const flakeInput = flakeLock.nodes[depName];
```
tebriel, line 74: because we bailed on 66 when `depName` was root TS can now safely assume `flakeInput` is a `NixInput`

lib/modules/manager/nix/schema.ts
```diff
@@ -29,15 +29,26 @@ const OriginalInput = z.object({
...
-    nodes: z.record(z.string(), NixInput),
+    nodes: z
+      .object({
+        root: z
+          .object({
+            inputs: z.record(z.string(), z.string()).optional(),
+          })
+          .optional(),
+      })
+      .catchall(NixInput),
```
tebriel, line 51: [catchall()](https://v3.zod.dev/?id=catchall) allows all keys but makes sure they're all in the format of `NixInput`, except our `root` node, of course which we previously defined.

lib/modules/manager/nix/extract.ts
viceice: use for-of loop

viceice: https://github.com/renovatebot/renovate/blob/main/docs/development/best-practices.md#best-practices

lib/modules/manager/nix/schema.ts
```diff
@@ -29,15 +29,26 @@ const OriginalInput = z.object({
   url: z.string().optional(),
 });
 
-const NixInput = z.object({
+export const NixInput = z.object({
```
viceice, line 32:
```suggestion
const NixInput = z.object({
```
not externally used

lib/modules/manager/nix/schema.ts
```diff
@@ -29,15 +29,26 @@ const OriginalInput = z.object({
...
   inputs: z.record(z.string(), z.string().or(z.array(z.string()))).optional(),
   locked: LockedInput.optional(),
   original: OriginalInput.optional(),
 });
 
+export type NixInput = z.infer<typeof NixInput>;
+
```
viceice, line 39:
```suggestion
```
unused

lib/modules/manager/nix/schema.ts
viceice: you can use zod 4 , zod v3 is exporting it from a subdir

viceice: you can use zod 4 , zod v3 is exporting it from a subdir

viceice:
https://zod.dev/library-authors

there it shows the sub exports

RahulGautamSingh (commented):
LGTM except open comments

PR #40924, opened 2026-02-04
chore(docker): add type information for `created` of OCI image config

jamietanna:
## Changes

As a step towards #38656, we can discover the Docker image config blob's
`created` timestamp, which will correspond with the `releaseTimestamp.`

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/datasource/docker/schema.ts
+++ b/lib/modules/datasource/docker/schema.ts
@@ -41,6 +41,7 @@ export const OciImageConfig = z.object({
   // This is required by the spec, but probably not present in the wild.
   architecture: z.string().nullish(),
   config: z.object({ Labels: z.record(z.string()).nullish() }).nullish(),
+  created: z.string(),
 });
 export type OciImageConfig = z.infer<typeof OciImageConfig>;
 


jamietanna:
Detecting this is straightforward, but in terms of trying to make sure that each tag/digest has a `releaseTimestamp`, that's more awkward

From my hacky testing, we need to be careful with rate limits that may occur, as we discover all the tags in a paginated API (that returns only tags) but then looking up the `created` will result in n*API calls

lib/modules/datasource/docker/schema.ts
```diff
@@ -41,6 +41,7 @@ export const OciImageConfig = z.object({
   // This is required by the spec, but probably not present in the wild.
   architecture: z.string().nullish(),
   config: z.object({ Labels: z.record(z.string()).nullish() }).nullish(),
+  created: z.string(),
```
viceice, line 44:
```suggestion
  created: z.string().nullish(),
```
I think it's sometimes missing/ not required field

viceice:
> Detecting this is straightforward, but in terms of trying to make sure that each tag/digest has a `releaseTimestamp`, that's more awkward
> 
> From my hacky testing, we need to be careful with rate limits that may occur, as we discover all the tags in a paginated API (that returns only tags) but then looking up the `created` will result in n*API calls

yes, that's why we haven't done it yet.
we can't cache tag-timestamp relation, because it can change. we can only cache digest to timestamp for long time.

digest can usually simply fetched by head request which doesn't count to rate limit.

lib/modules/datasource/docker/schema.ts
RahulGautamSingh: I think it has a fallback but better to be safe.

PR #40965, opened 2026-02-05
refactor(lint): Fix lint violations

zharinov:
## Changes

Fix underlying lint violations so ESLint passes without disable comments. Remaining suppressions (where the code is correct but lint disagrees) are left for a follow-up.

### Comment-only removals
- Remove `/* eslint-disable */` from `.markdownlint-cli2.mjs` (ESLint doesn't lint this file)

### Code fixes
- **Floating promises**: `void` prefix instead of `eslint-disable` (`lib/renovate.ts`, `tools/generate-schema.ts`)
- **Unnecessary `async`**: Remove from `helmfile/utils.ts` (already returns `Promise` via `localPathExists()`)
- **Test async patterns**: Remove async wrapper in `presets/local/index.spec.ts` (throws synchronously), replace `async` arrow with `Promise.resolve()` in `branch/index.spec.ts`
- **`throw lastError`**: Add non-null assertion in `git/index.ts` and `bitbucket-server/index.ts` (guaranteed set by retry loop)
- **Redundant type constituents**: `'local' | (string & {})` preserves autocomplete (`config/types.ts`)
- **Indexed object style**: Replace mapped types with `Record<string, ...>` (`util/stats.ts`)

### Test fixes
- **`no-conditional-expect`**: Unwind single-element loop in `kotlin-script/index.spec.ts`; replace manual assertions with Zod schema validation in `versioning-metadata.spec.ts`
- **`no-sparse-arrays`**: Explicit `undefined` values in `pr/labels.spec.ts`
- **`no-commented-out-tests`**: Delete dead commented-out test blocks in `codecommit/index.spec.ts`
- **`no-standalone-expect`**: Remove unnecessary precondition assertion from `git/index.spec.ts` `beforeAll`

## Context

## AI assistance disclosure

- [x] Yes — substantive assistance


lib/config/presets/internal/config.ts
```diff
@@ -1,7 +1,5 @@
 import type { Preset } from '../types.ts';
 
-/* eslint sort-keys: ["error", "asc", {caseSensitive: false, natural: true}] */
```
viceice: how do we enforce this in future when new items are added?

lib/modules/manager/deps-edn/types.ts
```diff
@@ -7,7 +7,6 @@ export type TokenTypes<T> = keyof UnionToIntersection<T[keyof T]>;
 
 export type ParsedEdnPrimitive = string | null;
 export type ParsedEdnArray = ParsedEdnData[];
-// Interface not possible due to circular reference
```
viceice, line 10: please do not remove the comment

lib/modules/manager/helmfile/utils.ts
```diff
@@ -18,8 +18,7 @@ export function kustomizationsKeysUsed(release: HelmRelease): boolean {
 }
 
 /** Returns true if a helmfile release uses a local chart with a kustomization.yaml file **/
-// eslint-disable-next-line require-await
```
viceice, line 21: why did eslint not complaining here?

lib/config/presets/internal/config.ts
zharinov: Fixed here: https://github.com/renovatebot/renovate/pull/40976

lib/modules/manager/deps-edn/types.ts
zharinov: Interface is possible now due to... evolved TypeScript?

lib/modules/manager/helmfile/utils.ts
zharinov: It would complain, but we removed `async`

lib/modules/manager/deps-edn/types.ts
zharinov: Never mind, removed this change

lib/modules/versioning/versioning-metadata.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { readdirSync } from 'node:fs';
 import { readFile } from 'fs-extra';
+import { z } from 'zod';
```
viceice: should we start to use zod v4 imports here?
```suggestion
import { z } from 'zod/v4';
```

lib/util/stats.ts
```diff
@@ -113,30 +113,33 @@ interface DatasourceCacheDataPoint {
   action: 'hit' | 'miss' | 'set' | 'skip';
 }
 
-/* eslint-disable @typescript-eslint/consistent-indexed-object-style */
 export interface DatasourceCacheReport {
-  long: {
-    [datasource in string]: {
```
viceice, line 119: please do git blame, I suggested same in initial PR 😅

lib/workers/repository/update/branch/index.spec.ts
```diff
@@ -108,9 +108,8 @@ describe('workers/repository/update/branch/index', () => {
 
     beforeEach(() => {
       scm.branchExists.mockResolvedValue(false);
-      reuse.shouldReuseExistingBranch.mockImplementation(
-        // eslint-disable-next-line require-await, @typescript-eslint/require-await
-        async (config) => config,
+      reuse.shouldReuseExistingBranch.mockImplementation((config) =>
+        Promise.resolve(config),
```
viceice, line 112:
```suggestion
      reuse.shouldReuseExistingBranch.mockResolvedValue(config
```

zharinov: No, it's incorrect. The config is being modified by the unit tests themselves. It requires deeper changes to make this work.

lib/util/stats.ts
zharinov: Extracted couple of interfaces to avoid nesting

viceice (approved):

PR #40982, opened 2026-02-06
ci: build docs from this repo

jamietanna:
## Changes

As noted in #39878, we should migrate our docs build into the main
Renovate repository.

However, we unfortunately need to rebuild the docs after a release is
made, so we can make sure that the version itself is embedded into the
docs site.

This adds on ~60s of build time, so isn't too bad performance wise.

## Context

- [x] This closes an existing Issue: Closes: #39878

## AI assistance disclosure

- [ ] Yes — other (please describe): GPT-4.1

The public repository: <URL>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -830,6 +830,63 @@ jobs:
 
       - run: df -h
 
+  build-deploy-docs-site:
+    if: ${{ github.repository == 'renovatebot/renovate' && github.ref_name == 'main' && github.event_name == 'push' }}
+    needs: release
+
+    permissions:
+      pages: write
+      id-token: write
+
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          show-progress: false
+
+      - name: Setup Node.js
+        uses: ./.github/actions/setup-node
+        with:
+          node-version: ${{ needs.setup-build.outputs.node-version }}
+          os: ${{ runner.os }}
+
+      - name: Setup PDM
+        uses: pdm-project/setup-pdm@94a823180e06fcde4ad29308721954a521c96ed0 # v4.4
+        with:
+          python-version-file: .python-version
+          version: ${{ env.PDM_VERSION }}
+          cache: true
+
+      - name: Install pdm dependencies
+        run: pdm install
+
+      - name: Determine latest release tag (whether we've released as part of this build or not)
+        id: latest_release
+        run: |
+          latest_release=$(gh release view --json tagName -q .tagName)
+          echo "release_tag=$latest_release" >> $GITHUB_OUTPUT
+        env:
+          GH_TOKEN: ${{ github.token }}
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -828,8 +831,77 @@ jobs:
...
+      - name: Upload GitHub Pages artifact
+        uses: actions/upload-pages-artifact@7b1f4a764d45c48632c6b24a0339c27f5614fb0b # v4.0.0
+        with:
+          path: tools/mkdocs/site
+
+      - name: Deploy to GitHub Pages
+        uses: actions/deploy-pages@d6db90164ac5ed86f2b6aed7e0febac5b3c0c03e # v4.0.5
+
+      - name: Deploy to GitHub Pages
+        uses: actions/deploy-pages@d6db90164ac5ed86f2b6aed7e0febac5b3c0c03e # v4.0.5
```
jamietanna:
```suggestion
```

jamietanna:
Right now, this will build under a subdirectory of the docs domain (which I've been testing with and then manually unpublished)

Once this is merged, I'll switch things over - it might require a rebuild afterwards to take effect again

.github/workflows/build.yml
```diff
@@ -830,6 +830,63 @@ jobs:
...
+        id: latest_release
+        run: |
+          latest_release=$(gh release view --json tagName -q .tagName)
+          echo "release_tag=$latest_release" >> $GITHUB_OUTPUT
+        env:
+          GH_TOKEN: ${{ github.token }}
+
+      # Rebuild the docs site to make sure that we have the version specified in the JSON Schema and page footer
+      - name: Build mkdocs for Renovate ${{ steps.latest_release.outputs.release_tag }}
+        run: pnpm mkdocs build --version ${{ steps.latest_release.outputs.release_tag }}
```
viceice, line 878: wh. not doing it in prepare release? there the version is already known. 🤔

then you simply upload the artifacts and download on publish docs step

secustor (approved):

.github/workflows/build.yml
jamietanna: What happens if:

1. Release 4x.y.z is released
2. Commit `chore(deps): this doesn't get released gets pushed`
3. `tools/prepare-release.ts` is called

What would the version be at this time? `4x.y.z`? Or `4x.y.z+1`?

viceice: it's not called because no release required. it's only required when a new release will made. 🤔 do we really need to push docs changes whithout a release? we don't do that currently because of:
- https://github.com/renovatebot/renovatebot.github.io/pull/739

so i don't see a need to do it again. we usually have at least one release daily and we never hat to fix a docs issue asap yet.

jamietanna: Ah, yeah so as part of this work (under https://github.com/renovatebot/renovate/issues/39878) I'd like to move back to always pushing docs changes, even without a release

I realise we release _fairly_ often, but sometimes there aren't releases for a day, which can mean that docs changes (to clarify something important) can then get missed - I'm sure I've seen that 1-2 times since I've become a maintainer

So the intent here is that we'd _always_ update the docs site on each commit

It's also a clearer flow - "merge to `main` means docs get updated" rather than a contributor thinking "oh, what semantic commit type did I have? will that get released"?

jamietanna: I'm going to merge this as-is - if we feel we would prefer to migrate it to only when there's a release, we can change it back 👍

PR #39568, opened 2025-11-26
docs: add Renovate CLI version to docs footer

jamietanna:
## Changes

As part of wiring in the version of the CLI into the docs build process,
we should also show this version to users, so they're aware what version
of the docs they're looking at.

To put this into the footer, the simplest place is to overwrite the
existing `copyright.html`[0], and remove the (unused) copyright
configuration.

We make the inclusion of the version optional, to make it possible to
merge + release this before the docs site has the version wired in.

[0]: https://github.com/squidfunk/mkdocs-material/blob/9.7.0/material/templates/partials/copyright.html
## Context

Please select one of the below:

- [ ] This closes an existing Issue: #
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [x] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #40988, opened 2026-02-06
chore(tools/mkdocs): ensure `RENOVATE_VERSION` is passed to `mkdocs`

jamietanna:
## Changes

As part of #39568, we add the Renovate
version into the footer based on the environment variable.

When we set the `--version` flag for the `generateDocs()` call, we
should also make sure that `mkdocs` is called with `RENOVATE_VERSION`.

## Context

## AI assistance disclosure

The public repository: <URL>


tools/mkdocs.ts
```diff
@@ -29,6 +29,9 @@ program
     const res = exec('pdm', args, {
       cwd: 'tools/mkdocs',
       stdio: 'inherit',
+      env: {
+        RENOVATE_VERSION: opts.version ?? '',
+      },
```
jamietanna, line 35: This might actually break it, I think this overwrites `env`?

viceice: yes, you need to pass existing env too

viceice (commented):
see comment

tools/mkdocs.ts
```diff
@@ -29,6 +29,10 @@ program
     const res = exec('pdm', args, {
```
viceice, line 29: btw: we should change the exec implementation to use execa for better error handling

viceice (approved):

tools/mkdocs.ts
jamietanna: #41062

PR #40989, opened 2026-02-06
fix(pep440): handle v-prefixed versions gracefully

felipecrs:
## Changes

Fixes handling of v-prefixed versions (e.g., `v0.7.15`) in pep440 versioning that were previously causing `Invalid pep440 currentValue` errors. The fix treats v-prefixed bare versions as pinned versions, bypassing the range parser which doesn't support them, even though Poetry handles them correctly.

## Context

## AI assistance disclosure

Claude Opus 4.6 was used to research the issue, implement the fix, and add tests. Manual review and testing were performed to ensure correctness.

The public repository:

- https://github.com/felipecrs/renovate-repro-poetry-v-prefix/pull/3


lib/modules/versioning/pep440/range.spec.ts
```diff
@@ -35,4 +35,14 @@ describe('modules/versioning/pep440/range', () => {
...
   });
+
+  it('handles v-prefixed version as currentValue', () => {
+    const res = getNewValue({
+      currentValue: 'v0.7.15',
+      rangeStrategy: 'auto',
+      newVersion: '0.8.0',
+      currentVersion: '0.7.15',
+    });
+    expect(res).toBe('0.8.0');
```
viceice: shouldn't it retain the v prefix?

felipecrs: Not necessarily. Poetry just happens to accept the v prefix, but I don't think it's recommended.

But I agree Renovate should probably take as less opinions as possible.

I'll check if the v prefix can be retained cleanly.

felipecrs: fef729ac18dc43708620eaf21aa62da027cafd7a

lib/modules/versioning/pep440/range.ts
```diff
@@ -117,6 +117,17 @@ export function getNewValue({
     return newVersion;
   }
 
+  // Handle bare versions (e.g., "v0.7.15") that don't strictly equal
+  // currentVersion due to normalization, treating them as pinned while
+  // preserving the v-prefix.
+  if (parseVersion(currentValue)) {
+    const vPrefix = regEx(/^([vV])/).exec(currentValue);
```
viceice: use named capture groups

viceice (commented):
needs test to get full coverage

lib/modules/versioning/pep440/range.spec.ts
```diff
@@ -35,4 +35,24 @@ describe('modules/versioning/pep440/range', () => {
...
+  });
+
+  it('handles bare version that differs from currentVersion without v-prefix', () => {
+    const res = getNewValue({
+      currentValue: '1.0.0.0',
+      rangeStrategy: 'auto',
+      newVersion: '1.2.3',
+      currentVersion: '1.0.0',
+    });
+    expect(res).toBe('1.2.3');
```
RahulGautamSingh, line 56: changing from 4 to 3 parts?

felipecrs: Yes, that's a bit weird.

But I think it would be even more weird add a `.0` to the version found in the datasource.

Don't you think?

In that sense, it may make sense to also drop the `v` prefix.

@viceice any thoughts?

Both cases are human mistakes that Poetry tolerates, from what I can tell. Maybe Renovate should just "fix" it (i.e. drop the `v` prefix and drop the fourth digit).

felipecrs: Just FYI, after reflecting on this throughout the day I believe this PR as-is is the best compromise.

viceice: the behavior isn't changed by this PR. just an added test which shows a weird existing behavior, so not in scope of this PR

viceice (approved):

RahulGautamSingh (approved):

felipecrs:
Thanks a lot for the review and merge, @RahulGautamSingh and @viceice!

PR #41040, opened 2026-02-09
fix(deps): ensure `tar` override works for `npm` installs

jamietanna:
## Changes

It appears that the override we specified for `tar` only takes effect
when Renovate is built via `pnpm`.

However, if users install via `npm install`, the npm doesn't
(rightfully) understand the `pnpm.overrides`.

This is a follow-up to #40766.

## Context

## AI assistance disclosure

The public repository: <URL>


package.json
```diff
@@ -397,5 +397,8 @@
       "esbuild": "0.27.2",
       "vite": "8.0.0-beta.11"
     }
+  },
+  "overrides": {
+    "tar": "7.5.7"
```
jamietanna, line 402: I think this will mean it's _always_ that version, so we'll need to make sure to update this once `yarnpkg/core` is updated to use a new version? https://github.com/yarnpkg/berry/issues/7039

viceice: I'm not sure if we really should do this 🤔

jamietanna: Thoughts on alternatives? Or do we think there's not a good option for folks installing via `npm`?

viceice: I think it's not worth the effort because the vulnerable code isn't used from renovate. we only use the parser which doesn't need tar at all.

viceice (commented):
i still think we should not do this, it can cause bad side effects

PR #41130, opened 2026-02-12
fix(gitlab): skip branch status when no pipeline found, retry after PR creation

### LINKED ISSUE #41241: Extra branch pipeline created when setting status checks with MR pipelines (GitLab)
### Discussed in https://github.com/renovatebot/renovate/discussions/41129

<sup>Originally posted by **fgreinacher** February 12, 2026</sup>

Self-hosted Renovate CLI

GitLab.com and GitLab Self-hosted

v43

When a GitLab project is configured to use MR (merge request) pipelines, Renovate creates an unwanted additional branch pipeline when setting stability/confidence status checks:

image

This causes multiple problems:

  - Wasted CI resources / minutes: The extra pipeline runs CI jobs unnecessarily, consuming runners and GitLab CI minutes quota
  - Branch pipeline may fail: Projects configured for MR-only pipelines often use $CI_MERGE_REQUEST_* variables in their CI config that aren't available in   branch pipelines, causing failures
  - Confusing status: Two pipelines on the same commit make it unclear which result matters
  - May block automerge: A failing branch pipeline can prevent Renovate from automerging, even though the MR pipeline succeeded

#### Root cause

**Status checks are set before the MR exists.** In `processBranch()`, `setStability()` and `setConfidence()` are called before `ensurePr()`. For new branches, the MR hasn't been created yet, so GitLab has no MR pipeline. `setBranchStatus()` posts the commit status via `POST /projects/:id/statuses/:sha` without a `pipeline_id`, which causes GitLab to create a branch pipeline.

#### Expected behavior

Renovate should associate the commit status with the existing MR pipeline and not trigger a separate branch pipeline.

#### Current behavior

An additional branch pipeline is created every time Renovate sets a stability or confidence status check on a branch that uses MR pipelines.

### Reproduction

See https://gitlab.com/fgreinacher/renovate-status-check-extra-branch-pipeline/-/merge_requests/7/pipelines
  

### Logs (if relevant)

n/a</div>

fgreinacher:
## Changes

When a GitLab project uses MR pipelines, Renovate was creating an unwanted branch pipeline by posting commit statuses without a `pipeline_id`. This happened because `setBranchStatus()` posted the status even when no pipeline could be found after retrying.

This PR makes two changes:

1. **Skip posting status when no pipeline is found** (`lib/modules/platform/gitlab/index.ts`): After the retry loop in `setBranchStatus()`, if `RENOVATE_X_GITLAB_SKIP_STATUS_WITHOUT_PIPELINE` is set and no `pipeline_id` was found, skip posting the status to avoid creating an unwanted branch pipeline.

2. **Retry setting status checks after PR/MR creation** (`lib/workers/repository/update/branch/index.ts`): Extract `setBranchStatusChecks()` helper and call it in two places:
   - At the original location (before early returns) — works for all platforms and existing MRs with pipelines
   - Inside the `with-pr` block after `ensurePr()` — retries on GitLab after MR creation so the MR pipeline can be found

The skip behavior is gated behind the `RENOVATE_X_GITLAB_SKIP_STATUS_WITHOUT_PIPELINE` experimental environment variable. The retry after PR creation is always active and is a no-op on other platforms since `setStatusCheck` skips when `existingState === state`.

## Context

- [x] This closes an existing Issue, Closes: #41241

## AI assistance disclosure

  - Claude Code helped me find the right place to make this change and analyze the code flow for potential issue with the slightly changed execution order.

The public repository: https://gitlab.com/fgreinacher/renovate-status-check-extra-branch-pipeline-fixed/-/merge_requests/4/pipelines

:hammer_and_pick: with :heart: by [Siemens](https://opensource.siemens.com/)


lib/workers/repository/update/branch/index.ts
RahulGautamSingh, line 1: But these status checks are for branch not MR

This will now cause branch only statuses for github to not be applied till PR is created. And, having a PR always is not necessary

fgreinacher: Good point! I have reworked the logic a bit so that we keep the original calls and additionally set the status after PR creation. What do you think about this?

lib/modules/platform/gitlab/index.ts
```diff
@@ -997,6 +996,11 @@ export async function setBranchStatus({
     logger.warn('Failed to retrieve commit pipeline');
   }
 
+  if (options.pipeline_id === undefined) {
```
viceice: this still doesn't set the branch status when no pipeline exists 🤔

fgreinacher: Yes true, that's the drawback of this change. But I'm not sure there are many cases where folks don't have any kind of pipeline and still rely on the branch status. If we'd like to stay fully backward compatible we might also put this behavior behind an experimental config option to start with?

viceice: yes, let's use an experimental option to opt-in. in next major we can then flip it. and remove the options maybe on some later major

fgreinacher: Alright, done. The new behavior is now opt-in via `RENOVATE_X_GITLAB_SKIP_STATUS_WITHOUT_PIPELINE`. Let me know what you think!

docs/usage/self-hosted-experimental.md
```diff
@@ -82,6 +82,11 @@ If set, Renovate will use this as a delay to proceed with an automerge.
 
 Default value: `250` (milliseconds).
 
+## `RENOVATE_X_GITLAB_SKIP_STATUS_WITHOUT_PIPELINE`
+
+If set to any value, Renovate will skip setting a branch status check on GitLab when no pipeline is found for the commit.
```
viceice:
```suggestion
If set to `true` value, Renovate will skip setting a branch status check on GitLab when no pipeline is found for the commit.
```

makes later migration to boolean easier

lib/modules/platform/gitlab/index.ts
```diff
@@ -997,6 +996,14 @@ export async function setBranchStatus({
     logger.warn('Failed to retrieve commit pipeline');
   }
 
+  if (
+    options.pipeline_id === undefined &&
+    env.RENOVATE_X_GITLAB_SKIP_STATUS_WITHOUT_PIPELINE
```
viceice:
```suggestion
    env.RENOVATE_X_GITLAB_SKIP_STATUS_WITHOUT_PIPELINE === 'true'
```

viceice (commented):
needs some code coverage

viceice (approved):

Issue #40848, opened 2026-02-02
Remove deprecated npm-based presets


jamietanna:
### Describe the proposed change(s).

Via https://github.com/renovatebot/renovate/blob/dc9f86873dfeb115cfe808ed71a83b0ccf2e6337/docs/usage/config-presets.md?plain=1#L28-L31, we deprecated npm-based presets in #14444 / [31.95.0](https://github.com/renovatebot/renovate/releases/tag/31.95.0) (2022-02-28)

PR #41143, opened 2026-02-13
feat: add support for jsonc as a config file suffix

rtpg:
## Changes

Based off of [this discussion](https://github.com/renovatebot/renovate/discussions/38060), I'm adding support for config files ending in `.jsonc`, so that people who want to can specify that their config is JSONC with the suffix.

## Context

Doing this allows for JSONC files to be automatically identified by stuff like linters.

## AI assistance disclosure

I used AI for my initial search of trying to find places that would need changing. 

- [X] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).

I was having trouble getting 

The public repository: <URL>

### DIFF
--- a/lib/config/app-strings.ts
+++ b/lib/config/app-strings.ts
@@ -5,13 +5,17 @@ import { regEx } from '../util/regex.ts';
 const configFileNames = [
   'renovate.json',
   'renovate.json5',
+  'renovate.jsonc',
   '.github/renovate.json',
   '.github/renovate.json5',
+  '.github/renovate.jsonc',
   '.gitlab/renovate.json',
   '.gitlab/renovate.json5',
+  '.gitlab/renovate.jsonc',
   '.renovaterc',
   '.renovaterc.json',
   '.renovaterc.json5',
+  '.renovaterc.jsonc',
   'package.json',
 ];
 
--- a/lib/workers/global/config/parse/util.ts
+++ b/lib/workers/global/config/parse/util.ts
@@ -52,6 +52,7 @@ export async function getParsedContent(file: string): Promise<RenovateConfig> {
     case '.yaml':
     case '.yml':
       return parseSingleYaml(await readSystemFile(file, 'utf8'));
+    case '.jsonc':
     case '.json5':
     case '.json':
       return parseJson(
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -12,13 +12,17 @@ You can store your Renovate configuration file in one of these locations:
 
 1. `renovate.json`
 1. `renovate.json5`
+1. `renovate.jsonc`
 1. `.github/renovate.json`
 1. `.github/renovate.json5`
+1. `.github/renovate.jsonc`
 1. `.gitlab/renovate.json`
 1. `.gitlab/renovate.json5`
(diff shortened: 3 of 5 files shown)


RahulGautamSingh (changes requested):
Please link a working example of this,

lib/config/app-strings.ts
```diff
@@ -5,13 +5,17 @@ import { regEx } from '../util/regex.ts';
 const configFileNames = [
   'renovate.json',
   'renovate.json5',
+  'renovate.jsonc',
```
viceice, line 8: We should really refactor it to be more generic. so first try `renovate.json` then replace extention and try json5 and jsonc.

maybe just use a glob here? `renovate.{json,json5,jsonc}`, so it finds first match. 🤔

lib/config/app-strings.ts
```diff
@@ -5,13 +5,17 @@ import { regEx } from '../util/regex.ts';
...
   '.github/renovate.json5',
+  '.github/renovate.jsonc',
   '.gitlab/renovate.json',
   '.gitlab/renovate.json5',
+  '.gitlab/renovate.jsonc',
   '.renovaterc',
   '.renovaterc.json',
   '.renovaterc.json5',
+  '.renovaterc.jsonc',
   'package.json',
```
viceice, line 19: @jamietanna didn't we removed the support in v43? if not we should in v44. i think it's already deprecated? 🤔

jamietanna: Deprecated, not yet removed: https://github.com/renovatebot/renovate/issues/40848 

(Happy for us to do it in 44)

lib/config/app-strings.ts
jamietanna: Glob sounds reasonable - means it's a bit more predictable

Related: https://github.com/renovatebot/renovate/issues/40608 (would be nice to get this auto-updating?)

viceice: what should we do with this pr? should we first refactor (which i would prefer)?

jamietanna: For getting the docs auto-updating:

- https://github.com/renovatebot/renovate/pull/41250
- https://github.com/renovatebot/renovate/pull/41252
- https://github.com/renovatebot/renovate/pull/41253
- https://github.com/renovatebot/renovate/pull/41254

(which I would see as a required change before we get this in)

I agree that it would be good to glob - are you thinking that we would replace these lines with i.e. `

```suggestion
  'renovate.json{,5,c}',
```

?

lib/config/app-strings.ts
jamietanna, line 1: Missing a change further down in this file + would be good to check that the filtering works correctly for platform-specific filenames

lib/workers/global/config/parse/util.ts
jamietanna, line 1: Would also be good to have a test here to make sure that we don't break this in the future

lib/config/app-strings.ts
jamietanna: https://github.com/renovatebot/renovate/pull/41258 is an attempt at the usage of brace expansions, if this is what you're thinking?

jamietanna:
Replaced by https://github.com/renovatebot/renovate/pull/43677

PR #41171, opened 2026-02-13
feat(presets/monorepos): add k8s.io repos

felipecrs:
## Changes

This PR adds a monorepo group for most of the k8s.io go packages.

The real source of these packages is https://github.com/kubernetes/kubernetes, but there's an automation that splits them from the monorepo into individual repositories and then publishes them to the go registry. They are all tagged at the same time and with the same tag. The full list of packages is fortunately documented and can be found here:

- https://github.com/kubernetes/kubernetes/tree/638bdc9df6a7491b6270ee5ec244389e6b499e9c/staging/README.md

While the name of these packages are all prefixed with `k8s.io`, we should not rely on that because there are certain `k8s.io` packages that don't belong together, like `k8s.io/klog/v2` which has its own release cycle and versioning.

## Context

## AI assistance disclosure

The public repository: N/A


viceice (dismissed):
🙈  hopefully this won't change a lot

felipecrs:
> 🙈 hopefully this won't change a lot

I also hope so!

lib/data/monorepo.json
```diff
@@ -423,6 +423,39 @@
       "https://github.com/junit-team/junit5",
       "https://github.com/junit-team/junit-framework"
     ],
+    "k8s.io": [
```
viceice:
```suggestion
    "k8s-io": [
```

dot isn't allowed

felipecrs: @viceice, I see.

But when generating the PR, having it named as `k8s.io` would make a lot more sense for go developers, as that's how these packages are named:

[k8s.io/api](https://github.com/kubernetes/api)
[k8s.io/apiextensions-apiserver](https://github.com/kubernetes/apiextensions-apiserver)
[k8s.io/apimachinery](https://github.com/kubernetes/apimachinery)
[k8s.io/apiserver](https://github.com/kubernetes/apiserver)
[k8s.io/cli-runtime](https://github.com/kubernetes/cli-runtime)
[k8s.io/client-go](https://github.com/kubernetes/client-go)
etc.

Should I really change to `k8s-io` or could I add it to the "exception" list?

https://github.com/renovatebot/renovate/blob/c531748f09f5508bf4122e16ade75d7ac5ad98db/lib/config/presets/internal/monorepos.spec.ts#L4-L19

viceice: exception list is for legacy only, so rename

viceice (approved):

PR #41180, opened 2026-02-14
fix(datasource/python-version): gracefully handle python.org 429 rate limiting

wgordon17:
## Summary

- When python.org returns HTTP 429 (rate limited), the datasource now falls back to containerbase prebuild releases instead of throwing `ExternalHostError` which aborts the entire repository
- Non-429 errors continue to be handled by `handleGenericErrors()` as before
- Prebuild releases still go through the EOL deprecation loop

## Test plan

- [ ] New test: `falls back to prebuild releases on 429` — mocks python.org returning 429, verifies releases come from prebuild data with EOL applied
- [ ] Existing 500 and error tests unchanged and passing

### DIFF
--- a/lib/modules/datasource/python-version/index.spec.ts
+++ b/lib/modules/datasource/python-version/index.spec.ts
@@ -80,6 +80,39 @@ describe('modules/datasource/python-version/index', () => {
       ).toBeNull();
     });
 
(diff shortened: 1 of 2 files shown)


jamietanna (approved):

jamietanna:
Thanks!

PR #41187, opened 2026-02-14
feat: lib years merge confidence badge

setchy:
## Changes

Add `Lib Year` merge confidence badge

Screenshot 2026-02-14 at 3 47 06 PM

## Context

New badge to display the age of the current version - https://libyear.com/

This new badge helps convey the age of the current version, which is a useful data-point when contrasted with the update/patch `Age`

## AI assistance disclosure

The public repository: <URL>


RahulGautamSingh (changes requested):
~We can be more descriptive. LibYear and Age together might confuse new users regarding which ages are being conveyed.

Maybe `Release Age` would be less confusing~

I think we can go with this, as the refs to documentation on the Age and LibYear would help users know more incase they're confused.

docs/usage/bot-comparison.md
```diff
@@ -81,7 +81,8 @@ If you're an advanced user, you may use the [`dependabot-core` repository](https
 
 Renovate shows four _Merge Confidence_ badges in its PRs:
 
-- **Age**: The age of the package
+- **Lib Year**: The age of the current package version
```
RahulGautamSingh, line 84: Nitpick: Let's put this atfer labels starting with "A".

lib/config/options/index.ts
```diff
@@ -2903,6 +2905,7 @@ const options: Readonly<RenovateOptions>[] = [
     freeChoice: true,
     mergeable: true,
     default: {
+      'Lib Year': '[Lib Year](https://docs.renovatebot.com/merge-confidence/)',
```
RahulGautamSingh, line 2908: I'd rather have this link to official libyear page

lib/config/options/index.ts
```diff
@@ -2884,6 +2884,8 @@ const options: Readonly<RenovateOptions>[] = [
       Pending: '{{{displayPending}}}',
       References: '{{{references}}}',
       'Package file': '{{{packageFile}}}',
+      'Lib Year':
+        "{{#if newVersion}}![libYear](https://developer.mend.io/api/mc/badges/age/{{datasource}}/{{replace '/' '%2f' packageName}}/{{{currentVersion}}}?slim=true){{/if}}",
```
RahulGautamSingh, line 2888: If we already calculate this for each extracted dep during the run, it'll be better to just re-use the value here instead of fetching from mend.

RahulGautamSingh (changes requested):

lib/config/options/index.ts
setchy: This fetches the SVG badge which is different to the raw libYear value.

docs/usage/bot-comparison.md
setchy: I did consider making it alphabetic, but kept it in the order which the PR body columns reference them for consistency with the other badges (Confidence, etc)

lib/config/presets/internal/merge-confidence.preset.ts
```diff
@@ -31,6 +31,7 @@ export const presets: Record<string, Preset> = {
         prBodyColumns: [
           'Package',
           'Change',
+          'Lib Year',
```
viceice, line 34: I don't like to have another badge by default. I think it's already very overloaded. 🤔

lib/config/options/index.ts
viceice: should the PR be in draft until it's supported server side? currently it's same as age badge. 🤔

jamietanna (changes requested):
Will review on Monday - got some thoughts generally :)

lib/config/options/index.ts
setchy: My thinking was we can use the same server side endpoint to create the badge svg for Lib Years since the only difference is calculating the Age using current version vs new version.

lib/config/presets/internal/merge-confidence.preset.ts
setchy: This is the `all-badges` preset ;)

setchy: I did leave it out of the default columns ;)

setchy:
Non-urgent, but am keen to hear your thoughts @jamietanna 🙏

docs/usage/bot-comparison.md
```diff
@@ -81,7 +81,8 @@ If you're an advanced user, you may use the [`dependabot-core` repository](https
 
 Renovate shows four _Merge Confidence_ badges in its PRs:
```
jamietanna: If we were doing this:

```suggestion
Renovate shows five _Merge Confidence_ badges in its PRs:
```

jamietanna (changes requested):
OK, a few thoughts:

- Right now, this badge doesn't provide any value, because Mend's backend doesn't provide an API to show LibYears for a given dependency
- Work to add that is possible, but...
- Renovate itself already has the LibYears data 
  - albeit it's calculated at the end of the run, so would need to move after the `lookup` split
- With that data, we could provide (in the PR) i.e. `5.2`
- This also doesn't really fit under "Merge Confidence" as a thing - so we would not want it to be documented there, or included in the "Merge Confidence badges" 

Also, from my time in Open Source Program Offices and working on metadata around dependencies (including work with CHAOSS), I've spent time considering [LibYears as a metric](https://chaoss.community/kb/metric-libyears/).

There are a lot of ecosystem-specific gotchas that lead to a tonne of nuance needed when taking in the LibYear value, and this is an area I want us to be careful on messaging with - a low LibYear != "this will be easy to update to" or "you're in a good place"

If we do want to add this to the PR (which I think is fine, as an opt-in, it'd need to be collected from Renovate's calculation itself, _and_ I think we need to make sure we have some documentation about LibYears and what they mean, so folks can understand the nuance moreso

setchy:
> _Right now, this badge doesn't provide any value, because Mend's backend doesn't provide an API to show LibYears for a given dependency_
> _Work to add that is possible, but..._

I agree, it _could_ be a new Mend backend API for telemtry/metrics, etc, but doesn't seem like a firm req since its still just an age badge based on `currentVersion` vs `newVersion`.

Alternatively, there is `currentVersionAgeInDays` which could just be used as-is without a svg badge

How about this pivot...
- Don't add this as a `Merge Confidence` concept.
- Add `Current Age` or `Lib Years` to [prBodyDefinitions](https://docs.renovatebot.com/configuration-options/#prbodydefinitions)
- Leave it as an opt-in config via `prBodyColumns`
- Add a preset / helper to make this easier

setchy:
Quickly mocked this up

Screenshot 2026-02-27 at 9 29 23 AM


```
{
      "description": "Configure PR Body Columns for major, minor and patch updates to include merge confidence data",
      "matchUpdateTypes": [
        "major",
        "minor",
        "patch"
      ],
      "prBodyColumns": [
        "Package",
        "Package file",
        "Current Age",
        "Change",
        "Age",
        "Adoption",
        "Passing",
        "Confidence"
      ]
    },
  "prBodyDefinitions": {
    "Current Age": "![currentAge](https://developer.mend.io/api/mc/badges/age/{{datasource}}/{{replace '/' '%2f' packageName}}/{{{currentVersion}}}?slim=true)",
  },
,
```

It may be clearer if the current `Age` column was `Update Age` (or something along those lines)

PR #41244, opened 2026-02-17
fix(libyear): skip disabled dependencies

YashSaini15:
## Changes

This PR updates libyear calculation to skip disabled dependencies (`enabled: false`) entirely.

Following discussion, it was confirmed that disabled dependencies should not have updates available and therefore should not contribute to LibYears calculations or related logging.

This change skips them at the start of the loop and updates the tests accordingly.

## Context

## AI assistance disclosure

Details:
AI assistance was used to help analyse the existing codebase, identify the correct location for the fix, and draft the initial implementation and unit test structure. All code was reviewed, adapted, and validated locally.

The public repository: N/A (unit tests added and full test suite run locally)

### DIFF
--- a/lib/workers/repository/process/libyear.spec.ts
+++ b/lib/workers/repository/process/libyear.spec.ts
@@ -197,13 +197,13 @@ describe('workers/repository/process/libyear', () => {
         {
           libYears: {
             managers: {
-              npm: 2,
+              npm: 1,
             },
-            total: 2,
+            total: 1,
           },
           dependencyStatus: {
-            outdated: 2,
-            total: 2,
+            outdated: 1,
+            total: 1,
           },
         },
         'Repository libYears',
@@ -211,13 +211,13 @@ describe('workers/repository/process/libyear', () => {
       expect(addLibYears).toHaveBeenCalledExactlyOnceWith(config, {
         libYears: {
           managers: {
-            npm: 2,
+            npm: 1,
           },
-          total: 2,
+          total: 1,
         },
         dependencyStatus: {
-          outdated: 2,
-          total: 2,
+          outdated: 1,
+          total: 1,
         },
       });
     });
--- a/lib/workers/repository/process/libyear.spec.ts
+++ b/lib/workers/repository/process/libyear.spec.ts
@@ -301,7 +301,7 @@ describe('workers/repository/process/libyear', () => {
       );
     });
 
-    it('ignores disabled dependencies without currentVersionTimestamp', () => {
+    it('ignores disabled dependencies', () => {
       const packageFiles: Record<string, PackageFile[]> = {
         npm: [
           {
--- a/lib/workers/repository/process/libyear.spec.ts
+++ b/lib/workers/repository/process/libyear.spec.ts
@@ -300,5 +300,48 @@ describe('workers/repository/process/libyear', () => {
         'Repository libYears',
       );
     });
+
+    it('ignores disabled dependencies without currentVersionTimestamp', () => {
+      const packageFiles: Record<string, PackageFile[]> = {
+        npm: [
+          {
+            packageFile: 'package.json',
+            deps: [
+              {
+                depName: 'disabled-dep',
+                datasource: 'npm',
+                currentVersion: '1.0.0',
+                enabled: false,
+                updates: [
+                  {
+                    newVersion: '2.0.0',
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+      };
+
+      calculateLibYears(config, packageFiles);
+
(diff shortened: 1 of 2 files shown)


viceice (dismissed):

jamietanna (changes requested):
This doesn't seem to close the issue it's mentioned it will?

jamietanna (commented):
I'm also not sure if this should be the case - I wonder if we should split this out into a separate metric, as it would be useful for folks to know that "these dependencies you said that you don't want updating are now 500 LibYears behind"

jamietanna:
I'll also need to confirm how (if at all) this affects how Mend presents these calculations in various places

YashSaini15:
> This doesn't seem to close the issue it's mentioned it will?

Thanks for checking.
The original issue references a docs-side timestamp check that no longer exists in the current codebase. After tracing the behavior in main, the same underlying problem still occurs during libyear processing: disabled dependencies (enabled: false) are logged as missing currentVersionTimestamp.

This PR addresses that behavior in lib/workers/repository/process/libyear.ts, which is where the timestamp check currently lives. Functionally, it resolves the same issue described in #40725, but in its updated location.

If you prefer, I’m happy to remove the “Closes #40725” linkage and treat this as a standalone fix instead.

viceice (commented):
LGTM for me 🤷‍♂️

jamietanna:
> The original issue references a docs-side timestamp check that no longer exists in the current codebase

Yeah that issue is for https://docs.renovatebot.com/key-concepts/minimum-release-age/#which-datasources-support-release-timestamps which isn't the same thing as here - removing the `Closes` would be appreciated!

I'll take another look at this next week, as I'm juggling a few things and so won't be able to confirm our internal usage of this and how this may affect it until now - thanks for your patience!

YashSaini15:
> > The original issue references a docs-side timestamp check that no longer exists in the current codebase
> 
> Yeah that issue is for https://docs.renovatebot.com/key-concepts/minimum-release-age/#which-datasources-support-release-timestamps which isn't the same thing as here - removing the `Closes` would be appreciated!
> 
> I'll take another look at this next week, as I'm juggling a few things and so won't be able to confirm our internal usage of this and how this may affect it until now - thanks for your patience!

Thanks for the clarification — makes sense.
I’ve removed the “Closes #40725” linkage and left this as a standalone fix.
Happy to wait while you confirm internal usage.

jamietanna:
Is the primary intent here to avoid the log line appearing, when the package is disabled?

jamietanna (changes requested):
If we're trying to avoid the log line, let's refactor this (once https://github.com/renovatebot/renovate/pull/41488 is merged) to only avoid the log line when disabled dependencies are called

YashSaini15:
> Is the primary intent here to avoid the log line appearing, when the package is disabled?

My original intent was to skip disabled dependencies entirely — both suppressing the “No currentVersionTimestamp” log and excluding them from libyear calculations — because `enabled: false` generally means “ignore this dependency.”

However, I’m happy to adjust this PR to only suppress the log message for disabled dependencies if that better fits the expected behavior.

Please let me know the preferred direction for this change.

jamietanna:
In discussion with @viceice I'm now more confident that it _shouldn't_ be possible for an `enabled: false` dependency to also have `updates` available, so we shouldn't be receiving any LibYears data from them

So I think continuing with the approach you had - of skipping the entire loop - is fine. Once you pull in from `main` you can update the test from https://github.com/renovatebot/renovate/pull/41488 to show that it'll no longer use that data

Please also make sure the PR description is updated :)

YashSaini15:
Rebased onto latest main and updated the test to reflect that disabled dependencies are skipped entirely. All tests passing locally.

jamietanna (changes requested):
Can you please update the test name (L144 in lib/workers/repository/process/libyear.spec.ts) ? As it's no longer true

lib/workers/repository/process/libyear.spec.ts
```diff
@@ -300,5 +300,48 @@ describe('workers/repository/process/libyear', () => {
         'Repository libYears',
       );
     });
+
+    it('ignores disabled dependencies without currentVersionTimestamp', () => {
```
jamietanna:
```suggestion
    it('ignores disabled dependencies', () => {
```

jamietanna (approved):
This _shouldn't_ break anything, as it _shouldn't_ be possible for `enabled: false` to have `updates`, but if we see any regressions, we'll revert this, and only apply the logic to the log line

viceice (approved):

PR #41323, opened 2026-02-19
fix(workers/global): migrate string-based `repositoryCacheForceLocal`

jamietanna:
## Changes

For users who had previously set `RENOVATE_X_REPO_CACHE_FORCE_LOCAL` to
a non-empty string, we can make sure to auto-migrate them to the
expected value of `true`.

Because this is an environment variable, we cannot use a migration, and
so instead need to handle it when we're performing the manual migration.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/workers/global/config/parse/env.ts
+++ b/lib/workers/global/config/parse/env.ts
@@ -1,4 +1,4 @@
-import { isArray } from '@sindresorhus/is';
+import { isArray, isNonEmptyString } from '@sindresorhus/is';
 import JSON5 from 'json5';
 import { getOptions } from '../../../../config/options/index.ts';
 import type { AllConfig } from '../../../../config/types.ts';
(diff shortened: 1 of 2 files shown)


lib/workers/global/config/parse/env.ts
```diff
@@ -117,6 +117,9 @@ function massageConvertedExperimentalVars(
       // special case to use a more consistent prefix with other `repositoryCache` options
       if (key === 'RENOVATE_X_REPO_CACHE_FORCE_LOCAL') {
         newKey = 'RENOVATE_REPOSITORY_CACHE_FORCE_LOCAL';
+        if (isNonEmptyString(env[key])) {
+          env[key] = 'true';
+        }
```
viceice, line 122:
```suggestion
        } else {
          continue;
        } 
```
don't forward any other value?

viceice (commented):
otherwise lgtm

lib/workers/global/config/parse/env.ts
jamietanna: If it is empty, it doesn't seem to do anything, so this shouldn't be needed? 

But good thinking

viceice (approved):

PR #41333, opened 2026-02-19
feat(logger): take function parameters into account for deduplication

jamietanna:
## Changes

Previously, the `log.once` call would only take into account the "call
site" of the file + line number.

However, we have areas in the codebase (currently, and as part of future
changes) where we need to be able to call the `logger.once` functions
from the same line, but with different parameters.

We then need to make sure that the parameters are correctly passed in to
our `logOnceFn`, so we can then use them.

We use `safe-stable-stringify` to provide a more performant means to
stringify, and then hash the result to avoid large keys.

Because the `logger.once` should only be used with our logger, we can
also make the parameters required, and update tests accordingly.

## Context

## AI assistance disclosure

- [ ] Yes — other (please describe): Claude Sonnet 4.5 (GitHub Copilot + `crush`)

The public repository: <URL>


jamietanna:
Codecov isn't showing a report, as it's still thinking we're running against the old base commit 🤔

viceice (dismissed):

lib/logger/once.ts
```diff
@@ -1,3 +1,6 @@
+import { createHash } from 'crypto';
```
viceice:
```suggestion
import { createHash } from 'node:crypto';
```

to explicit show we want the node built-in

viceice (commented):
otherwise lgtm

viceice (approved):

PR #41386, opened 2026-02-22
fix: resolve cache TTL lazily in `PackageHttpCacheProvider`

zharinov:
## Changes

`PackageHttpCacheProvider` resolved `cacheTtlOverride` eagerly in its constructor.
The Maven datasource creates its instance at module level, before `GlobalConfig.set()`, so the override was silently ignored.

Defer `resolveTtlValues()` to usage time so it always reads current config.

## Context

Discussion: https://github.com/renovatebot/renovate/discussions/41320

## AI assistance disclosure

- [x] Code inspection only


RahulGautamSingh (approved):

jamietanna (approved):
Would be good to see if there's a test we can write for this case, but also whether there are other automated ways to make sure this doesn't happen again across other datasources

PR #41423, opened 2026-02-23
docs(versioning): auto-generate `pin` documentation

jamietanna:
## Changes

As a follow-up from #36261, we should auto-generate this in our
documentation to keep it up-to-date.

If the `getPinnedValue` function is present, we should imply
`rangeStrategy=pin`.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/tools/docs/versioning.ts
+++ b/tools/docs/versioning.ts
@@ -1,5 +1,8 @@
 import { codeBlock } from 'common-tags';
-import { getVersioningList } from '../../lib/modules/versioning/index.ts';
+import {
+  type VersioningApi,
+  getVersioningList,
+} from '../../lib/modules/versioning/index.ts';
 import type { RangeStrategy } from '../../lib/types/versioning.ts';
 import { readFile, updateFile } from '../utils/index.ts';
 import {
@@ -19,6 +22,7 @@ interface Versioning {
   urls: string[];
   supportsRanges: boolean;
   supportedRangeStrategies?: RangeStrategy[];
+  api: VersioningApi;
 }
 
 export async function generateVersioning(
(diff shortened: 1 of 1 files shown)


jamietanna:
Currently:

- https://docs.renovatebot.com/modules/versioning/cargo/

Now:

```markdown
# Cargo Versioning

## Identifier

 `cargo` 

## References

 - [https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)

## Ranges/Constraints

✅ Ranges are supported.

Valid `rangeStrategy` values are: `bump`, `replace`, `pin`
```

tools/docs/versioning.ts
```diff
@@ -31,8 +35,20 @@ export async function generateVersioning(
...
+      id,
+      displayName,
+      urls,
+      supportsRanges,
+      supportedRangeStrategies,
+      api,
+    } = definition;
+
+    let strategies = supportedRangeStrategies ?? [];
+    if (api.getPinnedValue && !strategies.includes('pin')) {
```
viceice, line 48: did you verified that? shouldn't we fix the supported range strategies?

jamietanna: As far as I can tell, this is what the logic from https://github.com/renovatebot/renovate/pull/36261 does and the change we made, but we didn't consider the documentation changes

viceice: ok, but pin is not a range strategy. it should probably have its own section, eg: supports version pinning.

jamietanna: Is it not? 🤔 https://github.com/renovatebot/renovate/blob/2eb25c139676c0106f7ad06352842c163902079f/lib/types/versioning.ts#L1-L9

viceice: ok, validated again. you're right

viceice (approved):

PR #41434, opened 2026-02-23
refactor(cache): Convert package cache backends to classes

zharinov:
## Changes

Moves package cache backends into `impl/` as classes extending a shared `PackageCacheBase` abstract class. Adds `backend.ts` to manage backend lifecycle. `index.ts` now delegates to `backend.ts` for raw operations.

**Migration from `redis.ts` entry point:**

`get`, `set`, `destroy` signatures are unchanged. Only `init` changes:

```ts
// Old
import { init, get, set, destroy } from 'renovate/dist/util/cache/package/redis';
await init(url, prefix);

// New
import { init, get, set, destroy } from 'renovate/dist/util/cache/package/backend';
await init({ redisUrl: url, redisPrefix: prefix });
```

First step of incremental port of #39489.

## Context

## AI assistance disclosure


lib/util/cache/package/sqlite.ts
```diff
@@ -75,11 +76,6 @@ export class SqlitePackageCache {
     client
       .prepare('CREATE INDEX IF NOT EXISTS expiry ON package_cache (expiry)')
       .run();
-    client
-      .prepare(
-        'CREATE INDEX IF NOT EXISTS namespace_key ON package_cache (namespace, key)',
-      )
-      .run();
```
viceice, line 82: why this is removed 🤔

viceice (commented):
otherwise LGTM

lib/util/cache/package/sqlite.ts
zharinov: `(namespace, key)` is already a PK

viceice (approved):

PR #41454, opened 2026-02-24
refactor: consolidate previous groups into `kubernetes` monorepo

felipecrs:
## Changes

I added the `k8s-io` monorepo group in https://github.com/renovatebot/renovate/pull/41171 without realizing that the `kubernetes-go` group already existed and had almost the same package list included.

This PR renames that `k8s-io` monorepo group to `kubernetes`, which is a more consistent name considering the dependencies originate from https://github.com/kubernetes/kubernetes.

This PR also removes the old `kubernetes` group in favor of the new `kubernetes` monorepo group, therefore consolidating the two redundant groups into one.

## Context

## AI assistance disclosure

The public repository: https://github.com/felipecrs/renovate-repro-kubernetes-group


lib/config/presets/internal/group.preset.ts
```diff
@@ -356,15 +358,19 @@ const staticGroups = {
...
+          'k8s.io/externaljwt**',
+          'k8s.io/kms**',
           'k8s.io/kube-aggregator**',
           'k8s.io/kube-controller-manager**',
           'k8s.io/kube-proxy**',
           'k8s.io/kube-scheduler**',
           'k8s.io/kubectl**',
           'k8s.io/kubelet**',
-          'k8s.io/legacy-cloud-providers**',
+          'k8s.io/legacy-cloud-providers**', // This no longer receives updates, but should still be grouped if used
```
felipecrs: It had the same versioning as the other packages until it stopped receiving updates. And it's no longer part of the [official](https://github.com/kubernetes/kubernetes/blob/bebefd8382f3df4953afc76b5787a147deb0798f/staging/README.md) list.

Maybe I should remove this. Not sure. Any opinions?

viceice (commented):
I would prefer to do it the other way around and use the monorepo style

felipecrs:
> I would prefer to do it the other way around and use the monorepo style

However this `kubernetes-go` group existed for years. Doing that would be somewhat breaking. Renovate would probably forget or close existing `kubernetes-go` PRs.

@viceice, please let me know if you want that anyway. I personally don't mind.

lib/config/presets/internal/group.preset.ts
felipecrs: I think we can keep it.

felipecrs:
> I would prefer to do it the other way around and use the monorepo style

Never mind my last comment: I added a migration for it and this is now done.

viceice (commented):
can you test on a real repo? otherwise LGTM

felipecrs:
@viceice, done. All good:

- https://github.com/felipecrs/renovate-repro-kubernetes-group/pulls

felipecrs:
@viceice, after more thought, I decided to rename the `k8s-io` monorepo group into `kubernetes`, which I think is more consistent with the monorepo name itself (https://github.com/kubernetes/kubernetes), and also looks nicer. I've added a migrations.

Hope you are ok with the changes, please let me know otherwise, I can revert it.

lib/data/monorepo.json
```diff
@@ -425,7 +425,7 @@
       "https://github.com/junit-team/junit5",
       "https://github.com/junit-team/junit-framework"
     ],
-    "k8s-io": [
+    "kubernetes": [
```
viceice, line 436: this needs reorder, they should be in alphabetical order

viceice (approved):

PR #41455, opened 2026-02-24
feat(cache): Implement L1 and L2 for package cache

zharinov:
## Changes

Even without a persistent backend (Redis, SQLite, file), Renovate now deduplicates package cache lookups within a single run via an in-memory L1 layer. Previously, no-backend meant every `get()` returned `undefined` — now identical lookups are served from memory.

The L1 layer also sits in front of configured backends, avoiding redundant disk/network reads for keys already seen in the same process.

Adds `softReset()` to clear L1 between repositories without tearing down the backend connection.

Second step of incremental port of #39489, building on #41434.

## Context

## AI assistance disclosure


lib/util/cache/package/package-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+import { getMutex } from '../../mutex.ts';
+import { PackageCacheStats } from '../../stats.ts';
+import { PackageCacheBase } from './impl/base.ts';
+import { getTtlOverride } from './ttl.ts';
+import type { PackageCacheNamespace } from './types.ts';
+
+const DEFAULT_TIMEOUT_MS = 2 * 60 * 1000;
+
+export class PackageCache extends PackageCacheBase {
+  readonly memory = new Map<string, unknown>();
```
viceice: Could this be a new issue with too much memory usage? Especially the hosted app is already on limit on some oss projects 🤔 

@jamietanna WDYT? Should this have a configurable limit with LRU?

viceice (commented):
LGTM beside the memory resource 🤔

zharinov:
In current codebase, everything that goes to package cache, goes to the memory cache first IIRC. The LRU could be beneficial here, yes.

viceice (dismissed):

lib/util/cache/package/package-cache.ts
jamietanna: Yes, we'd need to have a configurable limit here

Do we know what level we'd want it to be for users who're i.e. running in GitHub Actions, or with ~2GB of memory? How much would we want to allocate here?

viceice:
conflicted

viceice (commented):
- conflicted
- needs a limit (LRU?, see conversation)

zharinov:
64Mb for L1 (mem) cache by default

PR #41462, opened 2026-02-24
fix(github): prevent PR cache sync from exhausting all pages

### LINKED ISSUE #40426: Initial seeding of PR Cache on a repository with many PRs can lead to the job running forever
As reported in #40076, a repository with _many_ closed Pull Requests could lead to Renovate taking significant time to first scan a repository, as it continues through many pages of data, and may eventually time out.

(In the report's case, there are ~90k PRs closed on that repository, and they have a 30m job execution time limit on the Mend-hosted infrastructure)

It appears that - although this was believed to be purely based on onboarding, is down to this being the PR Cache initialisation.

We should see if we can make this process more effective, for instance fetching more PRs at once (especially in the case a repo has _many_ PRs) or (_insert amazing idea here when someone has one_).

zharinov:
## Changes

On repos with many non-Renovate PRs, the sync path exhausts all pages because username filtering empties pages before the stop condition is checked, and `reconcile()` returns `true` for empty arrays.

**Core fix:** The stop condition now operates on the **unfiltered** page (before username filtering). Pages sorted `updated_at desc` — once the oldest item predates the cached watermark, all subsequent pages are stale.

**Key decisions:**

- **Watermark advances from unfiltered page data**, not just reconciled Renovate PRs. Without this, repos with only non-Renovate activity would re-scan the same pages every run.
- **Cutoff is snapshotted once before the loop.** `reconcile()` updates `lastModified` as it runs — reading it inside the loop creates a moving target.
- **Strict `<` comparison** (not `<=`). GitHub timestamps have second-level granularity, so ties at page boundaries must continue to the next page.
- **Derived cutoff from cached items** when `lastModified` is missing (older caches, `updateItem()` path).
- **MAX_SYNC_PAGES = 100** safety net for extremely stale caches.

## Context

- [x] This closes an existing Issue, Closes: #40426

## AI assistance disclosure

### DIFF
--- a/lib/modules/platform/github/pr.ts
+++ b/lib/modules/platform/github/pr.ts
@@ -152,6 +152,7 @@ export async function getPrCache(
         needNextPageFetch &&= !opts.paginate;
       }
 
+      // Safety net: cutoff-based stop should always fire first
       if (
         !isInitial &&
         needNextPageFetch &&
--- a/lib/modules/platform/github/pr.ts
+++ b/lib/modules/platform/github/pr.ts
@@ -67,7 +67,17 @@ export async function getPrCache(
 
   // Snapshot before the loop — reconcile() updates lastModified as it
   // processes items, so reading it inside the loop would create a moving target.
-  const lastModifiedRaw = prApiCache.getLastModified();
+  // If lastModified is missing but items exist (populated via updateItem()),
+  // derive cutoff from the newest cached item.
+  let lastModifiedRaw = prApiCache.getLastModified();
+  if (!lastModifiedRaw && !isInitial) {
+    const items = prApiCache.getItems();
+    for (const item of items) {
+      if (!lastModifiedRaw || item.updated_at > lastModifiedRaw) {
+        lastModifiedRaw = item.updated_at;
+      }
+    }
+  }
   const cutoffTime = lastModifiedRaw ? DateTime.fromISO(lastModifiedRaw) : null;
 
   try {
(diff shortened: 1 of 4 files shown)


lib/modules/platform/github/pr.ts
```diff
@@ -114,13 +129,27 @@ export async function getPrCache(
...
+      }
+
       needNextPageFetch = !!parseLinkHeader(linkHeader)?.next;
 
       if (pageIdx === 1) {
         needNextPageFetch &&= !opts.paginate;
       }
 
+      if (!isInitial && needNextPageFetch && pageIdx >= MAX_SYNC_PAGES) {
+        logger.warn(
```
jamietanna, line 162: I can see why we're thinking a `WARN`, but am wondering if this should be a `DEBUG` - I'm wondering what action we'd be expecting a human to take if they see this?

zharinov: It's just like a salient alternative to infinite loop, which shouldn't happen anymore though.

viceice: so add a comment that it should never happen?

viceice (dismissed):

viceice (approved):

lib/modules/platform/github/pr.ts
```diff
@@ -100,6 +117,17 @@ export async function getPrCache(
 
       let { body: page } = res;
 
+      if (!isInitial && cutoffTime && isNonEmptyArray(page)) {
+        // Advance watermark so next run doesn't re-scan these pages,
```
jamietanna, line 121: What does `watermark` mean in this context?

jamietanna (commented):
Looks great thank you 💜 

Once this is released, I'll ship it to the Dev Platform so I can confirm it helps out with https://github.com/renovatebot/renovate/discussions/40076

lib/modules/platform/github/pr.ts
```diff
@@ -12,6 +13,8 @@ import { ApiCache } from './api-cache.ts';
 import { coerceRestPr } from './common.ts';
 import type { ApiPageCache, GhPr, GhRestPr } from './types.ts';
 
+const MAX_SYNC_PAGES = 100;
```
jamietanna, line 16: Let's make this an experiemental config: https://github.com/renovatebot/renovate/issues/41485

PR #41534, opened 2026-02-26
feat(swift): update Package.resolved when updating Package.swift

bdolewski-intellias:
## Changes

Add `Package.resolved` lockfile support to the Swift Package Manager module. When Renovate updates dependency versions in `Package.swift`, it now also updates the corresponding pins in `Package.resolved` files.

### Background for non-iOS maintainers

Swift Package Manager (SPM) uses `Package.swift` as the manifest (like `package.json`) and `Package.resolved` as the lockfile (like `package-lock.json`). The lockfile is a JSON file that pins each dependency to an exact version and git commit SHA. Many iOS projects commit this file to ensure reproducible builds.

Until now, Renovate updated `Package.swift` but left `Package.resolved` untouched, making the lockfile stale after every Renovate PR.

### Approach: Direct JSON modification

Instead of running `swift package resolve` (which would require a Swift toolchain, a macOS runner, and potentially private repo authentication), this PR **directly modifies the JSON** in `Package.resolved`. For each updated dependency, it:

1. Matches the pin by URL (with `.git` suffix, trailing slash, and case normalization)
2. Resolves the new git commit SHA via Renovate's existing `getDigest()` datasource
3. Does a targeted regex replacement of `version` and `revision` values within the matched pin block, preserving Xcode's exact formatting (spaces before colons, indentation)

This is the same version-and-revision data that `swift package resolve` would produce, just without needing the toolchain.

### Known limitations

- **v3 `originHash`**: `Package.resolved` v3 (Xcode 15+) includes an `originHash` field computed over the full dependency graph. We cannot recompute it without the Swift toolchain, so it's preserved as-is. `xcodebuild -resolvePackageDependencies` will recompute it at build time — the version and revision pins we set will already be correct, so the resolve is effectively a no-op for those fields.
- **v1 format**: Not supported (Xcode 11, 2019). Gracefully skipped with a debug log.
- **Branch-pinned deps**: Pins with `"branch": "main"` instead of a version are naturally skipped.

## Context

This is not my first contribution to Renovate. I previously helped identify and fix a bug with the `v` prefix for SPM packages (https://github.com/renovatebot/renovate/discussions/28810) - that issue was rendering Renovate unusable for any SPM package with releases tagged as `v1.0.0` instead of `1.0.0`.

I'm an iOS developer, not a TypeScript/Java developer. I rely on Renovate to keep our iOS project's 50+ SPM dependencies up to date, and the stale `Package.resolved` after every PR has been a persistent pain point that required a manual GitHub Actions workaround.

## AI assistance disclosure

This entire PR was authored by **Claude Code (Opus 4.6)**. I designed the approach and guided the implementation through plan mode, but all TypeScript code, tests, and fixtures were generated by the AI. I reviewed the output and verified tests pass locally.

21 test cases covering: null returns for edge cases, single/multi pin updates, URL normalization (`.git`, trailing slash, case), git-tags/github-tags/gitlab-tags datasources, `newDigest` passthrough, `getDigest` failure handling, v2/v3 format support, formatting preservation, and already-up-to-date detection.

```
 Test Files  4 passed (4)
      Tests  42 passed (42)  # 21 new + 21 existing
```


dcespedes-ecw (approved):

lib/modules/manager/swift/artifacts.spec.ts
```diff
@@ -0,0 +1,585 @@
...
+    '../../datasource/index.ts',
+  );
+  return {
+    ...actual,
+    getDigest: vi.fn(),
+  };
+});
+
+const v2Fixture = Fixtures.get('Package.resolved.v2.json');
+const v3Fixture = Fixtures.get('Package.resolved.v3.json');
```
viceice: please inline the fixtures to the tests

lib/modules/manager/swift/artifacts.spec.ts
```diff
@@ -0,0 +1,585 @@
...
+
+describe('modules/manager/swift/artifacts', () => {
+  beforeEach(() => {
+    vi.resetAllMocks();
+  });
```
viceice:
```suggestion
```
not needed. AI tricked you

lib/modules/manager/swift/artifacts.ts
```diff
@@ -0,0 +1,241 @@
...
+}
+
+async function resolveCommitSha(
+  dep: {
+    newDigest?: string;
+    datasource?: string;
+    depName?: string;
+    packageName?: string;
+    registryUrls?: string[] | null;
+  },
```
viceice: please use proper types
```suggestion
  dep: PackageDependency,
```

lib/modules/manager/swift/artifacts.ts
```diff
@@ -0,0 +1,241 @@
...
+    .replace(regEx(/\/$/), '')
+    .toLowerCase();
+}
+
+function matchPinForDep(
+  dep: {
+    depName?: string;
+    datasource?: string;
+    registryUrls?: string[] | null;
+  },
```
viceice:
```suggestion
  dep: PackageDependency,
```

lib/modules/manager/swift/artifacts.ts
```diff
@@ -0,0 +1,241 @@
...
+  for (const resolvedFile of resolvedFiles) {
+    const content = await readLocalFile(resolvedFile, 'utf8');
+    if (!content) {
+      logger.debug({ resolvedFile }, 'swift: could not read Package.resolved');
+      continue;
+    }
+
+    let parsed: PackageResolvedJson;
+    try {
+      parsed = JSON.parse(content) as PackageResolvedJson;
```
viceice: would be nice to have a zod schema to safely parse this

bdolewski-intellias: Fixed

lib/modules/manager/swift/artifacts.ts
bdolewski-intellias: Fixed

lib/modules/manager/swift/artifacts.ts
bdolewski-intellias: Fixed

lib/modules/manager/swift/artifacts.spec.ts
bdolewski-intellias: Fixed

lib/modules/manager/swift/artifacts.spec.ts
bdolewski-intellias: Fixed

bdolewski-intellias:
@viceice thanks for the thorough review, really appreciate it. I've just pushed fixes to all the comments that you made.

viceice (approved):

pixlwave:
Noting that this resulted in a corrupt Package.resolved file for us, caused by prefixing the version with a `v`. I've opened a discussion here:

https://github.com/renovatebot/renovate/discussions/41780

pixlwave:
Also, I should say: Thank you very much for this PR, other than the issue above it appears to work great, even detecting multiple files in the same project 🎉

PR #41536, opened 2026-02-26
fix(crate): respect config.json in indices

lehmanju:
## Changes

- #41452.

This PR implements proper handling of cargo registries by respecting their config.json file. Each cargo registry serves this file at the root. It contains base urls for downloading and optionally for the web api.

https://doc.rust-lang.org/cargo/reference/registry-index.html

The change implements this behavior by always consulting `config.json` to retrieve the correct base urls. Additionally, it removes most of the special treatment for "crates.io" since it is just a normal cargo registry.

A potentially breaking change is, that the user-defined registry "https://crates.io" does not work anymore since it has been replace with "sparse+https://index.crates.io". For most users this shouldn't be a problem, since they just use the default registry. A possible solution is to automatically map the old url to the new sparse version.

Also, the `registryUrl` field for a detected dependency changes as can be seen in the test files.

## Context

## AI assistance disclosure

### DIFF
--- a/lib/modules/datasource/crate/index.spec.ts
+++ b/lib/modules/datasource/crate/index.spec.ts
@@ -21,12 +21,24 @@ import type { RegistryInfo } from './types.ts';
 vi.mock('simple-git');
 const simpleGit = vi.mocked(_simpleGit);
 
-const API_BASE_URL = CrateDatasource.CRATES_IO_API_BASE_URL;
-
-const baseUrl =
-  'https://raw.githubusercontent.com/rust-lang/crates.io-index/master/';
+const API_BASE_URL = 'https://crates.io';
+const DL_BASE_URL = 'https://static.crates.io/crates';
+const CRATES_IO_REGISTRY_URL = 'sparse+https://index.crates.io/';
+const CRATES_IO_REGISTRY_URL_PARSED = 'https://index.crates.io/';
 
 const datasource = CrateDatasource.id;
+
+const cratesIoConfig = {
+  dl: DL_BASE_URL,
+  api: API_BASE_URL,
+};
+
+function mockCratesIoConfig(): void {
+  httpMock
+    .scope(CRATES_IO_REGISTRY_URL_PARSED)
+    .get('/config.json')
+    .reply(200, cratesIoConfig);
+}
 
 function setupGitMocks(delayMs?: number): {
   mockClone: MockedFunction<SimpleGit['clone']>;
@@ -74,7 +86,7 @@ function setupErrorGitMock(): {
 function mockCratesApiCallFor(crateName: string, response?: httpMock.Body) {
   httpMock
     .scope(API_BASE_URL)
-    .get(`/crates/${crateName}?include=`)
+    .get(`/api/v1/crates/${crateName}?include=`)
     .reply(response ? 200 : 404, response);
 }
 
@@ -132,7 +144,10 @@ describe('modules/datasource/crate/index', () => {
 
     it('returns null for missing registry url', async () => {
       // FIXME: should not call default registry?
-      httpMock.scope(baseUrl).get('/no/n_/non_existent_crate').reply(404, {});
+      httpMock
+        .scope(CRATES_IO_REGISTRY_URL_PARSED)
+        .get('/no/n_/non_existent_crate')
+        .reply(404, {});
       expect(
         await getPkgReleases({
           datasource,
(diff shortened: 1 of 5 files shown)


lib/modules/datasource/crate/schema.ts
```diff
@@ -1,6 +1,11 @@
 import { z } from 'zod/v3';
 import { MaybeTimestamp } from '../../../util/timestamp.ts';
 
+export const RegistryConfigSchema = z.object({
+  dl: z.string(),
+  api: z.string().optional(),
+});
```
viceice, line 7:
```suggestion
});
export type RegistryConfigSchema = z.infer<typeof RegistryConfigSchema>;
```

to be usable as type

lib/modules/datasource/crate/index.ts
```diff
@@ -153,25 +148,77 @@ export class CrateDatasource extends Datasource {
...
   }
 
+  /**
+   * Fetches the registry's config.json which provides the API base URL
+   * and download URL template.
+   * See: https://doc.rust-lang.org/cargo/reference/registry-index.html#index-configuration
+   */
+  private async fetchRegistryConfig(
+    info: RegistryInfo,
+  ): Promise<{ dl: string; api?: string } | null> {
```
viceice:
```suggestion
  ): Promise<RegistryConfigSchema | null> {
```

lib/modules/datasource/crate/index.ts
```diff
@@ -153,25 +148,77 @@ export class CrateDatasource extends Datasource {
...
+    const cacheKey = `crate-datasource/registry-config/${info.rawUrl}`;
+    const cached = memCache.get<{ dl: string; api?: string }>(cacheKey);
```
viceice:
```suggestion
    const cached = memCache.get<RegistryConfigSchema>(cacheKey);
```

lib/modules/datasource/crate/index.ts
```diff
@@ -153,25 +148,77 @@ export class CrateDatasource extends Datasource {
...
+      }
+    } else if (info.isSparse) {
+      try {
+        const configUrl = joinUrlParts(info.rawUrl, 'config.json');
+        const { body } = await this.http.getJsonUnchecked<unknown>(configUrl);
+        const parsed = RegistryConfigSchema.safeParse(body);
+        if (parsed.success) {
+          memCache.set(cacheKey, parsed.data);
+          return parsed.data;
+        }
```
viceice:
```suggestion
        const { body } = await this.http.getJson(configUrl, RegistryConfigSchema);
        memCache.set(cacheKey, parsed.data);
        returnbody;
```
simplify

lib/modules/datasource/crate/index.ts
```diff
@@ -153,25 +148,77 @@ export class CrateDatasource extends Datasource {
...
+    const cached = memCache.get<{ dl: string; api?: string }>(cacheKey);
+    if (cached) {
+      return cached;
+    }
+
+    if (info.clonePath) {
+      try {
+        const configPath = upath.join(info.clonePath, 'config.json');
+        const content = await readCacheFile(configPath, 'utf8');
+        const parsed = RegistryConfigSchema.safeParse(JSON.parse(content));
```
viceice:
```suggestion
        const parsed = Json.pipe(RegistryConfigSchema).safeParse(content);
```

lib/modules/datasource/crate/index.ts
```diff
@@ -442,12 +484,13 @@ export class CrateDatasource extends Datasource {
...
-    registryUrl: string | undefined,
-  ): boolean {
-    // We only cache public releases, we don't want to cache private
-    // cloned data between runs.
-    return registryUrl === 'https://crates.io';
+  private static isCratesIo(registryUrl: string | undefined): boolean {
+    if (!registryUrl) {
+      return false;
+    }
+    const normalized = registryUrl.replace(/^sparse\+/, '');
```
viceice:
```suggestion
    const normalized = registryUrl.replace(regEx(/^sparse\+/), '');
```
use our regex util

lib/modules/datasource/crate/index.spec.ts
```diff
@@ -16,18 +16,29 @@ import * as memCache from '../../../util/cache/memory/index.ts';
...
-  'https://raw.githubusercontent.com/rust-lang/crates.io-index/master/';
+const API_BASE_URL = 'https://crates.io';
+const DL_BASE_URL = 'https://static.crates.io/crates';
+const CRATES_IO_REGISTRY_URL = 'sparse+https://index.crates.io/';
+const CRATES_IO_REGISTRY_URL_PARSED = 'https://index.crates.io/';
 
 const datasource = CrateDatasource.id;
 
+const cratesIoConfig = {
+  dl: DL_BASE_URL,
```
viceice, line 32: use the type here, so typescript will detect any mismatch in future

viceice (commented):
nearly LGTM

@zharinov can you check the caching thing? maybe we can use one of our decorator functions to simplify?

lib/modules/datasource/crate/index.spec.ts
lehmanju: done

zharinov (approved):
Memory cache doesn't have decorator, and I think it's not worth it anyways

viceice (approved):

altaiezior:
@lehmanju when using a custom registry the getCrateMetadata method fails as it tries to query the metadata using /api/v1 but in registries such as of nexus and artifact-keeper the metadata API is not available, it is only part of crates.io and hence can be one of the reasons why crates.io was earlier being used here.

lehmanju:
Yes, I know. Thats why I have merged a logging fix recently. The /api/v1 endpoint is still part of the spec but as you say not implemented for most custom implementations. It should fail silently with the latest version.

altaiezior:
@lehmanju do you think `failed to download crate metadata` should also be suppressed as it would otherwise create confusion of the call being failure

Secondly can we somehow make a quick check if the repository supports metadata API, cache that for repository and don't waste our calls if it is not supported.

In metadata we usually fetch the source repository / homepage currently, so when using custom registry a user would miss these data points, do you think falling back to crates.io for metadata would be a better choice.

We can have feature controls via configurations

lehmanju:
I don't understand, the warning suppression is already merged (#41601) and afaik there is no way of checking the implemented registry features.

altaiezior:
> I don't understand, the warning suppression is already merged (#41601) and afaik there is no way of checking the implemented registry features.

I just thought if we can save some network calls, usually when I have to debug why a dependency is not getting updated I use the debug mode and I see these failure logs before I reliased may be they don't even matter.

PR #41537, opened 2026-02-26
fix(cargo): skip `--precise` for range-bumped crates to avoid stale `Cargo.lock` errors

### LINKED ISSUE #39137: Renovate fails to update an unpinned Cargo version, when attempting to update it

As noted in https://github.com/renovatebot/renovate/discussions/38778, we're seeing a case where:

- Cargo specificies an imprecise version number (`0.12`) with a lockfile locked to a specific version (`0.12.23`)
- Renovate detects this correctly: 
  ```json
            {
              "currentValue": "0.12",
              "datasource": "crate",
              "depName": "reqwest",
              "depType": "dependencies",
              "lockedVersion": "0.12.23",
              "updates": [
                {
                  "newVersion": "0.12.24",
                  "newValue": "0.12.24",
                }
              ],
              "currentVersion": "0.12.23",
              "isSingleVersion": false,
              "fixedVersion": "0.12.23"
            }

  ```
- When Renovate attempts to upgrade this version, it uses the `lockedVersion`/`fixedVersion` instead of the `currentValue`, resulting in the `cargo` CLI failing

Interestingly, running:

```
$ install-tool rust 1.90.0
$ cargo --version
cargo 1.90.0 (840b83a10 2025-07-30)
$ cargo update --config net.git-fetch-with-cli=true --manifest-path subdir2/Cargo.toml --package reqwest@0.12.23 --precise 0.12.24
    Updating crates.io index
    Updating reqwest v0.12.23 -> v0.12.24
      Adding subdir v0.1.0 (/app/subdir)
      Adding subdir2 v0.1.0 (/app/subdir2)
note: pass `--verbose` to see 32 unchanged dependencies behind latest
```

Seems to work for me - but seems to fail when `renovate` runs it.

just-in-chang:
## Changes

Skips updating a specific entry with the `--precise` flag for Crates where the `Cargo.toml` range was bumped, delegating the resolution to the later update call with the `--workspace` flag. 

## Context

- [x] This closes an existing Issue, Closes: #39137

I put a bit of thought into this and iterated over a few possible solutions:
1. Omitting the `@lockedVersion` from `--package`. It keeps the `--precise` but doesn't work when there are multiple versions of the same crate in the `Cargo.lock`.
2. Proactively reading `Cargo.lock` to skip crates already at their target version. This introduced unnecessary complexity and duplicated the existing retry logic in `artifacts.ts`.
3. [Current Approach] Skips `--precise` entirely for range-bumped crates, letting the later `--workspace` update command re-resolve them. This avoids both the stale version and ambiguity issues cleanly.

## AI assistance disclosure

Claude Code was used to help generate the unit and real repository tests.

The public repository: https://github.com/just-in-chang/renovate-39137-tests

Please read the `README.md` there for more details about the tests.

### DIFF
--- a/lib/modules/manager/cargo/artifacts.ts
+++ b/lib/modules/manager/cargo/artifacts.ts
@@ -51,6 +51,12 @@ async function cargoUpdatePrecise(
   // First update individual dependencies to their `newVersion`. Necessary when
   // using the `update-lockfile` rangeStrategy which doesn't touch Cargo.toml.
   for (const dep of updatedDeps) {
+    // Range bumped for an existing crate. The old lockedVersion is gone from Cargo's
+    // dependency graph, so let the --workspace update at the end re-resolve it instead.
+    if (dep.currentValue && dep.newValue && dep.currentValue !== dep.newValue) {
+      continue;
+    }
+
     cmds.push(
       `cargo update --config net.git-fetch-with-cli=true` +
         ` --manifest-path ${quote(manifestPath)}` +
--- a/lib/modules/manager/cargo/artifacts.ts
+++ b/lib/modules/manager/cargo/artifacts.ts
@@ -51,8 +51,9 @@ async function cargoUpdatePrecise(
   // First update individual dependencies to their `newVersion`. Necessary when
   // using the `update-lockfile` rangeStrategy which doesn't touch Cargo.toml.
   for (const dep of updatedDeps) {
-    // Range bumped for an existing crate. The old lockedVersion is gone from Cargo's
-    // dependency graph, so let the --workspace update at the end re-resolve it instead.
+    // If the range is bumped in Cargo.toml, the old lockedVersion may no longer
+    // exist in Cargo's dependency graph, so let the --workspace update at the
+    // end re-resolve it instead.
     if (dep.currentValue && dep.newValue && dep.currentValue !== dep.newValue) {
       continue;
     }
--- a/lib/modules/manager/cargo/artifacts.spec.ts
+++ b/lib/modules/manager/cargo/artifacts.spec.ts
@@ -160,6 +160,89 @@ describe('modules/manager/cargo/artifacts', () => {
     ]);
   });
 
+  it('skips precise update when manifest range has changed', async () => {
+    fs.findLocalSiblingOrParent.mockResolvedValueOnce('Cargo.lock');
+    fs.readLocalFile.mockResolvedValueOnce('Old Cargo.lock');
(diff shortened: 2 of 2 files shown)


jamietanna:
Would you mind also checking that it works against https://github.com/tommyjcarpenter/rust-repro-example-38778 ?

lib/modules/manager/cargo/artifacts.ts
```diff
@@ -51,6 +51,13 @@ async function cargoUpdatePrecise(
   // First update individual dependencies to their `newVersion`. Necessary when
   // using the `update-lockfile` rangeStrategy which doesn't touch Cargo.toml.
   for (const dep of updatedDeps) {
+    // If the range is bumped in Cargo.toml, the old lockedVersion may no longer
+    // exist in Cargo's dependency graph, so let the --workspace update at the
+    // end re-resolve it instead.
```
viceice, line 56: Wouldn't this mean the real locked version can be higher than the range renovate bumped to?

Maybe we should extract the locked version in extract, so bump isn't done at all, because the resolved value is already updated?

viceice (commented):
Sorry, i don't know how cargo works for resolving.

lib/modules/manager/cargo/artifacts.ts
just-in-chang: Thanks for the review!

> Wouldn't this mean the real locked version can be higher than the range renovate bumped to?

The version updated by the `--workspace` update at the end will always satisfy the range bounds used in `Cargo.toml`. It just might resolve to a higher version than what is specified in `newVersion` specifically. It's also consistent with the existing behavior by calling the `--workspace` update at the end.

> Maybe we should extract the locked version in extract, so bump isn't done at all, because the resolved value is already updated?

If I understand correctly, the extraction is already happening! The bump is a separate concern though from when new version exceeds the current range or `rangeStrategy` is configured to do so (i.e. setting it to `bump`).

just-in-chang:
> Would you mind also checking that it works against https://github.com/tommyjcarpenter/rust-repro-example-38778 ?

@jamietanna Updated my public test repo to include it in the test cases and also updated the test suite to be more comprehensive. Please take a look!

For the new E2E tests, I'm spinning up two Docker containers containing a few Rust fixtures per test: one with the upstream `main` branch and one with my Renovate fork. We utilize Docker Compose and Gitea along with the two Docker containers because Renovate expects a real Git platform. We then show that the upstream Renovate fails and my fork containing the fix succeeds.

viceice (approved):

PR #41588, opened 2026-03-02
fix(gomod): write full pseudo-version instead of bare hash on digest updates

qixiang:
## Changes

When a Go dependency uses a pseudo-version (v0.0.0-timestamp-hash) and Renovate detects a newer pseudo-version, updateType is set to 'digest' via the gomod special-case in lookup/index.ts. The digest update code path was then writing only the 12-character bare commit hash, producing invalid go.mod syntax:
```
  - knative.dev/pkg v0.0.0-20250312035536-b7bbf4be5dbd
  + knative.dev/pkg 4a022ed9999a
```
Historical cause:

In 2018, the Go module proxy did not yet exist. Renovate queried the GitHub API for the latest commit SHA and had no way to obtain the correct commit timestamp, so it fabricated a pseudo-version using the wall-clock time when Renovate ran. The result was syntactically valid but semantically wrong. Go 1.11 and 1.12 did not validate pseudo-version timestamps, and `go get` would normalise the fabricated version to the correct one anyway, so this worked in practice.

Go 1.13 (2019) introduced strict pseudo-version timestamp validation, rejecting any pseudo-version whose timestamp does not match the actual commit time. Renovate's fix (31ce47297) was to drop the fabricated timestamp entirely and write just the bare commit hash, then rely on `go get` to expand it into a canonical pseudo-version. However, when the `go get` command or other subsequent operations fail, Renovate may leave the go.mod file containing dependencies pinned to bare commit hashes.

Why the fix is now safe:

Since the 2024 goproxy datasource changes, both newValue and newDigest are extracted from the same proxy version string via pseudoVersionToRelease(). They always reference the same commit, so writing newValue directly produces correct, syntactically valid go.mod output without depending on `go get` to clean it up afterward.

This should fix the issue mentioned in discussion - #33596 

## Context

- [] This closes an existing Issue, Closes: # 

## AI assistance disclosure

The public repository: <none>

### DIFF
--- a/lib/modules/manager/gomod/update.spec.ts
+++ b/lib/modules/manager/gomod/update.spec.ts
@@ -370,6 +370,36 @@ describe('modules/manager/gomod/update', () => {
       expect(res).toEqual(gomod2);
     });
 
+    it('updates pseudo-version with digest updateType', () => {
+      const fileContent = codeBlock`
+        module example.com/test
+        require (
+          knative.dev/pkg v0.0.0-20250312035536-b7bbf4be5dbd
+          k8s.io/utils v0.0.0-20251002143259-bc988d571ff4
+        )
+      `;
+      const upgrade = {
+        depName: 'knative.dev/pkg',
+        managerData: { lineNumber: 2, multiLine: true },
+        updateType: 'digest' as const,
+        currentValue: 'v0.0.0-20250312035536-b7bbf4be5dbd',
+        currentDigest: 'b7bbf4be5dbd',
+        newValue: 'v0.0.0-20260120122510-4a022ed9999a',
+        newDigest: '4a022ed9999a',
+        depType: 'require',
+      };
+      const res = updateDependency({
+        fileContent,
+        packageFile: 'go.mod',
+        upgrade,
+      });
+      // Should write the full pseudo-version, not just the digest
+      expect(res).toContain(
+        'knative.dev/pkg v0.0.0-20260120122510-4a022ed9999a',
+      );
+      expect(res).not.toContain('knative.dev/pkg 4a022ed9999a');
+    });
+
     it('handles multiline mismatch', () => {
       const upgrade = {
         depName: 'github.com/fatih/color',
--- a/lib/modules/manager/gomod/update.ts
+++ b/lib/modules/manager/gomod/update.ts
@@ -80,22 +80,41 @@ export function updateDependency({
     }
     let newLine: string;
     if (upgrade.updateType === 'digest') {
-      const newDigestRightSized = upgrade.newDigest!.substring(
-        0,
-        upgrade.currentDigest!.length,
-      );
-      if (lineToChange.includes(newDigestRightSized)) {
-        return fileContent;
+      // Since the 2024 goproxy datasource changes, newValue and newDigest are
+      // both extracted from the same proxy version string and always reference
+      // the same commit, so newValue can be written directly for pseudo-versions.
+      if (upgrade.newValue?.startsWith('v0.0.0-')) {
+        logger.debug(
+          { depName: currentName, lineToChange, newValue: upgrade.newValue },
+          'gomod: updating pseudo-version digest',
+        );
+        newLine = lineToChange.replace(
+          // TODO: can be undefined? (#22198)
+          updateLineExp!,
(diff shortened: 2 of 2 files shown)


jamietanna:
I think we usually recommend folks use `postUpdateOptions: goModTidy` for this, but getting it sorted without needing that is also pretty useful!

lib/modules/manager/gomod/update.ts
```diff
@@ -80,22 +80,41 @@ export function updateDependency({
...
+          'gomod: updating pseudo-version digest',
+        );
+        newLine = lineToChange.replace(
+          // TODO: can be undefined? (#22198)
+          updateLineExp!,
+          `$<depPart>$<divider>${upgrade.newValue}`,
+        );
+      } else {
+        // Defensive fallback for non-pseudo-version digest updates.
+        // Unreachable for gomod in practice: currentDigest is only extracted
```
viceice, line 98: isn't this reachable when you bypass go proxy

qixiang: When the Go proxy is bypassed, Renovate uses `GoDirectDatasource` which fetches git tags from the source repo directly. In this case, it returns raw git tags and have no knowledge of pseudo versions, pseudo versions are generated by the Go module proxy and don't exist as real git tags.

For the else branch to be reachable, two conditions must hold simultaneously:
1. upgrade.updateType === 'digest'
2. upgrade.newValue does not start with `v0.0.0-`

https://github.com/renovatebot/renovate/blob/51ef4acd87adcadc1ef8350df2d76c92afaecc65/lib/workers/repository/process/lookup/index.ts#L511-L518

The gomod special override that sets `updateType = 'digest'` requires `update.newValue?.startsWith('v0.0.0-')` as a pre-condition. In case of using `GoDirectDatasource`, `updateType` is not `digest` and this else clause is unreachable.

My understanding of the Renovate codebase is limited, so it's possible that I may have overlooked something.

lib/modules/manager/gomod/update.spec.ts
```diff
@@ -295,6 +295,33 @@ describe('modules/manager/gomod/update', () => {
       expect(res).toEqual(gomod2);
     });
 
+    it('updates pseudo-version with digest updateType', () => {
+      const fileContent = `module example.com/test
```
RahulGautamSingh: use `codeBlock` utility

lib/modules/manager/gomod/update.spec.ts
```diff
@@ -295,6 +295,33 @@ describe('modules/manager/gomod/update', () => {
...
+
+require (
+	knative.dev/pkg v0.0.0-20250312035536-b7bbf4be5dbd
+	k8s.io/utils v0.0.0-20251002143259-bc988d571ff4
+)
+`;
+      const upgrade = {
+        depName: 'knative.dev/pkg',
+        managerData: { lineNumber: 3, multiLine: true },
+        updateType: 'digest' as UpdateType,
```
RahulGautamSingh: Don't think this assertion is needed. 

IF TS complaines give this object the needed type or use `satisfies`

RahulGautamSingh (changes requested):

lib/modules/manager/gomod/update.spec.ts
qixiang: Done

lib/modules/manager/gomod/update.spec.ts
qixiang: Done

viceice (dismissed):

RahulGautamSingh (approved):

jamietanna (dismissed):

jamietanna:
@qixiang can you please sign the CLA? It might mean you need to amend your Git authorship email, per https://github.com/renovatebot/renovate/pull/41588#issuecomment-4221318442

viceice (commented):
needs test fix and cla signing (maybe the commit author / editor was wrong, so fix it and force-push)

qixiang:
@jamietanna @viceice My GitHub-associated email has changed, so I need to re-sign the CLA and amend my commit to update the email address. I force-pushed the branch after rebasing the change and fixed the test failure.

jamietanna (approved):

PR #41589, opened 2026-03-02
Revert "fix(crate): respect config.json in indices"

lehmanju:
Reverts renovatebot/renovate#41536

Sorry, this doesn't work. The index is not hosted at the "dl" endpoint as I assumed but at the baseUrl directly. I do not know how I could oversee such a crucial detail. Sometimes I am just blind, I guess.


lehmanju:
The config.json is not actually needed to download the crate releases, only for the web api (crate metadata). I will provide a new PR that will implement the relevant changes correctly, but for now I think it is better to revert and restore functionality.

lehmanju:
wait, let me check sth first.

lehmanju:
Nvm, just panicked. I will fix the logging though, because now every registry that doesn't support the metadata API will emit a warning, but debug should be enough, since it can be silently ignored. Also additional logging for release requests may be helpful.

PR #41633, opened 2026-03-03
fix(datasource/github): skip GraphQL requests for non-GitHub registry URLs

miketheman:
## Changes

When a non-GitHub registryUrl (e.g. https://pypi.org/pypi/) reaches the GitHub GraphQL datasource fetcher — via user config applying registryUrls or defaultRegistryUrls to github-tags/github-releases datasources — the fetcher blindly constructs a GraphQL endpoint from it:

```
  getApiBaseUrl('https://pypi.org/pypi/')
  → 'https://pypi.org/pypi/api/v3/'  (assumes GHE)
  → 'https://pypi.org/pypi/api/'     (strip v3)
  → POST /pypi/api/graphql            (append graphql)
```

This produces always-failing POST requests to hosts like PyPI that have no GraphQL endpoint (HTTP 405).

Add an `isGithubHost()` check in `GithubGraphqlDatasourceFetcher.query()` that validates the registryUrl before issuing requests. Non-GitHub URLs now return empty results immediately with a debug log.

## Context

## AI assistance disclosure

- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation). This PR was written primarily by Claude Code with my guidance and review.

### DIFF
--- a/lib/util/github/graphql/datasource-fetcher.spec.ts
+++ b/lib/util/github/graphql/datasource-fetcher.spec.ts
@@ -116,6 +116,16 @@ describe('util/github/graphql/datasource-fetcher', () => {
       http = new GithubHttp();
     });
 
+    it('returns empty result for non-GitHub registryUrl', async () => {
+      const res = await Datasource.query(
+        { packageName: 'foo/bar', registryUrl: 'https://pypi.org/pypi/' },
+        http,
+        adapter,
+      );
+
+      expect(res).toEqual([]);
+    });
+
     it('can perform query and receive result', async () => {
       httpMock
         .scope('https://api.github.com/')
--- a/lib/util/github/graphql/datasource-fetcher.ts
+++ b/lib/util/github/graphql/datasource-fetcher.ts
@@ -11,7 +11,8 @@ import type {
   GithubHttpOptions,
 } from '../../http/github.ts';
 import type { HttpResponse } from '../../http/types.ts';
-import { getApiBaseUrl } from '../url.ts';
+import { parseUrl } from '../../url.ts';
+import { getApiBaseUrl, getSourceUrlBase } from '../url.ts';
 import { GithubGraphqlMemoryCacheStrategy } from './cache-strategies/memory-cache-strategy.ts';
 import { GithubGraphqlPackageCacheStrategy } from './cache-strategies/package-cache-strategy.ts';
 import type {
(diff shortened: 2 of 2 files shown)


viceice (changes requested):
I don't like this solution. the current behavior is intended and it's clearly a user error if that happens.

we can add a warning when the graphql returns a 404.

viceice:
please don't remove PR template parts

miketheman:
> I don't like this solution. the current behavior is intended and it's clearly a user error if that happens.

Can you help me understand this? The "user" here is the renovate service, and it's mis-issuing API calls to a service that doesn't support them, wasting resources.

> we can add a warning when the graphql returns a 404.

I'm not sure what that does or helps - since in this case, the response is a 405 Method Not Allowed, not 404.

Happy to take any guidance to support the effort in removing these extraneous calls.


> please don't remove PR template parts

Sorry about that - I hate seeing checklist summaries on issues that appear incomplete, like "4 of 12 tasks". If removing some of the checks is undesirable, maybe add that in the PR template or contributing guidelines?

viceice:
> > I don't like this solution. the current behavior is intended and it's clearly a user error if that happens.
> 
> Can you help me understand this? The "user" here is the renovate service, and it's mis-issuing API calls to a service that doesn't support them, wasting resources.

the user is the person who configured the registry urls in a wrong way.

> > we can add a warning when the graphql returns a 404.
> 
> I'm not sure what that does or helps - since in this case, the response is a 405 Method Not Allowed, not 404.
> 
> Happy to take any guidance to support the effort in removing these extraneous calls.

405 is also ok to me to log a warning here.

> > please don't remove PR template parts
> 
> Sorry about that - I hate seeing checklist summaries on issues that appear incomplete, like "4 of 12 tasks". If removing some of the checks is undesirable, maybe add that in the PR template or contributing guidelines?

those are not tasks here and some can change in lifetime of PR.

miketheman:
> the user is the person who configured the registry urls in a wrong way.
 
As we don't operate the renovate managed service, can you assist in helping determine who the "user" is on your end that could correct this?

> 405 is also ok to me to log a warning here.

That's fine by me, I'm also fine with instead blocking these requests at the PyPI firewall edge, but it would be better and more beneficial to everyone to not issue requests that we know for certain will fail.

As it is, Renovate has a non-trivial failed request rate for PyPI - of the uncached requests over the past month:

- 52% were a 301 response (we redirect callers with missing trailing slashes, we could definitely stop that, and simply 404)
- 29% were 404
- 15% were 405 (since we don't implement a `POST /pypi/api/graphql` or `POST /pypi/api/graphql/` endpoint), and triggered this investigation
- The remaining percentage are 200 responses.

Hopefully this breakdown paints the picture that 95% of all traffic to PyPI generated by renovate user agents is probably useless and wasteful, and there are means to correct this and thus reduce your overall impact on a free service - as well as your own free service not needing to issue calls and spend bandwidth on things that are not going to work.

viceice:
`user` -> the people who configure the `renovate.json` with the wrong registry urls.

viceice:
please create a minimal reproduction, the custom registry urls can be dummy so they intentionally fail

jamietanna:
@RahulGautamSingh will pick up getting a minimal repro together - I'm raising an Issue to cover the work to fix it, too

jamietanna:
OK, we've added a config warning in https://github.com/renovatebot/renovate/pull/41682 to warn folks that they shouldn't be doing this

Based on the DMs I've had with Mike, we've at least got folks using Renovate 41.x, 42.x and 43.x (and there may be older users) - so even if we backport this, it's not certain if they'll upgrade and then address the warning - unfortunately it's a case where we're going to have to wait for folks to do the right thing

I wouldn't disagree with PyPI's choice to block these requests at the edge/firewall - it's more effort for y'all, which we're sorry about, but hopefully will limit the impact to your servers in the meantime

PR #41682, opened 2026-03-05
fix(config): warn if `registryUrls` is set at top-level in config

RahulGautamSingh:
## Context

- [x] This closes an existing Issue, Closes: #41663 

## AI assistance disclosure

The public repository: <URL>


viceice (changes requested):
this is not what I had in my mind to fix the issue. it's nearly same as the original PR, where I told that's the wrong solution.

there should be config validation warning instead to not use registry urls at top level config. 

later it should be an error because it will almost always cause errors

lib/config/validation.ts
```diff
@@ -237,6 +237,13 @@ export async function validateConfig(
         });
       }
     }
+    if (key === 'registryUrls' && !parentPath && isNonEmptyArray(val)) {
+      warnings.push({
+        topic: 'Configuration Warning',
+        message:
+          'Setting `registryUrls` at the top level of your config has no effect. Use `registryUrls` inside `packageRules` to apply it to specific packages.',
```
RahulGautamSingh: @jamietanna  Any suggestions on what the warning should be?

jamietanna: Does it have no effect - I thought it did, and breaks things?

jamietanna (commented):
We also need to warn if `defaultRegistryUrls`

lib/config/validation.ts
viceice: first step is to warn. maybe open a discussion and link to it for futher help? Maybe some users are currently depend on this and we want to help them to properly reconfigure it?

viceice (commented):
add same for `defaultRegistryUrls`

viceice (commented):
why duplicate tests?

RahulGautamSingh:
> why duplicate tests?

one for global config test another for repo config

viceice (approved):
ok, I don't see a benefit because the check is always done but ok

PR #41705, opened 2026-03-05
feat: drop redundant `(major|minor|patch)` PR suffixes

felipecrs:
## Changes

The PR title for group updates have a special handling: if `separateMajorMinor: true`, the PR title will have a `(major)` suffix for major updates.

This is different than non-group updates where no such suffix is added, even under the same circumstances.

Normally, when the new versions inside a group are all the same, Renovate will include the version in the PR title, like `Update group name to v3 (major)`.

However, when the new versions inside a group are not all the same, Renovate will not include the new version to the PR title, like `Update group name (major)`. In that case, it's important to add `(major)` suffix to distinguish between other updates like minor or patch.

This PR changes it so that the `(major)` (or `(minor)` and `(patch)`) suffix is only added to the PR title when the PR title doesn't yet contain a version.

This makes PR titles more consistent between non-group and group updates when possible, while still keeping the suffix when useful. Also saves a few characters for those who care.

image

## Context

## AI assistance disclosure

The public repository: https://github.com/felipecrs/renovate-repro-remove-major-suffix-group/pulls


lib/workers/repository/updates/generate.ts
```diff
@@ -141,18 +141,23 @@ function compilePrTitle(
...
-          : '';
-      upgrade.prTitle +=
-        upgrade.updateType === 'minor' && upgrade.separateMinorPatch
-          ? ' (minor)'
-          : '';
-      upgrade.prTitle +=
-        upgrade.updateType === 'patch' && upgrade.separateMinorPatch
-          ? ' (patch)'
-          : '';
+      const hasVersionInTitle = Boolean(
```
viceice:
```suggestion
      const hasVersionInTitle = isNonEmptyStringAndNotWhitespace(
```

viceice (dismissed):

secustor (commented):
This will be breaking for some people.
I have seen users parsing these PR titles for `(major)`, if labels can not be used.

felipecrs:
> This will be breaking for some people. I have seen users parsing these PR titles for `(major)`, if labels can not be used.

@secustor interesting, even though `(major)` is only there for groups and not individual updates?

My point is: if they are already able to parse individual updates without `(major)` suffix (probably by extracting the version from the PR title and checking if only contains the _major_ component like `v3`), then that logic should also handle this new case... without having to refactor it, I suppose.

secustor:
I think it is not blocking as we are not considering PR title ( AFAIK ) as public API. @jamietanna WDYT?

lib/workers/repository/updates/generate.spec.ts
```diff
@@ -1184,6 +1187,48 @@ describe('workers/repository/updates/generate', () => {
...
+      const minorWithVersionInTitle = [
+        {
+          ...requiredDefaultOptions,
+          manager: 'some-manager',
+          depName: 'some-dep',
+          branchName: 'some-branch',
+          prTitle: 'some-title',
+          newValue: '1.2.0',
+          newVersion: '1.2.0',
+          commitMessageExtra: 'to 1.2.0',
```
RahulGautamSingh, line 1222: pr title says only major suffix is being removed, but this is minor.

lib/workers/repository/updates/generate.ts
```diff
@@ -141,18 +146,23 @@ function compilePrTitle(
...
-      upgrade.prTitle +=
-        upgrade.updateType === 'minor' && upgrade.separateMinorPatch
-          ? ' (minor)'
-          : '';
-      upgrade.prTitle +=
-        upgrade.updateType === 'patch' && upgrade.separateMinorPatch
-          ? ' (patch)'
-          : '';
+      const hasVersionInTitle = isNonEmptyStringAndNotWhitespace(
+        template.compile(upgrade.commitMessageExtra ?? '', upgrade),
```
RahulGautamSingh, line 145: this is user customisable, we should be more thorough in checking that this is a version like string

lib/workers/repository/updates/generate.spec.ts
felipecrs: AFAIK there is no such thing as a `(minor)` suffix.

I believe I added this test case only to satisfy the code coverage requirements.

felipecrs: NVM, of course there is `(minor)` suffix. Let me think better about this.

felipecrs: Ok, this is listed in the PR description:

> This PR changes it so that the `(major)` (or `(minor)` and `(patch)`) suffix is only added to the PR title when the PR title doesn't yet contain a version.

So, I'll fix the pr title to make that clearer. 

Thanks!

lib/workers/repository/updates/generate.ts
felipecrs: Very nice catch. Thank you!

Done in https://github.com/renovatebot/renovate/pull/41705/commits/afaa040c0b29e838c788abe68fb70974d66fecd3.

RahulGautamSingh (approved):

secustor (approved):
Approved, though waiting for a review of @jamietanna

jamietanna (approved):

PR #41765, opened 2026-03-09
feat: introduce support for minimumMinorAge

tomkerkhove:
## Changes

Adds `minimumMinorAge` — a new config option that defers adoption of new minor versions until the minor version's **first release** has been available for a specified duration. Unlike `minimumReleaseAge` (which checks each individual release's age), `minimumMinorAge` checks the age of the minor version introduction (e.g., `x.y.0`) and then allows the latest patch within that mature minor.

**Config & types:**
- New `minimumMinorAge` option (`string`, default `null`) in config options and `RenovateSharedConfig`

**Core logic (`filter-checks.ts`):**
- Pre-computes a map of `"major.minor"` → first release timestamp from the ascending-sorted releases
- For each candidate, if its minor differs from current version's minor, checks the first minor release's age against `minimumMinorAge`
- Respects `minimumReleaseAgeBehaviour` for missing timestamps
- Composes cleanly with existing `minimumReleaseAge` — both checks must pass

**Documentation:**
- Added `minimumMinorAge` section to `docs/usage/configuration-options.md` with description, example configuration, and release timeline
- Added detailed documentation in `docs/usage/key-concepts/minimum-release-age.md` covering behavior, examples, combining with `minimumReleaseAge`, and missing timestamp handling

**Example:**

```json
{
  "packageRules": [
    {
      "matchDatasources": ["npm"],
      "minimumMinorAge": "7 days"
    }
  ]
}
```

Given releases `1.1.0` (Aug 1), `1.1.3` (Aug 8): on Aug 2nd stays on `1.0.0` (1.1.0 is 1 day old); on Aug 9th upgrades to `1.1.3` (1.1.0 is 8 days old).

Relates to https://github.com/renovatebot/renovate/discussions/35691

## Context

## AI assistance disclosure

8 new tests covering: blocking immature minors, selecting latest patch of mature minor, fallback to older mature minor, no-op for same-minor patches, composition with `minimumReleaseAge`, `packageRules` integration, `internalChecksFilter=strict` pending behavior, and blocking minor versions with missing first-release timestamps when `minimumReleaseAgeBehaviour=timestamp-required`.

---

💡 You can make Copilot smarter by setting up custom instructions, customizing its development environment and configuring Model Context Protocol (MCP) servers. Learn more [Copilot coding agent tips](https://gh.io/copilot-coding-agent-tips) in the docs.


tomkerkhove:
Let me know what you think - Based on that I can sort out how I get CLA happy because I got help from Copilot as I'm not familiar with this language =)

Chrimle:
Suggestion: `minimumMinorReleaseAge`? The current phrasing sounds odd.

tomkerkhove:
I'm OK with that if @viceice  agrees

lib/workers/repository/process/lookup/filter-checks.spec.ts
```diff
@@ -336,5 +336,269 @@ describe('workers/repository/process/lookup/filter-checks', () => {
...
+          releaseTimestamp: '2021-01-06T00:00:00.000Z' as Timestamp,
+        },
+        {
+          version: '1.1.3',
+          releaseTimestamp: '2021-01-08T00:00:00.000Z' as Timestamp,
+        },
+      ];
+
+      it('blocks new minor versions that are not old enough', async () => {
+        dateUtil.getElapsedMs.mockReset();
```
viceice, line 361: i think this can be dropped, mocks are rest globally after each test
```suggestion
```

lib/workers/repository/process/lookup/filter-checks.spec.ts
```diff
@@ -336,5 +336,269 @@ describe('workers/repository/process/lookup/filter-checks', () => {
...
+          'minor',
+          clone(minorReleases),
+        );
+        expect(res.pendingChecks).toBeFalse();
+        expect(res.pendingReleases).toHaveLength(3);
+        expect(res.release?.version).toBe('1.0.1');
+      });
+
+      it('allows minor versions that are old enough and picks latest patch', async () => {
+        dateUtil.getElapsedMs.mockReset();
```
viceice:
```suggestion
```

lib/workers/repository/process/lookup/filter-checks.spec.ts
```diff
@@ -336,5 +336,269 @@ describe('workers/repository/process/lookup/filter-checks', () => {
...
+          'minor',
+          clone(minorReleases),
+        );
+        expect(res.pendingChecks).toBeFalse();
+        expect(res.pendingReleases).toHaveLength(0);
+        expect(res.release?.version).toBe('1.1.3');
+      });
+
+      it('falls back to latest patch of older mature minor when latest minor is too young', async () => {
+        dateUtil.getElapsedMs.mockReset();
```
viceice:
```suggestion
```

lib/workers/repository/process/lookup/filter-checks.ts
```diff
@@ -128,6 +163,45 @@ export async function filterInternalChecks(
...
+
+      if (minimumMinorAgeMs) {
+        const candidateMajor = versioningApi.getMajor(candidateRelease.version);
+        const candidateMinor = versioningApi.getMinor(candidateRelease.version);
+
+        if (candidateMajor !== null && candidateMinor !== null) {
+          const candidateMinorKey = `${candidateMajor}.${candidateMinor}`;
+
+          // Only apply if the minor version is different from the current version
+          if (candidateMinorKey !== currentMinorKey) {
```
viceice: this also applies to rollbacks, candidate is lower than current but newer. 🤔

viceice (commented):
@secustor @jamietanna WDYT about this feature?

jamietanna:
I've seen this requested before - sorry I've not yet had a chance to reply to this PR - and IMO:

- We should be careful to validate + document how this works with `minimumReleaseAge` - which takes priority?
- We already see questions about "why isn't my PR being created when I use `minimumReleaseAge`", so we should take care to make sure that this provides the debugability needed, especially if it takes precedence
- Reading https://github.com/renovatebot/renovate/discussions/35691, Rhys was thinking that we'd support major and minor versions here - I wonder if we should make sure they're both implemented at this point, to make the logic clearer?
- I wonder if what we may instead want is "don't give me .0 releases, let me wait until the .1 or .2", in which case this may not be the right thing
  - what happens with the behaviour of this if .1 never comes? Does .0 get used? Is that what we want?

tomkerkhove:
Those are very valid concerns!

> * We should be careful to validate + document how this works with `minimumReleaseAge` - which takes priority?

Agreed! I'd say existing one wins, or we block it and don't allow using both?

> * Reading [Upgrade to latest patch version for a given minor version after x days #35691](https://github.com/renovatebot/renovate/discussions/35691), Rhys was thinking that we'd support major and minor versions here - I wonder if we should make sure they're both implemented at this point, to make the logic clearer?

If we agree on that, we can do that - I'm open to rename settings how you like

> * I wonder if what we may instead want is "don't give me .0 releases, let me wait until the .1 or .2", in which case this may not be the right thing
>   
>   * what happens with the behaviour of this if .1 never comes? Does .0 get used? Is that what we want?

Yes, if only .0 stays for the configured amount we'd like to use that. If there are no bugs, .1 will never come imo

secustor:
IMO we should rather work this in the current `minimumReleaseAge` feature. 

I could see for example: 

```json
{
  "minimumReleaseAge:" {
    "default": "3 days" // generally wait for 3 days before updating
    "delay<update Type>": "6 days", // wait up to 6 days for `<updateType>`
  }
}
```

tomkerkhove:
@secustor It was discussed in https://github.com/renovatebot/renovate/discussions/35691 and not to be integrated

tomkerkhove:
Having said that, if @rarkins agrees then I can change it to use your approach - I'm OK with that

rarkins:
FYI I'm taking an indefinite break from the project, so I have no opinion on the matter!

tomkerkhove:
> IMO we should rather work this in the current `minimumReleaseAge` feature.
> 
> I could see for example:
> 
> ```json
> {
>   "minimumReleaseAge:" {
>     "default": "3 days" // generally wait for 3 days before updating
>     "delay<update Type>": "6 days", // wait up to 6 days for `<updateType>`
>   }
> }
> ```

@viceice / @RahulGautamSingh are you OK with this proposal? If so, I can start making the change

viceice:
sure, remember to include migrations

tomkerkhove:
@secustor @viceice Feel free to have a look, I've integrated it in to minimumReleaseAge

lib/config/validation.ts
RahulGautamSingh, line 1:
```js
minimumReleaseAge: {
      default: "3 days",
      delayMinr: "7 days"
    }
```

This will not be flagged as wrong in the current validation. Why is there no validation at all for the fields within the object?

RahulGautamSingh (changes requested):

lib/config/options/index.ts
```diff
@@ -2079,8 +2079,10 @@ const options: Readonly<RenovateOptions>[] = [
   },
   {
     name: 'minimumReleaseAge',
-    description: 'Time required before a new release is considered stable.',
+    description:
+      'Time required before a new release is considered stable. Accepts a duration string (e.g. `"3 days"`) or an object with per-update-type overrides (e.g. `{"default": "3 days", "delayMinor": "6 days"}`). When using the object form, `delay<UpdateType>` keys (`delayMajor`, `delayMinor`) check the age of the first release in the version group.',
```
secustor, line 2083:
```suggestion
    description: 'Time required before a new release is considered stable',
```

should be short

lib/config/options/index.ts
```diff
@@ -2079,8 +2079,10 @@ const options: Readonly<RenovateOptions>[] = [
...
     type: 'string',
+    allowObject: true,
```
secustor, line 2085:
```suggestion
    type: 'object',
```
Why not simply migrating from string to object instead?

secustor (commented):
missing coverage

PR #41782, opened 2026-03-10
fix(swift): don't use `v` prefix with `Package.resolved`

jamietanna:
## Changes

As noted in #41780, in the case that a `newVersion` has the `v` prefix,
we use that version, which then results in `xcodebuild` failing to build
the project.

We should instead use `newValue`, which does not include the `v` prefix.

To make this clear in our tests, we can align the test data with the
expected input from our Datasources.

Because we need to use the `newVersion` for digest lookups, we need to
make sure that we still require it (or log when it's not found for some
reason).

## Context

## AI assistance disclosure

The public repository: https://github.com/JamieTanna-Mend-testing/element-x-ios/pull/1

### DIFF
--- a/lib/modules/manager/swift/artifacts.ts
+++ b/lib/modules/manager/swift/artifacts.ts
@@ -181,8 +181,21 @@ export async function updateArtifacts({
     let updated = content;
 
     for (const dep of updatedDeps) {
-      const newVersion = dep.newVersion;
-      if (!newVersion) {
+      const newValue = dep.newValue;
+      if (!newValue) {
+        continue;
+      }
+
+      // shouldn't happen
+      if (!dep.newVersion) {
+        logger.debug(
+          {
+            depName: dep.depName,
+            newVersion: dep.newVersion,
+            newValue: dep.newValue,
+          },
+          'swift: found a newValue but not a newVersion',
+        );
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


lib/modules/manager/swift/artifacts.ts
```diff
@@ -181,8 +181,8 @@ export async function updateArtifacts({
     let updated = content;
 
     for (const dep of updatedDeps) {
-      const newVersion = dep.newVersion;
-      if (!newVersion) {
+      const newValue = dep.newValue;
```
jamietanna, line 184: I'm not sure if we want to have a `newVersion` fallback and trim the `v` prefix if possible?

lib/modules/manager/swift/artifacts.ts
```diff
@@ -196,16 +196,19 @@ export async function updateArtifacts({
...
           'swift: pin already at target version',
         );
         continue;
       }
 
-      const newRevision = await resolveCommitSha(dep, newVersion);
-      updated = updatePinInJson(updated, pin, newVersion, newRevision);
+      const newRevision = await resolveCommitSha(
+        dep,
+        dep.newVersion ?? newValue,
```
jamietanna: ?

```suggestion
        newValue,
```

lib/modules/manager/swift/artifacts.ts
viceice: i don't think so, all other managers are only using `newValue` in artifacts.

it's computed in `generateUpdate`

https://github.com/renovatebot/renovate/blob/2eb25c139676c0106f7ad06352842c163902079f/lib/workers/repository/process/lookup/generate.ts#L57

viceice (commented):
`newValue` seems the right value not `newVersion`

viceice (approved):

jamietanna:
This might be one that'd good to get onto Dev Platform - not sure if I'll get a chance before I go OOO - so might be one for @Gabriel-Ladzaretti next week, if there are other things to go too

PR #41796, opened 2026-03-10
docs: Fix typo in newDigestShort description

davidxia:


lib/util/template/index.ts
```diff
@@ -188,7 +188,7 @@ export const allowedFields = {
   manager: 'The (package) manager which detected the dependency',
   newDigest: 'The new digest value',
   newDigestShort:
-    'A shorted version of newDigest, for use when the full digest is too long to be conveniently displayed',
+    'A shortened version of newDigest, for use when the full digest is too long to be conveniently displayed',
```
davidxia, line 191: How many characters is this?

jamietanna: As in how many characters we'll use for the shortened digest?

jamietanna: It seems like right now it's `.substring(0, 7)`, but that may not always be true

jamietanna (approved):

PR #41813, opened 2026-03-11
fix(datasource/docker): use custom `www-authenticate` parser

viceice:
## Changes
- allows parsing multiple `www-authenticate` challenges and only use `bearer`

## Context

- https://github.com/renovatebot/renovate/discussions/41774
 
## AI assistance disclosure

The public repository: <URL>


jamietanna:
Should this close https://github.com/renovatebot/renovate/issues/41789 ?

jamietanna (commented):
Have we confirmed if there's any additional test coverage from `auth-header` that we can add here? Anything else from the RFC(s)?

viceice:
> Have we confirmed if there's any additional test coverage from `auth-header` that we can add here? Anything else from the RFC(s)?

there could be a lot of validation, but i think it's not required for our use case.

lib/util/http/www-authenticate.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+export const BearerScheme = 'bearer' as const;
+
+export interface SimpleChallenge {
+  scheme: 'basic';
+  params?: string;
+}
+export interface ParamsChallenge {
+  scheme: 'bearer' | ' digest';
```
jamietanna: These were previously capitalised i.e. `Bearer` and `Digest` 

Is this intentional? Do we know if there will be knock-on effects internally?

jamietanna: ❓ is the whitespace intentional?
```suggestion
  scheme: 'bearer' | 'digest';
```

viceice: they are case insensitive and only used on one place in docker datasource yet

lib/util/http/www-authenticate.ts
```diff
@@ -0,0 +1,161 @@
...
+            args.push(t.value);
+            break;
+          case 'pair':
+            params[t.key] = t.value;
+            break;
+        }
+      }
+
+      if (args.length) {
+        result.push({ scheme: c.scheme.toLowerCase(), params: args[0] });
```
jamietanna, line 151: As above, do we know if there will be any issues with this now lowercase'd?

lib/util/http/www-authenticate.ts
viceice: instead of doing upper-case comaper i now do lowercase on parsing as suggested by the http spec

jamietanna: Might be worth a specific test to document this as behaviour?

lib/modules/datasource/docker/common.ts
```diff
@@ -149,15 +146,20 @@ export async function getAuthHeaders(
...
+    const authenticateHeader = challenges.find(
+      (c): c is ParamsChallenge => c.scheme === BearerScheme,
+    );
+
     // If realm isn't an url, we should directly use auth header
     // Can happen when we get a Basic auth or some other auth type
     // * WWW-Authenticate: Basic realm="Artifactory Realm"
     // * Www-Authenticate: Basic realm="https://123456789.dkr.ecr.eu-central-1.amazonaws.com/",service="ecr.amazonaws.com"
     // * www-authenticate: Bearer realm="https://ghcr.io/token",service="ghcr.io",scope="repository:user/image:pull"
     // * www-authenticate: Bearer realm="https://auth.docker.io/token",service="registry.docker.io"
```
jamietanna, line 159: Should we also add a comment to indicate an example from Forgejo 14.0.3?

jamietanna (commented):
I don't see an example of the `www-authenticate` header we were getting back from Forgejo, to confirm that we can handle it?

viceice:
> I don't see an example of the `www-authenticate` header we were getting back from Forgejo, to confirm that we can handle it?

they are validated in `lib/util/http/www-authenticate.spec.ts`. I'll add another test

jamietanna (approved):

jamietanna:
A concern I had (before looking at this code in more detail) is that it may be a risky change if we're changing too much about authentication

Reading more about this + the code changes, it's _only_ parsing the `www-authenticate` header (not changing how we do things

PR #41852, opened 2026-03-12
fix(config)!: don't allow options with a parent to be set in CLI

jamietanna:
## Changes

As per #40813, it does not make sense for an option that has a parent to
be configurable through CLI arguments, as it is difficult to set them
with how we currently derive the flags.

We can also enforce this with a test.

## Context

- [x] This closes an existing Issue, Closes: #40813

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/config/options/index.spec.ts
+++ b/lib/config/options/index.spec.ts
@@ -91,13 +91,13 @@ describe('config/options/index', () => {
     }
   });
 
-  describe('options with a parent cannot be cli=true', () => {
+  describe('options with a parent cannot be set on the command-line', () => {
     const opts = getOptions();
     for (const option of opts) {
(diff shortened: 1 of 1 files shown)


lib/config/options/index.spec.ts
```diff
@@ -90,4 +90,19 @@ describe('config/options/index', () => {
       }
     }
   });
+
+  describe('options with a parent cannot be set on the command-line', () => {
+    const opts = getOptions();
+    for (const option of opts) {
+      if (option.name !== 'enabled' && option.name !== 'managerFilePatterns') {
+        if (!isUndefined(option.parents)) {
```
viceice, line 98: no conditional tests, use filter in outer loop

PR #41859, opened 2026-03-12
docs(config/options): add `supportsTemplating` field

RahulGautamSingh:
## Changes
- Add `supportsTemplating` field to `RenovateOptionBase `
- Show `supportsTemplating` field in doc option table OR a mkdocs tip linking out to https://docs.renovatebot.com/templates/
- Update tests

## Context

- [x] This closes an existing Issue, Closes: #41826 

## AI assistance disclosure

<details>

<summary> Image </summary>

Image

</details>


tools/docs/config.ts
```diff
@@ -131,6 +131,9 @@ function genTable(obj: [string, string][], type: string, def: any): string {
       ) {
         el[1] = `<code>${el[1]}</code>`;
       }
+      if (el[0] === 'supportsTemplating' && el[1]) {
+        el[1] = `<a href="templates/">true</a>`;
+      }
```
jamietanna:
```suggestion
```

tools/docs/config.ts
```diff
@@ -201,6 +204,14 @@ function genExperimentalMsg(el: Record<string, any>): string {
   return warning + '\n';
 }
 
+function genTemplatingMsg(): string {
```
jamietanna, line 211: I think this looks nicer, based on your screenshot!

Potentially right under the table?

tools/docs/config.ts
```diff
@@ -201,6 +204,14 @@ function genExperimentalMsg(el: Record<string, any>): string {
...
+  return (
+    '\n<!-- prettier-ignore -->\n!!! tip "This option supports Renovate\'s template syntax"\n' +
```
jamietanna, line 213: Using `tip` for this is a good shout 🤓

jamietanna (commented):
I think we ned to update https://github.com/renovatebot/renovate/blob/bb135afeae7e01098ff08323b22daa3343a124d2/lib/util/template/index.ts#L117 accordingly, too? Then the list remains the same in terms of what we have made available + documented

Actually sorry, no, these are fields that are made avaialble when templating - it'd be good if we could make this a property of the config options to make it clear what's available when templating in a stronger typed means

Edit: created https://github.com/renovatebot/renovate/issues/42237

lib/config/options/index.spec.ts
```diff
@@ -67,6 +67,19 @@ describe('config/options/index', () => {
     }
   });
 
+  it('supportsTemplating options should be of type string or array with subType string', () => {
```
jamietanna: Can we make this a `describe` with an `it`? Then it'll be clearer when things fail what option is failing

lib/config/options/index.spec.ts
```diff
@@ -67,6 +67,19 @@ describe('config/options/index', () => {
...
+    const opts = getOptions();
+
+    opts
+      .filter((option) => option.supportsTemplating)
+      .forEach((option) => {
+        const valid =
+          option.type === 'string' ||
+          (option.type === 'array' && option.subType === 'string');
```
jamietanna, line 96: ❓ I don't think it should be an array? Our types don't allow that

tools/docs/config.ts
```diff
@@ -320,11 +328,19 @@ export async function generateConfig(dist: string, bot = false): Promise<void> {
...
 
+        if (el.supportsTemplating) {
+          sectionContent += genTemplatingMsg();
+        }
+
+        configOptionsRaw[headerIndex] += sectionContent;
+
+        if (el.experimental) {
+          configOptionsRaw[footerIndex] += genExperimentalMsg(el);
+        }
```
jamietanna, line 391: Should we move this into a separate PR if this seems to be new? (And although good, technically unrelated - one we can get in separately)

RahulGautamSingh: I think it's fine here.

lib/config/options/index.spec.ts
RahulGautamSingh: Kept both string and array since existing options like `labels`, ``addLabels`, `prBodyNotes` use `supportsTemplating` with `array` type

viceice (dismissed):

jamietanna:
Did we say that `extends` supports templating, via https://github.com/renovatebot/renovate/discussions/42039#discussioncomment-16279357 ?

RahulGautamSingh:
> Did we say that `extends` supports templating, via [#42039 (comment)](https://github.com/renovatebot/renovate/discussions/42039#discussioncomment-16279357) ?

It is a special case, I wouldn't count it.

jamietanna:
A special case because they're using `customEnvVariables`? Does that mean every field is templatable for env vars if there is `customEnvVariables` set?

RahulGautamSingh:
`extends` is special because it's hardcoded to go through `template.compile()` at preset resolution time but with an empty input `{}`. So only env vars (from `customEnvVariables`/`process.env`) are available.. none of the usual Renovate template variables like `depName`, `branchName` etc. exist at that point. 

It's not the same as `supportsTemplating` which gives access to full dependency/upgrade context during later stages.

jamietanna:
Interesting - where would be good to get that documented? I'd like us (me!) to not forget that in the future

RahulGautamSingh:
> Interesting - where would be good to get that documented? I'd like us (me!) to not forget that in the future

Shared Config Presets IMO

jamietanna:
I'd say that's either something that needs to go under `extends` or maybe the general templating

jamietanna (changes requested):
According to https://docs.renovatebot.com/templates/#exposed-config-options there are a few options we've not called out - intentional? Which is right/wrong?

```
group
prBodyColumns
prBodyDefinitions
separateMajorMinor
separateMinorPatch
separateMultipleMinor
```

I'd say that the templates page should also make sure that this includes the autogenerated list, rather than a static list from `tools/docs/templates.ts` and `lib/util/template/index.ts`

RahulGautamSingh:
These are two different concepts:                                                                                                   
  - `exposedConfigOptions`: config values available to read inside templates (e.g., `{{separateMajorMinor}}`)                                
  - `supportsTemplating`: config values whose own value gets run through `template.compile()`

lib/config/options/index.spec.ts
```diff
@@ -67,6 +69,87 @@ describe('config/options/index', () => {
...
+      });
+    }
+  });
+
+  describe('template.compile usage must have supportsTemplating', () => {
+    const allOptions = getOptions();
+    const supportsTemplating = new Set(
+      allOptions.filter((o) => o.supportsTemplating).map((o) => o.name),
+    );
+    function collectTsFiles(dir: string): string[] {
```
viceice: please no nested functions

lib/config/options/index.spec.ts
```diff
@@ -67,6 +69,87 @@ describe('config/options/index', () => {
...
+      const results: string[] = [];
+      for (const entry of readdirSync(dir, { withFileTypes: true })) {
```
viceice: use async io, describe can also take an async function 😉

viceice (commented):
otherwise LGTM

docs/usage/templates.md
```diff
@@ -15,6 +15,13 @@ Some are configuration options passed through, while others are generated as par
 
 `logJSON` and `releases` are only allowed in `commitBody` template.
 
+## Options that support templating
+
+<!-- Autogenerate in https://github.com/renovatebot/renovate -->
+<!-- Autogenerate end -->
+
+<!-- Automatically insert options that support templating here -->
```
jamietanna, line 20: We could use this as the start of the template string, as it's more self-documenting (and it may avoid conflicts in the future)

viceice (dismissed):

jamietanna:
I'd say this isn't a user-facing `feat`, so either `chore` or `docs` makes sense here

lib/config/options/index.spec.ts
```diff
@@ -67,6 +86,74 @@ describe('config/options/index', () => {
...
+    const sourceDir = join(__dirname, '..', '..');
+    const sourceFiles = (
+      await Promise.all([
+        collectTsFiles(join(sourceDir, 'workers')),
+        collectTsFiles(join(sourceDir, 'modules', 'datasource', 'custom')),
+        collectTsFiles(join(sourceDir, 'modules', 'manager', 'custom')),
+      ])
+    ).flat();
+
+    const directPattern =
```
jamietanna, line 119: Can we also find `compile` imports i.e. https://github.com/renovatebot/renovate/blob/032a5cb289515bef32628d44ef7e20e643d2a755/lib/util/package-rules/index.ts#L10 ?

jamietanna (commented):
Some additional cases we may need to catch, otherwise LGTM

jamietanna (approved):

PR #41866, opened 2026-03-13
feat(datasource/rpm): add primary_db metadata provider

kpumuk:
## Changes

The RPM datasource currently scans the full `primary.xml.gz` metadata file for every package lookup. This becomes expensive for repositories with large metadata files and many dependencies. Some RPM repositories also publish `primary_db` metadata, which provides the same package information through an indexed SQLite database.

This PR adds `primary_db` support while retaining the existing XML provider:

- Discover both `primary` and `primary_db` entries from `repomd.xml`.
- Prefer `primary_db` in the default `auto` mode and query it using the built-in `node:sqlite` module.
- Fall back to `primary.xml.gz` when `primary_db` is absent or fails to open or query in `auto` mode.
- Treat an SQLite lookup with no matching package as authoritative instead of falling back to XML.
- Allow source selection through the registry URL fragment:
  - `#rpmMetadataSource=auto`: prefer `primary_db`, with XML fallback
  - `#rpmMetadataSource=primary_db`: require SQLite metadata
  - `#rpmMetadataSource=primary`: require XML metadata
- Preserve existing version formatting, deduplication, caching, and XML parsing behavior.
- Document metadata selection and cover discovery, fallback, validation, caching, and source-specific behavior with unit tests.

For repositories that publish `primary_db`, repeated package lookups use indexed SQLite queries instead of repeatedly scanning the full XML metadata file.

## Context

## AI assistance disclosure

- [ ] No - I did not use AI for this contribution.
- [ ] Yes - minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes - other (please describe):

Used Codex to help design, implement, test, benchmark, document, and review the change.

Used private repository with 60 dependencies from AL2023.

### DIFF
--- a/lib/modules/datasource/rpm/index.spec.ts
+++ b/lib/modules/datasource/rpm/index.spec.ts
@@ -135,6 +135,55 @@ function mockPrimaryDbResponse(primaryDbGzip: Buffer): void {
     });
 }
 
+function parseRegistryUrlForTest(
+  datasource: RpmDatasource,
+  url: string,
+): { metadataSource: string; registryUrl: string } {
+  return (
+    datasource as unknown as {
+      parseRegistryUrl: (url: string) => {
+        metadataSource: string;
+        registryUrl: string;
+      };
+    }
+  ).parseRegistryUrl(url);
+}
+
+function getAutoReleasesForTest(
+  datasource: RpmDatasource,
+  metadata: { repomdUrl: string },
+  packageName: string,
+  url: string,
+): Promise<unknown> {
+  return (
+    datasource as unknown as {
+      getAutoReleases: (
+        metadata: { repomdUrl: string },
+        packageName: string,
+        url: string,
+      ) => Promise<unknown>;
+    }
+  ).getAutoReleases(metadata, packageName, url);
+}
+
+function mockPrimaryDbProviderFailure(
+  datasource: RpmDatasource,
+  err: unknown,
+): void {
+  vi.spyOn(
+    (
+      datasource as unknown as {
+        providers: {
+          primary_db: {
+            getReleases: () => Promise<unknown>;
+          };
+        };
+      }
+    ).providers.primary_db,
+    'getReleases',
+  ).mockRejectedValue(err);
+}
+
 describe('modules/datasource/rpm/index', () => {
   let cacheDirResult: DirectoryResult | null;
   let rpmDatasource: RpmDatasource;
@@ -354,6 +403,17 @@ describe('modules/datasource/rpm/index', () => {
       await expect(
         rpmDatasource.getPrimaryGzipUrl(registryUrl),
       ).rejects.toThrow(`No href found in ${registryUrl}repomd.xml`);
+    });
+
+    it('throws when only primary_db metadata is present', async () => {
+      mockRepomdResponse({
+        primaryDbHref: 'repodata/somesha256-primary.sqlite.gz',
+        primaryHref: '',
+      });
+
+      await expect(
+        rpmDatasource.getPrimaryGzipUrl(registryUrl),
+      ).rejects.toThrow(`No primary data found in ${registryUrl}repomd.xml`);
     });
   });
 
@@ -1053,6 +1113,35 @@ describe('modules/datasource/rpm/index', () => {
       ).rejects.toThrow('Invalid rpmMetadataSource in RPM registry URL:');
     });
 
+    it('strips explicit auto rpmMetadataSource from registryUrl', async () => {
+      mockRepomdResponse();
+      mockPrimaryXmlResponse(
+        buildPrimaryXml(codeBlock`
+          <package type="rpm">
+            <name>example-package</name>
+            <arch>x86_64</arch>
+            <version epoch="0" ver="1.0" rel="2.azl3"/>
+          </package>
+        `),
+      );
+
+      await expect(
+        rpmDatasource.getReleases({
+          registryUrl: `${registryUrl}#rpmMetadataSource=auto`,
+          packageName: 'example-package',
+        }),
+      ).resolves.toEqual({
+        releases: [{ version: '1.0-2.azl3' }],
+      });
+    });
+
+    it('keeps invalid registryUrl unchanged when parsing metadata source', () => {
+      expect(parseRegistryUrlForTest(rpmDatasource, 'not a url')).toEqual({
+        metadataSource: 'auto',
+        registryUrl: 'not a url',
+      });
+    });
+
     it('falls back to primary.xml.gz when primary_db is absent', async () => {
       mockRepomdResponse();
       mockPrimaryXmlResponse(
@@ -1101,6 +1190,31 @@ describe('modules/datasource/rpm/index', () => {
       });
 
       expect(releases).toEqual({
+        releases: [{ version: '2.0-1.azl3' }],
+      });
+    });
+
+    it('falls back to primary.xml.gz when primary_db rejects with a non-Error', async () => {
+      mockRepomdResponse({
+        primaryDbHref: 'repodata/somesha256-primary.sqlite.gz',
+      });
+      mockPrimaryDbProviderFailure(rpmDatasource, 'not-a-sqlite-error');
+      mockPrimaryXmlResponse(
+        buildPrimaryXml(codeBlock`
+          <package type="rpm">
+            <name>example-package</name>
+            <arch>x86_64</arch>
+            <version epoch="0" ver="2.0" rel="1.azl3"/>
+          </package>
+        `),
+      );
+
+      await expect(
+        rpmDatasource.getReleases({
+          registryUrl,
+          packageName: 'example-package',
+        }),
+      ).resolves.toEqual({
         releases: [{ version: '2.0-1.azl3' }],
       });
     });
@@ -1262,6 +1376,17 @@ describe('modules/datasource/rpm/index', () => {
           packageName: 'example-package',
         }),
       ).rejects.toThrow(`No primary data found in ${registryUrl}repomd.xml`);
+    });
+
+    it('returns null when repository metadata has no usable source', async () => {
+      await expect(
+        getAutoReleasesForTest(
+          rpmDatasource,
+          { repomdUrl: `${registryUrl}repomd.xml` },
+          'example-package',
+          registryUrl,
+        ),
+      ).resolves.toBeNull();
     });
   });
 });
--- a/lib/modules/datasource/rpm/index.spec.ts
+++ b/lib/modules/datasource/rpm/index.spec.ts
@@ -1,3 +1,4 @@
+import { readFile } from 'node:fs/promises';
 import { Readable } from 'node:stream';
 import { gzipSync } from 'node:zlib';
 import { codeBlock } from 'common-tags';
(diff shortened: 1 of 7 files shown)


docs/usage/configuration-options.md
```diff
@@ -4553,6 +4553,31 @@ Renovate's "rollback" feature exists to propose a downgrade to the next-highest
 Renovate does not create these rollback PRs by default, so this functionality needs to be opted-into.
 We recommend you do this selectively with `packageRules` and not globally.
 
+## rpmMetadataSource
```
viceice: I don't think this is needed at all. why should we make it configurable?

lib/modules/datasource/rpm/providers/sqlite.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    try {
+      db = new Sqlite(extractedFile, {
+        fileMustExist: true,
+        readonly: true,
+      });
+
+      const rows = db
+        .prepare('select version, release from packages where name = ?')
+        .all(packageName) as RpmSqlitePackageRow[];
```
viceice: use async iterator

lib/modules/datasource/rpm/providers/xml.ts
viceice, line 1: do this refactor in a separate PR to be merged first

viceice (changes requested):

docs/usage/configuration-options.md
kpumuk: I first implemented it as a straight change, but then thought it might be considered a breaking change in the behaviour. Another reason is that sqlite files might be larger than xml, and some might want to switch back to xml.

As a result, I added a config option to allow people to switch back and report, if there are any issues to avoid breaking behaviour.

lib/modules/datasource/rpm/providers/sqlite.ts
kpumuk: Unfortunately, better-sqlite3 does not expose an async iterator. Its type is `iterate(...): IterableIterator<Result>`. Wrapping this into async wrapper won't change the state of things, and just pretend we have an async interface :-(

I will switch from `.all()` to `.iterate()` to avoid materializing all rows.

lib/modules/datasource/rpm/providers/xml.ts
kpumuk: Moved XML extraction to https://github.com/renovatebot/renovate/pull/41910

docs/usage/configuration-options.md
viceice: then it should be an experimental feature to opt in, so we flip it enabled after some time and when no issues arise we drop the option again.

viceice: let's encode it into the registry url. we're doing something similar in java and deb datasource

lib/modules/datasource/rpm/providers/common.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export function formatRpmVersion(
+  ver: RpmVersionValue,
+  rel?: RpmVersionValue,
+): string | null {
+  if (ver === undefined || ver === null) {
+    return null;
+  }
+
+  const version = String(ver);
```
viceice:
```suggestion
  const version = `ver`;
```

use stering template or `.toString()`

lib/modules/datasource/rpm/providers/common.ts
```diff
@@ -0,0 +1,62 @@
...
+  url: string,
+): Promise<Buffer> {
+  try {
+    const response = await http.getBuffer(url);
+    if (response.body.length === 0) {
+      logger.debug(`Empty response body from getting ${url}.`);
+      throw new Error(`Empty response body from getting ${url}.`);
+    }
+
+    return await gunzipAsync(response.body);
```
viceice: why still not streaming to a tmp file like we do in deb datasource? So we don't need to have the full buffer in memory.

lib/modules/datasource/rpm/providers/sqlite.ts
```diff
@@ -0,0 +1,96 @@
+import upath from 'upath';
+import { sqlite } from '../../../../expose.ts';
```
viceice: use `node:sqlite` which is now unflagged from experimental warnings in node v24.15.0 (which is our used version)

lib/modules/datasource/rpm/providers/sqlite.ts
viceice: yes, use iterate

lib/modules/datasource/rpm/providers/xml.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  async getReleases(
+    primaryGzipUrl: string,
+    packageName: string,
+  ): Promise<ReleaseResult | null> {
+    const decompressedBuffer = await getGunzippedBuffer(
+      this.http,
+      primaryGzipUrl,
+    );
```
viceice: stream to file and use file stream to pipe to sax parser

lib/modules/datasource/rpm/index.ts
```diff
@@ -98,147 +232,62 @@ export class RpmDatasource extends Datasource {
...
+    }
+
+    return metadata.primaryGzipUrl;
+  }
+
   async getReleasesByPackageName(
     primaryGzipUrl: string,
     packageName: string,
   ): Promise<ReleaseResult | null> {
-    let response: HttpResponse<Buffer>;
```
viceice: can you do the xml parsing move in a seperate PR? a refactor and a new feature in same pr makes review hard and time consuming.

kpumuk: Yes, it was already extracted into https://github.com/renovatebot/renovate/pull/41910 and waiting for review

lib/modules/datasource/rpm/providers/xml.ts
kpumuk: Please check https://github.com/renovatebot/renovate/pull/41910 for the implementation. It also cut the memory usage significantly

lib/modules/datasource/rpm/providers/common.ts
kpumuk: Addressed in https://github.com/renovatebot/renovate/pull/41910

lib/modules/datasource/rpm/providers/sqlite.ts
kpumuk: Applied along with a switch to `node:sqlite`

docs/usage/configuration-options.md
kpumuk: I didn't about this, implemented `http://example.com/repo/repodata/#rpmMetadataSource=primary` similar to how Nuget does it

kpumuk:
I have refactored this MR almost from the ground up based on the initial feedback:

* XML data source refactoring moved to https://github.com/renovatebot/renovate/pull/41910 (it also improved parsing and memory usage dropped significantly)
* Mode switcher moved to URL, as `rpmMetadataSource` parameter (e.g. `http://example.com/repo/repodata/#rpmMetadataSource=primary`)
* Reworked caching layer to ensure we don't cache broken results

The performance results still stand, updating 50 dependencies from Amazon Linux 2023 takes 5–6 seconds. XML parsing is brutal, as re-parsing the whole database every time is expensive (measured in minutes in my laptop).

kpumuk:
@viceice Could you take a look please?

lib/modules/datasource/rpm/index.spec.ts
RahulGautamSingh, line 1: Don't create function counterparts just for testing. Instead use `getReleases` and the internal functions will be tested ie. `parseRegistryUrlForTest`, `getAutoReleasesForTest`

Most of these have a public path (getReleases with a #rpmMetadataSource= fragment) and should go through it.

lib/modules/datasource/rpm/providers/sqlite.ts
```diff
@@ -0,0 +1,63 @@
...
+    );
+    const { DatabaseSync: Sqlite } = await import('node:sqlite');
+    const db = new Sqlite(extractedFile, {
+      readOnly: true,
+    });
+
+    try {
+      const rows = db
+        .prepare('select version, release from packages where name = ?')
+        .iterate(packageName) as Iterable<RpmSqlitePackageRow>;
```
RahulGautamSingh: this assertion bypasses the type system.

the row shape, not validated, so an invalid/errorenous DB column would surface as `undefined` rather than an error.

RahulGautamSingh (changes requested):

lib/modules/datasource/rpm/index.spec.ts
kpumuk: Addressed. Removed `parseRegistryUrlForTest` and `getAutoReleasesForTest`; coverage now goes through `getReleases()`, including `rpmMetadataSource` fragments.

lib/modules/datasource/rpm/providers/sqlite.ts
kpumuk: Removed the row assertion and validate version and release as `string | null`, missing or malformed values now throw.

kpumuk:
Rebased and resolved conflicts: updated the RPM specs to use async `gzip()` and await response helpers, aligning them with the new `node/no-sync` lint rule while preserving test behavior.

lib/modules/datasource/rpm/index.spec.ts
RahulGautamSingh: Thanks. `parseRegistryUrlForTest` and `getAutoReleasesForTest` are gone, but `getPrimaryGzipUrl` and `getReleasesByPackageName` are now public with no production caller, the new `getReleases` routes through `this.providers`. They exist only for the spec, which is the same pattern. Remove both and cover those cases through `getReleases` with `#rpmMetadataSource=` fragments.

lib/modules/datasource/rpm/index.ts
```diff
@@ -47,40 +75,184 @@ export class RpmDatasource extends Datasource {
...
     } catch (err) {
       this.handleGenericErrors(err);
     }
   }
 
-  getReleases(config: GetReleasesConfig): Promise<ReleaseResult | null> {
-    return withCache(
+  async getReleases(config: GetReleasesConfig): Promise<ReleaseResult | null> {
+    const parsedRegistryUrl = config.registryUrl
+      ? this.parseRegistryUrl(config.registryUrl)
```
RahulGautamSingh: `parseRegistryUrl` runs here for the cache key and again inside `_getReleases`. Parse once and pass the result down. 

This copy is also outside the try/catch, so an invalid `#rpmMetadataSource` throws straight out of `getReleases` and skips `handleGenericErrors`.

RahulGautamSingh (changes requested):

lib/modules/datasource/rpm/index.ts
kpumuk: Addressed. Removed the test-only lookup methods and updated the specs to use `getReleases()` with `#rpmMetadataSource=`. The registry URL is now parsed once inside the error-handling path and passed through the lookup.

kpumuk:
Rebased to pick up major version bump https://github.com/renovatebot/renovate/actions/runs/30486180399/job/90692307160?pr=41866

RahulGautamSingh (approved):
LGTM

kpumuk:
Rebased to address merge conflicts.

kpumuk:
The conflicts came from the upstream refactor that [removed gzip-specific naming from the RPM datasource](https://github.com/renovatebot/renovate/pull/46264). The SQLite provider now uses the same format-neutral metadata URL and decompression terminology as the XML provider, with no intended behavioral changes.

kpumuk:
Rebased onto `upstream/main`. The RPM conflicts came from #46102, which introduced `Datasource.cached()` and renamed `_getReleases` to `fetchReleases`. I kept those upstream conventions while preserving this PR’s source-specific cache keys and SQLite/XML provider selection. I also retained the custom-registry API change from #40497.

PR #41882, opened 2026-03-13
fix(manager/flux): extract both tag and digest for OCIRepository

### LINKED ISSUE #29768: Let `flux` manager update tags and digests (`flux` currently ignores tags)

Renovate's `flux` manager does _not_ support updating tags _and_ digests yet:

https://github.com/renovatebot/renovate/blob/8975c9bda67880fd4c91bdc32534e24f98a90e0e/lib/modules/manager/flux/extract.ts#L214

`@viceice` says Renovate should support updating tags _and_ digests. `@viceice` will look next week if there's an easy way to add support.

Related Discussion:

- #29521

robinschneider:
## Changes

This PR allows using both tag and digest for the Flux manager for `OCIRepository`.
Flux ignores the tag if there is an additional digest, but a digest is not human readable, so it is better to use them simultaneously. Flux does not support tag@digest, is uses separate fields for these.

## Context

- [x] This closes an existing Issue, Closes: #29768

## AI assistance disclosure

I used Gemini 3.1 Pro to support for this PR. I have carefully checked and tested the code.
It took many iterations to get it done.
It helped me to understand what parts of renovate need to be modified in order to achieve updating Flux tag and digest.

## Documentation

## How I've tested my work

The public repository: https://github.com/robinschneider/renovate-flux/pull/1

### DIFF
--- a/lib/modules/manager/flux/extract.ts
+++ b/lib/modules/manager/flux/extract.ts
@@ -307,16 +307,21 @@ function resolveResourceManifest(
           // Attempt to find the exact block spanning `tag: ...` to `digest: ...` (or vice-versa) in the source content
           const escapedTag = escapeRegExp(resource.spec.ref.tag);
           const escapedDigest = escapeRegExp(resource.spec.ref.digest);
-          // Regex handles both ordering: tag before digest, or digest before tag.
-          // Allows optional single or double quotes around the values.
-          const multilineRegEx = regEx(
-            `(tag\\s*:\\s*["']?${escapedTag}["']?[\\s\\S]+?digest\\s*:\\s*["']?${escapedDigest}["']?)|(digest\\s*:\\s*["']?${escapedDigest}["']?[\\s\\S]+?tag\\s*:\\s*["']?${escapedTag}["']?)`,
+          const tagFirstRegex = regEx(
+            `ref\\s*:[\\s\\S]+?tag\\s*:\\s*["']?${escapedTag}["']?[\\s\\S]+?digest\\s*:\\s*["']?${escapedDigest}["']?`,
           );
-          const match = multilineRegEx.exec(content);
+          const digestFirstRegex = regEx(
+            `ref\\s*:[\\s\\S]+?digest\\s*:\\s*["']?${escapedDigest}["']?[\\s\\S]+?tag\\s*:\\s*["']?${escapedTag}["']?`,
+          );
+
+          const tagFirstMatch = tagFirstRegex.exec(content);
+          const digestFirstMatch = digestFirstRegex.exec(content);
+          const match = tagFirstMatch ?? digestFirstMatch;
+
           if (match) {
             combinedDep.replaceString = match[0];
             // We use the literals directly for replacement to preserve any surrounding quotes if they exist in the match
-            if (match[1]) {
+            if (tagFirstMatch) {
               // tag came first
               combinedDep.autoReplaceStringTemplate = match[0]
                 .replace(resource.spec.ref.tag, '{{newValue}}')
--- a/lib/modules/manager/flux/extract.ts
+++ b/lib/modules/manager/flux/extract.ts
@@ -308,10 +308,10 @@ function resolveResourceManifest(
           const escapedTag = escapeRegExp(resource.spec.ref.tag);
           const escapedDigest = escapeRegExp(resource.spec.ref.digest);
           const tagFirstRegex = regEx(
-            `ref\\s*:[\\s\\S]+?tag\\s*:\\s*["']?${escapedTag}["']?[\\s\\S]+?digest\\s*:\\s*["']?${escapedDigest}["']?`,
+            `ref\\s*:[\\s\\S]+?["']?tag["']?\\s*:\\s*["']?${escapedTag}["']?[\\s\\S]+?["']?digest["']?\\s*:\\s*["']?${escapedDigest}["']?`,
           );
           const digestFirstRegex = regEx(
-            `ref\\s*:[\\s\\S]+?digest\\s*:\\s*["']?${escapedDigest}["']?[\\s\\S]+?tag\\s*:\\s*["']?${escapedTag}["']?`,
+            `ref\\s*:[\\s\\S]+?["']?digest["']?\\s*:\\s*["']?${escapedDigest}["']?[\\s\\S]+?["']?tag["']?\\s*:\\s*["']?${escapedTag}["']?`,
           );
 
           const tagFirstMatch = tagFirstRegex.exec(content);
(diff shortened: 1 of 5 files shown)


matheuscscp:
@secustor Looks like this PR is fixing an important bug in the flux manager, can you pls take a look? :pray:

lib/modules/manager/flux/extract.ts
```diff
@@ -292,29 +293,72 @@ function resolveResourceManifest(
...
+            }
+          } else {
+            logger.debug(
+              { file: manifest.file, name: resource.metadata.name },
+              'Could not find exact tag/digest block in content, auto-replace might fail',
+            );
+            // fallback (will likely fail auto-replace if not formatted this way, but avoids breaking tests outright)
+            combinedDep.replaceString = `${resource.spec.ref.tag}@${resource.spec.ref.digest}`;
+            combinedDep.autoReplaceStringTemplate =
+              '{{newValue}}@{{newDigest}}';
```
secustor: As I understand this will lead to errors with Flux

lib/modules/manager/flux/extract.ts
```diff
@@ -292,29 +293,72 @@ function resolveResourceManifest(
...
+              // tag came first
+              combinedDep.autoReplaceStringTemplate = match[0]
+                .replace(resource.spec.ref.tag, '{{newValue}}')
+                .replace(resource.spec.ref.digest, '{{newDigest}}');
+            } else {
+              // digest came first
+              combinedDep.autoReplaceStringTemplate = match[0]
+                .replace(resource.spec.ref.digest, '{{newDigest}}')
+                .replace(resource.spec.ref.tag, '{{newValue}}');
+            }
```
secustor: In which scenario would the order here matter?

lib/modules/manager/flux/extract.ts
```diff
@@ -292,29 +293,72 @@ function resolveResourceManifest(
...
+          // Set currentValue to the tag so the docker datasource can look up the image's new digest
+          combinedDep.currentValue = resource.spec.ref.tag;
+
+          // Attempt to find the exact block spanning `tag: ...` to `digest: ...` (or vice-versa) in the source content
+          const escapedTag = escapeRegExp(resource.spec.ref.tag);
+          const escapedDigest = escapeRegExp(resource.spec.ref.digest);
+          // Regex handles both ordering: tag before digest, or digest before tag.
+          // Allows optional single or double quotes around the values.
+          const multilineRegEx = regEx(
+            `(tag\\s*:\\s*["']?${escapedTag}["']?[\\s\\S]+?digest\\s*:\\s*["']?${escapedDigest}["']?)|(digest\\s*:\\s*["']?${escapedDigest}["']?[\\s\\S]+?tag\\s*:\\s*["']?${escapedTag}["']?)`,
```
secustor: Split this into two to simplify the regex and additionally match `ref:` before to anchor this in the correct object

lib/modules/manager/flux/extract.ts
secustor, line 475: Don't duplicate the state instead use another type which combines `manifest` and `content`

lib/modules/manager/flux/extract.ts
```diff
@@ -364,9 +409,13 @@ export async function extractAllPackageFiles(
 ): Promise<PackageFile<FluxManagerData>[] | null> {
   const manifests: FluxManifest[] = [];
   const results: PackageFile<FluxManagerData>[] = [];
+  const fileContents = new Map<string, string>(); // Add a map to cache content
```
secustor: Use Record instead of Map

lib/modules/manager/flux/extract.ts
matheuscscp: Yeah, `OCIRepository.spec.ref.tag` cannot contain the digest

lib/modules/manager/flux/extract.ts
robinschneider: The order matters because `replace()` only replaces the first occurrence of a string.

If the tag value happens to be a substring of the digest string (e.g. tag `123` and digest `sha256:abcd123`), and we process the tag first even though the digest comes earlier in the file, `.replace()` might accidentally overwrite a part of the digest instead of the actual tag.

By processing the replacements in the order they physically appear in the YAML file (top-to-bottom), we prevent this issue. The first field is safely replaced by its literal placeholder like `'{{newDigest}}'`, which guarantees it won't conflict with the second replacement.

lib/modules/manager/flux/extract.ts
```diff
@@ -367,10 +416,12 @@ export async function extractAllPackageFiles(
 
   for (const file of packageFiles) {
     const content = await readLocalFile(file, 'utf8');
-    // TODO #22198
-    const manifest = readManifest(content!, file);
-    if (manifest) {
-      manifests.push(manifest);
+    if (content) {
+      // TODO #22198
```
viceice:
```suggestion
```

lib/modules/manager/flux/extract.ts
```diff
@@ -292,29 +295,74 @@ function resolveResourceManifest(
...
+              // digest came first
+              combinedDep.autoReplaceStringTemplate = match[0]
+                .replace(resource.spec.ref.digest, '{{newDigest}}')
+                .replace(resource.spec.ref.tag, '{{newValue}}');
+            }
+          } else {
+            logger.debug(
+              { file: manifest.file, name: resource.metadata.name },
+              'Could not find exact tag/digest block in content, skipping extraction',
+            );
```
viceice, line 390: so this would skip?

```yaml
---
apiVersion: source.toolkit.fluxcd.io/v1
kind: OCIRepository
metadata:
  name: valkey
spec:
  interval: 12h
  url: oci://ghcr.io/visualon/bitnamicharts/valkey
  ref:
    digest: sha256:761c3189c482d0f1f0ad3735ca05c4c398cae201d2169f6645280c7b7b2ce6fc
    # tag is for renovate and will be ignored by flux
    tag: 5.4.7
```

🤔 

instead of regex we should use the yaml document parser, then find the nodes, from them we can get start and end, so we can easily substring replace string from content.

see npm for parsing, the nodes have a Range, which tell you the positions in source.

https://github.com/renovatebot/renovate/blob/a4456c139c20fa02d5447705c40a52c4441fb100/lib/modules/manager/npm/update/dependency/pnpm.ts#L45

robinschneider: This would break the regex.
I adjusted it to use a yaml document parser as you suggested.
I also added more example configurations to the repository that I linked above for testing

secustor (approved):
LGTM now

PR #41910, opened 2026-03-13
refactor(datasource/rpm): extract xml metadata provider

kpumuk:
## Changes

This pull request was split from https://github.com/renovatebot/renovate/pull/41866, and only contains the basic refactoring of the RPM data source to extract `primary` (XML) package format.

Previously, the RPM datasource mixed repository metadata discovery and XML package parsing inside a single class. That made the code harder to extend for additional metadata backends and made the later SQLite work more invasive than it needed to be.

> [!NOTE]
> This refactor intentionally keeps one small behavior improvement that was already folded into the extraction: malformed RPM XML entries with a `<version>` element missing the `ver` attribute are now ignored instead of producing a release with version `undefined`.
>
> That is a behavior change for invalid metadata, but it is a safer outcome and aligns better with the intended RPM parsing behavior.

In the https://github.com/renovatebot/renovate/pull/41910#discussion_r2945024337 discussion an additional improvement was proposed to use a file on disk for unpacking the XML, which led to significant memory savings. In the benchmark over 50 Amazon Linux 2 dependencies, the results are:

| Version | Avg Wall Time (3 runs) | Avg Peak RSS (3 runs) |
|---|---:|---:|
| Before (`upstream/main`) | 416.55s | 1578.2 MiB |
| After (`rpm-xml-provider`) | 416.93s (`+0.09%`) | 533.8 MiB (`-66.2%`) |

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/datasource/rpm/index.ts
+++ b/lib/modules/datasource/rpm/index.ts
@@ -54,7 +54,7 @@ export class RpmDatasource extends Datasource {
       const primaryGzipUrl = await this.getPrimaryGzipUrl(registryUrl);
       return await this.getReleasesByPackageName(primaryGzipUrl, packageName);
     } catch (err) {
-      this.handleGenericErrors(err as Error);
+      this.handleGenericErrors(err);
     }
   }
 
--- a/lib/modules/datasource/rpm/index.ts
+++ b/lib/modules/datasource/rpm/index.ts
@@ -1,16 +1,10 @@
-import { Readable } from 'node:stream';
-import { gunzip } from 'node:zlib';
-import sax from 'sax';
-import { promisify } from 'util';
 import { XmlDocument } from 'xmldoc';
 import { logger } from '../../../logger/index.ts';
 import { withCache } from '../../../util/cache/package/with-cache.ts';
-import type { HttpResponse } from '../../../util/http/index.ts';
 import { joinUrlParts } from '../../../util/url.ts';
 import { Datasource } from '../datasource.ts';
-import type { GetReleasesConfig, Release, ReleaseResult } from '../types.ts';
-
-const gunzipAsync = promisify(gunzip);
+import type { GetReleasesConfig, ReleaseResult } from '../types.ts';
+import { RpmXmlMetadataProvider } from './providers/xml.ts';
 
 export class RpmDatasource extends Datasource {
   static readonly id = 'rpm';
@@ -18,8 +12,11 @@ export class RpmDatasource extends Datasource {
   // repomd.xml is a standard file name in RPM repositories which contains metadata about the repository
   static readonly repomdXmlFileName = 'repomd.xml';
 
+  private readonly xmlProvider: RpmXmlMetadataProvider;
+
   constructor() {
     super(RpmDatasource.id);
+    this.xmlProvider = new RpmXmlMetadataProvider(this.http);
   }
 
   /**
@@ -52,14 +49,12 @@ export class RpmDatasource extends Datasource {
     if (!registryUrl || !packageName) {
       return null;
     }
+
     try {
       const primaryGzipUrl = await this.getPrimaryGzipUrl(registryUrl);
-      if (!primaryGzipUrl) {
-        return null;
-      }
       return await this.getReleasesByPackageName(primaryGzipUrl, packageName);
     } catch (err) {
-      this.handleGenericErrors(err);
+      this.handleGenericErrors(err as Error);
     }
   }
 
@@ -75,10 +70,37 @@ export class RpmDatasource extends Datasource {
     );
   }
 
-  // Fetches the primary.xml.gz URL from the repomd.xml file.
-  private async _getPrimaryGzipUrl(
+  private getPrimaryRepodataUrl(
+    xml: XmlDocument,
     registryUrl: string,
-  ): Promise<string | null> {
+    repomdUrl: string,
+  ): string {
+    const primaryData = xml.childWithAttribute('type', 'primary');
+
+    if (!primaryData) {
+      throw new Error(`No primary data found in ${repomdUrl}`);
+    }
+
+    const locationElement = primaryData.childNamed('location');
+    if (!locationElement) {
+      throw new Error(`No location element found in ${repomdUrl}`);
+    }
+
+    const href = locationElement.attr.href;
+    if (!href) {
+      throw new Error(`No href found in ${repomdUrl}`);
+    }
+
+    // replace trailing "repodata/" from registryUrl, if it exists, with a "/" because href includes "repodata/"
+    const registryUrlWithoutRepodata = registryUrl.replace(
+      /\/repodata\/?$/,
+      '/',
+    );
+
+    return joinUrlParts(registryUrlWithoutRepodata, href);
+  }
+
+  private async _getPrimaryGzipUrl(registryUrl: string): Promise<string> {
     const repomdUrl = joinUrlParts(
       registryUrl,
       RpmDatasource.repomdXmlFileName,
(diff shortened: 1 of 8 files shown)


lib/modules/datasource/rpm/index.ts
```diff
@@ -52,14 +49,12 @@ export class RpmDatasource extends Datasource {
...
+
     try {
       const primaryGzipUrl = await this.getPrimaryGzipUrl(registryUrl);
-      if (!primaryGzipUrl) {
-        return null;
-      }
       return await this.getReleasesByPackageName(primaryGzipUrl, packageName);
     } catch (err) {
-      this.handleGenericErrors(err);
+      this.handleGenericErrors(err as Error);
```
RahulGautamSingh: needed? or AI being extra.

viceice: is that change really needed?

lib/modules/datasource/rpm/index.ts
```diff
@@ -75,10 +70,37 @@ export class RpmDatasource extends Datasource {
     );
   }
 
-  // Fetches the primary.xml.gz URL from the repomd.xml file.
-  private async _getPrimaryGzipUrl(
+  private getPrimaryRepodataUrl(
```
viceice: I think this can be moved to separate file as it no longer needs this context?

lib/modules/datasource/rpm/providers/common.ts
```diff
@@ -0,0 +1,62 @@
...
+    releases: uniqueVersions.map((version) => ({ version })),
+  };
+}
+
+export async function getGunzippedBuffer(
+  http: Http,
+  url: string,
+): Promise<Buffer> {
+  try {
+    const response = await http.getBuffer(url);
```
viceice: this should be refactored to use a cache file, so we don't hold the huge buffer in memory

lib/modules/datasource/rpm/providers/xml.ts
```diff
@@ -0,0 +1,98 @@
...
+        logger.debug(`SAX parsing error in ${primaryGzipUrl}: ${err.message}`);
+        setImmediate(() => saxParser.removeAllListeners());
+        reject(err);
+      });
+      saxParser.on('end', () => {
+        settled = true;
+        setImmediate(() => saxParser.removeAllListeners());
+        resolve();
+      });
+      Readable.from(decompressedBuffer).pipe(saxParser);
```
viceice: we should use an available pipeline here and await this and the outer promise .  maybe the outer isn't needed then.

viceice (commented):
you can do the buffer change in a follow up PR if it feels better to you

lib/modules/datasource/rpm/index.ts
kpumuk: A left-over from me fighting linter in another place. Will drop it

lib/modules/datasource/rpm/index.ts
kpumuk: Makes sense. I can extract the repomd parsing helper into a small helper module so index.ts is just fetch/cache/orchestration.

lib/modules/datasource/rpm/providers/xml.ts
kpumuk: A direct `pipeline(Readable.from(buffer), saxParser)` would look simpler, but it does not work correctly here. `sax.createStream()` is duplex, and with no consumer on its readable side the pipeline hangs.

So if we switch this to `pipeline(...)`, it needs a small writable adapter in front of the SAX parser. That keeps the pipeline lifecycle handling, while still letting SAX drive parsing and surface parser errors correctly.

```diff
diff --git a/lib/modules/datasource/rpm/providers/xml.ts b/lib/modules/datasource/rpm/providers/xml.ts
index 06bd1bb7e..16d6a51c6 100644
--- a/lib/modules/datasource/rpm/providers/xml.ts
+++ b/lib/modules/datasource/rpm/providers/xml.ts
@@ -1,5 +1,6 @@
-import { Readable } from 'node:stream';
+import { Readable, Writable } from 'node:stream';
 import sax from 'sax';
+import { pipeline } from 'stream/promises';
 import { logger } from '../../../../logger/index.ts';
 import type { Http } from '../../../../util/http/index.ts';
 import type { ReleaseResult } from '../../types.ts';
@@ -67,25 +68,34 @@ export class RpmXmlMetadataProvider {
       }
     });
 
-    await new Promise<void>((resolve, reject) => {
-      let settled = false;
-      saxParser.on('error', (err: Error) => {
-        if (settled) {
-          return;
-        }
-        settled = true;
-        logger.debug(`SAX parsing error in ${primaryGzipUrl}: ${err.message}`);
-        setImmediate(() => saxParser.removeAllListeners());
-        reject(err);
-      });
-      saxParser.on('end', () => {
-        settled = true;
-        setImmediate(() => saxParser.removeAllListeners());
-        resolve();
-      });
-      Readable.from(decompressedBuffer).pipe(saxParser);
+    let parserError: Error | undefined;
+    saxParser.on('error', (err: Error) => {
+      if (parserError) {
+        return;
+      }
+
+      parserError = err;
+      logger.debug(
+        `SAX parsing error in ${primaryGzipUrl}: ${
+          err instanceof Error ? err.message : err
+        }`,
+      );
     });
 
+    await pipeline(
+      Readable.from(decompressedBuffer),
+      new Writable({
+        write(chunk, _encoding, callback) {
+          saxParser.write(chunk);
+          callback(parserError);
+        },
+        final(callback) {
+          saxParser.end();
+          callback(parserError);
+        },
+      }),
+    );
+
     const result = buildReleaseResult(releases);
     if (!result) {
       logger.trace(
```

lib/modules/datasource/rpm/providers/common.ts
kpumuk: This is a great suggestion. In fact, I got my CI worker killed today due to OOM, and I strongly suspect this code. I will try to generalize this to fit both this branch and the primary_db/sqlite (which already materializes the database on disk, but it still unzips it in memory).

Between two choices `withCache()` with pre-defined TTL and file-based freshness comparison (similar to how Debian data-source does) I went with `If-Modified-Since` and timestamps. Let me know if the `withCache()` is preferred.

The benchmark is promising. I did 3 runs over the 50 Amazon Linux 2023 dependencies:

| Branch | Run 1 Wall Time | Run 2 Wall Time | Run 3 Wall Time | Avg Wall Time | Run 1 Peak RSS | Run 2 Peak RSS | Run 3 Peak RSS | Avg Peak RSS |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| `upstream/main` | 418.05s | 404.64s | 426.97s | 416.55s | 1425.1 MiB | 1432.1 MiB | 1877.5 MiB | 1578.2 MiB |
| `rpm-xml-provider` | 419.05s | 420.19s | 411.54s | 416.93s | 535.7 MiB | 534.9 MiB | 531.0 MiB | 533.8 MiB |
| Delta (`rpm-xml-provider` vs `upstream/main`) | +1.00s | +15.55s | -15.43s | +0.37s (`+0.09%`) | -889.4 MiB | -897.2 MiB | -1346.5 MiB | -1044.4 MiB (`-66.2%`) |

lib/modules/datasource/rpm/index.ts
kpumuk: https://github.com/renovatebot/renovate/pull/41910/commits/4819c3056340a5a67b9d354c5521bc65152d386b

lib/modules/datasource/rpm/index.ts
kpumuk: https://github.com/renovatebot/renovate/pull/41910/commits/ca5b227e94c8dec7fa51b148c56f18ef8170fdfc

lib/modules/datasource/rpm/providers/common.ts
kpumuk: https://github.com/renovatebot/renovate/pull/41910/commits/23e860fd64a00ff89a38b7400dff702bdb9d681a

I have added `extension` parameter to all helpers, so I can re-use it later for `primary_db` change

kpumuk:
@viceice would you mind re-checking this proposal? it did grew a little bit from the initial simple refactoring, but I think even on its own without `primary_db` this change provides a significant memory optimization.

kpumuk:
Added one more commit after finding an interesting edge case.

Before this change, a metadata refresh would extract directly into the shared cached XML file. If the downloaded archive was bad or truncated, we could partially overwrite a previously good cache entry and still return that path, leaving later lookups to read corrupted metadata. There was also a race where one lookup could be parsing the file while another was rewriting it in place.

The fix is to extract into a temporary file first and only replace the shared cache file after extraction succeeds. That means readers see either the old complete file or the new complete file, and a failed refresh preserves the previous known-good cache.

This is not unique to RPM. The Debian datasource currently uses the same in-place extraction pattern, so it is susceptible to the same class of issue and should get the same temp-file + atomic-replace treatment in a follow-up.

viceice (approved):

PR #41913, opened 2026-03-13
feat(pep621): use exclude-newer during uv lock when minimumReleaseAge is set

thejoeejoee:
## Changes

When `minimumReleaseAge` is configured, Renovate now passes `UV_EXCLUDE_NEWER` as an environment variable to `uv lock`.

Additionally, if `pyproject.toml` already defines `[tool.uv] exclude-newer`, Renovate compares it with its own computed date and uses the more restrictive (older) one. This prevents Renovate from accidentally widening the package resolution window beyond what the project intended.

📖 **Related**: uv documents recommended `exclude-newer` usage for dependency bots in their [integration guide](https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependency-cooldown).

Key changes:
- `lib/modules/manager/types.ts` — added `minimumReleaseAge` to `UpdateArtifactsConfig`
- `lib/modules/manager/pep621/schema.ts` — added `exclude-newer` to `UvConfig` Zod schema (handles both TOML string and Date types)
- `lib/modules/manager/pep621/processors/uv.ts` — `UV_EXCLUDE_NEWER` env var with `min(renovate_date, pyproject_date)` logic
- `lib/modules/manager/pep621/processors/uv.spec.ts` — 28 tests covering env var usage, min-date comparison, invalid values, and edge cases

## Context

- [x] This closes an existing Issue, Closes: #41654

## AI assistance disclosure

opencode (Claude Opus 4.6 / Sonnet 4.6) used for test writing, and code review iteration

### DIFF
--- a/docs/usage/key-concepts/minimum-release-age.md
+++ b/docs/usage/key-concepts/minimum-release-age.md
@@ -44,15 +44,7 @@ To protect against this, it's recommended to ensure that your package manager co
 There is ongoing work to [integrate more closely with package manager checks](https://github.com/renovatebot/renovate/issues/41652) to make sure that Renovate's minimum release age configuration is specified when calling package managers that support it.
 If you have a package manager you'd like supported, please raise a [Suggest an Idea Discussion](https://github.com/renovatebot/renovate/discussions/new?category=suggest-an-idea).
 
-#### npm
-
-When `minimumReleaseAge` is configured, Renovate passes `--before=<date>` to npm commands during lock file generation.
-This ensures that npm only resolves package versions that were available before the cooldown threshold, protecting against newly published (and potentially malicious) transitive dependencies.
-
-The `--before` date is calculated as `now - minimumReleaseAge`.
-If a `before=<date>` or `min-release-age=<days>` setting already exists in the project's `.npmrc`, Renovate uses the stricter (older) of the two dates.
-
-### uv
+#### uv
 
 <!-- prettier-ignore -->
 For uv, Renovate reads `[tool.uv] exclude-newer` from `pyproject.toml` and uses the more restrictive (older) date between it and `minimumReleaseAge`.
--- a/docs/usage/key-concepts/minimum-release-age.md
+++ b/docs/usage/key-concepts/minimum-release-age.md
@@ -55,6 +55,7 @@ If a `before=<date>` or `min-release-age=<days>` setting already exists in the p
 <!-- prettier-ignore -->
 !!! note
     For uv, Renovate reads `[tool.uv] exclude-newer` from `pyproject.toml` and uses the more restrictive (older) date between it and `minimumReleaseAge`.
+    The `exclude-newer` value can be an RFC 3339 timestamp (e.g. `2025-01-01T00:00:00Z`) or a friendly duration (e.g. `2 weeks`), but ISO 8601 durations (e.g. `P14D`) are not supported.
     See uv's [dependency bot integration guide](https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependency-cooldown) for details on configuring `exclude-newer`.
 
 ### What happens if the datasource and/or registry does not provide a release timestamp, when using `minimumReleaseAge`?
--- a/docs/usage/key-concepts/minimum-release-age.md
+++ b/docs/usage/key-concepts/minimum-release-age.md
@@ -52,11 +52,12 @@ This ensures that npm only resolves package versions that were available before 
 The `--before` date is calculated as `now - minimumReleaseAge`.
 If a `before=<date>` or `min-release-age=<days>` setting already exists in the project's `.npmrc`, Renovate uses the stricter (older) of the two dates.
 
+### uv
+
 <!-- prettier-ignore -->
-!!! note
-    For uv, Renovate reads `[tool.uv] exclude-newer` from `pyproject.toml` and uses the more restrictive (older) date between it and `minimumReleaseAge`.
-    The `exclude-newer` value can be an RFC 3339 timestamp (e.g. `2025-01-01T00:00:00Z`) or a friendly duration (e.g. `2 weeks`), but ISO 8601 durations (e.g. `P14D`) are not supported.
-    See uv's [dependency bot integration guide](https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependency-cooldown) for details on configuring `exclude-newer`.
+For uv, Renovate reads `[tool.uv] exclude-newer` from `pyproject.toml` and uses the more restrictive (older) date between it and `minimumReleaseAge`.
+The `exclude-newer` value can be an RFC 3339 timestamp (e.g. `2025-01-01T00:00:00Z`) or a friendly duration (e.g. `2 weeks`), but ISO 8601 durations (e.g. `P14D`) are not supported.
+See uv's [dependency bot integration guide](https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependency-cooldown) for details on configuring `exclude-newer`.
 
 ### What happens if the datasource and/or registry does not provide a release timestamp, when using `minimumReleaseAge`?
 
--- a/docs/usage/key-concepts/minimum-release-age.md
+++ b/docs/usage/key-concepts/minimum-release-age.md
@@ -44,6 +44,19 @@ To protect against this, it's recommended to ensure that your package manager co
 There is ongoing work to [integrate more closely with package manager checks](https://github.com/renovatebot/renovate/issues/41652) to make sure that Renovate's minimum release age configuration is specified when calling package managers that support it.
 If you have a package manager you'd like supported, please raise a [Suggest an Idea Discussion](https://github.com/renovatebot/renovate/discussions/new?category=suggest-an-idea).
 
+#### npm
+
+When `minimumReleaseAge` is configured, Renovate passes `--before=<date>` to npm commands during lock file generation.
+This ensures that npm only resolves package versions that were available before the cooldown threshold, protecting against newly published (and potentially malicious) transitive dependencies.
+
+The `--before` date is calculated as `now - minimumReleaseAge`.
+If a `before=<date>` or `min-release-age=<days>` setting already exists in the project's `.npmrc`, Renovate uses the stricter (older) of the two dates.
+
+<!-- prettier-ignore -->
+!!! note
+    For uv, Renovate reads `[tool.uv] exclude-newer` from `pyproject.toml` and uses the more restrictive (older) date between it and `minimumReleaseAge`.
+    See uv's [dependency bot integration guide](https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependency-cooldown) for details on configuring `exclude-newer`.
+
 ### What happens if the datasource and/or registry does not provide a release timestamp, when using `minimumReleaseAge`?
 
 <!-- prettier-ignore -->
--- a/lib/modules/manager/pep621/processors/uv.spec.ts
+++ b/lib/modules/manager/pep621/processors/uv.spec.ts
@@ -14,7 +14,6 @@ import { getPkgReleases as _getPkgReleases } from '../../../datasource/index.ts'
 import { PypiDatasource } from '../../../datasource/pypi/index.ts';
 import type { UpdateArtifact, UpdateArtifactsConfig } from '../../types.ts';
 import { parsePyProject } from '../extract.ts';
-import { PyProject } from '../schema.ts';
 import { depTypes } from '../utils.ts';
 import { UvProcessor } from './uv.ts';
 
@@ -1056,17 +1055,6 @@ describe('modules/manager/pep621/processors/uv', () => {
         });
         expect(logger.logger.debug).toHaveBeenCalledWith(
           "Invalid exclude-newer value 'not-a-date' in pyproject.toml, ignoring",
-        );
-      });
-
-      it('handles exclude-newer as Date object in schema', () => {
-        const result = PyProject.parse({
-          tool: {
-            uv: { 'exclude-newer': new Date('2026-03-05T00:00:00.000Z') },
-          },
-        });
-        expect(result.tool?.uv?.['exclude-newer']).toBe(
-          '2026-03-05T00:00:00.000Z',
         );
       });
     });
--- a/lib/modules/manager/pep621/processors/uv.spec.ts
+++ b/lib/modules/manager/pep621/processors/uv.spec.ts
@@ -934,6 +934,27 @@ describe('modules/manager/pep621/processors/uv', () => {
         });
       });
 
+      it('sets UV_EXCLUDE_NEWER for non-maintenance upgrade-package command', async () => {
+        const updatedDeps = [
+          { packageName: 'dep1', depType: depTypes.dependencies },
+        ];
+        await processor.updateArtifacts(
+          {
+            packageFileName: 'folder/pyproject.toml',
+            newPackageFileContent: '',
+            config: {
+              minimumReleaseAge: '3 days',
+            },
+            updatedDeps,
+          },
+          parsePyProject('')!,
+        );
+        expect(execSnapshots[0].cmd).toBe('uv lock --upgrade-package dep1');
+        expect(execSnapshots[0].options?.env).toMatchObject({
+          UV_EXCLUDE_NEWER: '2026-03-10T00:00:00.000Z',
+        });
+      });
+
       it('skips UV_EXCLUDE_NEWER when minimumReleaseAge absent', async () => {
         await processor.updateArtifacts(
           {
--- a/lib/modules/manager/pep621/processors/uv.spec.ts
+++ b/lib/modules/manager/pep621/processors/uv.spec.ts
@@ -14,6 +14,7 @@ import { getPkgReleases as _getPkgReleases } from '../../../datasource/index.ts'
 import { PypiDatasource } from '../../../datasource/pypi/index.ts';
 import type { UpdateArtifact, UpdateArtifactsConfig } from '../../types.ts';
 import { parsePyProject } from '../extract.ts';
+import { PyProject } from '../schema.ts';
 import { depTypes } from '../utils.ts';
 import { UvProcessor } from './uv.ts';
 
@@ -1293,6 +1294,17 @@ describe('modules/manager/pep621/processors/uv', () => {
         expect(execSnapshots[0].options?.env).toMatchObject({
           UV_EXCLUDE_NEWER: '2026-03-05T00:00:00.000Z',
         });
+      });
+
+      it('handles exclude-newer as Date object in schema', () => {
+        const result = PyProject.parse({
+          tool: {
+            uv: { 'exclude-newer': new Date('2026-03-05T00:00:00.000Z') },
+          },
+        });
+        expect(result.tool?.uv?.['exclude-newer']).toBe(
+          '2026-03-05T00:00:00.000Z',
+        );
       });
     });
   });
--- a/lib/modules/manager/pep621/processors/uv.spec.ts
+++ b/lib/modules/manager/pep621/processors/uv.spec.ts
@@ -897,9 +897,8 @@ describe('modules/manager/pep621/processors/uv', () => {
       let execSnapshots: ReturnType<typeof mockExecAll>;
 
       beforeEach(() => {
-        vi.spyOn(Date, 'now').mockReturnValue(
-          new Date('2026-03-13T00:00:00.000Z').getTime(),
-        );
+         vi.useFakeTimers();
+         vi.setSystemTime(new Date('2026-03-13T00:00:00.000Z'));
         execSnapshots = mockExecAll();
         GlobalConfig.set(adminConfig);
         fs.findLocalSiblingOrParent.mockResolvedValueOnce('uv.lock');
@@ -915,6 +914,7 @@ describe('modules/manager/pep621/processors/uv', () => {
 
       afterEach(() => {
         vi.restoreAllMocks();
+        vi.useRealTimers();
       });
 
       it('sets UV_EXCLUDE_NEWER from minimumReleaseAge', async () => {
--- a/lib/modules/manager/pep621/processors/uv.spec.ts
+++ b/lib/modules/manager/pep621/processors/uv.spec.ts
@@ -894,29 +894,31 @@ describe('modules/manager/pep621/processors/uv', () => {
     });
 
     describe('UV_EXCLUDE_NEWER with minimumReleaseAge', () => {
-      beforeEach(() => {
-        vi.restoreAllMocks();
-      });
+      let execSnapshots: ReturnType<typeof mockExecAll>;
 
-      it('sets UV_EXCLUDE_NEWER env var on lockfile maintenance with minimumReleaseAge', async () => {
+      beforeEach(() => {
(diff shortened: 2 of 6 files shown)


lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -215,6 +216,20 @@ export class UvProcessor extends BasePyProjectProcessor {
       } else {
         cmd = generateCMD(updatedDeps);
       }
+
+      if (config.minimumReleaseAge) {
+        const ms = toMs(config.minimumReleaseAge);
+        if (ms !== null && ms !== undefined) {
```
RahulGautamSingh: use `@sindresorhus/is` 's `isNullOrUndefined` method instead.

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -215,6 +216,20 @@ export class UvProcessor extends BasePyProjectProcessor {
...
+          const excludeNewer = new Date(Date.now() - ms).toISOString();
```
RahulGautamSingh: We have a DateTime utility please use that.

RahulGautamSingh (changes requested):
Please, use the PR template

Churro (changes requested):
I think this could benefit from some refinement:
* With uv, `--exclude-newer` can also be defined in `pyproject.toml` or `uv.toml` (see [here](https://docs.astral.sh/uv/reference/settings/#pip_exclude-newer)). A CLI value always takes precedence but this creates a problem if the `pyproject.toml` setting is not the same value as `minimumReleaseAge`:
  * If it is more restrictive (older date), a less restrictive renovate `minimumReleaseAge` would override it and potentially allow packages the user explicitly wanted to exclude.
  *  If it is less restrictive (newer date or absent), renovate correctly tightens the constraint.
  * **Suggestion:** Read `exclude-newer` from `pyproject.toml` and, if present, use the more restrictive (older) date of the two, i.e., `min(renovate_minimum_release_age, pyproject_exclude_newer_date)`.
* Using [`UV_EXCLUDE_NEWER`](https://docs.astral.sh/uv/reference/environment/#uv_exclude_newer) instead of `--exclude-newer` would be more consistent with the existing code style (`UV_EXTRA_URL`, `UV_INDEX_*` patterns).
* `--exclude-newer` works only since uv 0.2.22. Unlikely that someone still uses such old versions but, strictly speaking, it would be necessary to check if a uv constraint is set and if < 0.2.22 to omit this setting. The elegance of the approach via `UV_EXCLUDE_NEWER` env var is that incompatible uv versions would simply ignore the setting.
* Might also be worth a hint in renovate docs, considering that uv also explicitly exemplifies using this combination: https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependency-cooldown

lib/modules/manager/pep621/processors/uv.ts
thejoeejoee: fixed

lib/modules/manager/pep621/processors/uv.ts
thejoeejoee: fixed

thejoeejoee:
Thanks, all four points are addressed in the later commits — min() of both dates, UV_EXCLUDE_NEWER env var (so old uv versions just ignore it), and a docs note linking the uv dependency guide.

lib/modules/manager/types.ts
```diff
@@ -50,6 +50,7 @@ export interface UpdateArtifactsConfig {
   skipArtifactsUpdate?: boolean;
   lockFiles?: string[];
   toolSettings?: ToolSettingsOptions;
+  minimumReleaseAge?: string | null;
```
Churro: I don't think this is set to `null` anywhere. Shouldn't `minimumReleaseAge?: string` be enough?

Or perhaps even better would be to use the same type as here for consistency:
https://github.com/renovatebot/renovate/blob/9a8f3201b4c04f6170ee444ea53c42c2661366dd/lib/config/types.ts#L113

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -215,6 +217,39 @@ export class UvProcessor extends BasePyProjectProcessor {
...
         cmd = generateCMD(updatedDeps);
       }
+
+      // See https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependency-cooldown
+      if (config.minimumReleaseAge) {
+        const ms = toMs(config.minimumReleaseAge);
+        if (isNullOrUndefined(ms)) {
+          logger.debug(
+            { minimumReleaseAge: config.minimumReleaseAge },
+            'Invalid minimumReleaseAge value, skipping UV_EXCLUDE_NEWER for uv lock',
```
Churro: See https://github.com/renovatebot/renovate/blob/9a8f3201b4c04f6170ee444ea53c42c2661366dd/docs/development/best-practices.md?plain=1#L160

```suggestion
            `Invalid minimumReleaseAge value '${config.minimumReleaseAge}', skipping UV_EXCLUDE_NEWER for uv lock`,
```

docs/usage/key-concepts/minimum-release-age.md
```diff
@@ -44,6 +44,11 @@ To protect against this, it's recommended to ensure that your package manager co
 There is ongoing work to [integrate more closely with package manager checks](https://github.com/renovatebot/renovate/issues/41652) to make sure that Renovate's minimum release age configuration is specified when calling package managers that support it.
 If you have a package manager you'd like supported, please raise a [Suggest an Idea Discussion](https://github.com/renovatebot/renovate/discussions/new?category=suggest-an-idea).
 
+<!-- prettier-ignore -->
+!!! note
+    For uv, Renovate reads `[tool.uv] exclude-newer` from `pyproject.toml` and uses the more restrictive (older) date between it and `minimumReleaseAge`.
```
Churro: uv [supports](https://docs.astral.sh/uv/reference/settings/#exclude-newer):

> Accepts RFC 3339 timestamps (e.g., 2006-12-02T02:07:43Z), a "friendly" duration (e.g., 24 hours, 1 week, 30 days), or an ISO 8601 duration (e.g., PT24H, P7D, P30D).

ISO 8601 durations (e.g. "P14D") are not supported by this implementation (friendly dates only if added in this PR, see below) - ISO 8601 durations will be ignored, so please add a note to the docs.

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -215,6 +217,39 @@ export class UvProcessor extends BasePyProjectProcessor {
...
+              zone: 'utc',
+            });
+            if (pyprojectDate.isValid) {
+              if (pyprojectDate < excludeNewerDate) {
+                excludeNewerDate = pyprojectDate;
+              }
+            } else {
+              logger.debug(
+                { excludeNewer: pyprojectExcludeNewer },
+                'Invalid exclude-newer value in pyproject.toml, ignoring',
```
Churro:
```suggestion
                `Invalid exclude-newer value '${pyprojectExcludeNewer}' in pyproject.toml, ignoring`,
```

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -215,6 +217,39 @@ export class UvProcessor extends BasePyProjectProcessor {
...
+            'Invalid minimumReleaseAge value, skipping UV_EXCLUDE_NEWER for uv lock',
+          );
+        } else {
+          let excludeNewerDate = DateTime.now().minus(ms).toUTC();
+
+          const pyprojectExcludeNewer = project.tool?.uv?.['exclude-newer'];
+          if (pyprojectExcludeNewer) {
+            const pyprojectDate = DateTime.fromISO(pyprojectExcludeNewer, {
+              zone: 'utc',
+            });
```
Churro, line 240: `DateTime.fromISO()` works only with absolute timestamps but fails with exclude-newer = "2 weeks",  "30 days", etc. 

Add a fallback like this here:
```ts
            if (!pyprojectDate.isValid) {
              const durationMs = toMs(pyprojectExcludeNewer);
              if (!isNullOrUndefined(durationMs)) {
                pyprojectDate = DateTime.now().minus(durationMs).toUTC();
              }
            }
```

lib/modules/manager/pep621/processors/uv.spec.ts
```diff
@@ -891,5 +892,420 @@ describe('modules/manager/pep621/processors/uv', () => {
         },
       ]);
     });
+
+    describe('UV_EXCLUDE_NEWER with minimumReleaseAge', () => {
+      beforeEach(() => {
```
Churro:
```suggestion
      afterEach(() => {
```

Otherwise, `Date.now` will still be spied if there's a test coming after the `describe(...)`. Not now but probably in the future

Churro (changes requested):
Please revise the tests as a whole and make them meaningful.

There's no point in ~500 lines of AI slop just to satisfy code coverage. In particular visible with these:
- `sets UV_EXCLUDE_NEWER on per-package update`
   -> same path as `sets UV_EXCLUDE_NEWER env var on lockfile ...`,  only difference is --upgrade-package vs --upgrade, which is already tested elsewhere
- `does not set UV_EXCLUDE_NEWER on empty string minimumReleaseAge` 
    -> identical path to `does not set UV_EXCLUDE_NEWER when no minimumReleaseAge` ('' is falsy)
- `does not set UV_EXCLUDE_NEWER when no minimumReleaseAge even if pyproject has exclude-newer`
    -> same path as `does not set UV_EXCLUDE_NEWER when no minimumReleaseAge`, pyproject value is never reached
- `handles pyproject exclude-newer as local date without time component`
   -> same path as `uses pyproject exclude-newer when more restrictive...` , just different input format. Not a distinct code branch
- `handles exclude-newer as Date object in schema` 
   -> Schema test, not an `updateArtifacts` code path. Belongs in schema tests if anywhere.

**// EDIT: Here's a suggestion of what it could look like:**

```ts
    describe('UV_EXCLUDE_NEWER with minimumReleaseAge', () => {
      let execSnapshots: ReturnType<typeof mockExecAll>;

      beforeEach(() => {
        vi.spyOn(Date, 'now').mockReturnValue(
          new Date('2026-03-13T00:00:00.000Z').getTime(),
        );
        execSnapshots = mockExecAll();
        GlobalConfig.set(adminConfig);
        fs.findLocalSiblingOrParent.mockResolvedValueOnce('uv.lock');
        fs.readLocalFile.mockResolvedValueOnce('test content');
        fs.readLocalFile.mockResolvedValueOnce('changed test content');
        getPkgReleases.mockResolvedValueOnce({ releases: [{ version: '3.11.1' }] });
        getPkgReleases.mockResolvedValueOnce({ releases: [{ version: '0.2.35' }] });
      });

      afterEach(() => {
        vi.restoreAllMocks();
      });

      it('sets UV_EXCLUDE_NEWER from minimumReleaseAge', async () => {
        await processor.updateArtifacts(
          {
            packageFileName: 'folder/pyproject.toml',
            newPackageFileContent: '',
            config: { isLockFileMaintenance: true, minimumReleaseAge: '3 days' },
            updatedDeps: [],
          },
          parsePyProject('')!,
        );

        expect(execSnapshots[0].cmd).not.toContain('--exclude-newer');
        expect(execSnapshots[0].options?.env).toMatchObject({
          UV_EXCLUDE_NEWER: '2026-03-10T00:00:00.000Z',
        });
      });

      it('skips UV_EXCLUDE_NEWER when minimumReleaseAge is absent', async () => {
        await processor.updateArtifacts(
          {
            packageFileName: 'folder/pyproject.toml',
            newPackageFileContent: '',
            config: { isLockFileMaintenance: true },
            updatedDeps: [],
          },
          parsePyProject('')!,
        );

        expect(execSnapshots[0].options?.env?.UV_EXCLUDE_NEWER).toBeUndefined();
      });

      it('skips UV_EXCLUDE_NEWER on unparseable minimumReleaseAge', async () => {
        await processor.updateArtifacts(
          {
            packageFileName: 'folder/pyproject.toml',
            newPackageFileContent: '',
            config: { isLockFileMaintenance: true, minimumReleaseAge: 'invalid garbage' },
            updatedDeps: [],
          },
          parsePyProject('')!,
        );

        expect(execSnapshots[0].options?.env?.UV_EXCLUDE_NEWER).toBeUndefined();
        expect(logger.logger.debug).toHaveBeenCalledWith(
          "Invalid minimumReleaseAge value 'invalid garbage', skipping UV_EXCLUDE_NEWER for uv lock",
        );
      });

      it('uses pyproject ISO date when more restrictive than minimumReleaseAge', async () => {
        // pyproject (2026-03-05) is earlier than minimumReleaseAge (3 days = 2026-03-10)
        const pyproject = parsePyProject(codeBlock`
          [tool.uv]
          exclude-newer = "2026-03-05T00:00:00.000Z"
        `)!;

        await processor.updateArtifacts(
          {
            packageFileName: 'folder/pyproject.toml',
            newPackageFileContent: '',
            config: { isLockFileMaintenance: true, minimumReleaseAge: '3 days' },
            updatedDeps: [],
          },
          pyproject,
        );

        expect(execSnapshots[0].options?.env).toMatchObject({
          UV_EXCLUDE_NEWER: '2026-03-05T00:00:00.000Z',
        });
      });

      it('uses minimumReleaseAge when pyproject ISO date is less restrictive', async () => {
        // pyproject (2026-03-12) is later than minimumReleaseAge (3 days = 2026-03-10)
        const pyproject = parsePyProject(codeBlock`
          [tool.uv]
          exclude-newer = "2026-03-12T00:00:00.000Z"
        `)!;

        await processor.updateArtifacts(
          {
            packageFileName: 'folder/pyproject.toml',
            newPackageFileContent: '',
            config: { isLockFileMaintenance: true, minimumReleaseAge: '3 days' },
            updatedDeps: [],
          },
          pyproject,
        );

        expect(execSnapshots[0].options?.env).toMatchObject({
          UV_EXCLUDE_NEWER: '2026-03-10T00:00:00.000Z',
        });
      });

      it('uses pyproject relative duration when more restrictive than minimumReleaseAge', async () => {
        // 2 weeks before 2026-03-13 = 2026-02-27, earlier than 3 days (2026-03-10)
        const pyproject = parsePyProject(codeBlock`
          [tool.uv]
          exclude-newer = "2 weeks"
        `)!;

        await processor.updateArtifacts(
          {
            packageFileName: 'folder/pyproject.toml',
            newPackageFileContent: '',
            config: { isLockFileMaintenance: true, minimumReleaseAge: '3 days' },
            updatedDeps: [],
          },
          pyproject,
        );

        expect(execSnapshots[0].options?.env).toMatchObject({
          UV_EXCLUDE_NEWER: '2026-02-27T00:00:00.000Z',
        });
      });

      it('ignores unparseable pyproject exclude-newer and uses minimumReleaseAge', async () => {
        const pyproject = parsePyProject(codeBlock`
          [tool.uv]
          exclude-newer = "not-a-date"
        `)!;

        await processor.updateArtifacts(
          {
            packageFileName: 'folder/pyproject.toml',
            newPackageFileContent: '',
            config: { isLockFileMaintenance: true, minimumReleaseAge: '3 days' },
            updatedDeps: [],
          },
          pyproject,
        );

        expect(execSnapshots[0].options?.env).toMatchObject({
          UV_EXCLUDE_NEWER: '2026-03-10T00:00:00.000Z',
        });
        expect(logger.logger.debug).toHaveBeenCalledWith(
          "Invalid exclude-newer value 'not-a-date' in pyproject.toml, ignoring",
        );
      });
    });
```

lib/modules/manager/pep621/processors/uv.spec.ts
```diff
@@ -891,5 +892,183 @@ describe('modules/manager/pep621/processors/uv', () => {
...
+        );
+        expect(execSnapshots[0].options?.env).toMatchObject({
+          UV_EXCLUDE_NEWER: '2026-03-10T00:00:00.000Z',
+        });
+        expect(logger.logger.debug).toHaveBeenCalledWith(
+          "Invalid exclude-newer value 'not-a-date' in pyproject.toml, ignoring",
+        );
+      });
+
+      it('handles exclude-newer as Date object in schema', () => {
```
Churro: As mentioned before, I think this would better belong into a `schema.spec.ts`, as it tests the schema definition, not the implementation in `uv.ts`

docs/usage/key-concepts/minimum-release-age.md
```diff
@@ -44,6 +44,12 @@ To protect against this, it's recommended to ensure that your package manager co
 There is ongoing work to [integrate more closely with package manager checks](https://github.com/renovatebot/renovate/issues/41652) to make sure that Renovate's minimum release age configuration is specified when calling package managers that support it.
 If you have a package manager you'd like supported, please raise a [Suggest an Idea Discussion](https://github.com/renovatebot/renovate/discussions/new?category=suggest-an-idea).
 
+<!-- prettier-ignore -->
```
Churro, line 49: Nitpick but instead of this note box I think it would be better to add `### uv` since this is part of a broader effort (https://github.com/renovatebot/renovate/issues/41652) and there might be further sections upcoming for other managers too.

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -215,6 +217,43 @@ export class UvProcessor extends BasePyProjectProcessor {
...
+                excludeNewerDate = pyprojectDate;
+              }
+            } else {
+              logger.debug(
+                `Invalid exclude-newer value '${pyprojectExcludeNewer}' in pyproject.toml, ignoring`,
+              );
+            }
+          }
+
+          extraEnv.UV_EXCLUDE_NEWER = excludeNewerDate.toISO()!;
```
Churro:
```suggestion
          extraEnv.UV_EXCLUDE_NEWER = excludeNewerDate.toISO();
```

Missed that before but I believe this assertion is not needed. Neither at compile time, nor at runtime because `excludeNewerDate.toISO()` is always valid at this point.

Churro (changes requested):
Otherwise LGTM

docs/usage/key-concepts/minimum-release-age.md
RahulGautamSingh: ++

lib/modules/manager/pep621/processors/uv.spec.ts
```diff
@@ -891,5 +892,183 @@ describe('modules/manager/pep621/processors/uv', () => {
...
       ]);
     });
+
+    describe('UV_EXCLUDE_NEWER with minimumReleaseAge', () => {
+      let execSnapshots: ReturnType<typeof mockExecAll>;
+
+      beforeEach(() => {
+        vi.spyOn(Date, 'now').mockReturnValue(
+          new Date('2026-03-13T00:00:00.000Z').getTime(),
+        );
```
RahulGautamSingh:
```suggestion
         vi.useFakeTimers();
         vi.setSystemTime(new Date('2026-03-13T00:00:00.000Z'));
```

lib/modules/manager/pep621/processors/uv.spec.ts
```diff
@@ -891,5 +892,183 @@ describe('modules/manager/pep621/processors/uv', () => {
...
+        getPkgReleases.mockResolvedValueOnce({
+          releases: [{ version: '3.11.1' }],
+        });
+        getPkgReleases.mockResolvedValueOnce({
+          releases: [{ version: '0.2.35' }],
+        });
+      });
+
+      afterEach(() => {
+        vi.restoreAllMocks();
```
RahulGautamSingh, line 915:
```suggestion
        vi.restoreAllMocks();
        vi.useRealTimers();
```

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -215,6 +217,43 @@ export class UvProcessor extends BasePyProjectProcessor {
       } else {
         cmd = generateCMD(updatedDeps);
       }
+
+      // See https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependency-cooldown
+      if (config.minimumReleaseAge) {
```
RahulGautamSingh: We could also check `minimumReleaseAgeBehavior` here and only apply incase it's not `timestamp-optional` as simple API doesn't return any releaseTimestamp

RahulGautamSingh (changes requested):
LGTM except open comment and suggested changes.

RahulGautamSingh (commented):
It would be worth adding one test with a non-maintenance  update to confirm the env var is set for regular `--upgrade-package` commands too.

Churro:
From my perspective, the code is fine but only for the "happy paths":

1. Fresh lockfile - ✅ 
2. Existing lockfile, all within window - ✅ 
3. Existing lockfile, some deps too new - 🟥 needs fallback or graceful degradation
4. Lockfile convergence over time - 🟥 natural consequence of no. 3's fallback

I've built https://github.com/renovate-demo/41913-uv-lockfile-maintenance to demonstrate the lockfile maintenance case. When run with the code from this PR, we see in logs:

```
DEBUG: updateArtifacts for lockFileMaintenanceFiles (repository=renovate-demo/41913-uv-lockfile-maintenance, branch=renovate/lock-file-maintenance)
DEBUG: Executing command (repository=renovate-demo/41913-uv-lockfile-maintenance, branch=renovate/lock-file-maintenance)
       "command": "uv lock --upgrade"
DEBUG: rawExec err (repository=renovate-demo/41913-uv-lockfile-maintenance, branch=renovate/lock-file-maintenance)
       "err": {
         "cmd": "uv lock --upgrade",
         "stderr": "Using CPython 3.13.0\nResolving despite existing lockfile due to addition of global exclude newer 2025-03-28T11:29:06.431Z\n  × No solution found when resolving dependencies:\n  ╰─▶ Because only flask<=3.1.0 is available and your project depends on\n      flask>=3.1.3, we can conclude that your project's requirements are\n      unsatisfiable.\n",
         "stdout": "",
         "options": { ... },
         "exitCode": 1,
         "name": "ExecError",
       },
       "durationMs": 258
DEBUG: Failed to update uv lock file (repository=renovate-demo/41913-uv-lockfile-maintenance, branch=renovate/lock-file-maintenance)
```

So summarizing, I think a fallback similar to the one I proposed in https://github.com/renovatebot/renovate/pull/42552 is needed. Let's probably pause with this PR until a way forward is agreed upon regarding `npm`.

docs/usage/key-concepts/minimum-release-age.md
```diff
@@ -44,6 +44,13 @@ To protect against this, it's recommended to ensure that your package manager co
 There is ongoing work to [integrate more closely with package manager checks](https://github.com/renovatebot/renovate/issues/41652) to make sure that Renovate's minimum release age configuration is specified when calling package managers that support it.
 If you have a package manager you'd like supported, please raise a [Suggest an Idea Discussion](https://github.com/renovatebot/renovate/discussions/new?category=suggest-an-idea).
 
+#### uv
+
+<!-- prettier-ignore -->
+For uv, Renovate reads `[tool.uv] exclude-newer` from `pyproject.toml` and uses the more restrictive (older) date between it and `minimumReleaseAge`.
+The `exclude-newer` value can be an RFC 3339 timestamp (e.g. `2025-01-01T00:00:00Z`) or a friendly duration (e.g. `2 weeks`), but ISO 8601 durations (e.g. `P14D`) are not supported.
```
simonhaenisch, line 51: There was a recent change in uv, now you set `exclude-newer-span` for relative durations, and `exclude-newer` is for absolute timestamps only?

See https://github.com/astral-sh/uv/pull/19024.

PR #41918, opened 2026-03-13
feat(changelog): add generic changelog source

bdovaz:
## Changes

Create a new generic changelog source to handle cases where data sources such as `unity3d` return markdown directly in the `changelogUrl` field.

API example: https://services.api.unity.com/unity/editor/release/v1/releases?stream=LTS

A markdown example: https://storage.googleapis.com/live-platform-resources-prd/templates/assets/6000_3_11f1_500c5f352f/6000_3_11f1_500c5f352f.md

## Context

It comes from [here](https://github.com/renovatebot/renovate/discussions/40139) and [here](https://renovatebot.slack.com/archives/CAFH752JU/p1769957637309399)

## AI assistance disclosure

The public repository: <URL>


bdovaz:
I see that the `ChangeLogSource` class and everything related to it (repositories, tags, etc) are tightly coupled. And the generic approach “doesn't quite fit”:
1. In the constructor, I pass the data source as null because I’m not going to use it and I can’t (the Unity3D data source doesn’t resolve the repository because Unity is closed-source, so there are no tags or anything similar to resolve).
2. Because of the previous point, I haven’t been able to implement `getCompareURL`.
3. I implement `getAPIBaseUrl`, but in my case, it’s obviously “not an API base URL.”
4. I return `true` in `hasValidRepository`, but in reality, there isn’t a repository at all.

The `getChangeLogJSON` function (in index.ts) works with `sourceUrl`, but as I mentioned earlier, Unity is not open source and there is no source URL, nor does the API work that way, so it directly returns the Markdown in `changelogUrl`. Because of this, I had to make slight modifications to that function.

In the `getChangeLogJSON` function (from source.ts), I had to modify how it resolves the cache key due to the reasons mentioned above.

Despite all of the above, I've reached the point where the `(changelog)` link appears, which leads to the Markdown file: https://github.com/bdovaz/renovate_test/pull/26

I haven't been able to get it to render inline because the `sectionize` function in `release-notes.ts` doesn't seem to like Unity's Markdown. I guess because it isn't formatted correctly (it only has sections 3 and 4). But that's another issue for the future that could be addressed in a separate PR.

lib/workers/repository/update/pr/changelog/index.ts
```diff
@@ -12,24 +12,28 @@ export * from './types.ts';
...
-    if (!(sourceUrl && currentVersion && newVersion)) {
+    if (!((registryUrl || sourceUrl) && currentVersion && newVersion)) {
       return null;
     }
     const versioningApi = allVersioning.get(versioning);
     if (versioningApi.equals(currentVersion, newVersion)) {
       return null;
     }
+
+    const platformUrl = sourceUrl ?? registryUrl!;
```
viceice, line 26: why this is neeed 🤔

lib/workers/repository/update/pr/changelog/generic/source.ts
```diff
@@ -0,0 +1,25 @@
+import { BranchUpgradeConfig } from '../../../../../types.ts';
```
viceice: you probably need to rebase onto main or merge main and fix this

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -212,7 +213,9 @@ async function releaseNotesResult(
   }
   const { baseUrl, repository } = project;
   const releaseNotes: ChangeLogNotes = releaseMatch;
-  if (detectPlatform(baseUrl) === 'gitlab') {
+  if (detectPlatform(baseUrl) === 'generic') {
```
viceice: store `detectPlatform(baseUrl)` call into a variable and reuse (perf)

lib/workers/repository/update/pr/changelog/source.ts
```diff
@@ -89,6 +89,7 @@ export abstract class ChangeLogSource {
     const versioning = config.versioning!;
     const currentVersion = config.currentVersion!;
     const newVersion = config.newVersion!;
+    const changelogUrl = config.changelogUrl!;
```
viceice, line 92: Are you sure `changelogUrl` is always set here?

lib/workers/repository/update/pr/changelog/index.ts
bdovaz: As I mentioned in my original post:

> The getChangeLogJSON function (in index.ts) works with sourceUrl, but as I mentioned earlier, Unity is not open source and there is no source URL, nor does the API work that way, so it directly returns the Markdown in changelogUrl. Because of this, I had to make slight modifications to that function.

In other words, I try to use `sourceUrl` as usual, and if that fails, I fall back to `registryUrl`. Unless you can suggest a better option for me to use later in `detectPlatform`, this is the only solution I could come up with that I know will always be set.

In the case of Unity, in the link to an API call that I included in the PR description, you can see that the Markdown links have nothing to do with Unity. They’re hosted on Google Cloud.

Image

I can't use `changelogUrl` because it doesn't refer to anything in Unity, and I can't filter by it.

lib/workers/repository/update/pr/changelog/source.ts
bdovaz: I don't have much experience with the codebase, but at least I know that in the test I ran [mentioned in this comment](https://github.com/renovatebot/renovate/pull/41918#issuecomment-4058748765), it goes through this code and ends up generating the link with the text `(changelog)`.

I know that `sourceUrl` is empty in the case of the `unity3d` data source, so if `changelogUrl` were also empty in the subsequent calls to `getCacheKey`, the code would throw an error—but it doesn't.

bdovaz:
@rarkins I've made further changes to ensure that the changelog appears in the body of the PR, including all changelogs from the current version to the new one. Test PR: https://github.com/bdovaz/renovate_test/pull/26

The thing is, as you'll see, I've had to make a few exceptions because of the way Unity works.

bdovaz:
@viceice I completed what you asked me to do some weeks ago and have answered your questions. I'm waiting for more feedback.

I want to save the tests for last, once the implementation is correct.

bdovaz:
@RahulGautamSingh could someone please review this pull request? It's been on hold for several weeks without any feedback.

lib/workers/repository/update/pr/changelog/generic/source.ts
```diff
@@ -0,0 +1,25 @@
+import type { BranchUpgradeConfig } from '../../../../../types.ts';
+import { ChangeLogSource } from '../source.ts';
+
+export class GenericChangeLogSource extends ChangeLogSource {
+  constructor() {
+    super('generic', null!);
```
RahulGautamSingh, line 6: `null!` here bypasses the type system. `ChangeLogSource` calls `getAllTags` (uses `this.datasource`) and `getCompareURL` (which you override to throw) inside `getChangeLogJSON`. it works today only because Unity returns no tags so neither path fires.

imo the base class contract doesn't fit a raw URL fetcher. either make `datasource` optional in the base, or don't extend `ChangeLogSource` at all.

lib/workers/repository/update/pr/changelog/generic/source.ts
```diff
@@ -0,0 +1,25 @@
...
+    return config.changelogUrl!;
+  }
+
+  override getCompareURL(
+    baseUrl: string,
+    repository: string,
+    prevHead: string,
+    nextHead: string,
+  ): string {
+    throw new Error('Method not implemented.');
```
RahulGautamSingh, line 19: this will throw if ever called. `source.ts:186-193` calls it whenever `prevHead`/`nextHead` resolve to non-empty strings.

lib/workers/repository/update/pr/changelog/generic/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export async function getReleaseNotesMd(
+  releaseNotesUrl: string,
+  depName?: string,
+): Promise<ChangeLogFile | null> {
+  logger.trace('generic.getReleaseNotesMd()');
+
+  if (depName === 'Unity Editor') {
+    return null;
+  }
```
RahulGautamSingh, line 22: why is `generic` hardcoding Unity? if it's truly generic this check shouldn't exist.

lib/workers/repository/update/pr/changelog/generic/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    );
+
+    const release = response.body.results[0];
+
+    const changelogFile = await getReleaseNotesMd(release.releaseNotes.url);
+
+    changelogNotes.push({
+      url: release.releaseNotes.url,
+      notesSourceUrl: release.releaseNotes.url,
+      body: changelogFile!.changelogMd,
```
RahulGautamSingh, line 60: `changelogFile!` will NPE — `getReleaseNotesMd` returns `null` on empty body (line 28).

lib/workers/repository/update/pr/changelog/generic/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      tag: _release.version,
+    });
+  }
```
RahulGautamSingh, line 63: this whole function only does anything for `Unity Editor` / `m_EditorVersion` and otherwise returns `[]`. belongs in the unity3d datasource, not here.

RahulGautamSingh (commented):
the abstraction here doesn't really fit. it's named `generic` but contains hardcoded Unity branches in 4 places (`common.ts`, `generic/index.ts` x2, `release-notes.ts`). either make it actually generic (fetch the url, return the body, no `depName` checks) or rename to `unity` and move into the `unity3d` datasource module.

also CI has 4 failing tests, please check.

lib/util/common.ts
```diff
@@ -49,6 +50,9 @@ export function detectPlatform(
   if (hostname?.includes('forgejo')) {
     return 'forgejo';
   }
+  if (hostname === 'services.api.unity.com') {
+    return 'generic';
+  }
```
RahulGautamSingh, line 55: hardcoding `services.api.unity.com` in the platform detector leaks Unity into a generic util. push this into the unity3d module.

lib/workers/repository/update/pr/changelog/index.ts
```diff
@@ -12,24 +12,28 @@ export * from './types.ts';
 export async function getChangeLogJSON(
   config: BranchUpgradeConfig,
 ): Promise<ChangeLogResult | null> {
-  const { sourceUrl, versioning, currentVersion, newVersion } = config;
+  const { registryUrl, sourceUrl, versioning, currentVersion, newVersion } =
+    config;
   try {
-    if (!(sourceUrl && currentVersion && newVersion)) {
+    if (!((registryUrl || sourceUrl) && currentVersion && newVersion)) {
```
RahulGautamSingh, line 18: this changes behavior for *every* datasource, not just unity. now any package without `sourceUrl` but with a `registryUrl` will trigger `detectPlatform` on the registry url — which can match `github`/`gitlab`/`bitbucket` substrings and start fetching changelogs from the wrong place. needs regression coverage.

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -212,7 +215,10 @@ async function releaseNotesResult(
   }
   const { baseUrl, repository } = project;
   const releaseNotes: ChangeLogNotes = releaseMatch;
-  if (detectPlatform(baseUrl) === 'gitlab') {
+  const platform = detectPlatform(baseUrl);
+  if (platform === 'generic') {
+    releaseNotes.url = baseUrl;
```
RahulGautamSingh, line 220: `baseUrl` is computed by `ChangeLogSource.getBaseUrl` from `config.sourceUrl`. if `sourceUrl` is missing (the case this PR is enabling) `baseUrl` is `''` and the user gets an empty url.

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -367,6 +375,19 @@ export async function getReleaseNotesMd(
...
+  if (
+    project.depName === 'Unity Editor' &&
+    project.packageName === 'm_EditorVersion'
+  ) {
+    return {
+      body: await linkifyBody(project, changelogMd),
+      url: changelogUrl!,
+      notesSourceUrl: changelogUrl!,
+    };
+  }
```
RahulGautamSingh, line 389: same magic-string Unity check as `generic/index.ts:44`. duplicated logic — consolidate or move to unity3d.

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -383,12 +404,21 @@ export async function getReleaseNotesMd(
...
+              ? getNotesSourceUrl(baseUrl, repository, project, changelogFile)
+              : changelogUrl!;
+          } else {
+            notesSourceUrl = getNotesSourceUrl(
+              baseUrl,
+              repository,
+              project,
+              changelogFile,
+            );
+          }
```
RahulGautamSingh, line 420: both branches do the same thing when `repository` is truthy. collapses to `notesSourceUrl = repository ? getNotesSourceUrl(...) : changelogUrl!` without the type check.

lib/workers/repository/update/pr/changelog/generic/source.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  override hasValidRepository(repository: string): boolean {
+    return true;
+  }
+}
```
RahulGautamSingh, line 25: no spec for this file or `generic/index.ts`. please add.

PR #42032, opened 2026-03-20
fix(manager/flux): support Helm semver ranges for OCI HelmRepository sources

QuentinBtd:
## Changes
This PR fixes Flux `HelmRelease` extraction for OCI Helm repositories when `spec.chart.spec.version` uses a Helm range (e.g. `1.2.x`).
Previously, OCI sources were mapped to the Docker datasource, but range values like `1.2.x` were interpreted with Docker versioning and therefore not updated as expected.  
Now, for Flux OCI Helm sources, Renovate keeps `datasource=docker` but sets `versioning=helm` when the current value is a valid Helm range and not a plain Docker version.
Also added regression coverage:
- `extract.spec.ts`: OCI via `registryAliases` with `version: "1.2.x"`
- `extract.spec.ts` + new fixture `helmOCIReleaseRange.yaml`: OCI `HelmRepository` + `HelmRelease` with `version: 1.2.x`
## Context
## AI assistance disclosure

Used AI assistance to analyze Renovate Flux extraction logic, implement code changes, and draft/add regression tests and fixture updates.
The public repository: private Gerrit repository (yes, Gerrit...)
image

### DIFF
--- a/lib/modules/manager/flux/__fixtures__/helmOCIReleaseRange.yaml
+++ b/lib/modules/manager/flux/__fixtures__/helmOCIReleaseRange.yaml
@@ -1,0 +1,16 @@
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
+metadata:
+  name: arc-assets
+  namespace: dev
+spec:
+  interval: 30m
+  chart:
+    spec:
+      chart: actions-runner-controller-charts/gha-runner-scale-set
+      version: 0.4.x
+      sourceRef:
+        kind: HelmRepository
+        name: actions-runner-controller
+        namespace: flux-system
+      interval: 30m
--- a/lib/modules/manager/flux/extract.ts
+++ b/lib/modules/manager/flux/extract.ts
@@ -124,6 +124,31 @@ function resolveGitRepositoryPerSourceTag(
   }
 }
 
+/**
+ * Flux always treats `spec.chart.spec.version` as a Helm chart semver
+ * expression, including ranges, regardless of whether the chart is served
+ * from a classic Helm repository or an OCI registry:
+ * - https://github.com/fluxcd/source-controller/blob/main/internal/helm/repository/oci_chart_repository.go#L142-L144
+ * - https://github.com/fluxcd/source-controller/blob/main/internal/helm/repository/oci_chart_repository.go#L170-L171
+ * - https://github.com/fluxcd/source-controller/blob/main/internal/helm/repository/oci_chart_repository.go#L257-L277
+ *
+ * Renovate's Docker datasource defaults to `docker` versioning, which does
+ * not understand Helm/semver ranges such as `*`, `0.9.x` or `0.*` (all
+ * exercised in Flux's own test suite, see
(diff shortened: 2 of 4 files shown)


lib/modules/manager/flux/__fixtures__/helmOCIReleaseRange.yaml
secustor, line 1: Please inline this

lib/modules/manager/flux/extract.ts
```diff
@@ -120,6 +125,13 @@ function resolveHelmRepository(
         if (repo.spec.type === 'oci' || isOCIRegistry(repo.spec.url)) {
           // Change datasource to Docker
           dep.datasource = DockerDatasource.id;
+          if (
+            dep.currentValue &&
+            !dockerVersioningApi.isVersion(dep.currentValue) &&
+            helmVersioningApi.isValid(dep.currentValue)
+          ) {
+            dep.versioning = helmVersioningId;
+          }
```
secustor: I don't really like to have this logic here. 

Please share some references / docs how Flux interprets this versions. 
If the behavior differs enough from existing system, we would put this rather in a new `versioning` module

QuentinBtd: Thanks for the feedback - agreed that the logic should be justified.
The reason for using `versioning: helm` here is that Flux treats `spec.chart.spec.version` as a Helm chart **semver expression** (including ranges), even when the source is OCI.

- https://github.com/fluxcd/source-controller/blob/main/internal/helm/repository/oci_chart_repository.go#L142-L144  
  `GetChartVersion` says the version is expected to be a semver constraints-compatible string.
- https://github.com/fluxcd/source-controller/blob/main/internal/helm/repository/oci_chart_repository.go#L170-L171  
  If version is not a concrete semver, Flux interprets it as a semver range and finds the best matching tag.
- https://github.com/fluxcd/source-controller/blob/main/internal/helm/repository/oci_chart_repository.go#L257-L277  
  Uses `semver.NewConstraint(ver)` and `verConstraint.Check(v)`.
- https://github.com/fluxcd/source-controller/blob/main/internal/helm/repository/oci_chart_repository.go#L283  
  Error wording confirms constraint semantics: `could not locate a version matching provided version string %s`.
Related wiring (HelmRelease -> HelmChart):
- https://github.com/fluxcd/helm-controller/blob/main/internal/reconcile/helmchart_template.go#L226  
  `Version: template.Spec.Version` (passed through unchanged)
- https://github.com/fluxcd/helm-controller/blob/main/api/v2/helmrelease_types.go#L362-L366  
  HelmRelease template version is documented as a semver expression.
- https://github.com/fluxcd/source-controller/blob/main/api/v1/helmchart_types.go#L38-L42  
  HelmChart version is documented as a semver expression.
So OCI does not introduce different version syntax semantics in Flux; it remains Helm semver/range semantics.

secustor: The `*`, `0.9.x` and `0.*` pattern require a separate versioning 

https://github.com/fluxcd/source-controller/blob/5376a7ead3e636f6bef21dc7d06e9c2fe90808c6/internal/helm/repository/oci_chart_repository_test.go#L95-L184

PR #42051, opened 2026-03-22
feat(manager/npm): pass --before to npm install when minimumReleaseAge is set

Churro:
## Changes

When `minimumReleaseAge` is configured, Renovate now passes `--before=<date>` to `npm install` during lock file generation. This prevents npm from resolving newly published transitive dependencies that haven't yet passed the cooldown threshold.

`--before` is used rather than `--min-release-age` because `--before` has been available since npm 6 (released in 2018), whereas `--min-release-age` was only introduced in npm 11. 

If the project's `.npmrc` already contains a `before=` or `min-release-age=` setting, Renovate uses the stricter (earlier) of the two dates so that existing project-level constraints are never weakened.

## Context

- [x] This closes an existing Issue, Closes: #41657 

## AI assistance disclosure

The public repository: https://github.com/renovate-demo/41657-npm-lockfile-before
with `"minimumReleaseAge": "5 days"`

```
DEBUG: Generating package-lock.json for . (repository=renovate-demo/41657-npm-lockfile-before, branch=renovate/vite-8.x)
DEBUG: Spawning npm install to create ./package-lock.json (repository=renovate-demo/41657-npm-lockfile-before, branch=renovate/vite-8.x)
DEBUG: Updating lock file only (repository=renovate-demo/41657-npm-lockfile-before, branch=renovate/vite-8.x)
DEBUG: Setting npm --before=2026-03-17T14:14:08.740Z based on minimumReleaseAge=5 days (repository=renovate-demo/41657-npm-lockfile-before, branch=renovate/vite-8.x)
DEBUG: No node constraint found - using latest (repository=renovate-demo/41657-npm-lockfile-before, branch=renovate/vite-8.x)
DEBUG: Executing command (repository=renovate-demo/41657-npm-lockfile-before, branch=renovate/vite-8.x)
       "command": "npm install --package-lock-only --no-audit --ignore-scripts --before=2026-03-17T14:14:08.740Z"
DEBUG: exec completed (repository=renovate-demo/41657-npm-lockfile-before, branch=renovate/vite-8.x)
       "durationMs": 3499,
       "stdout": "\nup to date in 3s\n\n8 packages are looking for funding\n  run `npm fund` for details\n",
```

### DIFF
--- a/lib/modules/manager/npm/post-update/npm.ts
+++ b/lib/modules/manager/npm/post-update/npm.ts
@@ -148,7 +148,10 @@ export async function generateLockFile(
       const ms = toMs(config.minimumReleaseAge);
       if (ms === null) {
         logger.debug(
-          `Invalid minimumReleaseAge value: ${config.minimumReleaseAge}, skipping --before for npm install`,
+          {
+            minimumReleaseAge: config.minimumReleaseAge,
+          },
+          'Invalid minimumReleaseAge, skipping --before for npm install',
         );
       } else {
         let beforeDate = DateTime.now().minus(ms).toUTC();
@@ -156,14 +159,22 @@ export async function generateLockFile(
         const npmrcDate = parseNpmrcCooldownDate(npmrcContent);
         if (npmrcDate && npmrcDate < beforeDate) {
           logger.debug(
-            `Using stricter .npmrc cooldown date ${npmrcDate.toISO()} over minimumReleaseAge date ${beforeDate.toISO()}`,
+            {
+              npmrcDate: npmrcDate.toISO(),
+              beforeDate: beforeDate.toISO(),
+            },
+            'Using stricter .npmrc cooldown date over minimumReleaseAge date',
           );
           beforeDate = npmrcDate;
         }
 
         const beforeISO = beforeDate.toISO();
         logger.debug(
-          `Setting npm --before=${beforeISO} based on minimumReleaseAge=${config.minimumReleaseAge}`,
+          {
+            beforeISO,
+            minimumReleaseAge: config.minimumReleaseAge,
+          },
+          'Setting npm --before based on minimumReleaseAge',
         );
         cmdOptions += ` --before=${beforeISO}`;
       }
--- a/lib/modules/manager/npm/post-update/npm.ts
+++ b/lib/modules/manager/npm/post-update/npm.ts
@@ -47,7 +47,7 @@ const npmrcMinReleaseAgeRegex = regEx(
 
 export function parseNpmrcCooldownDate(
   npmrcContent: string | null,
-): DateTime | null {
+): DateTime<true> | null {
   const dateStr = npmrcContent?.match(npmrcBeforeRegex)?.[1];
   if (dateStr) {
     const parsed = DateTime.fromISO(dateStr, { zone: 'utc' });
--- a/lib/modules/manager/npm/post-update/npm.ts
+++ b/lib/modules/manager/npm/post-update/npm.ts
@@ -1,5 +1,6 @@
 // TODO: types (#22198)
 import { isNonEmptyString, isString } from '@sindresorhus/is';
+import { DateTime } from 'luxon';
 import semver from 'semver';
 import { quote } from 'shlex';
 import upath from 'upath';
@@ -22,6 +23,8 @@ import {
   renameLocalFile,
 } from '../../../../util/fs/index.ts';
 import { minimatch } from '../../../../util/minimatch.ts';
+import { toMs } from '../../../../util/pretty-time.ts';
+import { regEx } from '../../../../util/regex.ts';
 import { Result } from '../../../../util/result.ts';
 import { trimSlashes } from '../../../../util/url.ts';
 import type { PostUpdateConfig, Upgrade } from '../../types.ts';
@@ -34,6 +37,35 @@ import {
   getPackageManagerVersion,
   lazyLoadPackageJson,
 } from './utils.ts';
+
+const npmrcBeforeRegex = regEx(
+  /^\s*before\s*=\s*"?(\d{4}-\d{2}-\d{2}(?:T[\d:.]+Z?)?)"?(?:\s+[;#].*)?\s*$/m,
+);
+const npmrcMinReleaseAgeRegex = regEx(
+  /^\s*min-release-age\s*=\s*"?(\d+)"?(?:\s+[;#].*)?\s*$/m,
+);
+
+export function parseNpmrcCooldownDate(
+  npmrcContent: string | null,
+): DateTime | null {
+  const dateStr = npmrcContent?.match(npmrcBeforeRegex)?.[1];
+  if (dateStr) {
+    const parsed = DateTime.fromISO(dateStr, { zone: 'utc' });
+    if (parsed.isValid) {
+      return parsed;
+    }
+    logger.debug(`Invalid before date in .npmrc: ${dateStr}, ignoring`);
+  }
+
+  const daysStr = npmrcContent?.match(npmrcMinReleaseAgeRegex)?.[1];
+  if (daysStr) {
+    return DateTime.now()
+      .minus({ days: parseInt(daysStr, 10) })
+      .toUTC();
+  }
+
+  return null;
+}
 
 async function getNpmConstraintFromPackageLock(
   lockFileDir: string,
@@ -67,6 +99,7 @@ export async function generateLockFile(
   filename: string,
   config: Partial<PostUpdateConfig> = {},
   upgrades: Upgrade[] = [],
+  npmrcContent: string | null = null,
 ): Promise<GenerateLockFileResult> {
   // TODO: don't assume package-lock.json is in the same directory
   const lockFileName = upath.join(lockFileDir, filename);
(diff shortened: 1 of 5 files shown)


lib/modules/manager/npm/post-update/npm.ts
```diff
@@ -35,6 +38,35 @@ import {
...
+      return parsed;
+    }
+    logger.debug(`Invalid before date in .npmrc: ${dateStr}, ignoring`);
+  }
+
+  const daysStr = npmrcContent?.match(npmrcMinReleaseAgeRegex)?.[1];
+  if (daysStr) {
+    return DateTime.now()
+      .minus({ days: parseInt(daysStr, 10) })
+      .toUTC();
```
RahulGautamSingh, line 64: should add a debug log here too

lib/modules/manager/npm/post-update/npm.ts
```diff
@@ -111,6 +144,31 @@ export async function generateLockFile(
       cmdOptions += ' --ignore-scripts';
     }
 
+    if (config.minimumReleaseAge) {
+      const ms = toMs(config.minimumReleaseAge);
+      if (ms === null) {
```
RahulGautamSingh, line 149:
```suggestion
      if (isNull(ms)) {
```

lib/modules/manager/npm/post-update/npm.ts
```diff
@@ -111,6 +144,31 @@ export async function generateLockFile(
...
+        logger.debug(
+          `Invalid minimumReleaseAge value: ${config.minimumReleaseAge}, skipping --before for npm install`,
+        );
```
RahulGautamSingh, line 155:
```suggestion
        logger.debug( {minimumReleaseAge: config.minimumReleaseAge }
          `Invalid minimumReleaseAge, skipping --before for npm install`,
        );
```

lib/modules/manager/npm/post-update/npm.ts
```diff
@@ -111,6 +144,31 @@ export async function generateLockFile(
...
+      } else {
+        let beforeDate = DateTime.now().minus(ms).toUTC();
+
+        const npmrcDate = parseNpmrcCooldownDate(npmrcContent);
+        if (npmrcDate && npmrcDate < beforeDate) {
+          logger.debug(
+            `Using stricter .npmrc cooldown date ${npmrcDate.toISO()} over minimumReleaseAge date ${beforeDate.toISO()}`,
+          );
```
RahulGautamSingh, line 167: same change as the above debug log, we do this to make it easier to match the log statements for `logLevelRemap`

lib/modules/manager/npm/post-update/npm.ts
```diff
@@ -111,6 +144,31 @@ export async function generateLockFile(
...
+          beforeDate = npmrcDate;
+        }
+
+        const beforeISO = beforeDate.toISO();
+        logger.debug(
+          `Setting npm --before=${beforeISO} based on minimumReleaseAge=${config.minimumReleaseAge}`,
+        );
+        cmdOptions += ` --before=${beforeISO}`;
```
RahulGautamSingh, line 179: same change as above debug statement,

RahulGautamSingh (changes requested):
LGTM, except some nitpicks

lib/modules/manager/npm/post-update/npm.ts
```diff
@@ -111,6 +144,31 @@ export async function generateLockFile(
       cmdOptions += ' --ignore-scripts';
     }
 
+    if (config.minimumReleaseAge) {
```
RahulGautamSingh, line 147: We should also check if it's '0 days' or not, cause that is a valid input atm.

RahulGautamSingh (changes requested):

lib/modules/manager/npm/post-update/npm.ts
RahulGautamSingh: Let's wait for this to be merged/discussed first: https://github.com/renovatebot/renovate/pull/42053

lib/modules/manager/npm/post-update/npm.ts
Churro: No big ideal to change but I actually oriented this at the documented best practices:

https://github.com/renovatebot/renovate/blob/eddf500d24a86d72ba890060294212f80eef955b/docs/development/best-practices.md?plain=1#L160-L161

Do they need a refresh?

lib/modules/manager/npm/post-update/npm.ts
Churro: I'm curious to learn why you think this is better? `isNull()` is only used 5x throughout the entire renovate code base. Idiomatically `=== null` is standard syntax...

lib/modules/manager/npm/post-update/npm.ts
Churro: I didn't add one because the regex matches only \d+ so technically there's no invalid date to ignore

lib/modules/manager/npm/post-update/npm.ts
RahulGautamSingh: I am also not very sure. I suggested cause we use the `is` library.

lib/modules/manager/npm/post-update/npm.ts
RahulGautamSingh: I remember creating a PR for moving all log statement to display message and metadata separately. So, I think this message might need a update, will leave the decision to @viceice @jamietanna 

Till we agree on a approach, your change is fine.

RahulGautamSingh (approved):

lib/modules/manager/npm/post-update/npm.ts
jamietanna: Oh interesting - yes I think we should update the guidance - I can pick that up (unless you have proposed wording you'd like to start with @Churro)

Churro: I've changed it to use logger metadata now. As for best practice guidance, I'll leave that up to you, in case you might want to make further adjustments.

lib/modules/manager/npm/post-update/npm.ts
```diff
@@ -35,6 +38,35 @@ import {
...
+const npmrcBeforeRegex = regEx(
+  /^\s*before\s*=\s*"?(\d{4}-\d{2}-\d{2}(?:T[\d:.]+Z?)?)"?(?:\s+[;#].*)?\s*$/m,
+);
+const npmrcMinReleaseAgeRegex = regEx(
+  /^\s*min-release-age\s*=\s*"?(\d+)"?(?:\s+[;#].*)?\s*$/m,
+);
+
+export function parseNpmrcCooldownDate(
+  npmrcContent: string | null,
+): DateTime | null {
```
viceice:
```suggestion
): DateTime<true> | null {
```
should we add the type parameter to signal it only returns valid date objects?

lib/modules/manager/npm/post-update/npm.ts
```diff
@@ -111,6 +144,42 @@ export async function generateLockFile(
       cmdOptions += ' --ignore-scripts';
     }
 
+    if (config.minimumReleaseAge) {
+      const ms = toMs(config.minimumReleaseAge);
+      if (ms === null) {
+        logger.debug(
```
viceice, line 150: should this a warning, as it should never happen?

viceice (commented):
otherwise LGTM

lib/modules/manager/npm/post-update/npm.ts
Churro: It's likely to happen since `minimumReleaseAge` is not strictly validated before this point. In similar cases, when `toMs` returns `null` it gets silently coerced to 0 but this would be a bit pointless here.

If I pass `"minimumReleaseAge": "somefu days"` I get:
```
DEBUG: Invalid minimumReleaseAge, skipping --before for npm install (repository=renovate-demo/41657-npm-lockfile-before, branch=renovate/vite-8.x)
```

Still change to `logger.warn`?

viceice: oh, I thought they're validated when resolving the config. 🤔

@jamietanna @secustor wdyt about this? debug or warn log message?

secustor: I remember there has been a PR regarding the coercion  in the last few days. 

Tending towards debug level.

jamietanna:
There looks to be a separate attempt at this which we should also take a look at: https://github.com/renovatebot/renovate/pull/42145 (in case there's anything from each set of changes to take into account)

lib/modules/manager/npm/post-update/npm.ts
jamietanna: Perhaps - as a separate PR - we should add more strict validation to `minimumReleaseAge` as part of config validation? That'll surface this earlier, and we could make it a config error instead of warning to stop us even getting this far?

lib/modules/manager/npm/post-update/npm.ts
jamietanna: Out of interest, why do you feel this won't be straightforward to `logLevelRemap`?

Churro:
> There looks to be a separate attempt at this which we should also take a look at: #42145 (in case there's anything from each set of changes to take into account)

The approach in #42145 seems to address npm too but only on a very basic level. Nothing relevant for the scope of this PR.

lib/modules/manager/npm/post-update/npm.ts
Churro: Ok, then I'll mark this as resolved for now. Hope that works for you 😊

ZackKanter:
> > There looks to be a separate attempt at this which we should also take a look at: #42145 (in case there's anything from each set of changes to take into account)
> 
> The approach in #42145 seems to address npm too but only on a very basic level. Nothing relevant for the scope of this PR.

Thanks, I missed your PR when creating mine. I've rebased #42145 on your branch and added support for pnpm in c22b45f8bcbabdf812a1474dd9b3a9028f1b2ed6.

viceice (approved):

jamietanna:
(Aside: this didn't close the issue)

jamietanna:
Looks like we may have some unintended side effects: https://github.com/renovatebot/renovate/issues/41657#issuecomment-4142032301

viceice:
> 43.93.0

because the full url was used to link to issue



> Looks like we may have some unintended side effects: [#41657 (comment)](https://github.com/renovatebot/renovate/issues/41657#issuecomment-4142032301)

should we revert this?

jamietanna:
Yeah I think a revert is probably best - once that's released I'll ship that to the Developer Platform

PR #42054, opened 2026-03-22
feat: add customUpdateCommands to allow overwrite update logic

secustor:
## Changes

This is a draft for a new `customUpdateCommands` configuration. Which if applied overwrites the replacement logic. 

This iteration also supports `lockfileMaintenance`

## Context

This is useful in combination with CustomManager, if the a simply replacing is not enough as external tooling could use the packageFile as state ( which version is currently applied ) 

## AI assistance disclosure

The public repository: <URL>

TODOs:
- [ ] can we unify this with postUpgradeTasks? 
- [ ] or at least reuse some parts?


docs/usage/configuration-options.md
```diff
@@ -1363,6 +1363,113 @@ It will be compiled using Handlebars and the regex `groups` result.
...
+      }
+    }
+  ],
+  "allowedCommands": ["^yarn backstage-cli versions:bump"]
+}
+```
+
+The `customUpdateCommands` configuration consists of four fields:
+
+### customUpdateCommands.commands
```
jamietanna: I think this is now going to break, given https://github.com/renovatebot/renovate/pull/42204 😬 

(as the env var is possible through the existing `.commands`)

lib/config/options/index.ts
```diff
@@ -244,6 +244,16 @@ const options: Readonly<RenovateOptions>[] = [
     default: 'update',
     cli: false,
   },
+  {
+    name: 'customUpdateCommands',
```
jamietanna, line 249: Do we think this is "advanced" use functionality?

docs/usage/configuration-options.md
```diff
@@ -1363,6 +1363,113 @@ It will be compiled using Handlebars and the regex `groups` result.
...
+
+Each command must match at least one of the patterns defined in [`allowedCommands`](./self-hosted-configuration.md#allowedcommands) (a global-only configuration option) in order to be executed.
+If the list of allowed commands is empty then no commands will be executed.
+
+`customUpdateCommands` can be applied to specific dependencies using `packageRules`.
+For running commands _after_ a file has already been updated by Renovate, see [`postUpgradeTasks`](#postupgradetasks) instead.
+
+Example — using the Backstage CLI to bump all Backstage packages at once:
+
+```json
```
````suggestion
```json title="Example — using the Backstage CLI to bump all Backstage packages at once"
````

?

File: docs/usage/configuration-options.md [hunk 4]
```diff
@@ -1363,6 +1363,113 @@ It will be compiled using Handlebars and the regex `groups` result.
...
+
+All file changes produced by the commands (additions, modifications, and deletions) are captured via `git status` and committed by Renovate.
+This makes `customUpdateCommands` well-suited for tools like the [Backstage CLI](https://backstage.io/docs/tooling/cli/commands/#versionsbump), which updates `backstage.json`, `package.json`, `yarn.lock`, and other files in a single invocation.
+
+Each command must match at least one of the patterns defined in [`allowedCommands`](./self-hosted-configuration.md#allowedcommands) (a global-only configuration option) in order to be executed.
+If the list of allowed commands is empty then no commands will be executed.
+
+`customUpdateCommands` can be applied to specific dependencies using `packageRules`.
+For running commands _after_ a file has already been updated by Renovate, see [`postUpgradeTasks`](#postupgradetasks) instead.
+
```
jamietanna, line 1406: I'd also like to see an example that _isn't_ using a Custom Manager if possible - it may read as "this can only be used with Custom Managers"

docs/usage/configuration-options.md
```diff
@@ -1363,6 +1363,113 @@ It will be compiled using Handlebars and the regex `groups` result.
...
 
+## customUpdateCommands
+
+Update commands are commands that are executed by Renovate _instead of_ the regular file update (auto-replace) step.
+The intention is to handle cases where an external tool must be invoked to propagate a dependency update across multiple files — for example updating a lockfile, regenerating code, or running a CLI that modifies several files at once.
+
+All file changes produced by the commands (additions, modifications, and deletions) are captured via `git status` and committed by Renovate.
+This makes `customUpdateCommands` well-suited for tools like the [Backstage CLI](https://backstage.io/docs/tooling/cli/commands/#versionsbump), which updates `backstage.json`, `package.json`, `yarn.lock`, and other files in a single invocation.
+
+Each command must match at least one of the patterns defined in [`allowedCommands`](./self-hosted-configuration.md#allowedcommands) (a global-only configuration option) in order to be executed.
```
jamietanna, line 1401: Might be worth keeping a similar note to what we have re `Post-upgrade tasks are blocked by default for security reasons ...`?

docs/usage/configuration-options.md
```diff
@@ -1363,6 +1363,113 @@ It will be compiled using Handlebars and the regex `groups` result.
...
+    }
+  ],
+  "packageRules": [
+    {
+      "matchDepNames": ["backstage/backstage"],
+      "customUpdateCommands": {
+        "commands": [
+          "yarn backstage-cli versions:bump --release {{{newValue}}}"
+        ],
+        "fileFilters": [
```
jamietanna, line 1443: I guess we don't feel we need an `executionMode` here?

lib/config/migrations/migrations-service.ts
```diff
@@ -112,6 +112,7 @@ export class MigrationsService {
     ['masterIssueLabels', 'dependencyDashboardLabels'],
     ['regexManagers', 'customManagers'],
     ['baseBranches', 'baseBranchPatterns'],
+    ['updateCommands', 'customUpdateCommands'],
```
jamietanna: 🤔 I don't see us ever having one of these - accidental change?

lib/workers/repository/update/branch/execute-update-commands.ts
```diff
@@ -0,0 +1,182 @@
...
+      );
+    }
+
+    if (allowedCommands!.some((pattern) => regEx(pattern).test(compiledCmd))) {
+      try {
+        logger.trace({ cmd: compiledCmd }, 'Executing update command');
+
+        const execOpts: ExecOptions = {
+          shell: GlobalConfig.get(
+            'allowShellExecutorForPostUpgradeCommands',
```
jamietanna, line 70: If we're referencing this, can we make sure we update the docs for it to note it's used for both?

lib/workers/repository/update/branch/execute-update-commands.ts
```diff
@@ -0,0 +1,182 @@
...
+        { rawCmd: cmd, compiledCmd },
+        'Update command has been compiled',
+      );
+    }
+
+    if (allowedCommands!.some((pattern) => regEx(pattern).test(compiledCmd))) {
+      try {
+        logger.trace({ cmd: compiledCmd }, 'Executing update command');
+
+        const execOpts: ExecOptions = {
```
jamietanna, line 68: Can we keep the same warning as we have in `lib/workers/repository/update/branch/execute-post-upgrade-commands.ts`?

jamietanna (commented):
Few comments, but happy with the overall design and approach - especially as it's very close code wise to how we're doing `postUpgradeTasks` (which I know is intentional)

jamietanna:
IMO, let's keep the duplication in place between them - I prefer the "rule of 3" rather than "don't repeat yourself" - there are a few things that are different between them, and we won't know until - maybe - we add a third thing like this where the duplication is and what's special to each case

docs/usage/configuration-options.md
secustor: Hmm should be fine as it is child 🤔

lib/config/options/index.ts
secustor: I would classify it as that.   
We can also migrate it internally from `customUpdateCommands` to `updateCommands` so we have the option to use them in manager. 

Potentially with inspiration from this [hook proposal](https://github.com/renovatebot/renovate/discussions/35198).

docs/usage/configuration-options.md
secustor: Make sense, I will try to come up with one.

docs/usage/configuration-options.md
secustor: The expectation this would always run against a single dependency, like we are editing the file for each update. Though for sure users will probably find an edge case for this.

lib/config/migrations/migrations-service.ts
secustor: I had it first named `updateCommands` and then pivoted to `customUpdateCommands`

docs/usage/configuration-options.md
secustor: We haven't such a note in `postUpgradeTasks`, but we can add such one to `allowedCommands`

lib/workers/repository/update/branch/execute-update-commands.ts
secustor: Do you mean? 
 https://github.com/renovatebot/renovate/blob/cde4f00c786011ff3f3cae133dd0299db5eb4b01/lib/workers/repository/update/branch/execute-post-upgrade-commands.ts#L136-L137 

That should be rather added to the documentation rather code.

PR #42068, opened 2026-03-23
fix(scm-manager): invalid base url, due to double slashes and a fixed context path

zT-1337:
## Changes

The Base URL for the SCM-Manager was constructed with this line:
`const baseUrl = `${endpoint}/scm/api/v2;`

There are two issues with this line of code.
First it falsely assumes that `endpoint` will not have a trailing slash, resulting in double slashes within the `baseUrl`.
Secondly the first segment of the path (`/scm`) is configurable and can be changed to any valid path, therefore this part can not be hardcoded and needs to be part of the `endpoint` configuration.

This PR fixes this two errors.

## Context

## AI assistance disclosure

### DIFF
--- a/lib/modules/platform/scm-manager/index.ts
+++ b/lib/modules/platform/scm-manager/index.ts
@@ -60,7 +60,7 @@ export async function initPlatform({
     throw new Error('Init: SCM-Manager API token not configured');
   }
 
-  const baseUrl = `${endpoint}scm/api/v2`;
+  const baseUrl = `${endpoint}api/v2`;
   setBaseUrl(baseUrl);
 
   try {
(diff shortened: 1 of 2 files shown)


lib/modules/platform/scm-manager/index.ts
```diff
@@ -60,7 +60,7 @@ export async function initPlatform({
     throw new Error('Init: SCM-Manager API token not configured');
   }
 
-  const baseUrl = `${endpoint}/scm/api/v2`;
+  const baseUrl = `${endpoint}api/v2`;
```
viceice: use our `ensureTrailingSlash` util, so it doesn't matter if it's there or not

zT-1337: Do you mean like this?
```
const baseUrl = ensureTrailingSlash(`${endpoint}api/v2`);
```

viceice (approved):

PR #42083, opened 2026-03-23
fix(http): detect x-access-token prefix before hostType-based auth branching

zharinov:
## Changes

When a GitHub App installation token (`ghs_*`) is used, the platform init prefixes it with `x-access-token:`. The `applyAuthorization` function previously only detected this prefix inside the `GITHUB_API_USING_HOST_TYPES` branch, making it unreachable for datasources not in that list (e.g. `github-digest`). This resulted in sending `Bearer x-access-token:ghs_*` instead of `token ghs_*`, causing `401 Bad credentials`.

This PR moves the `x-access-token:` prefix detection before the hostType-based branching chain, so it works for any hostType — current and future.

## Context

- [x] Closes: #42077

## AI assistance disclosure

### DIFF
--- a/lib/util/http/auth.ts
+++ b/lib/util/http/auth.ts
@@ -36,6 +36,17 @@ export function applyAuthorization<GotOptions extends AuthGotOptions>(
       } else {
         options.headers.authorization = `${authType} ${options.token}`;
       }
+    } else if (options.token.startsWith('x-access-token:')) {
+      // GitHub App installation token — prefix is set by the GitHub platform init
+      const appToken = options.token.replace('x-access-token:', '');
+      options.headers.authorization = `token ${appToken}`;
+      // v8 ignore else -- TODO: add test #40625
+      if (isString(options.headers.accept)) {
+        options.headers.accept = options.headers.accept.replace(
+          'application/vnd.github.v3+json',
+          'application/vnd.github.machine-man-preview+json',
+        );
+      }
     } else if (
       options.hostType &&
       FORGEJO_API_USING_HOST_TYPES.includes(options.hostType)
@@ -53,17 +64,6 @@ export function applyAuthorization<GotOptions extends AuthGotOptions>(
       GITHUB_API_USING_HOST_TYPES.includes(options.hostType)
     ) {
       options.headers.authorization = `token ${options.token}`;
-      if (options.token.startsWith('x-access-token:')) {
-        const appToken = options.token.replace('x-access-token:', '');
-        options.headers.authorization = `token ${appToken}`;
-        // v8 ignore else -- TODO: add test #40625
-        if (isString(options.headers.accept)) {
-          options.headers.accept = options.headers.accept.replace(
-            'application/vnd.github.v3+json',
-            'application/vnd.github.machine-man-preview+json',
-          );
-        }
-      }
     } else if (
       options.hostType &&
       GITLAB_API_USING_HOST_TYPES.includes(options.hostType)
--- a/lib/util/http/auth.ts
+++ b/lib/util/http/auth.ts
@@ -1,4 +1,4 @@
-import { isNonEmptyString, isString } from '@sindresorhus/is';
+import { isNonEmptyString } from '@sindresorhus/is';
 import {
   FORGEJO_API_USING_HOST_TYPES,
   GITEA_API_USING_HOST_TYPES,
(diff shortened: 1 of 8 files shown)


jamietanna:
Can we please show evidence that we've managed to test this with a GitHub App and validated it resolves the issue?

Although this does fix, I'd like to see if (maybe as a follow-up) we can find a more resilient fix - I'd expect there'd be something we can do that works across other all GitHub datasources appropriately

zharinov:
Doing another way of fixing it, and trying to test...

zharinov:
The public repository: https://github.com/renovate-testing/github-actions-refs-test

### Reproduction evidence

Created a GitHub App (`renovate-testing-fix-42077`), installed it on the test repo, and ran Renovate locally with the installation token.

**Without fix** — `github-digest` sends `Bearer x-access-token:ghs_*` → 401:
```
authHostType: "github-digest"
authHeader:   "Bearer x-access-token:ghs_bhuJ..."
→ 401 Unauthorized: "Bad credentials"
  datasource: "github-digest", packageName: "peter-evans/create-pull-request"
```

**With fix** — `github-digest` sends `token ghs_*` → success:
```
authHostType: "github-digest"
authHeader:   "token ghs_ftLtk2dNQVlLgIRZYuLq..."
→ Repository finished successfully
```

Note: the bug only manifests when `github-digest` queries a repo that has no corresponding `github-tags` lookup (they share a GraphQL-level memory cache). The test repo was updated with `peter-evans/create-pull-request@main` — a branch-only dep targeting a unique repo — to trigger the issue.

jamietanna (dismissed):
Good stuff, and a better approach now

Can you please also get a PR raised to add a test for this? We can merge this as-is, but if you can create a test branched off before this PR was raised (to validate it'd catch this issue) and then when `main` is merged into your branch, it's fixed the issue, that'd be great

jamietanna:
I'd like a second review from @renovatebot/maintainers to double-check if there are any side effects we may need to watch out for here

RahulGautamSingh (approved):
LGTM

lib/util/http/auth.ts
```diff
@@ -36,6 +36,17 @@ export function applyAuthorization<GotOptions extends AuthGotOptions>(
       } else {
         options.headers.authorization = `${authType} ${options.token}`;
       }
+    } else if (options.token.startsWith('x-access-token:')) {
+      // GitHub App installation token — prefix is set by the GitHub platform init
+      const appToken = options.token.replace('x-access-token:', '');
+      options.headers.authorization = `token ${appToken}`;
```
viceice: i think we should simply use standarized `Bearer` instead of `token`

- https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2026-03-10#about-authentication

viceice (commented):
we should drop `x-access-token` handling. it should no longer be used at all.

- #38952

as a starter we can detect it on config validation and migrate and warn.
later we can emit an error.

jamietanna:
I agree that we should drop it - when I looked at it, it was a chunk of work (with a large opportunity to break things) so we didn't get it in before v43

We probably shouldn't config migrate warn it - it's Renovate setting that prefix, rather than the user themselves

lib/util/http/auth.ts
secustor: Do we know why this is here? I can not find any references for this on the Rest API in the past. 
The only ones I can get my hands on is for cloning via GitHub app installation tokens. 

> You can also use an installation access token to authenticate for HTTP-based Git access. Your app must have the "Contents" repository permission. You can then use the installation access token as the HTTP password. Replace TOKEN with the installation access token: git clone https://x-access-token:TOKEN@github.com/owner/repo.git.

https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation#about-authentication-as-a-github-app-installation

lib/util/http/auth.ts
```diff
@@ -53,17 +57,6 @@ export function applyAuthorization<GotOptions extends AuthGotOptions>(
...
     ) {
       options.headers.authorization = `token ${options.token}`;
-      if (options.token.startsWith('x-access-token:')) {
-        const appToken = options.token.replace('x-access-token:', '');
-        options.headers.authorization = `token ${appToken}`;
-        // v8 ignore else -- TODO: add test #40625
-        if (isString(options.headers.accept)) {
-          options.headers.accept = options.headers.accept.replace(
-            'application/vnd.github.v3+json',
-            'application/vnd.github.machine-man-preview+json',
```
viceice, line 63: is that header no longer needed?

viceice (commented):
otherwise LGTM

lib/util/http/auth.ts
zharinov: Yeah, it's not the "preview" the last 6 years 🙂

secustor (approved):
LGTM based on the current docs

lib/util/http/auth.ts
jamietanna: https://developer.github.com/changes/2020-08-20-graduate-machine-man-and-sailor-v-previews/

jamietanna:
Let's merge this and see if we can get some feedback on whether this resolves things - if things are looking good later this week, we'll ship this to the Mend Developer Platform

jamietanna:
@zharinov to confirm, have you re-tested this works since the latest changes (`Bearer ...`)?

zharinov:
Tested, all good

PR #42112, opened 2026-03-24
fix(config): invert registration of manager objects

jamietanna:
Co-authored-by: Claude Sonnet 4.6 <jamie.tanna+claude-code@mend.io>

## Changes

As part of future changes, we need to reduce the risk of a dependency
cycle in `config/options`, which requires removing the dependency on the
Manager APIs.

This flips this requirement by making sure that when the managers are
initialised, we then load the configuration.

Using `fix(...)` to release this and get faster feedback if it's breaking.

## Context

## AI assistance disclosure

- [ ] Yes — other (please describe): Claude Sonnet 4.6 (Claude Code)

The public repository: <URL>


lib/modules/manager/index.ts
```diff
@@ -18,6 +19,8 @@ export const getManagerList = (): string[] => managerList;
 export const getManagers = (): Map<string, ManagerApi> => managers;
 export const allManagersList = [...managerList, ...customManagerList];
 
+registerManagerOptions(new Map([...managers, ...customManagers]));
```
viceice, line 22: we should get rid of this somehow

viceice (commented):
lgtm

lib/modules/manager/index.ts
jamietanna: As in, to stop having the default manager config registered? I assume we do need it for some things 🤔

jamietanna:
Any thoughts on other ways to handle this, or ways that we could avoid the ordering needed - i.e. will this meant that `tools/` need to import the manager API to get the default options registered? 🤔

viceice:
> Any thoughts on other ways to handle this, or ways that we could avoid the ordering needed - i.e. will this meant that `tools/` need to import the manager API to get the default options registered? 🤔 

tools are already doing that. I wonder how can we get the manager default options without adding them to the options.

one idea would be to move them to a lazy loaded core preset which is resolved in global worker when initializing the config 🤔

viceice (commented):
we need to check if manager API is still loaded early enough to be registered before we call the config loading.

secustor (commented):
LGTM, though this needs definitively testing against a real repo

jamietanna:
Replaced by https://github.com/renovatebot/renovate/pull/43048

PR #42135, opened 2026-03-25
feat(gradle): add `!!`-notation support for strictly and prefer constraints

ldt-fweichert:
## Changes

Adds Support for gradles `!!` shorthand notation for describing both a `strictly` and a `preferred` version constraint. The behavior is briefly explained in the [official documentation](https://docs.gradle.org/current/userguide/dependency_versions.html#sec:rich-version-constraints).

The syntax is: `<strict version (range)>!!<preferred version>`

Using this is very common in big multi-module projects as it 1.: allows for loose constraints enabling easy updates and 2.: stabilizes the dependency-tree by using a specific preferred version if possible.

The current implementation of this PR updates the preferred version to be exactly the `newVersion` only if it the original preferred version was also referenced in the original strictly version part.

I am open to suggestions about changing this specific behavior. Important to the usecase is just that `[1.2.3, )!!1.2.3` gets updated to `[1.3.0, )!!1.3.0` instead of beeing ignored entirely.

(This is somewhat of a late follow-up to [this PR](https://github.com/renovatebot/renovate/pull/33453))

## Context

## AI assistance disclosure

The public repository: https://github.com/ldt-fweichert/renotave-gradle-repro


viceice (commented):
LGTM. @Churro can you have a look?

lib/modules/versioning/gradle/index.ts
```diff
@@ -190,6 +190,10 @@ function minSatisfyingVersion(
   }, null);
 }
 
+const strictlyShorthandRegex = regEx(
+  /^(?<baseRange>[^!]+)!!(?<preferredVal>[-._+a-zA-Z0-9]+)$/,
+);
+
```
Churro:
```suggestion

```

lib/modules/versioning/gradle/index.ts
```diff
@@ -220,6 +224,30 @@ function getNewValue({
     }
   }
 
+  const strictlyShorthandMatchGroups =
+    strictlyShorthandRegex.exec(currentValue)?.groups;
+  if (strictlyShorthandMatchGroups) {
+    const { baseRange, preferredVal } = strictlyShorthandMatchGroups;
```
Churro:
```suggestion
  const mavenRange = parseMavenBasedRange(currentValue);
  if (mavenRange?.preferredVal) {
    const { leftVal, rightVal, preferredVal } = mavenRange;
    const baseRange = currentValue.slice(
      0,
      currentValue.lastIndexOf(`!!${preferredVal}`),
    );
```

lib/modules/versioning/gradle/index.ts
```diff
@@ -220,6 +224,30 @@ function getNewValue({
...
+    if (newBaseRange === null) {
+      return null;
+    }
+
+    let newPreferredVal = preferredVal;
+    const preferredValueWasReferenced =
+      baseRange.includes(preferredVal) && !newBaseRange.includes(preferredVal);
+    if (preferredValueWasReferenced) {
+      newPreferredVal = newVersion;
+    }
```
Churro:
```suggestion
    const preferredIsBoundary =
      preferredVal === leftVal || preferredVal === rightVal;
    const newParsed = parseMavenBasedRange(newBaseRange);
    const preferredStillPresent =
      newParsed?.leftVal === preferredVal ||
      newParsed?.rightVal === preferredVal;
    const newPreferredVal =
      preferredIsBoundary && !preferredStillPresent
        ? newVersion
        : preferredVal;
```

Churro (changes requested):
Only one thing that's probably an optimization:

* By adding `!!` to `parseMavenBasedRange` in `compare.ts`, `isValid()` and `matches()` can now also handle these constraints. 
* The `strictlyShorthandRegex` in `getNewValue()` would then introduce a second, independent way to split out the `!!` suffix. To me this seems a bit redundant, since `parseMavenBasedRange` already returns `preferredVal`, so `getNewValue()` could reuse it directly.

Otherwise your change is fine. I didn't spot any potential side effects.

lib/modules/versioning/gradle/index.ts
ldt-fweichert: applied suggestion

lib/modules/versioning/gradle/index.ts
ldt-fweichert: applied suggestion

lib/modules/versioning/gradle/index.ts
ldt-fweichert: applied suggestion

ldt-fweichert:
Thanks for the quick feedback. I've applied the suggestion

Churro (approved):

lib/modules/versioning/gradle/compare.ts
```diff
@@ -232,6 +232,9 @@ interface MavenBasedRange {
   rightBound: RangeBound;
   rightBoundStr: string;
   rightVal: string | null;
+  // The existence of preferredVal infers the "strictly" keyword "!!"
```
RahulGautamSingh:
```suggestion
  // The existence of preferredVal implies the "strictly" keyword "!!"
```

lib/modules/versioning/gradle/index.ts
```diff
@@ -220,6 +220,35 @@ function getNewValue({
...
+    const baseRange = currentValue.slice(
+      0,
+      currentValue.lastIndexOf(`!!${preferredVal}`),
+    );
+    const newBaseRange = mavenVersion.getNewValue({
+      currentValue: baseRange,
+      rangeStrategy,
+      newVersion,
+    });
+    // istanbul ignore if: the implementation has a non-null return type
```
RahulGautamSingh:
```suggestion
    // v8 ignore if: the implementation has a non-null return type
```

lib/modules/versioning/gradle/index.spec.ts
RahulGautamSingh, line 1: Missing pin strategy tests for `!!`

RahulGautamSingh (changes requested):

lib/modules/versioning/gradle/index.spec.ts
ldt-fweichert: Good catch! There was actually a bug where pin strategy would still try to update the preferred version. I've changed the implementation so that pinning disables the preferred version update entirely as it makes no sense to have a preferred version in addition to a pinned one.

lib/modules/versioning/gradle/index.ts
```diff
@@ -220,6 +220,37 @@ function getNewValue({
     }
   }
 
+  const mavenRange = parseMavenBasedRange(currentValue);
+  // if there is a preferred version, handle it. Ignore pin strategy, where
+  // having a preferred version wouldn't make any sense.
+  if (mavenRange?.preferredVal && rangeStrategy !== 'pin') {
```
viceice:
```suggestion
  if (mavenRange?.preferredVal) {
```

`pin` is never / no longer passed as range strategy to `getNewValue`

- #36261

ldt-fweichert: I've removed the pin check and range strategy entirely

viceice (approved):

PR #42142, opened 2026-03-25
docs: explicitly document `depTypes` each manager supports

jamietanna:
## Changes

A long-standing question we receive is "what `depTypes` can I match
against in my project".

Although the best recommendation we usually have is "check the
`packageFiles with updates" log line (as a very key log line for users),
we can go one step further and try and create a stronger contract.

To do this, I've used Claude Sonnet 4.6 to work through the codebase,
determine areas that a `depType` is set in a manager (statically or
dynamically) and generate a structured, typed, means to document how
they're used.

Through a few iterations of the process, and attempts at human-written
documentation, we've settled on:

- prioritising autogenerated documentation (for humans to read) to note
  what `depTypes` are known, and what they are
- `knownDepTypes`, to describe any known, static, `depType`s and their
  human-readable meaning
  - with the `prettyDepType`, if used
- `supportsDynamicDepTypesNote`, to add a Markdown note to the generated
  documentation

While here, we also:

- make sure to note this in the "adding a new package manager"
  documentation
- remove any hand-written references to `depTypes` in each manager's
  documentation page
- correct the Gleam `depTypes`
- make sure to use the `knownDepTypes`'s `prettyDepType` definition if
  we know it, for a given `depType`

The descriptions have been generated by Claude Sonnet, with some
human-written tweaks where appropriate.

[0]: https://github.com/renovatebot/renovate/discussions/15176

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/manager/ansible-galaxy/dep-types.ts
+++ b/lib/modules/manager/ansible-galaxy/dep-types.ts
@@ -1,0 +1,12 @@
+import type { DepTypeMetadata } from '../types.ts';
+
+export const knownDepTypes = [
+  {
+    depType: 'galaxy-collection',
+    description: 'Ansible Galaxy collection',
+  },
+  {
+    depType: 'role',
+    description: 'Ansible role',
+  },
+] as const satisfies readonly DepTypeMetadata[];
--- a/lib/modules/manager/bazel/dep-types.ts
+++ b/lib/modules/manager/bazel/dep-types.ts
@@ -1,0 +1,76 @@
+import type { DepTypeMetadata } from '../types.ts';
+
+export const knownDepTypes = [
+  {
+    depType: 'container_pull',
+    description: 'A Docker container image pulled via `container_pull`',
+  },
+  {
+    depType: '_container_pull',
+    description:
+      'A Docker container image pulled via the private `_container_pull` rule',
+  },
+  {
+    depType: 'git_repository',
+    description: 'A Git repository dependency via `git_repository`',
+  },
+  {
+    depType: '_git_repository',
+    description:
+      'A Git repository dependency via the private `_git_repository` rule',
+  },
+  {
+    depType: 'new_git_repository',
+    description:
+      'A Git repository with a custom BUILD file via `new_git_repository`',
+  },
+  {
+    depType: '_new_git_repository',
+    description:
+      'A Git repository with a custom BUILD file via the private `_new_git_repository` rule',
+  },
+  {
+    depType: 'go_repository',
+    description: 'A Go module dependency via `go_repository`',
+  },
+  {
+    depType: '_go_repository',
+    description: 'A Go module dependency via the private `_go_repository` rule',
+  },
+  {
+    depType: 'http_archive',
+    description: 'An archive downloaded over HTTP via `http_archive`',
+  },
+  {
+    depType: '_http_archive',
+    description:
+      'An archive downloaded over HTTP via the private `_http_archive` rule',
+  },
+  {
+    depType: 'http_file',
+    description: 'A file downloaded over HTTP via `http_file`',
+  },
+  {
+    depType: '_http_file',
+    description:
+      'A file downloaded over HTTP via the private `_http_file` rule',
+  },
+  {
+    depType: 'oci_pull',
+    description: 'An OCI container image pulled via `oci_pull`',
+  },
+  {
+    depType: '_oci_pull',
+    description:
+      'An OCI container image pulled via the private `_oci_pull` rule',
+  },
+  {
+    depType: 'maven_install',
+    description: 'A Maven artifact installed via `maven_install`',
+  },
+  {
+    depType: '_maven_install',
+    description:
+      'A Maven artifact installed via the private `_maven_install` rule',
+  },
+] as const satisfies readonly DepTypeMetadata[];
--- a/lib/modules/manager/gleam/readme.md
+++ b/lib/modules/manager/gleam/readme.md
@@ -1,10 +1,5 @@
 Renovate can update `gleam.toml` and/or `manifest.toml` files.
 
-The `gleam` manager can update these `depTypes`:
-
-- `dependencies`
-- `dev-dependencies`
-
 ### How Renovate updates `gleam.toml` files
 
 The `gleam` manager extracts dependencies for the `hex` datasource, and uses Renovate's implementation of Hex SemVer to evaluate `gleam.toml` updates.
--- a/lib/modules/manager/gomod/dep-types.ts
+++ b/lib/modules/manager/gomod/dep-types.ts
@@ -1,0 +1,30 @@
+import type { DepTypeMetadata } from '../types.ts';
+
+export const knownDepTypes = [
+  {
+    depType: 'golang',
+    description:
+      'The `go` directive specifying the minimum version of Go that the code is source compatible with',
+  },
+  {
+    depType: 'toolchain',
+    description:
+      'The `toolchain` directive specifying the Go toolchain version',
+  },
+  {
+    depType: 'require',
+    description: 'A direct module dependency',
+  },
+  {
+    depType: 'indirect',
+    description: 'An indirect (transitive) module dependency',
+  },
+  {
+    depType: 'replace',
+    description: 'A module replacement directive',
+  },
+  {
+    depType: 'tool',
+    description: `A tool dependency using Go 1.24+'s \`tool\` directive`,
+  },
+] as const satisfies readonly DepTypeMetadata[];
--- a/lib/modules/manager/gradle/dep-types.ts
+++ b/lib/modules/manager/gradle/dep-types.ts
@@ -1,0 +1,21 @@
+import type { DepTypeMetadata } from '../types.ts';
+
+export const knownDepTypes = [
+  {
+    depType: 'plugin',
+    description: 'A Gradle plugin dependency',
+  },
+  {
+    depType: 'dependencies',
+    description: 'A standard Gradle dependency',
+  },
+  {
+    depType: 'devDependencies',
+    description: 'A dependency from a `buildSrc` project',
+  },
+  {
+    depType: 'test',
+    description:
+      'A test dependency from the `gradle-consistent-versions` plugin lock file',
+  },
+] as const satisfies readonly DepTypeMetadata[];
--- a/lib/modules/manager/npm/dep-types.ts
+++ b/lib/modules/manager/npm/dep-types.ts
@@ -1,0 +1,62 @@
+import type { DepTypeMetadata } from '../types.ts';
+
+export const knownDepTypes = [
+  {
+    depType: 'dependencies',
+    prettyDepType: 'dependency',
+    description: 'Listed under `dependencies`',
+  },
+  {
+    depType: 'devDependencies',
+    prettyDepType: 'devDependency',
+    description: 'Listed under `devDependencies`',
+  },
+  {
+    depType: 'optionalDependencies',
+    prettyDepType: 'optionalDependency',
+    description: 'Listed under `optionalDependencies`',
+  },
+  {
+    depType: 'peerDependencies',
+    prettyDepType: 'peerDependency',
+    description: 'Listed under `peerDependencies`',
+  },
+  {
+    depType: 'engines',
+    prettyDepType: 'engine',
+    description: 'Listed under `engines`',
+  },
+  {
+    depType: 'volta',
+    prettyDepType: 'volta',
+    description: 'Listed under `volta`',
+  },
+  {
+    depType: 'resolutions',
+    prettyDepType: 'resolutions',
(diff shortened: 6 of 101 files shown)


lib/workers/repository/updates/generate.ts
```diff
@@ -300,7 +301,10 @@ export function generateBranchConfig(
       upgrade.displayPending = '';
     }
     upgrade.prettyDepType =
-      upgrade.prettyDepType ?? upgrade.depType ?? 'dependency';
+      upgrade.prettyDepType ??
+      getPrettyDepType(upgrade.manager, upgrade.depType!) ??
```
jamietanna: Needs a test first

lib/modules/manager/types.ts
```diff
@@ -133,11 +133,12 @@ export interface LookupUpdate {
  */
 export interface PackageDependency<
   T = Record<string, any>,
+  DT extends string = string,
```
jamietanna: IMO, I'd prefer something like `DEP_TYPE` or `DEPTYPE` to make it clearer, but not sure if that's idiomatic?

lib/modules/manager/ansible-galaxy/dep-types.ts
```diff
@@ -0,0 +1,15 @@
...
+  {
+    depType: 'galaxy-collection',
+    description:
+      'Ansible Galaxy collection from `requirements.yml` or `galaxy.yml`',
+  },
+  {
+    depType: 'role',
+    description: 'Ansible role from `requirements.yml`',
+  },
+] as const satisfies DepTypeMetadata[];
```
viceice: why not adding the type to the variable? 🤔

is it for the type below which is probably used in a later PR?

lib/workers/repository/updates/generate.ts
viceice: should we do it in a separate PR? 🤔

lib/modules/manager/ansible-galaxy/dep-types.ts
jamietanna: You mean why not i.e.

```typescript
export const knownDepTypes: DepTypeMetadata[] = [
  {
    depType: 'galaxy-collection',
    description:
      'Ansible Galaxy collection from `requirements.yml` or `galaxy.yml`',
  },
  {
    depType: 'role',
    description: 'Ansible role from `requirements.yml`',
  },
] as const;
```

?

If so, no reason - hadn't spotted this in my review of the generated code - I'll correct 👍

lib/workers/repository/updates/generate.ts
jamietanna: This change to add the logic to derive this? Or the tests themselves?

lib/modules/manager/ansible-galaxy/dep-types.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const knownDepTypes: DepTypeMetadata[] = [
+  {
+    depType: 'galaxy-collection',
+    description:
+      'Ansible Galaxy collection from `requirements.yml` or `galaxy.yml`',
+  },
+  {
+    depType: 'role',
+    description: 'Ansible role from `requirements.yml`',
```
jamietanna: I'm not sure if we should be explicit here, especially if `managerFilePatterns` is used - thoughts?

lib/modules/manager/gradle/dep-types.ts
```diff
@@ -0,0 +1,24 @@
...
+    description: 'A Gradle plugin dependency',
+  },
+  {
+    depType: 'dependencies',
+    description: 'A standard Gradle dependency',
+  },
+  {
+    depType: 'devDependencies',
+    description:
+      'A dependency from a `buildSrc` project without Kotlin source files',
```
jamietanna: https://github.com/renovatebot/renovate/blob/e43164d75136a7fc7da80a86306274d8d93c10b3/lib/modules/manager/gradle/extract.ts#L278-L281 - not sure if we want to keep this info in

lib/modules/manager/npm/dep-types.ts
```diff
@@ -0,0 +1,61 @@
...
+    description: 'Listed under the top-level `pnpm` field',
+  },
+  {
+    depType: 'pnpm.overrides',
+    prettyDepType: 'overrides',
+    description: 'Listed under `pnpm.overrides`',
+  },
+];
+
+export type NpmExtractDepType = (typeof knownDepTypes)[number]['depType'];
```
jamietanna:
```suggestion
export type NpmDepType = (typeof knownDepTypes)[number]['depType'];
```

lib/modules/manager/ansible-galaxy/dep-types.ts
secustor: I would not name files here as it will be only confusing if it differs from `managerFilePatterns`

lib/modules/manager/bazel/dep-types.ts
```diff
@@ -0,0 +1,78 @@
+import type { DepTypeMetadata } from '../types.ts';
+
+export const knownDepTypes: DepTypeMetadata[] = [
+  {
+    depType: 'container_pull',
+    description: 'A Docker container image pulled via `container_pull`',
+  },
+  {
+    depType: '_container_pull',
```
secustor, line 9: Are these underline prefixed fields errors? 

If not, we should rename this to `private_*`

lib/modules/manager/gradle/dep-types.ts
RahulGautamSingh: Seems removable, after confirming with a test

RahulGautamSingh (commented):
Haven't gone through each manager changes but the feature change itself seems fine to me.

lib/modules/manager/gleam/readme.md
```diff
@@ -3,7 +3,7 @@ Renovate can update `gleam.toml` and/or `manifest.toml` files.
 The `gleam` manager can update these `depTypes`:
 
 - `dependencies`
-- `dev-dependencies`
+- `devDependencies` (taken from the `dev-dependencies` section)
```
secustor: The section should be deleted

lib/modules/manager/gradle/dep-types.ts
jamietanna: As in remove the logic? (Mind raising an Issue if so?)

lib/modules/manager/gleam/readme.md
jamietanna: I think I'd left it in as it was more focussed on the Gleam-specific types being removed, but agree it's confusing - removed

lib/workers/repository/updates/generate.ts
viceice: this logic change, it's runtime related change not docs only. so i prefer to move this logic change to a separate PR

lib/modules/manager/ansible-galaxy/dep-types.ts
```diff
@@ -0,0 +1,14 @@
...
+export const knownDepTypes: DepTypeMetadata[] = [
+  {
+    depType: 'galaxy-collection',
+    description: 'Ansible Galaxy collection',
+  },
+  {
+    depType: 'role',
+    description: 'Ansible role',
+  },
+] as const;
```
viceice:
```suggestion
];
```

const doesn't cahnge the result type

lib/modules/manager/ansible-galaxy/dep-types.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type AnsibleGalaxyDepType = (typeof knownDepTypes)[number]['depType'];
```
viceice: doesn't make sense, it's now always `string` from `DepTypeMetadata.depType`.

Do we really need those? I don't see a good usage.

lib/modules/manager/bazel/dep-types.ts
viceice: we can change later, this is just what the current implementation returns as `depType`

lib/modules/manager/gomod/dep-types.ts
```diff
@@ -0,0 +1,34 @@
...
+  },
+  {
+    depType: 'tool',
+    description: `A tool dependency using Go 1.24+'s \`tool\` directive`,
+  },
+];
+
+export type GomodDepType = (typeof knownDepTypes)[number]['depType'];
+
+export type GomodDep = PackageDependency<Record<string, any>, GomodDepType>;
```
viceice: doesn't make sense yet, `GomodDepType` is coerce to `string`. that why you used `as const satifies DepTypeMetadata` before? 🤔

lib/workers/repository/updates/generate.ts
jamietanna: https://github.com/renovatebot/renovate/pull/42193

lib/modules/manager/types.ts
```diff
@@ -192,7 +193,10 @@ export interface PackageDependency<
   isAbandoned?: boolean;
 }
 
-export interface Upgrade<T = Record<string, any>> extends PackageDependency<T> {
+export interface Upgrade<
+  T = Record<string, any>,
+  D extends string = string,
+> extends PackageDependency<T, D> {
```
viceice:
```suggestion
  DepType extends string = string,
> extends PackageDependency<T, DepType> {
```

use same as above

lib/modules/manager/types.ts
```diff
@@ -280,10 +284,32 @@ export interface GlobalManagerConfig {
...
+  prettyDepType?: string;
+  /** Human-readable description of what this depType represents */
+  description: string;
+}
+
 interface ManagerApiBase extends ModuleApi {
   defaultConfig: Record<string, unknown>;
 
   categories?: Category[];
+  knownDepTypes?: DepTypeMetadata[];
```
viceice: also add a tsdoc here too. we should do it for all "public" api interfaces in future.

lib/modules/manager/bazel/dep-types.ts
jamietanna: Moved into https://github.com/renovatebot/renovate/issues/42195 - agreed we should do something in the future about this

lib/modules/manager/gomod/dep-types.ts
jamietanna: Ah yes, that was why it was there!

jamietanna: Trying to enforce this does look like a much larger change 🤔

lib/modules/manager/pixi/dep-types.ts
```diff
@@ -0,0 +1,16 @@
...
+ * `feature-${name}` depTypes at runtime, where `name` is the
+ * feature name defined in the pixi configuration.
+ *
+ * Top-level dependencies (under `[dependencies]` or `[pypi-dependencies]`)
+ * do not have a depType set.
+ *
+ * There are no fixed/enumerable depType values for this manager.
+ */
+export const knownDepTypes: DepTypeMetadata[] = [];
+
```
jamietanna: Probably remove?

```suggestion
```

lib/modules/manager/gomod/dep-types.ts
jamietanna: 42ee883ab8 is what we'll need to do everywhere - do we want to?

viceice: I think we should skip that in this PR because it's totally unused yet.

viceice (commented):
see comment. I'm not sure about the benefit vs the code complexity 🤔

jamietanna:
@viceice I'm largely trying to make it "too hard not to" document the `depType`s

By making sure it's enforced through our type system, we shouldn't have any cases where we have contributions which (accidentally) don't document a new `depType`

That being said, I'm happy splitting this out into:

1. Document existing types + add to autogenerated docs
2. https://github.com/renovatebot/renovate/pull/42193
3. Adding strong types to enforce this more heavily

Sound good? Then we can focus on getting the start of the value, and work out how we want to enforce this more strongly (if at all)

viceice:
sounds good

tools/docs/manager.ts
```diff
@@ -146,6 +146,39 @@ export async function generateManagers(
         .join(', ');
       md += `This manager supports extracting the following datasources: ${escapedDatasources}.\n\n`;
       md += formatUrls(urls);
+      if (
+        definition.knownDepTypes?.length ||
+        definition.supportsDynamicDepTypesNote
+      ) {
+        md += '## Dependency types\n\n';
+        if (definition.knownDepTypes?.length) {
```
jamietanna: Should we display "we don't currently have any documented `depTypes`"?

jamietanna:
There haven't been any other changes in a1495c8682b88cc3615c2fcc3c3f621dbccdc933..origin/HEAD that have added / changed `depTypes`

viceice (approved):

PR #42155, opened 2026-03-26
docs: show all Managers in sidebar

jamietanna:
## Changes

As we do with the other modules, we should include Managers too.

Although there are many, it can be a little jarring not seeing the list
of Managers when viewing a specific one.

This requires we also show the "display name" (`Ansible`) in the
sidebar, and then use the more verbose `Automated Dependency Updates for
Ansible` as the header.

To do this, we can utilise the `sidebar_label` (which was previously
used in Docusarus, but unused by Material for MkDocs) via a custom hook
to inject the label for the sidebar.

Unfortunately this requires a custom hook, which will make moving away
from Material for MkDocs (to Zensical) a little less straightforward.

## Context

## AI assistance disclosure

The public repository: <URL>


jamietanna:
Back to draft while we improve the page title:

Screenshot 2026-03-26 at 11 42 21

viceice (commented):
we need to check the titles, we should use the short manager name

jamietanna:
Now working:

Screenshot 2026-03-26 at 13 47 46

tools/docs/manager.ts
```diff
@@ -86,12 +86,13 @@ export async function generateManagers(
 
     let md = codeBlock`
       ---
-      title: ${getTitle(manager, displayName, isCustomMgr)}
-      sidebar_label: ${displayName}
+      title: ${displayName}
```
viceice: doesn't this also change the `title` html meta attribute which is important for search engines?

jamietanna: yes the `<title>` is now `Ansible - Renovate Docs`

The page body still contains the full title

I'll see what options we have to preserve this

tools/mkdocs/mkdocs-hooks/sidebar-label.py
jamietanna, line 1: It's not ideal that we need to do this - it does seem to work though 🤔

jamietanna:
(Moving to draft so we can decide if it's the right approach)

tools/docs/manager.ts
viceice: the hook is probably the best we can do for now, so we retain title as is?

jamietanna:
Notice that the `<title>` is unchanged, but we now see the page title + sidebar label corrected - do we want that?

Screenshot 2026-04-02 at 14 58 23

RahulGautamSingh (commented):
Looks good

viceice (approved):
seems ok to me if the page meta stays same for search engines

PR #42156, opened 2026-03-26
feat(config): Add BLoC monorepo

Sander0542:
## Changes

Monorepo for Flutter/Dart BLoC package. The packages are not released under the same version, but they are dependent on eachother.

Repo: https://github.com/felangel/bloc

Releases example: https://github.com/felangel/bloc/releases

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -123,6 +123,7 @@
     "babel": "https://github.com/babel/babel",
     "backstage": "https://github.com/backstage/backstage",
     "baset": "https://github.com/igmat/baset",
+    "bloc": "https://github.com/felangel/bloc",
     "brave": "https://github.com/openzipkin/brave",
     "bugsnag-js": "https://github.com/bugsnag/bugsnag-js",
     "bull-board": "https://github.com/felixmosh/bull-board",


viceice (commented):
please extend PR description and share links.

are all packages have the same version?

Sander0542:
> please extend PR description and share links.
> 
> are all packages have the same version?

I have updated the PR description detailing the lack of a singular release, but providing reasoning for having this as a monorepo.

viceice:
> The packages are not released under the same version, but they are dependent on eachother.

Does that mean updates fail without grouping? Grouping different versions mean the PR is immortal and can no longer be ignored.

Sander0542:
I had a problem where the versions would not update separetly, but when I combined the commits of the PRs it would work.

Note: This is not always the case, but with the last major version upgrade it was causing issues.

viceice (changes requested):
then this won't be merged as monorepo because it causes immortal PRs and the packages don't always need to be grouped.

Sander0542:
What should be the way forward for this PR?

viceice:
> What should be the way forward for this PR?

I think there is no way forward yet. so close and forget for now. maybe we create an optional monorepo preset with cases like this in future.

PR #42162, opened 2026-03-26
feat(report): add `reportFormatting` option to format JSON reports with Prettier

felipecrs:
## Changes

This PR adds an option `reportFormatting`, which when enabled, formats the JSON report output with Prettier before writing it to disk or to S3.

This is useful in case the report is being consumed by humans, as it makes it easier to read, and, most importantly, to compare with previous reports.

## Context

## AI assistance disclosure

The public repository: https://github.com/felipecrs/renovate-tutorial


lib/instrumentation/reporting.spec.ts
```diff
@@ -136,6 +138,31 @@ describe('instrumentation/reporting', () => {
     );
   });
 
+  it('write formatted report if reportFormatting is enabled', async () => {
+    const formatMock = vi.fn().mockResolvedValue('{"formatted":true}');
+    vi.mocked(prettier).mockReturnValue({ format: formatMock } as never);
```
viceice: use partial helper instead of never

felipecrs: I refactored it to avoid mocking Prettier in this test, following the example of https://github.com/renovatebot/renovate/blob/e8a5c70862d71cebd2904b9cf493a6b66e3b3351/lib/workers/repository/config-migration/branch/migrated-data.spec.ts#L163.

viceice (dismissed):

jamietanna (commented):
IMO, we should maybe make this:

- `JSON.stringify`
- `prettier`
- Not pretty-printed

I'd be happy only using `prettier` instead of `JSON.stringify` - there's no "breaking change" with this as it's ""just"" formatting, _and_ this is experimental, so it's fine to change that

I'd also say that our default should really be not formatted - although with compression it can be the case it doesn't provide that much overhead, I've found that there can be meaningful impact between pretty-printing and not with large repositories

felipecrs:
Hey @jamietanna,

> IMO, we should maybe make this:
> 
> * `JSON.stringify`
> * `prettier`
> * Not pretty-printed
> 
> I'd be happy only using `prettier` instead of `JSON.stringify` - there's no "breaking change" with this as it's ""just"" formatting, _and_ this is experimental, so it's fine to change that

You mean like changing `reportFormatting: boolean` to `reportFormatting: 'stringify' | 'prettier' | 'disabled'`, right?

I can do that. But would someone really care about using `stringify` over `prettier`? I can't imagine a use case.

> I'd also say that our default should really be not formatted - although with compression it can be the case it doesn't provide that much overhead, I've found that there can be meaningful impact between pretty-printing and not with large repositories

I totally agree. That's why I made this option come disabled by default. :)

RahulGautamSingh (approved):
LGTM

jamietanna:
I don't know if folks would prefer one over the other - I can imagine that `prettier` is slower than `JSON.stringify`?

(We may actually want to use `safe-stable-stringify` for a non-`prettier` option) 

I'm happy this being `prettier` vs `off` (so a boolean would be OK)

This would have a default of `false`, so we'll now be no longer pretty-printing

jamietanna (changes requested):

felipecrs:
Hey @jamietanna, thanks for the review.

> I don't know if folks would prefer one over the other - I can imagine that `prettier` is slower than `JSON.stringify`?

It is definitely slower, but I don't think it matters for this use case. Here's a benchmark:

```console
❯ ls -lah report.json
-rw-r--r-- 1 felipecrs felipecrs 462K Mar 30 10:14 report.json

❯ pnpm exec vitest bench run reportFormatting.bench.ts

 ✓ reportFormatting.bench.ts > formatting 3117ms
     name                       hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · prettier               6.7622  135.98  162.56  147.88  155.58  162.56  162.56  162.56  ±4.17%       10
   · safe-stable-stringify  462.63  1.9512  4.1346  2.1616  2.0983  3.7632  3.7766  4.1346  ±2.34%      232

 BENCH  Summary

  safe-stable-stringify - reportFormatting.bench.ts > formatting
    68.41x faster than prettier
```

```ts
import { readFile } from 'node:fs/promises';
import { format } from 'prettier';
import { stringify } from 'safe-stable-stringify';
import { bench, describe } from 'vitest';

const text = await readFile('report.json', 'utf8');
const json = JSON.parse(text);

describe('formatting', () => {
  bench('prettier', async () => {
    await format(text, { parser: 'json' });
  });

  bench('safe-stable-stringify', () => {
    stringify(json, null, 2);
  });
});
```

> I'm happy this being `prettier` vs `off` (so a boolean would be OK)
>
> This would have a default of `false`, so we'll now be no longer pretty-printing

Sorry, I'm confused.

That is exactly what this PR does in its current state, but you _requested changes_. Can you please clarify what changes you are requesting?

BTW, I can add the `stringify` option if that's what you mean (using `safe-stable-stringify`), no problem at all.

lib/instrumentation/reporting.ts
```diff
@@ -77,6 +78,14 @@ export function finalizeReport(): void {
   }
 }
 
+async function getReportBody(config: RenovateConfig): Promise<string> {
+  const json = JSON.stringify(report);
+  if (!config.reportFormatting) {
+    return json;
+  }
+  return prettier().format(json, { parser: 'json' });
+}
```
jamietanna, line 87: Let's change this so the default is _not pretty printed whatsoever_ - then if someone opts-in, they can get it pretty-printed

jamietanna:
(Also, reminder to not @-mention reviewers/maintainers, we're following our notifications :))

lib/instrumentation/reporting.ts
felipecrs: That is what these lines are doing:

https://github.com/renovatebot/renovate/blob/b25af7542cdd8b489fc63a43bf4463538b8b0458/lib/instrumentation/reporting.ts#L82-L85

It avoids pretty-printing when `configFormatting: false` (the default), retaining the exact same behavior of before.

felipecrs:
> (Also, reminder to not @-mention reviewers/maintainers, we're following our notifications :))

ACK. Apologies for that.

lib/instrumentation/reporting.ts
jamietanna: Ah yes, I'd misread this as it pretty-printing it 👍

lib/config/options/index.ts
```diff
@@ -505,6 +505,15 @@ const options: Readonly<RenovateOptions>[] = [
...
     experimental: true,
   },
+  {
+    name: 'reportFormatting',
+    description:
+      'If set to `true`, Renovate will format the report JSON with Prettier before writing.',
+    globalOnly: true,
+    type: 'boolean',
+    default: false,
+    experimental: true,
```
jamietanna, line 515: CLI definitely not needed:

```suggestion
    experimental: true,
    cli: false,
```

felipecrs: Hm... even though `reportType` and `reportPath` have CLI flags?

```console
❯ pnpm start --help | grep -- --report
  --report-type <string>                                       Set how, or if, reports should be generated.
  --report-path <string>                                       Path to where the file should be written. In case of `s3` this has to be a full S3 URI.
  --report-formatting [boolean]                                If set to `true`, Renovate will format the report JSON with Prettier before writing.
```

jamietanna: Yes - we can add it in the future if we feel we need to, but I'm trying to keep the CLI arguments fairly lean, and especially as this is an experimental feature (and ""only"" handles output formatting) I don't feel we need to have a CLI option for it

felipecrs: Fair enough.

And I agree with that. The `--help` is too convoluted already.

jamietanna (approved):

PR #42164, opened 2026-03-26
fix(npm): use packageName for pnpm overrides with range selectors in minimumReleaseAgeExclude

oikarinen:
## Changes

When a `pnpm.overrides` entry uses a version range selector (e.g. `"fast-xml-parser@<=5.3.5": "5.5.7"`), Renovate was generating an invalid `minimumReleaseAgeExclude` entry like `fast-xml-parser@<=5.3.5@5.5.7` instead of `fast-xml-parser@5.5.7`, causing `pnpm install` to fail with `ERR_PNPM_INVALID_MINIMUM_RELEASE_AGE_EXCLUDE`.

The root cause: `depName` for pnpm overrides with range selectors contains the full override key (needed for correctly updating `package.json`), but `minimumReleaseAgeExclude` requires exact package name + version pairs.

Fix by using `upgrade.packageName ?? upgrade.depName` when constructing `minimumReleaseAgeExclude` entries. `packageName` holds the bare npm package name (`fast-xml-parser`) as parsed by `@pnpm/parse-overrides`, while `depName` retains the full key. Falls back to `depName` for all other dependency types where `packageName` is not set, preserving existing behaviour.

## Context

## AI assistance disclosure

Made with Cursor (claude-4.6-sonnet). AI was used for root cause analysis, implementation, tests, and documentation.

### DIFF
--- a/lib/modules/manager/npm/artifacts.spec.ts
+++ b/lib/modules/manager/npm/artifacts.spec.ts
@@ -703,6 +703,41 @@ minimumReleaseAgeExclude:
       expect(res).toBeNull();
     });
 
+    it('uses packageName (bare package name) for pnpm overrides with range selectors', async () => {
+      fs.getSiblingFileName.mockReturnValueOnce('pnpm-workspace.yaml');
+      fs.localPathExists.mockResolvedValueOnce(true);
+      fs.readLocalFile.mockResolvedValueOnce(
+        codeBlock`minimumReleaseAge: 4320`,
+      ); // for pnpm-workspace.yaml
+      const res = await updateArtifacts({
+        packageFileName: 'package.json',
+        updatedDeps: [
+          {
+            ...validDepUpdate,
+            depName: 'fast-xml-parser@<=5.3.5',
+            packageName: 'fast-xml-parser',
+            depType: 'pnpm.overrides',
+            currentValue: '5.3.5',
+            newVersion: '5.5.7',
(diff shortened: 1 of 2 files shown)


RahulGautamSingh (commented):
Looks good, But please sign the CLA first.

lib/modules/manager/npm/artifacts.spec.ts
```diff
@@ -703,6 +703,41 @@ minimumReleaseAgeExclude:
...
+        newPackageFileContent: 'some new content',
+        config,
+      });
+      expect(res).toStrictEqual([
+        {
+          file: {
+            type: 'addition',
+            path: 'pnpm-workspace.yaml',
+            contents:
+              'minimumReleaseAge: 4320\nminimumReleaseAgeExclude:\n  # Renovate security update: fast-xml-parser@5.5.7\n  - fast-xml-parser@5.5.7\n',
```
viceice, line 735: use `codeBlock` string template helper and properly format

viceice (commented):
otherwise LGTM

Squixx:
This seems stale? Anyone looking to get this merged? @viceice  What would be the policy on this? should i make a new pr to address this if the author is MIA?

viceice:
> This seems stale? Anyone looking to get this merged? @viceice  What would be the policy on this? should i make a new pr to address this if the author is MIA? 

feel free to open a new PR. you can retain the current commits, it's not a must

PR #42173, opened 2026-03-27
feat(manager): add Apache Ant manager with basic inline version extraction

RahulGautamSingh:
## Changes
  - Add new ⁠ ant ⁠ package manager that extracts Maven dependencies from Ant ⁠ build.xml ⁠ files
  - Support inline ⁠ version= ⁠ attributes on ⁠ <dependency> ⁠ elements inside ⁠ <artifact:dependencies> ⁠ blocks
  - Support ⁠ scope ⁠ attribute mapping to ⁠ depType ⁠ (defaults to ⁠ compile ⁠)
  - Add XML version updater with single/double quote support
  - Register manager in ⁠ api.ts ⁠ with ⁠ build.xml ⁠ file matching

  ## Context

  Please select one of the following:

  - [x] This closes an existing Issue, Closes: #42178
  - [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or
   implementation

  ## AI assistance disclosure

  Did you use AI tools to create any part of this pull request?

  Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

  - [ ] No — I did not use AI for this contribution.
  - [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
  - [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, and documentation).
  - [ ] Yes — other (please describe):

  ## Documentation (please check one with an [x])

  - [x] I have updated the documentation, or
  - [ ] No documentation update is required

  ## How I've tested my work (please select one)

  I have verified these changes via:

  - [ ] Code inspection only, or
  - [x] Newly added/modified unit tests, or
  - [ ] No unit tests, but ran on a real repository, or
  - [ ] Both unit tests + ran on a real repository

Public repository: https://github.com/RahulGautamSingh/ant-basic

PR #42175, opened 2026-03-27
feat(manager/ant): add property resolution and .properties file support

RahulGautamSingh:
## Changes
- Resolve same-file `<property name="..." value="..." />` references in version attributes
- Load external `.properties` files via `<property file="..." />`
- Implement first-definition-wins semantics matching Ant's property immutability
- Detect and skip circular and unresolvable property references
- Support chained property resolution (property values referencing other properties)
- Add `updateDependency` function for both XML `<property>` value attributes and `.properties` file values
- Track `sharedVariableName` and redirect `fileReplacePosition` to the property source file
- Group dependencies by their property source file, with `editFile` set for cross-file updates

### Why `updateDependency` instead of auto-replace

The default auto-replace function calls `extractPackageFile` (single-file extraction without property resolution) to confirm the update succeeded. For property-backed dependencies, this always fails because `extractPackageFile` sees the raw `${placeholder}` instead of the resolved value. 

## Context

- [x] This closes an existing Issue, Closes: #42179

> *Stacked PR 2/5* — depends on #42173

## AI assistance disclosure

- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, and documentation).

## Public Repos:

1. Re-ran on basic inline dependency versions: https://github.com/RahulGautamSingh/ant-basic/pull/2
2. Updating property version in same file: https://github.com/RahulGautamSingh/ant-inline-props/pulls
3. Updating property version in different  file: https://github.com/RahulGautamSingh/ant-external-props/pulls


lib/modules/manager/ant/extract.spec.ts
```diff
@@ -1,157 +1,710 @@
...
-          depName: 'junit:junit',
-          currentValue: '4.13.2',
-          depType: 'test',
-          registryUrls: [],
-        }),
-      ],
-    });
+  beforeEach(() => {
+    vi.resetAllMocks();
   });
```
viceice: not needed

https://github.com/renovatebot/renovate/blob/83b89553262c67be6b706675536135654b57c6c2/vitest.config.mts#L88

@zharinov can we add a lint rule to flag usage of that function?

lib/modules/manager/ant/extract.ts
viceice, line 1: can you split this file so it's not getting that huge? especially there is more to come.

lib/modules/manager/ant/update.ts
```diff
@@ -0,0 +1,44 @@
+import { logger } from '../../../logger/index.ts';
+import type { UpdateDependencyConfig } from '../types.ts';
+
+export function updateDependency({
```
viceice, line 5: i still don't get why we need this and auto replace doesn't work 🤔

RahulGautamSingh: Explained in PR description.

RahulGautamSingh: For the external `.properties`, `confirmIfDepUpdated` call within the auto-replace fn re-extracts and then checks:                                                      
     
```ts                                                                                                                               
  // auto-replace.ts:63                                                                                                                  
  if (upgrade.depName !== newUpgrade.depName) {                                                                                          
    return false; // depName mismatch                                                                                                    
  }   
```                                                                                                                                   
                                                                                                                                         
 A `.properties` file like `slf4j.version=1.7.36` has no information about `depName` (`org.slf4j:slf4j-api`), `groupId`, or `artifactId`. 

That metadata only exists in `build.xml`. So `extractPackageFile`("versions.properties") can't return a dep with a matching `depName` — it would  fail the `depName` check.  


ie. 
For external `.properties` files: `updateDependency` is necessary because `extractPackageFile` can't reconstruct dep metadata from a `.properties` file alone

RahulGautamSingh: I thought since we will also need to handle the `import` files in future PRs ,having a separate function to do the updates will be cleaner and modular.

We could do away with the `updateDepedency` and handle it via `replaceString` .. adding a bit of complexity to the extract function. Would you prefer that over `updateDependecy`?

lib/modules/manager/ant/extract.spec.ts
```diff
@@ -1,157 +1,706 @@
...
-  it('returns null for invalid XML', () => {
-    expect(extractPackageFile('<<< not xml >>>', 'build.xml')).toBeNull();
-  });
+  describe('property resolution', () => {
+    it('resolves inline property references', async () => {
+      fs.readLocalFile.mockResolvedValue(codeBlock`
+        <project>
+          <property name="slf4j.version" value="1.7.36"/>
+          <artifact:dependencies>
+            <dependency groupId="org.slf4j" artifactId="slf4j-api" version="${'${slf4j.version}'}" />
```
viceice:
```suggestion
            <dependency groupId="org.slf4j" artifactId="slf4j-api" version="\${slf4j.version}" />
```

only escaping the `$` seems more readable

lib/modules/manager/ant/extract.spec.ts
```diff
@@ -1,157 +1,706 @@
...
-    ).toBeNull();
-  });
+    it('resolves properties from external .properties files', async () => {
+      fs.readLocalFile.mockImplementation((file: string) => {
+        if (file === 'build.xml') {
+          return Promise.resolve(codeBlock`
+            <project>
+              <property file="versions.properties"/>
+              <artifact:dependencies>
+                <dependency groupId="org.slf4j" artifactId="slf4j-api" version="${'${slf4j.version}'}" />
```
viceice:
```suggestion
                <dependency groupId="org.slf4j" artifactId="slf4j-api" version="\${slf4j.version}" />
```

lib/modules/manager/ant/properties.ts
```diff
@@ -0,0 +1,224 @@
...
+
+/**
+ * Resolve chained property references within the property map itself.
+ * E.g., if prop A = "${B}" and prop B = "1.0", resolve A to "1.0".
+ * Marks circular properties by setting val to a placeholder that will be caught later.
+ */
+export function resolveChainedProps(props: Record<string, AntProp>): void {
+  const resolved = new Map<string, string | null>(); // null = circular
+
+  function resolve(key: string, chain: Set<string>): string | null {
```
viceice: can we please avoid the nested function?

lib/modules/manager/ant/properties.ts
```diff
@@ -0,0 +1,224 @@
+import type { XmlDocument, XmlElement } from 'xmldoc';
+import { regEx } from '../../../util/regex.ts';
+import { isXmlElement } from '../nuget/util.ts';
+import type { PackageDependency } from '../types.ts';
+import type { AntProp } from './types.ts';
+
+const placeholderRegex = regEx(/\$\{([^}]+)}/g);
```
viceice: it's not a good idea to cache regex with global flag, they can cause strange behaviors

lib/modules/manager/ant/extract.spec.ts
```diff
@@ -1,157 +1,706 @@
...
+              skipReason: 'recursive-placeholder',
+            }),
+          ],
+        },
+      ]);
     });
+  });
 
-    const result = await extractAllPackageFiles({}, ['build.xml', 'build.xml']);
+  describe('parsePropertiesFile', () => {
```
viceice: move to `properties.spec.ts`

lib/modules/manager/ant/update.ts
jamietanna, line 5: Might be worth adding some wording (from the PR) about why this over auto-replace

jamietanna (dismissed):

jamietanna (changes requested):
Can we please use JSDoc syntax?

lib/modules/manager/ant/update.ts
```diff
@@ -0,0 +1,46 @@
+import { logger } from '../../../logger/index.ts';
+import type { UpdateDependencyConfig } from '../types.ts';
+
+// For external .properties files: updateDependency is necessary
+// because extractPackageFile can't reconstruct dep metadata from a .properties file alone
```
jamietanna: (Might need `prettier` to fix)

```suggestion
/** For external .properties files: updateDependency is necessary because extractPackageFile can't reconstruct dep metadata from a .properties file alone */
```

jamietanna (dismissed):

zharinov (approved):
Clankers flag the problematic sharing of properties between multiple `build.xml` files and property precedence that doesn't match tool's resolution mechanics. But hey, as far as it's acceptable...

jamietanna:
@zharinov mind elaborating more? Trying to work out if Renovate should be more/less strict?

lib/modules/manager/ant/extract.ts
```diff
@@ -87,39 +116,102 @@ export function extractPackageFile(
...
-  const results: PackageFile[] = [];
   const visitedFiles = new Set<string>();
+  const allProps: Record<string, AntProp> = {};
+  const allRawDeps: RawDep[] = [];
 
   for (const packageFile of packageFiles) {
-    const result = await walkXmlFile(packageFile, visitedFiles);
-    if (result) {
-      results.push(result);
+    await walkXmlFile(packageFile, visitedFiles, allProps, allRawDeps);
```
zharinov, line 228: @jamietanna `allProps` and `allRawDeps` are shared between files, could be a problem

lib/modules/manager/ant/extract.ts
```diff
@@ -87,39 +116,102 @@ export function extractPackageFile(
...
+  } catch {
+    logger.debug(`ant manager: could not parse XML ${packageFile}`);
+    return;
+  }
+
+  // Collect property file references first (order matters for first-definition-wins)
+  const propertyFileRefs = collectPropertyFileRefs(doc);
+
+  // Collect inline properties (first-definition-wins: inline before file refs)
+  collectProperties(doc, content, packageFile, allProps);
```
zharinov: @jamietanna this ideally must be ordered as it appears in XML, not by property type

jamietanna:
Good checking, thank you (and our robot overlords)!

jamietanna:
Test failures + coverage, then should be good to go - would appreciate Sergei's re-review before we merge, too

lib/modules/manager/ant/properties.ts
```diff
@@ -0,0 +1,185 @@
...
+export function findAttrValuePosition(
+  content: string,
+  node: XmlElement,
+  attrName: string,
+): number {
+  const startTag = node.startTagPosition!;
+  const tagEnd = content.indexOf('>', startTag);
+  const tagContent = content.slice(startTag, tagEnd + 1);
+
+  const attrPattern = regEx(`${attrName}\\s*=\\s*(?:"([^"]*)"|'([^']*)')`);
```
viceice: do we need to escape the `attrName` ?

RahulGautamSingh: Escape but It only gets literal strings so not a real issue

viceice (approved):

jamietanna:
Failing due to GitHub issues - retrying build

PR #42201, opened 2026-03-27
chore: expose known environment variables

jamietanna:
## Changes

To provide more programmatic means to determine what environment
variables configure Renovate configuration options - and whether they're
Repo, Inherit or Global config options - we can expose these in two
forms:

- `getEnvOptionsMap()`
- a static JSON file, `renovate-env-options.ts`

This will provide programmatic use-cases from a plain JSON file that can
be used more readily - especially as it's distributed in the docs and
built package - as well as a function that can be used by users of
Renovate-as-a-library.

## Context

## AI assistance disclosure

The public repository: <URL>


lib/config/options/env-options.ts
```diff
@@ -0,0 +1,31 @@
...
+  for (const option of getOptions()) {
+    const envName = getEnvName(option);
+    if (envName) {
+      entries.push([
+        envName,
+        {
+          configName: option.name,
+          globalOnly: option.globalOnly ?? false,
+          inheritConfigSupport: option.inheritConfigSupport ?? false,
+          type: option.type,
```
jamietanna, line 24: maybe not needed?

lib/config/options/env-options.ts
```diff
@@ -0,0 +1,31 @@
...
+  const entries: [string, EnvOptionInfo][] = [];
+  for (const option of getOptions()) {
+    const envName = getEnvName(option);
+    if (envName) {
+      entries.push([
+        envName,
+        {
+          configName: option.name,
+          globalOnly: option.globalOnly ?? false,
+          inheritConfigSupport: option.inheritConfigSupport ?? false,
```
jamietanna, line 23: Possibly not needed - if you're able to set environment variables, it doesn't matter?

lib/workers/global/config/parse/env.ts
```diff
@@ -1,13 +1,15 @@
...
+import { getEnvName } from '../../../../config/options/env.ts';
 import { getOptions } from '../../../../config/options/index.ts';
 import type { AllConfig } from '../../../../config/types.ts';
 import { logger } from '../../../../logger/index.ts';
 import { parseJson } from '../../../../util/common.ts';
 import { coersions } from './coersions.ts';
-import type { ParseConfigOptions } from './types.ts';
 import { migrateAndValidateConfig } from './util.ts';
 
+export { getEnvName };
```
jamietanna: do we want to do this? Or update all call sites?

tools/docs/env-options.ts
```diff
@@ -0,0 +1,13 @@
+import { getEnvOptionsMap } from '../../lib/config/options/env-options.ts';
+import { prettier } from '../../lib/expose.ts';
+import { updateFile } from '../utils/index.ts';
+
+export async function generateEnvOptions(dist: string): Promise<void> {
```
jamietanna, line 5: I'd say we should ideally have a JSON schema for this file

lib/config/options/env-options.ts
```diff
@@ -0,0 +1,31 @@
...
+  globalOnly: boolean;
+  inheritConfigSupport: boolean;
+  type: string;
+}
+
+export type EnvOptionsMap = Record<string, EnvOptionInfo>;
+
+export function getEnvOptionsMap(): EnvOptionsMap {
+  const entries: [string, EnvOptionInfo][] = [];
+  for (const option of getOptions()) {
```
secustor, line 15: Are you also want options which are not part of config options? 

E.g. OpenTelemetry, LOG_LEVEL or else when we are reading env directly

lib/workers/global/config/parse/env.ts
secustor: No

lib/config/options/env-options.ts
jamietanna: At this point, I'm happy us only exposing `RENOVATE_` options - we can add additional environment variables in the future

lib/config/options/env-options.ts
RahulGautamSingh: Good to have since the JSON file will be used as single source of truth by those who will use it

tools/docs/env-options.ts
```diff
@@ -0,0 +1,13 @@
...
+  const map = getEnvOptionsMap();
+
+  const config = await prettier().format(JSON.stringify(map), {
+    filepath: 'renovate-env-options.json',
+  });
+
+  await updateFile(`${dist}/renovate-env-options.json`, `${config}\n`);
```
viceice, line 12: why in dist folder 🤔

viceice (commented):
I don't see the use case but won't block this

tools/docs/env-options.ts
jamietanna: Similar to the JSON Schema, we want to make sure this is distributed in the built package / docs site

viceice (approved):

lib/config/options/env-options.ts
jamietanna: I'll leave it for now

tools/docs/env-options.ts
jamietanna: We can add this in a follow-up

PR #42025, opened 2026-03-20
feat(manager): add rebar3 manager for Erlang dependency extraction

Taure:
## Summary

Adds a new `rebar3` manager that extracts dependencies from Erlang/OTP `rebar.config` files and reads locked versions from `rebar.lock` files.

- Supports all rebar3 dependency formats: hex packages (with version constraints, bare atoms, pkg overrides), git dependencies (tag/branch/ref), git subdirectory dependencies, and profile-scoped dependencies (test)
- Reuses existing `hex` datasource, `hex` versioning, `git-tags` and `github-tags` datasources
- Adds `erlang` to the list of manager categories

This is part 1 of 3 for the rebar3 manager. Lock file updating (artifacts) and range strategy support will follow in subsequent PRs.

### Dependency formats supported

| Format | Example |
|--------|---------|
| Hex (version) | `{cowboy, "~> 2.9"}` |
| Hex (pinned) | `{cowboy, "2.9.0"}` |
| Hex (bare atom) | `cowboy` |
| Hex (pkg override) | `{app, {pkg, alt_name}}`, `{app, "1.0", {pkg, alt_name}}` |
| Git (tag) | `{app, {git, "url", {tag, "1.0"}}}` |
| Git (branch) | `{app, {git, "url", {branch, "main"}}}` |
| Git (ref) | `{app, {git, "url", {ref, "abc123"}}}` |
| Git subdirectory | `{app, {git_subdir, "url", {tag, "1.0"}, "path"}}` |
| Profile deps | `{profiles, [{test, [{deps, [...]}]}]}` |

### Files

| File | Purpose |
|------|---------|
| `lib/modules/manager/rebar3/index.ts` | Manager config and exports |
| `lib/modules/manager/rebar3/extract.ts` | Parse `rebar.config` + read `rebar.lock` |
| `lib/modules/manager/rebar3/readme.md` | Manager documentation |
| `lib/constants/category.ts` | Added `erlang` category |
| `lib/modules/manager/api.ts` | Registered `rebar3` manager |

## AI disclosure

This PR was written primarily by Claude Code.

## Test plan

- [x] `extract.spec.ts` — 6 tests: invalid files, no deps, all dep formats, locked versions, simple config, commented deps
- [x] Verified existing manager index tests still pass with new manager registered

PR #42223, opened 2026-03-29
feat(manager): add rebar manager for Erlang dependency extraction

Taure:
## Summary

Adds a new `rebar` manager that extracts dependencies from Erlang/OTP `rebar.config` files and reads locked versions from `rebar.lock` files.

- Supports hex packages (versioned, bare atom, pkg overrides), git dependencies (tag/branch/ref), git subdirectory dependencies, and profile-scoped dependencies
- Reuses existing `hex` datasource, `hex` versioning, `git-tags` and `github-tags` datasources
- Adds `erlang` to the list of manager categories

This is part 1 of 3 for the rebar manager. Lock file updating (artifacts) and range strategy support follow in stacked PRs.

### Changes from previous PR (#42025)
- Renamed `rebar3` → `rebar` (no version in manager name)
- Inlined fixtures using `codeBlock` template helper
- Mocked filesystem in tests instead of using real fixture files

## AI disclosure

## Test plan

- [x] `extract.spec.ts` — 18 tests covering all dep formats, lockfile parsing, edge cases


lib/modules/manager/rebar/readme.md
```diff
@@ -0,0 +1,30 @@
...
+- Hex packages: `{cowboy, "~> 2.9"}`, `{cowboy, "2.9.0"}`
+- Hex packages with alternate names: `{app, {pkg, hex_name}}`, `{app, "1.0", {pkg, hex_name}}`
+- Git dependencies: `{app, {git, "url", {tag, "1.0"}}}`, `{app, {git, "url", {branch, "main"}}}`, `{app, {git, "url", {ref, "abc123"}}}`
+- Git subdirectory dependencies: `{app, {git_subdir, "url", {tag, "1.0"}, "path"}}`
+
+Dependencies from `profiles` (e.g., test profile) are also extracted with appropriate `depType`.
+
+The `rebar3` tool is used to keep the `rebar.lock` file up-to-date.
+
+### `lockFileMaintenance`
```
viceice: remove below, not yet implemented

lib/modules/manager/rebar/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import * as hexVersioning from '../../versioning/hex/index.ts';
+
+export { extractPackageFile } from './extract.ts';
+
+export const url = 'https://rebar3.org';
+export const categories: Category[] = ['erlang'];
+
+export const defaultConfig = {
+  managerFilePatterns: ['/(^|/)rebar\\.config$/'],
+  versioning: hexVersioning.id,
```
viceice: should be set by datasource, not manager 
```suggestion
```

viceice (commented):
fix type issues

Taure:
Sorry about this. :( I tried to stack them but I couldn't stack them in a good way where you could only see the changes. Maybe would have been better to have all three in one PR?

viceice:
> Sorry about this. :( I tried to stack them but I couldn't stack them in a good way where you could only see the changes. Maybe would have been better to have all three in one PR?

stacking means the second PR needs to target the source branch of the first PR.
if you're not able then just leave them as is for now to avoid CI time waste. we can concentrate on extract PR first.

lib/modules/manager/rebar/extract.spec.ts
```diff
@@ -0,0 +1,362 @@
...
+      expect(res?.deps).toHaveLength(2);
+    });
+
+    it('handles nested brackets in deps opening line', async () => {
+      const config = codeBlock`
+        {deps, [[{cowboy, "2.13.0"}]
+        ]}.
+      `;
+      const res = await extractPackageFile(config, 'rebar.config');
+      expect(res).toBeDefined();
```
RahulGautamSingh: Check what deps were actually extracted

lib/modules/manager/rebar/extract.spec.ts
RahulGautamSingh, line 1: Add a test for duplicate dep names (same dep in prod and test) to verify the map-based dedup behavior

RahulGautamSingh (changes requested):

oleg-sigida-happening:
Hi all! Could we please keep working on this PR?

lib/modules/manager/rebar/extract.spec.ts
Frost: That's done here, right? https://github.com/renovatebot/renovate/pull/42223/changes#diff-b56c13c0c75a7974ce340da4c53ac280341dff7f1a8508b9f6f3e2e17eec8505R368-R387

RahulGautamSingh (commented):
@Taure are you working on the PR?

Taure:
Yes will pick this up again. Thank you for reaching out.

lib/modules/manager/rebar/extract.ts
```diff
@@ -0,0 +1,359 @@
...
+      dep.packageName = `${github.owner}/${github.repo}`;
+    } else {
+      dep.datasource = GitTagsDatasource.id;
+      dep.packageName = url;
+    }
+
+    if (refType === 'tag') {
+      dep.currentValue = refValue;
+    } else if (refType === 'branch') {
+      dep.currentValue = refValue;
```
RahulGautamSingh:
```suggestion
    if (refType === 'tag' || refType === 'branch) {
      dep.currentValue = refValue;
```

lib/modules/manager/rebar/extract.ts
RahulGautamSingh, line 1: The style guide requires named capturing groups when capturing multiple groups. `githubUrlRegExp` already does it right `(?<owner>, ?<repo>)`; the rest should follow.

RahulGautamSingh (changes requested):

Taure:
Sorry for the long gap - picking this back up.

Both review points are done: every regex now uses named capturing groups (the style guide asks for them even on single captures, so `bareAtomRegExp` got one too), and the duplicated tag/branch arms are collapsed.

Rebasing onto current main surfaced `no-stateful-global-regex`, so the two lock-file regexes moved from `exec()` loops to `matchAll()`.

While re-reading the extractor I found a few real bugs and fixed them here rather than leave them for a follow-up:

- `currentDigest` was being taken from `rebar.lock`. Since the digest never appears in `rebar.config`, `doAutoReplace` can't satisfy `digestReplaced` and every git tag/branch update failed. It now mirrors `mix`: `lockedVersion` from the lock, digest only from an in-config `{ref, "..."}`.
- Profile detection was a positional scan for the last `{test,` line. A stock `{alias, [{test, [eunit, ct]}]}` mislabelled every prod dep, non-test profiles after the test profile all became `test`, and CRLF shifted the offsets. It now tracks the enclosing term stack, and uses the profile name as `depType`.
- Comment stripping wasn't string-aware, so a percent-encoded git URL truncated the line and unbalanced bracket depth for the rest of the file.
- The deps list was located via the first `[` on the line, which dropped every dep of a single-line profile.

Also added `dep-types.ts` - `knownDepTypes` landed in #42142 while this branch was idle, and without it the generated docs page claims the manager has no `depType` values.

`extract.ts` is at 100% branch coverage. Lock file updating still lands in #42224, which is why extract returns `lockFiles` with no `updateArtifacts` yet.

RahulGautamSingh (commented):
LGTM

viceice (approved):
ok. lockfile update come later?

PR #42026, opened 2026-03-20
feat(manager): add rebar3 lock file and artifact support

Taure:
## Summary

Adds `updateArtifacts` to the rebar3 manager (from #42025) to support lock file updates via `rebar3 upgrade`.

- Runs `rebar3 upgrade <dep>` for individual dependency updates
- Runs `rebar3 upgrade` for lockFileMaintenance
- Includes command injection prevention (validates dependency names)
- Handles error cases: missing lock file, exec failures, no changes

This is part 2 of 3 for the rebar3 manager. Depends on #42025.

### Files

| File | Purpose |
|------|---------|
| `lib/modules/manager/rebar3/artifacts.ts` | Run `rebar3 upgrade` to update lock file |
| `lib/modules/manager/rebar3/artifacts.spec.ts` | 8 test cases |
| `lib/modules/manager/rebar3/index.ts` | Added `updateArtifacts` and `supportsLockFileMaintenance` exports |

## AI disclosure

This PR was written primarily by Claude Code.

## Test plan

- [x] `artifacts.spec.ts` — 8 tests: skip conditions, lock updates, lockFileMaintenance, error handling, injection prevention

PR #42224, opened 2026-03-29
feat(manager): add rebar lock file and artifact support

Taure:
## Summary

Adds `updateArtifacts` to the rebar manager to support lock file updates via `rebar3 upgrade`.

- Runs `rebar3 upgrade <dep>` for individual dependency updates
- Runs `rebar3 upgrade --all` for lockFileMaintenance
- Uses `findLocalSiblingOrParent` for lock file lookup (supports monorepos)
- Includes command injection prevention via `shlex.quote`

This is part 2 of 3 for the rebar manager. Stacked on #42223 — merge that first to see only the incremental diff here.

### Changes from previous PR (#42026)
- Simplified lock file lookup to use `findLocalSiblingOrParent` directly
- Moved early return for missing lock file before write operations (perf)
- Removed hardcoded `^26` erlang constraint fallback
- Removed `/* v8 ignore */` — all paths now covered by tests

## AI disclosure

## Test plan

- [x] `artifacts.spec.ts` — 11 tests: skip conditions, lock updates, lockFileMaintenance, error handling, injection prevention


viceice (changes requested):
same as other PR. 

why new PR again?

PR #42225, opened 2026-03-29
feat(manager): add rebar range strategy support

Taure:
## Summary

Adds `getRangeStrategy` to the rebar manager for range-aware dependency updates.

- Uses `update-lockfile` for simple ranges with `auto` strategy
- Falls back to `widen` for complex ranges (e.g., `>= 1.0.0 and < 2.0.0`)

This is part 3 of 3 for the rebar manager. Stacked on #42224 → #42223 — merge those first to see only the incremental diff here.

## AI disclosure

## Test plan

- [x] `range.spec.ts` — 5 tests: bump/auto/pin strategies with simple and complex ranges


viceice (changes requested):
target the dependent PR to only show relevant changes

PR #42284, opened 2026-03-31
feat(instrumentation): allow exporting OTLP JSON file

jamietanna:
## Changes

We've recently been investing in improvements to Renovate's
OpenTelemetry instrumentation to improve debuggability and
supportability.

However, not everyone has an immediately available OpenTelemetry
Collector, so the added instrumentation isn't useful to those users.

Instead, we can provide a way to export to the OTLP file format, which
provides a means to export the data to disk and then consume it later.

This will also support use-cases like running in a CI platform like
GitHub Actions and GitLab CI, where OpenTelemetry Collectors may not be
exposed in that environment, but a post-run execution can export it.

We now need to make sure that `RENOVATE_USE_CLOUD_METADATA_SERVICES` is
never set, as `isTracingEnabled` will now resolve true, and the Cloud
Metadata Resource Detectors will be used.

## Context

## AI assistance disclosure

The public repository: <URL>

---

Once merged, I'll get Issues raised in the GitHub Action + GitLab CI usages - should we also do the same i.e. in Forgejo?

### DIFF
--- a/docs/usage/opentelemetry.md
+++ b/docs/usage/opentelemetry.md
@@ -63,6 +63,22 @@ The default values is `all`, which means all of the above detectors are used. Se
 - `github`
 - `env`
 
+### File exporter
+
+The OpenTelemetry data can also be exported into a newline-delimited JSON (NDJSON) file, in OTLP JSON format.
+
+This is configured using the `RENOVATE_TRACING_FILE_EXPORTER_PATH` environment variable.
+
+You can then ingest this data into other tools, for instance:
+
+```sh title="Exporting the JSONL OTLP file to a local OpenTelemetry traces API"
+while IFS= read -r line; do
+  curl -s -X POST http://localhost:4318/v1/traces \
+    -H 'Content-Type: application/json' \
+    -d "$line"
+done < /tmp/traces.jsonl
+```
+
 ## Supported OTLP data
 
 ### Traces
--- a/lib/instrumentation/file-exporter.spec.ts
+++ b/lib/instrumentation/file-exporter.spec.ts
@@ -1,0 +1,181 @@
+import type { Resource } from '@opentelemetry/resources';
+import type { ReadableSpan } from '@opentelemetry/sdk-trace-base';
+import type { DirectoryResult } from 'tmp-promise';
+import { dir } from 'tmp-promise';
+import upath from 'upath';
+import { partial } from '~test/util.ts';
+import { FileSpanExporter } from './file-exporter.ts';
+
+vi.mock('node:fs/promises', () => ({
+  appendFile: vi.fn().mockResolvedValue(undefined),
+}));
+
+const { appendFile } = vi.mocked(await import('node:fs/promises'));
+
+function createMockSpan(overrides: Partial<ReadableSpan> = {}): ReadableSpan {
+  return partial<ReadableSpan>({
+    name: 'test-span',
+    kind: 0,
+    spanContext: () => ({
+      traceId: '00000000000000000000000000000001',
+      spanId: '0000000000000001',
+      traceFlags: 1,
+    }),
+    startTime: [0, 0],
+    endTime: [1, 0],
+    status: { code: 0 },
+    attributes: {},
+    links: [],
+    events: [],
+    duration: [1, 0],
+    ended: true,
+    resource: partial<Resource>({
+      attributes: {},
+      merge: vi.fn(),
+    }),
+    instrumentationScope: { name: 'test' },
+    droppedAttributesCount: 0,
+    droppedEventsCount: 0,
+    droppedLinksCount: 0,
+    parentSpanContext: undefined,
+    ...overrides,
+  });
+}
+
+// ExportResultCode values from @opentelemetry/core
+const ExportResultCode = { SUCCESS: 0, FAILED: 1 };
+
+describe('instrumentation/file-exporter', () => {
+  let tmpDir: DirectoryResult;
+  beforeEach(async () => {
+    tmpDir = await dir({ unsafeCleanup: true });
+  });
+
+  afterAll(async () => {
+    await tmpDir?.cleanup();
+  });
+
+  it('writes OTLP JSON line to file', async () => {
+    const tracesPath = upath.join(tmpDir.path, 'test-traces.jsonl');
+
+    const exporter = new FileSpanExporter(tracesPath);
+    const spans = [createMockSpan()];
+
+    const result = await new Promise<{ code: number }>((resolve) => {
+      exporter.export(spans, resolve);
+    });
+
+    expect(result.code).toBe(ExportResultCode.SUCCESS);
+    expect(appendFile).toHaveBeenCalledTimes(1);
+    expect(appendFile).toHaveBeenCalledWith(
+      tracesPath,
+      expect.stringMatching(/^\{.*"resourceSpans".*\}\n$/),
+      'utf-8',
+    );
+
+    // Verify the written content is valid JSON
+    const writtenContent = appendFile.mock.calls[0][1] as string;
+    const parsed = JSON.parse(writtenContent.trim());
+    expect(parsed).toHaveProperty('resourceSpans');
+  });
+
+  it('reports FAILED on write error', async () => {
+    appendFile.mockRejectedValueOnce(new Error('disk full'));
+    const exporter = new FileSpanExporter(
+      upath.join(tmpDir.path, 'test-traces.jsonl'),
+    );
+
+    const result = await new Promise<{ code: number }>((resolve) => {
+      exporter.export([createMockSpan()], resolve);
+    });
+
+    expect(result.code).toBe(ExportResultCode.FAILED);
+  });
+
+  it('reports FAILED after shutdown', async () => {
+    const exporter = new FileSpanExporter(
+      upath.join(tmpDir.path, 'test-traces.jsonl'),
+    );
+    await exporter.shutdown();
+
+    const result = await new Promise<{ code: number }>((resolve) => {
+      exporter.export([createMockSpan()], resolve);
+    });
+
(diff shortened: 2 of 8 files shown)


lib/instrumentation/file-exporter.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+// ExportResultCode values from @opentelemetry/core
+const ExportResultCode = { SUCCESS: 0, FAILED: 1 };
+
+describe('instrumentation/file-exporter', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('writes OTLP JSON line to file', async () => {
+    const exporter = new FileSpanExporter('/tmp/test-traces.jsonl');
```
jamietanna: Should be Renovate's temp directory / one for tests

docs/usage/opentelemetry.md
```diff
@@ -65,6 +65,22 @@ The default values is `all`, which means all of the above detectors are used. Se
 - `github`
 - `env`
 
+### File exporter
+
+The OpenTelemetry data can also be exported into a JSON Lines (JSONL) file, in OTLP JSON format.
```
viceice: It seems we should update all references to `ndjson` to `jsonnl` as that's the recommend name since 2025?
Or at least update the docs for log format that it's `jsonnl` and not plain `json` ?

- <https://en.wikipedia.org/wiki/JSON_streaming#Newline-delimited_JSON>

jamietanna: Let's cover that in #42348 - for this one, let's keep it consistent

jamietanna:
```suggestion
The OpenTelemetry data can also be exported into a newline-delimited JSON (NDJSON) file, in OTLP JSON format.
```

lib/instrumentation/file-exporter.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+      merge: vi.fn(),
+    },
+    instrumentationLibrary: { name: 'test' },
+    instrumentationScope: { name: 'test' },
+    droppedAttributesCount: 0,
+    droppedEventsCount: 0,
+    droppedLinksCount: 0,
+    parentSpanId: undefined,
+    ...overrides,
+  } as unknown as ReadableSpan;
```
viceice: can we `partial<ReadableSpan>(...)` ?

lib/instrumentation/file-exporter.spec.ts
viceice: `tmp-promise` if there's something written to disk

docs/usage/opentelemetry.md
```diff
@@ -65,6 +65,22 @@ The default values is `all`, which means all of the above detectors are used. Se
 - `github`
 - `env`
 
+### File exporter
```
secustor, line 66: Should we eventually also write logs in this file 🤔 
In the end these are just another signal. I for example would have expected this.

jamietanna: I think - at the point we support OTLP Logs - then that'd make sense

lib/instrumentation/file-exporter.spec.ts
jamietanna: Sorry yep, adding now

secustor:
OTEL seems to define the stdout there as a default. https://opentelemetry.io/docs/specs/otel/protocol/file-exporter/
So in that case we route all logs and traces to stdout as OTEL data

jamietanna:
Claude Sonnet 5

````markdown
Code review complete — 6 findings survived verification. Ranked by severity:

**Correctness (2):**

1. **`lib/instrumentation/file-exporter.ts:42,45` — `shutdown()`/`forceFlush()` don't wait for in-flight writes.** `export()` kicks off `void this._writeSpans(spans).then(...)` without awaiting it. `shutdown()` just sets `_stopped = true` and resolves immediately; `forceFlush()` is a no-op. Renovate's shutdown path (`lib/renovate.ts`) calls `await otel.shutdown()` right after the final `'run'` span ends, and under `RENOVATE_X_HARD_EXIT` calls `process.exit()` immediately after that resolves — so the last (and most important) span's `appendFile()` write can be silently dropped or truncated. Fix: track the in-flight write promise and await it in both `shutdown()` and `forceFlush()`.

2. **`lib/instrumentation/utils.ts:8` — `isTracingEnabled()` now also fires on `RENOVATE_TRACING_FILE_EXPORTER_PATH` alone**, which pulls in `detectResources()`'s full cloud-metadata detector set (AWS/GCP/Azure). Someone who only wants a local trace file for debugging now unknowingly triggers real network probes to cloud metadata endpoints on every run, adding latency/timeouts especially in sandboxed or air-gapped environments. Worth either scoping detector selection independently of file-export-only mode, or documenting that file export implies the same cloud-detection behavior as other tracing modes.

**Cleanup (4):**

3. **`lib/instrumentation/file-exporter.ts:38`** — `if (serialized === undefined) return;` is dead code: the concrete `JsonTraceSerializer.serializeRequest` implementation never returns `undefined`, and it's untested. Given the repo requires 100% coverage, either drop the branch or mark it `/* v8 ignore */` with a reason per `docs/development/best-practices.md`.
4. **`lib/instrumentation/file-exporter.ts:11`** — `_filePath`, `_stopped`, `_writeSpans` use underscore-prefixed private members, which the repo's style guide explicitly says to avoid (`private` already provides encapsulation).
5. **`package.json:165`** — `@opentelemetry/otlp-transformer` breaks alphabetical ordering of the `@opentelemetry/*` block; belongs after `instrumentation-redis`, not right after `exporter-trace-otlp-http`.
6. ~~`file-exporter.spec.ts:46` redundant `vi.clearAllMocks()`~~ — already fixed; the workflow reviewed a snapshot taken before I re-removed it in the previous turn.

Most impactful to address: #1 (silent data loss on shutdown) and #2 (unexpected network calls). Want me to apply fixes for these?
````

**Comment by jamietanna** [comment]
> OTEL seems to define the stdout there as a default. https://opentelemetry.io/docs/specs/otel/protocol/file-exporter/ So in that case we route all logs and traces to stdout as OTEL data

Nice! However, this is more for use-cases where we want regular Renovate stdout to be there, though 🤔

**Comment by secustor** [comment]
Seems like they haven't defined a env variable for it yet as it is in `Development`

 https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#in-development-exporter-selection 
suggest maybe something like `otlp/file` 🤔 
and `OTEL_EXPORTER_OUTPUT_FILE=<path>`

File: lib/instrumentation/file-exporter.spec.ts [hunk 5]
```diff
@@ -0,0 +1,181 @@
...
+    });
+
+    const shutdownPromise = exporter.shutdown().then(() => {
+      events.push('shutdown-resolved');
+    });
+
+    // let any already-queued microtasks run; the write is still pending, so
+    // neither the export callback nor shutdown() should have resolved yet
+    await Promise.resolve();
+    await Promise.resolve();
```
viceice, line 141: I think there's a vitest function to run all tasks from microqueue , which is more reliable than this

lib/instrumentation/file-exporter.ts
```diff
@@ -0,0 +1,58 @@
+import { appendFile } from 'node:fs/promises';
+import { diag } from '@opentelemetry/api';
+import { JsonTraceSerializer } from '@opentelemetry/otlp-transformer';
+import type { ReadableSpan, SpanExporter } from '@opentelemetry/sdk-trace-base';
+
+// ExportResultCode values from @opentelemetry/core (not a direct dependency)
+const SUCCESS = 0;
+const FAILED = 1;
```
viceice:
```suggestion
const SUCCESS = 0 as const;
const FAILED = 1 as const;
```
so the type isn't number?

viceice (commented):
otherwise LGTM

lib/instrumentation/file-exporter.spec.ts
jamietanna: According to Claude Sonnet 5:

> Both suspicions confirmed empirically:
> - `vi.runAllTicks()` requires `vi.useFakeTimers()` and even then only flushes `process.nextTick`-based microtasks — it does not flush plain `.then()` chains, so it wouldn't actually fix what viceice is worried about.

https://vitest.dev/api/vi.html#vi-runallticks

(That being said, I don't know enough about the details - does that make sense?

jamietanna:
@secustor that's a fair point - I'm not sure about speculating on what OTel may use names wise

Yes, this is an experimental feature, but there may be some things being built on top of it (I have some plans 🤓) and I'd want to avoid an OTel library bump leading to this functionality breaking if `OTEL_TRACES_EXPORTER=otlp/file` and/or `OTEL_EXPORTER_OUTPUT_FILE` have a different meaning that then breaks/changes behaviour in newer OTel versions 🤔

lib/instrumentation/index.ts
```diff
@@ -74,6 +77,12 @@ export function init(): void {
     spanProcessors.push(new BatchSpanProcessor(exporter));
   }
 
+  if (isFileExporterEnabled()) {
+    spanProcessors.push(
+      new SimpleSpanProcessor(new FileSpanExporter(getFileExporterPath())),
```
secustor: You should add a batch processor, else the system will write single span on disk on its own

jamietanna: In practice I've found (on a small repo) it does many, but fair shout

secustor (approved):

lib/instrumentation/file-exporter.spec.ts
viceice: ok, what about `waitFor` or `waitUntil` ?

https://vitest.dev/api/vi.html#vi-waituntil

PR #42286, opened 2026-03-31
feat: S3-backed persistence for `persistRepoData`

ZackKanter:
## Changes

Adds S3-backed persistence for `persistRepoData` so Renovate can reuse `.git/` data across runs on ephemeral workers.

Example config:

```json
{
  "persistRepoData": true,
  "persistRepoDataType": "s3://my-bucket/renovate-git/",
  "s3Endpoint": "https://s3.us-east-1.amazonaws.com"
}
```

### Why S3 vs blobless cloning

The S3 implementation proposed here should be an additional option rather than a replacement for blobless cloning because it solves a different problem. 

Blobless clone makes a cold clone cheaper by skipping most blobs up front, but each run still starts from the remote Git host and has to renegotiate objects again. S3-backed repo persistence is meant for environments where runners are stateless or short-lived, such as containers or Fargate tasks, and therefore cannot keep a local .git directory between runs. In those cases, storing the repository metadata in S3 allows a later run to restore the previous Git state and perform an incremental fetch instead of paying the cost of a fresh clone every time. 

That can reduce repeated network transfer and repeated load on the source platform for repositories that Renovate processes often. At the same time, S3 adds operational complexity and storage overhead, so it is not universally better than blobless cloning. Blobless remains the simpler choice for one-off or infrequent runs, while S3 persistence is useful as an opt-in remote cache for ephemeral runners that need cross-run reuse of Git state.

### Implementation details

- Added `persistRepoDataType` as a new global config option.
- Added S3 restore before repository init so an existing `.git/` archive can be reused and `git fetch` can run against restored repository state instead of starting from a fresh clone.
- Added S3 archive after repository processing to persist `.git/` back to S3 for later runs.
- The upload path streams the generated archive directly to S3 instead of buffering it fully in memory, which is important for large repositories.
- The persisted archive is now a zstd-compressed tarball rather than gzip.
- Reused the existing shared S3 utilities in `lib/util/s3.ts`.
- Kept local cleanup behavior unchanged; S3 persistence is additive and does not decide whether the worktree is deleted.
- The repository worker imports the S3 persistence helpers directly, avoiding any ambiguity about emitted import extensions during compilation.

### Patterns followed

- S3 key layout mirrors `RepoCacheS3`: `{prefix}{platform}/{repository}/git-data.tar.zst`.
- Missing trailing slashes on the configured S3 prefix are normalized with a warning, matching existing S3 cache behavior.
- Restore error handling follows the existing S3 cache patterns: `NoSuchKey` falls back quietly, while other errors warn and fall back.
- Restore validates the extracted archive by checking for `.git/HEAD`; if extraction succeeds but the result is not a usable Git directory, the partial `.git/` directory is cleaned up.
- The upload path streams the generated tar archive through zstd compression into S3 instead of buffering the full archive in memory first; buffering the whole `.git` archive would not be viable for the large repositories this feature is meant to help.
- The shared S3 client configuration is reused so uploads and downloads also work with S3-compatible endpoints.

## Context

## AI assistance disclosure

Verified with a local direct dry run against an S3-compatible endpoint

### DIFF
--- a/docs/usage/self-hosted-configuration.md
+++ b/docs/usage/self-hosted-configuration.md
@@ -1170,7 +1170,7 @@ Use an `s3://` URL to archive the `.git/` directory to S3 after each run and res
 ```ts title="Set persistRepoDataType to an S3 URI to enable S3 backed git data persistence"
 {
   persistRepoData: true,
-  persistRepoDataType: 's3://bucket-name/renovate-git/';
+  persistRepoDataType: 's3://bucket-name/renovate-git/',
 }
 ```
 
--- a/docs/usage/self-hosted-configuration.md
+++ b/docs/usage/self-hosted-configuration.md
@@ -1155,6 +1155,42 @@ Set this to `true` if you want Renovate to persist repo data between runs.
 The intention is that this allows Renovate to do a faster `git fetch` between runs rather than `git clone`.
 It also may mean that ignored directories like `node_modules` can be preserved and save time on operations like `npm install`.
 
+## persistRepoDataType
+
+<!-- prettier-ignore -->
+!!! warning "This feature is flagged as experimental"
+    Experimental features might be changed or removed at any time.
+
+Set the storage backend for persisted repository git data when [`persistRepoData`](#persistrepodata) is `true`.
+
+By default, Renovate keeps repository data on the local filesystem (`local`).
+On ephemeral infrastructure (e.g., AWS Fargate, Kubernetes Jobs), local data is lost between runs.
+Use an `s3://` URL to archive the `.git/` directory to S3 after each run and restore it before the next run, enabling fast `git fetch` instead of full `git clone`.
+
+```ts title="Set persistRepoDataType to an S3 URI to enable S3 backed git data persistence"
+{
+  persistRepoData: true,
+  persistRepoDataType: 's3://bucket-name/renovate-git/';
+}
+```
+
+Renovate uses the [AWS SDK for JavaScript V3](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/welcome.html) to connect to the S3 instance.
+Therefore, Renovate supports all the authentication methods supported by the AWS SDK.
+Read more about [the default credential provider chain for AWS SDK for JavaScript V3](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-credential-providers/#fromnodeproviderchain).
+
+The S3 object key follows the pattern `{prefix}{platform}/{repository}/git-data.tar.gz`.
+
+This option reuses [`s3Endpoint`](#s3endpoint) and [`s3PathStyle`](#s3pathstyle) for S3-compatible storage.
+
+<!-- prettier-ignore -->
+!!! tip
+    You may set a folder hierarchy as part of `persistRepoDataType`.
+    For example, `persistRepoDataType: 's3://bucket-name/dir1/.../dirN/'`.
+
+<!-- prettier-ignore -->
+!!! note
+    This option persists git repository data (the `.git/` directory) to enable fast fetches, and is separate from the repository cache ([`repositoryCacheType`](#repositorycachetype)) and the lookup cache ([`redisUrl`](#redisurl)). If the S3 archive is missing (first run) or corrupted, Renovate falls back to a full `git clone` automatically.
+
 ## platform
 
 ## prCacheSyncMaxPages
--- a/lib/util/git/s3-persist.ts
+++ b/lib/util/git/s3-persist.ts
@@ -1,0 +1,127 @@
+import { PassThrough, Readable } from 'node:stream';
+import { pipeline } from 'node:stream/promises';
+import type {
+  GetObjectCommandInput,
+  PutObjectCommandInput,
+} from '@aws-sdk/client-s3';
+import { GetObjectCommand, PutObjectCommand } from '@aws-sdk/client-s3';
+import fs from 'fs-extra';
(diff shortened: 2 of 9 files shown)


lib/workers/repository/index.spec.ts
```diff
@@ -1,32 +1,85 @@
...
   describe('renovateRepository()', () => {
     let config: RenovateConfig;
+    let localDir: string;
 
-    beforeEach(() => {
-      config = getConfig();
-      config.localDir = '';
+    beforeEach(async () => {
+      GlobalConfig.reset();
+      localDir = await fs.mkdtemp('/tmp/renovate-repository-');
```
viceice: use `tmp-promise`

lib/util/git/s3-persist.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (!(await fs.pathExists(gitDir))) {
+      logger.debug('archiveGitDataToS3() - no .git directory found, skipping');
+      return;
+    }
+
+    const { bucket, key } = getS3Key(s3Url, platform, repository);
+
+    try {
+      const archiveStream = tar.create(
+        { gzip: true, cwd: localDir, portable: true },
```
viceice: why gzip, we can use brotli or zstd which are much better to save bandwidth .

- <https://nodejs.org/docs/latest-v24.x/api/zlib.html>

lib/workers/repository/index.ts
```diff
@@ -80,6 +80,23 @@ export async function renovateRepository(
       try {
         await fs.ensureDir(localDir);
         logger.debug('Using localDir: ' + localDir);
+        if (
+          repoConfig.persistRepoData &&
+          repoConfig.persistRepoDataType?.startsWith('s3://')
+        ) {
+          const { restoreGitDataFromS3 } =
+            await import('../../util/git/s3-persist.ts');
```
viceice: will ts-down update the extension on compile?

viceice (commented):
please extedn PR description why storing repos on s3 is better than cloning with blobless filter.

lib/workers/repository/index.spec.ts
ZackKanter: Fixed in b084628506f539d812a9d969f84dadfc23dd4faf

lib/util/git/s3-persist.ts
ZackKanter: Fixed in e8c54f859a917ae66c2801155967e6ef59c7d0b1

lib/workers/repository/index.ts
ZackKanter: Good point — changed to a static import so we don’t depend on tsdown rewriting the .ts extension here.

ZackKanter:
> please extedn PR description why storing repos on s3 is better than cloning with blobless filter.

Updated.

viceice:
@secustor @jamietanna wdyt?

docs/usage/self-hosted-configuration.md
```diff
@@ -1155,6 +1155,42 @@ Set this to `true` if you want Renovate to persist repo data between runs.
 The intention is that this allows Renovate to do a faster `git fetch` between runs rather than `git clone`.
 It also may mean that ignored directories like `node_modules` can be preserved and save time on operations like `npm install`.
 
+## persistRepoDataType
```
secustor, line 1158: Is there an usecase to make this a different destination than `repoCacheType`? 

Do you have tested the improvement for this? I'm curios what the performance improvement looks like when this is enabled.

docs/usage/self-hosted-configuration.md
```diff
@@ -1155,6 +1155,42 @@ Set this to `true` if you want Renovate to persist repo data between runs.
...
+  persistRepoData: true,
+  persistRepoDataType: 's3://bucket-name/renovate-git/',
+}
+```
+
+Renovate uses the [AWS SDK for JavaScript V3](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/welcome.html) to connect to the S3 instance.
+Therefore, Renovate supports all the authentication methods supported by the AWS SDK.
+Read more about [the default credential provider chain for AWS SDK for JavaScript V3](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-credential-providers/#fromnodeproviderchain).
+
+The S3 object key follows the pattern `{prefix}{platform}/{repository}/git-data.tar.gz`.
```
secustor, line 1181: We should have host in there too. 
There is the possibility for collisions that way. 

E.g. A company having a GitHub.com and GHE.com setup

docs/usage/self-hosted-configuration.md
secustor, line 1158: Personally I don't like putting type and destination in field as it makes the code messier and allow only one S3 strategy, but I understand it is following `repoCache` pattern.

lib/util/git/s3-persist.ts
```diff
@@ -0,0 +1,147 @@
...
+  });
+}
+
+export async function archiveGitDataToS3(
+  localDir: string,
+  s3Url: string,
+  platform: string,
+  repository: string,
+): Promise<void> {
+  await instrument('archiveGitToS3', async () => {
```
secustor, line 115: I would rather have `persistRepoData` span with attributes showing the difference.

secustor (commented):
First skim

jamietanna:
> Do you have tested the improvement for this? I'm curios what the performance improvement looks like when this is enabled.

I'd also +1 this - is this something y'all are using at Stedi, Zack? 

Note that we're also in the process of making it possible to use shallow cloning, which may help too

PR #42283, opened 2026-03-31
fix(datasource/github-digest): add github-digest to GITHUB_API_USING_HOST_TYPES

Shegox:
## Changes

We noticed when consuming GitHub Actions from a GitHub Enterprise Server, the new GitHub digest lookups (#40225 & #40226) don't work out of the box as the platform credentials are not used for lookup.
Resulting in such warnings (and failed lookups) for packages on our internal GitHub Enterprise Server:

grafik

```json
{
  "warnings": [
    "Failed to look up github-digest package $ORG/$REPO: no-result"
  ],
  "files": [
    ".github/workflows/build.yaml"
  ]
}
```

This PR adds the datasource `github-digest` to the `GITHUB_API_USING_HOST_TYPES` to align with how other (github) datasources are handled. While new datasources, which consume `api.github.com`, should not be added there, I think it is reasonable to add datasources using custom GitHub Enterprise Servers endpoints (like `github-digest`). 

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: # 
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

Tested on an internal repository on our GitHub Enterprise Server: 
grafik

The public repository: <URL>

PR #42287, opened 2026-03-31
fix(http): fallback to github hostType for GHE platform endpoint

zharinov:
## Changes

When running against GitHub Enterprise Server, datasources not listed in `GITHUB_API_USING_HOST_TYPES` (e.g. `github-digest`) fail to authenticate because the existing fallback only covers `api.github.com` URLs.

This adds a host-matching fallback: when the active platform is `github` and the request URL targets the same host as the configured endpoint, credentials from the `github` hostType rule are applied.

This makes the existing `api.github.com` workaround unnecessary for GHE — the fallback is now derived from the platform configuration rather than hardcoded URLs.

## Context

- Closes: #42283

The issue in #42283 added `github-digest` to `GITHUB_API_USING_HOST_TYPES` as a workaround, but this violates the deprecation guidance in that constant. This fix addresses the root cause: the host-rule fallback logic didn't account for GHE platform endpoints.

## AI assistance disclosure

### DIFF
--- a/lib/util/http/host-rules.spec.ts
+++ b/lib/util/http/host-rules.spec.ts
@@ -570,6 +570,57 @@ describe('util/http/host-rules', () => {
     });
   });
 
+  describe('GHE platform endpoint fallback', () => {
+    beforeEach(() => {
+      GlobalConfig.set({
+        platform: 'github',
+        endpoint: 'https://ghe.example.com/',
+      });
+      hostRules.clear();
+      hostRules.add({
+        hostType: 'github',
+        matchHost: 'ghe.example.com',
+        token: 'ghe-token',
+      });
+    });
+
+    afterEach(() => {
+      GlobalConfig.reset();
+    });
+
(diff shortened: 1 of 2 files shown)


lib/util/http/host-rules.spec.ts
```diff
@@ -570,6 +570,57 @@ describe('util/http/host-rules', () => {
...
+      hostRules.add({
+        hostType: 'github',
+        matchHost: 'ghe.example.com',
+        token: 'ghe-token',
+      });
+    });
+
+    afterEach(() => {
+      GlobalConfig.reset();
+    });
```
viceice: remove, the set in before each is overwriting previous config

viceice (commented):
otherwise LGTM

jamietanna:
cc @Shegox not sure if you fancy testing this in your environment pre-merge? (If that's possible)

Shegox (approved):
Tested the pull request as well against our local setup and can confirm that it works as expected. 👍 

Image

jamietanna (approved):

PR #42303, opened 2026-04-01
fix(swift): don't write `from:` range to `Package.resolved`

enomoto:
## Changes

When a dependency uses a range specifier like `from: "1.0.0"` in Package.swift, the `newValue` produced by the versioning module contains the range expression (e.g. `from: "603.0.0"`) rather than a plain version string. Using `newValue` directly to update `Package.resolved` results in the lock file containing the range expression instead of a resolved version, which corrupts it.

Fix by using `dep.newVersion` with the `v` prefix stripped, which is the actual resolved version and the correct value for `Package.resolved`.

This is a regression introduced by #41782, which fixed the `v` prefix issue by switching from `newVersion` to `newValue` — but `newValue` can contain range expressions for `from:`-style dependencies.

## Context

- [x] This closes an existing Issue, Closes: #42188

## AI assistance disclosure

This PR was written with Claude Code (claude-sonnet-4-6).

A failing test was written first to reproduce the bug, then the fix was applied to make it pass. All existing tests (including those added in #41782) continue to pass.

### DIFF
--- a/lib/modules/manager/swift/artifacts.ts
+++ b/lib/modules/manager/swift/artifacts.ts
@@ -208,17 +208,19 @@ export async function updateArtifacts({
         continue;
       }
 
+      const resolvedVersion = dep.newVersion.replace(/^v/, '');
+
       // Skip if already up-to-date
-      if (pin.state.version === newValue) {
+      if (pin.state.version === resolvedVersion) {
         logger.debug(
-          { depName: dep.depName, newValue },
+          { depName: dep.depName, resolvedVersion },
           'swift: pin already at target version',
         );
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


lib/modules/manager/swift/artifacts.ts
```diff
@@ -208,17 +208,19 @@ export async function updateArtifacts({
         continue;
       }
 
+      const resolvedVersion = dep.newVersion.replace(/^v/, '');
```
viceice:
```suggestion
      const resolvedVersion = dep.newVersion.replace(regEx(/^v/), '');
```

viceice (commented):
otherwise LGTM

viceice (approved):

PR #42310, opened 2026-04-01
fix(swift): strip range prefix from newVersion in Package.resolved updates

bdolewski-intellias:
## Changes

Fix a bug in the Swift `updateArtifacts` function where `dep.newVersion` may contain a Swift Package Manager range prefix (e.g. `from: "37.0.0"` instead of `37.0.0`), producing corrupted JSON in `Package.resolved`:

```json
"version" : "from: "37.0.0""
```

This is a follow-up to #41534, which added `Package.resolved` lockfile support to the Swift manager.

### Root cause

`dep.newVersion` can arrive wrapped in a range operator from `Package.swift` syntax. The existing code only strips the `v` prefix but not range operators like `from:`, `exact:`, `upToNextMajor:`, or `upToNextMinor:`. When the range-prefixed value is written directly into the `"version"` field, the resulting JSON is invalid (embedded unescaped quotes).

### Fix

Add an `extractPlainVersion()` helper that strips Swift range prefixes before writing to `Package.resolved`. The function handles:
- `from: "1.0.0"` → `1.0.0`
- `exact: "1.0.0"` → `1.0.0`
- `upToNextMajor: "1.0.0"` → `1.0.0`
- `upToNextMinor: "1.0.0"` → `1.0.0`
- `v1.0.0` → `1.0.0` (existing behavior preserved)

## Context

- Follow-up to #41534 (feat: update Package.resolved when updating Package.swift)
- [x] I have read the [Renovate Docs](https://docs.renovatebot.com/)

## AI assistance disclosure

7 new parametrized test cases for `extractPlainVersion()` plus an integration test that reproduces the exact corruption scenario. All 35 tests pass.

```
 ✓ lib/modules/manager/swift/artifacts.spec.ts (35 tests) 11ms
```

### DIFF
--- a/lib/modules/manager/swift/artifacts.spec.ts
+++ b/lib/modules/manager/swift/artifacts.spec.ts
@@ -3,7 +3,7 @@ import { GitTagsDatasource } from '../../datasource/git-tags/index.ts';
 import { GithubTagsDatasource } from '../../datasource/github-tags/index.ts';
 import { GitlabTagsDatasource } from '../../datasource/gitlab-tags/index.ts';
 import * as datasource from '../../datasource/index.ts';
-import { updateArtifacts } from './artifacts.ts';
+import { extractPlainVersion, updateArtifacts } from './artifacts.ts';
 
 vi.mock('../../../util/fs/index.ts');
 vi.mock('../../datasource/index.ts', async () => {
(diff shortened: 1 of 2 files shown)


jamietanna:
Looks like a duplicate of https://github.com/renovatebot/renovate/pull/42303 ?

bdolewski-intellias:
> Looks like a duplicate of #42303 ?

@jamietanna I wanted to polish that PR (hence the draft first) but you've got a great point. I'll take a look if I haven't accidentally duplicated #42303 :)

bdolewski-intellias:
@jamietanna you're right and #42310 is even better than my fix. Closing it.

PR #42326, opened 2026-04-02
feat(terraform): support minimumReleaseAge for lockfiles

arnaud-dezandee:
## Changes

Adds `minimumReleaseAge` support to Terraform lock file maintenance.

During lock file maintenance Renovate resolves provider versions itself rather than delegating to Terraform. It now selects the newest provider version which satisfies the existing lock file constraints *and* passes `minimumReleaseAge` / `minimumReleaseAgeBehaviour`, instead of always taking the newest satisfying version. If every satisfying release is still pending, the lock file is left unchanged.

## Context

Related: #41652

## AI assistance disclosure

AI was used for code exploration, implementation drafting, test generation and PR text. Originally GitHub Copilot; the rescope onto the upstream helper, the import cycle diagnosis, tests and docs were done with Claude Code (Claude Opus 5).


lib/modules/manager/terraform/lockfile/index.ts
```diff
@@ -23,8 +27,71 @@ import {
   writeLockUpdates,
 } from './util.ts';
 
+function getLatestAllowedVersion(
+  lock: ProviderLock,
+  releases: Release[],
+  config: UpdateArtifact['config'],
```
viceice:
```suggestion
  config: UpdateArtifactsConfig,
```

lib/modules/manager/terraform/lockfile/index.ts
```diff
@@ -23,8 +27,71 @@ import {
...
+      `Skipping lockfile maintenance release ${release.version} without releaseTimestamp`,
+    );
+  }
+
+  return null;
+}
+
 async function updateAllLocks(
   locks: ProviderLock[],
+  config: UpdateArtifact['config'],
```
viceice:
```suggestion
  config: UpdateArtifactsConfig,
```

lib/modules/manager/terraform/lockfile/index.ts
viceice, line 110: pass minimum release age here and the `getPkgReleases` internals will already filter for you, so no need for manual filtering

viceice (commented):
seems independed to datasource changes and if the `getPkgReleases` call is corrected, then it should already work for terraform registry because it already returns the time stamp.

arnaud-dezandee:
Applied the UpdateArtifactsConfig type fix, thanks.
I looked into passing `minimumReleaseAge` to `getPkgReleases` but I don't see it on `GetPkgReleasesConfig`
Am I missing something, or would that need to be added first?

joseinclan682-cpu (changes requested):
[](url)

joseinclan682-cpu (approved):

lib/modules/manager/terraform/lockfile/index.ts
arnaud-dezandee: I looked into passing minimumReleaseAge to getPkgReleases but I don't see it on GetPkgReleasesConfig
Am I missing something, or would that need to be added first?

viceice: i see, it's happening in worker

https://github.com/renovatebot/renovate/blob/7344b93bb0e79ca2269eefe957f4d4bbf291d5b0/lib/workers/repository/process/lookup/filter-checks.ts#L26

i think we need to refactor / extract a util function to do the filtering, so it can be reused instead of duplicate the same logic.

@jamietanna @secustor WDYT?

arnaud-dezandee: I extracted a `checkMinimumReleaseAge` util function only for now. It's the minimal logic that is required for the lockfile side. Did you mean to refactor a larger part of the `filterInternalChecks` function?

jamietanna: (Thinking out loud) I'm wondering where's best to have this logic - we are doing some filtering at the datasource layer (i.e. `constraintsFiltering`) but the `minimumReleaseAge` checks in the worker code does several other checks - I'm not sure if filtering it in the Datasource is correct as it may miss out on information like "this is a vulnerability alert so we should still allow it through"

(For this refactor, it'd be good as a separate PR, before this PR goes in, so we can change it in a cleaner way than including in with these changes - makes life simpler for everyone :) )

jamietanna: I do agree we should have a helper utility that does this centrally and can be applied wherever needed, though

arnaud-dezandee: The refactor landed in #44824 / #44825 while this sat, so I've dropped my competing extraction and rebuilt on `checkMinimumReleaseAge`.

Consuming it from a manager isn't possible where it currently lives. Importing `filter-checks.ts` forms an import cycle, since it pulls in `config/index.ts`, which loads every manager via `manager/api.ts`:

```
ReferenceError: Cannot access 'managers' before initialization
    at lib/modules/manager/index.ts:16
```

So I moved `checkMinimumReleaseAge` to `lib/util/minimum-release-age.ts`. No behaviour change, tests moved with it, and the other three helpers stay put as they're lookup-worker concerns.

@jamietanna you wanted the refactor as a separate PR ahead of this one. Happy to split it out, it's ~4 files on its own.

jamietanna: Hah yes, this utility was added for some other functionality, I'd forgotten we'd talked about it here.

Yes please, the moving of `checkMinimumReleaseAge` into `lib/util/minimum-release-age.ts` would be a good prefactor PR - thanks in advance for raising that!

docs/usage/key-concepts/minimum-release-age.md
```diff
@@ -141,14 +141,22 @@ Depending on your manager, datasource and the given package(s), it may be that s
...
 
 You can validate which update types may have release timestamps by following something similar to how [verify if the registry you're using](#which-registries-support-release-timestamps).
 
+#### Lock file maintenance
+
+Where Renovate resolves the versions itself instead of delegating to the package manager, `minimumReleaseAge` and `minimumReleaseAgeBehaviour` are applied to that resolution:
+
+- `terraform`, when updating `.terraform.lock.hcl`: Renovate selects the newest provider version which satisfies the existing lock file constraints _and_ passes the minimum release age check.
+
+For `npm`, Renovate instead passes the cutoff to the package manager as [`--before`](#npm).
```
jamietanna:
```suggestion
Generally, Renovate delegates lock file maintenance to the package manager.

In some cases, Renovate performs the resolution of versions itself, which leads to Renovate applying `minimumReleaseAge` and `minimumReleaseAgeBehaviour` to the version resolution. This applies to the following managers:

- [Terraform manager](../modules/manager/terraform/index.md): when updating `.terraform.lock.hcl`: Renovate selects the newest provider version which satisfies the existing lock file constraints _and_ passes the minimum release age check
```

jamietanna (commented):
Otherwise LGTM, thanks!

(And feel free to rebase your branch if that'd be more helpful for you as the author)

lib/modules/manager/terraform/lockfile/index.ts
```diff
@@ -24,8 +27,60 @@ import {
...
+    );
+    logger.once.debug(
+      {
+        depName: lock.packageName,
+        versions: versionsWithoutTimestamp,
+        check: 'minimumReleaseAge',
+      },
+      `${versionsWithoutTimestamp.length} release(s) did not have a releaseTimestamp, but as we're running with minimumReleaseAgeBehaviour=timestamp-optional, proceeding`,
+    );
+  }
```
jamietanna, line 93: ❓ we're not checking whether we are running in `minimumReleaseAgeBehaviour=timestamp-optional` here? so this seems like it'll fire regardless, which we don't want

arnaud-dezandee: Good catch. Added a guard against version downgrade bug too

lib/modules/manager/terraform/lockfile/index.ts
```diff
@@ -24,8 +27,77 @@ import {
...
+        return false;
+      }
+
+      return true;
+    })
+    .map((release) => release.version);
+
+  if (versionsWithoutTimestamp.length) {
+    const logMeta = {
+      depName: lock.packageName,
```
jamietanna, line 75: Might be worth making sure these lines are clearly part of the Terraform manager?

```suggestion
      manager: 'terraform',
      depName: lock.packageName,
```

But this would also de-duplicate them, so if there's already a log line that hits them (i.e. in another manager) then they'll show here too, so maybe not wanted

viceice:
conflicted

PR #42328, opened 2026-04-02
feat: use custom conventional commits detector

viceice:
## Changes
Use a simplified detector to remove a lot transitive dependencies (including deprecated).
It now also supports new commit styles:
- `fix!: some breaking change`
- `feat(scope)!: scoped breaking change`

## Context

## AI assistance disclosure

The public repository: <URL>


package.json
```diff
@@ -202,7 +202,6 @@
     "changelog-filename-regex": "2.0.1",
     "clean-git-ref": "2.0.1",
     "commander": "14.0.3",
-    "conventional-commits-detector": "1.0.3",
```
jamietanna, line 205: Have we looked at alternatively using i.e. `@semantic-release/commit-analyzer`? As it's [what semantic-release uses](https://github.com/semantic-release/semantic-release/blob/1c485a5836623fa1d7803152a9c742934ff96f5f/package.json#L30C6-L30C40)?

(I've not checked if it has similar functionality, passing thought)

lib/util/git/semantic.spec.ts
```diff
@@ -34,5 +34,23 @@ describe('util/git/semantic', () => {
...
+        'fix:',
+        'some:invalid',
+      ]);
+      const res = await detectSemanticCommits();
+      expect(res).toBe('disabled');
+    });
+
+    it('detects true on breaking changes', async () => {
+      config.semanticCommits = undefined;
+      git.getCommitMessages.mockResolvedValue(['fix!: foo']);
```
jamietanna, line 51: Can we also check a scoped breaking change, as that's one of the new things we've noted we add support for

lib/util/git/semantic.ts
```diff
@@ -27,3 +27,22 @@ export async function detectSemanticCommits(): Promise<DetectedSemanticCommit> {
...
+/**
+ * Detect semantic commits by counting the number of commit messages that match the Angular convention and comparing it to the number of commit messages that do not match the convention.
+ * The Angular convention is defined as: `<type>(<scope>): <description>`, where `<type>` is a word, `<scope>` is an optional word, and `<description>` is a string.
+ *
+ * @see Inspired by {@link https://github.com/conventional-changelog/conventional-commits-detector|conventional-commits-detector} and {@link https://www.conventionalcommits.org|conventional commits specification}.
+ * @param commitMessages commit messages to check
+ * @returns A number greater than zero if more semantic commits than non-semantic commits, less than zero if more non-semantic commits than semantic commits, or zero if equal number of semantic and non-semantic commits
+ */
+function detect(commitMessages: string[]): number {
+  const angular = regEx(/^(\w*)(?:\((.*)\))?!?: (.*)$/);
```
RahulGautamSingh, line 40:
```suggestion
  const angular = regEx(/^(\w+)(?:\((\w\-./]*)\))?!?: (.*)$/);
```

else it will match `(): foo` 

should not allow empty scope or be very generous and allow anything

lib/util/git/semantic.ts
```diff
@@ -16,9 +16,9 @@ export async function detectSemanticCommits(): Promise<DetectedSemanticCommit> {
   }
   const commitMessages = await getCommitMessages();
   logger.trace(`commitMessages=${JSON.stringify(commitMessages)}`);
-  const type = conventionalCommitsDetector(commitMessages);
+  const type = detect(commitMessages);
```
RahulGautamSingh: we should rename this `type` variable to he more infromative

RahulGautamSingh (changes requested):

lib/util/git/semantic.ts
viceice: it's a copy of
https://github.com/conventional-changelog/conventional-changelog/blob/61573ecba6aea718b96365c96993d4fa7d53c6fd/packages/conventional-changelog-conventionalcommits/src/parser.js#L3

it's the same what `conventional-commits-detector` used without the `!?` breaking change indicator

https://github.com/conventional-changelog/conventional-commits-detector/blob/6fa4155c7d1560c1b5e5ef2410ef606149f4d710/src/detectors.js#L4

which is the same as
https://github.com/conventional-changelog/conventional-changelog/blob/61573ecba6aea718b96365c96993d4fa7d53c6fd/packages/conventional-changelog-angular/src/parser.js#L3

package.json
viceice: i don't see a benefit to add a lot new transitive deps again for such a trivial regex match over commit messages.

https://github.com/semantic-release/commit-analyzer/blob/e8123811d10070965547bd4ab5cd82fc0db897ac/package.json#L21-L30

lib/util/git/semantic.ts
viceice: sugestions?

lib/util/git/semantic.ts
RahulGautamSingh: Yeah, seems to work sd we are only interested in form/structure

lib/util/git/semantic.ts
RahulGautamSingh: `semanticCommitScore` or `semanticCommitDelta`

jamietanna (approved):

mfeltscher:
This change breaks a whole bunch of our repositories since we're using merge commits that count towards the newly introduced score. Would it make sense to adjust this to look at the last 10 commits and use semantic commit messages if at least one commit contains such commit syntax? From my understanding that was the case before anyways.

viceice:
please post here:

- https://github.com/renovatebot/renovate/discussions/42353

PR #42384, opened 2026-04-05
refactor(lint): Switch to Oxlint

zharinov:
## Changes

Port the remaining ESLint rules, complete switch to Oxlint.

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: # 
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [x] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #42387, opened 2026-04-06
refactor(lint): Improve pnpm check

zharinov:
## Changes

Make `pnpm check` faster and more focused for local development:

- Targeted checks: `pnpm check lib/util/http` runs only relevant linters and tests instead of the full suite
- Multiple targets: `pnpm check lib/foo.ts lib/bar/` works with any mix of files and directories
- `--fix` flag: Replaces `--no-fix` — now check-only by default, opt in to auto-fix with `--fix`
- Faster execution: Lint and tests run in parallel instead of waiting for lint to finish first
- Cleaner coverage output: Shows only uncovered file paths and line numbers, no noise

## Context

## AI assistance disclosure


zharinov:
- Blocked by: #42384

.github/workflows/build.yml
```diff
@@ -195,7 +195,7 @@ jobs:
           os: ${{ runner.os }}
           save-cache: true
 
-  lint-eslint:
+  lint-oxlint:
```
viceice, line 198: @jamietanna you need to change the branch rule

tools/check/index.ts
```diff
@@ -1,22 +1,25 @@
...
  *
  * Options:
- *   --no-fix          Skip auto-fix (eslint-fix, prettier-fix)
+ *   --fix             Run fixers only (oxlint-fix, biome-fix, prettier-fix)
  *   --no-test         Skip tests
- *   --full            Run full vitest instead of only affected shards
  *   --base <branch>   Compare against different base branch (default: main)
  */
 
+import { readdirSync } from 'node:fs';
```
viceice, line 14: can we use async api please?

.github/workflows/build.yml
jamietanna: Yep, I'll do the branch protection status check change when this is merged via Merge Queue? (I'll make sure I'm the one to merge this)

jamietanna: Done as part of https://github.com/renovatebot/renovate/pull/42384

jamietanna:
Why remove branch-based logic? (I've not used it before)

zharinov:
Less magic is better. Just point the files you want to check instead of trying to detect them (bad heuristics).

viceice (approved):

PR #42398, opened 2026-04-06
fix(manager/github-actions): disable updates for bare SHA pins without version comment

shaanmajid:
## Changes

Disables GitHub Actions dependencies pinned to a bare SHA without a version comment, since Renovate cannot determine which branch or tag the SHA belongs to.

- #42245

## Context

(Discussion at #42245).

## AI assistance disclosure

The public repository: https://github.com/shaanmajid/gha-naked-sha-test-42398


lib/modules/manager/github-actions/extract.spec.ts
```diff
@@ -524,6 +524,78 @@ describe('modules/manager/github-actions/extract', () => {
...
+              - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v4
+        `,
+        'workflow.yml',
+      );
+      expect(res?.deps[0]).toMatchObject({
+        depName: 'actions/checkout',
+        currentDigest: 'c85c95e3d7251135ab7dc9ce3241c5835cc595a9',
+        currentValue: 'v4',
+      });
+      expect(res?.deps[0].enabled).toBeUndefined();
```
viceice: use single `toEqual`

lib/modules/manager/github-actions/readme.md
```diff
@@ -39,6 +39,9 @@ If you want to automatically pin action digests add the `helpers:pinGitHubAction
 }
 ```
 
+Actions pinned to a bare SHA without a version comment are disabled by default, because Renovate cannot determine which branch or tag the SHA belongs to.
+To enable updates for these dependencies, add a package rule matching `matchManagers: ["github-actions"]` with `enabled: true`.
```
viceice: or add a tag / branch name as version comment which is the preferred option

jamietanna: Although this does work, I'd probably recommend a different package rule for this - also ++ to `@viceice`'s comment that the recommendation should be "don't use a bare SHA"

lib/modules/manager/github-actions/extract.ts
```diff
@@ -130,6 +130,10 @@ function extractRepositoryAction(
     dep.currentValue = ref;
   }
 
+  if (!dep.currentValue) {
+    dep.enabled = false;
```
jamietanna, line 134: Should we also set a `skipReason`?

lib/modules/manager/github-actions/readme.md
shaanmajid: Incorporated the add version comment suggestion. What package rule would you prefer?

lib/modules/manager/github-actions/extract.ts
shaanmajid: `unversioned-reference` seems appropriate?

viceice (dismissed):

lib/modules/manager/github-actions/readme.md
```diff
@@ -39,6 +39,10 @@ If you want to automatically pin action digests add the `helpers:pinGitHubAction
 }
 ```
 
+Actions pinned to a bare SHA without a version comment are disabled by default, because Renovate cannot determine which branch or tag the SHA belongs to.
+To enable updates, add a tag or branch name as a version comment, as shown above.
+You can also re-enable these with a package rule matching `matchManagers: ["github-actions"]` and `enabled: true`, but Renovate updates the digest to the latest commit on the repository default branch, which may not match the original SHA's source.
```
jamietanna: IMO:

```suggestion
```

Or if we are doing it, let's put a full code block (as it's then validated) that explains a little more how we can target this more appropriately

shaanmajid:
Sorry, didn't mean to dismiss stale review -- think it automatically did that when applying the suggestion.

jamietanna (approved):
All good, thanks!

PR #42487, opened 2026-04-08
feat(manager/docker-compose): add ability to ignore package based on comment

FoxxMD:
## Changes

This change adds some logic to the [docker-compose](https://docs.renovatebot.com/modules/manager/docker-compose/) manager so that individual service/packages can be skipped using a trailing comment on a service's `image` line.

After a docker-compose file has been successfully parsed, during package extract, it checks for a comment on the raw yaml document associated with the `image` line and sets `skipReason` if it matches the existing `renovate: ignore` format. Example:

```yaml
version: "3"
services:
  nginx:
    image: quay.io/nginx:0.0.1 #renovate: ignore
```

## Context

## AI assistance disclosure

It's unclear if this should be documented as the other uses of `renovate: ignore` in [pip_requirements](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/pip_requirements/extract.ts#L41) and [jenkins](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/jenkins/extract.ts#L92) do not appear to be documented either.

The public repository: <URL>

### DIFF
--- a/lib/modules/manager/docker-compose/extract.spec.ts
+++ b/lib/modules/manager/docker-compose/extract.spec.ts
@@ -229,5 +229,66 @@ describe('modules/manager/docker-compose/extract', () => {
         ],
       });
     });
+
+    it('ignores dependency when skip comment is present', () => {
+      const compose = codeBlock`
+      version: "3"
+      services:
+        nginx:
+          image: quay.io/nginx:0.0.1 #renovate: ignore
+    `;
+      const res = extractPackageFile(compose, '', {});
+      expect(res).toEqual({
+        deps: [
+          {
+            autoReplaceStringTemplate:
+              '{{depName}}{{#if newValue}}:{{newValue}}{{/if}}{{#if newDigest}}@{{newDigest}}{{/if}}',
+            currentDigest: undefined,
+            currentValue: '0.0.1',
+            datasource: 'docker',
+            depName: 'quay.io/nginx',
+            packageName: 'quay.io/nginx',
+            replaceString: 'quay.io/nginx:0.0.1',
+            skipReason: 'ignored',
+          },
+        ],
+      });
+    });
+
+    it('ignores dependency when skip comment is present in fragment', () => {
+      const compose = codeBlock`
+        ---
+        x-shared_setting: &shared_settings
+          image: debian:11 #renovate: ignore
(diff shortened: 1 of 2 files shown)


viceice (commented):
i don't really like this

@jamietanna @secustor WDYT?

jamietanna:
I seem to remember us having similar asks in the past, but I'd say that using `packageRules` for this is our preference

FoxxMD:
Let me quickly explain the motivation:

In the homelab/self-hosted space most people organize and deploy their services using docker compose stacks. Most of the time none of these stacks are related. And often users aggregate them into monorepos where compose.yaml files are separated by folders.

Elsewhere in the homelab space, there are other discovery/meta services that use [labels](https://docs.docker.com/engine/manage-resources/labels/) in compose services to build their behavior. Examples: [traefik](https://doc.traefik.io/traefik/expose/docker/basic/), [homepage](https://gethomepage.dev/configs/docker/#automatic-service-discovery), [dozzle](https://dozzle.dev/guide/container-names#custom-names/)...The running theme in this space is that _responsibility for the service behavior should be defined next to/by the service itself._

This is counter to how renovate does this and it's obviously not feasible or desired to flip this model but the `#renovate: ignore` mechanism does give back a small bit of that control that would be familiar and extremely useful for this demographic: since most of these compose files are unrelated it would be useful to be able to proactively (and in-situ) ignore updates for supporting services. This, as opposed to, having to keep a `packageRules` entry in `renovate.json` updated with a growing list of arbitrary/unrelated images each time a user decides to deploy a new stack.

___

@viceice 

> i don't really like this

Is is the implementation you don't like, or the general idea? 

If it's the general idea, can I ask why `#renovate: ignore` is implemented at all, for pip/jenkins?

secustor:
I agree here with `@viceice` this seems like a good idea first, but if at all this would have to be general feature supported for all managers and not only `docker-compose` it will confuse users why this is the case. 

In regard IF there should be a feature like this, this would have to be opt in as in bigger setup ( thinking monorepos using CODEOWNERS scenarios ) it can not be desired that user can inline deactivate tracking / updating

FoxxMD:
Alright, that's fair.

Thanks for the consideration and the reply :+1:

PR #42491, opened 2026-04-08
fix(git): don't pass extraCloneOpts to ls-remote when local clone exists

Churro:
## Changes

This PR fixes a case where `git ls-remote` fails during a recursive `renovateRepository()` call after automerge. In `fetchBranchCommits()`, skip injecting `extraCloneOpts` as `-c` flags when `.git/HEAD` already exists in `localDir`.

With token auth, `getExtraCloneOpts()` sets up http.extraHeader, which gets persisted in `.git/config`. Until now, the directory survived the recursive call and `fetchBranchCommits` added those same options again via `-c`, producing duplicate HTTP headers that servers reject with HTTP 400. See https://github.com/renovatebot/renovate/discussions/42210#discussioncomment-16355646 for details.

## Context

- [x] This closes an existing Issue, Closes: #42305

## AI assistance disclosure

The public repository: <URL>


RahulGautamSingh (approved):
LGTM

DidierLoiseau:
I confirm this fixes the issue for me according to my local testing.

viceice (commented):
i don't like that solution. it slows down all platforms just for a bitbucket issue.
we should find a better fix.
eg don't pass extra header options to git fetch, so we don't set it twice. or ensure we remove that option from git repo after run

Churro:
> i don't like that solution. it slows down all platforms just for a bitbucket issue. we should find a better fix. eg don't pass extra header options to git fetch, so we don't set it twice. or ensure we remove that option from git repo after run

Agreed, I've changed the approach now to don't pass extra header options to `ls-remote`, if the repo already exists locally.

lib/util/git/index.ts
```diff
@@ -208,7 +208,9 @@ async function fetchBranchCommits(preferUpstream = true): Promise<void> {
     preferUpstream && config.upstreamUrl ? config.upstreamUrl : config.url;
   logger.debug(`fetchBranchCommits(): url=${url}`);
   const opts = ['ls-remote', '--heads', url];
-  if (config.extraCloneOpts) {
+  const localDir = GlobalConfig.get('localDir')!;
+  const repoExists = await fs.pathExists(upath.join(localDir, '.git/HEAD'));
+  if (config.extraCloneOpts && !repoExists) {
```
viceice, line 213: did you checked the call sites? doesn't the repo always already exist when we're here?  haven't checked it yet 🤔

Churro: Yes, I checked all callers.

- On the very first run, when the `git` instance is constructed, `initRepo()` -> `fetchBranchCommits()` is called **before** `syncGit()` called, so `localDir` is empty and `.git/HEAD` does not exist yet.
- For calls from `syncGit()`:
  - `extraCloneOpts` is set + no `.git/HEAD` exists -> pass -c flags
  - `extraCloneOpts`  is set + `.git/HEAD exists` -> skip -c flags because .git/config already has them from clone

All subsequent operations e.g., `prepareCommit`, `pushCommit`, etc. call `syncGit()` which returns early because `gitInitialized = true`, so they rely on the `http.extraHeader` in `.git/config`.

marquisesenior383-ctrl:
I didn't make it Lang did he keeps logging into my accounts

viceice (approved):

PR #42531, opened 2026-04-10
feat(cache): add a in-memory expiry map to package file cache

Thooms:
## Changes

This PR introduces a small in-memory map that stores expiry times for keys in the `PackageCacheFile` implementation.
We noticed in our deployment that it was starting to bottleneck on a non-small amount of repositories.

This allows for faster expiry lookup, which makes for the bulk of the loop on destroy(). Note that this doesn't remove the overhead of looping through all entries, but at least doesn't load the data from disk with a deserialization.

Result of the provided benchmark:

```
 ✓ lib/util/cache/package/impl/file.bench.ts > PackageCacheFile destroy() 8481ms
     name                                                           hz      min      max     mean      p75      p99     p995     p999     rme  samples
   · cold path — no expiryMap (disk read per entry)             2.8974   338.00   350.81   345.14   347.57   350.81   350.81   350.81  ±0.82%       10
   · warm path — expiryMap populated via set() (no disk read)  20.1797  48.4270  50.5818  49.5548  50.1291  50.5818  50.5818  50.5818  ±0.82%       11

 BENCH  Summary

  warm path — expiryMap populated via set() (no disk read) - lib/util/cache/package/impl/file.bench.ts > PackageCacheFile destroy()
    6.96x faster than cold path — no expiryMap (disk read per entry)
```    

## Context

## AI assistance disclosure

Tested also on an internal repository with indeed a ~50% speedup on cache cleanup.


zharinov:
Hey, will be able to review the PR in couple hours

zharinov (commented):
I don't agree with general direction of how this should be fixed. Stay tuned, I'm working on the PR which should help solve your problem using another approach.

Thooms:
Hey, thanks for the review; I just took a look at your PR: it was my first approach, before deciding to move to a simpler way (which also paves the way to a faster listing procedure, which is the bulk of the time).

That being said, happy to go either way, I'm merely interested in having this path run faster, no real strong opinion on the implem!

zharinov:
Yeah, I thought it's going to be 20 minutes adventure, it turned to be a real rabbit hole due to garbage collection concerns. And here I am, 20 commits later still not happy about `cacache`-based solution in general...

Anyways, I think the best course of action is to merge this PR and move on for now. The only thing concerning me is the lifecycle of the in-memory object. In some situations it could get big, so maybe limit its size to something about 50–100Mb?

Out of curiosity, would you consider SQLite if we harden its implementation a bit and promote to non-experimental?

zharinov:
Oh, and I think we maybe don't want to keep the benchmarks in the codebase. They're going to be an additional burden to maintain, unless we decided to maintain the baseline perf across all cache backends.

viceice (commented):
please drop benchmarks

Thooms:
Happy to remove the benchmark, I provided it to back the reasoning behind my PR, but no strong reason to keep it around in the codebase.

> so maybe limit its size to something about 50–100Mb

Let me see if I can use a small LRU or something, overall I don't think it's an issue but doesn't hurt to limit damage in extreme cases.

> Out of curiosity, would you consider SQLite if we harden its implementation a bit and promote to non-experimental?

Probably, but right now we're choosing to operate the simplest way we can, thus this PR!

zharinov:
> Let me see if I can use a small LRU or something, overall I don't think it's an issue but doesn't hurt to limit damage in extreme cases

MRU fits better? 🙂

> Probably, but right now we're choosing to operate the simplest way we can, thus this PR!

Yeah, SQLite could've solve this better. The model of `cacache` is append-only + periodic GC, and we're fighting against this model by making it incremental-ish. SQLite would help to amortise cleanup cost better. I guess at the moment of introducing `cacache`, the huge volumes of data weren't the concern. Like, just nuke the cache bro.

Thooms:
> MRU fits better? 🙂

I don't think it actually matters here, given that we lookup on everything, I'm merely using an LRU for the size-limiting properties here

As for SQLite, happy to try it out!

viceice (approved):
`lru-cache` v11 is already a transitive production dependency

zharinov (approved):

PR #42543, opened 2026-04-11
feat(cache): Optimize package cache `file` backend

zharinov:
## Changes

This refactors the package cache `file` backend to store expiry/version in `cacache` index metadata instead of inside the cached payload.

- New entries store the compressed value directly as cache content, with `expiry` and `version` in index metadata.
- Reads check expiry from metadata first, so current-format entries can usually be validated without reading and parsing the cached blob.
- Cleanup (`destroy()`) uses that same metadata to remove expired/invalid index entries more cheaply.
- Valid legacy entries are still supported and are lazily migrated to the new format during cleanup, without recompressing their stored value.
- Content GC is now safer: removing a cache key does not immediately delete the underlying cached blob, because another live key may still point to the same content. Cleanup first treats removed blobs as "maybe unused", keeps track of blobs still referenced by surviving entries, and only deletes blobs that are no longer referenced by anything live.

## Context

- Supersedes: #42531

## AI assistance disclosure

- [x] Yes — substantive assistance (code and tests were authored/refined with AI support).

### DIFF
--- a/lib/util/cache/package/impl/file.spec.ts
+++ b/lib/util/cache/package/impl/file.spec.ts
@@ -2,14 +2,33 @@ import cacache from 'cacache';
 import { DateTime } from 'luxon';
 import { type DirectoryResult, dir } from 'tmp-promise';
 import upath from 'upath';
-import { compressToBase64, decompressFromBuffer } from '../../../compress.ts';
+import {
+  compressToBase64,
+  compressToBuffer,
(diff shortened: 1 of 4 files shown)


viceice (commented):
conflicted 🙃

bbodenmiller:
Does this also address issue where old items in `content-v2` are never garbage collected even though they were removed from index?

PR #42557, opened 2026-04-12
fix(cache): Improve SQLite cache error handling

zharinov:
## Changes

- Harden the SQLite package cache so non-critical cache failures do not break execution:
  - `set()` now catches serialization/compression/write failures, logs once, and continues
  - `get()` now returns `undefined` on malformed cached data or SQLite read failures and logs once instead of throwing
  - `destroy()` now catches cleanup and close failures separately, still attempts `close()`, and logs/continues
- Set a short SQLite `busy_timeout` so cache operations fail fast under contention instead of hanging
- Expand SQLite cache tests to cover malformed stored data, failed reads, failed writes, and `destroy()` cleanup/close failure paths using the real database

## Context

## AI assistance disclosure


.github/workflows/build.yml
```diff
@@ -523,7 +523,7 @@ jobs:
           pnpm nyc check-coverage -t ./coverage/nyc \
             --statements 99.88 \
             --branches 97.84 \
-            --functions 99.96 \
+            --functions 99.95 \
```
viceice: why? 😢

lib/util/cache/package/impl/sqlite.ts
```diff
@@ -51,14 +53,15 @@ export class PackageCacheSqlite extends PackageCacheBase {
...
-  private readonly client: Database;
+  readonly client: Database;
 
   private constructor(client: Database) {
     super();
     this.client = client;
 
     client.pragma('journal_mode = WAL');
     client.pragma("encoding = 'UTF-8'");
+    client.pragma(`busy_timeout = ${PackageCacheSqlite.busyTimeoutMs}`);
```
viceice, line 64: we should allow override via env otherwise renovate will fail  often on slow disks

lib/util/cache/package/impl/sqlite.ts
```diff
@@ -51,14 +53,15 @@ export class PackageCacheSqlite extends PackageCacheBase {
   private readonly deleteExpiredRows: Statement<unknown[]>;
   private readonly countStatement: Statement<unknown[]>;
 
-  private readonly client: Database;
+  readonly client: Database;
```
viceice, line 56: why public? 🤔

viceice (commented):
please avoid coverage drops

.github/workflows/build.yml
zharinov:
```suggestion
            --functions 99.96 \
```

Accidental floating math I guess. Let's revert, rebase and try again.

lib/util/cache/package/impl/sqlite.ts
zharinov: Way more easy to test, not visible outside of cache internals anyways

viceice (commented):
still concerned about the busy timeout. we recently had big issues on forgejo with too low value and concurrent access

zharinov:
What access patterns for SQLite could be?

For many Renovate instances accessing single SQLite database? 1–5 sec maybe. But would anyone setup many instances on one machine, sharing a single database? 🤔

viceice:
> What access patterns for SQLite could be?
> 
> For many Renovate instances accessing single SQLite database? 1–5 sec maybe. But would anyone setup many instances on one machine, sharing a single database? 🤔

it's ürobably safe here as we only have one shared connection and no real multithreading. On Forgejo we've multiple connections and they can cause issues with too low busy timeout.

viceice (approved):

Issue #42567, opened 2026-04-12
feat(datasource/go): expose `constraints` for discovered `go` directive


jamietanna:
Via https://github.com/renovatebot/renovate/discussions/42459

jamietanna:
This also needs:

- https://github.com/renovatebot/renovate/issues/31831
  - as the `go` directive extracted in https://github.com/renovatebot/renovate/pull/42582 will need its own versioning
- A way to match `1.24.0` with `1.24.1` (to allow an update within a patch bump)
- https://github.com/renovatebot/renovate/pull/42582

Issue #31831, opened 2024-10-07
Allow separate versioning for constraints 


rarkins:
### Describe the proposed change(s).

Ref: #26151

Idea: support a new option `constraintsVersioning` for the rare cases where constraints need a separate versioning to the lookups

PR #42582, opened 2026-04-13
feat(manager/gomod): extract the `go` and `toolchain` directives as `constraints`

jamietanna:
## Changes

As part of future changes in #42567 to surface the `constraints` of a
given Go module, we should first extract the constraints from the
package file.

To make this simpler for users, we can make sure we extract these at
parse time:

- the `toolchain` directive becomes the `go` constraint, as it is the
  toolchain actually running Go, and mirrors how a `constraints.go`
  would be expected to control the `go` toolchain version for
  Containerbase
- the `go` directive becomes a new `go-mod` constraint, which keeps it
  separate from the version of the Go toolchain actually used to run

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/manager/gomod/extract.spec.ts
+++ b/lib/modules/manager/gomod/extract.spec.ts
@@ -1,5 +1,7 @@
 import { codeBlock } from 'common-tags';
 import { Fixtures } from '~test/fixtures.ts';
+import { getDefaultVersioning } from '../../datasource/common.ts';
+import * as allVersioning from '../../versioning/index.ts';
 import { extractPackageFile } from './index.ts';
 
 const gomod1 = Fixtures.get('1/go-mod');
@@ -122,6 +124,12 @@ describe('modules/manager/gomod/extract', () => {
             datasource: 'go',
           },
         ],
+        extractedConstraints: {
+          '%goMod': '~1.23.x',
+        },
+        constraintsVersioning: {
+          '%goMod': 'semver-coerced',
+        },
       });
     });
 
@@ -172,6 +180,12 @@ describe('modules/manager/gomod/extract', () => {
             versioning: 'loose',
           },
         ],
+        extractedConstraints: {
+          '%goMod': '~1.25.x',
+        },
+        constraintsVersioning: {
+          '%goMod': 'semver-coerced',
+        },
       });
     });
 
@@ -236,6 +250,13 @@ describe('modules/manager/gomod/extract', () => {
             datasource: 'go',
           },
         ],
+        extractedConstraints: {
+          '%goMod': '~1.23.x',
+          golang: '1.23.3',
+        },
+        constraintsVersioning: {
+          '%goMod': 'semver-coerced',
+        },
       });
     });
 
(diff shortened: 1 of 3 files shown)


lib/modules/manager/gomod/extract.spec.ts
```diff
@@ -121,6 +121,9 @@ describe('modules/manager/gomod/extract', () => {
             datasource: 'go',
           },
         ],
+        extractedConstraints: {
+          'go-mod': '1.23',
```
jamietanna: Or maybe `go-directive`?

viceice:
```suggestion
          'go': '1.23',
```

it should match what we use for `constraints` / `install-tool`

lib/modules/manager/gomod/extract.ts
```diff
@@ -72,5 +72,31 @@ export function extractPackageFile(content: string): PackageFileContent | null {
...
+  const toolchainDirective = packageFile.deps.find(
+    (dep) =>
+      dep.depName === 'go' &&
+      dep.depType === 'toolchain' &&
+      dep.datasource === 'golang-version',
+  );
+  if (toolchainDirective?.currentValue) {
+    packageFile.extractedConstraints ??= {};
+    packageFile.extractedConstraints.go = toolchainDirective.currentValue;
+  }
```
viceice, line 105: we should prefer toolchain over directive and only set `go` as `extractedConstraints`. see other comment.

lib/modules/manager/gomod/extract.spec.ts
```diff
@@ -232,6 +238,10 @@ describe('modules/manager/gomod/extract', () => {
...
```
jamietanna: It'd be nice if we could extract this as a `^1.23` (or in a way that allows `1.23` to be treated as `1.23.x`) to allow patch bumps

lib/modules/manager/gomod/extract.spec.ts
jamietanna: So this is a case where we want to keep them separate - the `go` directive in `go.mod` is the _source compatibility_ of the code, which is different to the `toolchain` / `go` constraint that will actually run the code

viceice:
```suggestion
          'goMod': '1.23',
```
or

```suggestion
          '%go-mod': '1.23',
```

what about a prefix to better distinghish from real tool constraints? so we don't have an issue if we add a `go-mod` tool later. tool names will ever only be lowercase. `@` will be used for npm scopes, so we can't use that.

jamietanna: Making it an explicit alternative sounds good - note that https://github.com/renovatebot/renovate/pull/42589 has highlighted we do already have `gomodMod` for https://github.com/renovatebot/renovate/blob/50a2bacb2b0c5eacdc7b98f06df0cb2c5fd94160/lib/modules/manager/gomod/artifacts.ts#L71-L96 

(But we don't want to use that here)

jamietanna:
We'll want to rethink this to tweak with how we do it in https://github.com/renovatebot/renovate/blob/HEAD/lib/modules/manager/gomod/artifacts.ts#L195 - we should have these consistent, so we're aware we'll always have a `constraints?.go` set to the toolchain to use

lib/modules/manager/gomod/extract.ts
```diff
@@ -72,5 +72,44 @@ export function extractPackageFile(content: string): PackageFileContent | null {
...
+    (dep) =>
+      dep.depName === 'go' &&
+      dep.depType === 'golang' &&
+      dep.datasource === 'golang-version',
+  );
+  if (goDirective?.currentValue) {
+    packageFile.extractedConstraints ??= {};
+    // TODO #31831
+    packageFile.extractedConstraints['%goMod'] =
+      convertGoDirectiveToSemVerRange(goDirective.currentValue);
```
jamietanna: 🤔 

```
 WARN: Invalid constraint used with strict constraintsFiltering (repository=local)
       "packageName": "golang.org/x/mod",
       "constraint": "^1.24.x",
       "versioning": "semver",
```

jamietanna:
OK, this is blocked by https://github.com/renovatebot/renovate/issues/31831

lib/modules/manager/gomod/extract.spec.ts
```diff
@@ -495,6 +510,91 @@ describe('modules/manager/gomod/extract', () => {
...
+            managerData: {
+              lineNumber: 1,
+            },
+            depName: 'go',
+            depType: 'golang',
+            currentValue: goDirective,
+            datasource: 'golang-version',
+            versioning: 'go-mod-directive',
+            commitMessageTopic: 'go module directive',
+          },
```
jamietanna, line 548: Should we maybe return `constraintsVersioning` here, to make sure it's explicitly returned by the `gomod` manager, rather than a config-level default?

jamietanna (commented):
It should be possible to set i.e.

```
	"constraints": {
		"%goMod": "~1.25.x"
	},
```

If you want to be extra explicit?

lib/modules/manager/gomod/extract.spec.ts
viceice: you would need to repeat it for every dep. i don't like it, it will blow up the object even more 🤔

viceice (approved):

lib/modules/manager/gomod/extract.spec.ts
jamietanna: Sorry, should've resolved this - I was going to do so, but then I found that it was more straightforward to do at a packageFile level

It also makes a bit more sense to be per file level - a dependency probably doesn't know enough about itself and other things to be able to override it

We can see in the future, but this functionality is very likely to be sparingly used

PR #42099, opened 2026-03-24
feat(mise): add lock file support

altendky:
## Summary

Adds lock file support to the mise package manager, enabling Renovate to:
- Extract locked versions from `mise.lock` files
- Update lock files when dependencies change
- Support lock file maintenance mode

Closes #40568

### New Files
- **`lib/modules/manager/mise/lockfile.ts`** - `getLockFileName()` utility for deriving lock file paths
- **`lib/modules/manager/mise/lockfile.spec.ts`** - Tests for lock file path derivation
- **`lib/modules/manager/mise/artifacts.ts`** - `updateArtifacts()` and `updateLockedDependency()` functions
- **`lib/modules/manager/mise/artifacts.spec.ts`** - Tests for artifact updates

### Modified Files
- **`lib/modules/manager/mise/index.ts`** - Added exports for lock file support
- **`lib/modules/manager/mise/schema.ts`** - Added `MiseLockFile` Zod schema
- **`lib/modules/manager/mise/extract.ts`** - Made async, added `lockedVersion` extraction
- **`lib/modules/manager/mise/extract.spec.ts`** - Converted to async, added lock file tests

## Implementation Details

- Follows patterns from `nix` and `cargo` managers
- Handles TEMPORARY_ERROR for Renovate's retry mechanism
- Combines stdout/stderr/message in artifact errors for better debugging
- `updateLockedDependency()` returns `already-updated` when version matches

PR #42591, opened 2026-04-13
feat(mise): add lock file support

### LINKED ISSUE #40568: feat(manager/mise): support lock file maintenance
Mise supports has now experimental support for lock files and we should support [lockfileMaintenance](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance) for it. 

https://mise.jdx.dev/dev-tools/mise-lock.html

zeitlinger:
## Summary

Adds lock file support to the mise manager. This is a replacement for #42099 (@altendky's draft) with two additional fixes identified during review.

**What's included (from #42099):**
- Extract `lockedVersion` from `mise.lock` during package file extraction
- `updateArtifacts()`: run `mise lock [tools]` to update the lock file after dependency changes
- `updateLockedDependency()`: return `already-updated` when the locked version already matches
- Lock file maintenance support via `mise lock`
- `supportsLockFileMaintenance`, `lockFileNames` exports

**Additional fixes (not in #42099):**

**Fix 1 — `--local` flag for local config files**

`mise lock` without `--local` updates `mise.lock`, not `mise.local.lock`. Processing `mise.local.toml` was silently updating the wrong lock file.

**Fix 2 — `MISE_ENV` for environment-specific configs**

`mise lock` for `mise.test.toml` needs `MISE_ENV=test` to select the right config and update `mise.test.lock`. Without it, mise defaults to `mise.toml`.

Closes #40568


RahulGautamSingh (commented):
LGTM

lib/modules/manager/mise/artifacts.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+import { updateArtifacts, updateLockedDependency } from './artifacts.ts';
+import * as lockfile from './lockfile.ts';
+
+vi.mock('../../../util/exec/env.ts');
+vi.mock('../../../util/fs/index.ts');
+vi.mock('../../../util/git/index.ts');
+vi.mock('../../../util/host-rules.ts', () => mockDeep());
+vi.mock('./lockfile.ts', async (importOriginal) => ({
+  ...(await importOriginal<typeof lockfile>()),
+}));
```
viceice: why this? doesn't make sense to me to return the original module

lib/modules/manager/mise/artifacts.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+const updateToolCmd = 'mise lock node';
+const updateMultipleToolsCmd = 'mise lock node python';
+const lockfileMaintenanceCmd = 'mise lock';
+
+describe('modules/manager/mise/artifacts', () => {
+  beforeEach(() => {
+    env.getChildProcessEnv.mockReturnValue({
+      ...envMock.basic,
+      LANG: 'en_US.UTF-8',
+      LC_ALL: 'en_US',
```
viceice, line 39: why this?

lib/modules/manager/mise/artifacts.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+    });
+    GlobalConfig.set(adminConfig);
+    docker.resetPrefetchedImages();
+    hostRules.find.mockReturnValue({ token: undefined });
```
viceice: do not mock hostrules, just add a rule

lib/modules/manager/mise/artifacts.ts
```diff
@@ -0,0 +1,162 @@
...
+      },
+    ];
+  } catch (err) {
+    // istanbul ignore if: not worth testing
+    if (err.message === TEMPORARY_ERROR) {
+      throw err;
+    }
+
+    const errorOutput = [err.stdout, err.stderr, err.message]
+      .filter(Boolean)
```
viceice:
```suggestion
      .filter(isString)
```

lib/modules/manager/mise/artifacts.ts
```diff
@@ -0,0 +1,162 @@
...
+    ];
+  }
+}
+
+/**
+ * Check if a dependency is already at the target version in the lock file.
+ * Returns 'already-updated' if the locked version matches newVersion,
+ * 'unsupported' otherwise (letting Renovate proceed with the update).
+ */
+export function updateLockedDependency(
```
viceice: move to `./update-locked.ts`

lib/modules/manager/mise/extract.spec.ts
```diff
@@ -9,29 +10,34 @@ const miseFilename = 'mise.toml';
 const mise1toml = Fixtures.get('Mise.1.toml');
 
 describe('modules/manager/mise/extract', () => {
+  beforeEach(() => {
+    // By default, mock no lock file found
+    fs.readLocalFile.mockResolvedValue(null);
+  });
```
viceice:
```suggestion
```

leave default, it returns `undefined` when no explicit return value is set.

lib/modules/manager/mise/lockfile.ts
```diff
@@ -0,0 +1,100 @@
...
+  env?: string;
+}
+
+/**
+ * Parses the config file name to determine its type (local, env-specific, or default).
+ * Used to derive the correct lock file name and `mise lock` flags.
+ */
+export function getConfigType(configPath: string): MiseConfigType {
+  const filename = upath.basename(configPath);
+  const isLocal = filename.includes('.local.');
```
viceice:
```suggestion
  const isLocal = filename.endsWith('.local.toml');
```

lib/modules/manager/mise/readme.md
```diff
@@ -11,6 +11,24 @@ Renovate supports all standard mise configuration file patterns:
...
+When a lock file is present:
+
+- Dependencies will have their `lockedVersion` extracted from the lock file
+- Renovate can update lock files when dependencies change
+- Lock file maintenance is supported via the `lockFileMaintenance` option
+
+Renovate recognizes environment-specific lock files:
+
+- `mise.lock` - default lock file
+- `mise.local.lock` - local configuration lock file
```
viceice: shouldn't the local variants added to git ignore file? 🤔 

https://mise.jdx.dev/configuration/environments.html

lib/modules/manager/mise/lockfile.ts
```diff
@@ -0,0 +1,100 @@
...
+  // Patterns: mise.{env}.toml, mise.{env}.local.toml, .mise.{env}.toml, config.{env}.toml
+  const envMatch = regEx(
+    /^(?:\.?mise|config)\.([^.]+)(?:\.local)?\.toml$/,
```
viceice: use named capture group

lib/modules/manager/mise/extract.ts
```diff
@@ -44,31 +52,58 @@ export function extractPackageFile(
   }
 
   const deps: PackageDependency[] = [];
-  const tools = misefile.tools;
-
-  if (tools) {
```
viceice, line 53: why removal of this condition? can't that be undefined or null?

lib/modules/manager/mise/artifacts.spec.ts
zeitlinger: Removed that partial mock entirely. It was unnecessary here, so the test now uses the normal module behavior.

lib/modules/manager/mise/artifacts.spec.ts
zeitlinger: Removed the extra default setup. The test no longer needs that blanket mock state.

lib/modules/manager/mise/artifacts.spec.ts
zeitlinger: Switched this test to use real host-rule setup instead of mocking the module.

lib/modules/manager/mise/artifacts.ts
zeitlinger: Applied. This now uses `.filter(isString)`.

lib/modules/manager/mise/artifacts.ts
zeitlinger: Moved `updateLockedDependency()` into `./update-locked.ts` and re-exported it from the manager index.

lib/modules/manager/mise/extract.spec.ts
zeitlinger: Removed the redundant default `readLocalFile` setup from the spec.

lib/modules/manager/mise/lockfile.ts
zeitlinger: Applied. Local config detection now uses `endsWith('.local.toml')`.

lib/modules/manager/mise/readme.md
zeitlinger: Clarified the docs here. Local variants are usually paired with ignored local config files, so I called that out explicitly instead of presenting them like normal committed lockfiles.

lib/modules/manager/mise/lockfile.ts
zeitlinger: Applied. The env-matching regex now uses a named capture group.

lib/modules/manager/mise/extract.ts
zeitlinger: I checked this again and the guard is still present on current HEAD where it matters, so I didn’t make an extra code change for this thread.

lib/modules/manager/mise/artifacts.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (token) {
+    extraEnv.GITHUB_TOKEN = token;
+  }
+
+  const execOptions: ExecOptions = {
+    cwdFile: packageFileName,
+    extraEnv,
+    toolConstraints: [
+      {
+        toolName: 'mise',
```
viceice, line 67: needs to be added here
https://github.com/renovatebot/renovate/blob/53dcdba84db00e01a1edebeb0cebfc30d3105084/lib/util/exec/types.ts#L14

zeitlinger: Applied. I added `mise` to the Containerbase tool definitions and restored the `toolConstraints` usage in `artifacts.ts`, so the manager now uses the intended path instead of the temporary workaround.

lib/modules/manager/mise/artifacts.ts
```diff
@@ -0,0 +1,112 @@
...
+        constraint: config.constraints?.mise,
+      },
+    ],
+    docker: {},
+  };
+
+  try {
+    await exec(cmd, execOptions);
```
viceice, line 75: shouldn't we delete the lockfile when `config.isLockFileMaintenance==true`? we typically do it for other managers 🤔

lib/modules/manager/mise/artifacts.ts
```diff
@@ -0,0 +1,112 @@
...
+  const localFlag = isLocal ? ' --local' : '';
+
+  let cmd: string;
+  if (config.isLockFileMaintenance) {
+    cmd = `mise lock${localFlag}`;
+  } else {
+    const tools = updatedDeps
+      .map(({ depName }) => depName)
+      .filter(isNonEmptyStringAndNotWhitespace)
+      .map((depName) => quote(depName))
```
viceice:
```suggestion
      .map(quote)
```
simplify

viceice (commented):
otherwise LGTM

lib/modules/manager/mise/artifacts.ts
zeitlinger: I don't think we should delete first for mise, because that can actively degrade the lockfile's security metadata during maintenance.

mise.lock can already contain stronger recorded assertions for a tool/platform pair, and mise guarantees that an update will not switch that existing entry to a less secure verification method. But that guarantee only helps if the existing lockfile entry is still present. If we delete the file first, we remove the stronger assertion before mise gets a chance to preserve it, so regeneration can produce a weaker entry than the one the user had before.

So the problem isn't just "we lose metadata"; it's that delete-and-recreate can bypass mise's non-degradation behavior for existing lock entries. With `locked_verify_provenance=false`, users then directly trust the downgraded assertion set. With it enabled, installs re-verify what exists, but we still may have already thrown away the stronger recorded method.

viceice: but it ensures that all deps are updated to their latest allowed version?

viceice (commented):
conflicted

lib/modules/manager/mise/artifacts.ts
zeitlinger: Yes, `mise lock` still refreshes the lockfile for all tools and keeps versions within the ranges/prefixes from `mise.toml`, so we do not need a delete-and-recreate step to get the latest allowed versions. I want to preserve the existing file because that lets mise update in place and keep any stronger per-platform metadata/checksums that are already recorded instead of throwing them away first.

zeitlinger:
> conflicted

fixed

RahulGautamSingh (approved):

zeitlinger:
can this be merged?

RahulGautamSingh (approved):
@viceice Able to take a look.

lib/util/exec/types.ts
jamietanna, line 1: The changes unrelated to the addition of `mise` need to be reverted here

jamietanna (changes requested):
See changes noted for `lib/util/exec/types.ts`. Otherwise, LGTM

lib/util/exec/types.ts
zeitlinger: Reverted the unrelated cleanup in `lib/util/exec/types.ts`; that file’s diff is now scoped to adding `mise` to the tool definitions.

jamietanna: Not sure if the `VersioningName` needs to be there?

Also CI is failing

lib/util/exec/containerbase.ts
```diff
@@ -148,6 +148,11 @@ const allToolConfig: Record<ToolName, ToolConfig> = {
     packageName: 'containerbase/maven-prebuild',
     versioning: mavenVersioningId,
   },
+  mise: {
+    datasource: 'github-releases',
+    packageName: 'jdx/mise',
+    versioning: npmVersioningId,
```
mxmehl:
```suggestion
    versioning: semverVersioningId,
```

Would that fix the tests? This is also how it's configured in `pdm` which uses github-releases as datasource.

zeitlinger: I ended up with 'npm'

jamietanna (approved):

altendky:
thanks for working this through @zeitlinger ! :tada:

jamietanna:
Related: this does not work out-of-the-box due to mise's trust functionality: https://github.com/renovatebot/renovate/discussions/43562

PR #42627, opened 2026-04-14
feat(dashboard): print schedule(s) for awaiting PRs

setchy:
## Changes

Group updates in the Dependency Dashboard awaiting-schedule section under schedule and timezone headings. Each schedule is printed once, while the existing per-update checkboxes remain available beneath it.

Schedule reporting is explicitly opt-in via `dependencyDashboardReportSchedules: true`. This keeps existing dashboards unchanged and avoids increasing issue size unless users request the extra context.

The revived implementation also:

- updates the branch to current `main`
- preserves `dependencyDashboardCategory` as the outer grouping
- includes the effective timezone in schedule headings
- supports both cron and Renovate Later.js schedule syntax
- shares cron description formatting with PR configuration descriptions

## Context

This revisits the original need with the dashboard-size feedback from the earlier discussion incorporated. Grouping avoids repeating the same schedule against every checkbox. The proposed dashboard configuration object did not land, so this uses the existing flat `dependencyDashboard*` option model and defaults the new detail to off.

- [x] This does not close an issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

- [x] Yes, substantive assistance. GitHub Copilot was used to update the branch, resolve conflicts, modernize the implementation, and add tests and documentation.

## Documentation

## Testing

- [x] Updated unit tests

Validated locally with:

- `pnpm vitest lib/util/schedule.spec.ts lib/workers/repository/dependency-dashboard.spec.ts --run` (79 tests)
- `pnpm type-check`
- `pnpm test-schema`
- `pnpm biome`
- `pnpm oxlint`
- focused Prettier checks

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -1553,6 +1553,14 @@ This helps you identify dependencies that may need attention due to lack of main
 
 Set this to `false` if you prefer not to see abandoned packages in your dependency dashboard.
 
+## dependencyDashboardReportSchedules
+
+Controls whether schedule descriptions are reported in the dependency dashboard..
+
+When enabled (default), Renovate will display the schedule details for any PR awaiting schedule..
+
+Set this to `false` if you prefer not to see schedule descriptions in your dependency dashboard.
+
 ## dependencyDashboardTitle
 
 Configure this option if you prefer a different title for the Dependency Dashboard.
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -2107,6 +2107,13 @@ const options: Readonly<RenovateOptions>[] = [
     default: true,
   },
   {
+    name: 'dependencyDashboardReportSchedules',
+    description:
+      'Controls whether waiting schedule descriptions are reported in the dependency dashboard.',
+    type: 'boolean',
+    default: true,
+  },
+  {
     name: 'internalChecksAsSuccess',
     description:
       'Whether to consider passing internal checks such as `minimumReleaseAge` when determining branch status.',
--- a/lib/config/types.ts
+++ b/lib/config/types.ts
@@ -411,6 +411,7 @@ export interface RenovateConfig
   dependencyDashboardLabels?: string[];
   dependencyDashboardOSVVulnerabilitySummary?: 'none' | 'all' | 'unresolved';
   dependencyDashboardReportAbandonment?: boolean;
+  dependencyDashboardReportSchedules?: boolean;
   mode?: 'silent' | 'full';
   packageFile?: string;
   packageRules?: PackageRule[];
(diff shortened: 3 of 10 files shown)


viceice (commented):
we need an option to opt-in or opt-out because the dashboard can already be very huge and truncated

eg: https://codeberg.org/forgejo/forgejo/issues/2779

setchy:
> we need an option to opt-in or opt-out because the dashboard can already be very huge and truncated
> 
> eg: https://codeberg.org/forgejo/forgejo/issues/2779

😅

setchy:
Added a opt-out config called `dependencyDashboardReportSchedules` - feedback welcome

viceice:
wonder if we should have a dashboard options object/ array to opt in into all those features, so we can unify them in a single config option. 🤔

@jamietanna @secustor wdyt?

jamietanna:
Yep, I was thinking the same - would we use i.e.

```json
{
  "dependencyDashboard": {
    "enabled": true
  }
}
```

and then have a migration from existing options? Seems reasonable to me

setchy:
> wonder if we should have a dashboard options object/ array to opt in into all those features, so we can unify them in a single config option. 🤔
> 
> @jamietanna @secustor wdyt?

I was thinking similar... a broader cleanup / restructure of the `dependencyDashboard*` config options

secustor:
Definitely in favour of moving this to an object, but I'm worried we are starting to overload the dashboard with information.
Especially as this is on by default, together with other sections which are turned by default it becomes hard to read. 

`https://github.com/backstage/backstage/issues/22682` is an example, even if it is a especial big project

setchy:
I personally find the Dependency Dashboards one of the most valuable features of Renovate - it's a super useful single pane of glass / control plan for a repositories dependency state.

I see it's utility being even greater if we can surface further insights in a user-configurable way (opt-in or opt-out).

For instance:
- _"which status checks are pending?"_
- _"when is the next schedule we're waiting for?"_
- _"what conditions are contributing to PRs being rate-limited?"_
- _"for deprecations/replacements, which package paths/files are these from?"_
- _"for the detected dependencies tree, showing some iconography / linkage back to the abandoned and vulnerable lists"_

There is also the known work to sync/align the Mend-hosted Dependency Dashboard since that's a different implementation that has drifted in recent times...

viceice:
I'm ok if most of them are opt-in

jamietanna:
Moving to draft so we can raise a prefactor PR first so we can convert the `dependencyDashboard` to an object

setchy:
@jamietanna - I refreshed this branch with the help of Claude this morning. 

It's in a ready for review state, but wanted to check if the dashboard config object refactor is still a prereq or something we can do independently?

PR #42672, opened 2026-04-15
feat(gerrit): Add gitUrl support

spmiller:
## Changes

This change adds support for the existing `gitUrl` parameter to the `gerrit` platform. 

The existing behaviour (when `gitUrl` is not specified) is left unchanged, and is mapped to the `default` and `endpoint` values of `gitUrl`.

## Context

## AI assistance disclosure

- [X] Yes — other (please describe):
I asked Claude which API provided the download schemes in Gerrit. No AI was used for the coding.

I added Gerrit as a supported platform for the parameter `gitUrl` in `config/options/index.ts`

My Gerrit server is not publicly available, but with my changes Renovate was able to clone the repo over SSH.


lib/modules/platform/gerrit/utils.ts
```diff
@@ -32,34 +32,61 @@ export const REQUEST_DETAILS_FOR_PRS: GerritRequestDetail[] = [
...
+  httpUrl.username = opts.username;
+  httpUrl.password = opts.password;
+  httpUrl.pathname = joinUrlParts(
+    httpUrl.pathname,
     'a',
     encodeURIComponent(repository),
   );
+
+  const sshUrl = new URL(
+    `ssh://${endpointUrl.host}:29418${joinUrlParts(endpointUrl.pathname, repository)}`,
```
felipecrs: Can you please extract the port number to a constant?

lib/modules/platform/gerrit/utils.ts
```diff
@@ -32,34 +32,61 @@ export const REQUEST_DETAILS_FOR_PRS: GerritRequestDetail[] = [
...
   if (!(opts.username && opts.password)) {
     throw new Error(
       'Init: You must configure a Gerrit Server username/password',
     );
   }
-  url.username = opts.username;
-  url.password = opts.password;
-  url.pathname = joinUrlParts(
-    url.pathname,
+  const httpUrl = new URL(endpointUrl.toString());
```
felipecrs: Nitpick, but maybe using `clone` from `util/clone.ts` would be nicer.

lib/modules/platform/gerrit/utils.ts
```diff
@@ -32,34 +32,61 @@ export const REQUEST_DETAILS_FOR_PRS: GerritRequestDetail[] = [
...
+  httpUrl.username = opts.username;
+  httpUrl.password = opts.password;
+  httpUrl.pathname = joinUrlParts(
+    httpUrl.pathname,
     'a',
     encodeURIComponent(repository),
   );
+
+  const sshUrl = new URL(
```
felipecrs: You should probably use `parseUrl` here too.

felipecrs (changes requested):
Otherwise, LGTM.

felipecrs:
@spmiller, this reminded me of a very old discussion:

- https://github.com/gitkraken/vscode-gitlens/issues/1964

In that discussion, the Gerrit URL would be something like `<host>/gerrit/<projectPath>`, as well as the HTTP clone URL, while the SSH clone URL would be something like `<host>/<projectPath>`, without the path prefix.

Maybe you can extract the prefix from the configured platform URL and "subtract" it from the generated SSH URL?

spmiller:
> @spmiller, this reminded me of a very old discussion:
> 
> Maybe you can extract the prefix from the configured platform URL and "subtract" it from the generated SSH URL?

Hey @felipecrs, this old discussion is gold! Having some actual config examples makes it much easier to solve the problem. I think the behaviour in my PR would not work for their server.

It got me thinking, maybe my original approach (query the server to get the download schemes) wasn't so bad after all? It was an extra API call (`config/server/info`) but I could Promise.all them quite easily to get it in parallel. We wouldn't have to have extra logic to extract the prefix.

The trade-off is more code in the client/types files. Perhaps it's worth it?

This is the original commit I made in case you want to compare:
https://github.com/spmiller/renovate/commit/bb187fa99002b2f4800279c3ef4847578e8642d2#diff-9e28b3970977a05aec269cd0b4fb67427dc51bdb6a1e10281f1dc214c69c56f8

marquisesenior383-ctrl:
No stop whatever you doing now 5019

felipecrs:
@spmiller, you only need to make these changes:

```diff
diff --git a/lib/modules/platform/gerrit/utils.spec.ts b/lib/modules/platform/gerrit/utils.spec.ts
index a6f7b7a72..7515b781f 100644
--- a/lib/modules/platform/gerrit/utils.spec.ts
+++ b/lib/modules/platform/gerrit/utils.spec.ts
@@ -100,7 +100,7 @@ describe('modules/platform/gerrit/utils', () => {
           'https://gerrit.example.com/context',
           'ssh',
         );
-        expect(repoUrl).toBe('ssh://gerrit.example.com:29418/context/web/apps');
+        expect(repoUrl).toBe('ssh://gerrit.example.com:29418/web/apps');
       });
     });
   });
diff --git a/lib/modules/platform/gerrit/utils.ts b/lib/modules/platform/gerrit/utils.ts
index 5f24e3635..9d7ffde83 100644
--- a/lib/modules/platform/gerrit/utils.ts
+++ b/lib/modules/platform/gerrit/utils.ts
@@ -62,7 +62,7 @@ export function getGerritRepoUrl(
   );
 
   const sshUrl = new URL(
-    `ssh://${endpointUrl.host}:29418${joinUrlParts(endpointUrl.pathname, repository)}`,
+    `ssh://${endpointUrl.host}:29418/${repository}`,
   );
 
   const url = pickUrlFromGitUrl(gitUrl, sshUrl, httpUrl);

```

felipecrs (approved):
LGTM.

lib/modules/platform/gerrit/utils.ts
```diff
@@ -32,34 +35,64 @@ export const REQUEST_DETAILS_FOR_PRS: GerritRequestDetail[] = [
...
+  );
+  if (!sshUrl) {
+    throw new Error(CONFIG_GIT_URL_UNAVAILABLE);
+  }
+
+  const url = pickUrlFromGitUrl(gitUrl, sshUrl, httpUrl);
   logger.trace(
     { url: url.toString() },
     'using URL based on configured endpoint',
   );
```
felipecrs: Coverage is now failing because of the change I asked (to use `parseUrl`). Sorry about that.

https://app.codecov.io/gh/renovatebot/renovate/pull/42672?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=renovatebot#598202c3fbcb3052e676a66671afb2ec-R71

Feel free to change it back to `new URL()` for now, then let's see what the maintainers think.

spmiller: > Feel free to change it back to `new URL()` for now, then let's see what the maintainers think.

Thanks, I think that's the pragmatic option. I rewrote the method to allow testing the coverage and it looked dreadful!

spmiller:
> @spmiller, you only need to make these changes:

Okay, I've made those changes and addressed your other review comments. Thanks for your help so far!

lib/modules/platform/gerrit/utils.ts
```diff
@@ -32,34 +35,60 @@ export const REQUEST_DETAILS_FOR_PRS: GerritRequestDetail[] = [
...
+function pickUrlFromGitUrl(
+  gitUrl: GitUrlOption | undefined,
+  sshUrl: URL,
+  httpUrl: URL,
+): string {
+  switch (gitUrl) {
+    case 'ssh':
+      return sshUrl.toString();
+    case 'endpoint':
+      return httpUrl.toString();
```
viceice:
```suggestion
```

let fall through when same

lib/modules/platform/gerrit/utils.ts
```diff
@@ -32,34 +35,60 @@ export const REQUEST_DETAILS_FOR_PRS: GerritRequestDetail[] = [
...
   if (!(opts.username && opts.password)) {
     throw new Error(
       'Init: You must configure a Gerrit Server username/password',
     );
   }
-  url.username = opts.username;
-  url.password = opts.password;
-  url.pathname = joinUrlParts(
-    url.pathname,
+  const httpUrl = clone(endpointUrl);
```
viceice: why cloning? the `endpointUrl` is only used in this function, so use it here again?

better do the switch case here and don't compute multiple urls to later throw the other away. only compute ssh when `gitUrl=ssh`, use current otherwise. will make the code much simpler

spmiller:
Hi @viceice, thanks for the review. I have updated the code based on your feedback and it has simplified things a lot.

I pushed the URL creation into separate methods to keep `getGerritRepoUrl` small but I can inline them if that fits the style of this repo better.

Thanks for your help!

lib/modules/platform/gerrit/utils.ts
```diff
@@ -32,33 +34,51 @@ export const REQUEST_DETAILS_FOR_PRS: GerritRequestDetail[] = [
...
+      : createHttpUrl(endpointUrl, endpoint, repository);
+  logger.trace({ url }, 'using URL based on configured endpoint');
+
+  return url;
+}
+
+function createSshUrl(url: URL, repository: string): string {
+  return new URL(
+    `ssh://${url.host}:${DEFAULT_SSH_PORT}/${repository}`,
+  ).toString();
```
viceice:
```suggestion
  return `ssh://${url.host}:${DEFAULT_SSH_PORT}/${repository}`;
```
why not simply inline?

spmiller: Creating a new URL added some validation that the host, port and repository combination would create a valid URL. If you don't think that's required I can commit your suggestion.

Thanks!

viceice: not required here

spmiller: Thanks @viceice, I've updated the PR.

viceice (approved):

RahulGautamSingh (commented):
LGTM

felipecrs:
@spmiller, I believe you need to mark all conversations as **resolved** for the auto-merge to kick in.

viceice (approved):
yes, I don't have the time to search all open conversations 😅

spmiller:
> I believe you need to mark all conversations as **resolved** for the auto-merge to kick in.
> yes, I don't have the time to search all open conversations 😅

Thanks guys! I am new to this repo, and I really appreciate how helpful you have been in getting this in.

PR #42689, opened 2026-04-16
feat(presets/best-practices): pin Actions to SemVer

jamietanna:
## Changes

A "best practice" that we generally recommend is to pin your GitHub
Actions to SemVer.

We should promote this up to `config:best-practices`, where all users
can benefit from the additional visibility we receive for the full
SemVer tag in comments.

I'll wait for both maintainers to review this, given it's `config:best-practices`, before we merge.

I don't believe this should need to wait until next major.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/config/presets/internal/config.preset.ts
+++ b/lib/config/presets/internal/config.preset.ts
@@ -8,6 +8,8 @@ export const presets: Record<string, Preset> = {
       'config:recommended',
       'docker:pinDigests',
       'helpers:pinGitHubActionDigests',
+      // as well as helpers:pinGitHubActionDigests, so it can be ignored without removing the regular pinning
+      'helpers:pinGitHubActionDigestsToSemver',
       ':configMigration',
       ':pinDevDependencies',
       'abandonments:recommended',


viceice (dismissed):

lib/config/presets/internal/config.preset.ts
```diff
@@ -8,6 +8,8 @@ export const presets: Record<string, Preset> = {
       'config:recommended',
       'docker:pinDigests',
       'helpers:pinGitHubActionDigests',
+      // as well as helpers:pinGitHubActionDigests, so it can be ignored without removing the regular pinning
+      'helpers:pinGitHubActionDigestsToSemver',
```
secustor, line 12: I remember in the past there have been Discussions which reported when both of them are enabled. 
Are these resolved?

jamietanna: ~~What do you mean, sorry?~~

I'll take a look and confirm!

RahulGautamSingh (commented):
Some review notes inline.

jamietanna:
We shouldn't do this until we can make sure that PRs are correctly `pinDigest`'d on first merge (i.e. to avoid https://github.com/ecosyste-ms/ecosystems-go/pull/40 and https://github.com/ecosyste-ms/ecosystems-go/pull/41 both being raised)

jamietanna:
Might be we'd want to use `rangeStrategy=pin` here?

secustor:
> Might be we'd want to use `rangeStrategy=pin` here?

I have been not sure how this would effect, the regex versioning. 

So it seems this will not work: 

```
rangeStrategy=pin on renovatebot/renovate#42689 is not viable — with the preset's regex versioning isSingleVersion('v7') is true, so lookup's pin branch never fires and rangeStrategy resets to replace; the real duplicate comes from the digest branch guarded by the manager-only digestOneAndOnly flag, though the churn is one-time and self-healing.
```

<details>

<summary> Full Opus 5 output </summary>

```
I traced the actual code path and verified the key predicates against the installed Renovate (44.4.6) plus current main sources.

Verdict: not viable — rangeStrategy=pin is a no-op here

Why it can't fire. helpers:pinGitHubActionDigestsToSemver sets versioning: regex:^v?(?<major>\d+)(\.(?<minor>\d+)\.(?<patch>\d+))?$. The regex versioner extends GenericVersioningApi, where isSingleVersion === isValid === isVersion, and it declares supportsRanges = false. Verified:

regex v7   isValid=true isVersion=true isSingleVersion=true

The pin update in lib/workers/repository/process/lookup/index.ts:529 is gated on rangeStrategy === 'pin' && !versioningApi.isSingleVersion(compareValue). Since v7 is a single version, the branch never runs — and index.ts:545 then silently resets rangeStrategy back to 'replace'. Identical output to today.

Even if it fired, it wouldn't fix the cited case. The pin-suppression at index.ts:818-822 only guards the pinDigest branch (repo not yet digest-pinned). In ecosystems-go the repo already had digests, so the duplicate came from the digest branch at index.ts:809-817, which is guarded by digestOneAndOnly — a manager/extract-level flag on PackageDependency (only gomod and bazel set it), not something a preset can set.

What actually causes the double PR

- extractVersion: ^(?<version>v?\d+\.\d+\.\d+)$ drops the floating v7 tag from the release list.
- Generic matches() is exact triple equality: matches('v7.0.0','v7')=true, matches('v7.0.1','v7')=false. So currentVersion resolves to v7.0.0 and the move to v7.0.1 is classified a patch update → PR #41 (comment rewrite only).
- Independently, currentDigest is set, so a digest update is emitted with newValue='v7' and the new sha → PR #40.
- Both survive the strip filter at index.ts:933-947 and converge on the same sha.

The mitigation worth raising on the PR

The duplication is one-time and self-healing, not recurring. Once the comment reads # v7.0.1, the next run's digest update has newValue='v7.0.1', whose resolved digest equals currentDigest, so it's stripped at index.ts:945-946 and only the version PR remains. The cost is bounded by "actions currently pinned with a floating-major comment", paid once at rollout — but that is exactly every existing config:best-practices user, since they all already have helpers:pinGitHubActionDigests. Fresh onboarding also gets two (a pinDigest PR, then a patch PR).

I also checked the only shape where the pin idea could mechanically work — a versioning that treats v7 as a range. npm qualifies (isSingleVersion('v7')=false), but it's a dead end: getNewValue({currentValue:'v7', newVersion:'v7.0.1', rangeStrategy:'pin'}) returns 7, and isValid('main')=false breaks non-semver actions.

So the real options are: land a core change (extend the pin-suppression to the currentDigest branch, or give github-actions digestOneAndOnly), or accept the one-time churn and document it — which is a maintainer judgement call, not something the preset can paper over.

result: rangeStrategy=pin on renovatebot/renovate#42689 is not viable — with the preset's regex versioning isSingleVersion('v7') is true, so lookup's pin branch never fires and rangeStrategy resets to replace; the real duplicate comes from the digest branch guarded by the manager-only digestOneAndOnly flag, though the churn is one-time and self-healing.
```


</details>

Issue #15696, opened 2022-05-23
dotnet lockfiles aren't properly updated when using central package versioning


mdarocha:
### What would you like Renovate to be able to do?

In https://github.com/renovatebot/renovate/pull/14312 a fix was added, to update all referenced project's `packages.lock.json` files when updating a package in a project.

Hovewer, it seems this does not take into account a situation when a central versioning scheme is used (either [CentralPackageVersions](https://github.com/microsoft/MSBuildSdks/tree/main/src/CentralPackageVersions) or [NuGet central versioning](https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions)). In such a case, renovate only updates the `Packages.props`/`Directory.Packages.props` file containing the package version, and the same situation as described by https://github.com/renovatebot/renovate/issues/8010 happens - CI fails for the renovate MR due to `RestoreLockedMode` being set.

When bumping a package version in such a situation, lockfiles of all projects that use the package should be updated, as well as lockfiles of all projects that reference the projects.

Example reproduction: https://github.com/mdarocha/renovate-packages-lock-bug/pull/3

### If you have any ideas on how this should be implemented, please tell us here.

Seems like the check here: https://github.com/renovatebot/renovate/blob/5afed40ba318852c2d36e355fe1fcbfc9caf5fff/lib/modules/manager/nuget/artifacts.ts#L120 should be extended and properly implemented.

### Is this a feature you are interested in implementing yourself?

Maybe

viceice:
forked to https://github.com/renovate-reproductions/15696-renovate-packages-lock-bug

viceice:
Please add a readme to the project and describe current and expectecd behavior

viceice:
//cc @mattvaughan

mdarocha:
> Please add a readme to the project and describe current and expectecd behavior

Done

viceice:
PR open 🎉

Issue #8010, opened 2020-12-14
NuGet lock file not updated for referencing projects


lazyboy1:
**What Renovate type, platform and version are you using?**

Self hosted v24.2.5

**Describe the bug**

When an update found for a dependency in a project and a lock file exists, Renovate will run `dotnet restore` for that project to update the lock file. It should run `dotnet restore` for projects referencing the updated project as well, since their lock file also needs to be updated.

**Relevant debug logs**

Important part is that it's running the command `dotnet restore src/My.Project/My.Project.csproj`. Later in the PR pipeline `dotnet restore --locked-mode` is failing for a tests project that's referencing `My.Project`.

```
DEBUG: Starting search at index 290 (repository=**repo-name**, packageFile=src/My.Project/My.Project.csproj, branch=renovate/identitymodel-4.x)
       "depName": "IdentityModel"
DEBUG: Found match at index 290 (repository=**repo-name**, packageFile=src/My.Project/My.Project.csproj, branch=renovate/identitymodel-4.x)
       "depName": "IdentityModel"
DEBUG: found NuGet.config (repository=**repo-name**, branch=renovate/identitymodel-4.x)
       "nuGetConfigPath": "/tmp/renovate/repos/bitbucket/**repo-name**/nuget.config"
DEBUG: clearing registry URLs (repository=**repo-name**, branch=renovate/identitymodel-4.x)
DEBUG: adding registry URL (repository=**repo-name**, branch=renovate/identitymodel-4.x)
       "registryUrl": "https://api.nuget.org/v3/index.json"
DEBUG: adding registry URL (repository=**repo-name**, branch=renovate/identitymodel-4.x)
       "registryUrl": "**private registry url**"
DEBUG: Contents updated (repository=**repo-name**, packageFile=src/My.Project/My.Project.csproj, branch=renovate/identitymodel-4.x)
       "depName": "IdentityModel"
DEBUG: nuget.updateArtifacts(src/My.Project/My.Project.csproj) (repository=**repo-name**, branch=renovate/identitymodel-4.x)
DEBUG: found NuGet.config (repository=**repo-name**, branch=renovate/identitymodel-4.x)
       "nuGetConfigPath": "/tmp/renovate/repos/bitbucket/**repo-name**/nuget.config"
DEBUG: clearing registry URLs (repository=**repo-name**, branch=renovate/identitymodel-4.x)
DEBUG: adding registry URL (repository=**repo-name**, branch=renovate/identitymodel-4.x)
[...]

PR #42735, opened 2026-04-19
feat(nuget): update lock files when `Directory.Build.props` is changed

jeffpardy:
## Changes

`Directory.Build.props` is a standard MSBuild import file that can contain `<PackageReference>` elements inherited by all projects in the same directory and below. When Renovate updates packages referenced in this file, lock file (`packages.lock.json`) regeneration was skipped because `updateArtifacts` only handled `.csproj`/`.vbproj`/`.fsproj`, `Directory.Packages.props`/`Packages.props`, and `global.json`.

This extends the existing logic to treat `Directory.Build.props` the same way as central management files — all project files under its parent directory have their lock files regenerated via `dotnet restore`.

In my case, I use `Directory.Build.props` to set common packages across a monorepo. Renovate PRs that update packages in this file always fail CI because the dependent project lock files are stale. This requires manually pulling down the branch and running `dotnet restore` to regenerate the lock files — which defeats the purpose of automated dependency updates.

Changes:
- `package-tree.ts`: Added `DIRECTORY_BUILD_PROPS` constant, added it to the `parentDir` resolution, renamed `isCentralManagement` parameter to `isPropsFile` for clarity since it now covers both central management and `Directory.Build.props`
- `artifacts.ts`: Added `isDirectoryBuildProps` detection, updated the guard clause to allow `Directory.Build.props` through, passed the combined flag to `getDependentPackageFiles`
- Added 4 new test cases covering root-level and nested `Directory.Build.props` for both dependency resolution and lock file updates
- Added new test fixture directory

## Context

Related: #15696 (the analogous fix for `Directory.Packages.props`), #8010 (original lock file update issue)

## AI assistance disclosure

### DIFF
--- a/lib/modules/manager/nuget/artifacts.spec.ts
+++ b/lib/modules/manager/nuget/artifacts.spec.ts
@@ -191,6 +191,80 @@ describe('modules/manager/nuget/artifacts', () => {
     expect(execSnapshots).toBeEmptyArray();
   });
 
+  it('updates lock file for Directory.Build.props', async () => {
+    const execSnapshots = mockExecAll();
+    scm.getFileList.mockReset();
+    scm.getFileList.mockResolvedValueOnce(['project.csproj']);
+    fs.getSiblingFileName
+      .mockReturnValueOnce('packages.lock.json')
+      .mockReturnValueOnce('packages.lock.json');
+    git.getFiles.mockResolvedValueOnce({
+      'packages.lock.json': 'Current packages.lock.json',
+    });
+    fs.getLocalFiles.mockResolvedValueOnce({
+      'packages.lock.json': 'New packages.lock.json',
+    });
+
+    expect(
+      await nuget.updateArtifacts({
+        packageFileName: 'Directory.Build.props',
+        updatedDeps: [{ depName: 'dep' }],
+        newPackageFileContent: '{}',
+        config,
+      }),
+    ).toEqual([
+      {
+        file: {
+          contents: 'New packages.lock.json',
+          path: 'packages.lock.json',
+          type: 'addition',
+        },
+      },
+    ]);
+    expect(execSnapshots).toMatchObject([
+      {
+        cmd: 'dotnet restore project.csproj --force-evaluate --configfile /tmp/renovate/cache/__renovate-private-cache/nuget/nuget.config',
+      },
+    ]);
+  });
+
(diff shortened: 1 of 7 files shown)


lib/modules/manager/nuget/artifacts.spec.ts
```diff
@@ -191,6 +191,80 @@ describe('modules/manager/nuget/artifacts', () => {
     expect(execSnapshots).toBeEmptyArray();
   });
 
+  it('updates lock file for Directory.Build.props', async () => {
+    const execSnapshots = mockExecAll();
+    scm.getFileList.mockReset();
```
viceice, line 196:
```suggestion
```
shouldn't needed

lib/modules/manager/nuget/artifacts.spec.ts
```diff
@@ -191,6 +191,80 @@ describe('modules/manager/nuget/artifacts', () => {
...
+    expect(execSnapshots).toMatchObject([
+      {
+        cmd: 'dotnet restore project.csproj --force-evaluate --configfile /tmp/renovate/cache/__renovate-private-cache/nuget/nuget.config',
+      },
+    ]);
+  });
+
+  it('updates lock file for nested Directory.Build.props', async () => {
+    const execSnapshots = mockExecAll();
+    scm.getFileList.mockReset();
```
viceice, line 233: same

lib/modules/manager/nuget/artifacts.spec.ts
jeffpardy: The `mockReset()` is needed here. The `beforeEach` queues `scm.getFileList.mockResolvedValueOnce([])` which returns an empty project file list. Without the reset, `getDependentPackageFiles` gets `[]` from `getFileList`, finds no dependent projects, and the function returns `null`.

This is different from the existing `.csproj` tests where the empty list is fine — the `.csproj` file itself becomes a leaf node in the graph. For `.props` files, the file gets removed from deps (line 85-87 in `package-tree.ts`), so we need actual project files in the list to have any dependents. There are no existing artifact tests for `Directory.Packages.props` or `global.json` that go through this path, so this pattern hasn't come up before.

I confirmed by removing the reset — the nested test fails with `expected null to deeply equal [...]`.

viceice (approved):

PR #42737, opened 2026-04-19
fix: Revert "build(deps): update dependency simple-git to v3.36.0 (main)"

viceice:
Reverts renovatebot/renovate#42736

breaks renovate

### DIFF
--- a/package.json
+++ b/package.json
@@ -257,7 +257,7 @@
     "semver-stable": "3.0.0",
     "semver-utils": "1.1.4",
     "shlex": "3.0.0",
-    "simple-git": "3.36.0",
+    "simple-git": "3.35.2",
     "slugify": "1.6.9",
     "source-map-support": "0.5.21",
     "strip-json-comments": "5.0.3",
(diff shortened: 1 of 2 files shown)


valentjn:
So https://github.com/renovatebot/renovate/pull/42736#issuecomment-4275487537 was right? Why downvoting that?

secustor (approved):

jamietanna:
@valentjn you're likely seeing a :-1: for your comment:

> I'd create an issue, but this is forbidden here 
> (and I won't open a discussion **because this more feels like posting to `/dev/null` to me**)

Emphasis mine. This is an unhelpful comment

We (the Renovate maintainers) try and do what we can to respond to and action reports from the community as best as we can, but we are a significantly small team for the usage, and comments like this aren't always taken with the best intentions behind them

viceice:
> So https://github.com/renovatebot/renovate/pull/42736#issuecomment-4275487537 was right? Why downvoting that?

because of the dev null message, which is inappropriate

jamietanna (commented):
LGTM

PR #42744, opened 2026-04-19
feat(github-actions): add support for jakebailey/pyright-action version option

trim21:


lib/modules/manager/github-actions/community.ts
```diff
@@ -291,6 +291,32 @@ const ZizmorcoreZizmorAction = z
     };
   });
 
+const SetupPyright = z
+  .object({
+    uses: matchAction('jakebailey/pyright-action'),
+    with: z.object({ version: z.string().optional() }),
```
viceice, line 297: there's also `pylance-version` and `python-version` inputs 🤔

viceice (commented):
otherwise LGTM

lib/modules/manager/github-actions/community.ts
trim21: the pylance is released on vscode market and there is no way to support it unless we add a new source.

and the `python-version` here is actually a option of pyright's min support python version, I'm not very interested in adding it because pyright can infer it from `requires-python`

lib/modules/manager/github-actions/community.ts
```diff
@@ -291,6 +291,32 @@ const ZizmorcoreZizmorAction = z
...
+  })
+  .transform(({ with: val }): PackageDependency => {
+    let skipStage: StageName | undefined;
+    let skipReason: SkipReason | undefined;
+
+    if (!val.version || val.version === 'PATH') {
+      skipStage = 'extract';
+      skipReason = 'unspecified-version';
+    }
```
viceice, line 306: we should extract a helper function in a follow up PR to not write so much duplicate code

lib/modules/manager/github-actions/community.ts
```diff
@@ -291,6 +291,32 @@ const ZizmorcoreZizmorAction = z
...
+
+    return {
+      datasource: NpmDatasource.id,
+      depName: 'pyright',
+      versioning: npmVersioning.id,
+      packageName: 'pyright',
+      ...(skipStage && { skipStage }),
+      ...(skipReason && { skipReason }),
+      currentValue: val.version,
+      depType: 'uses-with',
```
viceice, line 316: same

viceice (approved):

PR #42769, opened 2026-04-20
fix(versioning/github-actions): correctly handle `major.minor` and `major`

jamietanna:
## Changes

As part of the final set of changes needed to support Immutable Releases
with GitHub Actions and the `github-actions` versioning, we need to
modify how our versioning works to handle correctly surfacing the
shortest version available.

To make this more resilient, and help us test-drive the change, we'll
also add an integration test that:

- extracts from a GitHub Actions workflow
- performs a lookup (with mocked data)
- determines the available updates that would be proposed

This provides more confidence in the change itself, as well as
validates the intended changes.

## Context

## AI assistance disclosure

The public repository: https://github.com/JamieTanna-Mend-testing/setup-uv/issues/3

### DIFF
--- a/lib/modules/manager/github-actions/integration.spec.ts
+++ b/lib/modules/manager/github-actions/integration.spec.ts
@@ -1,0 +1,947 @@
+import { codeBlock } from 'common-tags';
+import * as httpMock from '~test/http-mock.ts';
+import { partial } from '~test/util.ts';
+import { getConfig } from '../../../config/defaults.ts';
+import * as hostRules from '../../../util/host-rules.ts';
+import { Result } from '../../../util/result.ts';
+import * as lookup from '../../../workers/repository/process/lookup/index.ts';
+import type { LookupUpdateConfig } from '../../../workers/repository/process/lookup/types.ts';
+import { GithubTagsDatasource } from '../../datasource/github-tags/index.ts';
+import { id as githubActionsVersioningId } from '../../versioning/github-actions/index.ts';
+import type { PackageDependency } from '../types.ts';
+import { extractPackageFile } from './index.ts';
+
+describe('modules/manager/github-actions/integration', () => {
+  describe('when using versioning=github-actions', () => {
+    const getGithubTags = vi.spyOn(
+      GithubTagsDatasource.prototype,
+      'getReleases',
+    );
+    const getGithubDigest = vi.spyOn(
+      GithubTagsDatasource.prototype,
+      'getDigest',
+    );
+
+    let baseConfig: LookupUpdateConfig;
+
+    beforeEach(() => {
+      // TODO: fix types #22198
+      baseConfig = partial<LookupUpdateConfig>(getConfig() as never);
+      baseConfig.manager = 'github-actions';
+    });
+
+    afterEach(() => {
+      httpMock.clear(false);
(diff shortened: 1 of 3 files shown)


lib/modules/versioning/github-actions/index.ts
```diff
@@ -228,6 +228,14 @@ function getNewValue({
 
   const newParsed = parseVersion(newVersion);
   if (!newParsed) {
+    // If a major (i.e. `v7`), and the proposed update is a minor i.e. (`v7.6`), return the existing major version instead of updating to the new minor, as the major should have been re-tagged, too
+    if (isUndefined(range.minor)) {
```
viceice: should we check if `v7` exists? so maybe they stopped publishing it?

lib/modules/versioning/github-actions/index.ts
```diff
@@ -272,21 +292,36 @@ function getShortestMatchingVersion(
...
-    `${prefix}${newParsed.major}.${newParsed.minor}`,
-    `${prefix}${newParsed.major}.${newParsed.minor}.${newParsed.patch}`,
-    `${prefix}${newParsed.toString()}`,
-  ];
-
-  for (const option of options) {
-    if (allVersions.has(option)) {
-      return option;
+  const { major, minor, patch } = newParsed;
+  const versions = new Set(allVersions);
```
viceice, line 303: don't duplicate, it's already a set (perf)

lib/modules/versioning/github-actions/index.ts
jamietanna: Debug log? Or in that case, return the minor?

viceice: log and return the minor. same should be done if we're at `1.2` but only `1.3.0` to update to

lib/modules/versioning/github-actions/index.ts
viceice: forgot to push?

viceice (commented):
see comments

lib/modules/versioning/github-actions/index.ts
jamietanna: Yep, was going to once I'd updated the other :)

lib/modules/versioning/github-actions/index.ts
jamietanna: I've changed it but not added the log - thoughts?

viceice (dismissed):

lib/modules/manager/github-actions/integration.spec.ts
```diff
@@ -0,0 +1,947 @@
...
+    let baseConfig: LookupUpdateConfig;
+
+    beforeEach(() => {
+      // TODO: fix types #22198
+      baseConfig = partial<LookupUpdateConfig>(getConfig() as never);
+      baseConfig.manager = 'github-actions';
+    });
+
+    afterEach(() => {
+      httpMock.clear(false);
```
viceice: is that really needed I don't see host rules or http mock usage 🤔

jamietanna: Nope!

viceice (approved):

PR #42783, opened 2026-04-21
feat(mise): automagically support all tools through Mise Registry

jamietanna:
## Changes

As noted in #42250, we should try and auto-support tools from Mise where
possible.

Similar to other areas of the codebase, we should use the "static data"
pattern, and periodically sync the `mise registry --json` output with
`lib/data/mise-registry.json`, and then auto-detect the tools that we
support through the known backends the tools support.

This takes lower precedence than existing in-code tool definitions, so
we can still provide an "override" if we can support lookups or version
extraction more appropriately.

Because we now have many more tools we (mostly) support out-of-the-box,
we need to provide an improved documentation page, and:

- only link to tools we have an explicit tools URL for
- note where a backend may not be supported

To automate the updating of this data, we:

- make sure we have `mise` in GitHub Actions, before we run
- don't pin the `mise` version to make sure that we're always seeing the
  latest supported tools
- make sure the commits generated are now `feat(...)`, as they introduce
  functionality changes (in the case new tool(s) are supported)

> [!NOTE]
> This improves support from ~170 tool short names to just over 1000.

## Context

- [x] This closes an existing Issue: Closes #42250

## AI assistance disclosure

The public repository: https://github.com/JamieTanna-Mend-testing/mise/pull/2 wasn't supported before

### DIFF
--- a/.github/workflows/update-data.yml
+++ b/.github/workflows/update-data.yml
@@ -35,6 +35,14 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Install mise
+        uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
+        with:
+          # don't run `mise install`, but do install a version of `mise` itself
+          install: false
+          # explicitly always use the latest tool version
+          version: latest
+
       - name: Update static data
         run: pnpm run update-static-data
 
@@ -46,7 +54,7 @@ jobs:
         with:
           author: 'Renovate Bot <renovate@whitesourcesoftware.com>'
           branch: 'chore/update-static-data'
-          commit-message: 'fix(data): automatic update of static data'
+          commit-message: 'feat(data): automatic update of static data'
           committer: 'Renovate Bot <renovate@whitesourcesoftware.com>'
-          title: 'fix(data): automatic update of static data'
+          title: 'feat(data): automatic update of static data'
           assignees: viceice,secustor,jamietanna
--- a/lib/data/mise-registry.json
+++ b/lib/data/mise-registry.json
@@ -1,1736 +1,3797 @@
 {
-  "1password": ["vfox:mise-plugins/vfox-1password", "aqua:1password/cli"],
-  "1password-cli": ["vfox:mise-plugins/vfox-1password", "aqua:1password/cli"],
-  "aapt2": ["vfox:mise-plugins/vfox-aapt2"],
-  "acli": ["aqua:atlassian.com/acli"],
-  "act": ["aqua:nektos/act", "asdf:gr1m0h/asdf-act"],
-  "action-validator": [
-    "aqua:mpalmer/action-validator",
-    "asdf:mpalmer/action-validator",
-    "cargo:action-validator"
-  ],
-  "actionlint": [
-    "aqua:rhysd/actionlint",
-    "asdf:crazy-matt/asdf-actionlint",
-    "go:github.com/rhysd/actionlint/cmd/actionlint"
-  ],
-  "adr-tools": [
-    "aqua:npryce/adr-tools",
-    "asdf:https://gitlab.com/td7x/asdf/adr-tools"
-  ],
-  "ag": ["vfox:mise-plugins/vfox-ag", "asdf:mise-plugins/mise-ag"],
-  "age": ["aqua:FiloSottile/age", "asdf:threkk/asdf-age"],
-  "age-plugin-yubikey": [
-    "github:str4d/age-plugin-yubikey",
-    "asdf:joke/asdf-age-plugin-yubikey",
-    "cargo:age-plugin-yubikey"
-  ],
-  "agebox": ["aqua:slok/agebox", "asdf:slok/asdf-agebox"],
-  "aichat": ["aqua:sigoden/aichat"],
-  "air": ["aqua:air-verse/air", "asdf:pdemagny/asdf-air"],
-  "aks-engine": ["aqua:Azure/aks-engine", "asdf:robsonpeixoto/asdf-aks-engine"],
-  "allure": [
-    "github:allure-framework/allure2",
-    "asdf:mise-plugins/mise-allure"
-  ],
-  "allurectl": ["github:allure-framework/allurectl"],
-  "alp": ["aqua:tkuchiki/alp", "asdf:asdf-community/asdf-alp"],
-  "amass": ["github:owasp-amass/amass", "asdf:dhoeric/asdf-amass"],
-  "amazon-ecr-credential-helper": [
-    "aqua:awslabs/amazon-ecr-credential-helper",
-    "asdf:dex4er/asdf-amazon-ecr-credential-helper"
-  ],
-  "amazon-ecs-cli": ["aqua:aws/amazon-ecs-cli"],
-  "amp": ["npm:@sourcegraph/amp"],
-  "amplify": [
-    "github:aws-amplify/amplify-cli",
-    "asdf:LozanoMatheus/asdf-aws-amplify-cli"
-  ],
-  "android-sdk": ["vfox:mise-plugins/vfox-android-sdk"],
-  "ansible": ["pipx:ansible"],
-  "ansible-base": ["pipx:ansible-core"],
-  "ansible-core": ["pipx:ansible-core"],
-  "ant": ["vfox:mise-plugins/vfox-ant"],
-  "apko": ["aqua:chainguard-dev/apko", "asdf:omissis/asdf-apko"],
-  "apollo-ios": ["github:apollographql/apollo-ios"],
-  "apollo-ios-cli": ["github:apollographql/apollo-ios"],
-  "apollo-router": [
-    "github:apollographql/router",
-    "asdf:safx/asdf-apollo-router"
-  ],
-  "apollo-rover": ["github:apollographql/rover"],
-  "aqua": ["github:aquaproj/aqua"],
-  "arduino": ["aqua:arduino/arduino-cli", "asdf:egnor/asdf-arduino-cli"],
-  "arduino-cli": ["aqua:arduino/arduino-cli", "asdf:egnor/asdf-arduino-cli"],
-  "argc": ["github:sigoden/argc"],
-  "argo": ["aqua:argoproj/argo-workflows", "asdf:sudermanjr/asdf-argo"],
-  "argo-rollouts": [
-    "aqua:argoproj/argo-rollouts",
-    "asdf:abatilo/asdf-argo-rollouts"
-  ],
-  "argocd": ["aqua:argoproj/argo-cd", "asdf:beardix/asdf-argocd"],
-  "asciidoctorj": [
-    "vfox:mise-plugins/vfox-asciidoctorj",
-    "asdf:mise-plugins/mise-asciidoctorj"
-  ],
-  "assh": ["github:moul/assh", "asdf:mise-plugins/mise-assh"],
-  "ast-grep": [
-    "aqua:ast-grep/ast-grep",
-    "cargo:ast-grep",
-    "npm:@ast-grep/cli",
-    "pipx:ast-grep-cli"
-  ],
-  "astro": ["github:astronomer/astro-cli"],
-  "atlas": ["aqua:ariga/atlas", "asdf:komi1230/asdf-atlas"],
-  "atlas-community": ["aqua:ariga/atlas/community"],
-  "atmos": ["aqua:cloudposse/atmos", "asdf:cloudposse/asdf-atmos"],
-  "atuin": ["aqua:atuinsh/atuin", "cargo:atuin"],
-  "aube": ["github:endevco/aube"],
-  "auto-doc": ["github:tj-actions/auto-doc", "asdf:mise-plugins/mise-auto-doc"],
-  "aws": ["aqua:aws/aws-cli", "asdf:MetricMike/asdf-awscli"],
-  "aws-amplify": [
-    "github:aws-amplify/amplify-cli",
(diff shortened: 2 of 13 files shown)


lib/data/mise-registry.json
jamietanna, line 1: I've intentionally left this as a version from earlier this year, so we can then validate that this gets updated this weekend

jamietanna: I've not done that any more

lib/modules/manager/mise/extract.ts
```diff
@@ -113,9 +117,28 @@ function getToolConfig(
...
+      // Otherwise, see if we have any known short tool names that are in the `mise-registry.json` data file
+      for (const backendStr of getMiseRegistryBackends(toolName)) {
+        const [backendType, backendName] = backendStr.split(':', 2);
+        const result = getToolConfig(
+          backendType,
+          backendName,
+          version,
+          toolOptions,
+        );
+        if (result !== null) {
```
jamietanna, line 152: Not sure if the coverage here is straightforward to get - I think we need a tool that has multiple unsupported backends, which right now we don't have (nor may we going forwards, as it will change)

tools/static-data/generate-mise-registry.mjs
```diff
@@ -0,0 +1,25 @@
...
+import { updateJsonFile } from './utils.mjs';
+
+// use the JSON output, as the default output includes a banner
+const versionOutput = execSync('mise version --json', { encoding: 'utf8' });
+/** @type {{ version: string;  }} */
+const version = JSON.parse(versionOutput);
+console.log(`Generating mise registry using mise version ${version.version}`);
+const output = execSync('mise registry --json', { encoding: 'utf8' });
+/** @type {{ short: string; backends: string[] }[]} */
+const tools = JSON.parse(output);
```
secustor: We should do a schema validation here, especially we are running this against latest.

tools/static-data/generate-mise-registry.mjs
secustor, line 1: Shouldn't we able to rather use TS files here now?

lib/modules/manager/mise/upgradeable-tooling.ts
```diff
@@ -555,3 +556,7 @@ export const miseTooling: Record<string, ToolingDefinition> = {
   ...miseCoreTooling,
   ...miseRegistryTooling,
 };
+
+export function getMiseRegistryBackends(toolName: string): string[] {
+  return (miseRegistry as Record<string, string[]>)[toolName] ?? [];
```
secustor: I think we should schema parse here too and looking at the rest of the code there should a central Getter function which does lazy parse it

Also allows us already pre process the registry type in the transform

jamietanna:
Very fair shouts - will do!

tools/static-data/generate-mise-registry.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const registry = MiseRegistryJson.parse(
+  Object.fromEntries(
+    tools
+      .map(({ short, backends }): [string, string[]] => [short, backends])
+      .sort(([a], [b]) => a.localeCompare(b)),
+  ),
+);
+
+await updateJsonFile(
```
jamietanna, line 38: Aside: should we be calling prettier here? I know we also call it in GitHub Actions, but it may make it nicer when running tools locally

jamietanna:
In terms of whether we may be able to remove any of the inbuilt ones:

<details>
<summary> Claude Sonnet 4.6</summary>

```markdown
Here's a clear breakdown based on comparing hardcoded configs vs what the registry auto-generation would produce:

## Clearly removable (auto-gen produces identical result)

| Tool | Reason |
|------|--------|
| `npm` | Registry has `npm:npm` → same datasource + package |
| `rumdl` | Registry has `github:rvben/rumdl` → same datasource + package, no `extractVersion` in either |
| `swiftformat` | Registry has `github:nicklockwood/SwiftFormat` → same datasource + package, no `extractVersion` in either |

## Same datasource+package but missing `extractVersion` (borderline)

These have a `github:` registry entry with matching package names, but auto-gen via `createGithubToolConfig` doesn't add `extractVersion`. Without it, releases tagged `v1.2.3` would mismatch a user's `1.2.3` in mise.toml:

- `astro` → `github:astronomer/astro-cli` (hardcoded has `^v(...)`)
- `ccache` → `github:ccache/ccache` (hardcoded has `^v(...)`)
- `hivemind` → `github:DarthSim/hivemind` (hardcoded has `^v(...)`)
- `openfga` → `github:openfga/openfga` (hardcoded has `^v(...)`)
- `tusd` → `github:tus/tusd` (hardcoded has `^v(...)`)

These could be removed if `createGithubToolConfig` were updated to default-strip `v` prefixes, or left in place.

## Must keep

| Tool | Reason |
|------|--------|
| `kafka` | Not in registry at all (function-based config too) |
| `redis` | Only vfox/asdf backends in registry, no native support |
| `sqlite` | Only asdf backend in registry, special `^version-` extractVersion |
| `aws-vault` | Registry uses `ByteNess/aws-vault` (aqua fork), hardcoded uses `99designs/aws-vault` (original) |
| `pnpm` | Auto-gen picks `aqua:pnpm/pnpm` (GithubTags) first; hardcoded correctly uses NpmDatasource |
| `yarn` | Auto-gen picks `aqua:yarnpkg/berry` (GithubTags); hardcoded uses `@yarnpkg/cli` on NpmDatasource |
| `prettier` | Auto-gen gives `npm:prettier` (NpmDatasource); hardcoded uses GithubReleases — though npm is arguably *more* correct here |
| `localstack` | Registry maps to `localstack-cli` repo; hardcoded uses `localstack/localstack` |
| `protoc` | Registry maps to `protocolbuffers/protobuf/protoc` (aqua path); hardcoded uses `protocolbuffers/protobuf` |
| All other aqua-backed tools | Auto-gen uses `GithubTagsDatasource`; hardcoded uses `GithubReleasesDatasource` |
| `lychee` | Special `^lychee-v(...)` extractVersion pattern that aqua auto-gen can't replicate |
| `jq` | Special `^jq-(.*)` extractVersion pattern that auto-gen can't replicate |

The key systemic blocker for removing most entries is that the registry primarily resolves tools via `aqua:` backends, which use `GithubTagsDatasource`, whereas the hardcoded entries consistently use `GithubReleasesDatasource`. Until the auto-gen prefers `github:` backends (releases) over `aqua:` backends (tags) where both exist, the hardcoded entries remain the higher-quality option.
```

Follow-up: 

```markdown
Key change: with `github:` priority, `npm:` backends also get elevated. The notable new finding is **`pnpm`** — the registry now resolves to `npm:pnpm` (skipping `aqua:pnpm/pnpm`), which is an exact match with the hardcoded config.

Updated summary:

## Now clearly removable (auto-gen produces identical result)

| Tool | Registry preferred backend | Match |
|------|---------------------------|-------|
| `npm` | `npm:npm` → NpmDatasource, `npm` | exact |
| `pnpm` | `npm:pnpm` → NpmDatasource, `pnpm` | exact (**new**) |
| `rumdl` | `github:rvben/rumdl` → GithubReleases, same pkg, no extractVersion | exact |
| `swiftformat` | `github:nicklockwood/SwiftFormat` → GithubReleases, same pkg, no extractVersion | exact |

## Borderline (same datasource+package, but missing `^v` extractVersion)

These now use `github:` → GithubReleasesDatasource with matching package, but `createGithubToolConfig` adds no `extractVersion`. Since releases are tagged `v1.2.3` while users write `1.2.3`, there'd be a version mismatch without it:

`astro`, `ccache`, `hivemind`, `openfga`, `tusd`

These become removable if `createGithubToolConfig` defaults to stripping a `v` prefix, or if the `version_prefix` option is added to the registry entries.

## Still must keep

| Tool | Reason |
|------|--------|
| `kafka` | Not in registry |
| `redis`, `sqlite` | No native backend in registry (only vfox/asdf) |
| `aws-vault` | Registry has `ByteNess/aws-vault`, hardcoded uses `99designs/aws-vault` |
| `yarn` | Registry prefers `npm:@yarnpkg/cli-dist`, hardcoded uses `@yarnpkg/cli` |
| `lefthook` | Registry prefers `npm:lefthook` (NpmDatasource), hardcoded uses GithubReleases |
| `prettier` | Registry prefers `npm:prettier`, hardcoded uses GithubReleases |
| `localstack` | Registry has `localstack-cli`, hardcoded uses `localstack` |
| `protoc` | Registry has `protocolbuffers/protobuf/protoc` (aqua path), hardcoded uses `protocolbuffers/protobuf` |
| `lychee` | Special `^lychee-v` extractVersion not replicable by auto-gen |
| `jq` | Special `^jq-` extractVersion not replicable by auto-gen |
| All remaining aqua-backed | Still GithubTagsDatasource vs hardcoded GithubReleasesDatasource |
```

</details>

jamietanna:
> The key systemic blocker for removing most entries is that the registry primarily resolves tools via `aqua:` backends, which use `GithubTagsDatasource`, whereas the hardcoded entries consistently use `GithubReleasesDatasource`. Until the auto-gen prefers `github:` backends (releases) over `aqua:` backends (tags) where both exist, the hardcoded entries remain the higher-quality option.

is this maybe something we want to add to this logic?

Edit: I have done so

lib/modules/manager/mise/schema.ts
```diff
@@ -1,6 +1,8 @@
 import { z } from 'zod/v3';
 import { Toml } from '../../../util/schema-utils/index.ts';
 
+export const MiseRegistryJson = z.record(z.string(), z.array(z.string()));
```
viceice:
```suggestion
export const MiseRegistryJson = z.record(z.string(), z.array(z.string()).transform(sortBackends));
```

better sort once?

tools/static-data/generate-mise-registry.ts
viceice: i think we should add a post-commit hook again for prettier like we do on other repos.

lib/modules/manager/mise/schema.ts
jamietanna: How does this look? Or should we do it in the Zod schema itself?

secustor:
```suggestion
export const MiseRegistryJson = z.record(z.string(), z.array(z.string()).transform(createRecord));

function createRecord(input: string[]): Record<string,string> {
  const result:  Record<string,string> = {}
  for( const registry of input) {
    const [type, url] = registry.split(':');
    result[type] = url
  }
  return result
}
```

This is what I have meant in my https://github.com/renovatebot/renovate/pull/42783#discussion_r3122001014
Should be also make it easier to consume.

tools/static-data/generate-mise-registry.ts
```diff
@@ -0,0 +1,29 @@
...
+const MiseVersion = z.object({ version: z.string() });
+const MiseRegistry = z.array(
+  z.object({ short: z.string(), backends: z.array(z.string()) }),
+);
+
+// use the JSON output, as the default output includes a banner
+const versionOutput = execSync('mise version --json', { encoding: 'utf8' });
+const version = MiseVersion.parse(JSON.parse(versionOutput));
+console.log(`Generating mise registry using mise version ${version.version}`);
+const output = execSync('mise registry --json', { encoding: 'utf8' });
```
viceice: use `execa` and async? We're esm, so module scope async is allowed

lib/modules/manager/mise/schema.ts
```diff
@@ -1,6 +1,26 @@
...
+export const MiseRegistryJson = z
+  .record(z.string(), z.array(z.string()))
+  .transform((input) => {
+    const result: Record<string, Record<string, string>> = {};
+
+    for (const [key, val] of Object.entries(input)) {
+      result[key] = createRecord(val);
+    }
+    return result;
+  });
```
viceice: shouldn't we prepare on build time (perf)?

viceice: ☝️ ❓ 

Shouldn't this happen in `tools/static-data/generate-mise-registry.ts` so we don't need it here?

lib/modules/manager/mise/upgradeable-tooling.ts
```diff
@@ -1,3 +1,4 @@
+import miseRegistry from '../../../data/mise-registry.json' with { type: 'json' };
```
viceice, line 1: wonder if we should use a dynamic import, so we only import it if we really parse a mise file. 🤔 
we can improve in follow up pr.

lib/modules/manager/mise/schema.ts
jamietanna: Yep fair shout

lib/modules/manager/mise/upgradeable-tooling.ts
jamietanna: Happy to do so now if you'd prefer?

secustor (approved):

tools/static-data/generate-mise-registry.ts
jamietanna: I've added https://github.com/renovatebot/renovate/pull/42941 as a starting point

.github/workflows/update-data.yml
```diff
@@ -35,6 +35,14 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Install mise
+        uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
+        with:
+          # don't run `mise install`, but do install a version of `mise` itself
+          install: false
+          # explicitly always use the latest tool version
+          version: latest
```
jamietanna, line 44: Doesn't work: https://github.com/renovatebot/renovate/actions/runs/25266103377/job/74080798479

PR #42784, opened 2026-04-21
refactor(github-actions): simplify comunity action schema

viceice:
## Changes
Refactor to simple record for easier usage in docs (autogeneration)

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/manager/github-actions/community.ts
+++ b/lib/modules/manager/github-actions/community.ts
@@ -20,7 +20,7 @@ export interface CommunityActionConfig {
   /**
    * should return `true` if the version is invalid and should be skipped
    */
-  validator?: (value: string) => boolean;
+  isInvalid?: (value: string) => boolean;
 
   withSchema?: z.ZodEffects<
     z.ZodTypeAny,
@@ -32,7 +32,7 @@ type ActionSchema = z.ZodEffects<z.ZodTypeAny, PackageDependency>;
 
 function actionSchema(
   name: string,
-  { validator, withSchema, ...cfg }: CommunityActionConfig,
+  { isInvalid, withSchema, ...cfg }: CommunityActionConfig,
 ): ActionSchema {
   return z
     .object({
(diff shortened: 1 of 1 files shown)


lib/modules/manager/github-actions/community.ts
```diff
@@ -13,9 +10,48 @@ import * as condaVersioning from '../../versioning/conda/index.ts';
...
+  datasource: string;
+  depName?: string;
+  packageName: string;
+  versioning?: string;
+  extractVersion?: string;
+
+  /**
+   * should return `true` if the version is invalid and should be skipped
+   */
+  validator?: (value: string) => boolean;
```
jamietanna: Could tweak to:


```suggestion
  isInvalid?: (value: string) => boolean;
```

To make that more self-documenting?

jamietanna (commented):
I think this looks reasonable - would appreciate a second review from @secustor to confirm 

This looks a whole lot cleaner and more straightforward to read, but wondering if there may be "too much magic"?

lib/modules/manager/github-actions/community.ts
viceice: then i need to invert all conditions. currently it returns `true` for invalid versions. it didn't changed since previous implementation, see `parseValue` function.

jamietanna: Would it need to be inverted? I'm thinking that this would still return true if it is invalid

viceice: ups, I read `isValid` 🫣

jamietanna: Thought so 😉 

If you think it's not worth it, that's OK, but I _could_ have read `validator` as "it needs to return `true` if it is valid"

secustor (approved):
LGTM, but not sure if we will rack up a lot of special different schemas

jamietanna (approved):

PR #42800, opened 2026-04-21
fix(manager): add missing categories exports

RahulGautamSingh:
## Changes

Add missing `categories` exports to 7 managers: copier, fvm, gleam, osgi, pre-commit, unity3d, vendir

## Context

## AI assistance disclosure

Used Claude Code to identify managers missing categories and generate the additions.

## Documentation

## How I've tested my work

- [x] Code inspection only

### DIFF
--- a/lib/modules/manager/copier/index.ts
+++ b/lib/modules/manager/copier/index.ts
@@ -1,4 +1,8 @@
+import type { Category } from '../../../constants/index.ts';
 import { GitTagsDatasource } from '../../datasource/git-tags/index.ts';
+
+export const categories: Category[] = ['python'];
+
 import * as pep440 from '../../versioning/pep440/index.ts';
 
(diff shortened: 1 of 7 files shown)


jamietanna:
Should we require this from the API? (while allowing it to maybe be empty)

RahulGautamSingh:
> Should we require this from the API? (while allowing it to maybe be empty)

Yes, we could do that.

viceice (approved):

jamietanna:
Can we take a look at what impact that'd have / if it'd find any other cases where we're missing `categories`?

RahulGautamSingh:
That's behaviorally a no-op for `matchCategories` filtering. They're already unmatched today, but it makes the
 "no category fits" case explicit at the type level and prevents new managers from silently skipping it.

So, no impact

PR #42803, opened 2026-04-21
fix(config): only migrate standalone legacy template identifiers

fractal-6:
## Changes

Fix config migration so legacy template variables are only migrated when they appear as standalone identifiers.

This preserves the existing migration behavior for valid legacy template expressions such as `{{toVersion}}` and `{{#if toVersion}}`, while avoiding false-positive replacements inside larger identifiers like `correttoVersion`. As a result, strict config validation no longer fails for configs that were already valid.

## Context

## AI assistance disclosure

- [ ] Yes — substantive assistance (AI-generated non-trivial portions of code, tests, or documentation).

Used Cursor AI assistance for the migration implementation and regression tests.

The public repository: N/A


lib/config/migration.spec.ts
```diff
@@ -693,6 +693,31 @@ describe('config/migration', () => {
     expect(migratedConfig).toMatchSnapshot();
   });
 
+  it('does not migrate legacy template identifiers within larger identifiers', () => {
+    const config = {
```
RahulGautamSingh:
```suggestion
    const config = partial<RenovateConfig>{
```

lib/config/migration.spec.ts
```diff
@@ -693,6 +693,31 @@ describe('config/migration', () => {
...
+      autoReplaceStringTemplate:
+        'val correttoVersion = "{{{newValue}}}"\n    val imageSha256 = "{{{newDigest}}}"',
+    } as RenovateConfig;
```
RahulGautamSingh:
```suggestion
    };
```

RahulGautamSingh (changes requested):

lib/config/migration.ts
```diff
@@ -22,10 +22,45 @@ import type {
...
+
 export function fixShortHours(input: string): string {
   return input.replace(regEx(/( \d?\d)((a|p)m)/g), '$1:00$2');
 }
 
+function isTemplateIdentifierChar(char: string | undefined): boolean {
+  return !!char && templateIdentifierCharRegex.test(char);
+}
+
+function replaceStandaloneTemplateIdentifier(
```
RahulGautamSingh, line 41: Consider simplifying with a word boundary: `\b` already matches exactly between `[A-Za-z0-9_]` and any other character, so the helper can be reduced to a single `replace` without a callback or the `templateIdentifierCharRegex` / `isTemplateIdentifierChar` plumbing:

```ts
function replaceStandaloneTemplateIdentifier(input: string, from: string, to: string): string {
  return input.replace(regEx(`\\b${escapeRegExp(from)}\\b`, "g"), to);
}
```

lib/config/migration.ts
```diff
@@ -22,10 +22,45 @@ import type {
...
+  input: string,
+  from: string,
+  to: string,
+): string {
+  return input.replace(
+    regEx(escapeRegExp(from), 'g'),
```
RahulGautamSingh: Minor: `regEx(..., "g")` is intentionally not cached (see the `g`-bypass in `regEx`), so this rebuilds 6 RE2 instances for every string value on every (recursive) call to `migrateConfig`. Since `migratedTemplates` is a fixed table, consider hoisting the compiled regexes to module scope alongside it.

lib/config/migration.spec.ts
```diff
@@ -693,6 +693,31 @@ describe('config/migration', () => {
...
+    const { isMigrated, migratedConfig } =
+      configMigration.migrateConfig(config);
+    expect(isMigrated).toBeFalse();
+    expect(migratedConfig).toEqual(config);
```
RahulGautamSingh, line 705: Add a couple of cases like `toVersionSuffix`, `_toVersion`, `toVersion_`, and a `foo_toVersion_bar` .. style example so a future change to the boundary check cannot silently regress them.

fractal-6:
@RahulGautamSingh 
Thank you for your comment.

- Fixed the test typing by switching the config fixture to partial<RenovateConfig>(...) and dropping the cast.

- Expanded the negative coverage to include suffix and underscore-adjacent collisions as well: toVersionSuffix, _toVersion, toVersion_, and foo_toVersion_bar.

- Simplified the replacement logic to use \\b word boundaries, which still preserves the intended behavior for template expressions like {{toVersion}} / {{#if toVersion}} while avoiding replacements inside larger identifiers.

- Hoisted the replacement regex construction to module scope so the same compiled patterns are reused instead of being rebuilt on each migration call.

- Re-ran lib/config/migration.spec.ts; all tests are passing.

RahulGautamSingh (approved):

fractal-6:
@viceice Please check this when you have time 🙏

lib/config/migration.ts
```diff
@@ -22,6 +22,21 @@ import type {
...
+  fromVersion: 'currentVersion',
+  newValueMajor: 'newMajor',
+  newValueMinor: 'newMinor',
+  newVersionMajor: 'newMajor',
+  newVersionMinor: 'newMinor',
+  toVersion: 'newVersion',
+} as const;
+const migratedTemplatePatterns = Object.entries(migratedTemplates).map(
+  ([from, to]) => ({
+    pattern: regEx(`\\b${escapeRegExp(from)}\\b`, 'g'),
```
viceice: reusing global regex is a bad idea because they are state full. if you feed same string again then it won't match as expected.

use,`RegExp.escape`, see deprecating message.

lib/config/migration.ts
```diff
@@ -114,23 +129,16 @@ export function migrateConfig(
...
+      const migratedValue = Reflect.get(migratedConfig, key);
+      if (typeof migratedValue === 'string') {
+        let migratedStringValue = migratedValue;
+        for (const { pattern, replacement } of migratedTemplatePatterns) {
+          migratedStringValue = migratedStringValue.replace(
+            pattern,
+            replacement,
           );
         }
+        Reflect.set(migratedConfig, key, migratedStringValue);
```
viceice: why using reflect? we never use it before in our codebase

fractal-6:
@viceice 

Fixed.

I kept the RegExp.escape(...) usage, but added a local type assertion for RegExp because the current TypeScript lib definitions in this repo do not expose RegExpConstructor.escape yet, even though the runtime supports it.

Re-ran lib/config/migration.spec.ts; all tests are passing.

viceice:
> @viceice 
> 
> Fixed.
> 
> I kept the RegExp.escape(...) usage, but added a local type assertion for RegExp because the current TypeScript lib definitions in this repo do not expose RegExpConstructor.escape yet, even though the runtime supports it.

merge main and it'll be there

lib/config/migration.ts
viceice: can you please answer the question instead of simply replacing this?

viceice (commented):
please avoid huge refactor in a fix pr

fractal-6:
@viceice main was merged, but the repo still targets ES2024, while RegExp.escape is declared in lib.es2025.regexp.d.ts, so the local type assertion is still needed here.

fractal-6:
@viceice 
Trimmed this back to keep the PR focused on the fix. Removed the broader refactor and kept only the standalone template migration change plus the regression tests. Re-ran `lib/config/migration.spec.ts`; all tests are passing.

Issue #23249, opened 2023-07-07
Terraform: support pinning tags to digests


rarkins:
### Describe the proposed change(s).

Support pinning from:

```
source = "git@github.com:terraform-aws-modules/terraform-aws-kms.git?ref=v2.0.0"
```

to:

```
# renovate: version=v2.0.0
source = "git@github.com:terraform-aws-modules/terraform-aws-kms.git?ref=SOMESHA"
```

### Describe why we need/want these change(s).

Better reproducibility.

Blocked by #23248

PR #42810, opened 2026-04-22
feat(manager/terraform): support pinDigests for GitHub-sourced modules

### LINKED ISSUE #23248: Support updating digests in terraform

Supporting updating this type of annotated reference:

```
# renovate: version=v2.0.0
source = "git@github.com:terraform-aws-modules/terraform-aws-kms.git?ref=SOMESHA"
```

Where `v2.0.0` above is the tag.

### Describe why we need/want these change(s).

This is a prerequisite for pinning tags to SHAs for increased reproducibility.

Similar to GitHub Actions

adrien-mornet:
## Changes

When a Terraform module source uses a commit SHA as the `?ref=` value instead of a version tag, Renovate previously treated the SHA as an unknown version and left the dep unmanaged. This change makes Renovate properly handle SHA-pinned GitHub module sources.

The supported format (inline version comment after the SHA):
```hcl
module "example" {
  source = "github.com/terraform-aws-modules/terraform-aws-s3-bucket.git?ref=6c5e082b5d2fde77cb59c387a7f553dd2ed5da29" # v5.12.0
}
```

How it works:
1. `modules.ts` detects 40-char hex `?ref=` values as digests (`currentDigest`) instead of tags
2. `extract.ts` adds a post-processing pass over the raw file content (the HCL parser strips comments, so this must happen outside the parsed AST) to find the inline `# vX.Y.Z` comment and set `currentValue`, `replaceString`, and `autoReplaceStringTemplate`
3. When a new release is available, Renovate resolves the new tag to its commit SHA via `GithubTagsDatasource.getDigest()` and writes `?ref=<newSHA> # <newVersion>`

If there is no inline version comment, the dep is skipped with `unversioned-reference` (same behaviour as SHA-only GitHub Actions steps).

## Context

- [x] This closes an existing Issue, Closes: #23248

## AI assistance disclosure

This PR was written primarily by Claude Code (Anthropic). The implementation approach, architecture decisions, and understanding of the codebase were guided by a human engineer.

Two new test cases added to `extract.spec.ts`:
- SHA-pinned with inline version comment → correct `currentDigest`, `currentValue`, `replaceString`, `autoReplaceStringTemplate`
- SHA-pinned without version comment → `skipReason: 'unversioned-reference'`


jamietanna:
Mind rebasing to fix the commit message? Then we'll review - ty!

jamietanna:
Also, please follow the PR template :)

jamietanna:
(Question to maintainers): Would this solve both https://github.com/renovatebot/renovate/issues/23248 and https://github.com/renovatebot/renovate/issues/23249 ?

lib/modules/manager/terraform/extract.ts
```diff
@@ -55,6 +56,39 @@ export async function extractPackageFile(
...
 
+  // Post-process: for SHA-pinned GitHub module sources, extract version from
+  // inline comment in the raw file content (the HCL parser strips comments).
+  const sourceDigestWithCommentRegex = regEx(
+    /\?ref=(?<sha>[0-9a-f]{40})"(?:[^\S\r\n]*#[^\S\r\n]*(?<version>v?[^\s]+))?/gi,
+  );
+  for (const dep of dependencies) {
+    if (dep.currentDigest && dep.depType === 'module') {
+      const sha = dep.currentDigest;
+      sourceDigestWithCommentRegex.lastIndex = 0;
```
viceice: I think if you remove the global then you don't have to reset this 🤔

lib/modules/manager/terraform/extract.ts
```diff
@@ -55,6 +56,39 @@ export async function extractPackageFile(
...
+      let match: RegExpExecArray | null;
+      let found = false;
+      while ((match = sourceDigestWithCommentRegex.exec(content)) !== null) {
```
viceice: why not doing a match all before the dep loop?

or include the current digest into the regex to match the correct position

lib/modules/manager/terraform/extract.ts
adrien-mornet: Should be fixed @viceice :)

lib/modules/manager/terraform/extract.ts
adrien-mornet: Should be also fixed @viceice

viceice:
> (Question to maintainers): Would this solve both https://github.com/renovatebot/renovate/issues/23248 and https://github.com/renovatebot/renovate/issues/23249 ? 

the first, yes. the second, maybe.

lib/modules/manager/terraform/extract.ts
```diff
@@ -55,6 +56,39 @@ export async function extractPackageFile(
     dependencies.push(...deps);
   }
 
+  // Post-process: for SHA-pinned GitHub module sources, extract version from
+  // inline comment in the raw file content (the HCL parser strips comments).
+  // Scan the content once and build a lookup map before iterating over deps.
+  const shaCommentRegex = regEx(
+    /\?ref=(?<sha>[0-9a-f]{40})"(?:[^\S\r\n]*#[^\S\r\n]*(?<version>v?[^\s]+))?/gi,
```
RahulGautamSingh: The version capture group `v?[^\s]+` will match any non-whitespace token after `#`. Even a comment like `# updated 2026-04-01` would set `currentValue = "updated"`.  Tighten to something semver-shaped, e.g. `(?<version>v?\d+(?:\.\d+){0,2}\S*)`, or add a sanity check before back-filling `currentValue`.

lib/modules/manager/terraform/extract.ts
```diff
@@ -55,6 +56,39 @@ export async function extractPackageFile(
...
+  );
+  const shaInfoMap = new Map<
+    string,
+    { version: string | undefined; replaceString: string }
+  >();
+  for (const match of content.matchAll(shaCommentRegex)) {
+    const sha = match.groups!.sha;
+    shaInfoMap.set(sha, {
```
RahulGautamSingh: `shaInfoMap.set(sha, …)` unconditionally overwrites on every match. If two modules pin the same SHA .. module A as `?ref=SHA"` (no comment) and module B as `?ref=SHA" # v1.2.3` - the iteration order in `content.matchAll` decides the outcome, and A's `version: undefined` can affect B's value. 

Prefer the entry that has a `version`, or key by the full match (or its position) so each occurrence maps to its own dep.

lib/modules/manager/terraform/extractors/others/modules.ts
```diff
@@ -12,6 +12,7 @@ import type { TerraformDefinitionFile } from '../../hcl/types.ts';
 export const githubRefMatchRegex = regEx(
   /github\.com([/:])(?<project>[^/]+\/[a-z0-9-_.]+).*\?(depth=\d+&)?ref=(?<tag>.*?)(&depth=\d+)?$/i,
 );
+export const gitSha1Regex = regEx(/^[0-9a-f]{40}$/i);
```
RahulGautamSingh: `gitSha1Regex` is exported but only referenced inside this file. Drop the `export` unless an external consumer is planned.

lib/modules/manager/terraform/extract.ts
```diff
@@ -55,6 +56,47 @@ export async function extractPackageFile(
     dependencies.push(...deps);
   }
 
+  // Post-process: for SHA-pinned GitHub module sources, extract version from
+  // inline comment in the raw file content (the HCL parser strips comments).
+  // Scan the content once and build a lookup map before iterating over deps.
+  // The version comment must be semver-shaped so arbitrary trailing comments
+  // (e.g. `# updated 2026-04-01`) are not mistaken for a version.
+  const shaCommentRegex = regEx(
+    /\?ref=(?<sha>[0-9a-f]{40})"(?:[^\S\r\n]*#[^\S\r\n]*(?<version>v?\d+(?:\.\d+){0,2}\S*))?/gi,
```
RahulGautamSingh: the version group matches a bare date too. `# 2026-04-01` gets captured as version `2026-04-01`. the non-semver test only passes because `# updated ...` has a leading word. worth anchoring tighter, or at least noting the limit.

lib/modules/manager/terraform/extract.ts
```diff
@@ -55,6 +56,47 @@ export async function extractPackageFile(
...
+  );
+  const shaInfoMap = new Map<
+    string,
+    { version: string | undefined; replaceString: string }
+  >();
+  for (const match of content.matchAll(shaCommentRegex)) {
+    const sha = match.groups!.sha;
```
RahulGautamSingh: `match.groups!.sha` uses a non-null assertion but the next lines use `match.groups?.version`. pick one. a guard reads more consistent and does not bypass the type system.

lib/modules/manager/terraform/extract.ts
```diff
@@ -55,6 +56,54 @@ export async function extractPackageFile(
...
   }
 
+  // Post-process: for SHA-pinned GitHub module sources, extract version from
+  // inline comment in the raw file content (the HCL parser strips comments).
+  // Scan the content once and build a lookup map before iterating over deps.
+  // The version must be dotted-numeric (e.g. `v1.2.3`) so trailing comments
+  // such as `# updated ...`, a bare date `# 2026-04-01`, or a bare integer are
+  // not mistaken for a version. A dot-separated date (`# 2026.04.01`) remains
+  // an accepted but unlikely false positive.
+  const shaCommentRegex = regEx(
```
RahulGautamSingh: this anchors the SHA to a closing quote, so a source like `?ref=<sha>&depth=1` or `?depth=1&ref=<sha>` sets `currentDigest` in the extractor but is missed here and falls back to `unversioned-reference` even with a version comment.

Try `[?&]ref=` and allowing trailing params before the quote.

lib/modules/manager/terraform/extract.ts
```diff
@@ -55,6 +56,54 @@ export async function extractPackageFile(
...
+    /\?ref=(?<sha>[0-9a-f]{40})"(?:[^\S\r\n]*#[^\S\r\n]*(?<version>v?\d+(?:\.\d+){1,2}\S*))?/gi,
+  );
+  const shaInfoMap = new Map<
+    string,
+    { version: string | undefined; replaceString: string }
+  >();
+  for (const match of content.matchAll(shaCommentRegex)) {
```
RahulGautamSingh: this scans every terraform file even when no module uses a SHA ref.

lib/modules/manager/terraform/extract.ts
```diff
@@ -55,6 +56,54 @@ export async function extractPackageFile(
...
+
+  for (const dep of dependencies) {
+    if (dep.currentDigest && dep.depType === 'module') {
+      const info = shaInfoMap.get(dep.currentDigest);
+      if (info?.version) {
+        dep.currentValue = info.version;
+        // replaceString: SHA + closing quote + inline comment
+        // e.g. `6c5e082b...29" # v5.12.0`
+        dep.replaceString = info.replaceString;
+        dep.autoReplaceStringTemplate = '{{currentDigest}}" # {{newValue}}';
```
RahulGautamSingh: extraction is well covered, but nothing tests the update rewrite from `?ref=<sha> # v1.2.3` to `?ref=<newSHA> # <newVersion>`.

PR #42817, opened 2026-04-22
ci: never cancel in-progress release jobs

jamietanna:
## Changes
We currently try not to cancel a release when it's already running, but
we sometimes do end up with it being cancelled part-way through due to
our current `cancel-in-progress`.

We can adapt this to only trigger that case when in a PR/Merge Group
event, and then provide custom logic to cancel the release job (at the
start of the job and right before `semantic-release` runs) when we're
not the latest commit.

Follow-up to changes from https://github.com/renovatebot/renovate/pull/40630

## Context

- [x] This closes an existing Issue: Closes #40632

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -784,11 +784,14 @@ jobs:
       - name: Check for newer commits on ${{ github.ref_name }}
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REF_NAME: ${{ github.ref_name }}
+          EXPECTED_SHA: ${{ github.sha }}
+          RUN_ID: ${{ github.run_id }}
         run: |
-          latest_sha=$(git ls-remote origin "refs/heads/${{ github.ref_name }}" | cut -f1)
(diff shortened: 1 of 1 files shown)


viceice (commented):
zizmor warnings 😁

.github/workflows/build.yml
```diff
@@ -772,13 +772,28 @@ jobs:
...
+      actions: write
       contents: write
       issues: write
       pull-requests: write
       id-token: write
       packages: write
 
     steps:
+      # right at the start of the job, make sure we're not seeing another commit that's landed
+      - name: Check for newer commits on ${{ github.ref_name }}
```
jamietanna: Surprised that https://github.com/renovatebot/renovate/pull/42817/changes#r3147765508 doesn't flag here too

viceice (approved):

jamietanna:
Let's see how this goes 🤞

viceice (commented):
we should extract a action, so we can reuse it on our other expensive repos when this works as expected

PR #42819, opened 2026-04-22
chore(deps): migrate to typescript 7

jamietanna:


jamietanna:
v7 is out now, so we'll automagically update to it soon

PR #42822, opened 2026-04-22
fix(datasource/npm): require public Cache-Control to cache responses

ZackKanter:
### Current behavior

_Described in Discussion https://github.com/renovatebot/renovate/discussions/42792_

With the default `cachePrivatePackages: false`, the npm datasource is supposed to skip the shared package cache for private responses. This doesn't seem to work for GitHub Packages, which seems to omit a `Cache-Control` header and causes packages to be cached for the full hard TTL (7 days default). Every private `@org/*` packument ends up in the shared cache. Jobs within the 15-minute soft-TTL window then serve the cached response without any revalidation, so newly-published versions are invisible to Renovate until the soft TTL expires.

You can confirm this behavior in GitHub Packages by running:

```bash
curl -sS -D - -o /dev/null -H "Authorization: Bearer $(gh auth token)" https://npm.pkg.github.com/@foo%2Fbar
```

Response:

```
HTTP/2 200
content-type: application/json; charset=utf-8
date: Wed, 22 Apr 2026 13:13:22 GMT
x-github-request-id: D50C:63F1:1814E93:19EF5B3:69E8C968
```

No `Cache-Control`. No `ETag`. No `Last-Modified`. 

Under the hood, `lib/modules/datasource/npm/get.ts` configures:

```ts
checkAuthorizationHeader: false,
checkCacheControlHeader: true,
```

and `PackageHttpCacheProvider.cacheAllowed` gates the `Cache-Control` branch on `isString(resp.headers['cache-control'])`. When the header is absent, the branch is skipped, the authorization branch is skipped (disabled for npm), and execution falls through to `return true`. So `Cache-Control` behaves as a "block cache if explicitly private" gate, when for a shared cache it seems that it should be an "allow cache only if explicitly public" gate.

### New behavior

When `checkCacheControlHeader: true`, caching requires an affirmative `Cache-Control: public` signal. Missing or non-public → response is not cached.

```ts
if (this.checkCacheControlHeader) {
  const cacheControl = resp.headers['cache-control'];
  const isPublic =
    isString(cacheControl) &&
    cacheControl
      .toLocaleLowerCase()
      .split(regEx(/\s*,\s*/))
      .includes('public');

  if (!isPublic) {
    return false;
  }
}
```

Per-registry impact:

- `registry.npmjs.org`: unchanged — it sends `Cache-Control: public` and continues to cache.
- GitHub Packages and other registries that omit `Cache-Control`: no longer cached by default. Admins who know their registry is safe to share across users can still opt in via `cachePrivatePackages: true`.
- Maven and Typst datasources: unaffected — both set `checkCacheControlHeader: false`.

## Context

## AI assistance disclosure


viceice (commented):
i don't like this. this now completly disables caching of npm packages from my registry proxy.

@zharinov Can you have a look for a more elegant solution?

zharinov:
The if-statement fix is correct though: an absense of `cache-control` header must not make it cacheable.

lib/util/http/cache/package-http-cache-provider.ts
```diff
@@ -178,6 +178,24 @@ export class PackageHttpCacheProvider extends AbstractHttpCacheProvider {
...
+        true,
+      );
+      cachedResp.authorization = resp.authorization;
+
+      if (!this.cacheAllowed(cachedResp)) {
+        HttpCacheStats.incRemoteHits(url);
+        return cachedResp;
+      }
+    }
+
```
zharinov: This branch most likely is okay to omit.

zharinov (changes requested):

zharinov (approved):

secustor (approved):

PR #42843, opened 2026-04-23
feat(manager): add kas manager

tamilari:
## Changes

This pull request adds a new manager for updating kas dependencies.

Kas is a setup tool for Bitbake-based projects, primarily used in embedded Linux development.

This manager utilizes the `extractAllPackageFiles` function because kas configuration files can depend on each other and therefore need to be processed together. Since standard kas files and kas lock files are quite similar and share almost the same syntax, both will be updated using the same process.

The first step involves calling `kas dump` on the entry kas file(s) (which should be configured via the `managerFilePatterns` option) to expand and retrieve the final configuration. Subsequently, the include tree will be traversed, and each relevant dependency option for the full configuration will be extracted.

## Context

- [x] This closes an existing Issue, Closes: #41316 

## AI assistance disclosure

Many different tools were used for small code completions and general questions.


tamilari:
We are happy to add more tests to address the code coverage, but it would be great if someone could take a general look at the changes first :)

lib/modules/manager/kas/extract.ts
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    logger.debug({ packageDependency }, 'extracted dependency');
+    deps.push(packageDependency);
+  }
+  return deps.length > 0 ? { deps } : null;
+}
+
+export async function executeKasDump(file: string): Promise<KasDump | null> {
+  const cmd = `kas dump --format json ${file}`;
```
viceice, line 240: we don't run external tools for extract. can this be done differently in plain typescript / javascript ?

lib/modules/manager/kas/update.ts
```diff
@@ -0,0 +1,172 @@
+import { parse, weave } from 'jsonc-weaver';
+import { logger } from '../../../logger/index.ts';
+import { writeLocalFile } from '../../../util/fs/index.ts';
+import { escapeRegExp, regEx } from '../../../util/regex.ts';
+import { replaceAt } from '../../../util/string.ts';
+import type { UpdateDependencyConfig } from '../types.ts';
+import { isLockFilePath, isYamlFilePath } from './extract.ts';
+import type { KasRepo } from './schema.ts';
+
+export async function updateDependency({
```
viceice, line 12: why not using default auto replace?

viceice (commented):
- inline test fixtures to test with `codeBlock` template helper
- use `regEx(/.../)` helper for regex
- conflicted

lib/modules/manager/kas/update.ts
tamilari: The first problem is that the `confirmIfDepUpdated` function won't work with
the current approach. Even if we implement a kas `extractPackageFile` function
that can operate without a kas config dump, I assume this would require some
changes to the `doAutoReplace` and/or `confirmIfDepUpdated` function.

Next, extracting the `replaceString` for json files seems more difficult after
`jsonc-parser` was replaced with `jsonc-weaver`. However, directly using the
new weave function from `jsonc-weaver` appears to work quite well.

Additionally, this also enables an easy way to prevent unwanted updates (in
this case: no git commit update when the tag remains the same).

lib/modules/manager/kas/extract.ts
tamilari: Is there a specific reason for not running external tools for this? This
approach significantly simplifies the logic. With the result from `kas dump`
you can handle all the files separately. This kas logic is not trivial, and
maintaining both a Python and a Typescript version seems error-prone and not
particularly useful.

tamilari:
Sorry for the force-push; resolving the conflicts via the GitHub UI was not successful.

heiko-schabert:
@viceice  do you want to take an additional lock after the fixes?

viceice:
I usually don't review draft PRs

heiko-schabert:
> I usually don't review draft PRs

Make sense. The MR now left the DRAFT state and is ready :).

lib/modules/manager/kas/extract.ts
```diff
@@ -0,0 +1,368 @@
...
+  }
+  if (!repos) {
+    logger.debug({ packageFile }, 'no repos found in KAS file');
+    return null;
+  }
+  const repoStrings: Map<string, string> | null = isYamlFile
+    ? extractRepoStrings(content, packageFile)
+    : null;
+  const deps: PackageDependency[] = [];
+  for (const repoName in repos) {
```
RahulGautamSingh: use `for...of Object.keys(repos)` .`for...in` over objects is against best practices

lib/modules/manager/kas/extract.ts
```diff
@@ -0,0 +1,368 @@
...
+    }
+    const dumpRepo: KasRepo | null | undefined = kasDump.repos?.[repoName];
+    if (!dumpRepo) {
+      logger.debug({ packageFile }, 'no corresponding repo in dump. Skipping');
+      continue;
+    }
+    const overridesCommit: string | undefined =
+      kasDump.overrides?.repos?.[repoName]?.commit;
+    if (overridesCommit) {
+      dumpRepo.commit = overridesCommit;
```
RahulGautamSingh: `dumpRepo` is a reference into `kasDump.repos`, so this mutates the shared dump object. if the same repo appears in another included file, the second `_extractPackageFile` call sees the modified commit. capture the effective value locally instead: `const effectiveCommit = overridesCommit ?? dumpRepo.commit`

lib/modules/manager/kas/extract.ts
```diff
@@ -0,0 +1,368 @@
...
+    }
+    logger.trace(
+      { packageFile, dumpRepo, repo },
+      'corresponding dump repo details',
+    );
+
+    if (dumpRepo.type && repo.type === 'hg') {
```
RahulGautamSingh: after zod parsing, `dumpRepo.type` is always `'git'` or `'hg'` (the schema has `.default('git')`), so the `dumpRepo.type &&` guard is always truthy and does nothing. 

if the intent is to skip hg repos in either the project file or the dump, this should be `dumpRepo.type === 'hg' || repo.type === 'hg'`

lib/modules/manager/kas/extract.ts
```diff
@@ -0,0 +1,368 @@
...
+      logger.debug({ file, err }, 'Error parsing kas dump');
+    }
+    return null;
+  }
+}
+
+export async function extractAllPackageFiles(
+  config: ExtractConfig,
+  packageFiles: string[],
+): Promise<PackageFile[] | null> {
```
RahulGautamSingh: `null` is never returned here

lib/modules/manager/kas/update.ts
```diff
@@ -0,0 +1,164 @@
...
+        logger.warn(
+          { packageFile, depName },
+          'Parsed content or repos not found for non-yaml file.',
+        );
+        return fileContent;
+      }
+      let repo = repos[depName];
+      if (!repo || repo.name !== depName) {
+        for (const r of Object.values(repos)) {
+          if (r?.name === depName) {
```
RahulGautamSingh: no `break` after assignment. last matching repo wins, not first. use `Object.values(repos).find(r => r?.name === depName)` instead

lib/modules/manager/kas/extract-package-file.spec.ts
```diff
@@ -0,0 +1,618 @@
...
+
+function makeDump(
+  repos: KasDump['repos'],
+  overrides?: KasDump['overrides'],
+): KasDump {
+  return {
+    header: { version: 22 },
+    repos,
+    overrides,
+  } as KasDump;
```
RahulGautamSingh: use `satisfies KasDump` instead of `as KasDump`, or `partial<KasDump>()` from `test/util` if only a partial object is needed

tamilari:
Thanks for the review. Everything should be fixed now.

lib/modules/manager/kas/update.ts
tamilari: I realized that `repo.name` is not intended to identify a repo, so this poorly implemented code can simply be removed.

tamilari:
I've updated the branch to include the latest changes from main and incorporated the upgrade to Zod v4 for the kas manager. Everything seems to be working correctly now again.

lib/modules/manager/kas/extract.ts
```diff
@@ -0,0 +1,366 @@
...
+
+export function getLockFilePath(filePath: string): string | null {
+  if (isLockFilePath(filePath)) {
+    logger.trace(
+      { filePath },
+      'file path is already a lock file path. Returning null',
+    );
+    return null;
+  }
+  const lockFilePath = filePath.replace(/\.(yml|yaml|json)$/i, '.lock.$1');
```
RahulGautamSingh: use `regEx()` here instead of a raw literal, the rest of the file already does.

lib/modules/manager/kas/extract.ts
```diff
@@ -0,0 +1,366 @@
...
+    if (repoItem.key.range && repoNode?.range) {
+      const [keyStart] = repoItem.key.range;
+      const [, valueEnd] = repoNode.range;
+      map.set(repoName, content.substring(keyStart, valueEnd));
+    }
+  }
+  return map;
+}
+
+export function _extractPackageFile(
```
RahulGautamSingh: drop the `_` prefix, and avoid exporting just for tests. test through `extractAllPackageFiles` if you can.

lib/modules/manager/kas/extract.ts
```diff
@@ -0,0 +1,366 @@
...
+        );
+        replaceString = content;
+      }
+      logger.trace({ replaceString, repoName }, 'string to replace for repo');
+    }
+
+    const packageDependency: PackageDependency = {
+      depName: repoName,
+      packageName: git,
+      versioning: repo.branch ? looseVersioning : undefined,
```
RahulGautamSingh: versioning keys off `repo.branch` (file) but `currentValue` uses `dumpRepo.branch`. if the file omits branch but the dump has one, you get a branch value with no loose versioning. intended?

lib/modules/manager/kas/extract.ts
```diff
@@ -0,0 +1,366 @@
...
+      if (packageFileContent) {
+        results.push({
+          packageFile: file,
+          ...packageFileContent,
+        });
+      }
+    }
+
+    logger.debug(
+      { packageFiles, files: filesToExamine.entries() },
```
RahulGautamSingh: `filesToExamine.entries()` logs an iterator, and the array is already drained here. drop it or log the original list.

lib/modules/manager/kas/update.ts
```diff
@@ -0,0 +1,157 @@
...
+  const {
+    depName,
+    datasource,
+    currentValue,
+    newValue,
+    currentDigest,
+    newDigest,
+  } = upgrade;
+  logger.debug({ packageFile }, 'kas.updateDependency');
+  if (datasource === 'git-tags' && currentValue === newValue) {
```
RahulGautamSingh: use `GitTagsDatasource.id` instead of the `'git-tags'` literal.

lib/modules/manager/kas/update.ts
```diff
@@ -0,0 +1,157 @@
...
+    logger.debug(
+      { packageFile, depName, currentValue, newDigest },
+      'git tag version did not change. Skipping digest update.',
+    );
+    return fileContent;
+  }
+  if (isYamlFilePath(packageFile)) {
+    const replaceString = upgrade.replaceString;
+    const searchIndex: number = fileContent.indexOf(replaceString!);
```
RahulGautamSingh: `replaceString!` bypasses the type. guard with `is.string(replaceString)` and return early instead.

lib/modules/manager/kas/extract.ts
tamilari: wasn't intended, fixed :+1:

lib/modules/manager/kas/extract.ts
tamilari: Every function that was just exported for tests now gets tested through `extractAllPackageFiles`.

lib/modules/manager/kas/extract.ts
RahulGautamSingh, line 1: I am not sure about exec in extract.ts. Need to wait for a maintainer input for this.

RahulGautamSingh (commented):
Else LGTM

lib/modules/manager/kas/extract.ts
tamilari: Could you perhaps ask someone to take a look, or should we ask on the Slack channel?

lib/modules/manager/kas/extract.ts
viceice: we've multiple reasons, the important part is performance.

we only allow tools in artifacts updates.

we don't intend to do an exemption here.

@zharinov can we create a simple parser for extraction?

viceice: wait, those are yaml and json files which are simple to read. so no new parser needed at all. just some schema validation.

tamilari: But isn't the performance problem only relevant for users who have enabled the manager? And can't they be made aware of this limitation?
Yes, these yaml and json files are simple to read and parse, but interpreting them is not trivial. They can extend and override each other, including referencing kas files from other repositories. It only sounds logical to us to keep this logic in one place in kas. I can't think of a robust solution with a simple parser or schema validation. Am I misunderstanding something?

jamietanna: So, the reason that we don't execute tools - when we can - is:

- performance: it's generally much quicker to call a Typescript function, than it is to shell out to a new process
- security: a lot of package managers we work with may implicitly execute code, so we have to protect against that attack surface

@zharinov I'd appreciate your review on the complexity and/or domain-specific handling of the kas files - if the best option is to shell out to `kas` itself, that's fine

zharinov: For Renovate GitHub app I'd "slop fork" (in a good way) the relevant part of the resolver logic to make extract local and cheap just like for any other manager. But this means incompleteness which users may not tolerate. Seems like `kas dump` needs to make transitive fetching of repos to perform it's dump, which is something we may not want during extract.

For self-hosted, I wouldn't mind to run `kas dump`.

jamietanna: Thanks Sergei - chatting a little bit more offline, I agree that there's additional performance concerns (`kas dump` could lead to cloning repos)

If we're able to cache / control the performance implication for those repos, it's possible, but yes, keeping this as Typescript only makes most sense

[`allowCustomCrateRegistries`](https://docs.renovatebot.com/self-hosted-configuration/#allowcustomcrateregistries) is a similar case (https://github.com/renovatebot/renovate/discussions/18872) where due to caching we're not exposing this by default

RahulGautamSingh (commented):
@tamilari 

We cannot accept the PR as-is. You'll need to implement the extraction logic without replying on the exec call

Adding here this comment from jamie for more visibility:
https://github.com/renovatebot/renovate/pull/42843#discussion_r3735721464

> Thanks Sergei - chatting a little bit more offline, I agree that there's additional performance concerns (kas dump could lead to cloning repos)

> If we're able to cache / control the performance implication for those repos, it's possible, but yes, keeping this as Typescript only makes most sense

lib/modules/manager/kas/extract.ts
heiko-schabert: Thanks for the feedback!

Wewill refactor the implementation to handle the extraction logic natively in JavaScript/TypeScript without relying on `exec`. 

We will update the PR as soon as it's ready.

PR #42849, opened 2026-04-23
chore(azure): simplify calculation for `getMergeMethod`

jamietanna:
## Changes

Instead of the complicated check we were previously performing which:

- recovered from errors if `policyConfigurations` wasn't undefined, or
  other errors occurred
- trimmed the field names and checked if they were one of the
  `GitPullRequestMergeStrategy` options

We can instead rewrite this with stronger types and compile time checks
to confirm that the mapping from a policy configuration matches the
different `GitPullRequestMergeStrategy`s.

This provides a clearer intent for what this code is doing.

This changes behaviour slightly - previously, the first JSON key defined
that was a `GitPullRequestMergeStrategy` would be used (which is JSON
implementation specific), whereas now it is a defined order.

We also add coverage for the new behaviour.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/platform/azure/azure-helper.ts
+++ b/lib/modules/platform/azure/azure-helper.ts
@@ -117,6 +117,23 @@ export async function getCommitDetails(
   return results;
 }
 
+interface MergeStrategyPolicyConfiguration {
+  allowNoFastForward?: boolean;
+  allowSquash?: boolean;
+  allowRebase?: boolean;
+  allowRebaseMerge?: boolean;
+}
+
+const policyKeyByStrategy: Record<
+  GitPullRequestMergeStrategy,
+  keyof MergeStrategyPolicyConfiguration
+> = {
+  [GitPullRequestMergeStrategy.NoFastForward]: 'allowNoFastForward',
+  [GitPullRequestMergeStrategy.Squash]: 'allowSquash',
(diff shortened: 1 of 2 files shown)


jamietanna:
Wondering if we should do it as a "fix" to:

- get it released
- indicate that there may be functionality change

?

lib/modules/platform/azure/azure-helper.ts
```diff
@@ -165,30 +182,24 @@ export async function getMergeMethod(
...
-    logger.debug(
-      { policyConfigurations },
-      // TODO: types (#22198)
-      `getMergeMethod(branchRef=${branchRef!})=NoFastForward, as an error occured`,
-    );
-    return GitPullRequestMergeStrategy.NoFastForward;
+  // Note that this will iterate in the order of GitPullRequestMergeStrategy
+  for (const [key, policyKey] of Object.entries(policyKeyByStrategy)) {
+    if (policyConfigurations?.[policyKey] === true) {
+      const method = Number(key) satisfies GitPullRequestMergeStrategy;
```
viceice:
```suggestion
      const method = parseInt(key, 10) satisfies GitPullRequestMergeStrategy;
```
don't use number coersion

viceice (commented):
otherwise LGTM

viceice (approved):

PR #42852, opened 2026-04-23
fix(git): support repo cache with `getBranchUpdateDate`

jamietanna:
## Changes
As noted in #41953, the introduction of `getBranchUpdateDate` did not
take into account the use of a repository cache.

When using a repository cache, the repo is not cloned until it is needed
(and various metadata is used from the cache instead).

However, when using a cache (with a repo not yet cloned) this results in
`getBranchUpdateDate` failing (and a debug log appearing).

We can instead handle this by performing a repo sync at this point.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/util/git/index.ts
+++ b/lib/util/git/index.ts
@@ -581,6 +581,7 @@ export async function getBranchUpdateDate(
   if (!branchSha) {
     return null;
   }
+  await syncGit();
   try {
     return await getCommitDate(branchSha);
   } catch (err) {
--- a/lib/util/git/index.spec.ts
+++ b/lib/util/git/index.spec.ts
@@ -485,6 +485,41 @@ describe('util/git/index', { timeout: 10000 }, () => {
       // Restore original implementation
       fromISOSpy.mockRestore();
     });
+
+    it('works if running with a Repo Cache', async () => {
+      // Create a branch with a commit so it exists in the remote
+      const branchName = 'renovate/cache-test-branch';
+      await git.commitFiles({
+        branchName,
+        files: [
+          {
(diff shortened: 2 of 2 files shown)


lib/util/git/index.ts
```diff
@@ -581,6 +581,7 @@ export async function getBranchUpdateDate(
   if (!branchSha) {
     return null;
   }
+  await syncGit();
```
jamietanna, line 584: I'm not sure if this is the right call here - I wonder if we should instead try and get this from the cache?

viceice: yes, I think we should have this in cache 🤔

caching can be done in a follow up PR

viceice (approved):

lib/util/git/index.ts
felipecrs: Thanks a lot for the quick patch, @jamietanna.

> caching can be done in a follow up PR

And please let me know if you'd like help with that. 😊

jamietanna:
The main issue here is that it'll mean anyone using a repo cache (most folks at scale, Mend-hosted apps) won't have `commitHourlyLimit` working as expected, which may be a subset of users, but is still important to fix - so we'll get this in as-is, then follow-up with moving this into the cache

lib/util/git/index.ts
jamietanna: Thanks - I'm happy if you've got time to pick up full caching implementation PR, @felipecrs

felipecrs: Sure thing! I'll look into it.

PR #42887, opened 2026-04-24
feat: add generic statusCheckWhen option to control any renovate status check

pierluigilenoci:
## Summary

Adds a generic `statusCheckWhen` option that lets users control when each Renovate status check is set on branches. Keys mirror `statusCheckNames` (`artifactError`, `configValidation`, `mergeConfidence`, `minimumReleaseAge`); values are `"always"` | `"never"` | `"failed"`. Defaults preserve current behavior.

This makes it possible to mark `renovate/artifacts` as a required check in branch protection by setting `statusCheckWhen.artifactError = "always"`, and to silence checks entirely with `"never"`.

## Changes

- `lib/config/options/index.ts` — add the `statusCheckWhen` option with default `{ artifactError: "failed", configValidation: "always", mergeConfidence: "always", minimumReleaseAge: "always" }`.
- `lib/config/types.ts`, `lib/config/validation.ts` — types and per-key validation for the new option.
- `lib/workers/repository/update/branch/artifacts.ts` — honor `statusCheckWhen.artifactError`. The `"never"` debug line only fires when there are actual errors to skip (review feedback from @viceice).
- `lib/workers/repository/update/branch/status-checks.ts` — `setStability` and `setConfidence` honor `statusCheckWhen.minimumReleaseAge` / `mergeConfidence`. The `"never"` debug line is gated on the corresponding status being set.
- `lib/workers/repository/update/branch/{artifacts,status-checks}.spec.ts` — coverage for all modes (`always`/`never`/`failed`) and the no-log gating.
- `docs/usage/configuration-options.md` — describe the option, example, defaults table inside autogeneration markers; `platformAutomerge` note about pairing with `statusCheckWhen.artifactError = "always"`.
- `tools/docs/config.ts` — `generateStatusCheckWhenTable()` builds the defaults table from the option's `default` so the docs cannot drift from the source.

## Test plan

- [x] `pnpm test lib/workers/repository/update/branch/artifacts.spec.ts`
- [x] `pnpm test lib/workers/repository/update/branch/status-checks.spec.ts`
- [x] `pnpm test lib/config/validation.spec.ts`
- [x] `pnpm docs` regenerates the defaults table

Status: ready for review.


viceice (commented):
I think this should be more generic to allow control any renovate status check in future. 🤔

pierluigilenoci:
@viceice Good point — I've reworked this to be fully generic. Instead of the narrow `artifactsCheckWhen`, there's now a `statusCheckWhen` object that can control **any** renovate status check:

```json
{
  "statusCheckWhen": {
    "artifactError": "failed",
    "configValidation": "always",
    "mergeConfidence": "never",
    "minimumReleaseAge": "failed"
  }
}
```

Keys are the same `StatusCheckKey` values from `statusCheckNames`. Values: `"always" | "never" | "failed"`. Defaults preserve existing behavior. Object is `mergeable: true` so users only specify what they want to override.

As new status checks are added to Renovate, they automatically gain `statusCheckWhen` support. Let me know if this matches what you had in mind!

docs/usage/configuration-options.md
```diff
@@ -4867,6 +4867,40 @@ This option is mergeable, which means you only have to specify the status checks
...
+
+- `always`: Renovate always sets the status check on every branch commit — green on success, red on failure.
+- `never`: Renovate never sets this status check. Use this to skip a check entirely.
+- `failed`: Renovate only sets the status check when there is a failure. Successes are silent (no green check).
+
+Default values:
+
+| Key                 | Default  | Reasoning                                        |
+| ------------------- | -------- | ------------------------------------------------ |
+| `artifactError`     | `failed` | Legacy behavior — only reports artifact failures |
```
viceice: @jamietanna we should autogenerate this list

lib/workers/repository/update/branch/artifacts.ts
```diff
@@ -6,8 +6,19 @@ import type { BranchConfig } from '../../../types.ts';
 export async function setArtifactErrorStatus(
   config: BranchConfig,
 ): Promise<void> {
-  if (!config.artifactErrors?.length) {
-    // no errors
+  const mode = config.statusCheckWhen?.artifactError ?? 'failed';
+
+  if (mode === 'never') {
+    logger.debug(
```
viceice: don't log if there is nothing to skip (no error)

lib/workers/repository/update/branch/status-checks.ts
```diff
@@ -69,10 +69,23 @@ export interface StabilityConfig extends RenovateConfig {
 }
 
 export async function setStability(config: StabilityConfig): Promise<void> {
+  const mode = config.statusCheckWhen?.minimumReleaseAge ?? 'always';
+
+  if (mode === 'never') {
+    logger.debug(
```
viceice: same here

lib/workers/repository/update/branch/status-checks.ts
```diff
@@ -106,6 +119,15 @@ export interface ConfidenceConfig extends RenovateConfig {
 }
 
 export async function setConfidence(config: ConfidenceConfig): Promise<void> {
+  const mode = config.statusCheckWhen?.mergeConfidence ?? 'always';
+
+  if (mode === 'never') {
+    logger.debug(
```
viceice: same

docs/usage/configuration-options.md
jamietanna: Yep, it'd be good to add a `tools/docs` autogenerated option here

RahulGautamSingh (commented):
Should we add a note in the `platformAutomerge` docs as part of this PR, informing users to use the `renovate/artifacts` status check as mandatory?

jamietanna:
Yes we should!

jamietanna:
But I would also say that switching it on by default (in a follow-up PR) is fine, as it's non-breaking 👍🏽

docs/usage/configuration-options.md
pierluigilenoci: Done — added `<!-- status-check-when-defaults-begin/end -->` markers in `docs/usage/configuration-options.md` and a `generateStatusCheckWhenTable()` in `tools/docs/config.ts` that builds the table from the option `default` so it cannot drift from the source of truth in `lib/config/options/index.ts`.

lib/workers/repository/update/branch/artifacts.ts
pierluigilenoci: Done — gated on `hasErrors` so the "skipping" debug line only fires when there is actually something to skip.

lib/workers/repository/update/branch/status-checks.ts
pierluigilenoci: Done — gated on `config.stabilityStatus`, same pattern as artifacts.ts.

lib/workers/repository/update/branch/status-checks.ts
pierluigilenoci: Done — gated on `config.branchName && config.confidenceStatus`, same pattern.

docs/usage/configuration-options.md
pierluigilenoci: Done — added in this push (`tools/docs/config.ts:generateStatusCheckWhenTable()` driven by the option default).

pierluigilenoci:
Pushed updates addressing all four review comments from @viceice and @jamietanna:

1. **Autogenerated `statusCheckWhen` defaults table** — added `<!-- status-check-when-defaults-begin/end -->` markers in `docs/usage/configuration-options.md` and a `generateStatusCheckWhenTable()` in `tools/docs/config.ts` that derives the table from the option `default` field, so the docs always match `lib/config/options/index.ts`.
2. **Quieter logging** — `artifacts.ts` only logs the "never, skipping" line when there are actual errors to skip. Same gating in `status-checks.ts` for the stability and merge-confidence skips (gated on `stabilityStatus` / `branchName && confidenceStatus`). New spec cases assert the no-log path.
3. **`platformAutomerge` doc note** — added a note in the `platformAutomerge` section pointing readers to set `statusCheckWhen.artifactError = "always"` when they want to require the `renovate/artifacts` check via branch protection (since the default `"failed"` would otherwise leave the check absent on green PRs).

Branch was rebased onto current `main` while I was at it. Happy to switch the default on by default in a follow-up PR per @jamietanna's suggestion. Thanks for the review!

docs/usage/configuration-options.md
viceice: remove the manual list and only use a single replace marker. see recent changes of replace function

tools/docs/config.ts
```diff
@@ -279,6 +279,28 @@ function generateCacheNamespacesList(): string {
   return list;
 }
 
+function generateStatusCheckWhenTable(): string {
+  const option = options.find((o) => o.name === 'statusCheckWhen');
+  const defaults = (option?.default ?? {}) as Record<string, string>;
```
viceice: use `coerceObject` to simplify code

tools/docs/config.ts
```diff
@@ -455,6 +477,13 @@ export async function generateConfig(dist: string, bot = false): Promise<void> {
     );
   }
 
+  if (!bot) {
+    content = replaceContent(content, generateStatusCheckWhenTable(), {
+      replaceStart: '<!-- status-check-when-defaults-begin -->',
+      replaceStop: '<!-- status-check-when-defaults-end -->',
```
viceice: need refactor, there's now only a single replace marker

docs/usage/configuration-options.md
pierluigilenoci: Done. Switched to a single `<!-- status-check-when-defaults-begin -->` marker in the docs (removed the manual table and the end marker). The `replaceContent` call now uses the standard string-based API.

tools/docs/config.ts
pierluigilenoci: Done. Now using `coerceObject<Record<string, string>>(option?.default)` instead of `(option?.default ?? {}) as Record<string, string>`.

tools/docs/config.ts
pierluigilenoci: Done. Refactored to use the standard `replaceContent(content, text, marker)` signature with a single marker string, matching how all other replace calls work in this file.

pierluigilenoci:
@viceice Done — I've resolved the conversations. My usual practice is to let the reviewer close threads themselves once they're satisfied with the fix, hence the confusion. Apologies for the delay!

viceice:
> @viceice Done — I've resolved the conversations. My usual practice is to let the reviewer close threads themselves once they're satisfied with the fix, hence the confusion. Apologies for the delay!

you would know it if you read our contributer and best practices guides

PR #42909, opened 2026-04-26
feat(bun): add catalog extraction and update support

rhuanbarreto:
## Changes

Adds support for [Bun catalogs](https://bun.sh/docs/install/catalogs) in Renovate, following the same architecture as the existing pnpm/yarn catalog implementations.

- Extends the shared `extractCatalogDeps()` utility to accept `'bun'` as a catalog manager
- Extracts catalog dependencies from root `package.json` (both top-level and `workspaces`-nested `catalog`/`catalogs` fields)
- Introduces a bun-specific `updateDependency` that uses `jsonc-weaver` for format-preserving JSON updates, delegating non-catalog deps to the npm manager's `updateDependency`
- Documents catalog support in the bun manager readme

### Key design decisions (incorporating feedback from #39251)

- **Bun owns its update logic** — catalog updates are handled in `bun/update.ts`, not in the npm manager ([feedback](https://github.com/renovatebot/renovate/pull/39251#discussion_r1838995839))
- **Uses `jsonc-weaver`** for format-preserving JSON updates instead of custom string manipulation ([feedback](https://github.com/renovatebot/renovate/pull/39251#discussion_r1838995839))
- **No `any` types**

### File changes

| File | Change |
|------|--------|
| `npm/extract/common/catalogs.ts` | Add `BUN_CATALOG_DEPENDENCY` constant, widen `npmManager` type to include `'bun'` |
| `bun/extract.ts` | Add `bunCatalogsToArray()`, integrate catalog extraction into `extractAllPackageFiles()` |
| `bun/update.ts` | **New** — bun-specific `updateDependency` using `jsonc-weaver`, delegates non-catalog deps to npm |
| `bun/index.ts` | Export `updateDependency` from `./update.ts` instead of npm manager |
| `npm/dep-types.ts` | Add `bun.catalog.<name>` to dynamic depType documentation |
| `bun/readme.md` | Document catalog support, depType values, and `packageRules` usage |

Related discussion: https://github.com/renovatebot/renovate/discussions/36886
Previous stalled attempt: #39251 — this PR is a fresh implementation incorporating the maintainer feedback from that PR.

## Context

## AI assistance disclosure


rhuanbarreto:
This PR implements the feature requested in https://github.com/renovatebot/renovate/discussions/36886 (Support bun catalogs).

It's a fresh implementation that incorporates the feedback from the previous stalled attempt in #39251 by @LuggaPugga — specifically the review comments from @viceice about:
- Not handling bun catalog updates in the npm manager
- Using `jsonc-weaver` instead of custom JSON manipulation
- Avoiding rewriting the whole file

cc @jbergstroem who linked the discussion to the previous PR.

LuggaPugga:
Really appreciate your fresh implementation.
I’ve been quite busy recently and haven’t gotten around to refactoring mine yet 😅

lib/modules/manager/bun/update.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  if (catalogName === 'default') {
+    // Default catalog: look in `catalog` at top level and under `workspaces`
+    if (isPlainObject(parsedContents.catalog)) {
+      targets.push(parsedContents.catalog as Record<string, string>);
+    }
+    const workspaces = parsedContents.workspaces;
+    if (
+      isPlainObject(workspaces) &&
+      isPlainObject((workspaces as Record<string, unknown>).catalog)
```
RahulGautamSingh: `isPlainObject` from `@sindresorhus/is` is a type guard ie. it narrows `workspaces` to `Record<PropertyKey, unknown>`. The `as Record<string, unknown>` casts are redundant.

lib/modules/manager/bun/update.ts
```diff
@@ -0,0 +1,164 @@
...
+      packageFile: packageFileName,
+      upgrade,
+    });
+  }
+
+  const catalogName = catalogMatch.groups.catalogName;
+
+  let { newValue } = upgrade;
+  newValue = getNewGitValue(upgrade) ?? newValue;
+  newValue = getNewNpmAliasValue(newValue, upgrade) ?? newValue;
```
RahulGautamSingh, line 101: `getNewGitValue` / `getNewNpmAliasValue` aren't covered by any test and are unlikely to apply to catalog entries.

lib/modules/manager/bun/update.ts
```diff
@@ -0,0 +1,164 @@
...
+    return weave(fileContent, parsedContents);
+  } catch (err) {
+    logger.warn(
+      { err },
+      'Error weaving JSON to preserve formatting for bun catalog update, falling back to JSON.stringify',
+    );
+    // Fallback: detect indentation from original and stringify
+    const indentRe = /^(\s+)"/m;
+    const indent = indentRe.exec(fileContent)?.[1] ?? '  ';
+    return JSON.stringify(parsedContents, null, indent) + '\n';
```
RahulGautamSingh: This fallback only fires if `weave` throws on already-valid JSON, which shouldn't happen in practice. Return `null` and log instead .

lib/modules/manager/bun/extract.ts
```diff
@@ -67,12 +118,23 @@ export async function extractAllPackageFiles(
...
     const packageFile = getSiblingFileName(lockFile, 'package.json');
-    const res = await processPackageFile(packageFile, config);
-    if (res) {
+    const processResult = await processPackageFile(packageFile, config);
+    if (processResult) {
+      const { packageFileResult: res, packageJson } = processResult;
+
+      // Extract bun catalog dependencies from the root package.json
+      const bunCatalogs = bunCatalogsToArray(
+        packageJson as unknown as Record<string, unknown>,
```
RahulGautamSingh: Drop the double cast: type `bunCatalogsToArray`'s parameter as `NpmPackage` (or `Record<string, unknown>` and pass `packageJson` directly).

RahulGautamSingh (commented):
Thanks for the PR! Could you please look into the coverage issues?

lib/modules/manager/bun/update.ts
rhuanbarreto: Fixed

lib/modules/manager/bun/extract.ts
rhuanbarreto: Fixed

rhuanbarreto (commented):
Thanks for the review! ~~I tried to fix the patch coverage but I cannot see what's missing in codecov. Can you help me out?~~ I fixed it myself.

lib/modules/manager/bun/update.ts
rhuanbarreto: Fixed

lib/modules/manager/bun/update.ts
rhuanbarreto: Unfortunately I use that myself. It's not that unlikely. Will try my best to cover it

rhuanbarreto:
@RahulGautamSingh thanks for all feedback! All things fixed including coverage.

RahulGautamSingh (approved):
lgtm

hckhanh:
Is there anything that I can support with?

rhuanbarreto:
We are waiting for a review from @viceice or any contributors in the repo.

fskarmeta:
This would be really handy :)

lib/modules/manager/bun/extract.ts
```diff
@@ -44,11 +50,56 @@ export async function processPackageFile(
...
+    | Record<string, Record<string, string>>
+    | undefined;
+
+  const workspaces = packageJson.workspaces;
+  if (isPlainObject(workspaces)) {
+    catalog ??= workspaces.catalog as Record<string, string> | undefined;
+    catalogs ??= workspaces.catalogs as
+      | Record<string, Record<string, string>>
+      | undefined;
+  }
```
viceice: this should be done with a zod schema which does the necessary transform

lib/modules/manager/bun/update.spec.ts
```diff
@@ -0,0 +1,454 @@
...
+      });
+
+      expect(result).not.toBeNull();
+      const parsed = JSON.parse(result!);
+      expect(parsed.workspaces.catalogs.build.webpack).toBe('5.88.2');
+    });
+
+    it('preserves formatting when updating', () => {
+      // Use specific indentation to verify formatting preservation
+      const fileContent = [
```
viceice: can't we simply use `codeBlock` or `stripIndent` helpers?

lib/modules/manager/npm/dep-types.ts
```diff
@@ -66,4 +66,4 @@ export const knownDepTypes = [
 ] as const satisfies readonly DepTypeMetadata[];
 
 export const supportsDynamicDepTypesNote =
-  'Additionally, catalog dependencies produce dynamic `depType` values: `pnpm.catalog.<name>` for [pnpm catalogs](https://pnpm.io/catalogs#defining-catalogs) and `yarn.catalog.<name>` for [yarn catalogs](https://yarnpkg.com/features/catalogs).';
+  'Additionally, catalog dependencies produce dynamic `depType` values: `pnpm.catalog.<name>` for [pnpm catalogs](https://pnpm.io/catalogs#defining-catalogs), `yarn.catalog.<name>` for [yarn catalogs](https://yarnpkg.com/features/catalogs), and `bun.catalog.<name>` for [bun catalogs](https://bun.sh/docs/install/catalogs).';
```
viceice: sort alphabetical

lib/modules/manager/npm/extract/common/catalogs.ts
```diff
@@ -5,17 +5,26 @@ import { extractDependency, parseDepName } from './dependency.ts';
...
+const catalogPrefixes: Record<CatalogManager, string> = {
+  pnpm: PNPM_CATALOG_DEPENDENCY,
+  yarn: YARN_CATALOG_DEPENDENCY,
+  bun: BUN_CATALOG_DEPENDENCY,
+};
 
 /**
  * In order to facilitate matching on specific catalogs, we structure the
- * depType as `[pnpm|yarn].catalog.default`, `[pnpm|yarn].catalog.react17`, and so on.
+ * depType as `[pnpm|yarn|bun].catalog.default`, `[pnpm|yarn|bun].catalog.react17`, and so on.
```
viceice: alphabetical

viceice (commented):
are normal deps are grouped by catalog name is they reference one? would make sense to me like we do for maven Gradle variables. otherwise each single update would always contain same changes?

lib/modules/manager/bun/extract.ts
rhuanbarreto: Done. Added a `BunCatalogs` zod schema in `npm/schema.ts` with a `.transform()` that merges top-level and `workspaces`-nested `catalog`/`catalogs` fields (with `.catch(undefined)` on `workspaces` to handle the array-of-strings case). Follows the same pattern as `PnpmCatalogs`/`YarnCatalogs`. The `bunCatalogsToArray` function now takes the typed zod output and is private, matching `pnpmCatalogsToArray`/`yarnCatalogsToArray`.

lib/modules/manager/bun/update.spec.ts
rhuanbarreto: Done. Switched to `codeBlock` from `common-tags`.

lib/modules/manager/npm/dep-types.ts
rhuanbarreto: Done. Reordered to `bun` > `pnpm` > `yarn`.

lib/modules/manager/npm/extract/common/catalogs.ts
rhuanbarreto: Done. Sorted constants, type union, and record keys alphabetically.

rhuanbarreto:
> are normal deps grouped by catalog name if they reference one? would make sense to me like we do for maven Gradle variables. otherwise each single update would always contain same changes?

Good point. Currently, catalog deps are extracted with `depType: 'bun.catalog.<name>'` (e.g. `bun.catalog.default`), which allows matching via `packageRules`. However, they aren't automatically grouped into a single PR the way maven/Gradle variables are.

For that automatic grouping behavior, I think we'd need something similar to how `groupName` is set for variable-based updates. Would you like me to look into wiring that up in this PR, or should that be a follow-up?

viceice:
> > are normal deps grouped by catalog name if they reference one? would make sense to me like we do for maven Gradle variables. otherwise each single update would always contain same changes?
> 
> Good point. Currently, catalog deps are extracted with `depType: 'bun.catalog.<name>'` (e.g. `bun.catalog.default`), which allows matching via `packageRules`. However, they aren't automatically grouped into a single PR the way maven/Gradle variables are.
> 
> For that automatic grouping behavior, I think we'd need something similar to how `groupName` is set for variable-based updates. Would you like me to look into wiring that up in this PR, or should that be a follow-up?

can probably be done in a follow up PR

rhuanbarreto:
Great! Now we just need your approval and merge @viceice. I will fix the follow up PR soon.

RahulGautamSingh (approved):

lib/modules/manager/bun/readme.md
```diff
@@ -2,3 +2,26 @@ Used for updating bun projects.
...
+  },
+  "catalogs": {
+    "testing": {
+      "jest": "30.0.0"
+    }
+  }
+}
+```
+
+Catalog dependencies produce dynamic `depType` values: `bun.catalog.default` for the default catalog, and `bun.catalog.<name>` for named catalogs.
```
viceice: this should be in dep-types file, see other managers for dynamic dep-types note

lib/modules/manager/bun/update.ts
```diff
@@ -0,0 +1,152 @@
...
+import { logger } from '../../../logger/index.ts';
+import { BUN_CATALOG_DEPENDENCY } from '../npm/extract/common/catalogs.ts';
+import {
+  getNewGitValue,
+  getNewNpmAliasValue,
+} from '../npm/update/dependency/common.ts';
+import { updateDependency as npmUpdateDependency } from '../npm/update/index.ts';
+import type { UpdateDependencyConfig } from '../types.ts';
+
+const bunCatalogRe = new RegExp(
```
viceice:
```suggestion
const bunCatalogRe = regEx(
```
use our reg helper

lib/modules/manager/npm/schema.ts
```diff
@@ -1,6 +1,26 @@
 import { z } from 'zod/v4';
 import { Json, LooseRecord, Yaml } from '../../../util/schema-utils/index.ts';
 
+export const BunCatalogs = z
```
viceice: move to bun manager, not used in npm manager

lib/modules/manager/npm/dep-types.ts
```diff
@@ -66,4 +66,4 @@ export const knownDepTypes = [
 ] as const satisfies readonly DepTypeMetadata[];
 
 export const supportsDynamicDepTypesNote =
-  'Additionally, catalog dependencies produce dynamic `depType` values: `pnpm.catalog.<name>` for [pnpm catalogs](https://pnpm.io/catalogs#defining-catalogs) and `yarn.catalog.<name>` for [yarn catalogs](https://yarnpkg.com/features/catalogs).';
+  'Additionally, catalog dependencies produce dynamic `depType` values: `bun.catalog.<name>` for [bun catalogs](https://bun.sh/docs/install/catalogs), `pnpm.catalog.<name>` for [pnpm catalogs](https://pnpm.io/catalogs#defining-catalogs), and `yarn.catalog.<name>` for [yarn catalogs](https://yarnpkg.com/features/catalogs).';
```
viceice: wrong, npm manager doesn't do it, only bun manager. so move to bun.

rhuanbarreto:
@RahulGautamSingh if merge queues are on, shouldn't you remove the config that offers updating the branch against main?

lib/modules/manager/bun/readme.md
rhuanbarreto: Done. Moved the dynamic depType documentation to a new `bun/dep-types.ts` file with `supportsDynamicDepTypesNote`, exported from `bun/index.ts`. Removed the manual paragraph from the readme since it will be auto-generated by the docs tooling.

lib/modules/manager/bun/update.ts
rhuanbarreto: Done. Switched to `regEx()` from `../../../util/regex.ts`.

lib/modules/manager/npm/schema.ts
rhuanbarreto: Done. Moved `BunCatalogs` zod schema to `bun/schema.ts`. The bun manager now imports it from `./schema.ts` instead of `../npm/schema.ts`.

lib/modules/manager/npm/dep-types.ts
rhuanbarreto: Done. Removed the bun catalog reference from the shared manager's `supportsDynamicDepTypesNote`. The bun manager now has its own `supportsDynamicDepTypesNote` in `bun/dep-types.ts`.

rhuanbarreto:
@viceice is there anything I should do to have this PR merged? Is renovate interested in this PR at all? We have experienced little support to make this PR reach the main branch. Looks like it was downprioritized just because I used AI to assist on the coding.

RahulGautamSingh:
> @viceice is there anything I should do to have this PR merged? Is renovate interested in this PR at all? We have experienced little support to make this PR reach the main branch. Looks like it was downprioritized just because I used AI to assist on the coding.

It is not down prioritized, we just are facing enormous load due to lots of PRs being created and small number of people working.

gologames:
Bump. I had to remove Bun catalogs from my project because Renovate cannot update them yet

lib/modules/manager/bun/update.ts
```diff
@@ -0,0 +1,151 @@
...
+  parsedContents: Record<string, unknown>,
+  catalogName: string,
+): Record<string, string>[] {
+  const targets: Record<string, string>[] = [];
+  const workspaces = parsedContents.workspaces;
+
+  if (catalogName === 'default') {
+    // Default catalog: look in `catalog` at top level and under `workspaces`
+    if (isPlainObject(parsedContents.catalog)) {
+      targets.push(parsedContents.catalog as Record<string, string>);
```
viceice: why do we need all those casts? arn't the types narrow enough? they should be fixed then instead of all those unsafe casts.

if those are really needed then add a comment why they are needed

viceice (commented):
otherwise lgtm

maybe needs merge from main if it's to far behind (because of recent lint changes)

lib/modules/manager/bun/update.ts
rhuanbarreto: Fixed. I'm hardening the PR to avoid anything that is against the patterns in the repo.

rhuanbarreto:
@viceice I implemented all fixes and did a cleanup step simplifying the code. Also implemented dep types. Everything is in sync and passing. Just need to merge.

lib/modules/manager/bun/dep-types.ts
```diff
@@ -0,0 +1,12 @@
+import { knownDepTypes as npmKnownDepTypes } from '../npm/dep-types.ts';
+
+/**
+ * Bun reads the same `package.json` fields as npm, except the pnpm-specific
+ * ones.
+ */
+export const knownDepTypes = npmKnownDepTypes.filter(
+  ({ depType }) => !depType.startsWith('pnpm'),
```
RahulGautamSingh, line 8: bun uses the shared `extractPackageJson`, which emits `pnpm` and `pnpm.overrides` depTypes for any `package.json`, so this filter hides two depTypes bun really extracts. 

Only `pnpm-workspace.overrides` is pnpm-only. 

Please list the depTypes literally as `as const satisfies readonly DepTypeMetadata[]`, like every other manager does, instead of deriving them by filter.

lib/modules/manager/bun/schema.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (defaultCatalog) {
+      result.push({ name: 'default', dependencies: defaultCatalog });
+    }
+    const namedCatalogs = coerceObject(catalogs ?? workspaces?.catalogs);
+    for (const [name, dependencies] of Object.entries(namedCatalogs)) {
+      result.push({ name, dependencies });
+    }
+    return result;
+  })
+  .catch([]);
```
RahulGautamSingh, line 36: this outer `.catch([])` swallows every catalog when a single value is malformed.

`YarnCatalogs` scopes `.catch()` to the one field that can be malformed. Narrow it the same way, or log at debug before returning the empty array.

lib/modules/manager/bun/update.ts
```diff
@@ -0,0 +1,82 @@
...
+import { regEx } from '../../../util/regex.ts';
+import { BUN_CATALOG_DEPENDENCY } from '../npm/extract/common/catalogs.ts';
+import {
+  getNewGitValue,
+  getNewNpmAliasValue,
+} from '../npm/update/dependency/common.ts';
+import { updateDependency as npmUpdateDependency } from '../npm/update/index.ts';
+import type { UpdateDependencyConfig } from '../types.ts';
+
+const bunCatalogRe = regEx(`^${BUN_CATALOG_DEPENDENCY}\\.(?<catalogName>.+)$`);
```
RahulGautamSingh, line 13: `BUN_CATALOG_DEPENDENCY` is `bun.catalog`, so the interpolated dot is a wildcard here. Escape it, or build the pattern from `bun\\.catalog`.

lib/modules/manager/bun/update.ts
```diff
@@ -0,0 +1,82 @@
...
+
+function getCatalog(scope: unknown, catalogName: string): unknown {
+  if (!isPlainObject(scope)) {
+    return null;
+  }
+  if (catalogName === 'default') {
```
RahulGautamSingh, line 19: a catalog named `default` under `catalogs` extracts as `bun.catalog.default` but is unreachable here, so the update logs "Bun catalog entry not found" and returns null. 

Fall back to `scope.catalogs[catalogName]` when `scope.catalog` does not hold the dep.

RahulGautamSingh (changes requested):
The PR description is outdatd. 

The "File changes" table still lists `bunCatalogsToArray()`, `npm/dep-types.ts`, and readme depType/`packageRules` docs, none of which are in the diff now, and it does not mention `bun/schema.ts` or `bun/dep-types.ts`. 

Please refresh it.

lib/modules/manager/bun/dep-types.ts
rhuanbarreto: Done in 8ca5f9c079: the depTypes are now listed literally with `as const satisfies readonly DepTypeMetadata[]`, including `pnpm` and `pnpm.overrides`; only `pnpm-workspace.overrides` is left out.

lib/modules/manager/bun/schema.ts
rhuanbarreto: Done in 8ca5f9c079: `catalog`, `catalogs` and `workspaces` each have their own `.catch()`, and both catalog levels use `LooseRecord`, so a malformed value only drops its own entry. The outer catch now only triggers for non-object input and logs at debug. Covered in the new `schema.spec.ts` and an extract test.

lib/modules/manager/bun/update.ts
rhuanbarreto: Done in 8ca5f9c079: the prefix is built with `RegExp.escape(BUN_CATALOG_DEPENDENCY)`, with a test that `bunXcatalog.default` is not treated as a catalog.

lib/modules/manager/bun/update.ts
rhuanbarreto: Done in 8ca5f9c079: for `default`, the update now checks `scope.catalog` and falls back to `scope.catalogs.default`, with a test covering it.

rhuanbarreto:
Thanks for the review @RahulGautamSingh! I refreshed the PR description (the "File changes" table now covers `bun/schema.ts` and `bun/dep-types.ts` and drops the stale rows), addressed the inline comments in 8ca5f9c079, and merged `main` to resolve the conflict.

PR #42945, opened 2026-04-27
feat(vulnerability): Add vulnerability alerts to deps-edn + lein

elliotcourant:
deps-edn and lein both essentially use maven upstream for several packages. Some packages come from clojars directly but others are typical java packages. If there is an OSV alert for a maven package however, it will not propagate to the deps-edn or lein dependency update from renovate. This tries to adjust that so that clojure can derive dependency alerts from maven as well so that those dependencies are treated the same way.

Not sure if this qualifies as a bugfix or a feature? I was unable to find a github issue for this at the time of opening it but I'm happy to make a github issue and/or a reproduction of the problem in another repo if needed.

## Changes

Allow maven vulnerability alerts to propagate to clojure's deps-edn and lein dependency managers so that those dependency updates will not just be tagged as regular updates but can be treated as security updates in general.

## Context

## AI assistance disclosure

- [x] Yes — other (please describe): I used AI to help research where to make changes in the code and to validate that there was not another way to achieve this behavior using the existing code and configuration changes.

The public repository: <URL>

Here is an example PR though of how the vulnerabilities do not propagate today: 
https://github.com/treasuryprime/com.treasuryprime.clj-cron-parse/pull/11 

That PR contains an update for Clojure that references https://github.com/advisories/GHSA-vr64-r9qj-h27f but because the PR is for deps-edn only it does not tag the PR as security.

### DIFF
--- a/lib/util/vulnerability/ecosystem.ts
+++ b/lib/util/vulnerability/ecosystem.ts
@@ -22,9 +22,8 @@ export const datasourceToOsvEcosystem: Record<
   string,
   OsvEcosystem | undefined
 > = {
-  // ClojureDatasource extends MavenDatasource and OSV indexes Clojars and
-  // Maven Central artifacts under the same Maven ecosystem, so deps.edn and
-  // Leiningen packages share the maven entry below.
+  // ClojureDatasource extends MavenDatasource and OSV indexes Clojars and Maven Central artifacts under the same Maven
+  // ecosystem, so deps.edn and Leiningen packages share the maven entry below.
   [ClojureDatasource.id]: 'Maven',
   [CrateDatasource.id]: 'crates.io',
   [GoDatasource.id]: 'Go',
--- a/lib/util/vulnerability/ecosystem.ts
+++ b/lib/util/vulnerability/ecosystem.ts
@@ -43,14 +43,14 @@ export const datasourceToOsvEcosystem: Record<
  *
  * @see https://docs.github.com/en/rest/dependabot/alerts
  */
-export const githubEcosystemToDatasource: Record<GithubEcosystem, string> = {
-  actions: GithubTagsDatasource.id,
-  composer: PackagistDatasource.id,
-  go: GoDatasource.id,
-  maven: MavenDatasource.id,
-  npm: NpmDatasource.id,
-  nuget: NugetDatasource.id,
-  pip: PypiDatasource.id,
-  rubygems: RubygemsDatasource.id,
-  rust: CrateDatasource.id,
+export const githubEcosystemToDatasource: Record<GithubEcosystem, string[]> = {
+  actions: [GithubTagsDatasource.id],
+  composer: [PackagistDatasource.id],
+  go: [GoDatasource.id],
+  maven: [MavenDatasource.id, ClojureDatasource.id],
+  npm: [NpmDatasource.id],
+  nuget: [NugetDatasource.id],
+  pip: [PypiDatasource.id],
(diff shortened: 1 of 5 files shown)


lib/workers/repository/init/vulnerability.ts
```diff
@@ -126,8 +127,15 @@ export async function detectVulnerabilityAlerts(
       } catch (err) /* v8 ignore next */ {
         logger.warn({ err }, 'Error generating vulnerability PR notes');
       }
+      // GitHub reports JVM advisories under the maven ecosystem, but Renovate tags deps.edn and Leiningen dependencies
+      // with the clojure datasource. Widen the rule so a maven advisory also applies to those managers.
+      let matchDatasources = [datasource];
+      if (datasource === MavenDatasource.id) {
+        matchDatasources = [MavenDatasource.id, ClojureDatasource.id];
+      }
```
RahulGautamSingh: Optional style nit: this could be a const + ternary to avoid the `let`/reassignment:

```ts
const matchDatasources =
  datasource === MavenDatasource.id
    ? [MavenDatasource.id, ClojureDatasource.id]
    : [datasource];
```

Not blocking.

RahulGautamSingh (commented):
LGTM

Typo PR title:  **deps-end**  should be **deps-edn**

lib/workers/repository/init/vulnerability.ts
elliotcourant: I can tweak this shortly!

lib/util/vulnerability/ecosystem.ts
viceice: this should be updated to support multiple datasources.

lib/workers/repository/init/vulnerability.ts
```diff
@@ -126,8 +127,15 @@ export async function detectVulnerabilityAlerts(
       } catch (err) /* v8 ignore next */ {
         logger.warn({ err }, 'Error generating vulnerability PR notes');
       }
+      // GitHub reports JVM advisories under the maven ecosystem, but Renovate tags deps.edn and Leiningen dependencies
+      // with the clojure datasource. Widen the rule so a maven advisory also applies to those managers.
+      const matchDatasources =
+        datasource === MavenDatasource.id
+          ? [MavenDatasource.id, ClojureDatasource.id]
+          : [datasource];
+
```
viceice: see above, that would make future changes more simple

lib/util/vulnerability/ecosystem.ts
elliotcourant: At face value, changing this to be a map with an array value seems trivial. However this means that the opposite must be true as well? Going from datasource to ecosystem would also be a one to many, I'll try to poke around and see if there is a tidy way to do that without blowing up the diff.

elliotcourant: Okay maybe I was overthinking this, I made a change that I think will work. I'll push shortly after updating tests.

elliotcourant: Okay updated! The diff is a bit bigger now but this is now a proper one to many or many to one in both directions.

elliotcourant:
I'm not entirely sure why the CI failed but I will try to run things locally again to see if I broke something? The test suites I modified passed and surrounding ones too. So I'm not sure here.

viceice (approved):

RahulGautamSingh (approved):

elliotcourant:
Is there anything on this still waiting on me?

PR #43021, opened 2026-04-30
fix(log): ensure that log output is always flushed

jamietanna:
## Changes

In cases where the process ends very quickly after a log line is
written, it could lead to cases where that log line doesn't appear on
`stdout`.

We can make sure that we always explicitly call the callback.

This only affected the "pretty" output in the CLI, and only when log
lines were written fairly close to process end.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/logger/pretty-stdout.ts
+++ b/lib/logger/pretty-stdout.ts
@@ -132,6 +132,7 @@ export class RenovateStream extends Writable {
     _encoding: string,
     callback: (error?: Error | null) => void,
   ): void {
-    this.destination.write(formatRecord(data, this.colorize), callback);
+    this.destination.write(formatRecord(data, this.colorize));
+    callback();
   }
 }


viceice (approved):
🤷‍♂️ looks like i suggested to pass the callback to the destination, which doesn't work with `process.stdout`

jamietanna:
Now we know 😁 At least it's likely not affecting too many logs

felipecrs:
Thanks a lot for fixing it!

jamietanna:
This might not work for log files - needs some more testing

RahulGautamSingh (approved):

viceice (changes requested):
it breaks pretty file logger

jamietanna:
It looks like the pretty logger didn't work with the previous changes either - the `WARN` doesn't show - @viceice I'd say let's get this fixed for the more important use-case (the default logger) then let's follow-up for the pretty file logger

viceice (approved):
ok

jamietanna:
@felipecrs if you're able to take a look at that, it'd be appreciated - I'm low on bandwidth right now, and you're the only user :)

felipecrs:
> @felipecrs if you're able to take a look at that, it'd be appreciated - I'm low on bandwidth right now, and you're the only user :)

Of course! :)

felipecrs:
> @felipecrs if you're able to take a look at that, it'd be appreciated - I'm low on bandwidth right now, and you're the only user :)

I was unable to reproduce any issue with the pretty file logger in fact, even after this PR. 

I think we're good. :)

> it breaks pretty file logger

@viceice were you actually able to reproduce the issue or did you just assume? If yes, can you please share more details on how you reproduced it?

jamietanna:
I could reproduce the `LOG_FILE_FORMAT=pretty` not showing the validation error in the log, when testing against https://github.com/containerbase/.github/pull/3243 and running:

```
env LOG_FILE_FORMAT=pretty LOG_FILE=debug.log LOG_LEVEL=debug node ~/workspaces/renovate/lib/config-validator.ts --no-global --strict default.json
```

In this case, I don't see the `WARN` in the log file

We've seen that it either does/does not reproduce depending on local / whether it's running in CI - so very likely to be a timing issue / due to how buffers may (not) get flushed

felipecrs:
Awesome, I was able to reproduce the issue as well with the config validator. Thanks.

PR #43033, opened 2026-05-01
feat(config/validation): validate object entries in the `repositories` array

felipecrs:
## Changes

This extends the config-validator to also validate the repository configs defined as object entries in the [`repositories` array](https://docs.renovatebot.com/self-hosted-configuration/#repositories), which was previously not being validated.

Also fixes the `RenovateRepository` repository type to allow global-only properties ([ref](https://github.com/renovatebot/renovate/pull/43033#discussion_r3172633038)).

## Context

## AI assistance disclosure

Claude Opus 4.6 was used to generate the initial implementation of this feature, which was then reviewed and iterated several times by me.

The public repository: <URL>


lib/config/validation.spec.ts
```diff
@@ -2411,6 +2411,98 @@ describe('config/validation', () => {
...
     });
 
+    it('validates repositories[] object-entry config', async () => {
+      const config = {
+        repositories: [
+          'owner/repo1',
+          {
+            repository: 'owner/repo2',
+            extends: ['config:recommended'],
+            binarySource: 'global', // should not allow globalOnly options inside repository config
```
jamietanna: ⚠️ this is not true - this is intended behaviour - we would expect that global self-hosted config can specify `repositories`, and then set repo-specific global and repo-level config

jamietanna (changes requested):

lib/config/validation.spec.ts
felipecrs: Wow, I didn't know that, but that's indeed how the runtime behaves.

I will also fix the typing, as it's incorrectly flagging global-only stuff in the repository object-entry.

felipecrs: Thanks!

lib/config/validation.ts
```diff
@@ -1070,6 +1074,30 @@ async function validateGlobalConfig(
             }
           }
         }
+        if (key === 'repositories') {
+          for (const [subIndex, subval] of val.entries()) {
+            if (isPlainObject(subval)) {
+              if (!isNonEmptyString(subval.repository)) {
+                warnings.push({
```
viceice:
```suggestion
                errors.push({
```

isn't this a hard error?

lib/config/validation.ts
```diff
@@ -1070,6 +1074,30 @@ async function validateGlobalConfig(
...
+                'global',
+                repoEntryConfig,
+                false,
+                `${currentPath}[${subIndex}]`,
+              );
+              for (const warning of subValidation.warnings.concat(
+                subValidation.errors,
+              )) {
+                warnings.push(warning);
+              }
```
viceice:
```suggestion
              warnings.push(...subValidation.warnings);
              errors.push(...subValidation.errors);
```

should we push errors to errors? 

i see similar code below 🤔 @jamietanna @secustor WDYT?

lib/config/validation.ts
felipecrs: Definitely. Thanks.

lib/config/validation.ts
felipecrs: Yeah. I think so too. The similar code places are:

https://github.com/renovatebot/renovate/blob/60efe09c11c222c5c8c0c9442a9ea504858e1e28/lib/config/validation.ts#L372-L383

https://github.com/renovatebot/renovate/blob/60efe09c11c222c5c8c0c9442a9ea504858e1e28/lib/config/validation.ts#L847-L856

So, I've changed it. Thanks!

lib/config/validation.ts
```diff
@@ -1108,16 +1112,18 @@ async function validateGlobalConfig(
...
-          if (isObject(subval)) {
-            const subValidation = await validateConfig(
-              'global',
-              subval as AllConfig,
-              false,
-              `${currentPath}[${subIndex}]`,
-            );
-            warnings.push(...subValidation.warnings);
-            errors.push(...subValidation.errors);
+        if (key !== 'repositories') {
```
RahulGautamSingh: why not just, to prevent extra nesting

` if (isArray(val) && key !== 'repositories') {`

RahulGautamSingh (changes requested):

lib/config/validation.ts
felipecrs: Absolutely. That's a great idea. Thank you!

Done in https://github.com/renovatebot/renovate/pull/43033/commits/7887365787f9a7cecf8957c0b18ec0551a6bca84.

RahulGautamSingh (approved):

felipecrs:
@jamietanna, just FYI, I've merged main because new lint rules were introduced and I was wondering if they would catch issues here (so that I can fix them before they fail in merge queue).

lib/config/validation.ts
```diff
@@ -1167,6 +1171,26 @@ async function validateGlobalConfig(
             }
           }
         }
+      } else if (isArray(val)) {
+        for (const [subIndex, subval] of val.entries()) {
+          if (isPlainObject(subval)) {
```
jamietanna, line 1178: Via Claude Opus 4.8 - should there be an `else` here? I.e. to confirm it's an array of strings?

lib/config/validation.ts
```diff
@@ -94,7 +94,6 @@ const ignoredNodes = [
   'npmToken',
   'packageFile',
   'forkToken',
-  'repository',
```
jamietanna, line 97: Via Claude Opus 4.8, this may lead to issues in i.e. self-hosted configuration that specifies `repository`, which may not be intentional?

jamietanna (dismissed):

felipecrs:
Thanks a lot for fixing it up, @jamietanna!

viceice (approved):

lib/config/validation.ts
```diff
@@ -1167,6 +1173,46 @@ async function validateGlobalConfig(
...
+              false,
+              `${currentPath}[${subIndex}]`,
+            );
+            warnings.push(...subValidation.warnings);
+            errors.push(...subValidation.errors);
+          } else if (isString(subval)) {
+            if (!isNonEmptyStringAndNotWhitespace(subval)) {
+              warnings.push({
+                topic: 'Configuration Error',
+                message: `${currentPath}[${subIndex}]: each repository string entry entry must be a non-empty string`,
```
jamietanna, line 1206: Woops, typo

PR #43035, opened 2026-05-01
refactor: centralize SimpleGit creation

reduckted:
## Changes

This change moves the creation of `SimpleGit` instances into a single function. This aims to make the changes required for #43122 easier to manage by having one place to supply the `unsafe` options, and one place where the environment variables are built.

To help enforce this, I've extended the `no-restricted-imports` lint rule to restrict the import of `simpleGit` from the `simple-git` module, except in tests, tools, and the file that contains the new central function.

~One change that could be considered "breaking" is the `initRepo` function. It previously used `getEnv()` to get the environment variables (that uses `process.env`, which is unsafe). Due to the changes, this will now use `getChildEnv()`.~

## Context

See https://github.com/renovatebot/renovate/pull/42738#issuecomment-4351257205

## AI assistance disclosure

The public repository: N/A

I haven't added any _new_ tests, however the existing tests for consumers of this new function have been updated.


lib/util/git/index.ts
```diff
@@ -254,12 +272,10 @@ export async function initRepo(args: StorageConfig): Promise<void> {
   config.ignoredAuthors = [];
   config.additionalBranches = [];
   config.branchIsModified = {};
-  // TODO: safe to pass all env variables? use `getChildEnv` instead?
   git = instrumentGit(
-    simpleGit(GlobalConfig.get('localDir'), simpleGitConfig()).env({
-      ...getEnv(),
-      LANG: 'C.UTF-8',
-      LC_ALL: 'C.UTF-8',
+    createSimpleGit({
```
viceice: we'll do #43113 first

viceice:
can you please update to latest main? conflicted 🙃

reduckted:
There's a `GIT_SSH_COMMAND` environment variable that is set globally.
https://github.com/renovatebot/renovate/blob/6221de715b58d323e9dd246addb40656b95a4a83/lib/modules/datasource/git-refs/index.ts#L13

Does that still need to happen? If all SimpleGit creation is now using `getChildEnv`, then I don't think Git ever sees that environment variable (unless it's used for something else that I'm missing).

viceice:
> There's a `GIT_SSH_COMMAND` environment variable that is set globally.
> https://github.com/renovatebot/renovate/blob/6221de715b58d323e9dd246addb40656b95a4a83/lib/modules/datasource/git-refs/index.ts#L13
> 
> Does that still need to happen? If all SimpleGit creation is now using `getChildEnv`, then I don't think Git ever sees that environment variable (unless it's used for something else that I'm missing).

I think we should move that to the create simple git call. 🤔

lib/util/git/index.ts
```diff
@@ -255,15 +288,11 @@ export async function initRepo(args: StorageConfig): Promise<void> {
...
-          // TODO: Do we really need to set these?
-          LANG: 'C.UTF-8',
-          LC_ALL: 'C.UTF-8',
-        },
-      }),
-    ),
+    createSimpleGit({
+      config: { baseDir: GlobalConfig.get('localDir') },
+      // TODO: Do we really need to set these?
+      env: { LANG: 'C.UTF-8', LC_ALL: 'C.UTF-8' },
```
viceice: can now removed?

lib/util/git/index.ts
```diff
@@ -255,15 +288,11 @@ export async function initRepo(args: StorageConfig): Promise<void> {
...
-        env: {
-          // TODO: Do we really need to set these?
-          LANG: 'C.UTF-8',
-          LC_ALL: 'C.UTF-8',
-        },
-      }),
-    ),
+    createSimpleGit({
+      config: { baseDir: GlobalConfig.get('localDir') },
+      // TODO: Do we really need to set these?
```
viceice: also remove

lib/util/git/index.ts
reduckted: Fixed.

lib/util/git/index.ts
reduckted: Fixed.

viceice (approved):

lib/util/git/index.ts
```diff
@@ -255,15 +288,7 @@ export async function initRepo(args: StorageConfig): Promise<void> {
   config.additionalBranches = [];
   config.branchIsModified = {};
   git = instrumentGit(
-    simpleGit(GlobalConfig.get('localDir'), simpleGitConfig()).env(
-      getChildEnv({
-        env: {
-          // TODO: Do we really need to set these?
```
jamietanna, line 261: Thanks for working out why we set these!

jamietanna (approved):
Nice, thank you!

PR #43058, opened 2026-05-02
fix(manager/deno): set `--frozen=false` default to update lock file correctly

Hajime-san:
## Changes

Described:
- https://github.com/renovatebot/renovate/discussions/43055
- https://github.com/renovatebot/renovate/discussions/43063
- https://github.com/renovatebot/renovate/discussions/44478

Deno is the one of very few package manager that have frozen lock file setting in config. ~~Renovate should not occurs an error at the step of updating artifact when it specified `frozen: true`.~~

Renovate should always override the `frozen` flag due to if it would be specified `true`.

## Context

## AI assistance disclosure

The public repository: <URL>


lib/modules/manager/deno/artifacts.ts
```diff
@@ -90,6 +103,8 @@ export async function updateArtifacts(
 
     // "deno install" don't execute lifecycle scripts of package.json by default
     // https://docs.deno.com/runtime/reference/cli/install/#native-node.js-addons
+    // TODO: appending `--lockfile-only` is better
+    // https://docs.deno.com/runtime/reference/cli/install/#options-lockfile-only
     await exec('deno install', execOptions);
```
levibostian: Curious, is this TODO comment proposing that this `deno install` command gets modified to handle when the lock file is frozen? 

I ask because after reviewing the code, I agree, modifying the `deno install` command would be required if the lock file is frozen, otherwise you would still encounter the error *The lockfile is out of date. Run `deno install --frozen=false`, or rerun with `--frozen=false` to update it.*. 

I personally believe that this PR implementation is incomplete without modifying this command in some way.

Hajime-san: > Curious, is this TODO comment proposing that this deno install command gets modified to handle when the lock file is frozen?

Nope. It's just a disk efficiency flag.
- https://github.com/denoland/deno/pull/31376

> I ask because after reviewing the code, I agree, modifying the deno install command would be required if the lock file is frozen, otherwise you would still encounter the error The lockfile is out of date. Run deno install --frozen=false, or rerun with --frozen=false to update it..

IMO, I think Renovate should respect the settings in deno.json as they are as described in the following discussion. Therefore, it will not update the lock file unilaterally.
I think users could customize an extra setting for the deno manager in renovate.json if they want to update the lock file only when Renovate is running. For example, it would ignore frozen: true and update the lock file if there is a setting like [lockFileMaintenance](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance) due to this is written according to the user's wishes.

However, this is not a definitive statement since I am not a maintainer.

ref: 
- https://github.com/renovatebot/renovate/discussions/43055#discussioncomment-16789297

levibostian: Oh, I didn't realize that's how the lock file maintenence feature worked. 

Makes sense, thanks for explaining this!

viceice (commented):
conflicted

Hajime-san:
@viceice 
How do you think about the direction of this PR?
I think its reasonable that an extra settings for controlling the behaviour of lock file like below.

```ts
type denoFrozenLockfileMode = "respect" | "force-frozen" | "off"
```

```json
{
  "packageRules": [
    {
      "matchManagers": ["deno"],
      "denoFrozenLockfile": "off"
    }
  ]
}
```

However, this PR should close if the way of handling lock file is not an appropriate for maintainers opinion. And it would be better to open a discussion.

- https://github.com/renovatebot/renovate/discussions/43063
- https://github.com/renovatebot/renovate/discussions/43055

lib/modules/manager/deno/types.ts
```diff
@@ -7,6 +7,11 @@ export interface DenoManagerData extends Record<string, any> {
    * The file that refers to this import map, typically deno.json or deno.jsonc
    */
   importMapReferrer?: string;
+  /**
+   * Whether the lockfile is frozen. It need to skip updates to the lockfile on
+   * updating artifact step if it set to `true`.
+   */
+  frozenLockfile?: boolean;
```
viceice: how would the lockfile be updated when renovate doesn't ignore this flag. I'm pretty sure we use some flag to override on other npm managers

viceice (commented):
this will always cause lockfile errors because of mismatches between updated drops and the lockfile.

btw we need to validate the lockfile path be relative to the repo if path is changed.

lib/modules/manager/deno/types.ts
Hajime-san: It seems that yarn always overrides the frozen flag in config.

https://github.com/renovatebot/renovate/blob/766db6fc6f09592403dd21aec13138fe3f70cdd3/lib/modules/manager/npm/post-update/yarn.ts#L184
https://github.com/renovatebot/renovate/blob/766db6fc6f09592403dd21aec13138fe3f70cdd3/lib/modules/manager/npm/post-update/yarn.ts#L67-L69

We should always specify `--frozen=false`, and we don't need to add any variables to `DenoManagerData` if we follow this behavior.

In this case, the premise of this PR is invalidated. Would it be more appropriate to create a new PR instead of continuing here?

viceice: you can change this pr, so we don't loose history

lib/modules/manager/deno/artifacts.ts
```diff
@@ -95,12 +91,13 @@ export async function updateArtifacts(
...
     // TODO: appending `--lockfile-only` is better to reduce disk usage
     // https://docs.deno.com/runtime/reference/cli/install/#options-lockfile-only
-    let command = 'deno install';
-    if (args) {
-      command += args;
-    }
+    // deno.json(c) could have the `lock.frozen` field
+    // we should always override the `frozen` flag due to if it would be specified true
+    const args = ' --frozen=false';
+    command += args;
```
viceice: inline to a single string makes no sense to use a separate args variable here

viceice (approved):

Aire-One:
Hello, gentle ping because it seems the auto-merge didn't work.

RahulGautamSingh:
@Hajime-san can you update this branch so it's on track with main and if applicable resolve the review comment from @levibostian 

Not sure why automerge stalled :)

lib/modules/manager/deno/artifacts.ts
```diff
@@ -67,12 +67,8 @@ export async function updateArtifacts(
   try {
     await writeLocalFile(packageFileName, newPackageFileContent);
 
-    let args = '';
```
watsonarw, line 83: One thing to note when updating this branch, this will likely cause a subtle merge issue when updating due to the `additionalHostImports` logic [using the same pattern of adding args](https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/deno/artifacts.ts#L123-L130) as was removed in this PR.

I noticed it when raising https://github.com/renovatebot/renovate/pull/44480 before I knew about this PR.

viceice (approved):

viceice:
> Not sure why automerge stalled :)

because there was an old open conversation

Hajime-san:
@5ouma @levibostian @watsonarw 
Hi, sorry for confusing, give it a try newer release that was marged by team which allows to set deno config `"lock": { "frozen": true }` without an error.

Issue #41879, opened 2026-03-13
Add user-facing documentation for `BranchResult`s and `SkipReason`s


jamietanna:
### Describe the proposed change(s).

As a follow-up to https://github.com/renovatebot/renovate/issues/38851 and https://github.com/renovatebot/renovate/issues/41878

Relates to https://github.com/renovatebot/renovate/issues/6916

PR #43109, opened 2026-05-05
fix(types): generate and reuse types from config options

felipecrs:
## Changes

I noticed that `config-validator` was not flagging `goModTidy` (instead of `gomodTidy`) as an invalid config option, even though the JSON schema was flagging it as invalid.

There is a PR to fix the `config-validator` (https://github.com/renovatebot/renovate/pull/43110), but I also noticed that TypeScript was not flagging the same issue because it was simply not typed.

So I had this idea: what if we generate all the types from the config options, which is used by both the documentation and the JSON schema? This way, we can ensure that all three are always in sync, and we don't need to maintain the types manually.

This PR is relatively simple, but it changes a lot of files because it updates all files to use the generated types instead of the former ones. It also fixes some small type issues that I noticed in the process, besides removing many redundant types that are not needed anymore.

I could do it in a backwards compatible manner by aliasing the old types to the new ones, but I don't think we are considering these types as part of our public API, are we?

## Context

## AI assistance disclosure

Claude Sonnet 4.6 was employed to implement the idea and to scan for all the types that needed to be updated.

The public repository: <URL>


viceice (commented):
please alias to reduce code changes.

types are used by mend tools, do we should not break exports if not required

jamietanna:
This change does on the whole look reasonable (at a glance, I've not reviewed it in full) but we should also consider how this will affect https://github.com/renovatebot/renovate/issues/41878 and https://github.com/renovatebot/renovate/issues/41879 and related work we're looking at to improve the documentation

felipecrs:
> please alias to reduce code changes.
> 
> types are used by mend tools, do we should not break exports if not required

Done. I've kept all existing type exports and reverted the refactorings related to that.

PR #43111, opened 2026-05-06
fix(versioning/github-actions): ensure stable sorting for `v` prefix and short versions

viceice:
## Changes
Ensure stable sorting of versions:
- `v6 < v6.0 < v6.0.0`
- `1 < 1.0 < 1.0.0`
- `1.0.0 < v1.0.0`  
- `1 < v1 < 1.0 < v1.0 < 1.0.0 < v1.0.0`

## Context
- https://github.com/renovatebot/osv-offline/pull/1767

## AI assistance disclosure

The public repository: <URL>


lib/modules/versioning/github-actions/index.spec.ts
```diff
@@ -373,17 +373,18 @@ describe('modules/versioning/github-actions/index', () => {
       ${'v1.0.0'}  | ${'v1.0.0'}  | ${0}
       ${'v1.0.0'}  | ${'v1.0.1'}  | ${-1}
       ${'v1.0.1'}  | ${'v1.0.0'}  | ${1}
-      ${'v1.0.0'}  | ${'1.0.0'}   | ${0}
-      ${'1.0.0'}   | ${'v1.0.0'}  | ${0}
+      ${'v1.0.0'}  | ${'1.0.0'}   | ${1}
```
jamietanna, line 376: If we stripped the `v` when we did the compare, would this lead to "flip flopping" between the two? (as you'd hinted at on Slack)

viceice: yes, it then depends on the order of tags returned by the datasource.
if we don't do it we could get updates like this: `1.0.0 > v1.0.1 -> 1.2.0 > ...`

viceice: we only change the sort order, so `v1` is still equal to `1.0.0` and so on.

jamietanna (approved):

PR #40581, opened 2026-01-21
fix(git): correctly merge existing `GIT_CONFIG_` variables if set

jamietanna:
## Changes

As noted in #40580, we have cases where the `GIT_CONFIG` environment
variables may be set by a user, and Renovate doesn't correctly merge
them into the resulting variables, although it does use the existing
`GIT_CONFIG_COUNT` as its starting point for new variables.

To correctly handle this, we should make sure that if there are existing
variables set, we copy them over.

If the `GIT_CONFIG_COUNT` is set, but some value(s) are missing, we
should make sure to log a warning and default them to the empty string.

It's likely this may break Git, but be a more reasonable error than the
current behaviour.

## Context

Please select one of the below:

- [x] This closes an existing Issue: Closes #40580
- [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

The public repository: https://github.com/JamieTanna-Mend-testing/issue-40575

PR #43113, opened 2026-05-06
fix(git): pass only allowed child env

viceice:
## Changes
- only pass allowed child env to repo git commands
- separated from (#43035)
- probably breaking for users who expect git config / ssh agent env vars to be passed by default

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/util/git/index.ts
+++ b/lib/util/git/index.ts
@@ -181,7 +181,7 @@ export const GIT_MINIMUM_VERSION = '2.33.0'; // git show-current
 
 export async function validateGitVersion(): Promise<boolean> {
   let version: string | undefined;
-  const globalGit = instrumentGit(simpleGit());
+  const globalGit = instrumentGit(simpleGit().env(getChildEnv()));
   try {
     const { major, minor, patch, installed } = await globalGit.version();
     /* v8 ignore if -- TODO: add test #40625 */
@@ -254,13 +254,16 @@ export async function initRepo(args: StorageConfig): Promise<void> {
   config.ignoredAuthors = [];
   config.additionalBranches = [];
   config.branchIsModified = {};
-  // TODO: safe to pass all env variables? use `getChildEnv` instead?
   git = instrumentGit(
-    simpleGit(GlobalConfig.get('localDir'), simpleGitConfig()).env({
-      ...getEnv(),
-      LANG: 'C.UTF-8',
-      LC_ALL: 'C.UTF-8',
-    }),
+    simpleGit(GlobalConfig.get('localDir'), simpleGitConfig()).env(
+      getChildEnv({
+        env: {
+          // TODO: Do we really need so test these?
+          LANG: 'C.UTF-8',
+          LC_ALL: 'C.UTF-8',
+        },
+      }),
+    ),
   );
   gitInitialized = false;
   submodulesInitizialized = false;
--- a/lib/util/git/index.ts
+++ b/lib/util/git/index.ts
@@ -258,7 +258,7 @@ export async function initRepo(args: StorageConfig): Promise<void> {
     simpleGit(GlobalConfig.get('localDir'), simpleGitConfig()).env(
       getChildEnv({
         env: {
-          // TODO: Do we really need so test these?
+          // TODO: Do we really need to set these?
           LANG: 'C.UTF-8',
           LC_ALL: 'C.UTF-8',
         },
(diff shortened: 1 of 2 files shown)


jamietanna:
Yeah so this will break https://github.com/renovatebot/renovate/pull/40581 right?

jamietanna (dismissed):
I think this is reasonable - would also like @secustor's review before we merge

I think we should consider what this means for https://github.com/renovatebot/renovate/pull/40581 and what the migration for those users will be

viceice:
> I think this is reasonable - would also like @secustor's review before we merge

sure

> I think we should consider what this means for #40581 and what the migration for those users will be

they need to configure `customEnvVariables`[^1], so should be easy to transition if needed.

eg something like:
```js
const customEnvVariables = {
  'PAGER': 'less', // i don't think this should ever be passed to renovate
  'GIT_SSH_COMMAND': 'ssh -o BatchMode=yes',
};

for ([k,v] of Object.entries(process.env)) {
  if(/^GIT_CONFIG_/.test(k)) {
    customEnvVariables[k] = v
  }
}

module.exports = {
  // other config
  customEnvVariables,
};
```

[^1]: <https://docs.renovatebot.com/self-hosted-configuration/#customenvvariables>

lib/util/git/index.ts
```diff
@@ -254,13 +254,16 @@ export async function initRepo(args: StorageConfig): Promise<void> {
...
   git = instrumentGit(
-    simpleGit(GlobalConfig.get('localDir'), simpleGitConfig()).env({
-      ...getEnv(),
-      LANG: 'C.UTF-8',
-      LC_ALL: 'C.UTF-8',
-    }),
+    simpleGit(GlobalConfig.get('localDir'), simpleGitConfig()).env(
+      getChildEnv({
+        env: {
+          // TODO: Do we really need so test these?
```
reduckted:
```suggestion
          // TODO: Do we really need to set these?
```

reduckted: Looking at the history, I think the answer is yes, though it could definitely benefit from a comment explaining why.

https://github.com/renovatebot/renovate/commit/f31dacb8cd12fa747784387bd4bf25f2ecd96280#diff-3c967a96825db779ae3fa03f549890aa993ea47b1b3326e800e7a582f1b5b351

https://github.com/renovatebot/renovate/pull/17849

https://github.com/steveukx/git-js/blob/1bb14df0595794a9353d28ccdaeeb06c0b9bf2a5/docs/NON_ENGLISH_LOCALE.md

reduckted: I'm not sure why it's _only_ set here though. Maybe it should be set everywhere that a `SimpleGit` instance is created. I could add that to #43035.

secustor (dismissed):

lib/util/git/index.ts
viceice: please add it there

jamietanna (approved):

viceice:
merging, because @secustor approved previously and i only changed a comment and tests

rngcntr:
Just to let you know, this update broke the entire Renovate processing within our company :sweat_smile: 

We were using the configuration `gitAuthor: "name-of-the-bot@name-of-the-company.com"` and since we updated Renovate `v43.166.3`, all proposed upates that Renovate made were silently aborted with the log message: 
```
fatal: empty ident name (for <name-of-the-bot@name-of-the-company.com>)
```

Now the easy fix was to switch to `gitAuthor: "Name of the bot <name-of-the-bot@name-of-the-company.com>"` but we only noticed that by pure luck and did not expect a breaking change like that in a patch update.

mahmoud-mahdi:
> Just to let you know, this update broke the entire Renovate processing within our company 😅
> 
> We were using the configuration `gitAuthor: "name-of-the-bot@name-of-the-company.com"` and since we updated Renovate `v43.166.3`, all proposed upates that Renovate made were silently aborted with the log message:
> 
> ```
> fatal: empty ident name (for <name-of-the-bot@name-of-the-company.com>)
> ```
> 
> Now the easy fix was to switch to `gitAuthor: "Name of the bot <name-of-the-bot@name-of-the-company.com>"` but we only noticed that by pure luck and did not expect a breaking change like that in a patch update.

Same by us, after the update renovate does not considre the variables GIT_SSL_CAINFO and NODE_EXTRA_CA_CERTS anymore and we are getting certificate error:

`"message":"fatal: unable to access 'https://bitbucket.example.com/scm/iacks/argocd-apps.git/': server certificate verification failed. CAfile: none CRLfile: none\n","stack":"Error: fatal: unable to access `

viceice:
please open a new discussion what variables you are using any why. thfn reference this PR.


this change was made to harden renovate and it was never expected that the full env is passed to git for security reasons.

viceice:
`GIT_SSL_CAINFO` can probably be allowed to pass through, also the node specific variant

PR #43129, opened 2026-05-06
fix(platform): preserve explicit local dryRun

zeitlinger:
## Changes

When the local platform initializer is called with an explicit `dryRun` value, preserve it instead of always overwriting with `lookup`. Only `extract` is preserved — anything else (including `full`, which would error against the local platform) falls back to `lookup`.

## Context

## AI assistance disclosure

Used Claude Code for the initial diagnosis, code change, and test additions.

### DIFF
--- a/lib/modules/platform/local/index.ts
+++ b/lib/modules/platform/local/index.ts
@@ -10,9 +10,9 @@ import type {
 export const id = 'local';
 export const experimental = true;
 
-export function initPlatform(_params: PlatformParams): Promise<PlatformResult> {
+export function initPlatform(params: PlatformParams): Promise<PlatformResult> {
   return Promise.resolve({
-    dryRun: 'lookup',
+    dryRun: params.dryRun ?? 'lookup',
     endpoint: 'local',
     persistRepoData: true,
     requireConfig: 'optional',
--- a/lib/modules/platform/local/index.ts
+++ b/lib/modules/platform/local/index.ts
@@ -11,8 +11,9 @@ export const id = 'local';
 export const experimental = true;
 
 export function initPlatform(params: PlatformParams): Promise<PlatformResult> {
+  const dryRun = params.dryRun === 'extract' ? 'extract' : 'lookup';
   return Promise.resolve({
-    dryRun: params.dryRun ?? 'lookup',
+    dryRun,
     endpoint: 'local',
     persistRepoData: true,
     requireConfig: 'optional',
--- a/lib/modules/platform/local/index.spec.ts
+++ b/lib/modules/platform/local/index.spec.ts
@@ -13,13 +13,26 @@ describe('modules/platform/local/index', () => {
       `);
     });
 
-    it('preserves an explicit dryRun override', async () => {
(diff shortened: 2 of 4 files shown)


jamietanna:
I'm not sure I've seen this behaviour before - please share a case where you're seeing this happen (and the version of Renovate)

Also, when raising PRs, please follow our PR template

zeitlinger:
Observed with Renovate `43.141.6`, and `main` still appears to behave the same in `lib/modules/platform/index.ts` + `lib/modules/platform/local/index.ts`.

Concrete case:

```bash
LOG_LEVEL=debug LOG_FORMAT=json \
RENOVATE_CONFIG_FILE=.github/renovate.json5 \
renovate --platform=local --require-config=ignored --dry-run=extract
```

Expected: effective `dryRun=extract`, so repository processing stops after extraction.

Actual on the affected versions: local platform init sets `dryRun: 'lookup'`, and `initPlatform(config)` merges `...config, ...platformInfo`, so the explicit CLI/config `dryRun=extract` is overwritten. In the resulting `Combined config` log, `dryRun` is `lookup`, and repository processing continues into lookup.

The practical symptom we saw was that a caller intentionally invoking local mode with `--dry-run=extract` still got lookup-phase behavior and datasource/network activity.

The minimal code path behind the behavior is:
- `lib/modules/platform/local/index.ts` returns `dryRun: 'lookup'`
- `lib/modules/platform/index.ts` applies `...config, ...platformInfo`
- explicit caller dryRun loses to platform defaults

This PR only changes local platform behavior to preserve an explicit incoming `dryRun` while keeping the existing default of `lookup` when none was supplied.

zeitlinger:
> I'm not sure I've seen this behaviour before - please share a case where you're seeing this happen (and the version of Renovate)
> 
> Also, when raising PRs, please follow our PR template

My last comment is not a complete repo - please tell me if it's still needed.

RahulGautamSingh (changes requested):
Change the PR description to follow the template.

Else, LGTM.

lib/modules/platform/local/index.ts
```diff
@@ -10,9 +10,9 @@ import type {
 export const id = 'local';
 export const experimental = true;
 
-export function initPlatform(_params: PlatformParams): Promise<PlatformResult> {
+export function initPlatform(params: PlatformParams): Promise<PlatformResult> {
   return Promise.resolve({
-    dryRun: 'lookup',
+    dryRun: params.dryRun ?? 'lookup',
```
viceice: we should not allow full because it would run into errors

zeitlinger: Good catch — restricted to `extract` only; `full` and any other value now fall back to `lookup`. Updated in 31f90153d with a test for the `full → lookup` fallback.

zeitlinger:
Updated the PR description to follow the template. Thanks!

viceice (dismissed):

viceice:
docs should be updated

viceice (approved):

PR #43186, opened 2026-05-08
fix(github): discover public email if available

felipecrs:
## Changes

As part of investigating https://github.com/renovatebot/renovate/discussions/43164, I realized that Renovate was not auto-discovering my public email address properly.

That was happening because the `getUserEmail` function requires a token with the `user:email` scope, which I was not providing.

But I realized that the `getUserDetails` function already returns the user's public email address, so I updated the code to use that first, and only call `getUserEmail` if the email is not available in the user details (i.e., the user has not set their email to public).

This also saves an API call in such case.

I tested this against my own account when setting with public vs private email, and it works as expected.

Note I have not tested this against an App, but according to the GitHub API documentation it should work. It's harmless anyway since it will continue to fallback to `getUserEmail` when not available in the user details.

## Context

## AI assistance disclosure

The public repository: https://github.com/felipecrs/renovate-repro-tutorial


viceice (dismissed):

jamietanna (changes requested):
Can we also make sure we cover the case that `/user/emails` returns an error - i.e. in the case we don't have the `user` or `user:emails` scope

felipecrs:
> Can we also make sure we cover the case that `/user/emails` returns an error - i.e. in the case we don't have the `user` or `user:emails` scope

Note we don't error in that case, we just default to `renovate@whitesourcesoftware.com` instead. But it's nice to have it "documented" as a test case indeed. Done in 9e5ac0851b810918da53c65976d88d63046b9842.

jamietanna (approved):

PR #43222, opened 2026-05-10
feat(presets): add security:gomodIndirectSecurityUpdates preset

maxbrunet:
## Changes

Add a preset to allow users to opt-in to vulnerability-only updates for go.mod indirect dependencies.

## Context

Related to discussion #22487

## AI assistance disclosure

The public repository: https://github.com/maxbrunet/renovate-repro-22487

### DIFF
--- a/lib/config/presets/internal/security.preset.ts
+++ b/lib/config/presets/internal/security.preset.ts
@@ -1,6 +1,28 @@
+import { UpdateTypesOptions } from '../../types.ts';
 import type { Preset } from '../types.ts';
 
 export const presets: Record<string, Preset> = {
+  gomodIndirectSecurityUpdates: {
+    description:
+      'Enable go.mod indirect dependency updates only when vulnerabilities have been detected.',
+    extends: [':enableVulnerabilityAlerts'],
+    packageRules: [
+      {
+        description: 'Enable lookups for indirect go.mod dependencies',
+        enabled: true,
+        matchDepTypes: ['indirect'],
+        matchManagers: ['gomod'],
+      },
+      {
+        description:
+          'Disable non-security updates for indirect go.mod dependencies (vulnerability alerts use force to override this)',
+        enabled: false,
+        matchDepTypes: ['indirect'],
+        matchManagers: ['gomod'],
+        matchUpdateTypes: [...UpdateTypesOptions],
+      },
+    ],
+  },
   minimumReleaseAgeNpm: {
     description:
       'Wait until the npm package is three days old before raising the update. This a) introduces a short delay to allow for malware researchers and scanners to (possibly) detect any malicious behaviour in packages, and b) prevents the maintainer and/or NPM from unpublishing a package you already upgraded to, breaking builds.',


jamietanna:
https://github.com/maxbrunet/renovate-repro-22487 seems to be a 404 - is it a private repo by any chance?

jamietanna (commented):
I'd like to see it in use (see above) before we merge - it looks like it should work, but it'd be good to confirm

maxbrunet:
> is it a private repo by any chance?

Oops, I changed the visibility

jamietanna:
Would you mind adding another dependency (which has updates pending) to the project? I'd like to make sure that enabling this doesn't break the raising of PRs for direct dependencies

maxbrunet:
@jamietanna Done PTAL

jamietanna (approved):
@viceice mind a review to confirm this looks good from your side?

RahulGautamSingh (approved):

lib/config/presets/internal/security.preset.ts
```diff
@@ -1,6 +1,28 @@
+import { UpdateTypesOptions } from '../../types.ts';
 import type { Preset } from '../types.ts';
 
 export const presets: Record<string, Preset> = {
+  gomodIndirectSecurityUpdates: {
+    description:
+      'Enable go.mod indirect dependency updates only when vulnerabilities have been detected.',
+    extends: [':enableVulnerabilityAlerts'],
```
viceice, line 8: what does this enable? 🤔 aren't they enabled by default? 🤔

viceice (commented):
otherwise LGTM

lib/config/presets/internal/security.preset.ts
maxbrunet: So I had wondered the same and had asked the agent to dig, this is a bit out of context and old (line numbers), but I had tested manually and the agent was right, it is required.

> Yes — on GitHub, `vulnerabilityAlerts` has a **non-empty default** (lines 2306-2316), so `!input?.vulnerabilityAlerts` at line 51 is `false`, meaning the check passes. And since the default doesn't set `enabled: false`, line 54 also passes. So the GitHub vulnerability alerts path (`getVulnerabilityAlerts`) **runs automatically** when Dependabot security alerts are enabled on the repo.
> 
> However, there are two separate mechanisms:
> 
> 1. **GitHub-native vulnerability alerts** (`vulnerabilityAlerts`) — uses Dependabot alerts via the GitHub GraphQL API. This is active by default on GitHub if the repo has `hasVulnerabilityAlertsEnabled: true`. The `force` object spreads the default `vulnerabilityAlerts` config, which does NOT include `enabled: true`.
> 
> 2. **OSV vulnerability alerts** (`osvVulnerabilityAlerts`) — uses the OSV database. Must be explicitly opted in.
> 
> The problem is: the default `vulnerabilityAlerts` config doesn't include `enabled: true`. So even with Dependabot alerts enabled, the `force` object won't contain `enabled: true`, and Rule 2's `enabled: false` won't be overridden.
> 
> The user **must** explicitly set `vulnerabilityAlerts: { enabled: true }` regardless of which vulnerability source they use. This is what makes the `force: { enabled: true }` override work.

From what I understand, this ensures `enabled: true` is overwritten in the dependency config for vulnerability updates, the name of the preset (`:enableVulnerabilityAlerts`) might be misleading, something like `":setEnableForVulnerabilityUpdates"` could be more correct, especially reading [the description of the option](https://docs.renovatebot.com/configuration-options/#vulnerabilityalerts): _Config to apply when a PR is needed due to a vulnerability in the existing package version._ (The option name itself could be improved, e.g. `vulnerabilityUpdateConfig`)

jamietanna: I thought I'd raised something for this in the past, but https://github.com/renovatebot/renovate/issues/44153 would be a good way to try and unify the configuration, instead of adding more under osv-specific flags (if appropriate)

maxbrunet: @jamietanna I don't think this issue makes sense, `vulnerabilityAlerts` is already used by all security alert mechanism, the fields serve different purposes: `vulnerabilityAlerts` is a package config override object and `osvVulnerabilityAlerts` is a boolean feature flag. It is a naming problem imo

maxbrunet:
Anything blocking merging this PR now that we have Michael's review?

viceice (approved):

PR #42701, opened 2026-04-17
feat(poetry): split python-constrained dependency variants for #31406

wwade:
## Changes

Add the opt-in `splitPythonMarkers` configuration option for dependencies declared with standalone `python_version` or `python_full_version` environment markers, as well as native Poetry multiple-constraint arrays using `python`.

When enabled, Renovate gives each recognized marker variant a distinct branch prefix and commit-message suffix. Renovate also applies the intersection of the project Python constraint and dependency marker during release filtering. This allows dependencies with different version ranges for different Python versions to receive independent updates. Standalone `python_version` and `python_full_version` markers are supported. Compound markers containing `and` or `or` and non-Python markers are left unchanged.

Parse each entry in a native Poetry multiple-constraint array as an independent dependency, addressing the syntax reported in #31406. Preserve the complete PEP 508 dependency string as replacement metadata so Renovate updates the correct declaration when a package appears more than once with different markers. Preserve marker metadata while processing dependency groups and retain duplicate package names when their markers differ.

Apply the behavior to PEP 621 dependency declarations processed by the `pep621` and `poetry` managers.

The option is disabled by default, so existing extraction and update behavior remains unchanged unless users enable it.

## Context

Please select one of the following:

- [x] This closes an existing Issue, Closes: #31406
- [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

This builds on the dependency-level `extractedConstraints` support introduced separately in #43252.

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI-generated non-trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

OpenAI Codex GPT-5.

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

Documented the option, supported declaration formats, filtering behavior, and limitations in the configuration reference and the `pep621` and `poetry` manager documentation.

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

[...]

PR #43252, opened 2026-05-12
refactor(workers): allow extractedConstraints on PackageDependency

wwade:
## Changes

Add `extractedConstraints` to `PackageDependency` so managers can attach constraints discovered for an individual dependency.

Merge dependency-level extracted constraints into the constraints used during lookup while preserving user-configured constraints as the higher-precedence values.

This change separates extracted metadata from the user-facing `constraints` field and prepares for managers that derive constraints from individual dependency declarations.

## Context

- [x] This does not close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

This was extracted from #42701 so the shared worker and type changes can be reviewed independently from the PEP 621 and Poetry implementation.

## AI assistance disclosure

OpenAI Codex was used to both learn about the code base and help with the PR content.

This change adds an internal extraction field and does not introduce a user-facing configuration option.

The worker test now verifies that user-configured constraints take precedence over constraints extracted for an individual dependency.


lib/workers/repository/process/fetch.ts
```diff
@@ -53,6 +53,12 @@ async function lookup(
   const { depName } = dep;
   // TODO: fix types
   let depConfig = mergeChildConfig(packageFileConfig, dep);
+  if (dep.extractedConstraints) {
+    depConfig.constraints = {
+      ...depConfig.constraints,
+      ...dep.extractedConstraints,
+    };
+  }
```
RahulGautamSingh, line 61:
```suggestion
    depConfig.constraints = {
      ...depConfig.constraints,
      ...(dep.extractedConstraints &&{dep.extractedConstraints}),
    };
```

nitpick

RahulGautamSingh (changes requested):
And, please follow the PR Template

lib/workers/repository/process/fetch.ts
wwade: Thanks for the suggestion. I think this one does not work as written: `{ dep.extractedConstraints }` is invalid syntax because object shorthand cannot use a dotted expression.

The closest valid form would be `...(dep.extractedConstraints && { ...dep.extractedConstraints })`, but I'd prefer to keep the explicit `if` here since it avoids assigning `constraints` when there are no dependency-level extracted constraints and keeps the merge behavior clear.

With the change, i.e.

```console
[PARSE_ERROR] Error: Expected `,` or `}` but found `.`
    ╭─[ lib/workers/repository/process/fetch.ts:58:42 ]
    │
 58 │     ...(dep.extractedConstraints && { dep.extractedConstraints }),
    │                                     ┬    ┬
    │                                     ╰─────── Opened here
    │                                          │
    │                                          ╰── `,` or `}` expected
────╯
```

wwade:
@viceice bump

lib/workers/repository/process/fetch.ts
```diff
@@ -53,6 +53,12 @@ async function lookup(
   const { depName } = dep;
   // TODO: fix types
   let depConfig = mergeChildConfig(packageFileConfig, dep);
+  if (dep.extractedConstraints) {
+    depConfig.constraints = {
+      ...depConfig.constraints,
+      ...dep.extractedConstraints,
```
RahulGautamSingh: the ordering is opposite of what we expect.

we prefer user constraints to be preferred above extracted ones, so that user can control their choice via config.

RahulGautamSingh (changes requested):

lib/workers/repository/process/fetch.ts
wwade: thanks, makes sense.

RahulGautamSingh (approved):

wwade:
@viceice does it look ok to you now?

viceice (approved):

PR #42944, opened 2026-04-27
test(util/git): increase coverage ahead of `simple-git` upgrade

jamietanna:
## Changes

Before https://github.com/renovatebot/renovate/pull/42738

From what I can tell the key breakage was `GIT_SSH_COMMAND`:

- https://github.com/renovatebot/renovate/discussions/42739
- https://github.com/renovatebot/renovate/discussions/42772
- https://github.com/renovatebot/renovate/discussions/42740

Claude suggests that there is likely also an impact when `GIT_CONFIG_COUNT` is set, which is not currently tested

However, **??**

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: # 
- [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #43268, opened 2026-05-12
chore(utils): set unsafe flags for SimpleGit

reduckted:
## Changes

Adds the "unsafe" flags to SimpleGit that Renovate requires.

The only flag I'm not sure about is `allowUnsafePager`. It was required to get the tests to pass, but I'm not sure if Renovate is actually expecting a custom `PAGER` environment variable to be used, or whether that was just an example test added in #42944 to show that upgrading `simple-git` would break.

## Context

> [!NOTE]
> This is merging into the branch for #43122 

## AI assistance disclosure

The public repository: https://github.com/reduckted/test-renovate/actions/runs/25728233467/job/75546470636

Ran by checking out this fork: https://github.com/reduckted/test-renovate/actions/runs/25728233467/job/75546470636#step:2:4

### DIFF
--- a/lib/util/git/config.ts
+++ b/lib/util/git/config.ts
@@ -21,7 +21,6 @@ export function simpleGitConfig(): Partial<SimpleGitOptions> {
   const unsafe: SimpleGitOptions['unsafe'] = {
     allowUnsafeSshCommand: true, // For custom `GIT_SSH_COMMAND`.
     allowUnsafeConfigEnvCount: true, // For custom `GIT_CONFIG_COUNT`, `GIT_CONFIG_KEY_*` and `GIT_CONFIG_VALUE_*`.
-    allowUnsafePager: true,
   };
   if (getEnv().RENOVATE_X_CLEAR_HOOKS) {
     unsafe.allowUnsafeHooksPath = true;
--- a/lib/util/git/config.ts
+++ b/lib/util/git/config.ts
@@ -18,18 +18,23 @@ export function getNoVerify(): GitNoVerifyOption[] {
 }
 
 export function simpleGitConfig(): Partial<SimpleGitOptions> {
+  const unsafe: SimpleGitOptions['unsafe'] = {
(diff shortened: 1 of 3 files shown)


lib/util/git/config.ts
```diff
@@ -18,18 +18,23 @@ export function getNoVerify(): GitNoVerifyOption[] {
 }
 
 export function simpleGitConfig(): Partial<SimpleGitOptions> {
+  const unsafe: SimpleGitOptions['unsafe'] = {
+    allowUnsafeSshCommand: true, // For custom `GIT_SSH_COMMAND`.
+    allowUnsafeConfigEnvCount: true, // For custom `GIT_CONFIG_COUNT`, `GIT_CONFIG_KEY_*` and `GIT_CONFIG_VALUE_*`.
+    allowUnsafePager: true,
```
viceice: @jamietanna @secustor I think we should drop this and don't expect users to configure an explicit pager env. 🤔

jamietanna: Yeah - this came from me noticing that if we had `PAGER` set (when running tests locally) but wonder if we can only enable this for tests (if needed)?

reduckted: I don't think it's a problem now that the environment variables only come from `getChildEnv`.

viceice (approved):

package.json
```diff
@@ -254,7 +254,7 @@
     "semver-stable": "3.0.0",
     "semver-utils": "1.1.4",
     "shlex": "3.0.0",
-    "simple-git": "3.35.2",
+    "simple-git": "3.36.0",
```
viceice, line 257: looks like a wrong merge?

reduckted: Because #43122 was updated via force-push, this PR still included some of the original commits that updated `simple-git`. I've had to rebase this branch to get rid of those commits.

viceice (approved):

PR #43293, opened 2026-05-13
fix(git): allow customEnvVariables to override GIT_SSH_COMMAND

colinaaa:
## Changes

`createSimpleGit()` (introduced in #43035) hardcodes `GIT_SSH_COMMAND: 'ssh -o BatchMode=yes'` in the `env` parameter of `getChildEnv()`, which maps to `forcedEnv` — the highest priority layer. This silently overrides any user-configured `GIT_SSH_COMMAND` set via `customEnvVariables`.

This PR conditionally sets `GIT_SSH_COMMAND` in `forcedEnv`: the default `'ssh -o BatchMode=yes'` is only applied when `customEnvVariables` does **not** include `GIT_SSH_COMMAND`. When a user configures their own value, it flows through `globalConfigEnv` and takes effect.

**Before (broken):** `forcedEnv` always sets `GIT_SSH_COMMAND` → user's value from `customEnvVariables` is overridden.
**After (fixed):** `forcedEnv` skips `GIT_SSH_COMMAND` when `customEnvVariables` provides one → user's value wins.

```
extraEnv < parentEnv < globalConfigEnv (customEnvVariables) < userConfiguredEnv < forcedEnv
```

## Context

See https://github.com/renovatebot/renovate/discussions/43292

## AI assistance disclosure

Used AI to assist with code analysis, test authoring, and PR description drafting.

Added three test cases:
- `should allow customEnvVariables to override GIT_SSH_COMMAND` — verifies a custom value set via `setCustomEnv()` takes precedence over the default.
- `should allow customEnvVariables to override GIT_SSH_COMMAND alongside other custom vars` — verifies the override works when combined with other custom environment variables.
- `should not inherit GIT_SSH_COMMAND from process.env when no customEnvVariables are set` — verifies the host process's `GIT_SSH_COMMAND` does not leak through; Renovate uses its own default.

### DIFF
--- a/lib/util/git/index.spec.ts
+++ b/lib/util/git/index.spec.ts
@@ -1614,10 +1614,9 @@ describe('util/git/index', { timeout: 10000 }, () => {
       );
     });
 
-    it('should not inherit GIT_SSH_COMMAND from process.env when no customEnvVariables are set', async () => {
-      // If the host process happens to have GIT_SSH_COMMAND set,
-      // Renovate should still use its own default BatchMode value
-      // rather than inheriting the host's value.
+    it('should allow process.env GIT_SSH_COMMAND to override the default', async () => {
+      // GIT_SSH_COMMAND is declared in extraEnv, so the key is inherited
+      // from process.env via parentEnv (higher priority than extraEnv).
       process.env.GIT_SSH_COMMAND = 'ssh -o SomeHostOption=yes';
 
       const envSpy = vi.spyOn(SimpleGit.prototype, 'env');
@@ -1625,7 +1624,7 @@ describe('util/git/index', { timeout: 10000 }, () => {
       await expect(git.syncGit()).resolves.toBeUndefined();
       expect(envSpy).toHaveBeenCalledWith(
         expect.objectContaining({
-          GIT_SSH_COMMAND: 'ssh -o BatchMode=yes',
+          GIT_SSH_COMMAND: 'ssh -o SomeHostOption=yes',
         }),
       );
     });
--- a/lib/util/git/index.spec.ts
+++ b/lib/util/git/index.spec.ts
@@ -1571,6 +1571,73 @@ describe('util/git/index', { timeout: 10000 }, () => {
         }),
       );
     });
+
+    it('should allow customEnvVariables to override GIT_SSH_COMMAND', async () => {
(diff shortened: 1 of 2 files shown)


lib/util/git/index.spec.ts
```diff
@@ -1571,6 +1571,73 @@ describe('util/git/index', { timeout: 10000 }, () => {
...
+          CUSTOM_TOKEN: 'abc123',
+        }),
+      );
+    });
+
+    it('should not inherit GIT_SSH_COMMAND from process.env when no customEnvVariables are set', async () => {
+      // If the host process happens to have GIT_SSH_COMMAND set,
+      // Renovate should still use its own default BatchMode value
+      // rather than inheriting the host's value.
+      const original = process.env.GIT_SSH_COMMAND;
```
viceice: no need. env is already set in `beforeEach`

https://github.com/renovatebot/renovate/blob/cb71dd1d58cb6ed8d1216448a8d1686c5649f970/lib/util/git/index.spec.ts#L133
```suggestion
```

viceice (dismissed):

lib/util/git/index.spec.ts
colinaaa: Good catch! PR updated.

lib/util/git/index.ts
```diff
@@ -107,7 +108,12 @@ export function createSimpleGit({
         LANG: 'C.UTF-8',
         LC_ALL: 'C.UTF-8',
         // Git will prompt for known hosts or passwords, unless we activate BatchMode.
-        GIT_SSH_COMMAND: 'ssh -o BatchMode=yes',
+        // Only set the default when the user has not configured a custom
+        // GIT_SSH_COMMAND via customEnvVariables, so self-hosted users can
+        // override SSH behaviour (e.g. with a specific identity file).
+        ...(!customEnv.GIT_SSH_COMMAND && {
+          GIT_SSH_COMMAND: 'ssh -o BatchMode=yes',
```
viceice: We can simply pass it as `extraEnv` so parent env has higher prio and it can be inherited from process env.

viceice (approved):
please reuse PR next time instead of opening and closing multiple PR

jamietanna (approved):

PR #43316, opened 2026-05-13
fix: use `handlebars.parse()` for template validation

RahulGautamSingh:
## Changes
- use `handlebars.parse()` for template validation instead of `handlebars.compile()`

## Context

Ref: https://github.com/renovatebot/renovate/discussions/43297

## AI assistance disclosure

### DIFF
--- a/lib/config/validation.ts
+++ b/lib/config/validation.ts
@@ -285,10 +285,7 @@ export async function validateConfig(
       }
       if (optionSupportsTemplating.has(key) && val) {
         try {
-          // TODO: validate string #22198
-          let res = template.compile((val as string).toString(), config, false);
-          res = template.compile(res, config, false);
-          template.compile(res, config, false);
+          template.validate((val as string).toString());
         } catch {
           errors.push({
             topic: 'Configuration Error',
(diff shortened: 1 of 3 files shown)


lib/config/validation.ts
```diff
@@ -285,10 +285,7 @@ export async function validateConfig(
       }
       if (optionSupportsTemplating.has(key) && val) {
         try {
-          // TODO: validate string #22198
```
viceice, line 288: readd this to-do, I don't think this is fixed.

viceice (commented):
otherwise LGTM

lib/util/template/index.ts
```diff
@@ -359,3 +359,7 @@ export function safeCompile<T>(
     return '';
   }
 }
+
+export function validate(template: string): void {
```
jamietanna, line 368:
```suggestion
/**
 * Validate that the template is a valid Handlebars template string.
 *
 * NOTE: Does not validate that fields are set to valid field names.
 */
export function validate(template: string): void {
```

Or something similar

jamietanna (commented):
I think adding some docs to `validate` (so folks are aware it's only parsing, not compiling/validating fully) would be good

jamietanna (approved):

PR #41629, opened 2026-03-03
feat(workers/repository): raise artifact error if `pending` version used in an update

jamietanna:
## Changes

When using Minimum Release Age, and a package manager that doesn't
support being told an explicit version to update to (#41624) it is
possible that an artifact update leads to a different version of a
dependency being used compared to what Renovate is expecting.

This can lead to, at best, a surprising PR update, and at worst, a
supply chain attack.

We should report these cases more explicitly with an Artifact Error, to
allow the reviewers to decide what to do with the changes.

To do this, we need to re-extract package files after an update, and
determine if any of the version(s) of that dependency are using any of
the `pendingVersions`.

This may be slightly breaking to users, which is why this is a `feat`
not a `fix`, but isn't so breaking that it's worth leaving until the
next breaking change .

## Context

Please select one of the following:

- [x] This closes an existing Issue, Closes: #41622
- [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe): Claude Sonnet 4.6 (GitHub Copilot + `crush`)

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

The public repository: https://github.com/JamieTanna-Mend-testing/renovate-iss-41607/pull/1

When skipping checks:

```
 WARN: Artifact error would be reported due to a pending version in use which hasn't passed Minimum Release Age, but as we're running with minimumReleaseAgeBehaviour=timestamp-optional, proceeding. See debug logs for more in
formation (repository=JamieTanna-Mend-testing/renovate-iss-41607, branch=renovate/boto3-1.x-lockfile)
       "packageFileName": "pyproject.toml",
       "depName": "boto3",
       "expectedVersion": "1.42.55",
[...]

PR #43348, opened 2026-05-14
fix(workers/repository): skip pending-version check when re-extracted dep lacks resolved version

mado-m:
## Changes

`checkForPendingVersions` (introduced in #41629) currently throws `WORKER_FILE_UPDATE_FAILED` when the re-extracted dep has none of `lockedVersion`, `newVersion`, `currentVersion`, or `currentValue` populated. That throw aborts the entire `getUpdatedPackageFiles` flow for a single dep we cannot verify — which is disproportionate to the intent of the check (flag artifact updates that introduced a pending version).

This PR demotes the abort to **warn-and-continue**: if the resolved version for a single dep cannot be determined, the pending-version check is skipped for that dep only and the rest of the update proceeds.

```diff
 if (!resolvedVersion) {
-  logger.error(
-    { ..., depName, newVersion: resolvedVersion },
-    `No new version found for '${depName}' after updating '${packageFileName}'`,
-  );
-  throw new Error(WORKER_FILE_UPDATE_FAILED);
+  logger.warn(
+    { ..., depName },
+    `Could not determine resolved version for '${depName}' after updating '${packageFileName}'; skipping pending-version check`,
+  );
+  continue;
 }
```

### Concrete scenario observed

Triggered with `minimumReleaseAge: '7 days'` + the `terraform` manager / `terraform-provider` datasource. The upgrade object Renovate built for `DataDog/datadog` at the cooldown boundary:

```json
{
  "depName": "datadog",
  "datasource": "terraform-provider",
  "packageName": "datadog/datadog",
  "currentValue": "4.6.0",
  "currentVersion": "4.6.0",
  "updates": [{
    "newVersion": "4.7.0",
    "newVersionAgeInDays": 7,
    "pendingVersions": ["4.8.0", "4.9.0"]
  }]
}
```

The relevant log sequence:

```text
DEBUG: terraform.updateArtifacts(main.tf)
DEBUG: Found 1 provider deps
DEBUG: Pinning constraint for "datadog/datadog" to "4.7.0"
DEBUG: Creating hashes for datadog/datadog@4.7.0 (https://registry.terraform.io)
DEBUG: Writing updates to .terraform.lock.hcl
ERROR: No new version found for 'datadog' after updating 'main.tf'
       newVersion: undefined
WARN : Error updating branch: update failure
Error: The process '/usr/bin/docker' failed with exit code 1
```

The `.tf` file *is* successfully rewritten 4.6.0 → 4.7.0 — the `Contents updated` debug log fires earlier in the same run for the same file, and the lockfile artifact write begins. Then `checkForPendingVersions` re-extracts the updated content and, somehow, the re-extracted dep object surfaces no `lockedVersion` / `newVersion` / `currentVersion` / `currentValue`, so `resolvedVersion` is undefined and the function throws.

From the function's own perspective this is a false negative: it cannot determine whether 4.7.0 falls into `pendingVersions: ["4.8.0", "4.9.0"]` (it doesn't), but instead of skipping the check it aborts the whole worker. The actual package-file update succeeded; what failed is only this optional verification pass.

### Why warn-and-continue is the right behavior here

The check exists to surface a soft warning when an artifact update has resolved to a pending version. If the four version fields on the re-extracted dep are all undefined, the function simply cannot perform the comparison for that dep. The remaining deps in the same file can still be verified normally, and a `logger.warn` keeps the skip visible without taking down the update.

Investigating *why* the terraform re-extract surfaces an unpopulated dep here is out of scope — the goal of this PR is just to make `checkForPendingVersions` resilient so a single unverifiable dep does not abort the whole flow.

The existing `it('rejects when an updated dependency has no new version', ...)` test is updated to reflect the new warn-and-continue behavior.

## Context

## AI assistance disclosure

Used Claude (Anthropic) to locate the throw site and draft the code/test change. Reviewed and verified locally.

`pnpm vitest run lib/workers/repository/update/branch/get-updated.spec.ts` → all 58 tests pass.

### DIFF
--- a/lib/workers/repository/update/branch/get-updated.ts
+++ b/lib/workers/repository/update/branch/get-updated.ts
@@ -602,17 +602,16 @@ async function checkForPendingVersions(
       dep.currentVersion ??
       dep.currentValue;
     if (!resolvedVersion) {
-      logger.error(
+      logger.warn(
         {
           packageFile: packageFileName,
           manager,
           branchName: config.branchName,
           depName,
-          newVersion: resolvedVersion,
         },
-        `No new version found for '${depName}' after updating '${packageFileName}'`,
+        `Could not determine resolved version for '${depName}' after updating '${packageFileName}'; skipping pending-version check`,
       );
-      throw new Error(WORKER_FILE_UPDATE_FAILED);
+      continue;
     }
 
     if (resolvedVersion && upgradeInfo.pendingVersions.has(resolvedVersion)) {
--- a/lib/workers/repository/update/branch/get-updated.spec.ts
+++ b/lib/workers/repository/update/branch/get-updated.spec.ts
@@ -1667,8 +1667,7 @@ describe('workers/repository/update/branch/get-updated', () => {
       });
     });
 
-    // should never happen, but our types allow this
-    it('rejects when an updated dependency has no new version', async () => {
+    it('skips the pending-version check when re-extracted dep has no resolvable version', async () => {
       config.upgrades.push({
         packageFile: 'composer.json',
         manager: 'composer',
@@ -1699,19 +1698,16 @@ describe('workers/repository/update/branch/get-updated', () => {
         ],
       });
 
-      await expect(getUpdatedPackageFiles(config)).rejects.toThrowError(
-        'update-failure',
-      );
-
-      expect(logger.logger.error).toHaveBeenCalledWith(
+      const res = await getUpdatedPackageFiles(config);
+      expect(res.artifactErrors).toHaveLength(0);
+      expect(logger.logger.warn).toHaveBeenCalledWith(
         {
           packageFile: 'composer.json',
           manager: 'composer',
           branchName: 'renovate/pin',
           depName: 'some-dep',
-          newVersion: undefined,
         },
-        "No new version found for 'some-dep' after updating 'composer.json'",
+        "Could not determine resolved version for 'some-dep' after updating 'composer.json'; skipping pending-version check",
       );
     });
 


mado-m:
@viceice friendly ping — CLA is now signed. Would appreciate a review when you have a moment. The change is intentionally small (≈8 lines + an updated test) and aimed at making `checkForPendingVersions` resilient to a single unverifiable dep rather than aborting the whole `getUpdatedPackageFiles` flow. Happy to adjust scope/wording if a different direction is preferred. Thanks!

mado-m:
@viceice @RahulGautamSingh @jamietanna sorry for the extra ping.

This PR has been waiting for review for a few days. The change is intentionally small: it only makes `checkForPendingVersions` skip a pending-version check when a re-extracted dependency has no resolved version, instead of aborting the whole `getUpdatedPackageFiles` flow.

Most checks are passing, but `license/cla` still appears to be pending even though I believe I have completed the CLA. Could someone please help recheck the CLA status or let me know if there is anything else I need to do?

Once that blocker is cleared, I would appreciate a review or guidance on whether this is the right direction. Thanks!

jamietanna:
It's possible for you to re-check the CLA status by following the link ☝🏽 

Don't ping maintainers/collaborators on the repo. We're generally working on things as they come in and/or in priority that we determine

(I appreciate that this has waited for a bit, and that it may be blocking you, but we do also have many other PRs to look at, and we don't want to folks to think that if they @-mention us they get prioritised)

lib/workers/repository/update/branch/get-updated.ts
```diff
@@ -602,17 +602,16 @@ async function checkForPendingVersions(
       dep.currentVersion ??
       dep.currentValue;
     if (!resolvedVersion) {
-      logger.error(
+      logger.warn(
```
RahulGautamSingh, line 605: As per our best-practices, `WARN` messages should have a static msg component so they group correctly in metrics/error services. 

The values `depName`, `packageFile` are already in metadata. 

A static message like 'Could not determine resolved version; skipping pending-version check' would be preferable.

RahulGautamSingh (changes requested):

lib/workers/repository/update/branch/get-updated.ts
jamietanna: This is an existing piece of code (that I introduced) that I'd actually prefer to keep dynamic

(if we are changing it, let's do it outside of the scope of this change)

RahulGautamSingh: Sure,

mado-m:
CLA is now passing. Based on the discussion above, I'll leave the existing dynamic WARN message unchanged as it is outside the scope of this PR.

mado-m:
Per the discussion on the `get-updated.ts:605` thread, the dynamic WARN message was agreed to stay as-is (out of scope for this PR). Could the open *Changes requested* review be re-evaluated when convenient? Thanks!

RahulGautamSingh (approved):
lgtm

mado-m:
Just checking in on this. The latest feedback has been addressed/clarified and CI is green; is there anything else needed from my side to move this forward?

viceice (approved):

PR #43369, opened 2026-05-15
feat(npm): support devEngines.runtime and devEngines.packageManager

### LINKED ISSUE #38067: Update the packageManager and runtime versions in `devEngines`

Updating devEngines runtime or packageManager versions - https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines

Example repo: https://github.com/setchy/renovate-devengines/blob/main/package.json

eai04191:
## Changes

Adds native support in the `npm` manager for `devEngines.runtime` and `devEngines.packageManager` in `package.json`.

- Extracts and updates both the single-object and array forms.
- Supports Node.js, Bun, and Deno runtime versions using existing Renovate datasources.
- Keeps package manager versions aligned with the top-level `packageManager` field, including regeneration of existing Corepack hashes.
- Documents the supported Node.js and package manager version fields.

## Context

- [x] This closes an existing Issue, Closes: #38067

## AI assistance disclosure

- [ ] No
- [ ] Yes (minimal)
- [x] Yes (substantive assistance: AI-generated non-trivial portions of code, tests, or documentation).
- [ ] Yes (other)

The implementation, tests, documentation, and PR description were developed with substantive assistance from Claude Code and Codex. I reviewed all changes before submission.

`pnpm check --all lib/modules/manager/npm`

Corepack hash generation and isolation were also verified with Corepack 0.34.6 in `node:22-bookworm`.

byCodex


docs/usage/node.md
```diff
@@ -17,6 +17,7 @@ Renovate understands [codenames for Node.js LTS releases](https://github.com/nod
 Renovate can manage the Node.js version in the following files:
 
 - The [`engines`](https://docs.npmjs.com/files/package.json#engines) field in [`package.json`](https://docs.npmjs.com/files/package.json)
+- The [`devEngines.runtime`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines) field in [`package.json`](https://docs.npmjs.com/files/package.json)
```
viceice, line 20: what about `devEngines.packageManager` ?

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -117,6 +117,84 @@ function replaceAsString(
   throw new Error();
 }
 
+function updateDevEnginesDependency({
+  fileContent,
+  upgrade,
+}: Pick<UpdateDependencyConfig, 'fileContent' | 'upgrade'>): string | null {
+  const { depType, depName, newValue, managerData } = upgrade;
+  /* v8 ignore next 7 -- defensive: dispatcher already filtered */
```
viceice:
```suggestion
  /* v8 ignore if -- defensive: dispatcher already filtered */
```

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -117,6 +117,84 @@ function replaceAsString(
...
+    depType === 'devEngines.runtime' ? 'runtime' : 'packageManager';
+  try {
+    const parsedContents: NpmPackage = JSON.parse(fileContent);
+    const block = parsedContents.devEngines?.[subKey];
+    if (!block) {
+      return null;
+    }
+    let oldVersion: string | undefined;
+    if (isArray(block)) {
+      const idx = managerData?.devEnginesIndex as number | undefined;
```
viceice: add `devEnginesIndex` to `NpmManagerData` so this cast isn't needed

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -117,6 +117,84 @@ function replaceAsString(
...
+      oldVersion = item.version;
+      item.version = newValue;
+    } else {
+      if (block.name !== depName) {
+        return null;
+      }
+      oldVersion = block.version;
+      block.version = newValue;
+    }
+    /* v8 ignore next 3 -- defensive: extract filters out version-less items */
```
viceice:
```suggestion
    /* v8 ignore if -- defensive: extract filters out version-less items */
```

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -117,6 +117,84 @@ function replaceAsString(
...
+      return null;
+    }
+    if (oldVersion === newValue) {
+      return fileContent;
+    }
+    const searchString = JSON.stringify(oldVersion);
+    const newString = JSON.stringify(newValue);
+    const anchor = '"devEngines"';
+    let searchIndex = fileContent.indexOf(anchor);
+    /* v8 ignore next 3 -- defensive: extract verified devEngines presence */
```
viceice:
```suggestion
    /* v8 ignore if -- defensive: extract verified devEngines presence */
```

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -117,6 +117,84 @@ function replaceAsString(
...
+          newString,
+        );
+        if (dequal(parsedContents, JSON.parse(testContent))) {
+          return testContent;
+        }
+      }
+    }
+    /* v8 ignore next -- defensive: oldVersion guaranteed to appear in JSON */
+    return null;
+  } catch (err) /* v8 ignore next 4 */ {
```
viceice:
```suggestion
  } catch (err) /* v8 ignore next */ {
```

please add a test or remove try-catch if this will never throw

eai04191: Kept the try/catch and added a test that reaches it. JSON.parse on the package file can throw on malformed input, so I'd rather keep the guard.

Done in 987319277ca75389b86c3635c5d1e6acadcde7a5.

docs/usage/node.md
eai04191: `devEngines.packageManager` isn't a Node.js version (it's the package manager like pnpm/yarn/npm), so it doesn't really fit this "Node.js version" File Support list. There's also no package-manager-version docs page today. Where would you prefer it documented?

lib/modules/manager/npm/update/dependency/index.ts
eai04191: Done in 4dcc3983e09d62cebf6c5f3d1c590c60880f0c7b.

docs/usage/node.md
eai04191: After reconsidering this, I think the clearest approach is to document package manager versions on this page in a separate section. I added a “Package manager versions” section covering both the existing `packageManager` field and the new `devEngines.packageManager` field, while keeping `devEngines.runtime` in the Node.js version list.

Done in 59aa2e9182c3597bacdf1e20b686722076a14a03.

eai04191:
After taking another look at #38067, I noticed two cases that this PR does not yet handle:

1. The example linked from the issue specifies Node.js, Bun, and Deno in `devEngines.runtime`, but the current implementation only supports Node.js and marks Bun and Deno as `unknown-engines`.
Renovate’s GitHub Actions manager already uses the npm datasource for the runtime versions accepted by `setup-bun` and `setup-deno`, so using the same mapping here would be consistent with the existing implementation.

https://github.com/renovatebot/renovate/blob/87ee522caeb179231407b116486da1fa1beb77a7/lib/modules/manager/github-actions/community.ts

2. As discussed in the issue comments, `devEngines.packageManager.version` can include a Corepack hash.
Although this form is not documented in the npm CLI’s `package.json` documentation, it is explicitly supported by Corepack, which recommends including a hash.
The current implementation would remove an existing hash when updating the version, so it should preserve or regenerate the hash instead.

https://github.com/nodejs/corepack/blob/436b358a19f6d2592cff740078db1b06953c3578/README.md#devenginespackagemanager

I plan to address both cases in this PR. I’ll mark the PR as draft while I work on these changes and mark it ready for review again once they’re addressed.

eai04191:
Both follow-up items are now addressed: Bun and Deno runtime support has been added, and existing Corepack hashes are regenerated when package manager versions are updated. This PR is ready for review again.

lib/modules/manager/npm/extract/index.spec.ts
```diff
@@ -1131,13 +1131,252 @@ describe('modules/manager/npm/extract/index', () => {
...
+    });
+
+    it('skips devEngines items without name', async () => {
+      const pJson = {
+        name: 'demo',
+        devEngines: {
+          packageManager: { version: '9.0.0' } as {
+            name: string;
+            version: string;
+          },
```
viceice: use our `partial` helper instead

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -117,6 +117,84 @@ function replaceAsString(
...
+  ) {
+    return null;
+  }
+  const subKey: 'runtime' | 'packageManager' =
+    depType === 'devEngines.runtime' ? 'runtime' : 'packageManager';
+  try {
+    const parsedContents: NpmPackage = JSON.parse(fileContent);
+    const block = parsedContents.devEngines?.[subKey];
+    if (!block) {
+      return null;
```
viceice, line 144: we should log a warning here as this should never happen?

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -117,6 +117,84 @@ function replaceAsString(
...
+    }
+    let oldVersion: string | undefined;
+    if (isArray(block)) {
+      const idx = managerData?.devEnginesIndex;
+      if (typeof idx !== 'number') {
+        return null;
```
viceice, line 154: same here?

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -117,6 +117,84 @@ function replaceAsString(
...
+      }
+      const item = block[idx];
+      if (!item || item.name !== depName) {
+        return null;
```
viceice, line 167: same

lib/modules/manager/npm/update/dependency/index.ts
```diff
@@ -117,6 +117,84 @@ function replaceAsString(
...
+      }
+      oldVersion = item.version;
+      item.version = newValue;
+    } else {
+      if (block.name !== depName) {
+        return null;
```
viceice, line 177: same

lib/modules/manager/npm/artifacts.ts
```diff
@@ -112,15 +169,92 @@ async function handlePackageManagerUpdates(
...
+
+    for (const packageManagerUpdate of packageManagerUpdates) {
+      const { depName, newVersion } = packageManagerUpdate;
+      const locator = `${depName}@${newVersion}`;
+      let corepackVersion = generatedVersions.get(locator);
+      if (!corepackVersion) {
+        let corepackPackageFileContent: string | null;
+        if (packageManagerUpdate.depType === 'packageManager') {
+          await writeLocalFile(packageFileName, newPackageFileContent);
+          await exec(`corepack use ${locator}`, execOptions);
```
viceice, line 188: sadly #39844 was abandoned

lib/modules/manager/npm/extract/index.spec.ts
eai04191: Updated the test fixture to use the shared `partial` helper in fc9ca182e.

byCodex

lib/modules/manager/npm/update/dependency/index.ts
eai04191: Added a warning for a missing `devEngines` block in 37a03ee95.

byCodex

lib/modules/manager/npm/update/dependency/index.ts
eai04191: Added a warning when the array form is missing `devEnginesIndex` in 37a03ee95.

byCodex

lib/modules/manager/npm/update/dependency/index.ts
eai04191: Added a warning when the indexed array item is missing or does not match `depName` in 37a03ee95.

byCodex

lib/modules/manager/npm/update/dependency/index.ts
eai04191: Added a warning when the single-object entry does not match `depName` in 37a03ee95.

byCodex

lib/modules/manager/npm/artifacts.ts
eai04191: Ah, I wasn't aware of #39844. I may be able to pick that up in a separate PR, but I don’t think it’s within the scope of this one.

JaKXz:
Hi I am running into this issue, sorry to bump an older thread: what’s left/ needed for this PR to be merged? cc @viceice

PR #43373, opened 2026-05-15
fix(datasource/maven): cache 404s on `maven-metadata.xml` for 12h

jamietanna:
## Changes
As part of efforts to reduce the general overload on Maven Central[0]
from Renovate, we can look to reduce unnecessary HTTP lookups.

The Maven Datasource will use the `registryStrategy=merge`, which leads
to looking up a package in each of our registries, and merging them if
present.

However, this means that when looking up a package that /will only ever
be in a given registry/ (such as in
01bcd4fd43b22612ddbd4ffda31981d1d41a5da9) we're hitting other registries
with packages they are unlikely to have.

This is unnecessary, significantly increases traffic to Maven Central,
and leads to us re-fetching the same 404s which count towards rate
limits, and increased, unnecessary, server load on Maven Central.

We can instead introduce a new cache for 404s for the
`maven-metadata.xml`, for 24 hours, which should reduce traffic to Maven
Central (and other configured registries) if they only have a subset of
dependencies.

[0]: https://www.sonatype.com/blog/maven-central-and-the-tragedy-of-the-commons

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/datasource/maven/util.ts
+++ b/lib/modules/datasource/maven/util.ts
@@ -4,7 +4,7 @@ import { XmlDocument } from 'xmldoc';
 import { HOST_DISABLED } from '../../../constants/error-messages.ts';
 import { logger } from '../../../logger/index.ts';
 import { ExternalHostError } from '../../../types/errors/external-host-error.ts';
-import { getCacheType } from '../../../util/cache/package/index.ts';
+import * as packageCache from '../../../util/cache/package/index.ts';
 import { PackageHttpCacheProvider } from '../../../util/http/cache/package-http-cache-provider.ts';
 import { type Http, HttpError } from '../../../util/http/index.ts';
 import type { HttpOptions, HttpResponse } from '../../../util/http/types.ts';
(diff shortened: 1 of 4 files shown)


lib/modules/datasource/maven/util.ts
```diff
@@ -91,12 +91,35 @@ function selectCacheProvider(url: string): PackageHttpCacheProvider {
   return cacheProvider;
 }
 
+const METADATA_NOT_FOUND_NAMESPACE =
+  'datasource-maven:metadata-not-found' as const;
+const METADATA_NOT_FOUND_TTL_MINUTES = 60 * 24; // 24 hours
```
jamietanna: We could alternatively use a soft cache of 12 hours, and cache it up to 7 days?

viceice: sounds good

viceice (dismissed):

viceice:
we should add some jitter to the cache times, so they don't timeout on same time which again leads to rate limiting

viceice (approved):

viceice:
i wonder if we should add the jitter to cache layer in general 🤔

PR #43376, opened 2026-05-15
feat(self-hosted): make `user-agent` templatable

jamietanna:
## Changes

For Mend-hosted apps, it would be useful to provide a little bit more
information for where a Renovate HTTP request is coming from.

To provide more control over the generated field, we can make it
templatable, with the option to specify the `renovateVersion`.

As this now exposes Handlebars syntax to the user, we should be clear
that this `supportsTemplating`, and document with tests what this
allows.

Co-authored-by: Claude Sonnet 4.6 <jamie.tanna+claude-code@mend.io>

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/util/template/index.ts
+++ b/lib/util/template/index.ts
@@ -238,7 +238,8 @@ export const allowedFields = {
   releases: 'An array of releases for an upgrade',
   releaseNotes: 'A ChangeLogNotes object for the release',
   releaseTimestamp: 'The timestamp of the release',
-  renovateVersion: 'The currently running Renovate version',
+  renovateVersion:
+    'The currently running Renovate version. Only supported in the `user-agent` configuration option.',
   repository: 'The current repository',
   semanticPrefix: 'The fully generated semantic prefix for commit messages',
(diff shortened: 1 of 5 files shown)


viceice (dismissed):
why not making it templateable with version as a placeholder 🤔

lib/util/template/index.ts
```diff
@@ -238,6 +238,7 @@ export const allowedFields = {
   releases: 'An array of releases for an upgrade',
   releaseNotes: 'A ChangeLogNotes object for the release',
   releaseTimestamp: 'The timestamp of the release',
+  renovateVersion: 'The currently running Renovate version',
```
viceice: should we mention that it's only available for user agent?

viceice (dismissed):

lib/util/template/index.ts
jamietanna: I was wondering that. We don't really say where things are available, but I think being explicit here makes sense 👍🏽

viceice (approved):

PR #43394, opened 2026-05-16
feat(manager/npm): walk up to find inherited packageManager

esetnik:
## Summary

When generating a per-workspace lockfile, Renovate only checks the **sibling** `package.json` for a `packageManager` pin before falling back to the lockfile-derived constraint. In a monorepo where `packageManager` is set only on the root `package.json` (the standard recommendation), the gap silently routes Renovate onto whatever `pnpm@latest` / `npm@latest` / `yarn@latest` resolves to at the moment of the run.

This PR mirrors corepack's resolution semantics: walks parent directories from the lockfile directory looking for an ancestor `package.json` with a matching `packageManager` (or `devEngines.packageManager` / `engines` / `volta`) before falling through to the lockfile-derived fallback. Wired into pnpm, npm, and yarn — the same gap exists in all three managers.

Discussed in #43161; approved by @RahulGautamSingh.

## Why this matters

The original report (#43161) was triggered by a real regression: a monorepo with `packageManager: pnpm@10.33.4` set only on the root `package.json` had every Renovate dependency-update PR opened on 2026-05-07 silently switch to `pnpm@latest` (which had just rolled to pnpm 11.0.8 that day). Because pnpm 11 silently ignores `pnpm.overrides` and `pnpm.patchedDependencies` in `package.json` (filed upstream as pnpm/pnpm#11536), the regenerated lockfiles stripped CVE-remediation overrides — one PR merged before anyone caught it, restoring a known-vulnerable transitive dep to the develop branch.

The Renovate-side mitigation is independent of the pnpm-side bug: even if a future manager has no such silent breakage, allowing the lockfile-derived `latest` to win over a deliberate root-level pin is a recipe for cross-major surprises. Honouring the inherited pin matches how corepack itself resolves the binary at runtime.

## Change shape

- New `getInheritedPackageManagerVersion(name, lockFileDir)` in `lib/modules/manager/npm/post-update/utils.ts` — walks parent directories from `lockFileDir`, returns the first ancestor `packageManager` / `devEngines.packageManager` / `engines` / `volta` match for `name`. Skips `lockFileDir` itself (the sibling is already checked by the caller).
- Wired into the constraint chain in `pnpm.ts`, `npm.ts`, and `yarn.ts`, sitting between the sibling `getPackageManagerVersion` and the lockfile-derived fallback.
- New `utils.spec.ts` with 8 focused tests covering: simple walk-up, walking past intermediate directories, closest-ancestor-wins, no-pin-found, lockFileDir-is-repo-root, empty lockFileDir, `engines` field on ancestor, and ignoring a different manager pinned at the ancestor.
- Two new integration tests in `pnpm.spec.ts` exercising the walk-up via the full `generateLockFile` flow, including the "sibling pin wins, ancestor is never read" path.
- Existing tests in `npm.spec.ts`, `yarn.spec.ts`, and `pnpm.spec.ts` that count `readLocalFile` calls have been updated to reflect the extra ancestor read (3 → 4) and, where they used a chained `mockResolvedValueOnce` queue, a matching `'{}'` entry was added for the ancestor read.

## Test plan

- [x] `pnpm exec vitest run lib/modules/manager/npm` — 496 tests pass
- [x] `pnpm exec vitest run lib/modules/manager/npm/post-update --coverage --coverage.include='lib/modules/manager/npm/post-update/utils.ts'` — 100% statements / branches / functions / lines on the new helper
- [x] `pnpm exec biome check lib/modules/manager/npm/post-update/` — clean
- [x] `pnpm exec oxlint -c .oxlintrc.json lib/modules/manager/npm/post-update/` — 0 warnings, 0 errors
- [x] `pnpm exec tsc --noEmit` — clean

## AI assistance disclosure

Per the contributing guide: this PR was authored with assistance from Claude Code (Anthropic's CLI for Claude). Claude generated the implementation, the unit tests, the integration tests, the existing-test fixes for the new read count, the commit message, and this PR body. The design (walk-up that mirrors corepack, applied to pnpm/npm/yarn uniformly) was discussed and approved manually before implementation began, and the resulting code/tests were reviewed by me before push.

### DIFF
--- a/lib/modules/manager/npm/post-update/pnpm.ts
+++ b/lib/modules/manager/npm/post-update/pnpm.ts
@@ -25,6 +25,7 @@ import type { PnpmWorkspaceFile } from '../extract/types.ts';
 import { getNodeToolConstraint } from './node-version.ts';
 import type { GenerateLockFileResult, PnpmLockFile } from './types.ts';
 import {
+  getInheritedPackageManagerVersion,
   getNodeOptions,
   getPackageManagerVersion,
   lazyLoadPackageJson,
@@ -56,7 +57,8 @@ export async function generateLockFile(
       constraint:
         getPnpmConstraintFromUpgrades(upgrades) ?? // if pnpm is being upgraded, it comes first
         config.constraints?.pnpm ?? // from user config or extraction
-        getPackageManagerVersion('pnpm', await lazyPgkJson.getValue()) ?? // look in package.json > packageManager or engines
+        getPackageManagerVersion('pnpm', await lazyPgkJson.getValue()) ?? // look in sibling package.json > packageManager or engines
+        (await getInheritedPackageManagerVersion('pnpm', lockFileDir)) ?? // walk up to find an inherited packageManager (mirrors corepack)
         (await getConstraintFromLockFile(lockFileName)), // use lockfileVersion to find pnpm version range
     };
 
--- a/lib/modules/manager/npm/post-update/pnpm.ts
+++ b/lib/modules/manager/npm/post-update/pnpm.ts
@@ -51,13 +51,13 @@ export async function generateLockFile(
   let lockFile: string | null = null;
   const commands: string[] = [];
   try {
-    const lazyPgkJson = lazyLoadPackageJson(lockFileDir);
+    const lazyPkgJson = lazyLoadPackageJson(lockFileDir);
     const pnpmToolConstraint: ToolConstraint = {
       toolName: 'pnpm',
       constraint:
         getPnpmConstraintFromUpgrades(upgrades) ?? // if pnpm is being upgraded, it comes first
         config.constraints?.pnpm ?? // from user config or extraction
-        getPackageManagerVersion('pnpm', await lazyPgkJson.getValue()) ?? // look in sibling package.json > packageManager or engines
+        getPackageManagerVersion('pnpm', await lazyPkgJson.getValue()) ?? // look in sibling package.json > packageManager or engines
         (await getInheritedPackageManagerVersion('pnpm', lockFileDir)) ?? // walk up to find an inherited packageManager (mirrors corepack)
         (await getConstraintFromLockFile(lockFileName)), // use lockfileVersion to find pnpm version range
     };
@@ -86,7 +86,7 @@ export async function generateLockFile(
       extraEnv,
       docker: {},
       toolConstraints: [
-        await getNodeToolConstraint(config, upgrades, lockFileDir, lazyPgkJson),
+        await getNodeToolConstraint(config, upgrades, lockFileDir, lazyPkgJson),
         pnpmToolConstraint,
       ],
     };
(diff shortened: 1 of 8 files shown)


lib/modules/manager/npm/post-update/utils.ts
```diff
@@ -59,3 +60,47 @@ export function getPackageManagerVersion(
...
+    return null;
+  }
+
+  while (true) {
+    const pkg = await loadPackageJson(currentDir);
+    const version = getPackageManagerVersion(name, pkg);
+    if (version) {
+      const relPath =
+        currentDir === '.' ? 'package.json' : `${currentDir}/package.json`;
+      logger.debug(
```
RahulGautamSingh, line 91: Nit: could use `upath.join(currentDir, 'package.json')` here for consistency with how `loadPackageJson` constructs the path.

lib/modules/manager/npm/post-update/pnpm.ts
```diff
@@ -56,7 +57,8 @@ export async function generateLockFile(
       constraint:
         getPnpmConstraintFromUpgrades(upgrades) ?? // if pnpm is being upgraded, it comes first
         config.constraints?.pnpm ?? // from user config or extraction
-        getPackageManagerVersion('pnpm', await lazyPgkJson.getValue()) ?? // look in package.json > packageManager or engines
+        getPackageManagerVersion('pnpm', await lazyPgkJson.getValue()) ?? // look in sibling package.json > packageManager or engines
```
RahulGautamSingh: `lazyPgkJson` is a typo, should be `lazyPkgJson`. Not introduced by this PR, but since you're touching this line, could you pls fix it

esetnik: Done in 90c631af. Same typo had been copy-pasted into `yarn.ts`, so I fixed it there in the same commit — let me know if you'd rather I split that out.

[Written by Claude]

lib/modules/manager/npm/post-update/utils.ts
esetnik: Done in 90c631af.

[Written by Claude]

lib/modules/manager/npm/post-update/utils.ts
```diff
@@ -59,3 +60,46 @@ export function getPackageManagerVersion(
...
+ * `packageManager` pin set only at the monorepo root.
+ *
+ * `lockFileDir` itself is intentionally skipped — the caller already loads the
+ * sibling `package.json` directly via {@link getPackageManagerVersion}.
+ */
+export async function getInheritedPackageManagerVersion(
+  name: string,
+  lockFileDir: string,
+): Promise<string | null> {
+  const startDir = upath.normalize(lockFileDir || '.');
```
RahulGautamSingh:
```suggestion
  const startDir = upath.normalize(lockFileDir ?? '.');
```

RahulGautamSingh (commented):
lgtm otherwise

RahulGautamSingh (commented):
Failing coverage

RahulGautamSingh (approved):

PR #43422, opened 2026-05-18
fix: collect branch-mode postUpgradeTasks from all upgrades

JuJup:
-  https://github.com/renovatebot/renovate/discussions/42704

### Problem

When a branch has multiple upgrades from different package files and only some define `postUpgradeTasks` with `executionMode: "branch"`, the task is silently skipped.

This happens because `executePostUpgradeCommands` reads `config.postUpgradeTasks` to determine the branch-level task. That property is inherited from `config.upgrades[0]` via the `generateBranchConfig` spread (`config = { ...config, ...config.upgrades[0] }`). If the first sorted upgrade doesn't have `executionMode: "branch"`, the branch-level task is set to `undefined`. Meanwhile, the `updateUpgradeCommands` filter excludes any upgrade with `executionMode: "branch"` — so the task falls through both code paths.

### Example

See https://github.com/renovatebot/renovate/discussions/42704

### Fix

Scan `config.upgrades` for branch-mode tasks instead of relying on `config.postUpgradeTasks`. Identical tasks (from the same packageRule matching multiple upgrades) are deduplicated via `JSON.stringify` to preserve the "run once per branch" semantics. Falls back to `config.postUpgradeTasks` when no individual upgrade defines `executionMode: "branch"`.

### Tests

Added 6 unit tests for `executePostUpgradeCommands` covering:

- No changed files → returns null
- Branch-mode task on a non-first upgrade (the core bug)
- Upgrade without `postUpgradeTasks` alongside one with branch-mode
- First upgrade is branch-mode, second is update-mode (regression guard)
- Multiple upgrades with different branch-mode commands
- Mixed execution modes (update + branch on different upgrades)

## Context

## AI assistance disclosure

### DIFF
--- a/lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
+++ b/lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
@@ -342,17 +342,37 @@ export default async function executePostUpgradeCommands(
     return null;
   }
 
-  const branchUpgradeCommands: BranchUpgradeConfig[] = [
-    {
+  const uniqueBranchTasks = new Map<string, BranchUpgradeConfig>();
+  for (const upgrade of config.upgrades) {
+    if (upgrade.postUpgradeTasks?.executionMode === 'branch') {
+      const key = JSON.stringify(upgrade.postUpgradeTasks);
+      if (!uniqueBranchTasks.has(key)) {
+        uniqueBranchTasks.set(key, {
+          manager: upgrade.manager,
+          depName: upgrade.depName,
+          branchName: config.branchName,
+          postUpgradeTasks: upgrade.postUpgradeTasks,
+        });
+      }
+    }
+  }
+
+  // Fall back to branch-level config when no individual upgrade defines
+  // executionMode: 'branch', e.g. when postUpgradeTasks is only set at the
+  // top level of the BranchConfig.
+  if (
+    uniqueBranchTasks.size === 0 &&
+    config.postUpgradeTasks?.executionMode === 'branch'
+  ) {
+    uniqueBranchTasks.set(JSON.stringify(config.postUpgradeTasks), {
       manager: config.manager,
       depName: config.upgrades.map(({ depName }) => depName).join(' '),
       branchName: config.branchName,
-      postUpgradeTasks:
-        config.postUpgradeTasks!.executionMode === 'branch'
-          ? config.postUpgradeTasks
-          : undefined,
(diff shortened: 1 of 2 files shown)


JuJup:
The `cla-assistant` added the two comments, as my mail was wrong in the commit. I fixed that, but one of the two comments still shows the old error. Not sure how to fix that.

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -342,17 +342,37 @@ export default async function executePostUpgradeCommands(
...
+          postUpgradeTasks: upgrade.postUpgradeTasks,
+        });
+      }
+    }
+  }
+
+  // Fall back to branch-level config when no individual upgrade defines
+  // executionMode: 'branch', e.g. when postUpgradeTasks is only set at the
+  // top level of the BranchConfig.
+  if (
```
RahulGautamSingh, line 363: test this, I don't think it is possible

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -342,17 +342,37 @@ export default async function executePostUpgradeCommands(
     return null;
   }
 
-  const branchUpgradeCommands: BranchUpgradeConfig[] = [
-    {
+  const uniqueBranchTasks = new Map<string, BranchUpgradeConfig>();
```
RahulGautamSingh, line 345: test the dedup behavior

RahulGautamSingh (changes requested):

viceice (commented):
see comments

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
JuJup: Thanks. I think you're right, couldn't find a case where the fallback would be needed.
Removed it in 66b37a974c753cfa0736cdd4fd90714112f4e5e2

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
JuJup: I added an assertion to the test. That's sufficient?

viceice (commented):
please rebase and force push (remove the closes issue from commit message)

viceice (approved):

PR #43346, opened 2026-05-14
fix(datasource/maven): don't look up potential Gradle Plugins in Maven Central

jamietanna:
## Changes

As part of efforts to reduce the general overload on Maven Central[0]
from Renovate, we can look to avoid lookups that we know are likely to
never resolve.

For instance, via #43146, we can skip any packages that appear to be a
Gradle Plugin, as they will only ever be sourced from the  Gradle Plugin
Portal.

This may have some false positives - [we can see 137 `gradle.plugin` results on Maven Central](https://mvnrepository.com/search?q=gradle.plugin&), but I'm happy with these dependencies being skipped. If we need to, we can re-add an override for them in the future.

For instance, in the last 24 hours, Mend-hosted Renovate has seen 3507 rate limit errors related to `gradle.plugin` (out of a total 27137 events)

[0]: https://www.sonatype.com/blog/maven-central-and-the-tragedy-of-the-commons

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: # 
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [x] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #43432, opened 2026-05-18
fix(datasource/maven): treat rate-limit errors for Maven Central mirror as Maven Central

jamietanna:
## Changes

In #43346 we introduced knowledge of
the Maven Central mirror, but haven't taken it into account when looking
at rate limit errors.

To prevent this in the future, we can introduce a new helper function,
`isMavenCentral`, for other usage.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/datasource/maven/common.ts
+++ b/lib/modules/datasource/maven/common.ts
@@ -1,5 +1,16 @@
+import { parseUrl } from '../../../util/url.ts';
 export const id = 'maven';
 
 export const MAVEN_REPO = 'https://repo.maven.apache.org/maven2';
 export const MAVEN_CENTRAL_MIRROR = 'https://repo1.maven.org/maven2';
 export const MAVEN_CENTRAL_URLS = [MAVEN_REPO, MAVEN_CENTRAL_MIRROR];
+
+function getHost(url: string): string | undefined {
+  return parseUrl(url)?.host;
+}
+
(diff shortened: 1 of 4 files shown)


lib/modules/datasource/maven/common.ts
```diff
@@ -1,5 +1,16 @@
+import { parseUrl } from '../../../util/url.ts';
 export const id = 'maven';
 
 export const MAVEN_REPO = 'https://repo.maven.apache.org/maven2';
 export const MAVEN_CENTRAL_MIRROR = 'https://repo1.maven.org/maven2';
 export const MAVEN_CENTRAL_URLS = [MAVEN_REPO, MAVEN_CENTRAL_MIRROR];
```
jamietanna, line 6: For a follow-up - should these be set across `defaultRegistryUrls`?

viceice: we probably should filter for duplicates, so we don't use mirror urls and get another rate limit

viceice: maybe datasource can have an optional dedupe registry urls for this? 🤔

viceice (approved):

lib/modules/datasource/maven/common.ts
jamietanna: Ah yes, that's fair - we don't want to try and have both 😁

PR #43436, opened 2026-05-18
feat(publish): pin production dependency tree via npm-shrinkwrap.json

Squixx:
Add `npm-shrinkwrap.json` to the published npm tarball so that consumers installing renovate via `npx` or `npm install -g` get exact dependency versions matching those tested against `pnpm-lock.yaml`.

The shrinkwrap is generated from `pnpm-lock.yaml`'s resolved versions, ensuring the full transitive tree is pinned without floating semver ranges.

## Changes

When users run `npx renovate` or `npm install -g renovate`, npm resolves transitive dependency versions at install time from semver ranges — not from `pnpm-lock.yaml`, which is not shipped with the package. This means the installed tree can differ from what was tested, and may include package versions unknown at publish time.

This PR ships `npm-shrinkwrap.json` with the package. Unlike `package-lock.json`, npm reads a shrinkwrap from a published package when it is installed as root (i.e. `npx`, global install). All 596 production packages are pinned to the exact versions in `pnpm-lock.yaml`.

A new script `tools/generate-npm-shrinkwrap.mjs` (run via `pnpm update-npm-shrinkwrap`) generates the shrinkwrap by:

1. Parsing `pnpm-lock.yaml` and extracting every resolved `name@version`
2. Passing those as npm `overrides` when running `npm install --package-lock-only`, so npm produces a lockfile with registry URLs and integrity hashes matching pnpm's resolution
3. Stripping the `overrides` block from the output before writing `npm-shrinkwrap.json`

The `test-e2e` CI job now verifies the shrinkwrap is up to date (`git diff --exit-code npm-shrinkwrap.json`) before packing.

After any `pnpm-lock.yaml` change, run `pnpm update-npm-shrinkwrap` and commit the result.

## Context

## AI assistance disclosure

- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation). Tool: Claude (Sonnet 4.5 via Cursor).

- [x] Code inspection only

I dont (and shoudnt) have access to this part of your release pipeline, so proceed with caution.

But for a tool like renovate, reducing its attack surface would be nice.


jamietanna (commented):
We'd be happy to generate the shrinkwrap on publish, but not to commit it - mind amending? 

And setting up the CLA - looks like the PR author is different to commit author

Squixx:
> We'd be happy to generate the shrinkwrap on publish, but not to commit it - mind amending?
> 
> And setting up the CLA - looks like the PR author is different to commit author

Yes, makes sense in terms of maintenance. 

Commits are fixed. Normal instead of corp author now, CLA signed

.gitignore
```diff
@@ -40,3 +41,5 @@ test/datasource/nuget/_fixtures/obj
 /.agents
 /.claude
 /.opencode
+/AGENTS.md
+/CLAUDE.md
```
jamietanna:
```suggestion
```

.gitignore
```diff
@@ -39,4 +40,4 @@ test/datasource/nuget/_fixtures/obj
 # local development directories
 /.agents
 /.claude
-/.opencode
+/.opencode
```
jamietanna: Please re-add the newline at end-of-file

jamietanna (commented):
Will try and review today/tomorrow - no need to keep merging in `main` to this branch in the meantime

Squixx:
Any follow-up on this?

jamietanna:
Apologies for the delay. A quick review of this via Claude Sonnet 5 shows:

````markdown
## Review: #43436 — `feat(publish): pin production dependency tree via npm-shrinkwrap.json`

### Overview

Adds `npm-shrinkwrap.json` to the published npm tarball so `npx`/`npm install -g` consumers get the exact transitive dependency versions that were tested via `pnpm-lock.yaml`, instead of npm re-resolving semver ranges at install time. Mechanism: a new `tools/generate-npm-shrinkwrap.mjs` script parses `pnpm-lock.yaml`, builds npm `overrides` from the resolved versions, runs `npm install --package-lock-only` in a scratch dir against those overrides, and writes the result out as `npm-shrinkwrap.json`. `tools/prepare-release.ts` now calls this during the release build. 4 files changed, no tests (tools/ scripts are excluded from coverage requirements, so that's fine here).

### Critical correctness issue: version selection is not actually pinned to what pnpm resolved

`tools/generate-npm-shrinkwrap.mjs` builds its version map like this:

```js
for (const key of Object.keys(pnpmLock.packages ?? {})) {
  const name = key.slice(0, atIdx);
  const version = key.slice(atIdx + 1);
  pnpmVersions[name] ??= version;   // keeps the FIRST version seen
}
```

`pnpm-lock.yaml`'s `packages:` section frequently contains **multiple resolved versions of the same package name** (diamond dependencies) — verified in this repo's own lockfile:

```
type-fest@0.13.1
type-fest@0.8.1
type-fest@1.4.0
type-fest@2.19.0
type-fest@4.41.0
type-fest@5.6.0
```
(also `wrap-ansi` ×4, `string-width` ×4, `p-map` ×4, `get-stream` ×4, `ansi-styles` ×4, and ~15 more package names with 3+ resolved versions.)

Because YAML keys are emitted in ascending order and `??=` keeps the *first* assignment, `pnpmVersions[name]` ends up being whichever version sorts lowest — not the version any particular consumer actually needs. That value is then applied as a single flat `overrides["name"] = version` entry, which npm applies **tree-wide**, forcing every transitive consumer of that package name to the same version. But npm overrides only support per-name pinning at the root of the tree the way this script uses them — there's no per-parent scoping — so any dependent that pnpm intentionally resolved to a *different* coexisting version gets silently forced onto an incompatible one.

- **Failure scenario:** any production dependency that pulls in, say, `wrap-ansi@2.x` transitively (a real, distinct dependency chain from whatever pulls in `wrap-ansi@8.x`) gets forced to whichever version key happened to sort first in the lockfile — potentially a major-version mismatch with a completely different API. The resulting `npm-shrinkwrap.json` would **not** match what was tested against `pnpm-lock.yaml`, which is the entire premise of this PR.
- **Fix direction:** this needs to walk pnpm's per-consumer resolution (the `snapshots:` section in lockfileVersion 9, which encodes which exact version each dependent resolves for each dep) rather than flattening the deduplicated `packages:` section into one version per name. A flat `{name: version}` map is structurally incapable of representing multiple coexisting versions of the same package, which pnpm allows and this lockfile actually uses extensively.

### PR description doesn't match the diff

The PR body states:

> The `test-e2e` CI job now verifies the shrinkwrap is up to date (`git diff --exit-code npm-shrinkwrap.json`) before packing.

The diff only touches `.gitignore`, `package.json`, `tools/generate-npm-shrinkwrap.mjs`, and `tools/prepare-release.ts` (confirmed via `gh pr view --json files`) — there's no CI workflow change anywhere. As it stands, nothing enforces that `npm-shrinkwrap.json` stays in sync with `pnpm-lock.yaml`; it relies entirely on contributors remembering to run `pnpm update-npm-shrinkwrap` after every lockfile change, which is exactly the kind of manual step that goes stale. Given the PR's own stated goal is exact version matching, this gap should either be added before merge or the description corrected to not overclaim.

### Other observations (minor)

- `generateShrinkwrap()` in `tools/prepare-release.ts` runs unconditionally as part of every `build()` (i.e. every release), doing a live `npm install --package-lock-only` against the npm registry with no offline/failure fallback other than aborting the whole release — acceptable for a release pipeline but worth being aware this adds a network dependency + registry-availability risk to releases.
- `const count = Object.keys(lock.packages ?? {}).length;` in the generator counts the root `""` package entry along with real dependencies, so the logged "N packages pinned" count is off by one. Cosmetic only.
- I verified empirically (via `npm pack --dry-run`) that having `npm-shrinkwrap.json` in `.gitignore` while also listing it in `package.json`'s `files` array does **not** cause npm to exclude it from the tarball — `files` wins over `.gitignore` for packing purposes, so that part of the setup is correct, despite initially looking suspicious.

### Test coverage / conventions

- No `.spec` tests were added for the new script, but `tools/**` is fully excluded from the coverage config (`vitest.config.mts:103`), so this doesn't violate the repo's 100%-coverage rule.
- Author disclosed substantial AI assistance and explicitly flagged they haven't tested the release pipeline path — consistent with the version-selection bug above going unnoticed, since it wouldn't surface without inspecting the lockfile's duplicate-version entries.

### Recommendation

Do not merge as-is — the core version-pinning logic can silently produce a shrinkwrap with incompatible package versions for any package pnpm resolves at multiple versions (which this repo's own lockfile does for 20+ packages), undermining the feature's whole purpose. The missing CI freshness check should also be resolved or the PR description corrected.
````

Not sure if these are things you'd noticed yourself? 

We're generally approving of this being a thing that we do, but might need a few tweaks first ☝🏼 

Also, could be something we try on a "smaller" repo like https://github.com/renovatebot/osv-offline or similar, before we land it here

Squixx:
I stand corrected, after some more research i see pnpm dlx actually _does_ now support minReleaseAge https://github.com/pnpm/pnpm/issues/10338

which makes the whole point moot to be honest.

Maybe this consideration can be added for https://docs.renovatebot.com/getting-started/running/ to help point people in the right direction. But implementing this here would (imo) create more overhead that solves an edge case risk for ONLY npx. when there are better solutions available to run this securely.

I'll close this but would be willing to discuss / improve the docs on this if you'd like

PR #37075, opened 2025-07-19
feat: add `forgejo` as a standalone platform

pat-s:
## Changes

Adds `forgejo` as a standalone platform and decouples it as a synonym from `gitea`.
Forgejo is a hard-fork since some time now and APIs have started to diverge.
While it still works right now, a split is inevitable moving forward.

I pretty much copied & renamed the existing Gitea files for now, which should be suffcient as a starting point.

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [x] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

PR #43449, opened 2026-05-18
fix(platform/gitea): retry automerge until mergeable + reattempt on update

silenthooligan:
## Changes

Two gaps make platform-native automerge fail silently on Gitea and Forgejo. Forgejo runs on the Gitea code since #45825. This PR fixes both.

### 1. `createPr` enables automerge before the PR is mergeable

Gitea and Forgejo compute `mergeable` in the background after a push. Until that finishes, `POST /pulls/{index}/merge` returns `405 Please try again later`. `createPr` made one attempt right after creating the PR and only logged the failure. New PRs often never got automerge.

Seen with Renovate 43.184.0 against Gitea 1.26.1:

```
WARN: Gitea-native automerge: fail
  "statusCode": 405
  "url": ".../pulls/12/merge"
  "message": "Please try again later"
```

The same request succeeds with `201` a few seconds later.

### 2. No `reattemptPlatformAutomerge`

The branch worker calls `platform.reattemptPlatformAutomerge?.()` after a branch update. Gitea and Forgejo did not implement it. A PR created before automerge was enabled for it never got automerge.

### Fix

- New `tryPrAutomerge()` in `createPlatform()`, used by `createPr` and the new `reattemptPlatformAutomerge`. Both are exported from `gitea` and `forgejo`.
- It retries the merge call while the platform answers `405`. The wait between attempts is `250 * attempt²` ms, the same as GitLab's `tryPrAutomerge`.
- The new experimental global option `prMergeabilityCheckAttempts` sets the number of attempts. The default of `5` waits at most 7.5 s. A value of `0` still makes one attempt. The option is platform-independent for GitLab to adopt later. Only `gitea` and `forgejo` use it for now.
- Automerge is best-effort. Every error is logged and never fails `createPr`.
- A `409` from the merge endpoint is logged at debug as "already scheduled". Gitea does not cancel a scheduled automerge on push. Only `RemoveScheduledAutoMerge` does. Without this, every re-attempt after a branch update would log a warning.
- The version gate (Gitea 1.24.0+, Forgejo 10.0.0+) is unchanged.

## Context

Supersedes #45162. That PR was the Forgejo copy of this fix and is no longer needed after #45825.

## AI assistance disclosure

Claude (Claude Code) wrote substantial parts of the code, tests and docs. Claude Fable 5 did the earlier revisions. Claude Opus 5 did the port onto the shared implementation after #45825. The original Gitea fix, the design and the review of every change are mine.

Who answers review comments:

- [x] An agent will draft replies and @silenthooligan will read them before they are posted. **Name the account.**

The same tests run against both `gitea` and `forgejo`. They cover the retry on `405`, the attempts cap, a setting of `0`, no retry on other errors, and `reattemptPlatformAutomerge` (success, disabled, unsupported version, 409 already scheduled, merge failure). `gitea/index.ts` and `forgejo/index.ts` are at 100% on all four coverage metrics. `pnpm type-check`, oxlint, biome, prettier, markdownlint and `test/docs` pass on the branch after merging current `main`.


RahulGautamSingh (changes requested):
No documentation added for the new env vars in docs/usage/self-hosted-experimental.md. The GitLab equivalents are documented there.

silenthooligan:
Thanks for the review. Added docs in 5bbd42e04, and also corrected the env var spelling (`ATTEMPS` -> `ATTEMPTS`) since the Gitea var is brand new with no adoption. Left the GitLab equivalent alone to avoid breaking existing deployments.

silenthooligan:
Rebased onto current `main` and pushed a follow-up commit.

Recap since the review: the docs for the two env vars were the only thing requested, and those went in back in May (5bbd42e0). Today's commit brings the rest up to date with `main`:

- `timers/promises` import now uses the `node:` protocol
- the new debug messages use the inlined form the GitHub and GitLab adapters use, per `docs/development/best-practices.md`
- the coverage-ignore comment has a justification, now that the `v8-ignore-reason` rule from #44726 requires one

Locally `pnpm check --all lib/modules/platform/gitea` and `pnpm type-check` are clean, and `lib/modules/platform/gitea` is at 100% coverage on statements, branches, functions and lines.

I also opened #45162, which mirrors both fixes onto the Forgejo adapter. It was split out of Gitea in #37075 and still has the same single-attempt merge call and the same missing `reattemptPlatformAutomerge`.

lib/modules/platform/gitea/index.ts
```diff
@@ -202,6 +206,109 @@ async function fetchRepositories({
...
+ */
+export async function reattemptPlatformAutomerge({
+  number,
+  platformPrOptions,
+}: ReattemptPlatformAutomergeConfig): Promise<void> {
+  try {
+    await tryPrAutomerge(number, platformPrOptions);
+
+    logger.debug(`PR platform automerge re-attempted...prNo: ${number}`);
+  } catch (err) /* v8 ignore next -- defensive: automerge re-attempt failures are logged and swallowed, not simulated in specs */ {
```
RahulGautamSingh: the outer catch in `reattemptPlatformAutomerge` is actually reachable: `tryPrAutomerge` swallows the merge error, but `helper.getPR` can throw and would propagate here i.e. it's testable (mock `getPR` to reject).

RahulGautamSingh (changes requested):
PR description still says `RENOVATE_X_GITEA_AUTO_MERGEABLE_CHECK_ATTEMPS` (old spelling), please fix that as well.

lib/modules/platform/gitea/index.ts
silenthooligan: Agreed. Ignore removed, test added that rejects `helper.getPR` in the poll loop so the error reaches this catch. Module stays at 100% coverage.

silenthooligan:
Both addressed in 02a185ddd.

- Removed the `v8 ignore` on `reattemptPlatformAutomerge`, added a test that reaches the outer catch via a rejecting `helper.getPR`.
- Description: `ATTEMPS` -> `ATTEMPTS`. The remaining `ATTEMPS` further down is `RENOVATE_X_GITLAB_AUTO_MERGEABLE_CHECK_ATTEMPS`, which is GitLab's actual variable name, so I left it. The Gitea var has been `ATTEMPTS` in code and docs since 5bbd42e0.

Also pulled in the spec change @viceice asked for on #45162: `vi.mock('timers/promises')` instead of shrinking the delay via env var, so the retry tests do not sleep and assert the backoff schedule instead.

`pnpm check --all lib/modules/platform/gitea` and `pnpm type-check` clean. Module at 100% statements, branches, functions and lines.

lib/modules/platform/gitea/index.ts
```diff
@@ -202,6 +206,109 @@ async function fetchRepositories({
...
+  const baseDelay = parseInteger(
+    env.RENOVATE_X_GITEA_AUTO_MERGEABLE_CHECK_DELAY,
+    250,
+  );
+
+  // Wait for Gitea to finish computing mergeable before calling merge.
+  // Just after PR creation the field is false as a placeholder while the async
+  // check runs; a follow-up getPR returns the current value.
+  for (let attempt = 1; attempt <= retryTimes; attempt += 1) {
+    const pr = await helper.getPR(config.repository, prNumber, {
```
RahulGautamSingh, line 259: The mergeable poll loop is not wrapped, only `mergePR` is. In `createPr` this runs inside the outer try, so a transient `getPR` error rethrows past the 409 handler and fails PR creation even though the PR already exists. 

Please wrap the whole body so `tryPrAutomerge` is always best-effort, like the old isolated merge catch.

lib/modules/platform/gitea/index.ts
```diff
@@ -202,6 +206,109 @@ async function fetchRepositories({
...
+    }
+    if (attempt === retryTimes) {
+      logger.debug(
+        `PR not mergeable after ${retryTimes} attempts, merging anyway...prNo: ${prNumber}`,
+      );
+      break;
+    }
+    logger.debug(
+      `PR not yet in mergeable state. Retrying ${attempt}...prNo: ${prNumber}`,
+    );
```
RahulGautamSingh, line 273: These debug lines inline prNo: while success/fail below use { prNumber } metadata. pick one to be consistent

RahulGautamSingh (changes requested):

silenthooligan:
Merged `main` after #45825. Forgejo now runs on the Gitea code. This PR fixes both and replaces #45162.

Since the last review:
- The automerge logic is in `createPlatform()`. `gitea` and `forgejo` both export `reattemptPlatformAutomerge`.
- The env vars are replaced by the experimental global option `prMergeabilityCheckAttempts`, per @secustor's suggestion on #45162.
- The whole automerge attempt is best-effort. A failing mergeable check no longer fails `createPr`.
- Logging uses `{ prNumber }` metadata throughout.
- A `409` from the merge endpoint is logged at debug as "already scheduled". Gitea does not cancel a scheduled automerge on push. Without this, every re-attempt after a branch update would warn.

lib/modules/platform/gitea/index.ts
```diff
@@ -198,6 +204,78 @@ export function createPlatform(options: GiteaPlatformOptions): GiteaPlatform {
...
+  async function tryPrAutomerge(
+    prNumber: number,
+    platformPrOptions: PlatformPrOptions | undefined,
+  ): Promise<void> {
+    if (!platformPrOptions?.usePlatformAutomerge) {
+      return;
+    }
+
+    // `delete_branch_after_merge` is required to not have undesired
+    // behavior when renovate finds existing branches on next run.
```
RahulGautamSingh: move this above the  `delete_branch_after_merge: true` statement on line 257

lib/modules/platform/gitea/index.ts
```diff
@@ -198,6 +204,78 @@ export function createPlatform(options: GiteaPlatformOptions): GiteaPlatform {
...
+    const unsupportedReason = checkNativeAutomerge(defaults.version);
+    if (unsupportedReason !== null) {
+      logger.debug({ prNumber }, unsupportedReason);
+      return;
+    }
+
+    try {
```
RahulGautamSingh: Replace lines 231-276 (the whole try block) with a single retry loop around mergePR:

```ts
    // Cap at 1 so a 0 does not skip automerge entirely.
    const attempts = Math.max(
      GlobalConfig.get('prMergeabilityCheckAttempts'),
      1,
    );
    for (let attempt = 1; attempt <= attempts; attempt += 1) {
      try {
        await helper.mergePR(http, config.repository, prNumber, {
          Do:
            getMergeMethod(
              platformPrOptions.automergeStrategy,
              config.allowedMergeMethods,
            ) ?? config.mergeMethod,
          merge_when_checks_succeed: true,
          delete_branch_after_merge: true,
        });

        logger.debug({ prNumber }, `${name}-native automerge: success`);
        return;
      } catch (err) {
        // A push does not cancel a scheduled automerge, so re-attempting after
        // a branch update is answered with 409 Conflict.
        if (err.statusCode === 409) {
          logger.debug(
            { prNumber },
            `${name}-native automerge: already scheduled`,
          );
          return;
        }

        // The platform computes `mergeable` in the background after a push and
        // answers 405 until that is done.
        if (err.statusCode !== 405 || attempt === attempts) {
          logger.warn(
            { err, prNumber, platform: id },
            'Platform-native automerge: fail',
          );
          return;
        }

        logger.debug({ prNumber, attempt }, 'PR not yet mer
        await setTimeout(MERGEABLE_CHECK_DELAY_MS * attempt ** 2);
      }
    }
```

The `helper.getPR` poll and the outer try both go away. Every path returns, so nothing escapes into `createPr`.

RahulGautamSingh (changes requested):

lib/modules/platform/gitea/index.ts
```diff
@@ -66,6 +69,9 @@ import {
   usableRepo,
 } from './utils.ts';
 
+/** Base delay between automerge attempts, grows quadratically per attempt. */
+const MERGEABLE_CHECK_DELAY_MS = 250;
```
RahulGautamSingh: MERGEABLE_CHECK_DELAY_MS no longer names what it does, there is no mergeable check anymore. 
AUTOMERGE_RETRY_DELAY_MS fits better.

RahulGautamSingh (approved):
lgtm, except this minor one

lib/modules/platform/gitea/index.ts
```diff
@@ -198,6 +204,72 @@ export function createPlatform(options: GiteaPlatformOptions): GiteaPlatform {
...
+    // At least one attempt, so a 0 does not skip automerge entirely.
+    const attempts = Math.max(
+      GlobalConfig.get('prMergeabilityCheckAttempts'),
+      1,
+    );
+    for (let attempt = 1; attempt <= attempts; attempt += 1) {
+      try {
+        await helper.mergePR(http, config.repository, prNumber, {
+          Do:
+            getMergeMethod(
```
secustor, line 234: Why are you calling this on each attempt ?

lib/modules/platform/gitea/index.ts
```diff
@@ -198,6 +204,72 @@ export function createPlatform(options: GiteaPlatformOptions): GiteaPlatform {
...
+        // answers 405 until that is done.
+        if (err.statusCode !== 405 || attempt === attempts) {
+          logger.warn(
+            { err, prNumber, platform: id },
+            'Platform-native automerge: fail',
+          );
+          return;
+        }
+
+        logger.debug({ prNumber, attempt }, 'PR not yet mergeable, retrying');
```
secustor, line 267:
```suggestion
        logger.debug({ prNumber, attempt, delay }, 'PR not yet mergeable, waiting before retrying');
```
This makes it clear that it is expected that there are no logs expected until `delay`

lib/modules/platform/gitea/index.ts
```diff
@@ -198,6 +204,72 @@ export function createPlatform(options: GiteaPlatformOptions): GiteaPlatform {
...
+    prNumber: number,
+    platformPrOptions: PlatformPrOptions | undefined,
+  ): Promise<void> {
+    if (!platformPrOptions?.usePlatformAutomerge) {
+      return;
+    }
+
+    const unsupportedReason = checkNativeAutomerge(defaults.version);
+    if (unsupportedReason !== null) {
+      logger.debug({ prNumber }, unsupportedReason);
```
secustor, line 221: This log message is not helpful

PR #43494, opened 2026-05-20
feat(datasource): Add `vcpkg`

stemann:
## Changes

Add a new `vcpkg` datasource that reads version metadata from a vcpkg registry. A vcpkg registry is a Git repository with a fixed on-disk layout: per-port version history lives in `versions/<first-letter>-/<port>.json` plus a registry-wide `versions/baseline.json` index.

The datasource clones the registry to a local cache directory and reads the per-port file from disk, with one cache entry per registry URL so multi-registry setups stay isolated. The pattern mirrors the existing [`crate`](https://github.com/renovatebot/renovate/tree/main/lib/modules/datasource/crate) datasource, which solves the same problem for Cargo registries.

Each entry in a port file uses one of four scheme keys: `version`, `version-semver`, `version-date`, `version-string`. The datasource normalizes all four into a single `Release` version string, and appends a `#N` suffix when `port-version` is greater than zero. Comparison and ordering of the resulting strings is delegated to the companion vcpkg versioning module (separate PR, linked below) via `defaultVersioning: 'vcpkg'`.

Custom registries are first-class citizens via `registryUrls`, matching the maintainer convention documented in https://learn.microsoft.com/vcpkg/maintainers/registries. The Git tree object id from each version entry is exposed as `newDigest` so downstream lockfiles can pin to an exact tree.

See https://learn.microsoft.com/vcpkg/concepts/registries for the registry model overview.

This is the first of three planned vcpkg PRs:

1. **This PR**, the `vcpkg` datasource.
1. The companion `vcpkg` versioning module that interprets the four scheme keys and the `>=` lower-bound semantics.
1. The `vcpkg` manager that extracts dependencies from `vcpkg.json` manifest files.

A registry-maintainer side companion (port `portfile.cmake` upstream tracking) is planned as a follow-up.

## Context

- [ ] This closes an existing Issue, Closes: #
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

Related:

- #7135 (tracking issue for vcpkg support)
- #25217

## AI assistance disclosure

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

[...]

Issue #7135, opened 2020-08-30
Vcpkg support


JamieMagee:
**What would you like Renovate to be able to do?**

Vcpkg is a dependency manager for C++ projects. It's still in the early stages, but it's backed by Microsoft, and is likely to be widely used.

Currently vcpkg doesn't have support for versioning, but there is currently an RFC open at microsoft/vcpkg#11758. The format of `vcpkg.json` is likely to be:

```json
{
    "name": "project", 
    "version": "1.0.0",
    "version-scheme": "semver",
    "dependencies": [
        { "name": "zlib", "version": "1.2.11" },
        { "name": "rapidjson", "version": "2020-02-08", "version-scheme": "date" }]
}
```

There are more complex examples in the PR, including minimum versions, overrides, and exclusions. 

**Did you already have any implementation ideas?**

**Are there any workarounds or alternative ideas you've tried to avoid needing this feature?**

Also need to wait for registry support to be complete microsoft/vcpkg#13038

**Is this a feature you'd be interested in implementing yourself?**

Yes, whenever the RFC is complete.

rarkins:
Is it an alternative to Conan, or somehow complementary?

JamieMagee:
It's an alternative. Currently vcpkg appears to be much earlier in development than conan is, so if I had to prioritise one I'd prioritise conan right now.

applemonkey496:
Is there any plan to implement vcpkg support at this point?

JamieMagee:
I'm not actively working on it right now, and I don't think I'll be working on it in the near-term. I'm busy with #8647.

We can give help and guidance if it's something you're interested in implementing yourself? We have a Slack for contributors if you'd like to join.

HonkingGoose:
> Currently vcpkg appears to be much earlier in development than conan is, so if I had to prioritise one I'd prioritise conan right now.

> I'm not actively working on it right now, and I don't think I'll be working on it in the near-term. I'm busy with #8647.

I'll label this `priority-4-low` then, as we wanted to prioritize the Conan package manager first, and it doesn't seem like we'll get around to working on the vcpkg manager soon.

JamieMagee:
I agree with the `priority-4-low` label, but not with Conan over vcpkg. In the year since I made that comment I think vcpkg has overtaken Conan in terms of industry use (based on files checked in on GitHub and stars on each project).

Though if support for either comes via the community, we'll obviously accept it.

applemonkey496:
[...]

PR #43497, opened 2026-05-20
feat(versioning): Add `vcpkg` versioning

stemann:
## Changes

Add a `vcpkg` versioning module for [vcpkg](https://learn.microsoft.com/vcpkg/) per-port version schemes. The vcpkg registry per-port JSON declares one of four schemes:

- `version` is a relaxed dotted-number form like `3.1.4` or `1.2.3.4`, with an optional prerelease tail.
- `version-semver` is strict [Semantic Versioning](https://semver.org).
- `version-date` is an ISO calendar date in `YYYY-MM-DD` form, optionally followed by dot-separated disambiguation identifiers like `2024-01-15.1.2`.
- `version-string` is opaque text without a defined ordering.

Each entry can also carry a `port-version` integer for packaging-only changes, written as `<base>#<port-version>` and omitted when zero. The port-version acts as a tie-breaker after the base versions compare equal, so `1.2.3#1` is greater than `1.2.3#0`, while `1.2.4` is still greater than `1.2.3#1` because the base version wins.

The module detects the scheme heuristically from the input, validates dates against the calendar, and delegates ordered numeric comparison to the existing `semver` or `loose` modules so the underlying schemes do not need to be reimplemented.

vcpkg has no range expression syntax. Manifest `dependencies` carry a single `version>=` lower bound, and `overrides` carry an exact pin. The `matches`, `getSatisfyingVersion`, and `minSatisfyingVersion` functions implement the `>=` constraint for numeric and date schemes. Opaque strings have no ordering on the base, so a constraint over an opaque string is satisfied only by versions with the same base, with port-version still providing a `>=` comparison once the base matches.

This is a building block for the planned `vcpkg` datasource and manager.

Reference: https://learn.microsoft.com/vcpkg/users/versioning

## Context

Related:

- #7135 (tracking issue for vcpkg support)
- #25217

## AI assistance disclosure

Added `lib/modules/versioning/vcpkg/readme.md`.

The public repository: https://github.com/stemann/renovate-vcpkg-example

**Unit tests:** 131 table-driven cases covering each scheme, cross-scheme behavior, port-version interactions, date disambiguation tails, and the `>=` matches semantics. 100 % statements / lines / functions / branches.

**Real repository:** [`stemann/renovate-vcpkg-example@vcpkg-versioning`](https://github.com/stemann/renovate-vcpkg-example/tree/vcpkg-versioning) wires a `customManagers` regex on a `vcpkg.json` that pins `zlib` with `version>=: "1.2.13"`, with both `datasourceTemplate: vcpkg` and `versioningTemplate: vcpkg` set. The regex targets the companion `vcpkg` datasource ([#43494](https://github.com/renovatebot/renovate/pull/43494)) so the versioning module is exercised in its production context, with the `>=` matches semantics selecting the highest satisfying release. Running Renovate from a local branch combining this PR with #43494 produces:

    "managers": {"regex": {"fileCount": 1, "depCount": 1}}
    "depName": "zlib"
    "datasource": "vcpkg"
    "versioning": "vcpkg"
    "currentValue": "1.2.13"
    "newVersion": "1.3.2"
    "newValue": "1.3.2"
    "updateType": "minor"
    "registryUrl": "https://github.com/microsoft/vcpkg"

Reproducer:

    RENOVATE_TOKEN=<gh-pat> \
      RENOVATE_BASE_BRANCH_PATTERNS='["vcpkg-versioning"]' \
      RENOVATE_USE_BASE_BRANCH_CONFIG=merge \
      pnpm start \
        --platform=github \
        --dry-run=lookup \
        --autodiscover=false \
        stemann/renovate-vcpkg-example

### DIFF
--- a/lib/modules/versioning/vcpkg/index.spec.ts
+++ b/lib/modules/versioning/vcpkg/index.spec.ts
@@ -1,0 +1,340 @@
+import { coerce } from 'semver';
+import vcpkg from './index.ts';
+
+describe('modules/versioning/vcpkg/index', () => {
+  describe('.isValid(input)', () => {
+    it.each`
+      input                             | expected
+      ${'1.2.3'}                        | ${true}
+      ${'1.2.3-rc.1'}                   | ${true}
+      ${'1.2.3-rc.1+build.7'}           | ${true}
+      ${'1.2.3#1'}                      | ${true}
+      ${'1.2.3#0'}                      | ${true}
+      ${'1.2.3#42'}                     | ${true}
+      ${'1.2'}                          | ${true}
+      ${'1'}                            | ${true}
+      ${'1.2.3.4'}                      | ${true}
+      ${'1.2.3.4.5'}                    | ${true}
+      ${'1.2-alpha'}                    | ${true}
+      ${'2024-01-15'}                   | ${true}
+      ${'2024-01-15#3'}                 | ${true}
+      ${'2024-01-15.1'}                 | ${true}
+      ${'2024-01-15.0'}                 | ${true}
+      ${'2024-01-15.1.2.3'}             | ${true}
+      ${'2024-02-29'}                   | ${true}
+      ${'2023-02-29'}                   | ${false}
+      ${'2024-13-01'}                   | ${false}
+      ${'2024-02-31'}                   | ${false}
+      ${'2024-1-1'}                     | ${true}
+      ${'bla-bla-2024-08-fixed-typo'}   | ${true}
+      ${'bla-bla-2024-08-fixed-typo#2'} | ${true}
+      ${'2024.08.fixed'}                | ${true}
+      ${'opaque'}                       | ${true}
+      ${'opaque#5'}                     | ${true}
+      ${''}                             | ${false}
+      ${'   '}                          | ${false}
+      ${'#1'}                           | ${false}
+      ${'1.2.3#'}                       | ${true}
+      ${'1.2.3#abc'}                    | ${true}
+      ${null as unknown as string}      | ${false}
+      ${undefined as unknown as string} | ${false}
+    `('isValid("$input") === $expected', ({ input, expected }) => {
+      expect(vcpkg.isValid(input)).toBe(expected);
+    });
+
+    it('rejects non-string input', () => {
+      expect(vcpkg.isValid(123 as unknown as string)).toBeFalse();
+    });
+  });
+
+  describe('.isVersion(input)', () => {
+    it.each`
+      input                      | expected
+      ${'1.2.3'}                 | ${true}
+      ${'1.2'}                   | ${true}
+      ${'1.2.3.4'}               | ${true}
+      ${'2024-01-15'}            | ${true}
+      ${'2024-01-15#3'}          | ${true}
+      ${'bla-bla-2024-08-fixed'} | ${false}
+      ${'opaque#3'}              | ${false}
+      ${''}                      | ${false}
+      ${null}                    | ${false}
+      ${undefined}               | ${false}
+    `('isVersion("$input") === $expected', ({ input, expected }) => {
+      expect(vcpkg.isVersion(input)).toBe(expected);
+    });
+  });
+
+  describe('.isCompatible(version)', () => {
+    it('mirrors isValid', () => {
+      expect(vcpkg.isCompatible('1.2.3')).toBeTrue();
+      expect(vcpkg.isCompatible('1.2.3#7')).toBeTrue();
+      expect(vcpkg.isCompatible('opaque')).toBeTrue();
+      expect(vcpkg.isCompatible('')).toBeFalse();
+    });
+  });
+
+  describe('.isSingleVersion(version)', () => {
+    it('returns true for ordered schemes only', () => {
+      expect(vcpkg.isSingleVersion('1.2.3')).toBeTrue();
+      expect(vcpkg.isSingleVersion('1.2.3.4')).toBeTrue();
+      expect(vcpkg.isSingleVersion('2024-01-15')).toBeTrue();
+      expect(vcpkg.isSingleVersion('opaque')).toBeFalse();
+    });
+  });
+
+  describe('.isStable(version)', () => {
+    it.each`
+      input             | expected
+      ${'1.2.3'}        | ${true}
+      ${'1.2.3-rc.1'}   | ${false}
+      ${'1.2.3-rc.1#1'} | ${false}
+      ${'1.2'}          | ${true}
+      ${'1.2-alpha'}    | ${false}
+      ${'2024-01-15'}   | ${true}
+      ${'opaque-tag'}   | ${true}
+      ${'bogus_input!'} | ${true}
+      ${''}             | ${false}
+    `('isStable("$input") === $expected', ({ input, expected }) => {
+      expect(vcpkg.isStable(input)).toBe(expected);
+    });
+  });
+
+  describe('.getMajor/getMinor/getPatch(version)', () => {
+    it.each`
+      input           | major   | minor   | patch
+      ${'1.2.3'}      | ${1}    | ${2}    | ${3}
+      ${'1.2.3#4'}    | ${1}    | ${2}    | ${3}
+      ${'1.2.3-rc.1'} | ${1}    | ${2}    | ${3}
+      ${'1.2'}        | ${1}    | ${2}    | ${0}
+      ${'1.2.3.4'}    | ${1}    | ${2}    | ${3}
+      ${'5'}          | ${5}    | ${0}    | ${0}
+      ${'2024-01-15'} | ${null} | ${null} | ${null}
+      ${'opaque'}     | ${null} | ${null} | ${null}
+      ${''}           | ${null} | ${null} | ${null}
+    `('getMajor("$input")', ({ input, major, minor, patch }) => {
+      expect(vcpkg.getMajor(input)).toBe(major);
+      expect(vcpkg.getMinor(input)).toBe(minor);
+      expect(vcpkg.getPatch(input)).toBe(patch);
+    });
+
+    it('accepts SemVer object input', () => {
+      const sv = coerce('1.2.3')!;
+      expect(vcpkg.getMajor(sv)).toBe(1);
+      expect(vcpkg.getMinor(sv)).toBe(2);
+      expect(vcpkg.getPatch(sv)).toBe(3);
+    });
+  });
+
+  describe('.equals(a, b)', () => {
+    it.each`
+      a                  | b                  | expected
+      ${'1.2.3'}         | ${'1.2.3'}         | ${true}
+      ${'1.2.3'}         | ${'1.2.3#0'}       | ${true}
+      ${'1.2.3#0'}       | ${'1.2.3'}         | ${true}
+      ${'1.2.3#1'}       | ${'1.2.3#1'}       | ${true}
+      ${'1.2.3#1'}       | ${'1.2.3#2'}       | ${false}
+      ${'1.2.3'}         | ${'1.2.4'}         | ${false}
+      ${'1.2.3+build.7'} | ${'1.2.3+build.9'} | ${true}
+      ${'1.2'}           | ${'1.2.0'}         | ${false}
+      ${'2024-01-15'}    | ${'2024-01-15'}    | ${true}
+      ${'2024-01-15'}    | ${'2024-01-15#0'}  | ${true}
+      ${'2024-01-15#1'}  | ${'2024-01-16#1'}  | ${false}
+      ${'opaque'}        | ${'opaque'}        | ${true}
+      ${'opaque#1'}      | ${'opaque#1'}      | ${true}
+      ${'opaque#1'}      | ${'opaque#2'}      | ${false}
+      ${'opaque'}        | ${'other'}         | ${false}
+      ${'1.2.3'}         | ${'2024-01-15'}    | ${false}
+      ${'1.2.3'}         | ${'opaque'}        | ${false}
+      ${''}              | ${'1.2.3'}         | ${false}
+      ${'1.2.3'}         | ${''}              | ${false}
+    `('equals("$a", "$b") === $expected', ({ a, b, expected }) => {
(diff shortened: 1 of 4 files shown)


lib/modules/versioning/vcpkg/index.spec.ts
```diff
@@ -0,0 +1,340 @@
...
+      ${'opaque'}     | ${null} | ${null} | ${null}
+      ${''}           | ${null} | ${null} | ${null}
+    `('getMajor("$input")', ({ input, major, minor, patch }) => {
+      expect(vcpkg.getMajor(input)).toBe(major);
+      expect(vcpkg.getMinor(input)).toBe(minor);
+      expect(vcpkg.getPatch(input)).toBe(patch);
+    });
+
+    it('accepts SemVer object input', () => {
+      const sv = coerce('1.2.3')!;
```
viceice, line 122: why coerce? 🤔

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  // always yields three numeric parts.
+  const [yearStr, monthStr, dayStr] = head.split('-');
+  const year = Number.parseInt(yearStr, 10);
+  const month = Number.parseInt(monthStr, 10);
+  const day = Number.parseInt(dayStr, 10);
+  if (month < 1 || month > 12 || day < 1) {
+    return false;
+  }
+  // Manually validate day-of-month against actual month length.
+  const daysInMonth = new Date(Date.UTC(year, month, 0)).getUTCDate();
```
viceice, line 62: use luxon to validate

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  const strictSemver =
+    scheme === 'numeric' &&
+    strictSemverRegex.test(base) &&
+    semver.isValid(base);
+  if (scheme === 'date') {
+    const dateParts = matchDate(base);
+    /* v8 ignore next 3 -- detectScheme guarantees matchDate succeeds */
+    if (!dateParts) {
+      return null;
+    }
```
viceice, line 118:
```suggestion
    // detectScheme guarantees matchDate succeeds
    const dateParts = matchDate(base)!;
```

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  }
+  return a.portVersion - b.portVersion;
+}
+
+function isValid(input: string): boolean {
+  return parse(input) !== null;
+}
+
+function isVersion(input: string | undefined | null): boolean {
+  const parsed = parse(input ?? '');
```
viceice, line 184:
```suggestion
  const parsed = parse(input);
```

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  if (!parsed) {
+    return false;
+  }
+  return parsed.scheme !== 'string';
```
viceice, line 188: date and numeric scheme are not versions? I think that will cause issues later 🤔

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+function isStable(version: string): boolean {
+  const parsed = parse(version);
+  if (!parsed) {
+    return false;
+  }
+  if (parsed.scheme === 'numeric') {
+    // A pre-release suffix (`-...`) marks the version as unstable.
+    // Pure build metadata (`+...`) is ignored, matching SemVer.
+    const withoutBuild = parsed.base.replace(numericBuildRegex, '');
+    return !numericPrereleaseRegex.test(withoutBuild);
```
viceice, line 208: use semver-stable package?

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  }
+  // Strip any pre-release or build suffix before coercion so loose forms like
+  // `1.2-alpha` map to `1.2.0` instead of failing.
+  const stripped = base
+    .replace(numericBuildRegex, '')
+    .replace(numericPrereleaseRegex, '');
+  return coerce(stripped);
+}
+
+function getMajor(version: string | SemVer): number | null {
```
viceice, line 226: semver should never be passed to this function. where did that come from?

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  const input = typeof version === 'string' ? version : version.version;
+  const parsed = parse(input);
+  if (parsed?.scheme !== 'numeric') {
+    return null;
+  }
+  /* v8 ignore next -- coerce always succeeds for `numeric` bases */
+  return coerceNumeric(parsed.base, parsed.strictSemver)?.major ?? null;
```
viceice, line 233:
```suggestion
  /* coerce always succeeds for `numeric` bases */
  return coerceNumeric(parsed.base, parsed.strictSemver)!.major;
```
or use `coerceInt` util for fail safety

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  const parsed = parse(input);
+  if (parsed?.scheme !== 'numeric') {
+    return null;
+  }
+  /* v8 ignore next -- coerce always succeeds for `numeric` bases */
+  return coerceNumeric(parsed.base, parsed.strictSemver)?.patch ?? null;
+}
+
+function equals(version: string, other: string): boolean {
+  const a = parse(version);
```
viceice, line 257: why not simply call `compare(a,b)==0`?

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  let bestParsed: ParsedVersion | null = null;
+  for (const version of versions) {
+    if (!matches(version, range)) {
+      continue;
+    }
+    const parsed = parse(version);
+    /* v8 ignore next 3 -- parse succeeds for any matches-accepted input */
+    if (!parsed) {
+      continue;
+    }
```
viceice, line 336:
```suggestion
    // parse succeeds for any matches-accepted input
    const parsed = parse(version)!;
```

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  range: string,
+): string | null {
+  let bestStr: string | null = null;
+  let bestParsed: ParsedVersion | null = null;
+  for (const version of versions) {
+    if (!matches(version, range)) {
+      continue;
+    }
+    const parsed = parse(version);
+    /* v8 ignore next 3 -- parse succeeds for any matches-accepted input */
```
viceice, line 356: see other comment

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  return coerceNumeric(parsed.base, parsed.strictSemver)?.major ?? null;
+}
+
+function getMinor(version: string | SemVer): number | null {
+  const input = typeof version === 'string' ? version : version.version;
+  const parsed = parse(input);
+  if (parsed?.scheme !== 'numeric') {
+    return null;
+  }
+  /* v8 ignore next -- coerce always succeeds for `numeric` bases */
```
viceice, line 242: see other comment

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,390 @@
...
+  return coerceNumeric(parsed.base, parsed.strictSemver)?.minor ?? null;
+}
+
+function getPatch(version: string | SemVer): number | null {
+  const input = typeof version === 'string' ? version : version.version;
+  const parsed = parse(input);
+  if (parsed?.scheme !== 'numeric') {
+    return null;
+  }
+  /* v8 ignore next -- coerce always succeeds for `numeric` bases */
```
viceice, line 252: see other comment

lib/modules/versioning/vcpkg/index.ts
RahulGautamSingh, line 62: Replace with `DateTime.utc(year, month, 1).daysInMonth` from Luxon.

lib/modules/versioning/vcpkg/readme.md
```diff
@@ -0,0 +1,16 @@
...
+
+Renovate detects which scheme applies by inspecting the input, then delegates to the matching underlying versioning module (`semver`, `loose`, or a direct date comparison).
+
+A vcpkg port can also carry a `port-version` integer for changes that affect packaging without changing the upstream source.
+The canonical text form combines the base version and port-version with a `#` separator, like `1.2.3#1`, and the port-version is omitted when zero.
+Renovate treats the port-version as a tie-breaker after the base versions compare equal, so `1.2.3#1` is greater than `1.2.3#0`, while `1.2.4` is still greater than `1.2.3#1` because the base version wins.
+
+Vcpkg has no range expression syntax. Manifest `dependencies` carry a single `version>=` lower bound, and `overrides` carry an exact pin. Renovate treats `matches` as a `>=` comparison for the numeric and date schemes, so any candidate version at or above the constraint satisfies. Opaque strings have no ordering on the base, so only versions with the same base satisfy, with port-version still providing the `>=` comparison once the base matches.
```
RahulGautamSingh, line 16: one sentence per line in Markdown files.

lib/modules/versioning/vcpkg/index.ts
stemann: `semver-stable` only accepts strict semver, so relaxed `version` forms like `1.2.3.4` would read as unstable. I'd suggest to keep the regex.

lib/modules/versioning/vcpkg/index.ts
stemann: Used `DateTime.fromISO(head, { zone: 'utc' }).isValid` instead - the regex already pins `YYYY-MM-DD`, so validating the whole date drops the manual parsing entirely.

stemann: Used `DateTime.fromISO(head, { zone: 'utc' }).isValid` instead - the regex already pins `YYYY-MM-DD`, so validating the whole date drops the manual parsing entirely.

lib/modules/versioning/vcpkg/index.ts
stemann: They are - `scheme !== 'string'` excluded only opaque strings. But those were `isValid` and not `isVersion`, i.e. signalled as a range on a module with `supportsRanges = false`. `isVersion` now accepts any valid version.

lib/modules/versioning/vcpkg/index.ts
stemann: `compare` conflated equal with unorderable. It now returns `null` for unorderable, so `equals` is `compare(a, b) === 0`. Side effect: `isGreaterThan('opaque#2', 'opaque#1')` is now true, consistent with `matches` — spec updated.

lib/modules/versioning/vcpkg/index.spec.ts
stemann: Only to build a `SemVer` for the union parameter. Test dropped along with the narrowing.

lib/modules/versioning/vcpkg/index.ts
stemann: From `VersioningApi`, which declares `getMajor(version: string | SemVer)`. Narrowed to `string`.

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,338 @@
...
+  return coerce(stripped);
+}
+
+function getMajor(version: string): number | null {
+  const parsed = parse(version);
+  if (parsed?.scheme !== 'numeric') {
+    return null;
+  }
+  /* coerce always succeeds for `numeric` bases */
+  return coerceNumeric(parsed.base, parsed.strictSemver)!.major;
```
RahulGautamSingh, line 219: /* coerce always succeeds for numeric bases */ is not true. 

`numericRegex` accepts any digit run, but `semver.coerce` caps components at 16 digits and returns null above that:

```
getMajor('12345678901234567')    THREW Cannot read properties of null (reading 'major')
getMajor('99999999999999999999') THREW Cannot read properties of null (reading 'major')
```

Renovate calls `getMajor` on every release a datasource returns, so one long numeric version takes down the lookup. Handle the null (coerceInt, or return null).

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,338 @@
...
+  const a = parse(version);
+  const b = parse(other);
+  if (!a || !b) {
+    return 0;
+  }
+  return compare(a, b) ?? 0;
+}
+
+function matches(version: string, range: string): boolean {
+  // vcpkg has no range syntax; manifest dependencies are a `>=` lower bound.
```
RahulGautamSingh, line 266: matches as >= silently widens `allowedVersions`

so `allowedVersions: '1.2.3'` allows every version at or above `1.2.3` instead of pinning it.

lib/modules/versioning/vcpkg/index.ts
RahulGautamSingh, line 1: `isValid` and `isVersion` have identical bodies, and `isCompatible` calls one while `isSingleVersion` calls the other. Collapse to one.

lib/modules/versioning/vcpkg/index.ts
```diff
@@ -0,0 +1,338 @@
...
+  }
+  const result = compare(a, b);
+  return result !== null && result > 0;
+}
+
+function sortVersions(version: string, other: string): number {
+  const a = parse(version);
+  const b = parse(other);
+  if (!a || !b) {
+    return 0;
```
RahulGautamSingh, line 260: returning 0 for incomparable pairs is not a valid comparator.

A vcpkg port that switched its version field at some point (version-string → version, or version → version-date) gives you mixed versions..this won't handle that

RahulGautamSingh (changes requested):

PR #43511, opened 2026-05-21
docs: add a deep dive on our cache layers

jamietanna:
## Changes

- [ ] What is/isn't concurrency safe? Can S3/Redis be used by multiple Renovate instances at once?
- [ ]  What happens if inherited/extends config changes, and the repo hasn't changed?
- [ ] Why does the PR cache include all PRs on GitHub?https://github.com/renovatebot/renovate/discussions/43360
- [ ] https://github.com/renovatebot/renovate/discussions/29058

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: #41456 
- [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #43523, opened 2026-05-21
feat(config): rename `repositoryCacheHttpCacheTtlDays`

jamietanna:
## Changes

While working on #43511, it wasn't clear which cache(s) the
`httpCacheTtlDays` config option was applicable to.

Given this only affects the Repository Cache (and is unlikely to ever be
used for the Package Cache, as we have `cacheTtlOverride`), we can
rename it to make it explicit.

As it's being renamed, we need to make sure we have it handled as a
migration.

## Context

## AI assistance disclosure

The public repository: <URL>


viceice (commented):
I don't like the duplicate cache in name, otherwise LGTM 🤔

jamietanna:
Maybe `repositoryCacheHttpTtlDays` or `repositoryCacheMaximumHttpDays` or something like that?

(Keeping it duplicated does make it more discoverable when folks are looking for docs after the rename, but I agree that duplication isn't always helpful)

jamietanna:
I wonder if we should also do something similar with `cacheTtlOverride` and other options (via #43511) to make it super clear where they're used?

viceice:
sounds good to rename the others too.

`repositoryCacheMaximumHttpDays` sounds better

jamietanna:
Hmm, this is a bit harder with options like `cacheTtlOverride`, which maybe also want to be renamed 🤔

PR #43524, opened 2026-05-21
fix(manager/hermit): propagate git credentials to hermit install

### LINKED ISSUE #34544: Set git auth variables in Hermit manager to support private package updating

Hermit manager calls `git` under the hood, but currently no git credentials are passed through. They should be passed through similar to how Cargo does it: https://github.com/renovatebot/renovate/blob/3169d1a9dd55e2003a25cc24943709b7ad63bc55/lib/modules/manager/cargo/artifacts.ts#L35

wpfleger96:
## Summary

- Spread `getGitEnvironmentVariables()` into `extraEnv` on the Hermit manager's `ExecOptions`, matching the pattern already used by cargo, gomod, poetry, copier, vendir, and conan managers
- Fix `undefined` stderr bug in the outer catch block — non-`UpdateHermitError` exceptions rendered literal "undefined" in PR comments because the code unconditionally cast `err` to `UpdateHermitError`. Added an `instanceof` guard with a proper fallback using `err.message`
- Add test that mocks `hostRules` and asserts `GIT_CONFIG_*` env vars appear in exec snapshots (follows the copier test pattern)
- Document in `readme.md` that git credentials from `hostRules` are propagated to `hermit install`

## Context

`hermit install` runs without git credentials in its subprocess environment, so packages hosted in private repos (or any git-authenticated source) fail silently. This is the root cause of Hermit artifact update failures reported in #34544.

The fix follows the exact same pattern merged in PR #31931 (copier) and already present in 5 other managers.

## Disclosure

This PR was authored with AI assistance. CLA signature is on the commit.

Closes #34544


lib/modules/manager/hermit/readme.md
```diff
@@ -9,6 +9,8 @@ GITHUB_TOKEN
 
 These environment variable could be passed on via setting it in `customEnvironmentVariables`.
 
+Git credentials configured in `hostRules` are also automatically propagated to `hermit install` via `GIT_CONFIG_*` environment variables. This enables Hermit to fetch packages from private Git repositories without additional configuration.
```
RahulGautamSingh: one line per sentence

lib/modules/manager/hermit/artifacts.ts
```diff
@@ -219,6 +228,7 @@ async function updateHermitPackage(update: UpdateArtifact): Promise<void> {
   const execOptions: ExecOptions = {
     docker: {},
     cwdFile: update.packageFileName,
+    extraEnv: { ...getGitEnvironmentVariables(['hermit']) },
```
RahulGautamSingh:
```suggestion
    extraEnv: { getGitEnvironmentVariables(['hermit']) },
```

RahulGautamSingh (changes requested):

RahulGautamSingh (approved):
lgtm

secustor (approved):

PR #43551, opened 2026-05-22
fix(manager/circleci): parse config files as YAML 1.1

esetnik:
## Description

The `circleci` manager parsed config files as YAML 1.2, but CircleCI's own config processor uses YAML 1.1. When a config relies on YAML 1.1 features the parse fails, Renovate skips the entire file, and every `docker` and `orb` dep in it is silently dropped.

The most common trigger is a mapping with more than one `<<` merge key, which YAML 1.2 rejects as a duplicate key. CircleCI is fine with it.

This PR passes `{ version: '1.1' }` to `parseSingleYaml`, which:

- enables anchor merge keys (`<<`) natively
- swaps to the YAML 1.1 implicit-type schema, matching how CircleCI itself parses the file
- adds a regression test that covers the duplicate-merge-key case (the test asserts the deps list, not just that parsing didn't throw — the original bug was a *silent* skip, not a thrown error)

The previous revision of this PR documented a `%YAML 1.1` directive workaround; per review feedback (thanks @jamietanna and @secustor) the readme addition is reverted in favor of the actual fix.

Context: discussion #35786.

### DIFF
--- a/lib/modules/manager/circleci/readme.md
+++ b/lib/modules/manager/circleci/readme.md
@@ -2,6 +2,24 @@ The `circleci` manager extracts both `docker` as well as `orb` datasources from 
 
 If you need to change the versioning format, read the [versioning](../../versioning/index.md) documentation to learn more.
 
+### YAML aliases and merge keys
+
+Renovate parses CircleCI config files as YAML 1.2.
+Merge keys (`<<`) are a [YAML 1.1](https://yaml.org/type/merge.html) feature and are not part of YAML 1.2.
+A config that relies on merge keys can therefore fail to parse — for example, a mapping with more than one `<<` key is rejected as a duplicate key under YAML 1.2.
+When parsing fails, Renovate skips the entire file and misses every `docker` and `orb` dependency in it.
+
+To keep merge keys working, add a `%YAML 1.1` directive as the first line of the file:
+
+```yaml
+%YAML 1.1
+---
+version: 2.1
+# ... rest of your config
+```
+
+CircleCI itself processes merge keys regardless of the directive, so this does not change how your pipeline runs.
+
 ### Private orbs
 
 To get private orbs working you should:
--- a/lib/modules/manager/circleci/readme.md
+++ b/lib/modules/manager/circleci/readme.md
@@ -2,24 +2,6 @@ The `circleci` manager extracts both `docker` as well as `orb` datasources from 
 
 If you need to change the versioning format, read the [versioning](../../versioning/index.md) documentation to learn more.
 
-### YAML aliases and merge keys
-
-Renovate parses CircleCI config files as YAML 1.2.
-Merge keys (`<<`) are a [YAML 1.1](https://yaml.org/type/merge.html) feature and are not part of YAML 1.2.
-A config that relies on merge keys can therefore fail to parse — for example, a mapping with more than one `<<` key is rejected as a duplicate key under YAML 1.2.
-When parsing fails, Renovate skips the entire file and misses every `docker` and `orb` dependency in it.
-
-To keep merge keys working, add a `%YAML 1.1` directive as the first line of the file:
-
-```yaml
-%YAML 1.1
----
-version: 2.1
-# ... rest of your config
-```
-
-CircleCI itself processes merge keys regardless of the directive, so this does not change how your pipeline runs.
-
 ### Private orbs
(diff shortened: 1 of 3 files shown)


lib/modules/manager/circleci/readme.md
```diff
@@ -2,6 +2,24 @@ The `circleci` manager extracts both `docker` as well as `orb` datasources from
 
 If you need to change the versioning format, read the [versioning](../../versioning/index.md) documentation to learn more.
 
+### YAML aliases and merge keys
+
+Renovate parses CircleCI config files as YAML 1.2.
```
jamietanna: Could we instead force Renovate to parse it as YAML 1.1, if that's what CircleCI requires?

esetnik: Good question, I'll look into it.

jamietanna: Thanks! Would be best if we don't need to document a workaround :D

secustor: We can set document options to 1.1 by default for `circleci` which also controls the merging behaviour of `yaml` 
https://eemeli.org/yaml/#parse-options

esetnik: Good call — done in fe19995ee. The extractor now passes `{ version: '1.1' }` to `parseSingleYaml`, which matches CircleCI's own behavior and removes the need for the workaround docs.

Added a regression test that exercises the original failure case (a mapping with two `<<` merge keys); verified it fails without the fix (whole file silently skipped) and passes with it. The readme addition from this branch's first commit is reverted as part of the same commit.

esetnik:
[Written by Claude]

While auditing for the same blindspot in other YAML-parsing managers, I surveyed all 31 `parseSingleYaml` / `parseYaml` call sites in `lib/modules/manager` (+ workers). Three look like they'd benefit from the same `{ version: '1.1' }` treatment as `circleci`:

| Manager | Upstream parser | Current state | Issue |
|---|---|---|---|
| `gitlabci` + `gitlabci-include` | Ruby/Psych (1.1) | `{ uniqueKeys: false }` — heavier than needed; disables duplicate-key detection across the board | #43608 |
| `travis` | Ruby/Psych (1.1) | None | #43609 |
| `jenkins` | SnakeYAML (1.1) | None — lower-impact, plugin files rarely use merge keys | #43610 |

Decision rule: **file only if (upstream parses YAML 1.1) AND (file format idiomatically uses YAML 1.1 features in the wild).** Both conditions have to hold.

<details>
<summary>Full audit — all 31 call sites, why each was/wasn't filed</summary>

**Already YAML 1.2 upstream** (go-yaml v3, eemeli/yaml, compose-go) — no mismatch, left as-is:

- `crossplane`, `fleet`, `flux`, `glasskube`, `helm-requirements`, `helm-values`, `helmfile` (extract + artifacts), `helmsman`, `helmv3` (extract + artifacts), `kustomize`, `sveltos`, `tekton`, `vendir` — Kubernetes/Helm ecosystem on go-yaml v3
- `docker-compose` — compose-go is 1.2; also already passes `removeTemplates: true`
- `ocb` — OpenTelemetry Collector Builder, Go/Cobra, 1.2
- `woodpecker`, `velaci` — Go-based CI, less merge-key idiom

**YAML 1.1 upstream but file format doesn't use 1.1 features idiomatically**:

- `pre-commit` — PyYAML is 1.1 but `.pre-commit-config.yaml` is flat lists with no anchor culture. Theoretical risk only; would file if a user reports it.

**Same YAML library both sides** (eemeli/yaml on both Renovate and the tool) — version is identical on both ends, no mismatch possible:

- `npm/extract/pnpm.ts`, `npm/post-update/pnpm.ts` — pnpm
- `npm/post-update/index.ts` — Yarn `.yarnrc.yml`

**Not a manager extract**:

- `custom/jsonata/index.ts` — user supplies arbitrary input; Renovate shouldn't pick a YAML version for them
- `workers/global/config/parse/util.ts` — Renovate's own config file; should follow documented spec

**The case we just fixed**:

- `circleci` — this PR

</details>

@esetnik can open a separate PR for each of the three filed issues if any of those would be useful — happy to take direction on order or to combine them. Not blocking this PR.

viceice (approved):

secustor (approved):

PR #43566, opened 2026-05-24
feat(datasource/docker): add opt-in metadata timestamps

Ryangr0:
## Changes

This PR keeps Renovate's default OCI release timestamp behavior registry-controlled, while adding an explicit opt-in path for registries that do not expose supported publish timestamps for container images.

The new experimental `dockerReleaseTimestampSource=metadata` option allows non-Docker-Hub OCI registries to use client-controlled OCI metadata as a best-effort `releaseTimestamp` fallback:

1. Docker Hub remains unchanged and uses the registry-controlled `tag_last_pushed` value.
2. Non-Docker-Hub registries use metadata only when `dockerReleaseTimestampSource=metadata` is configured.
3. Metadata lookup checks `org.opencontainers.image.created` on the resolved image manifest first.
4. If the manifest annotation is absent or invalid, lookup falls back to the image config blob `created` field.
5. Invalid metadata timestamps are ignored.
6. Metadata lookup is bounded to the newest Docker releases to avoid unbounded manifest/blob requests.

Because OCI annotations and image config fields are publisher/build-system controlled, this PR keeps metadata timestamps opt-in instead of silently weakening `minimumReleaseAge` for all users.

Related issues:

Refs #37196  
Refs #38656  
Refs #38659

https://github.com/project-zot/zui/pull/502#issue-3881457672

## Context

> It doesn't close the previously referred issues and discussions because the security concern is not directly addressed, but it does provide a decent workaround for stalling PR's.

## AI assistance disclosure

- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation). --- Github Copilot with ChatGPT 5.5. (Logs available on request)

The public repository: <URL>

### DIFF
--- a/lib/modules/datasource/docker/index.ts
+++ b/lib/modules/datasource/docker/index.ts
@@ -77,7 +77,7 @@ const defaultConfig = {
 };
 
 const ociCreatedLabel = 'org.opencontainers.image.created';
-const maxOciCreatedTimestampLookups = 20;
+const maxMetadataTimestampLookups = 20;
 
 export class DockerDatasource extends Datasource {
   static readonly id = dockerDatasourceId;
@@ -90,7 +90,7 @@ export class DockerDatasource extends Datasource {
 
   override readonly releaseTimestampSupport = true;
   override readonly releaseTimestampNote =
-    'Docker Hub release timestamps are determined from the `tag_last_pushed` field in the results. Other OCI registries may use the `org.opencontainers.image.created` manifest annotation for the newest releases. **NOTE**: Currently, Docker Hub digests will receive the same release timestamp as the `tag_last_pushed`, which means that digests may appear newer than they are - see https://github.com/renovatebot/renovate/issues/38659';
+    'Docker Hub release timestamps are determined from the `tag_last_pushed` field in the results. Other registries only receive release timestamps from OCI image metadata if `dockerReleaseTimestampSource=metadata` is enabled. **NOTE**: Currently, Docker Hub digests will receive the same release timestamp as the `tag_last_pushed`, which means that digests may appear newer than they are - see https://github.com/renovatebot/renovate/issues/38659';
   override readonly sourceUrlSupport = 'package';
   override readonly sourceUrlNote =
     'The source URL is determined from the `org.opencontainers.image.source` and `org.label-schema.vcs-url` labels present in the metadata of the **latest stable** image found on the Docker registry.';
@@ -1130,7 +1130,7 @@ export class DockerDatasource extends Datasource {
     );
   }
 
-  private async getOciCreatedReleaseTimestamp(
+  private async getMetadataReleaseTimestamp(
     registryHost: string,
     dockerRepository: string,
     tag: string,
@@ -1141,21 +1141,36 @@ export class DockerDatasource extends Datasource {
         dockerRepository,
         tag,
       );
-      if (!manifest || !('annotations' in manifest)) {
+      if (!manifest) {
         return null;
       }
 
-      return asTimestamp(manifest.annotations?.[ociCreatedLabel]);
+      if ('annotations' in manifest) {
+        const annotationTimestamp = asTimestamp(
+          manifest.annotations?.[ociCreatedLabel],
+        );
+        if (annotationTimestamp) {
+          return annotationTimestamp;
+        }
+      }
+
+      const configResponse = await this.getImageConfig(
+        registryHost,
+        dockerRepository,
+        manifest.config.digest,
+      );
+
+      return asTimestamp(configResponse?.body.created);
     } catch (err) {
       logger.debug(
         { registryHost, dockerRepository, tag, err },
-        'Could not determine OCI created release timestamp',
+        'Could not determine Docker metadata release timestamp',
       );
       return null;
     }
   }
 
-  private async addOciCreatedReleaseTimestamps(
+  private async addMetadataReleaseTimestamps(
     registryHost: string,
     dockerRepository: string,
     releases: Release[],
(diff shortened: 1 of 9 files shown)


jamietanna:
Please follow our PR template, as requested by `AGENTS.md` and when creating a PR

Ryangr0:
> Please follow our PR template, as requested by `AGENTS.md` and when creating a PR

This is a draft, not remotely ready to be looked at. Just trying to understand the problem first. If this is a direction you'd be interested in exploring further I'd be willing to take a more serious look at a solution.

lib/modules/datasource/docker/readme.md
```diff
@@ -13,4 +13,20 @@ Here's an example from our `renovate/renovate` Dockerfile:
 LABEL org.opencontainers.image.source="https://github.com/renovatebot/renovate"
 ```
 
+## Release timestamps
+
+Renovate uses registry-controlled Docker Hub `tag_last_pushed` values as Docker release timestamps by default.
+Other Docker and OCI registries often do not expose registry-controlled publish timestamps through the OCI Distribution API.
```
polarathene, line 19: FWIW while "Docker" containers is quite widespread (_similar to "SSL" when modern use has been "TLS" for quite some time_), this isn't specific to Docker images (_a legacy image format_) but I guess that's more of an issue with the whole datasource module name itself too.

The OCI Distribution API doesn't have anything related to the registry publish timestamps information AFAIK. I'm not familiar with how the content here was derived, but earlier discussion on this PR mentions AI use, so perhaps this is an error from an LLM summarizing?

lib/modules/datasource/docker/readme.md
```diff
@@ -13,4 +13,20 @@ Here's an example from our `renovate/renovate` Dockerfile:
...
+
+If you enable `dockerReleaseTimestampSource=metadata`, Renovate can also use client-controlled OCI image metadata for non-Docker-Hub registries.
+Renovate checks the `org.opencontainers.image.created` manifest annotation first, then the image config `created` field.
+This can make `minimumReleaseAge` usable for registries such as GHCR, Quay, GCR, private registries, and pull-through caches that do not expose a supported registry timestamp.
```
polarathene, line 23: What is a client in this context? Probably better to phrase as:

> _Renovate can fallback to image metadata controlled by the publisher for image registries lacking an API integration for Renovate to securely source a publish timestamp._

I went with "image registries" above, assuming this is only intended for container images. If the datasource is intended to support OCI artifacts in addition to images, then OCI registries might be applicable (_but that'd likewise affect usage of "image" elsewhere in the document, so ignore me_).

lib/modules/datasource/docker/readme.md
```diff
@@ -13,4 +13,20 @@ Here's an example from our `renovate/renovate` Dockerfile:
...
+
+These metadata timestamps are weaker than registry-controlled timestamps because image publishers, build systems, or an attacker with publish access can set them to arbitrary values.
```
polarathene, line 25: "weaker than" => "insecure compared to"

lib/modules/datasource/docker/readme.md
```diff
@@ -13,4 +13,20 @@ Here's an example from our `renovate/renovate` Dockerfile:
...
+Some projects also rewrite image `created` timestamps for reproducible builds, so the value may be deterministic rather than a publish time.
+Keep the default `dockerReleaseTimestampSource=registry` if you use `minimumReleaseAge` as a strict supply-chain safety control.
+Only enable `metadata` when you accept this trade-off.
```
polarathene, line 28: These two tidbits are probably better positioned at the start of the section for the reader, who can then read the subsequent paragraphs about caveats of choosing to opt-in for the `metadata` workaround.

lib/modules/datasource/docker/readme.md
```diff
@@ -13,4 +13,20 @@ Here's an example from our `renovate/renovate` Dockerfile:
...
+
+To avoid excessive registry calls, Renovate only attempts metadata timestamp lookups for the newest Docker releases during a lookup.
```
polarathene, line 30: This could be better clarified. What is "Docker" in this context?

What is "newest" as opposed to whatever expectation this information is focused on addressing?
- My understanding without looking through code, is an image has already been determined and digest selected before the timestamp was inspected?
- If the timestamp is then too fresh, is image tag "newest - 1" then queried to identify the newest image published that satisfies the `minimumReleaseAge` constraint?

lib/modules/datasource/types.ts
```diff
@@ -1,6 +1,7 @@
 import type {
   ConstraintsFilter,
   CustomDatasourceConfig,
+  DockerReleaseTimestampSource,
```
polarathene, line 4: Like earlier feedback, I'd advise against introducing `Docker` like this for newly introduced types/fields.

lib/modules/datasource/docker/index.ts
```diff
@@ -86,7 +90,7 @@ export class DockerDatasource extends Datasource {
 
   override readonly releaseTimestampSupport = true;
   override readonly releaseTimestampNote =
-    'Only supported on Docker Hub: The release timestamp is determined from the `tag_last_pushed` field in the results. **NOTE**: Currently, digests will receive the same release timestamp as the `tag_last_pushed`, which means that digests may appear newer than they are - see https://github.com/renovatebot/renovate/issues/38659';
+    'Docker Hub release timestamps are determined from the `tag_last_pushed` field in the results. Other registries only receive release timestamps from OCI image metadata if `dockerReleaseTimestampSource=metadata` is enabled. **NOTE**: Currently, Docker Hub digests will receive the same release timestamp as the `tag_last_pushed`, which means that digests may appear newer than they are - see https://github.com/renovatebot/renovate/issues/38659';
```
polarathene, line 93: My understanding is if a registry provides the necessary API support, Renovate can add support that too.

For example [`zot` has support via their GraphQL API](https://github.com/project-zot/zui/pull/502#issue-3881457672) (_their UI uses this to present both the image creation from publisher and also a last tagged timestamp that is registry controlled when an image is published with a tag_).

polarathene (changes requested):
I've not gone over the full PR (no code review), you have several files where information I've provided feedback on is repeated but is inconsistent... where possible it should be in sync rather than multiple variations of saying the same thing.

thomas-be:
I was stumbling over the problem you described as a lot of spring-dependencies are not published with those timestamps so they were held back. I'm really interested in this feature. @Ryangr0 can you try to resolve the review comments?

esafak:
Can we get some movement on this? Quay, GHCR, ECR, GCR, Harbor — all get `getTags()` and no `releaseTimestamp`. Help needed @Ryangr0 ; should we take over?

PR #43573, opened 2026-05-24
feat(terraform-provider): use OpenTofu packages API for lockfile hashes

arnaud-dezandee:
## Changes

Use the new `packages` field that the OpenTofu Registry exposes on the per-platform provider download endpoint (introduced in [OpenTofu 1.12.0](https://github.com/opentofu/opentofu/pull/3434)) to populate `.terraform.lock.hcl` hashes without downloading and re-hashing each provider zip.

When the registry URL is `https://registry.opentofu.org`, `TerraformProviderHash.createHashes` now:

- Requests the `linux/amd64` download endpoint directly.
- Reads the `packages` map, which contains both `zh:` and `h1:` hashes for every platform.
- If `linux/amd64` is unavailable, queries `/versions` to select an available platform and requests that platform’s download endpoint.
- Falls back to the existing ZIP-download path when no usable `packages` metadata is available.

In the common case, this requires one HTTP request, with no ZIP downloads, extraction, or local hashing. Previously, generating hashes required registry metadata requests plus downloading and hashing every provider ZIP.

## Context

## AI assistance disclosure


lib/modules/manager/terraform/lockfile/hash.ts
```diff
@@ -12,6 +12,7 @@ import * as fs from '../../../../util/fs/index.ts';
 import { ensureCacheDir } from '../../../../util/fs/index.ts';
 import { Http } from '../../../../util/http/index.ts';
 import * as p from '../../../../util/promises.ts';
+import { TerraformDatasource } from '../../../datasource/terraform-module/base.ts';
```
RahulGautamSingh: `TerraformProviderDatasource` already gets `openTofuRegistryUrl` from `TerraformDatasource`. This import is unnecessary. 

Drop it and use `TerraformProviderDatasource.openTofuRegistryUrl` at the comparison below.

lib/modules/datasource/terraform-provider/index.spec.ts
```diff
@@ -483,6 +483,145 @@ describe('modules/datasource/terraform-provider/index', () => {
     });
   });
 
+  describe('getProviderPackages', () => {
+    const openTofuRegistryUrl = 'https://registry.opentofu.org';
```
RahulGautamSingh: Hardcoded URL string. Use `TerraformProviderDatasource.openTofuRegistryUrl` like `hash.spec.ts` does, so the test stays in sync if the constant changes.

lib/modules/datasource/terraform-provider/schema.ts
```diff
@@ -66,3 +66,44 @@ export const OpenTofuProviderDocsResponse = z
...
+  })
+  .transform(({ packages }): string[] | null => {
+    if (!packages) {
+      return null;
+    }
+    const hashes: string[] = [];
+    for (const pkg of Object.values(packages)) {
+      hashes.push(...pkg.hashes);
+    }
+    return hashes;
```
viceice:
```suggestion
    return Object.values(packages).flatMap({ hashes } => hashes);
```
simplify?

lib/modules/manager/terraform/lockfile/hash.ts
```diff
@@ -167,6 +167,24 @@ export class TerraformProviderHash {
     logger.debug(
       `Creating hashes for ${repository}@${version} (${registryURL})`,
     );
+
+    if (registryURL === TerraformProviderDatasource.openTofuRegistryUrl) {
```
viceice: also check if registry url is opentofu api

https://github.com/renovatebot/renovate/blob/dd1e0c88a95b997c7b13bb45e5f72bc07c5cc9df/lib/modules/datasource/terraform-module/index.ts#L69-L72

lib/modules/datasource/terraform-provider/index.ts
```diff
@@ -332,6 +333,55 @@ export class TerraformProviderDatasource extends TerraformDatasource {
...
+      repository,
+    );
+    const { body: versionsResponse } = await this.http.getJson(
+      `${baseUrl}/versions`,
+      TerraformRegistryVersions,
+    );
+    const versionEntry = versionsResponse.versions?.find(
+      (entry) => entry.version === version,
+    );
+    const platform = versionEntry?.platforms?.[0];
```
viceice: can we somehow reuse the platform info from previous API request? so we safe another request for all versions?

arnaud-dezandee: No previous request to reuse, sadly. The version list comes from `api.opentofu.org/.../index.json`, which only returns `{id, published}`, no platforms.
But we can still drop `/versions`: the download endpoint returns packages for all platforms no matter which os/arch is in the URL. So we can request linux/amd64 directly and fall back to /versions discovery on a 404 (rare). One request instead of two.

viceice: sounds good, please do

arnaud-dezandee:
This is ready if you got some time

lib/modules/datasource/terraform-provider/index.ts
```diff
@@ -332,6 +334,80 @@ export class TerraformProviderDatasource extends TerraformDatasource {
...
+      if (!(err instanceof HttpError) || err.response?.statusCode !== 404) {
+        throw err;
+      }
+      hashes = await this._getProviderPackagesForAvailablePlatform(
+        baseUrl,
+        version,
+      );
+    }
+
+    return hashes;
```
RahulGautamSingh: This keeps the type and flattens the flow.

```suggestion
   try {
  const { body } = await this.http.getJson(
    `${baseUrl}/${version}/download/linux/amd64`,
    OpenTofuProviderPackagesResponse,
  );
  return body;
} catch (err) {
  if (!(err instanceof HttpError) || err.response?.statusCode !== 404) {
    throw err;
  }
  return await this._getProviderPackagesForAvailablePlatform(baseUrl, version);
}
```

lib/modules/datasource/terraform-provider/index.ts
```diff
@@ -332,6 +334,80 @@ export class TerraformProviderDatasource extends TerraformDatasource {
     );
   }
 
+  /**
+   * The OpenTofu download endpoint exposes a `packages` map with the `zh:`/`h1:`
+   * hashes for every platform, avoiding a local download-and-rehash of each zip.
+   * The map is platform-independent, so we request `linux/amd64` directly and
```
RahulGautamSingh: reword so reader does not assume the hashes are the same per platform.

RahulGautamSingh (changes requested):

lib/modules/datasource/terraform-provider/index.ts
```diff
@@ -332,6 +334,77 @@ export class TerraformProviderDatasource extends TerraformDatasource {
...
+    );
+
+    try {
+      const { body } = await this.http.getJson(
+        `${baseUrl}/${version}/download/linux/amd64`,
+        OpenTofuProviderPackagesResponse,
+      );
+      return body;
+    } catch (err) {
+      if (!(err instanceof HttpError) || err.response?.statusCode !== 404) {
```
RahulGautamSingh: `_getProviderPackages` throws on any non-404 error (e.g. a transient 5xx on the fast-path download), so it will not fall back to the slow zip path in that case.

is this intentional?

arnaud-dezandee: Yes, intentional. A 404 is the expected soft case (that platform/version just isn't on the packages endpoint) so we fall back to platform discovery. A non-404 like a 5xx or network error is a real host error, and I'd rather surface it and retry next run than silently kick off a full zip download that could hide a registry problem. The zip path hits the same registry anyway, so it wouldn't buy much on an actual outage.

RahulGautamSingh: this rethrows a raw `HttpError` on any non-404, while the rest of the module wraps external failures in `ExternalHostError`. use the same for being consistent.

RahulGautamSingh (commented):
The description says step 1 calls `/versions` to pick a platform, then one download, "two HTTP calls". After the perf commit the code tries `linux/amd64` directly first and only hits `/versions` on a 404, so the happy path is one call. Please update the body to match.

lib/modules/datasource/terraform-provider/index.ts
```diff
@@ -332,6 +334,87 @@ export class TerraformProviderDatasource extends TerraformDatasource {
...
+    repository: string,
+    version: string,
+  ): Promise<string[] | null> {
+    const baseUrl = joinUrlParts(
+      TerraformProviderDatasource.openTofuRegistryUrl,
+      'v1/providers',
+      repository,
+    );
+
+    return this.http
```
RahulGautamSingh: use async/await

RahulGautamSingh (changes requested):

lib/modules/datasource/terraform-provider/index.ts
arnaud-dezandee: This leads to nested try/catches, that's why I used Promises chain first

RahulGautamSingh (approved):

arnaud-dezandee:
Ready when you have time :pray:

lib/modules/datasource/terraform-provider/schema.ts
```diff
@@ -142,3 +142,24 @@ export const TerraformRegistryBuildResponse = z.object({
...
+const OpenTofuProviderPackage = z.object({
+  hashes: LooseArray(z.string()).catch([]),
+});
+
+export const OpenTofuProviderPackagesResponse = z
+  .object({
+    packages: z.record(z.string(), OpenTofuProviderPackage).optional(),
+  })
+  .transform(({ packages }): string[] | null => {
+    const allHashes = Object.values(packages ?? {}).flatMap(
```
viceice: use `coerceObject`

viceice (commented):
otherwise LGTM

lib/modules/datasource/terraform-provider/schema.ts
```diff
@@ -142,3 +142,24 @@ export const TerraformRegistryBuildResponse = z.object({
...
+
+// OpenTofuProviderPackagesResponse — per-platform download endpoint exposes a
+// `packages` map containing the `zh:`/`h1:` hashes for every platform.
+const OpenTofuProviderPackage = z.object({
+  hashes: LooseArray(z.string()).catch([]),
+});
+
+export const OpenTofuProviderPackagesResponse = z
+  .object({
+    packages: z.record(z.string(), OpenTofuProviderPackage).optional(),
```
zharinov:
```suggestion
    packages: LooseRecord(OpenTofuProviderPackage).catch({}),
```

lib/modules/datasource/terraform-provider/schema.ts
zharinov: After using `LooseRecord`:

```suggestion
    const allHashes = Object.values(packages).flatMap(
```

Now, no `coerceObject` needed.

zharinov (changes requested):

lib/modules/datasource/terraform-provider/schema.ts
zharinov: This allows to collect items that are parsed correctly and ignore any values that doesn't match `OpenTofuProviderPackage` schema, instead of dropping all on one value error.

lib/modules/datasource/terraform-provider/schema.ts
arnaud-dezandee: fixed with zharinov suggestions

viceice (approved):

PR #43605, opened 2026-05-26
fix(util/fingerprint): incrementally hash input to avoid V8 max string length

mado-m:
## Changes

`fingerprint()` previously called `safeStringify(input)` to build a single canonical JSON string and then hashed it. On large monorepos this materialization can exceed V8's max string length and throw `RangeError: Invalid string length` from `ensurePr`, blocking PR creation entirely.

Reproduction (verified on a real monorepo with ~200 `go.mod` files and a single GHSA matching all of them): one vulnerability PR carries ~200 upgrades, each upgrade's `prBodyNotes` populated with the full advisory body by `vulnerabilities.ts`. The resulting `filteredPrConfig` is several hundred MB once stringified — `safe-stable-stringify`'s internal string concatenation hits the V8 limit and throws.

This PR replaces the materialize-then-hash approach with a recursive walker that emits canonical-JSON chunks directly into `crypto.createHash('sha512')`, never holding the full string in memory.

The canonicalization mirrors `safe-stable-stringify` in deterministic mode (sorted keys, `[Circular]` for cycles, `undefined`/`function`/`symbol` dropped in objects and replaced with `null` in arrays, `toJSON()` respected). Output is **byte-identical** to the previous `safeStringify` + `hash` combination for all inputs that fit within V8 limits, so previously stored fingerprints remain valid (no PR/branch re-render churn on upgrade).

Verified locally:
- All 18 representative cases (primitives, nested objects, arrays, `Date`, undefined/function/symbol drops, circular refs) produce **identical hashes** under both legacy and new impl.
- Legacy `safeStringify` throws `RangeError: Invalid string length` on a ~600 MB input that completes in ~560 ms under the new impl.

## Context

Refs Discussion: https://github.com/renovatebot/renovate/discussions/38189

## AI assistance disclosure

Tool: Claude Code (Opus 4.7). The implementation and tests were drafted with AI assistance; backward-compatibility was independently verified by comparing the new walker's output against `safeStringify(input)` + `hash(...)` for the cases listed above before commit.

The public repository: n/a (private monorepo where the original failure was reproduced; the new `fingerprint.spec.ts` covers the relevant invariants — equivalence with the legacy `safeStringify`+`hash` for representative cases, including circular refs and the various `undefined`/`function`/`symbol` drops).

### DIFF
--- a/lib/util/fingerprint.ts
+++ b/lib/util/fingerprint.ts
@@ -29,7 +29,7 @@ function isEmittable(value: unknown): boolean {
 // across hundreds of manifest files).
 function fingerprintInto(h: Hash, value: unknown, seen: WeakSet<object>): void {
   if (value === null || typeof value !== 'object') {
-    h.update(JSON.stringify(value)!);
+    h.update(JSON.stringify(value));
     return;
   }
 
--- a/lib/util/fingerprint.ts
+++ b/lib/util/fingerprint.ts
@@ -1,14 +1,35 @@
-import { createHash, type Hash } from 'node:crypto';
+import { type Hash, createHash } from 'node:crypto';
 
-// Walks `value` and emits deterministic JSON chunks directly into `h`,
+// Returns true when `value` would appear in the canonical JSON produced by
+// `safe-stable-stringify`. We pre-check before emitting because objects must
+// drop keys whose value resolves to nothing (incl. via toJSON → undefined),
+// and incremental hashing can't roll back already-emitted bytes.
+function isEmittable(value: unknown): boolean {
+  if (
+    value === undefined ||
+    typeof value === 'function' ||
+    typeof value === 'symbol'
+  ) {
+    return false;
+  }
+  if (value === null || typeof value !== 'object') {
+    return true;
+  }
+  const obj = value as { toJSON?: () => unknown };
+  if (typeof obj.toJSON === 'function') {
+    return isEmittable(obj.toJSON());
+  }
+  return true;
+}
+
+// Walks `value` and emits canonical JSON chunks directly into `h`,
 // matching `safe-stable-stringify` (deterministic mode) byte-for-byte
 // without materializing the full string. Required to avoid V8's max
 // string length when the input is large (e.g. a PR carrying upgrades
 // across hundreds of manifest files).
 function fingerprintInto(h: Hash, value: unknown, seen: WeakSet<object>): void {
   if (value === null || typeof value !== 'object') {
-    const s = JSON.stringify(value);
-    h.update(s ?? 'null');
+    h.update(JSON.stringify(value)!);
     return;
   }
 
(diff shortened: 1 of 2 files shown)


lib/util/fingerprint.ts
```diff
@@ -1,7 +1,92 @@
...
+    }
+    h.update(']');
+  } else {
+    const entries = value as Record<string, unknown>;
+    const keys = Object.keys(entries).sort();
+    h.update('{');
+    let first = true;
+    for (const k of keys) {
+      const v = entries[k];
+      if (!isEmittable(v)) {
```
viceice: we should pass a map to cache emitable check (perf)

viceice (commented):
otherwise LGTM

jamietanna:
This is similar to what happened with https://github.com/renovatebot/renovate/pull/43487 - is there maybe something wider we should look into?

lib/util/fingerprint.ts
mado-m: Done in 65844ed — threaded a `WeakMap<object, boolean>` through `isEmittable` (and into `fingerprintInto`) so each object reference is checked at most once, avoiding a second walk through any `toJSON` chain.

mado-m:
@jamietanna good catch — same root cause (eagerly materializing a canonical-JSON serialization of a large structure into a single V8 string).

A quick `grep` of `lib/` shows the pattern in a few places, split by whether the caller actually needs the string or just a hash of it:

| Site | Consumes the string as | Notes |
|---|---|---|
| `lib/util/fingerprint.ts` | hash only | this PR |
| `lib/util/cache/repository/impl/base.ts` (`isModified`) | hash only | same shape — could reuse `fingerprint()` once this lands |
| `lib/util/cache/repository/impl/base.ts` (`save`) | string → compress → base64 → disk write | needs a streaming JSON encoder, not just incremental hash |
| `lib/logger/with-sanitizer.ts` (`quickStringify(raw)`) | string → stream.write | #43487 sidestepped this by moving one huge log to TRACE, but any other log path that passes a big structure would hit the same wall |
| `lib/workers/repository/updates/generate.ts:530` | dedup key over a 4-element `string[]` | not at risk |
| `lib/util/clone.ts` (`quickStringify`) | clone via parse(stringify(...)) | input-size-bounded, separate concern |

So the "hash-only" sites can adopt this PR's incremental-hash trick; the "actually need the string" sites need a streaming JSON encoder that writes to a sink (compressor / log stream).

I'd suggest landing this PR as the focused fix for the `fingerprint()` site that's actively breaking users, and tracking the broader audit separately. Happy to file an umbrella issue and/or follow up with a `cache/repository/impl/base.ts:isModified` PR that reuses `fingerprint()` once this lands — whichever the maintainers prefer.

viceice (approved):

RahulGautamSingh (approved):

mado-m:
Just checking in on this. It looks approved and CI is green; is there anything else needed from my side before this can be merged?

jamietanna (approved):

PR #43607, opened 2026-05-26
docs: migrate all admonitions and lists to using 2-space indents

jamietanna:
## Changes

Correctly renders generated admonitions with 1 indent (of 2 spaces):

Screenshot 2026-05-26 at 17 52 11

As well as updating i.e.:

http://127.0.0.1:8000/modules/manager/gomod/#post-update-options

Screenshot 2026-05-26 at 18 27 35
Screenshot 2026-05-26 at 18 27 47
Screenshot 2026-05-26 at 18 27 29

## Context

Please select one of the below:

- [ ] This closes an existing Issue: Closes #
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [x] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #43615, opened 2026-05-26
ci(prettier): add prettier plugin for admonitions

secustor:
## Changes

Adds a local prettier plugin, which temporarily replaces `!!!` or `???` with `<!--mkdocs-admonition: -->` which fixes prettier removing the leading whitespace of the content

```
// before
!!! note
lore ipsum

// after 
!!! note
  lore ipsum
```

Also remove usages of `<!-- prettier-ignore -->` calls 

## Context

https://github.com/renovatebot/renovate/pull/43607#pullrequestreview-4366397020

Not sure what happens so far with dynamically generated admonitions e.g. 

https://github.com/renovatebot/renovate/blob/33e857e04a41e5a46c4eb7c1bbdb4f8d5900bdfb/tools/docs/config.ts#L218-L219

Looks like prettier is run after the generation too. 
Is it in the context of this repo now, that the docs page has been moved here? 

If yes, it should be fine to remove the ignores there too. 

## AI assistance disclosure

The public repository: <URL>


tools/prettier/plugins/mkdocs-admonitions/index.mjs
```diff
@@ -0,0 +1,190 @@
...
+  let fenceChar = '';
+  let fenceLen = 0;
+
+  let i = 0;
+  while (i < lines.length) {
+    const line = lines[i];
+
+    if (inFence) {
+      // Closing fence: same char, at least as many chars, no content after
+      const closeMatch = /^(\s{0,3})(`{3,}|~{3,})\s*$/.exec(line);
```
RahulGautamSingh:
```suggestion
      const closeMatch = /^(?:\s{0,3})(`{3,}|~{3,})\s*$/.exec(line);
```
This group is not being used so no point capturing it as one.

Then we can match index 1 below instead of 2.

Same for fenceMatch

RahulGautamSingh (commented):
minor nitpick

tools/prettier/plugins/mkdocs-admonitions/index.mjs
secustor: Switched to named groups instead

RahulGautamSingh (approved):

tools/prettier/plugins/mkdocs-admonitions/index.mjs
```diff
@@ -0,0 +1,188 @@
+// Custom prettier plugin to preserve MkDocs/Material admonitions (!!! / ??? / ???+).
+// Prettier 3.x does not understand this non-standard CommonMark extension
+// and corrupts the indented body by reflowing it into preceding paragraphs.
+//
+// Strategy: before the built-in markdown parser runs, replace each admonition
+// block with an HTML comment placeholder; after parsing, restore the original
+// text in the resulting html AST nodes. Prettier prints html nodes verbatim.
+//
+// Known limitations:
+//   - Nested admonitions are preserved verbatim (outer block is masked as one unit).
```
jamietanna, line 10: Do we have any of these to watch out for?

tools/prettier/plugins/mkdocs-admonitions/index.spec.mjs
```diff
@@ -0,0 +1,164 @@
...
+
+  it('preserves 3-space-indented body (mirrors configuration-options.md)', async () => {
+    const input = block`
+      !!! note
+         Body with three spaces.
+    `;
+    await expect(fmt(input)).resolves.toBe(input);
+  });
+
+  it('preserves 5-space-indented body (mirrors self-hosted-configuration.md)', async () => {
```
jamietanna, line 48: I'm not sure if we should have 3-/5-spaces in files? Is that right?

jamietanna (commented):
Couple of small questions, otherwise LGTM - good job!

tools/prettier/plugins/mkdocs-admonitions/index.mjs
secustor: I scanned before the codebase for occurrences of `!!!` and couldn't find any

tools/prettier/plugins/mkdocs-admonitions/index.spec.mjs
secustor: Right now the idea has been not to do any active write changes, but I can change it to always set it to 2 spaces in a follow up PR. 

I would like to avoid removing the ignore and mix them with other changes. 

The exception is   
https://github.com/renovatebot/renovate/pull/43615/changes#diff-be0a39f342380493f17c380bbd694f1f05d18bdd8d7cecf174369084952102deR4753 
which is clearly not correct

docs/usage/configuration-options.md
```diff
@@ -4831,12 +4747,10 @@ This option was formerly known as `updateLockFiles`.
 
 When this option is set, Renovate won't attempt to update artifacts such as lock files, so you will need to update them yourself, either manually or through secondary automation such as CI workflows.
 
-<!-- prettier-ignore -->
 !!! note
   When this option is used in package rules, along with grouped upgrades, artifact updating will only be skipped if every upgrade in the grouped branch wants to skip it.
 
-<!-- prettier-ignore -->
-!!!warning
+!!! warning
```
jamietanna, line 4753: The line below this should be unindented too, no? I'd expect Prettier to catch this line as incorrectly overly indented

jamietanna: Want me to get this one fixed before this goes in? Given it's not rendering correctly: https://docs.renovatebot.com/configuration-options/#skipartifactsupdate

secustor: I think it makes sense to do this as part of the follow up PR. If you don't want to wait, I can do it now.

jamietanna (approved):

docs/usage/configuration-options.md
jamietanna: All good - you happy to raise the follow-up PR (I assume you may have some other candidates in mind too)

viceice (approved):

PR #43629, opened 2026-05-27
feat: apply group settings when only one update is present

felipecrs:
## Changes

By default, Renovate titles PRs after the individual dependency when a group contains only a single update.

This PR adds a new configuration option, `groupSingleUpdates`, that allows users to choose whether Renovate should always title PRs after the group, even when the group contains only a single update.

Discussed and proposed in https://github.com/renovatebot/renovate/discussions/25588#discussioncomment-7486378. Note I chose a different option name than the one proposed in the discussion, as I think this name is more descriptive of the behavior it controls.

## Context

## AI assistance disclosure

This PR has been written with assistance of Claude Opus 4.6.

The public repository: <URL>


lib/workers/repository/updates/generate.ts
```diff
@@ -275,7 +275,9 @@ export function generateBranchConfig(
   const typesGroup =
     depNames.length > 1 && !hasGroupName && isTypesGroup(branchUpgrades);
   logger.trace(`groupEligible: ${groupEligible}`);
-  const useGroupSettings = hasGroupName && groupEligible;
+  const useGroupSettings =
+    hasGroupName &&
+    (groupEligible || branchUpgrades[0].groupSingleUpdates === true);
```
jamietanna, line 280: Could this still trigger when there are multiple upgrades? Should we check the length of `branchUpgrades`?

felipecrs: Yes, it can trigger with multiple `branchUpgrades`, but only when `groupEligible === false`, which means the same dep updating to the same version in multiple files.

I think applying group settings in that case is fine (considering user has explicitly set `groupName` and `groupSingleUpdates`).

Don't you think?

felipecrs: I added test coverage to cover that case. https://github.com/renovatebot/renovate/pull/43629/commits/e5fcc96e37b0fbf166143bd93ec6a0398df08d04

RahulGautamSingh (changes requested):
needs docs

felipecrs:
@RahulGautamSingh, I believe the auto-generated docs are good enough. I even had an extra sentence but thought it was superfluous.

https://github.com/renovatebot/renovate/pull/43629/commits/199d23320e5ef2cc42b626a67c5cd21dbc2f44d1

image

What do you think?

EDIT: Some examples:

- https://docs.renovatebot.com/configuration-options/#commitbodytable
- https://docs.renovatebot.com/configuration-options/#prheader
- https://docs.renovatebot.com/configuration-options/#prfooter

jamietanna (commented):
Thinking about this some more, I'm wondering if this should maybe be the default - what do @renovatebot/maintainers think?

IMO, it's a little confusing when a group only contains a single update and has a different title.

For instance, I noticed this earlier today on one of my projects (which I'd exepct there to be multiple dependencies in the group for)

Are there any other settings that may/may not be used when there's a group that may not be wanted if only one update is there?

felipecrs:
I do understand the rationale behind it: if there's only one update, titling the PR after update makes it a lot more descriptive than titling it after the group.

But I tend to agree, it is a common source of confusion for my users too.

> Are there any other settings that may/may not be used when there's a group that may not be wanted if only one update is there?

I can't think of anything, but have in mind anything under `group` is applicable in this case. It can hold arbitrary config.

https://docs.renovatebot.com/configuration-options/#group

I.e. I think it should be treated as a breaking change if we go that route.

reduckted:
Looking at the example posted here:
https://github.com/renovatebot/renovate/discussions/25588#discussioncomment-17078124

That sets `groupSingleUpdates` in a package rule. Can it also be applied at the top-level so that it applies to all groups (and could therefore be added to a common preset so that I don't have to edit the Renovate configs in dozens of repositories)?

felipecrs:
> Looking at the example posted here: [#25588 (comment)](https://github.com/renovatebot/renovate/discussions/25588#discussioncomment-17078124)
> 
> That sets `groupSingleUpdates` in a package rule. Can it also be applied at the top-level so that it applies to all groups (and could therefore be added to a common preset so that I don't have to edit the Renovate configs in dozens of repositories)?

Absolutely. It can be set at any level.

secustor:
> Thinking about this some more, I'm wondering if this should maybe be the default - what do @renovatebot/maintainers think?

SGTM, I ran into this too some times

jamietanna (changes requested):
Per comments (and +1'd by `secustor`) we'll make this on-by-default, and then folks who wish the old behaviour can disable it where they'd like to

felipecrs:
Alright. Working on it.

Question: should I add a breaking change notice to it, or would you rather handle that yourself during merge?

jamietanna:
I'm happy arguing this isn't a full "breaking change", so no need to document anything additionally

I'll see about whether it's worth adding a bit of a call out in the GitHub Release itself

docs/usage/configuration-options.md
```diff
@@ -2002,6 +2002,8 @@ For example, to group all non-major devDependencies updates together into a sing
   <br>
   Lock file maintenance will never be grouped with other dependency updates.
 
+## `groupSingleUpdates`
```
jamietanna, line 2005: I think adding some wording - especially now this has flipped to on-by-default - around why you may want to _not_ have this may be useful (if we can think of it)

felipecrs: Tried in https://github.com/renovatebot/renovate/pull/43629/commits/995c30ceb8e8d61c3f887834f35bb2c37a406acf. Hope you like it. :)

jamietanna (approved):

faenschi:
Hi! We've noticed that since this PR merged, our Maven dependency MR titles changed from:

` deps(deps): update dependency ch.foobar.ri:commonphases to v33`

to:

` deps(deps): update version.foobar.commonphases to v33`

This change appears to have been released in a patch/minor version update, but it was breaking for us—a lot of our PRs got recreated due to the title format change.

We understand this is due to the new groupSingleUpdates behavior—when a group contains only a single update, Renovate titles the PR using the group property name instead of the individual dependency name.

Questions:

 1. Was this a breaking change intended for a major version bump? It seems like this should have been opt-in rather than the default.
 2. Should groupSingleUpdates default to false for Maven/Java projects to maintain backward compatibility?
 3. Are there docs explaining the migration path for users affected by this change?

We've worked around this by setting "groupSingleUpdates": false in our Renovate presets, but other teams may hit this issue.

Thanks!

jamietanna:
Moving discussion into https://github.com/renovatebot/renovate/discussions/43756

PR #43635, opened 2026-05-27
feat(containerbase): add packageRules support for containerbase dynamic install

mistabanasik:
## Changes

Renovate doesn't use packageRules when installing tools with containerbase, so for example changing repository url is not possible with binarySource=install. 

This PR adds new option to instance global config, `containerbasePackageRules` which reuses `packageRules` for containerbase install-tool package resolutions.

Link to discussion:
https://github.com/renovatebot/renovate/discussions/43636

## Context

## AI assistance disclosure

Yes, I've used AI tools to generate test cases.

Code was tested on corporate self-hosted repository.


jamietanna:
I think we have some planned work around this `@viceice`, let's take a look and see how this proposal fits into it

mistabanasik:
Well now when I think of it, it might be better idea to reuse `packageRules` defined in global config.js. However if we add `packageRules` to `GlobalConfig.OPTIONS` they are removed from config passed to repository-level configuration in `GlobalConfig.set`. So we would have to add some kind of method like `GlobalConfig.update` so we can still pass `packageRules` down to repository-level config.

mistabanasik:
I've added another commit - we should use full config returned from `applyPackageRules` instead of picking only `packageName`, `datasource`, `versioning` and `extractVersion` fields. Thanks to that we can for example override registryUrl for node install-tool command, not only datasource.

mistabanasik:
@jamietanna @viceice any thoughts on this PR?

RahulGautamSingh (changes requested):
while we wait for maintainers review, could you please clean the pr title to not include the `...`

jamietanna:
Moving to draft until we get a chance to review the approach and work out if it matches where we want to go 👍🏼

felipecrs:
I didn't realize this PR (nor https://github.com/renovatebot/renovate/discussions/43636 or https://github.com/renovatebot/renovate/discussions/44608) and I proposed another PR that addresses the same issue:

- https://github.com/renovatebot/renovate/pull/44641

Took a bit different approach, though.

mistabanasik:
I've changed my PR to rely purely on packageRules config without additional containerbasePackageRules. With new depType=tool-constraint we can also specifically address some packageRules to only target containerbase dynamic installs via matchDepTypes.

@jamietanna I saw your comment in the other thread - well this PR + containerbase URL replace env variables are sufficient to make containerbase dynamic install work with self-hosted package mirrors. I think in the future there should be a way without needing to set URL replace env variables, but for now in my opinion this PR is sufficient.

felipecrs:
That's cool!

I closed [my PR](https://github.com/renovatebot/renovate/pull/44641) so there's one less to review.

docs/usage/configuration-options.md
```diff
@@ -3022,6 +3022,23 @@ If Renovate finds a dependency update available, and that dependency update is f
...
+    {
+      "matchDepNames": ["node"],
+      "matchDepTypes": ["tool-constraint"],
+      "overrideDatasource": "node-version",
+      "registryUrls": ["https://artifactory.example.com/nodejs"]
+    }
+  ]
+}
+```
+
```
felipecrs, line 3041: I believe repository-level `packageRules` will not be taken into account, and if so, I believe you should address this via documentation.

(and that's why I preferred to document in self-hosted configuration instead)

PR #43637, opened 2026-05-27
feat(gerrit): Add Gerrit serverInfo endpoint support

mistabanasik:
## Changes

When using Gerrit as a platform, Renovate doesn't use server info config (endpoint: a/config/server/info). That's why when using SSH authorization, URL is incorrectly set as:
`ssh://gerrit.example.com:29418/web/apps`
instead of correct:
`ssh://admin@gerrit.example.com:29418/web/apps`
returned from server info and visible in Gerrit web UI.

My PR fixes that and prioritizes server info endpoint when available.

Link to discussion:
https://github.com/renovatebot/renovate/discussions/43638

## Context

## AI assistance disclosure

I've used AI tools to generate test cases.

Code was tested on corporate self-hosted repository.


lib/modules/platform/gerrit/index.ts
```diff
@@ -135,6 +135,7 @@ export async function initRepo({
   gitUrl,
 }: RepoParams): Promise<RepoResult> {
   logger.debug(`initRepo(${repository}, ${gitUrl})`);
```
RahulGautamSingh, line 137: `getServerInfo()` has no error handling. If the endpoint is unavailable, older Gerrit version, permission denied, or a network error then `initRepo` will throw and the entire platform becomes unusable for that repo. 

Needs a try/catch that falls back gracefully, e.g.:
```ts
const serverInfo = await client.getServerInfo().catch(() => ({
  download: { archives: [], schemes: {} },
}));
```

lib/modules/platform/gerrit/utils.ts
```diff
@@ -46,18 +48,38 @@ export function getGerritRepoUrl(
 
   const url =
     gitUrl === 'ssh'
-      ? createSshUrl(endpointUrl, repository)
```
RahulGautamSingh, line 49: `createSshUrl` uses the server info URL directly after string substitution with no validation. 

Follow what `createHttpUrl` does ie. correctly wraps its result in `parseUrl(...)` and falls back on failure

RahulGautamSingh (commented):
A few issues across the changes — one critical, one moderate, and a couple of nits.

felipecrs:
We don't need to pull server info for that. Look like we just missed the username when constructing the SSH URL.

mistabanasik:
I've tried to sync Gerrit platform support with other platforms - GitLab, Gitea and Bitbucket rely on SSH url with username included from the repository server response.

As far as I can tell, no platform supports setting username for SSH directly.

felipecrs:
I'm not sure I get your point. I meant if this patch isn't enough:

```diff
diff --git a/lib/modules/platform/gerrit/index.ts b/lib/modules/platform/gerrit/index.ts
index 0f7972d90..220ec0ab3 100644
--- a/lib/modules/platform/gerrit/index.ts
+++ b/lib/modules/platform/gerrit/index.ts
@@ -146,7 +146,12 @@ export async function initRepo({
     labels: projectInfo.labels ?? {},
   };
   const baseUrl = defaults.endpoint!;
-  const url = getGerritRepoUrl(repository, baseUrl, gitUrl);
+  const url = getGerritRepoUrl(
+    repository,
+    baseUrl,
+    gitUrl,
+    config.gerritUsername!,
+  );
   configureScm(repository, config.gerritUsername!);
   await git.initRepo({ url, cloneSubmodules, cloneSubmodulesFilter });
 
diff --git a/lib/modules/platform/gerrit/utils.ts b/lib/modules/platform/gerrit/utils.ts
index 3c627aadc..b0f55478d 100644
--- a/lib/modules/platform/gerrit/utils.ts
+++ b/lib/modules/platform/gerrit/utils.ts
@@ -38,6 +38,7 @@ export function getGerritRepoUrl(
   repository: string,
   endpoint: string,
   gitUrl: GitUrlOption | undefined,
+  username: string,
 ): string {
   const endpointUrl = parseUrl(endpoint);
   if (!endpointUrl) {
@@ -46,15 +47,15 @@ export function getGerritRepoUrl(
 
   const url =
     gitUrl === 'ssh'
-      ? createSshUrl(endpointUrl, repository)
+      ? createSshUrl(endpointUrl, repository, username)
       : createHttpUrl(endpointUrl, endpoint, repository);
   logger.trace({ url }, 'using URL based on configured endpoint');
 
   return url;
 }
 
-function createSshUrl(url: URL, repository: string): string {
-  return `ssh://${url.host}:${DEFAULT_SSH_PORT}/${repository}`;
+function createSshUrl(url: URL, repository: string, username: string): string {
+  return `ssh://${username}@${url.host}:${DEFAULT_SSH_PORT}/${repository}`;
 }
 
 function createHttpUrl(url: URL, endpoint: string, repository: string): string {
```

mistabanasik:
Okay, but using this method we should also have option to change DEFAULT_SSH_PORT. When we use server info endpoint, we can get correct url in one request.

Also according to docs:
https://gerrit-review.googlesource.com/Documentation/user-upload.html

> git clone ssh://sshusername@hostname:29418/REPOSITORY_NAME.git
>
> In the command above, sshusername was configured as Username on the Profile tab of the Settings screen. If it is not set, propose a name and use Select Username to select the name.

I think this means there might be a case when sshusername is different than login - but I might be wrong.

felipecrs:
I pushed a PR for it:

- https://github.com/renovatebot/renovate/pull/43651

I was able to verify it myself against review.gerrithub.io.

felipecrs:
> Okay, but using this method we should also have option to change DEFAULT_SSH_PORT. When we use server info endpoint, we can get correct url in one request.
> 
> Also according to docs: [gerrit-review.googlesource.com/Documentation/user-upload.html](https://gerrit-review.googlesource.com/Documentation/user-upload.html)
> 
> > git clone ssh://sshusername@hostname:29418/REPOSITORY_NAME.git
> > In the command above, sshusername was configured as Username on the Profile tab of the Settings screen. If it is not set, propose a name and use Select Username to select the name.
> 
> I think this means there might be a case when sshusername is different than login - but I might be wrong.

If we ever encounter a situation where that is required, then we can consider it. For now, we can keep the implementation optimized by saving an HTTP request. That's my suggestion.

lib/modules/platform/gerrit/client.ts
```diff
@@ -43,6 +44,13 @@ class GerritClient {
     return res.body;
   }
 
+  async getServerInfo(): Promise<GerritServerInfo> {
+    const res = await this.gerritHttp.getJsonUnchecked<GerritServerInfo>(
```
viceice: this should use zod schema and checked Jason response

felipecrs: Just FYI none of the responses in the Gerrit platform have zod schemes yet.

(Maybe we should defer that to a later refactor PR?)

felipecrs:
To be clear, I don't recommend merging this PR.

It's legitimate, undoubtedly, but I don't think it's reasoning is strong enough to justify changing the approach discussed [here](https://github.com/renovatebot/renovate/pull/42672#issuecomment-4264380481).

I would recommend merging https://github.com/renovatebot/renovate/pull/43651 instead, which fixes this issue while keeping the existing approach.

lib/modules/platform/gerrit/client.ts
viceice: no, we nudge every new code to use schema validation. doing a huge refactor later makes review a lot harder and time consuming.

mistabanasik:
There are issues with current hardcoded URL, as mentioned in:
https://github.com/renovatebot/renovate/pull/42672#issuecomment-4259476169
Particularly case when Gerrit is configured like that:
> git clone "https://gerrit.ext.net.nokia.com/gerrit/<project>" - anonymous HTTPS,
> git clone "https://<user>@gerrit.ext.net.nokia.com/gerrit/a/<project>" - HTTPS,
> git clone "ssh://<user>@gerrit.ext.net.nokia.com:29418/<project>" - SSH.

My PR already addresses that - relying on a/config/server/info instead of hardcoded url. It also addresses problems with non-default port used by Gerrit instance and username. I don't think hardcoding URL creation is right way, when other platforms (GitLab, Bitbucket, Gitea) rely on server-returned information about Git URL. One additional call to Gerrit in my opinion isn't dealbreaker, because it's also cached.

So in 1 PR we address 3 different problems with Gerrit platform for a price of an additional Gerrit request. In my opinion it's totally worth it.

felipecrs:
> There are issues with current hardcoded URL, as mentioned in: [#42672 (comment)](https://github.com/renovatebot/renovate/pull/42672#issuecomment-4259476169) Particularly case when Gerrit is configured like that:
> 
> > git clone "[gerrit.ext.net.nokia.com/gerrit](https://gerrit.ext.net.nokia.com/gerrit/)" - anonymous HTTPS,
> > git clone "https://@gerrit.ext.net.nokia.com/gerrit/a/" - HTTPS,
> > git clone "ssh://@gerrit.ext.net.nokia.com:29418/" - SSH.
> 
> My PR already addresses that - relying on a/config/server/info instead of hardcoded url. It also addresses problems with non-default port used by Gerrit instance and username. I don't think hardcoding URL creation is right way, when other platforms (GitLab, Bitbucket, Gitea) rely on server-returned information about Git URL. One additional call to Gerrit in my opinion isn't dealbreaker, because it's also cached.

Renovate has no such issue. It was addressed [here](https://github.com/renovatebot/renovate/pull/42672#issuecomment-4264847539).

mistabanasik:
@RahulGautamSingh @viceice any thoughts on this PR?

RahulGautamSingh (changes requested):
~@mistabanasik  gerrit platform support is added & maintained @felipecrs so we don't want to merge this without his approval~

Ignore the above please.

@mistabanasik wait for maintainers comment please, they will decide on a move forward.

felipecrs:
PS: There's nothing wrong with this patch: there's just no need for it now. We can reconsider it when someone finds a situation of which the current logic is not capable of handling.

PR #43672, opened 2026-05-29
fix(rust-version): cache bust `manifests.txt` to avoid stale CDN responses

Turbo87:
## Changes

The CDN in front of `static.rust-lang.org` can serve a stale `manifests.txt` for up to a day, which hides newly released versions from the `rust-version` datasource. This was investigated in dependabot/dependabot-core#13583, where it was traced to inconsistent CDN edge propagation of `manifests.txt`. It affects stable releases too, not just nightly: at the moment 1.96.0 is missing from https://static.rust-lang.org/manifests.txt, but shows up in e.g. https://static.rust-lang.org/manifests.txt?t=123.

This appends a timestamp query param (`?t=<unix-ms>`) to the `manifests.txt` fetch so each request falls through to origin. Renovate's package cache (`withCache`) still controls how often we fetch, since `_getReleases` is only invoked on a cache miss, so the bust fires at most once per cache window and never adds load beyond Renovate's own TTL.

This mirrors the fix in dependabot-core (dependabot/dependabot-core#13604).

## Context

## AI assistance disclosure

The code and tests were written by Claude Code (Claude Opus 4.8) under my review and direction.

### DIFF
--- a/lib/modules/datasource/rust-version/index.spec.ts
+++ b/lib/modules/datasource/rust-version/index.spec.ts
@@ -1,5 +1,7 @@
+import { isString } from '@sindresorhus/is';
 import * as httpMock from '~test/http-mock.ts';
 import { logger } from '~test/util.ts';
+import { regEx } from '../../../util/regex.ts';
 import { getPkgReleases } from '../index.ts';
 import { RustVersionDatasource } from './index.ts';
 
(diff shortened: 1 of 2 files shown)


jamietanna:
Hmm, thoughts @viceice @zharinov ?

Turbo87:
let's actually skip this for now. it looks like my fix in https://github.com/rust-lang/simpleinfra/pull/1062 solved the root cause, so the cache busting shouldn't be necessary anymore. if it turns out that this still doesn't work we can always reopen the PR. sorry for the noise 😅

PR #43695, opened 2026-05-30
feat(manager/helmv3): Add currentAppVersion and newAppVersion

abelfodil:
## Changes

Extract current and new `appVersion` from `Chart.yaml` or `index.yaml` for better dashboarding customization.

The goal is to know if a chart update truly upgrades an application when updating.

For instance, charts contained within truecharts.org update frequently, but the underlying application might not necessarily change. Furthermore, it is hard to know what's the current vs new version without scrolling through a lot of changelogs.

## Context

https://github.com/renovatebot/renovate/discussions/20017

## AI assistance disclosure

The public repository: https://github.com/arch-anes/self-hosted-services

See [logs.txt](https://github.com/user-attachments/files/28428033/logs.txt)


lib/modules/datasource/docker/index.ts
```diff
@@ -1212,6 +1217,33 @@ export class DockerDatasource extends Datasource {
...
+        if (
+          currentValue &&
+          currentValue !== latestTag &&
+          tags.includes(currentValue)
+        ) {
+          const currentLabels = await this.getLabels(
+            registryHost,
+            dockerRepository,
+            currentValue,
+          );
```
abelfodil, line 1236: I understand this is not ideal, as it leads to a double request to the repo, and would gladly take suggestions to make it better. In essence, for OCI charts, the `currentAppVersion` is not known until inspecting the `Chart.yaml` file.

lib/modules/datasource/docker/index.ts
```diff
@@ -532,6 +533,9 @@ export class DockerDatasource extends Datasource {
             if (url) {
               labels[sourceLabel] = url;
             }
+            if (configResponse.body.appVersion) {
+              labels[appVersionLabel] = configResponse.body.appVersion;
```
abelfodil, line 537: I understand this is probably not the right field to populate. Before engaging in a heavy refactor, I wanted to align on an approach first.

RahulGautamSingh (changes requested):
The linked discussion has an unresolved objection from @viceice (core maintainer) that this feature "is not possible" because `appVersion` is not a dependency Renovate detects. Before this merges, it would be good to address that directly.

 And, also a few open questions from the discussion:

  1. @djmcgreal-cc pointed out that `appVersion` can be overridden by chart users and charts can have multiple sub-dependencies, making the field ambiguous.
  How does this implementation handle that ambiguity?

  2. The original request was to surface `appVersion` in the Dependency Dashboard. This PR instead adds it as PR body template variables (`currentAppVersion`,
  `newAppVersion`)?

The implementation itself looks okay to me as the data is available in `index.yaml`. Just not sure if it falls under the bot's scope of work.

abelfodil:
Hi @RahulGautamSingh, to answer your questions:

>  How does this implementation handle that ambiguity?

This doesn't attempt to fix the underlying issue. I agree that since charts can ship with multiple components, resolving exact versions isn't really possible without deeper inspection. That said, in a project like https://github.com/arch-anes/self-hosted-services, the `appVersion` correctly describes the main component >90% of the time (only 4 out of 66 charts have an override). I'm not sure if that's a typical project though.
The original feature request was mainly for dashboarding, not automation. While it doesn't solve the problem of knowing deployed versions with absolute certainty, it does make the operator's life easier.

> The original request was to surface appVersion in the Dependency Dashboard. This PR instead adds it as PR body template variables (`currentAppVersion`, `newAppVersion`)?

If one uses `currentAppVersion` and `newAppVersion` in `commitMessageExtra`, then the dashboard should be updated. Please don't hesitate to correct me if I'm wrong.

RahulGautamSingh (commented):
On the dashboard: you are right. Once `currentAppVersion` and `newAppVersion` are template variables, adding them to `commitMessageExtra` or `prBodyColumns` surfaces them in PRs and the Dependency Dashboard, so the original request is covered.


On the ambiguity: I am fine with the heuristic. These are informational template variables only and do not influence any update decision, so an occasional mismatch is low risk.


The part I cannot decide on my own is scope. @viceice raised in the discussion that `appVersion` is not a dependency Renovate detects. 

Since this changes what data Renovate surfaces, it needs a core maintainer sign-off before we invest more. @viceice, could you confirm in on whether this belongs in Renovate?

PR #43749, opened 2026-06-02
fix(datasource/aws-machine-image): return all AMI releases for minimumReleaseAge filtering

gjasny:
## Changes

Previously, _getReleases only returned the latest image in the releases array, which broke minimumReleaseAge filtering. When the latest AMI was too new, there were no older candidates to fall back to.

Fixed by:
- _getReleases now returns all matching AMIs as separate releases (sorted oldest→newest)
- _getDigest simplified to read directly from the images array to ensure latest digest is still returned
- Added reproducer test verifying all 3 images are returned

This aligns with how other datasources (npm, PyPI, etc.) work — they return complete version history to allow minimumReleaseAge filtering to find an appropriately-aged candidate.

## Context

## AI assistance disclosure

The public repository: <URL>


lib/modules/datasource/aws-machine-image/index.ts
```diff
@@ -129,8 +129,7 @@ export class AwsMachineImageDatasource extends Datasource {
       return null;
     }
 
-    const res = await this.getReleases({ packageName: serializedAmiFilter });
-    return res?.releases?.[0]?.newDigest ?? /* v8 ignore next */ null; // TODO: needs test
+    return images[images.length - 1].Name ?? /* v8 ignore next */ null;
```
viceice:
```suggestion
    return images.at(-1)?.Name ?? null;
```
add a test. shouldn't it now use the passed version to find the proper digest? otherwise it will propose the digest from latest image even if you set minimum release age?

lib/modules/datasource/aws-machine-image/index.ts
```diff
@@ -151,21 +150,17 @@ export class AwsMachineImageDatasource extends Datasource {
     packageName: serializedAmiFilter,
   }: GetReleasesConfig): Promise<ReleaseResult | null> {
     const images = await this.getSortedAwsMachineImages(serializedAmiFilter);
-    const latestImage = images[images.length - 1];
-    if (!latestImage?.ImageId) {
+    if (!images.length || !images[images.length - 1].ImageId) {
```
viceice:
```suggestion
    if (!images.length || !images.at(-1)!.ImageId) {
```

lib/modules/datasource/aws-machine-image/index.ts
gjasny: Isn't that what the `if (newValue) {` block is doing? The docker data source behaves likewise:

```
const newTag = isNonEmptyString(newValue) ? newValue : 'latest';
```

The other option would be returning `null` if newValue is undefined.

lib/modules/datasource/aws-machine-image/index.ts
gjasny: applied

lib/modules/datasource/aws-machine-image/index.ts
gjasny: I believe the `with matching newValue, with 3 matching image to return the matching image` test covers that.

gjasny:
Hello @viceice,

could you please take a look, again?

Thanks,
Gregor

RahulGautamSingh (approved):

kilitr:
+1, this would solve a use case we've been struggling with. Being able to apply minimumReleaseAge to AMIs would make our rollout process much more predictable. Happy to test if that's helpful.

jamietanna (approved):

PR #43753, opened 2026-06-03
fix(manager/flux): pin tag-only OCIRepository digests to spec.ref.digest

alexwaibel:
## Changes

Updates the Flux manager so `OCIRepository` resources with tag-only refs get proper digest-pinning suggestions by writing `spec.ref.digest` rather than incorrectly suggesting `tag@sha256:...` style values in `spec.ref.tag`.

## Context

- [x] This closes an existing Issue, Closes: #42082

I opened a discussion here before starting since I didn't see the original issue but it didn't generate any new conversation: https://github.com/renovatebot/renovate/discussions/43505

## AI assistance disclosure

I used GitHub Copilot with opus 4.6 high reasoning to write the majority of the code in this change, but spent a decent amount of my own time steering it to iterate on the results until it reached a state I felt comfortable signing my name to and submitting for others to review. This is my first contribution to the project so AI was useful for orienting myself.

The public repository: <URL>

### DIFF
--- a/lib/modules/manager/flux/extract.ts
+++ b/lib/modules/manager/flux/extract.ts
@@ -1,5 +1,6 @@
 import { isString } from '@sindresorhus/is';
 import {
+  type Document,
   type Scalar,
   type YAMLMap,
   isMap,
@@ -189,9 +190,12 @@ function resolveSystemManifest(
 /**
  * Returns all `spec.ref` map nodes for OCIRepository resources matching `resourceName`.
  */
-function findOCIRefNodes(content: string, resourceName: string): YAMLMap[] {
+function findOCIRefNodes(
+  docs: Document.Parsed[],
+  resourceName: string,
+): YAMLMap[] {
   const refNodes: YAMLMap[] = [];
-  for (const doc of parseAllDocuments(content, { strict: false })) {
+  for (const doc of docs) {
     const docContents = doc.contents;
     if (!isMap(docContents)) {
       continue;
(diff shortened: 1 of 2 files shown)


lib/modules/manager/flux/extract.ts
```diff
@@ -371,7 +416,10 @@ function resolveResourceManifest(
           // Set currentValue to the tag so the docker datasource can look up the image's new digest
           combinedDep.currentValue = resource.spec.ref.tag;
 
-          const refRange = extractOCIRefRange(content, resource.metadata.name);
+          const refRange = extractOCIRefTagAndDigestRange(
+            content,
```
viceice, line 428: please parse the file content once and pass it to the functions (perf)

lib/modules/manager/flux/extract.ts
```diff
@@ -405,9 +461,25 @@ function resolveResourceManifest(
...
+            dep.autoReplaceStringTemplate = `${refTagRange.replaceString.replace(
+              resource.spec.ref.tag,
+              '{{newValue}}',
+            )}{{#if newDigest}}${newline}${refTagRange.indentation}digest: {{newDigest}}{{/if}}`;
+          } else {
+            logger.debug(
+              { file: manifest.file, name: resource.metadata.name },
+              'Unable to locate tag node for replacement (may be YAML alias or alias reference), skipping replacement',
+            );
+            dep.skipReason = 'invalid-value';
```
viceice: can we fallback to a simple replace as before? digest pinning is then not possible.

viceice (commented):
otherwise LGTM

lib/modules/manager/flux/extract.ts
alexwaibel: Good catch, thank you for your thorough review. I addressed your comment but while working through this I think I've come across an issue that perhaps warrants a separate issue or discussion thread:

When `tag` uses a YAML alias, the source of truth could be far from where we need to insert the digest. Renovate's single `replaceString` per dependency cannot express non-contiguous edits. The same applies when both tag and digest fields already exist but one or both use aliases.

From what I've seen other managers (Docker, etc.) keep tag+digest on a single line, so they don't encounter this non-contiguous edit problem. Flux will ignore the tag anyway for image pulls, so perhaps we could remove the tag field and include tag in a comment for readability but I think this has other implications. For instance it seems the flux cli [will display both tag and digest for the resource if present](https://fluxcd.io/flux/components/source/ocirepositories/#ocirepository-status) so I could imagine someone preferring to keep both.

Ultimately though that bug already exists in this manager so I don't think this PR makes the situation worse. I think it can probably be tracked as a separate issue.

viceice (approved):

PR #43775, opened 2026-06-03
feat(docker): use schema validated `getJson`

secustor:
## Changes

Use `getJson` instead of `getJsonUnchecked` for the `docker` datasource. 

08/13

Ignored following occurrence as we are not using the content at all 

https://github.com/renovatebot/renovate/blob/c908d078ef2245e9d3c353894cffcfa97b9042fb/lib/modules/datasource/docker/common.ts#L68

## Context

## AI assistance disclosure

The public repository: <URL>


RahulGautamSingh (approved):

lib/modules/datasource/docker/index.ts
```diff
@@ -657,19 +659,10 @@ export class DockerDatasource extends Datasource {
...
-        tags: {
-          name: string;
-        }[];
-        has_additional: boolean;
-      }
-
-      // typescript issue :-/
-      // oxlint-disable typescript/no-unnecessary-type-assertion
-      const res = (await this.http.getJsonUnchecked<QuayRestDockerTags>(
+      const res: HttpResponse<QuayTagsResponse> = await this.http.getJson(
```
viceice, line 662:
```suggestion
      const res = await this.http.getJson(
```

is the type still required here?

lib/modules/datasource/docker/schema.ts
```diff
@@ -188,3 +188,35 @@ export const DockerHubTagsPage = z.object({
...
+/**
+ * Quay registry tags list response (Quay v1 API).
+ */
+export const QuayTagsResponse = z.object({
+  tags: LooseArray(
+    z.object({
+      name: z.string(),
+    }),
+  ),
+  has_additional: z.boolean(),
```
viceice: should this be optional? sadly no response model 😕 

https://docs.quay.io/api/swagger/#!/tag/listRepoTags

secustor: I followed basically our previous interfaces here, but we can default to `false` for safety

lib/modules/datasource/docker/index.ts
secustor: Yes, there is an implicit `any` in that case. 

Explanation by Opus 4.8
```
  Look at the loop body:

  let url: string | null = pageUrl(page);
  while (url && page <= 20) {
    const res = await this.http.getJson(url, QuayTagsResponse);  // res depends on url
    ...
    url = res.body.has_additional ? pageUrl(page) : null;        // url depends on res
  }

  - res is initialized from getJson(url, …) → its type depends on url.
  - url is reassigned from res.body.has_additional → its type depends on res.

  url is annotated string | null, but the narrowed type at the call site (line 662) still goes through
  control-flow analysis, and the loop-back edge at line 666 makes that narrowing depend on res.body. So
  when res has no annotation, TS must infer it from getJson(url, …) → needs url's type → needs the
  line-666 assignment → needs res.body → needs res. The cycle closes, TS reports TS7022 and falls back to
  any.
 ```

viceice (approved):

PR #43776, opened 2026-06-03
feat: add PNPM_CONFIG_TRUST_LOCKFILE to allowedEnv

kopach:
## Changes

Add PNPM_CONFIG_TRUST_LOCKFILE to the allowedEnv preset: https://docs.renovatebot.com/presets-global/

This will also allow reducing compute time on Renovate infrastructure!

## Context

Renovate would fail with OoM error when used with PNPM v11.1.3+. PNPM 11.1.3 introduces important security update which now checks every dependency in lockfile against security policies (not only newly added packages as it was before)

PNPM supports outing out of this feature either by parameter or env variable `PNPM_CONFIG_TRUST_LOCKFILE`, see https://github.com/pnpm/pnpm/pull/11878

This PR allows setting `PNPM_CONFIG_TRUST_LOCKFILE=true`, so the memory-hungry policy check mechanism can be disabled.

P.S.
it is generally safe to disable this security mechanism for renovate especially if we have other checks on PR which will do the same during pacakge install.

## AI assistance disclosure

	-  documentation is here https://docs.renovatebot.com/presets-global/ couldn't find any other reference in repo

The public repository: <URL>

### DIFF
--- a/lib/config/presets/internal/global.preset.ts
+++ b/lib/config/presets/internal/global.preset.ts
@@ -9,6 +9,7 @@ export const presets: Record<string, GlobalPreset> = {
       'RUSTC_BOOTSTRAP',
       'PNPM_WORKERS',
       'PNPM_MAX_WORKERS',
+      'PNPM_CONFIG_TRUST_LOCKFILE',
     ],
     description:
       'Hopefully safe environment variables to allow users to configure.',


lib/config/presets/internal/global.preset.ts
```diff
@@ -9,6 +9,7 @@ export const presets: Record<string, GlobalPreset> = {
       'RUSTC_BOOTSTRAP',
       'PNPM_WORKERS',
       'PNPM_MAX_WORKERS',
+      'PNPM_CONFIG_TRUST_LOCKFILE',
```
kopach, line 12: I think we can go even with `PNPM_*` (same as for `GO*`), but decided to be more conservative. Please let me know if you agree to have it with glob.

jamietanna:
How are you running Renovate? A Mend-hosted app, or self-hosted?

RahulGautamSingh (approved):

kopach:
> How are you running Renovate? A Mend-hosted app, or self-hosted?

@jamietanna, it's Mend-hosted app

lib/config/presets/internal/global.preset.ts
kopach: I have tested this env variable locally and can confirm, when it is set - policy check process of PNPM is skipped

jamietanna (commented):
Why would we like to use the environment variable configuration, rather than making it explicit [in a user's `pnpm-workspace.yaml`](https://pnpm.io/settings#trustlockfile)?

jamietanna:
(For future reference of maintainers reviewing this - in case we want to merge it when I'm OOO - I don't strongly disagree with this being available as an option, especially for Mend-hosted, but I want to understand why and if this is really something we want to make more of a recommended solution, or if it's something that can be managed through user configuration in their `pnpm-workspace.yaml`)

kopach:
> Why would we like to use the environment variable configuration, rather than making it explicit [in a user's `pnpm-workspace.yaml`](https://pnpm.io/settings#trustlockfile)?

@jamietanna, good question. The point is that this feature of PNPM is extremely valuable from a security perspective, and I don't want to disable it by default in entire repo for everyone (disabling it in `pnpm-workspace.yaml` would do just that). I want that each install (dev local or on CI) do trigger this verification process. However, for Renovate  runner it is a bit overkill for couple of reasons:
1. this PNPM's check was a response to potential cache pollution on GH runners (we don't have that here, also Renovate runners are isolated from CI pipeline)
2. I do have my other CI check which verify just that, so I don't need to run it also here once again.

kopach:
btw, @jamietanna would you be OK, if I change this to glob like `PNPM_*` ? There are many more potential tweaks to PNPM (including upcoming). All those should be safe to set from Renovate perspective. What do you think?

jamietanna:
If I'm honest, I'm still not sure what the benefit of this config would be.

[Via the docs](https://pnpm.io/settings#trustlockfile):

> A poisoned lockfile (one a contributor authored under a weaker policy than CI enforces) can slip through, so leave this false whenever outside collaborators can edit the lockfile.

Given the mix of contributors (Renovate, humans, other bots), I'm not sure if disabling it does make sense? And having Renovate fail to process a repo - if pnpm hit this issue - is surely a better case than it making its way into your CI?

> if I change this to glob like `PNPM_*` ? [...] All those should be safe to set from Renovate perspective. What do you think?

I'm not sure I agree, I'd prefer to be more conservative with what we make possible for users to specify

kopach:
> If I'm honest, I'm still not sure what the benefit of this config would be.

The main idea was to resolve OOM issue which happens now with recent versions of PNPM.
We could have `trustLockfile` enabled on Renovate and have another CI check to ensure all packages pass security policies.

Anyway, tested it another way (via pnpm-workspace) and this wouldn't help. So, PR can be closed. Thanks

PR #39710, opened 2025-12-02
test(config/validation): remove untested `bumpVersions` code

jamietanna:
## Changes

The test for this has been broken for some time, and only highlighted
when we start to look at the reason the test is failing - we're
providing a `bumpVersion` but we're intending it to be the
`bumpVersions` config.

Even if we fix the code, this then highlights that we have no validation
code in place, so the tests fail.

We'll follow-up to re-add validation in the future, but for now we can
remove a test that isn't testing what we're intending it to.

## Context

Please select one of the below:

- [ ] This closes an existing Issue: #
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [x] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #43799, opened 2026-06-04
test(config/validation): explicitly check error/warning messages

jamietanna:
## Changes

Previously, we had a number of tests checking for i.e.
`expect(warnings).toHaveLength(1);` which wouldn't necessarily be the
warning that we were expecting to see, nor would break if the message
changed (un)intentionally.

We can refactor this and:

- assert on the contents of `errors`/`warnings`
- check the `topic` only if it's interesting - for instance if it's a
  `Configuration Error`, but under `warnings`
- removed a `bumpVersion` test that doesn't make sense
- removed snapshot tests

This also reduces our usage of snapshot tests, as we're trying to remove
them where possible.

Also, refactor to use `toBeEmptyArray()`

## Context

## AI assistance disclosure

The public repository: <URL>


jamietanna:
We still have some `toMatchInlineSnapshot(...)`s which I think would be good to replace, but they're not as bad

viceice (dismissed):

lib/config/validation.spec.ts
```diff
@@ -93,7 +109,6 @@ describe('config/validation', () => {
         // @ts-expect-error -- invalid config
         config,
       );
-      expect(warnings).toHaveLength(2);
```
secustor, line 95: I remember being asked to add such length tests additionally to matchObject as the latter would also match if there are additional elements. 

Is this no longer true?

jamietanna: No, this is the case still. That being said, in a lot of the tests we have, we're not actually checking the values, only that there are values present, and as we've seen with i.e. https://github.com/renovatebot/renovate/pull/39710 that means we're not actually testing what we should do

_Technically_ if we hit this in the future and it's a bit of a pain, we could "no-op" the test

IMO, I'm happy with seeing how this goes, and if it does become awkward - I'd prefer us to have much stricter and clearer coverage of what we're changing in this file

secustor (approved):

PR #43870, opened 2026-06-08
fix(manager/npm): use packageName for pnpm override minimumReleaseAgeExclude

RahulGautamSingh:
## Changes

Fix malformed `minimumReleaseAgeExclude` entries when a pnpm `overrides` key carries a version constraint (e.g. `qs@<6.14.1`).

Previously the full override key was used as the dep name, producing invalid entries like `qs@<6.14.1@6.15.0` ("Invalid versions union") which blocked lockfile regeneration for security updates. The exclude entry now uses the parsed `packageName`, yielding `qs@6.15.0`.

## Context

## AI assistance disclosure

Used Claude Code (Claude Opus 4.8) to diagnose the bug, scope the fix in `lib/modules/manager/npm/artifacts.ts`, and add a regression test.

### DIFF
--- a/lib/modules/manager/npm/artifacts.spec.ts
+++ b/lib/modules/manager/npm/artifacts.spec.ts
@@ -387,6 +387,40 @@ minimumReleaseAgeExclude:
       ]);
     });
 
+    it('uses packageName for minimumReleaseAgeExclude when override key has a version constraint', async () => {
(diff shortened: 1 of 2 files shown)


RahulGautamSingh:
Needs testing on a test repo.

RahulGautamSingh:
It seems main already has this fix, maybe another pr was merged.

PR #43878, opened 2026-06-08
chore: improve renovate maintenance path

quyentonndbs:
Summary:
- Add or tighten focused edge-case tests or type assertions in test/util.ts, test/setup.ts related to TypeScript, frontend tooling, test stability, build fixes; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.


jamietanna:
I don't think we want to accept this - please re-raise, following our template, if you think this is still necessary

PR #41347, opened 2026-02-20
feat(platform-azure): initial bearer token auth support for Azure DevOps

bastianwenske:
## Changes

Add automatic support for Microsoft Entra ID (AAD) Bearer token authentication for Azure DevOps, alongside existing Personal Access Token (PAT) support. Token type is auto-detected — no new configuration options required.

### What changed:
- `util.ts`: Added `isProbablyJwt(token)` helper that detects JWT tokens by checking for 3 dot-separated base64url segments with a decodable JSON header containing `typ` or `alg` fields
- `util.ts`: `getStorageExtraCloneOpts()` now uses `isProbablyJwt()` instead of the `token.length === 52` heuristic to choose between `Bearer` and `Basic` auth headers for git clone
- `azure-got-wrapper.ts`: `getAuthenticationHandler()` uses `isProbablyJwt()` to select `BearerCredentialHandler` for JWT tokens or `PersonalAccessTokenCredentialHandler` for PATs, replacing the library's length-based heuristic
- Added documentation in `lib/modules/platform/azure/readme.md`

### Why:
The previous implementation relied on a fragile heuristic (token length === 52 or regex pattern) from the `azure-devops-node-api` library to distinguish PATs from Bearer tokens. Microsoft Entra ID tokens (OAuth/JWT) are increasingly the recommended authentication method for Azure DevOps Services, especially for service principals and managed identities. This change makes token type detection deterministic by inspecting the token structure (JWT format) rather than relying on length.

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: #
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

GitHub Copilot (Claude Opus 4.6) was used to design, implement, and test the feature, including: code changes across 4 files, unit tests (16 new test cases), documentation, and this PR description.

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [x] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
[...]

PR #43888, opened 2026-06-09
feat(platform-azure): initial bearer token auth support for Azure DevOps

jonnotjohn:
## Continuation note

This continues #41347 by @bastianwenske, who is the original author of this feature. The PR had been idle for a couple of months with a few requested changes outstanding; Bastian kindly gave his blessing for me to finish it and get it merged: https://github.com/renovatebot/renovate/pull/41347#issuecomment-4629157280

His original commits are preserved. Changes I made in addition to address the reviewer notes:

- Parse the JWT header with the codebase's `parseJsonc` util instead of raw `JSON.parse` (per @RahulGautamSingh / @viceice's comment)
- Fixed the "bearer token" test to use a real JWT (it previously used "testtoken" that actually called the PAT path), reusing the existing `buildJwt()` test helper
- Added `logger.debug` lines so the detected token type (Bearer vs PAT) is visible when troubleshooting auth
- Merged latest `main` and resolved the conflict

## Changes

Add automatic support for Microsoft Entra ID (AAD) Bearer token authentication for Azure DevOps, alongside existing Personal Access Token (PAT) support. Token type is auto-detected — no new configuration options required.

### What changed:

- `util.ts`: Added `isProbablyJwt(token)` helper that detects JWT tokens by checking for 3 dot-separated base64url segments with a decodable JSON header (parsed via `parseJsonc`) containing a `typ` or `alg` field
- `util.ts`: `getStorageExtraCloneOpts()` now uses `isProbablyJwt()` instead of the `token.length === 52` heuristic to choose between `Bearer` and `Basic` auth headers for git clone
- `azure-got-wrapper.ts`: `getAuthenticationHandler()` uses `isProbablyJwt()` to select `BearerCredentialHandler` for JWT tokens or `PersonalAccessTokenCredentialHandler` for PATs, replacing the library's length-based heuristic, and logs which type was detected
- Added documentation in `lib/modules/platform/azure/readme.md`

### Why:

The previous implementation relied on a fragile heuristic (token length === 52, or a regex) from the `azure-devops-node-api` library to distinguish PATs from Bearer tokens. Microsoft Entra ID tokens (OAuth/JWT) are increasingly the recommended authentication method for Azure DevOps Services, especially for service principals and managed identities. Inspecting the token structure (JWT format) makes detection deterministic rather than length-based.

## Context

## AI assistance disclosure

The original feature (by @bastianwenske) was developed with GitHub Copilot (Claude Opus 4.6), which generated non-trivial portions of the code, tests, and documentation. For this continuation I made the edits by hand, but used Claude to help guide them, orienting me in the codebase and advising on the review changes (e.g. which JSON util to use, the test-token fix). I reviewed every edit and verified the code paths and tests myself.

The Azure platform unit tests pass locally (`pnpm vitest lib/modules/platform/azure/`), including the tests covering JWT  vs PAT detection in `isProbablyJwt()`, `getStorageExtraCloneOpts()`, and `getAuthenticationHandler()`. I also confirmed via debugging that a JWT routes  to the Bearer handler and an opaque/PAT token routes to the PAT handler. `pnpm check` also passes locally.


viceice (dismissed):

lib/modules/platform/azure/azure-got-wrapper.ts
```diff
@@ -16,7 +22,12 @@ function getAuthenticationHandler(config: HostRule): IRequestHandler {
     return getBasicHandler(config.username, config.password, true);
   }
   // TODO: token can be undefined here (#22198)
-  return getHandlerFromToken(config.token!, true);
+  if (isProbablyJwt(config.token!)) {
```
RahulGautamSingh:
```suggestion
  if (config.token && isProbablyJwt(config.token!)) {
```

and getting rid of the comment todo above

RahulGautamSingh (changes requested):

lib/modules/platform/azure/azure-got-wrapper.ts
jonnotjohn: Done in f8d0102. I applied the guard and removed the todo. I also dropped the non-null assertions in the block because the guard makes them redundant.

lib/modules/platform/azure/azure-got-wrapper.ts
```diff
@@ -15,8 +21,12 @@ function getAuthenticationHandler(config: HostRule): IRequestHandler {
...
     return getBasicHandler(config.username, config.password, true);
   }
-  // TODO: token can be undefined here (#22198)
-  return getHandlerFromToken(config.token!, true);
+  if (config.token && isProbablyJwt(config.token)) {
+    logger.debug('Using Bearer authentication (JWT detected)');
+    return getBearerHandler(config.token, true);
+  }
+  logger.debug('Using PAT authentication');
+  return getPersonalAccessTokenHandler(config.token!, true);
```
viceice, line 30: the comment about missing token should be added back here as it's still an issue

jonnotjohn: Done in 4d2df2d. I added the todo comment above the return. I considered a guard/throw instead but the guard in azureObj() already guarantees the token will be defined at the return so the assertion is just to make TS happy.

lib/modules/platform/azure/util.ts
```diff
@@ -116,18 +118,44 @@ export function getRenovatePRFormat(azurePr: GitPullRequest): AzurePr {
...
+ * Bearer tokens from Azure DevOps Personal Access Tokens (PATs).
+ */
+export function isProbablyJwt(token: string): boolean {
+  const parts = token.split('.');
+  if (parts.length !== 3) {
+    return false;
+  }
+
+  try {
+    const header = parseJsonc(
```
secustor: Why `Jsonc`?

lib/modules/platform/azure/util.ts
```diff
@@ -116,18 +118,44 @@ export function getRenovatePRFormat(azurePr: GitPullRequest): AzurePr {
...
+ * Detects whether a token string is likely a JWT (JSON Web Token).
+ *
+ * JWTs consist of three base64url-encoded segments separated by dots.
+ * The first segment (header) must decode to valid JSON containing
+ * at least a `typ` or `alg` field.
+ *
+ * This is used to automatically distinguish Microsoft Entra ID (AAD)
+ * Bearer tokens from Azure DevOps Personal Access Tokens (PATs).
+ */
+export function isProbablyJwt(token: string): boolean {
```
secustor: I would prefer have this in a new`/lib/util/http/jwt.ts`

lib/modules/platform/azure/readme.md
```diff
@@ -18,6 +20,64 @@ Permissions for your PAT should be at minimum:
 
 Remember to set `platform=azure` somewhere in your Renovate config file.
 
+### Microsoft Entra ID Bearer Token (OAuth)
+
+Renovate also supports authenticating to Azure DevOps Services using Microsoft Entra ID (formerly Azure AD) Bearer tokens.
+This works with service principals, managed identities, and user-delegated tokens.
+
+Renovate **automatically detects** whether a token is a JWT (Bearer token) or a PAT based on its format.
+No additional configuration is needed — just set the `token` field to your Entra ID access token:
```
RahulGautamSingh:
```suggestion
No additional configuration is needed. Just set the `token` field to your Entra ID access token:
```

lib/util/http/jwt.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Detects whether a token string is likely a JWT (JSON Web Token).
+ *
+ * JWTs consist of three base64url-encoded segments separated by dots.
+ * The first segment (header) must decode to valid JSON containing
+ * at least a `typ` or `alg` field.
+ *
+ * This is used to automatically distinguish Microsoft Entra ID (AAD)
+ * Bearer tokens from Azure DevOps Personal Access Tokens (PATs).
+ */
+export function isProbablyJwt(token: string): boolean {
```
RahulGautamSingh: type the param `token: string | undefined` and return `false` on falsy

you will not need the two non-null assertions then

lib/util/http/jwt.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+      ).toString('base64url');
+      expect(isProbablyJwt(`${seg}.${seg}.${seg}.${seg}`)).toBeFalse();
+    });
+
+    it('returns false when header decodes to a non-object', () => {
+      const header = Buffer.from('"just a string"').toString('base64url');
+      const payload = Buffer.from(JSON.stringify({ sub: 'x' })).toString(
+        'base64url',
+      );
+      const sig = Buffer.from('sig').toString('base64url');
```
RahulGautamSingh: these couple of lines are repeated in this test file and in the other 2 test files of the PR. please extract a helper and dedupe

RahulGautamSingh (changes requested):

lib/modules/platform/azure/azure-got-wrapper.spec.ts
```diff
@@ -1,4 +1,5 @@
 import type * as _hostRules from '../../../util/host-rules.ts';
+import { buildTestJwt } from '../../../util/http/jwt.spec.ts';
```
secustor: If it is needed in multiple locations rather use `./test/` as location.

lib/util/http/jwt.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Detects whether a token string is likely a JWT (JSON Web Token).
+ *
+ * JWTs consist of three base64url-encoded segments separated by dots.
+ * The first segment (header) must decode to valid JSON containing
+ * at least a `typ` or `alg` field.
+ *
+ * This is used to automatically distinguish Microsoft Entra ID (AAD)
+ * Bearer tokens from Azure DevOps Personal Access Tokens (PATs).
```
secustor:
```suggestion
```
As this is now a generic handler this is not relevant.

secustor (commented):
else LGTM

lib/modules/platform/azure/util.ts
```diff
@@ -122,12 +123,12 @@ export function getStorageExtraCloneOpts(config: HostRule): GitOptions {
...
+  } else if (config.token && isProbablyJwt(config.token)) {
+    authType = 'bearer';
+    authValue = config.token;
+  } else {
     authType = 'basic';
     authValue = toBase64(`:${config.token}`);
-  } else {
-    authType = 'bearer';
-    authValue = config.token!;
   }
```
viceice, line 132: doesn't this break exising token which are longer than 52 chars which should be using bearer and now using basic auth?

lib/modules/platform/azure/readme.md
```diff
@@ -18,6 +20,64 @@ Permissions for your PAT should be at minimum:
...
+For **managed identities** (Azure-hosted Renovate):
+
+```bash
+az login --identity
+az account get-access-token --resource 499b84ac-1321-427f-aa17-267ca6975798 --query accessToken -o tsv
+```
+
+#### Requirements for service principals / managed identities
+
+1. The identity must be added to the Azure DevOps organization (Organization Settings → Users).
```
RahulGautamSingh, line 71: use 1. 1. 1. dor the numbered list

test/jwt-util.ts
```diff
@@ -0,0 +1,12 @@
+type StrObjNull = string | object | null;
+
+export function buildTestJwt(
+  header: StrObjNull,
+  payload: StrObjNull,
+  sig: StrObjNull,
+): string {
+  const encode = (segment: StrObjNull): string => {
```
RahulGautamSingh: move this out and don't use arrow functions

RahulGautamSingh (changes requested):

lib/modules/platform/azure/util.ts
jonnotjohn: This shouldn't break any existing tokens and it actually handles a case the old version didn't. Entra tokens in ADO are always JWTs, so they're guaranteed to hit the bearer path via the isProbablyJwt check. The only other auth type we support for ADO are PAT tokens which are basic auth. Microsoft changed ADO PAT tokens to 84 characters, so the old length check here would have incorrectly set a new PAT token to bearer but the new logic handles any length correctly.

The only case with a non-JWT bearer auth would be the legacy ADO OAuth, but Renovate doesn't document support for that and it's being retired this year anyway.

RahulGautamSingh: Thanks, that holds up. 

I checked main: the old code defaulted to bearer for anything that was not exactly 52 chars, so an 84-char PAT was already being sent as bearer incorrectly. 

Keying off the JWT structure fixes that and makes this path consistent with the new `getAuthenticationHandler`, where both default to PAT for non-JWT tokens. 

No concern from me, good to resolve.

RahulGautamSingh (approved):

secustor (approved):

PR #43891, opened 2026-06-09
fix(prune): skip branch pruning when branchPrefix is empty

RahulGautamSingh:
## Changes

When `branchPrefix` is empty (or whitespace-only), `pruneStaleBranches` now skips pruning entirely instead of filtering branches with `startsWith(config.branchPrefix)`.

An empty prefix makes `startsWith('')` match **every** branch in the repo (`getBranchList()` returns all remote heads via `git ls-remote --heads`), so unrelated branches — including the base branch — get treated as Renovate-managed and deleted as orphans. The `isBranchModified` check is only a heuristic and does not reliably protect clean branches.

Uses `isNonEmptyStringAndNotWhitespace`, consistent with the existing check in `calculateBaseBranchRegex` in the same file.

Trade-off: empty-prefix users no longer get automatic stale-branch pruning. This is unavoidable — with no static prefix, Renovate cannot reliably identify its own branches — and is strictly safer than deleting unrelated branches.

## Context

## AI assistance disclosure


lib/workers/repository/finalize/prune.ts
```diff
@@ -161,6 +161,15 @@ export async function pruneStaleBranches(
     logger.debug('No defaultBranch set - skipping branch pruning');
     return;
   }
+  if (!isNonEmptyStringAndNotWhitespace(config.branchPrefix)) {
+    // An empty branchPrefix matches every branch in the repo, so Renovate
+    // cannot reliably tell its own branches apart from unrelated ones. Skip
+    // pruning to avoid deleting non-Renovate branches as orphans.
+    logger.debug(
```
viceice, line 168: I think this should be a warning. setting branch prefix to an empty string isn't official supported because of the side effects. maybe we should treat that as a config error.


@secustor @jamietanna wdyt?

secustor: I agree this should be a config error and if we can not efficiently detect this is empty because of templating then it should be at least here a warning.

RahulGautamSingh: converted to warn

secustor (commented):
Needs merge from main, or something is triggering the breaking change. else LGTM

viceice (approved):

viceice:
> Needs merge from main, or something is triggering the breaking change. else LGTM

@RahulGautamSingh ☝️

viceice (approved):

secustor (approved):

PR #43901, opened 2026-06-09
fix(versioning): use BigInt to avoid integer precision loss in gradle, maven and loose

RahulGautamSingh:
## Changes

Version tokenizers parsed numeric parts with `parseInt`/`Number`, losing precision above `Number.MAX_SAFE_INTEGER` (~9e15). Timestamp-style versions (e.g. `2026051723231779060202`) rounded to the same float, so `compare()` returned `0` ("equal") and no updates were detected.

This switches numeric tokens to `BigInt` in the `gradle`, `maven` and `loose` schemes so large versions compare with full precision. `getMajor`/`getMinor`/`getPatch` keep their `number | null` contract via `Number(...)`.

Reported in discussion https://github.com/renovatebot/renovate/discussions/43676.

## Context

## AI assistance disclosure

Code and tests written with Claude Code (Claude Opus 4.8).


zharinov (changes requested):
Just added some edge case handling, please check it out and merge if you agree:

- https://github.com/RahulGautamSingh/renovate/pull/36

lib/modules/versioning/maven/index.ts
```diff
@@ -116,7 +120,42 @@ function isGreaterThan(a: string, b: string): boolean {
...
+  if (!(isVersion(a) && isVersion(b))) {
+    return false;
+  }
+
+  const left = getExactSection(a, type);
+  const right = getExactSection(b, type);
+  return left === right;
+}
+
+const isStable = (version: string): boolean => {
```
RahulGautamSingh:
```suggestion
function isStable(version: string): boolean {
```

lib/workers/repository/process/lookup/update-type.ts
```diff
@@ -26,9 +26,15 @@ export function getUpdateType(
   ) {
     return 'major';
   }
+  const newMinor = versioningApi.getMinor(newVersion)!;
+  const currentMinor = versioningApi.getMinor(currentVersion)!;
```
viceice, line 30: we should add a comment to correctly handle null here later because I think this assumption is wrong

viceice (commented):
otherwise LGTM

lib/workers/repository/process/lookup/update-type.ts
RahulGautamSingh: added guard instead

viceice (approved):

PR #43917, opened 2026-06-10
docs: switch from pdm to uv for docs dependencies

felipecrs:
## Changes

This PR switches the docs dependencies from pdm to uv. This is better because:

1. uv is way faster. It's written in Rust.
2. uv manages its own Python installation from `.python-version`, so we don't need to install Python through other means like Mise.
3. uv is distributed as a standalone binary, so we (and Mise) don't need to install it through the likes of pipx.

In my opinion uv is the best Python package manager right now.

## Context

## AI assistance disclosure

The public repository: <URL>


.github/workflows/build.yml
```diff
@@ -711,15 +711,15 @@ jobs:
...
 
-      - name: Setup PDM
-        uses: pdm-project/setup-pdm@973541a5febeafcfdadf8a51211435be6ecfd90f # v4.5
+      - name: Setup uv
+        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
         with:
-          python-version-file: .python-version
-          version: ${{ env.PDM_VERSION }}
-          cache: true
+          version: ${{ env.UV_VERSION }}
```
viceice: ist this action already supported native, so we can simply add the UV version here?

.gitignore
```diff
@@ -34,8 +34,8 @@ test/datasource/nuget/_fixtures/obj
 /docker-compose.yml
 /report*.json
 
-# pdm
-/.pdm-python
```
viceice, line 38: better leave it here for old checkouts

pyproject.toml
```diff
@@ -1,11 +1,13 @@
 [project]
+name = "renovate-docs"
+version = "0.0.0"
```
viceice, line 3: do we really need that?

.prettierignore
```diff
@@ -24,7 +24,7 @@ lib/modules/datasource/hex/v2/signed.ts
 **/tmp/
 /.venv/
 
-pdm.lock
```
viceice, line 27: same here

pyproject.toml
felipecrs: Yes, both are mandatory. That also bothers me.

- https://github.com/astral-sh/uv/issues/18114

.github/workflows/build.yml
felipecrs: Yes, that's totally right. Thanks for the catch.

viceice (dismissed):

felipecrs:
Had to dismiss your review because of merge conflict resolution, @viceice.

uv.lock
secustor, line 1: We should use pylock if possible

.prettierignore
```diff
@@ -24,9 +24,12 @@ lib/modules/datasource/hex/v2/signed.ts
...
+uv.lock
 pnpm-lock.yaml
 
+# For old checkouts
+pdm.lock
```
secustor, line 31: Should be removed IMO, we shouldn't keep old references floating around

.gitignore
```diff
@@ -34,8 +34,12 @@ test/datasource/nuget/_fixtures/obj
 /docker-compose.yml
 /report*.json
 
-# pdm
+# uv
+/.venv
+
+# For old checkouts
 /.pdm-python
+/pdm.lock
```
secustor, line 42: Same here

.prettierignore
felipecrs: Please check https://github.com/renovatebot/renovate/pull/43917#discussion_r3390691813.

I personally agree with @viceice. I ran into this situation a few times already.

Maybe the best is to keep it around for some time, maybe a year, and then remove it.

uv.lock
felipecrs: pylock doesn't have everything uv needs, and therefore they decided not to make it a first-class citizen:

https://docs.astral.sh/uv/concepts/projects/layout/#relationship-to-pylocktoml

uv can generate a pylock, and can use pylock to reinstall, but requires specific command line arguments. It's not seamless or transparent, cannot be configured as a preference in pyproject.toml for example.

felipecrs: So I'd recommend we stick with uv.lock for now and revisit this at another time.

Relevant discussion:

- https://github.com/astral-sh/uv/issues/12584

.github/workflows/build.yml
```diff
@@ -711,15 +710,15 @@ jobs:
...
-          version: ${{ env.PDM_VERSION }}
-          cache: true
+          version: 0.11.19
+          enable-cache: true
+          cache-python: true
 
-      - name: Install pdm dependencies
-        run: pdm install
+      - name: Install uv dependencies
+        run: uv sync
```
secustor:
```suggestion
        run: uv sync --locked
```
Or is UV detecting automatically that is running in CI ?

pyproject.toml
zanieb: Use a dependency group instead https://peps.python.org/pep-0735/ that's what they're for.

uv will sync the `dev` group by default.

.github/workflows/build.yml
felipecrs: No, and that's a very nice catch. Thank you. Fixed in 41d9101779532c2042ad40bd8f5f18b8c8549704.

pyproject.toml
felipecrs: @zanieb that works, but if I remove `[project]`, then uv warns:

```
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.14`.
```

And if I keep `project.requires-python` only, then it fails because `name` and `version` aren't set.

There's no clean way out. Unless I'm missing something.

zanieb: Ah you can add a `requires-python` range to the dependency-group for uv

https://docs.astral.sh/uv/concepts/projects/dependencies/#group-requires-python

Unfortunately they missed including that in the standard, but I think it'll be added in the future now that it's clear it's useful.

felipecrs: @zanieb that doesn't seem to be enough to clear the warning:

```console
$ cat <<'EOF' > pyproject.toml
[dependency-groups]
dev = ["mkdocs-material==9.7.6", "mkdocs-awesome-pages-plugin==2.10.1"]


[tool.uv.dependency-groups]
dev = {requires-python = ">=3.11"}


[tool.uv]
package = false
EOF

$ uv sync
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.14`.
Resolved 33 packages in 1ms
Checked 33 packages in 0.59ms
```

zanieb: That seems like a bug, I can look into that.

felipecrs: @zanieb or perhaps this warning shouldn't exist? There is no practical use for `requires-python` here anyway. If possible, we could simply omit it.

https://peps.python.org/pep-0621/#requires-python

It's not listed as mandatory.

zanieb: We need a range during resolution to ensure that your dependencies are compatible with the Python versions you'll use. We can assume a range based on the current Python version, but it can lead to unexpected results, hence the warning.

secustor (approved):

PR #43921, opened 2026-06-11
refactor(cache): prepare package cache for binary payloads

zharinov:
## Changes

Prepare the package cache for binary payloads.

- `compress.ts` adds `compressToBuffer` and `decompressFromBuffer` for buffer-based brotli. The existing base64 functions are reimplemented on top.
- `codec.ts` introduces a binary envelope format:
   - 8-byte header:
     - magic bytes (`0x11 0x91`) — collision-proof against legacy data
       - Redis and `cacache`: JSON starts with `{`
       - SQLite: valid brotli but never starts with `0x1191` per [RFC 7932 §9.1](https://datatracker.ietf.org/doc/html/draft-alakuijala-brotli-09#section-9.1)
     - version byte (`0x01`)
     - reserved byte
     - `cachedAt` as uint32 Unix seconds (not used in new model, but worth keeping)
   - body: raw brotli content
- `legacy.ts` adds decode-only support for the two old formats (JSON wrapper for file/Redis, raw brotli for SQLite). To be deleted couple months later.
- `PackageCacheBase` now owns `get()`. Backends expose `readRaw()` and `rm()` instead of duplicating read logic. The base handles format detection, legacy fallback, expiry, and corrupt-entry cleanup.

Read-path behavior changes:
- Corrupt or undecodable entries are now removed and logged once at debug level, instead of silently persisting.
- The new format degrades to a clean miss on every backend, so the follow-up PR can be safely rolled back.
- All backends still write their existing formats. No stored bytes change yet.

## Context

## AI assistance disclosure

### DIFF
--- a/lib/util/cache/package/codec.ts
+++ b/lib/util/cache/package/codec.ts
@@ -1,0 +1,84 @@
+import { DateTime } from 'luxon';
+import { compressToBuffer, decompressFromBuffer } from '../../compress.ts';
+
+const headerLength = 8;
+const prefixWord = 0x11910100;
+
+/**
+ * Format detection must not collide with anything older Renovate versions
+ * stored. Legacy entries are either JSON wrappers, which always start with `{`
+ * (`0x7b`), or raw brotli blobs from the SQLite backend.
+ *
+ * Brotli streams start with the WBITS field (RFC 7932, section 9.1). Standard
+ * brotli reserves `0010001` as an invalid bit pattern. The only byte values
+ * whose low seven bits match that pattern are `0x11` and `0x91`, so no valid
+ * legacy brotli stream can start with either byte.
+ *
+ * This does not account for `BROTLI_PARAM_LARGE_WINDOW`, which Renovate has
+ * never enabled. Both bytes also fall outside the base64 alphabet and printable
+ * JSON, giving the prefix extra margin against other legacy-looking payloads.
+ */
+const magic0 = 0x11;
+const magic1 = 0x91;
+
+const formatVersion = 0x01;
+const reserved = 0x00;
+
+export interface EnvelopeEntry {
+  value: unknown;
+  cachedAt: DateTime;
+}
+
+export function isEnvelope(data: Buffer): boolean {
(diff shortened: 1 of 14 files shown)


lib/util/cache/package/codec.ts
```diff
@@ -0,0 +1,84 @@
...
+export function isEnvelope(data: Buffer): boolean {
+  return data[0] === magic0 && data[1] === magic1;
+}
+
+export async function encodeEntry(
+  value: unknown,
+  cachedAt: DateTime,
+): Promise<Buffer> {
+  const cachedAtSeconds = cachedAtToSeconds(cachedAt);
+  const compressed = await compressToBuffer(JSON.stringify(value));
```
viceice, line 41: should we differentiate if the passed value is already a buffer? so we don't encode it as base64 in json?

lib/util/cache/package/impl/base.ts
```diff
@@ -14,4 +54,32 @@ export abstract class PackageCacheBase {
...
+  ): MaybePromise<void>;
+
+  private async removeInvalidEntry(
+    namespace: PackageCacheNamespace,
+    key: string,
+  ): Promise<void> {
+    try {
+      await this.rm(namespace, key);
+    } catch (err) {
+      logger.once.debug({ err }, 'Error while removing package cache value');
```
viceice, line 75: should we add namespace and key here? 🤔

viceice (commented):
otherwise LGTM

lib/util/cache/package/impl/base.ts
zharinov: It means "once per (namespace, key)" pair which probably not something we want for IO errors.

lib/util/cache/package/codec.ts
zharinov: Nobody does this so far, otherwise we'd get errors trying to use values after reading. And it complicates things, so I'd rather excluded Buffer from the broad type which cache API accepts/returns.

But convention also could be just fine.

viceice (approved):

PR #43922, opened 2026-06-11
feat(cache): use binary envelopes for package cache writes

zharinov:
## Changes

Package cache writes store binary envelopes instead of JSON-wrapped base64 strings.

- `PackageCacheBase` owns `set()`. It encodes the value into an envelope and passes the bytes to the backend's `writeRaw()`.
- Each backend stores opaque bytes and handles TTL with its native mechanism:
  - Redis writes the bytes with `EX`.
  - SQLite writes the bytes to its `data` blob and keeps its expiry column.
  - File cache writes the bytes with `cacache` and stores expiry in the index metadata.
- A non-positive TTL deletes the entry instead of writing a value that is already expired.
- File-cache cleanup reads expiry from `cacache` index metadata. It no longer opens and parses every content file, and the in-memory expiry map is gone.
- All backends share one codec, so SQLite payloads now use brotli quality 8 like the others.

Reads still fall back to the legacy JSON-wrapper decoder, so existing entries keep working. `legacy.ts` stays until pre-envelope entries age out.

Notes:

- New writes overwrite legacy entries in place. Keys are unchanged.
- The first file-cache cleanup after upgrading removes legacy entries that have no expiry metadata. Later runs rewrite them as envelopes.
- Smaller payloads: the JSON wrapper and base64 layer are gone.

## Context

## AI assistance disclosure

The public repository: N/A — local Redis smoke tests covered single-node Redis and a 3-node Redis cluster.


zharinov:
This is smooth for the SQLite and Redis backends. For the `cacache`, it's cache purge + cold start after update.

zharinov:
Just updated `cacache` case to migrate on read. If read migrates 1/10 items, 9/10 are still purged, and second run will be a cold start for sharded scenarios. Do we regard this as acceptable?

viceice (approved):
sounds ok to me

secustor (approved):
User will notice especially if they use nearly for ever cached hashes like Terraform Provider hashes, but this sounds for fine.

PR #43938, opened 2026-06-11
feat: add regex support for gitIgnoredAuthors configuration

gauravluckme:
## Changes

Adds regex support to `gitIgnoredAuthors` so Renovate can ignore commit author emails by pattern instead of exact match only. This is useful for GitLab Project Access Token bot accounts, where each token produces a new `project_{id}_bot_{hash}@noreply.abc.com` address and exact-match config becomes brittle.

## Context

Ref - #43930

## AI assistance disclosure

The public repository: <URL>


lib/util/git/index.ts
```diff
@@ -874,16 +874,15 @@ export async function isBranchModified(
...
-
-  for (const ignoredAuthor of ignoredAuthors) {
-    includedAuthors.delete(ignoredAuthor);
-  }
+  const includedAuthors = new Set(
+    [...committedAuthors].filter(
+      (committedAuthor) =>
+        committedAuthor !== gitAuthorEmail &&
+        !ignoredAuthors.some((ignoredAuthor) =>
+          matchesIgnoredAuthor(committedAuthor, ignoredAuthor),
```
viceice: I would prefer to simply use the match regex or glob helper

lib/util/git/index.ts
```diff
@@ -874,16 +874,13 @@ export async function isBranchModified(
...
-  for (const ignoredAuthor of ignoredAuthors) {
-    includedAuthors.delete(ignoredAuthor);
-  }
+  const includedAuthors = new Set(
+    [...committedAuthors].filter(
+      (committedAuthor) =>
+        committedAuthor !== gitAuthorEmail &&
+        !matchRegexOrGlobList(committedAuthor, ignoredAuthors),
+    ),
+  );
```
viceice: please use a for-of loop with an if condition, that would be much easier to read.

RahulGautamSingh (changes requested):
The switch from exact `Set` equality to `matchRegexOrGlobList` means every non-regex string is now interpreted as a `minimatch `glob and not an exact string. 

The breaking case is emails containing `[...]`

Example:

29139614+renovate[bot]@users.noreply.github.com 

`[bot]` is parsed as a glob character class (match one of b/o/t), so it no longer matches the literal `[bot]`. 

This can be breaking for some users.

lib/util/git/index.ts
```diff
@@ -875,15 +875,32 @@ export async function isBranchModified(
   }
   const { gitAuthorEmail, ignoredAuthors } = config;
 
-  const includedAuthors = new Set(committedAuthors);
+  const isIgnoredAuthor = (committedAuthor: string): boolean => {
```
RahulGautamSingh: convert to function declaration

lib/util/git/index.ts
```diff
@@ -875,15 +875,32 @@ export async function isBranchModified(
...
+    for (const ignoredAuthor of ignoredAuthors) {
+      const ignoredAuthorRegex = getRegexPredicate(ignoredAuthor);
```
RahulGautamSingh: when a user writes a malformed pattern like `/(invalid/`, `getRegexPredicate` returns `null`, so it falls through to exact-string comparison, which never matches. 

The author's ignore rule then silently does nothing. Add a debug/warn log so a bad pattern is visible.

RahulGautamSingh (changes requested):

RahulGautamSingh (commented):
lgtm

RahulGautamSingh (commented):
could you run some real tests, I will approve after that.

this one touches a lot of users so being cautious.

gauravluckme:
> could you run some real tests, I will approve after that.
> 
> this one touches a lot of users so being cautious.

I ran full test suite locally and `gitIgnoredAuthors` implementation passed all critical tests. Please find build & test validation result:

- Full `index.spec.ts`: 119/119 tests passed
- All `gitIgnoredAuthors` scenarios (exact email, regex, case-insensitive, literal brackets, malformed regex warning)
- Biome lint/format: passed
- String-match helpers: 25/25 passed

I also tested it locally by creating the image and running renovate with different scenarios after the run completed everything works as expected.

RahulGautamSingh (approved):

gauravluckme:
@viceice Can you please review this Merge Request!

lib/config/options/index.ts
```diff
@@ -1273,7 +1273,7 @@ const options: Readonly<RenovateOptions>[] = [
   {
     name: 'gitIgnoredAuthors',
     description:
-      'Git authors which are ignored by Renovate. Must conform to [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322).',
+      'Git author emails ignored by Renovate. Entries may be exact [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322) strings or regex patterns using Renovate regex syntax. Plain strings are exact matches (glob patterns are not supported).',
```
viceice: @secustor @jamietanna wonder if we should simply support glob or regex like we do for other fields. any chars which could be false glob things?

secustor: Yes, makes sense to simply reuse our common string pattern matching

docs/usage/configuration-options.md
```diff
@@ -1964,21 +1964,37 @@ Development Bot <dev-bot@my-software-company.com>
...
-This field accepts [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322)-compliant strings.
+Specify commit author emails ignored by Renovate.
+This field accepts exact [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322)-compliant email strings, glob patterns, or regex patterns using Renovate regex syntax.
 
 By default, Renovate will treat any PR as modified if another Git author has added to the branch.
 When a PR is considered modified, Renovate won't perform any further commits such as if it's conflicted or needs a version update.
 If you have other bots which commit on top of Renovate PRs, and don't want Renovate to treat these PRs as modified, then add the other Git author(s) to `gitIgnoredAuthors`.
 
-Example:
+Exact string example:
```
secustor: Rather link to our doc page for string pattern matching https://docs.renovatebot.com/string-pattern-matching/

See [allowedEnv](https://docs.renovatebot.com/self-hosted-configuration/#allowedenv) as example

lib/util/string-match.ts
```diff
@@ -8,12 +8,40 @@ export function isDockerDigest(input: string): boolean {
...
     return regExPredicate;
   }
 
+  // If the pattern doesn't contain glob wildcards, do exact string matching.
+  // This prevents emails with `[bot]` from being interpreted as glob character classes.
+  if (!isGlobPattern(pattern)) {
+    const isPositive = !pattern.startsWith('!');
+    const exactPattern = isPositive ? pattern : pattern.slice(1);
+    return (x: string): boolean => {
+      const matches = x.toLowerCase() === exactPattern.toLowerCase();
```
secustor: This will change the behaviour as we would no longer be case sensitive

secustor (changes requested):

lib/util/string-match.ts
```diff
@@ -8,13 +8,38 @@ export function isDockerDigest(input: string): boolean {
...
+ * (like GitHub bot emails `foo+renovate[bot]@example.com`) contain literal
+ * brackets that must not be interpreted as glob character classes.
+ */
+function isGlobPattern(pattern: string): boolean {
+  // Remove the '!' prefix if present (used for negation)
+  const cleanPattern = pattern.startsWith('!') ? pattern.slice(1) : pattern;
+  // Treat `*`, `?` (wildcards), `{` (brace expansion) and `\` (escape) as glob
+  // indicators. `[` is intentionally omitted so literal brackets in emails match
+  // exactly.
+  return /[*?{\\]/.test(cleanPattern);
```
secustor:
```suggestion
  return regEx(/[*?{\\]/.test(cleanPattern));
```
Use the `regEx` utility

lib/util/string-match.ts
```diff
@@ -8,13 +8,38 @@ export function isDockerDigest(input: string): boolean {
...
 
-  const mm = minimatch(pattern, { dot: true, nocase: true });
+  // Non-glob patterns may contain literal square brackets, e.g. GitHub bot emails
+  // like `foo+renovate[bot]@example.com`. Escape those brackets (and the backslash
+  // escape character itself) so minimatch treats them as literal characters instead
+  // of glob syntax. Case-sensitivity and all other matching behavior are left to
+  // minimatch, unchanged.
+  const globPattern = isGlobPattern(pattern)
+    ? pattern
+    : pattern.replace(/[\\[\]]/g, '\\$&');
```
secustor:
```suggestion
    : pattern.replace(regEx(/[\\[\]]/g), '\\$&');
```

lib/util/git/index.ts
```diff
@@ -875,15 +875,18 @@ export async function isBranchModified(
   }
   const { gitAuthorEmail, ignoredAuthors } = config;
 
-  const includedAuthors = new Set(committedAuthors);
-
-  // v8 ignore else -- TODO: add test #40625
-  if (gitAuthorEmail) {
-    includedAuthors.delete(gitAuthorEmail);
+  function isIgnoredAuthor(committedAuthor: string): boolean {
+    return matchRegexOrGlobList(committedAuthor, ignoredAuthors);
```
secustor: I don't really like the change of the core matching engine. 

@viceice @jamietanna WDYT about simply replacing here `[bot]@` with `@` similar we do in the email address schema? 
https://github.com/renovatebot/renovate/blob/969d1a1c49ff273218fe10a094bbd9bbf9a3ae6b/lib/util/schema-utils/index.ts#L448

jamietanna: In this case wouldn't it be replacing `project_{id}_bot_{hash}` with something else?

secustor: In that case it would be noop as there is no `[bot]@` in the pattern string or do you mean something else? 🤔

jamietanna: I guess trying to work out what you mean here - the use-case in this PR is for GitLab which has a different format than the GitHub `[bot]@`

secustor: The request is about Gitlab, though the feature we are changing is across all platforms. 

And [this workaround](https://github.com/renovatebot/renovate/pull/43938/changes#diff-d75125b02524c506d781e8d39dae138a1eea1ea71db4e54d0ccb4f25fbdc3f8cR16) is only necessary because of Github email addresses 

 ```suggestion
     return matchRegexOrGlobList(committedAuthor.replace('[bot]@', '@'), ignoredAuthors.replace('[bot]@', '@'));
```

**Review by secustor — DISMISSED** [review]
LGTM, but I would like to have a second pair of eyes on this

File: lib/util/git/index.spec.ts
Comment (RahulGautamSingh) [inline lib/util/git/index.spec.ts:1]: - should return false when ignored author contains literal brackets
- should return false when author matches exact email with brackets
these 2 tests are identical remove one

File: lib/util/git/index.ts [hunk 11]
```diff
@@ -998,15 +998,25 @@ export async function isBranchModified(
...
-  for (const ignoredAuthor of ignoredAuthors) {
-    includedAuthors.delete(ignoredAuthor);
+  const includedAuthors = new Set<string>();
+  for (const committedAuthor of committedAuthors) {
+    if (
+      committedAuthor !== gitAuthorEmail &&
+      !isIgnoredAuthor(committedAuthor)
+    ) {
+      includedAuthors.add(committedAuthor);
+    }
```
RahulGautamSingh, line 1010:
```ts
const normalizedIgnoredAuthors = ignoredAuthors.map((a) => a.replace('[bot]@', '@'));
const includedAuthors = new Set<string>();
for (const committedAuthor of committedAuthors) {
  const normalized = committedAuthor.replace('[bot]@', '@');
  if (committedAuthor !== gitAuthorEmail && !matchRegexOrGlobList(normalized, normalizedIgnoredAuthors)) {
    includedAuthors.add(committedAuthor);
  }
}
```
the `.map()` runs inside `isIgnoredAuthor`, once per committed `author. ignoredAuthors` doesn't change; hoist the mapped array above the loop and get away with the `isIgnoreAuthor` helper

RahulGautamSingh (changes requested):

secustor (dismissed):

lib/util/git/index.ts
```diff
@@ -999,15 +999,22 @@ export async function isBranchModified(
...
-
-  // v8 ignore else -- TODO: add test #40625
-  if (gitAuthorEmail) {
-    includedAuthors.delete(gitAuthorEmail);
-  }
+  // Normalize invalid GitHub app/bot email addresses (e.g.
+  // `renovate[bot]@users.noreply.github.com`) the same way `parseGitAuthor`
+  // does, so their literal brackets aren't treated as glob character classes.
+  const normalizedIgnoredAuthors = ignoredAuthors.map((author) =>
+    author.replace('[bot]@', '@'),
```
RahulGautamSingh: 1. Normalizing both the pattern and the input breaks correctly escaped regexes: `/renovate\[bot\]@users\.noreply\.github\.com/` can never match because the input has `[bot]@` stripped while the escaped pattern is left untouched. 

2. It also widens plain entries, `renovate@users.noreply.github.com` now matches the author `renovate[bot]@users.noreply.github.com`. 

Drop the normalization and instead check `ignoredAuthors.includes(committedAuthor)` first, then fall back to `matchRegexOrGlobList` on the raw strings. i.e.

```ts
if (
  committedAuthor !== gitAuthorEmail &&
  !ignoredAuthors.includes(committedAuthor) &&
  !matchRegexOrGlobList(committedAuthor, ignoredAuthors)
) {
  includedAuthors.add(committedAuthor);
}
```

The exact check keeps bracket bot emails working without rewriting either side.

lib/util/git/index.spec.ts
RahulGautamSingh, line 1: Add a test for a regex with escaped brackets, e.g. `/renovate\[bot\]@users\.noreply\.github\.com$/`, against the brackets branch.

docs/usage/configuration-options.md
```diff
@@ -2002,21 +2002,21 @@ Development Bot <dev-bot@my-software-company.com>
...
-Example:
-
 ```json
 {
   "gitIgnoredAuthors": ["some-bot@example.org"]
 }
 ```
 
+`gitIgnoredAuthors` values can be exact [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322)-compliant email strings, glob patterns, or regex patterns.
+For more details on the syntax and supported patterns, see Renovate's [String Pattern Matching documentation](./string-pattern-matching.md).
```
```suggestion
For more details on the syntax and supported patterns, see Renovate's [string pattern matching documentation](./string-pattern-matching.md).
```

for consistency with other places

RahulGautamSingh (changes requested):

viceice (commented):
LGTM beside the open conversation

viceice (approved):

secustor (approved):

PR #43970, opened 2026-06-12
fix: omit version from pr title if group has a single update

felipecrs:
## Changes

This addresses one specific shortcoming of `groupSingleUpdates` (https://github.com/renovatebot/renovate/pull/43629) as reported in https://github.com/renovatebot/renovate/discussions/43758#discussioncomment-17264388.

Rationale:

1. Renovate's adds the version to the PR title whenever the version is the same amongst all updates.
2. Imagine a situation where someone groups multiple updates together, like:

   ```json
   {
     "packageRules": [
       {
         "matchCategories": ["docker"],
         "groupName": "all docker dependencies"
       }
     ]
   }
   ```

3. Before `groupSingleUpdates`, if there was a single update, the PR title would be `Update alpine to 3.15.0`, which is ok.

4. Without this fix (current behavior), if there was a single update, the PR title would be `Update docker dependencies to 3.15.0`, which is not ideal.

5. After this fix, it becomes `Update docker dependencies`.

## Context

## AI assistance disclosure

The public repository:

- https://github.com/felipecrs/renovate-repro-group-single-updates/pull/6
  
  image

- https://github.com/felipecrs/renovate-repro-group-single-updates/pull/7

  image


RahulGautamSingh (commented):
>Before groupSingleUpdates, if there was a single update, the PR title would be Update alpine to 3.15.0, which is ok.

Do you mean `groupSingleUpdates` is affecting normal grouping behavior where there is only one dep in the group.?

It should not right. Since, it's a separate config option.

felipecrs:
> Do you mean groupSingleUpdates is affecting normal grouping behavior where there is only one dep in the group.?

If there is only one dependency in the group that means it's a single update, and that's what `groupSingleUpdates` is for. So, yeah. 😅

Or perhaps I'm missing something and in that case I'd appreciate if you can elaborate, @RahulGautamSingh.

lib/workers/repository/updates/generate.ts
```diff
@@ -312,6 +312,13 @@ export function generateBranchConfig(
       // Now overwrite original config with group config
       upgrade = mergeChildConfig(upgrade, upgrade.group);
       upgrade.isGroup = true;
+      if (!groupEligible) {
+        // Group settings here are only applied because of groupSingleUpdates.
+        // In this case, the group name is used instead of the dependency name,
+        // so the single update's version must not be appended to it.
+        delete upgrade.commitMessageExtra;
```
viceice, line 319: what if multiple updates with same version? is the to version shown in that case?

eg: monorepo updated to vXXX

felipecrs: Yes, that remains unchanged:

- https://github.com/felipecrs/renovate-repro-group-single-updates/pull/9
- https://github.com/felipecrs/renovate-repro-group-single-updates/pull/8

viceice (approved):

RahulGautamSingh (approved):

secustor (approved):

Chrimle:
Hi, this appears to have broken custom `commitMessageExtra` when not using grouped updates...

famod:
It broke more than that it seems: #44169

felipecrs:
Ok, I can confirm the issues. Added a reproduction here:

- https://github.com/felipecrs/renovate-repro-group-single-updates/commit/560f6b79f62a69b71d7bf52bd8c7dd90f23e9485

I'm investigating another way to fix the same.

felipecrs:
> Hi, this appears to have broken custom `commitMessageExtra` when not using grouped updates...

That's not really the case because it turns out Maven property-based updates are actually treated as a group internally by Renovate. Fix:

- https://github.com/renovatebot/renovate/pull/44176

It's true that the custom `commitMessageExtra` is lost in some cases with grouped updates, but that behavior was pre-existing to this PR. It just unintentionally extended that behavior to Maven property updates, unintentionally.

PR #43979, opened 2026-06-13
feat(git): add experimental gitShallowCloneDepth option

RahulGautamSingh:
## Changes

Adds an experimental `gitShallowCloneDepth` global config option that makes Renovate perform a shallow clone (`git clone --depth=N`) instead of the default blobless clone (`--filter=blob:none`).

- New `gitShallowCloneDepth` global, experimental option (`null` by default).
- When set to a positive integer, clones with `--depth=N --no-single-branch` instead of `--filter=blob:none`. `--no-single-branch` is required because `--depth` is single-branch by default, which would otherwise hide existing `renovate/*` branches and break branch management.
- Ignored (with a warning) when the platform requires a full clone.
- Eagerly unshallows the repository on a subsequent fetch (`git fetch --unshallow`) so long-lived clones regain full history.
- Reports `cloned` and `unshallowed` flags, plus the configured depth, in `GitOperationStats`.
- Documents on the option that the depth must exceed the number of commits the base branch can gain between runs; otherwise existing branches can be treated as modified and stop updating.

## Context

- [x] This closes an existing Issue, Closes: #39667

## AI assistance disclosure

AI (Claude Opus 4.8, via Claude Code) was used to implement the shallow-clone branch-visibility fix, write the accompanying unit test, and run the reproduction tests against a real repository.

The public repository: https://github.com/RahulGautamSingh/renovate-shallow-clone-test


RahulGautamSingh:
_AI Summary_

## `gitShallowCloneDepth` — test results

Tested from source with `LOG_LEVEL=debug`, depth `N=2`, against a live repo
(`RahulGautamSingh/renovate-shallow-clone-test`). "Before" = `--depth` only;
"After" = `--depth --no-single-branch` (this PR's fix).

### Q1 — clone time: blobless vs shallow (renovatebot/renovate, deep history)

| Strategy | Wall time | `.git` size | Local commits |
|---|---|---|---|
| blobless `--filter=blob:none` (current default) | **21.7s** | 44 MB | 25,967 |
| shallow `--depth=1` | **5.2s** | 9 MB | 1 |
| shallow `--depth=10` | **5.2s** | 8 MB | 10 |
| shallow `--depth=100` | **11.4s** | 9 MB | 100 |

→ Shallow is **~4× faster and ~5× smaller** on a deep repo. (Single sample, network-bound.)

### Scenario matrix (depth = 2)

**Before fix means using only the shallow flag without specifying depth

| # | Scenario | Mode | Before fix | After fix |
|---|---|---|---|---|
| 1 | First run (no existing branches) | ephemeral | ✅ PRs created | ✅ PRs created |
| 2 | Open PR, **0 commits** to HEAD, next run | ephemeral | ❌ `bad object`, `Error updating branch` | ✅ `isModified=false`, `isConflicted=false`, 0 errors |
| 3 | Open PR + one new PR to raise | ephemeral | ⚠️ new PR raised, existing PR errors | ✅ both handled |
| 4 | Update needed, **enough** history (≤ D−1 behind) | ephemeral | ❌ branch missing from clone | ✅ handled, "No file changes detected" |
| 5 | Update needed, **not enough** history (≥ D behind, e.g. D+1) | ephemeral | ❌ branch missing from clone | ⚠️ branch visible, 0 hard errors, but `isModified()=true` → PR frozen/stale |
| 6 | Subsequent run with `persistRepoData`, base +D+1 | persistent | ❌ `[rejected] (stale info)`, repo-changed error | ✅ `unshallowed=true`, `isModified=false`, `isConflicted=false`, 0 errors |

**Root cause:** `git clone --depth` implies `--single-branch`, so the clone fetched only the base branch and all `renovate/*` branches were absent. Adding `--no-single-branch` restores the full refspec while staying shallow.

### Answers to the key questions

**1. From a clean clone — how long does blobless vs shallow take?**
Blobless ≈ **21.7s / 44 MB** (full commit history). Shallow ≈ **5.2s / 8–9 MB** for depth ≤ 10, rising to ~11.4s at depth 100. Shallow is roughly **4× faster** because it skips downloading the full commit history.

**2. When a branch is created, and no commits to HEAD, does Renovate handle it on the next run?**
**Yes** (with this PR's `--no-single-branch` fix). The base hasn't moved, so the branch's merge-base is inside the shallow window; `isBranchModified`/`isBranchConflicted` both resolve correctly and the PR is managed normally.
*(Without the fix: no — the branch wasn't in the clone at all and every check threw `bad object`.)*

**3. When a branch is created, and `${gitShallowCloneDepth + 1}` commits to HEAD, does Renovate handle it on the next run?**
**Depends on mode:**
- **Ephemeral (fresh clone each run):** **No.** The branch is now visible, but its merge-base with the advanced base is beyond the depth window, so `isModified()` falsely returns `true` → the PR is treated as edited and silently stops updating. No crash. Mitigation: set depth larger than the max commits the base can gain between runs.
- **Persistent (`persistRepoData=true`):** **Yes.** The next run takes the fetch path and `--unshallow`s, restoring full history, so the D+1 branch is handled correctly.

jamietanna:
To confirm re the timings, you should be able to use OTel + Git operations statistics debug log line to see if there's anything interesting in there

lib/config/options/index.ts
```diff
@@ -271,6 +271,16 @@ const options: Readonly<RenovateOptions>[] = [
     inheritConfigSupport: true,
     cli: false,
   },
+  {
+    name: 'gitShallowCloneDepth',
+    description:
+      'If set to a positive integer, Renovate performs a shallow clone with the given depth instead of a blobless clone. The depth must exceed the number of commits the base branch can gain between Renovate runs, otherwise existing branches may be treated as modified and stop updating.',
```
jamietanna, line 277: AI generated? Might be a bit long for the usual description length

lib/config/options/index.ts
```diff
@@ -271,6 +271,16 @@ const options: Readonly<RenovateOptions>[] = [
...
+    type: 'integer',
+    default: null,
+    globalOnly: true,
+    experimental: true,
+    experimentalIssues: [39667],
```
jamietanna, line 282: Should this maybe be a new "Feedback Requested" Discussion? (I can create)

lib/util/stats.ts
```diff
@@ -758,12 +758,34 @@ export class GitOperationStats {
...
+      report[operationType] = timingReport;
+    }
+
+    report.cloned = memCache.get<boolean>('git-operations-cloned') ?? false;
+    report.unshallowed =
+      memCache.get<boolean>('git-operations-unshallowed') ?? false;
+
+    const shallowDepth = memCache.get<number>('git-operations-shallow-depth');
+    if (shallowDepth !== undefined) {
+      report.gitShallowCloneDepth = shallowDepth;
```
jamietanna, line 788: Does it make sense to combine these? We shouldn't ever need to unshallow unless we're running with a depth?

lib/util/git/index.ts
```diff
@@ -474,6 +474,16 @@ export const syncGit = withInstrumenting(
...
+          // Check if repo is shallow and unshallow if needed
+          const isShallow = (
+            await git.raw(['rev-parse', '--is-shallow-repository'])
+          ).trim();
+          if (isShallow === 'true') {
+            logger.debug('Repository is shallow, performing unshallow fetch');
+            await gitRetry(() => git.fetch(['--unshallow', 'origin']));
+            GitOperationStats.setUnshallowed();
+            logger.debug('Unshallow fetch completed');
+          }
```
jamietanna, line 486: I'd prefer we not unshallow unless we need to - this reads as if we're basically always unshallowing?

Given some weighty repos, this means that maybe the first clone is lightweight, but then any operations will require an unshallow?

jamietanna, line 486: ⚠️ can we please introduce a new `git.unshallow` function, which can then be its own `GitOperationType` for tracking in the `Git operations statistics` report?

lib/util/git/index.spec.ts
```diff
@@ -1619,6 +1619,110 @@ describe('util/git/index', { timeout: 30000 }, () => {
...
+      // Second run: fetch existing repo
+      await git.initRepo({ url: origin.path });
+      await git.syncGit();
+
+      const unshallowCalls = rawSpy.mock.calls.filter(
+        (call) =>
+          Array.isArray(call[0]) &&
+          (call[0] as string[]).includes('--unshallow'),
+      );
+      expect(unshallowCalls).toHaveLength(0);
```
jamietanna, line 1723: Could we instead spy on `git` to see if it calls `git.unshallow()`?

jamietanna:
For a "real world" test, I'd recommend using a different repo to really exercise this:

```
% gh api /repos/backstage/backstage --jq .size
5727783KB
   5593MB
      5GB
% gh api /repos/elastic/kibana --jq .size
15785145KB
   15415MB
      15GB
% gh api /repos/renovatebot/renovate --jq .size
  152255KB
     148MB
```

jamietanna (changes requested):

lib/util/git/index.ts
jamietanna: (We probably actually want to double-count it for the `fetch` and the `unshallow` types)

RahulGautamSingh: Removed the eager unshallow from `syncGit`. 

We first check and then unshallow now, for this I added `ensureBranchHistory(branch, base)` in git/index.ts that runs only when the clone is shallow, probes git merge-base, and unshallows only when the shared history is below the boundary.

RahulGautamSingh: Added `unshallow()` to `InstrumentedSimpleGit` and a new 'unshallow' GitOperationType, so it shows separately from fetch in the stats report.

lib/util/git/index.spec.ts
RahulGautamSingh: done

lib/util/stats.ts
RahulGautamSingh: keeping separate as a persistent clone can be shallow from a prior run's depth while the current run sets none, so they aren't redundant.

lib/config/options/index.ts
RahulGautamSingh: Fixed

lib/util/git/types.ts
```diff
@@ -178,6 +178,11 @@ export type GitOperationType =
    * The `git fetch` sub-command.
    */
   | 'fetch'
+  /**
+   * A `git fetch --unshallow`, tracked separately from `fetch` so the cost of
+   * converting a shallow clone to a full one is visible in the statistics.
```
jamietanna, line 183: Maybe

```suggestion
   * A `git fetch --unshallow`.
   *
   * This is tracked separately to `git fetch` to keep a clear measure of the unshallow cost separate to other operations.
```

jamietanna (commented):
We should also add `unshallowed` into the `INFO: Repository finished` log line

jamietanna:
@RahulGautamSingh let's also look at how this would look if i.e. there's 100 commits onto `main` between Renovate runs (if we're expecting 10) - I guess in that case, we'll need to unshallow, which will be expensive?

(although the self-hosted admin can bump their setting to 100, there may then be a bump to 400, etc, so trying to test what happens)

RahulGautamSingh:
_AI Summary of the tests on real repos_


@jamietanna tested depth=10 with 100 commits landing on `main` between runs.

**It unshallows and handles it correctly.** `ensureBranchHistory()` unshallows lazily, once per run, memoised: `isModified=false`, `isConflicted=false`, 0 warnings, 0 errors. Same in ephemeral and `persistRepoData` mode.

**And yes, it's expensive** — the unshallow refetches full history *with* blobs:

| Repo | blobless (default) | shallow `depth=10` | + `fetch --unshallow` |
| --- | --- | --- | --- |
| backstage (5 GB) | **192s / 669 MB** | 292s / 884 MB | **+1662s / 7.3 GB** |
| kibana (15 GB) | **388s / 1.2 GB** | 2952s / 9.3 GB | **+2024s / 16.6 GB** |

In ephemeral mode this is paid on **every run**, since the branch merge-base is always outside a 10-commit window once the base moves. Backstage ~28 min per run, kibana ~34 min. With `persistRepoData` it's one-time — the clone unshallows once and stays full.

**On bumping 10 → 100 → 400:** it doesn't help. A deeper clone makes the clone leg worse without removing the unshallow, because the trigger is the merge-base falling outside the window, not the depth itself. Note the shallow clone is already slower and larger than blobless before any unshallow — `--depth` fetches full trees and blobs at every ref tip, while blobless fetches the whole commit graph with zero blobs.


<details>

<summary> Long Summary </summary>



## Unshallow cost with 100 commits onto `main` (depth = 10)

@jamietanna re your question 
> yes, it unshallows. 


Tested from source with `LOG_LEVEL=debug` against a fresh repo (`RahulGautamSingh/renovate-unshallow-test`), plus raw `git` timings across six repos spanning 148 MB–15 GB and 731–23,592 refs. Single sample each, one machine, network-bound.

### Correctness: handled cleanly

Fresh repo, depth = 10, 100 commits pushed to `main` between runs:

| Run | Mode | Result |
| --- | --- | --- |
| 1 — initial | ephemeral | 2 PRs created, `unshallowed: false` |
| 2 — base +100 | ephemeral | unshallowed once (memoised), `isModified=false`, `isConflicted=false`, **0 warnings, 0 errors** |
| 3 — baseline | blobless | 0 errors |
| persist 1 | `persistRepoData` | unshallowed, repo left non-shallow |
| persist 2 — base +100 | same clone | `cloned: false`, `unshallowed: false`, 0 errors |

`ensureBranchHistory()` unshallows lazily, exactly once per run, and memoises. No `bad object` errors, no frozen PRs. Unit tests pass.

### Cost: blobless (current default) vs `gitShallowCloneDepth: 10`

| Repo | Refs (branches + tags) | blobless | shallow `depth=10` | + `fetch --unshallow` | shallow total |
| --- | --- | --- | --- | --- | --- |
| renovate (148 MB) | 12,815 | **15s / 47 MB** | 39s / 169 MB | +75s / 303 MB | 114s / 303 MB |
| git (319 MB) | 2,028 | **32s / 133 MB** | 65s / 181 MB | +73s / 334 MB | 138s / 334 MB |
| vim (240 MB) | 23,592 | **40s / 60 MB** | 80s / 233 MB | +63s / 431 MB | 143s / 431 MB |
| postgres (789 MB) | 731 | **53s / 187 MB** | 129s / 394 MB | +218s / 822 MB | 347s / 822 MB |
| backstage (5 GB) | 2,226 | **192s / 669 MB** | 292s / 884 MB | +1662s / 7.3 GB | 1954s / 7.3 GB |
| kibana (15 GB) | 2,541 | **388s / 1.2 GB** | 2952s / 9.3 GB | +2024s / 16.6 GB | 4976s / 16.6 GB |

The shallow clone is slower and larger than blobless on every repo tested, **before** any unshallow. On kibana, `--depth=10` takes 49 minutes and 9.3 GB to fetch 17,493 commits; blobless takes 6.5 minutes 6,119.

### Why

It isn't shallow negotiation  it's what each strategy downloads"

- **blobless** fetches the full commit graph with **zero blily on checkout.
- **shallow + `--no-single-branch`** fetches complete trees *and blobs* for N commits at **every ref tip — branches and tags**.

vim makes this concrete: 1 branch but 23,591 tags, so its `--depth=10` clone holds 23,951 commits and 233 MB. Ref count, not repo size, drives the cost, and I found no repo shape where shallow came out ahea

### The speed and the correctness are mutually exclusive

My earlier benchmark (`--depth=1` → 5.2s / 9 MB) measured `--no-single-branch`. That form is genuinely fast, but it's the bug we fixed:

| Repo | `--depth=10` alone (naive) | `--depth=10 --no-single-branch` (this PR) | blobless |
| --- | --- | --- | --- |
| renovate | **7s / 8 MB / 6 refs** | 39s / 169 MB / 286 refs | 15s / 47 MB |
| backstage | **98s / 210 MB / 3 refs** | 292s / 884 MB / 8

6 refs instead of 286, 3 instead of 810 — every `renovate/*oduced the `bad object` / `Error updating branch` failures.Restoring them costs 5.6× on renovate and 3× on backstage.

### The 100-commit case specifically

In ephemeral mode an active repo unshallows on **every run**, since a branch's merge-base is always outside a 10-commit window. Backstage pays 28 min per run, kibana 34 min. Raising the depth to 100 or 400 without removing the unshallow. With `persistRepoData` it's a one-time cost and the clone unshallows once and stays full.

</details>

PR #44022, opened 2026-06-16
fix(manager/mise): skip non-pinned lockedVersion from lock file

pataar:
## Changes

The mise manager assigned the raw `mise.lock` value as `lockedVersion`. For backends that don't pin to an exact version (e.g. `npm:` tools with `^`/`~`/range specifiers), mise records the requested specifier itself (e.g. `^1.18.4`). That range then flowed into the lookup's exact-version comparison (`semver.eq` via `getTimestamp`), throwing `TypeError: Invalid Version: ^1.18.4` and failing the whole Renovate job.

Now the manager only sets `lockedVersion` when the lock-file value is a concrete version under the dep's resolved versioning. Range-specced tools keep their range `currentValue` and simply get no `lockedVersion` — which is accurate, since mise didn't pin them.

## Context

## AI assistance disclosure

Claude Code (Opus 4.8) wrote the fix and the regression test, reviewed by the author.


pataar:
@jamietanna perhaps it's wise if you'd review this too

pataar:
> Hi there,
> 
> This is intended as a polite, automated _request_ that users avoid `@` mentioning repository maintainers like `@viceice`. Doing so causes annoying mobile notifications and makes it harder to maintain this repository.
> 
> We know it might be common elsewhere but we participate in hundreds of discussions a week and would need to turn off GitHub mobile notifications if we were mentioned in every one.
> 
> As a general rule, we will read and respond to all discussions in this repository, so there is no need to mention us.
> 
> Thanks, the Renovate team

Alright, good message. My apologies

lib/modules/manager/mise/extract.ts
```diff
@@ -100,6 +105,13 @@ export async function extractPackageFile(
   return result;
 }
 
+function isPinnedVersion(dep: PackageDependency, version: string): boolean {
+  const versioning = allVersioning.get(
+    dep.versioning ?? getDefaultVersioning(dep.datasource),
+  );
+  return versioning.isVersion(version);
```
jamietanna, line 114: It might be worth using `isSingleVersion` here - to make sure that it's only something that resolves to a specific version

pataar: Tried it, but that way '=1.18.4' becomes valid. That version would break as well.

jamietanna: And to confirm, "this would break" is Renovate's lookup phase? do you have some logs you can share for this? and/or a minimal repro to confirm

pataar:
```json
{
           "name": "renovate",
           "level": 50,
           "logContext": "8d0bbee1-5c93-477d-91e9-d9e6e1694aa2",
           "repository": "redacted",
           "baseBranch": "develop",
           "currentValue": "^6.0.3",
           "datasource": "npm",
           "packageName": "typescript",
           "followTag": null,
           "lockedVersion": "^6.0.3",
           "packageFile": "mise.toml",
           "pinDigests": false,
           "rollbackPrs": false,
           "updatePinnedDependencies": true,
           "err": {
             "message": "Invalid Version: ^6.0.3",
             "stack": "TypeError: Invalid Version: ^6.0.3\n    at new SemVer (/usr/local/renovate/node_modules/.pnpm/semver@7.8.3/node_modules/semver/classes/semver.js:56:13)\n    at compare (/usr/local/renovate/node_modules/.pnpm/semver@7.8.3/node_modules/semver/functions/compare.js:5:32)\n    at Object.eq [as equals] (/usr/local/renovate/node_modules/.pnpm/semver@7.8.3/node_modules/semver/functions/eq.js:4:29)\n    at file:///usr/local/renovate/lib/workers/repository/process/lookup/index.ts:56:21\n    at Array.find (<anonymous>)\n    at getTimestamp (file:///usr/local/renovate/lib/workers/repository/process/lookup/index.ts:53:35)\n    at lookupUpdates (file:///usr/local/renovate/lib/workers/repository/process/lookup/index.ts:353:45)\n    at file:///usr/local/renovate/lib/workers/repository/process/fetch.ts:81:12\n    at LookupStats.wrap (file:///usr/local/renovate/lib/util/stats.ts:39:20)\n    at file:///usr/local/renovate/lib/workers/repository/process/fetch.ts:141:14\n    at file:///usr/local/renovate/node_modules/.pnpm/p-map@6.0.0/node_modules/p-map/index.js:139:20"
           },
           "msg": "lookupUpdates error"
         },
```
         
This is one of the errors. Mise file is defined as this:

```toml
[tools]
node = "24.8.0"
"npm:typescript" = "^6.0.3"
```

pataar: Lockfile entry:
```toml
[[tools."npm:typescript"]]
version = "^6.0.3"
backend = "npm:typescript"
```

20jasper: > Tried it, but that way '=1.18.4' becomes valid. That version would break as well.

Shouldn't '=1.18.4' be allowed? For example, the cargo backend uses that syntax. Or do you mean it becomes invalid?

pataar: It's valid in both mise.toml and the lockfile, but the `isVersion` should return `false` as it can't be used correctly in the lookup phase

viceice: id highly depends on versioning, I think `=1.2.3` is a single version and allowed in look up when appropriate versioning is used

pataar: Alright, should I move to `isSingleVersion`?

pataar: Moved to `isSingleVersion`

viceice:
please avoid force push too

lib/modules/manager/mise/extract.spec.ts
```diff
@@ -1294,6 +1294,51 @@ describe('modules/manager/mise/extract', () => {
...
+      fs.readLocalFile.mockResolvedValueOnce(pinnedLockFileContent);
+      const content = codeBlock`
+        [tools]
+        "npm:typescript" = "=1.18.4"
+      `;
+      const result = await extractPackageFile(content, 'mise.toml');
+      expect(result?.deps[0]).toMatchObject({
+        depName: 'npm:typescript',
+        currentValue: '=1.18.4',
+        lockedVersion: '=1.18.4',
```
viceice: this looks wrong to me, that's not a valid npm version

lib/modules/manager/mise/extract.ts
```diff
@@ -100,6 +105,13 @@ export async function extractPackageFile(
   return result;
 }
 
+function isSingleVersion(dep: PackageDependency, version: string): boolean {
+  const versioning = allVersioning.get(
+    dep.versioning ?? getDefaultVersioning(dep.datasource),
```
viceice, line 112: are you sure versioning or datasource is already set here? the test case seems to suggest something different and a fallback to semver coerce

lib/modules/manager/mise/extract.spec.ts
pataar: True. In cargo it is though. Updated the tests to reflect this.

Screenshot 2026-06-20 at 17 04 56

lib/modules/manager/mise/extract.ts
pataar: I had to go back to `isVersion`. Cargo has valid `=1.2.3` versions which would be accepted by `isSingleVersion` and eventually break the lookup.

lib/modules/manager/mise/extract.ts
pataar: The datasource is always set by the time we reach this point, it is simply the backend that each tool maps to (npm to npm, cargo to crate, etc.). 
Since mise supports many of these, we resolve the versioning per datasource instead of assuming a single scheme in advance. We set the versioning explicitly only for the few tools that require a non-default one; for all the others, getDefaultVersioning(dep.datasource) takes care of it. 

The only situation where it falls back to semver-coerce is when there is no datasource at all, which corresponds basically to the deps we skip anyway

risu729:
I'm a mise contributor. `^1.18.4` or such semver ranges are not supported. It just works for the npm backend because mise just passes through the version as a pinned version, but I think that's not supported.
@jdx Sorry for the ping. Please correct me if I've misunderstood.

jdx:
correct

viceice:
> I'm a mise contributor. `^1.18.4` or such semver ranges are not supported. It just works for the npm backend because mise just passes through the version as a pinned version, but I think that's not supported. @jdx Sorry for the ping. Please correct me if I've misunderstood.

so this fix is technically not required because the user wrongly used ranges for the `mise.toml` ?

lib/modules/manager/mise/extract.ts
```diff
@@ -85,7 +87,12 @@ export async function extractPackageFile(
           lockFileParsed.data,
           dep.depName!,
         );
-        if (lockedVersion) {
+        /*
+         * mise records the requested specifier (e.g. "^1.18.4" or "=1.18.4") for
+         * backends that don't pin, so only treat concrete versions as a
+         * lockedVersion — anything else crashes the lookup's exact-version comparison.
+         */
+        if (lockedVersion && isConcreteVersion(dep, lockedVersion)) {
```
viceice, line 95: - https://github.com/jdx/mise/pull/10916

This should never be a range and is supported by mise accidentially. So we should log a warning like mise that ranges are not supported in `mise.toml`

risu729:
> so this fix is technically not required because the user wrongly used ranges for the `mise.toml` ?

Yes, mise now warns this unsupported range syntaxes.

viceice:
please add a warning log that this is wrong user config

viceice (commented):
merge conflicts

PR #44037, opened 2026-06-16
feat(datasource/packagist): mark abandoned packages as deprecated

pataar:
## Changes

Parse Packagist's `abandoned` field and surface it via `isDeprecated` /
`deprecationMessage`, matching how the npm datasource handles npm's `deprecated`
field. Each release of an abandoned package gets `isDeprecated: true`, the result
gets a package-level `deprecationMessage`, and when `abandoned` is a string the
message names the suggested replacement.

**For reviewers:** despite the field name, this is unrelated to Renovate's
abandonment-detection heuristic (`abandonmentThreshold`). Per #34127, "abandoned"
in Renovate means *derived from stale timestamps*, whereas "deprecated" means
*officially flagged by the registry*. Composer's `abandoned` is an official
maintainer signal, so it maps to deprecation — exactly like npm's `deprecated`.

Verified against a live `p2` response (`swiftmailer/swiftmailer`): in the
minified `composer/2.0` format the `abandoned` field is only present on the
newest version entry, but `MinifiedArray` propagates it to all older entries, so
every release is correctly flagged.

## Context

## AI assistance disclosure

AI (Claude Code, Opus 4.8) implemented the schema change and tests test-first; all output reviewed by me.

### DIFF
--- a/lib/modules/datasource/packagist/schema.ts
+++ b/lib/modules/datasource/packagist/schema.ts
@@ -49,6 +49,7 @@ export const ComposerRelease = z.object({
   source: z.object({ url: z.string() }).nullable().catch(null),
   time: MaybeTimestamp,
   require: z.object({ php: z.string() }).nullable().catch(null),
+  abandoned: z.union([z.string(), z.boolean()]).optional().catch(undefined),
 });
 export type ComposerRelease = z.infer<typeof ComposerRelease>;
 
@@ -95,6 +96,7 @@ export function extractReleaseResult(
   const releases: Release[] = [];
   let homepage: string | null | undefined;
   let sourceUrl: string | null | undefined;
+  let deprecationMessage: string | undefined;
 
   for (const composerReleasesArray of composerReleasesArrays) {
     for (const composerRelease of composerReleasesArray) {
@@ -109,6 +111,11 @@ export function extractReleaseResult(
 
       if (composerRelease.require?.php) {
         dep.constraints = { php: [composerRelease.require.php] };
+      }
+
+      if (composerRelease.abandoned) {
+        dep.isDeprecated = true;
+        deprecationMessage ??= getAbandonedMessage(composerRelease.abandoned);
       }
 
       releases.push(dep);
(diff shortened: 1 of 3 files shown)


jamietanna (commented):
Could we possibly use real packages that are marked as abandoned in our tests?

Helps clarify that there's an exact case (at least right now!)

I.e. https://packagist.org/packages/scheb/two-factor-bundle (has a replacement recommended) and https://packagist.org/packages/sonata-project/core-bundle where no replacement is present

Having both covered will also increase the coverage here

We could also possibly use this to also inject a `replacement`, but happy to do that as a follow-up

pataar:
> Could we possibly use real packages that are marked as abandoned in our tests?
> 
> Helps clarify that there's an exact case (at least right now!)
> 
> I.e. https://packagist.org/packages/scheb/two-factor-bundle (has a replacement recommended) and https://packagist.org/packages/sonata-project/core-bundle where no replacement is present
> 
> Having both covered will also increase the coverage here
> 
> We could also possibly use this to also inject a `replacement`, but happy to do that as a follow-up

Alright, good feedback. About the replacement pr: the downside is that we don't get a version in the replacement response in Packagist. So I guess we need to take the latest one? What's your vision on that?

lib/modules/datasource/packagist/schema.ts
```diff
@@ -111,6 +113,12 @@ export function extractReleaseResult(
         dep.constraints = { php: [composerRelease.require.php] };
       }
 
+      if (composerRelease.abandoned) {
+        dep.isDeprecated = true;
+        deprecationMessage ??= getAbandonedMessage(composerRelease.abandoned);
+        // TODO: when `abandoned` is a package name, emit replacementName/replacementVersion to open a replacement PR (follow-up)
```
jamietanna: We can either remove this, or:

```suggestion
        // TODO #44060 when `abandoned` is a package name, emit replacementName/replacementVersion to open a replacement PR
```

jamietanna (commented):
Otherwise LGTM - thanks! I've captured https://github.com/renovatebot/renovate/issues/44060 but it may need a bit more work before we can do this

pataar:
> Otherwise LGTM - thanks! I've captured #44060 but it may need a bit more work before we can do this

Thanks for putting it into an issue. If it's clear how to handle it, I'd be happy to implement it.

jamietanna (approved):

PR #44083, opened 2026-06-18
feat(manager/npm): read private registries from pnpm-workspace.yaml

pataar:
## Changes

Read private registries from `pnpm-workspace.yaml` (pnpm v11+). The `registry` and `registries` settings (default + scoped keys like `@my-org`) are now applied to dependency lookups — both to `catalog`/`catalogs`/`overrides` deps in the workspace file and to deps in member `package.json` files. Mirrors the existing `.yarnrc.yml` registry handling.

Notes:

- Auth is unchanged — pnpm keeps reading tokens from `.npmrc`/`hostRules`.
- `namedRegistries` is not read.
- Registry values containing `${...}` are ignored, matching pnpm v11.5.3+.

## Context

## AI assistance disclosure

- [x] Yes — substantive assistance (AI-generated non-trivial portions of code, tests, or documentation). Tool: Claude Code (Claude Opus 4.8), used for code, tests, and docs.

The public repository: N/A


pataar:
Unsure if I missed anything. 

Docs ref: https://pnpm.io/settings#registries

lib/modules/manager/npm/extract/pnpm.ts
```diff
@@ -328,3 +346,37 @@ function pnpmCatalogsToArray({
...
+  if (registries) {
+    for (const scope in registries) {
+      // `default` is the fallback registry, not a scope
+      if (scope === 'default') {
+        continue;
+      }
+      // pnpm scope keys keep the leading `@`, e.g. `@my-org`
+      if (
+        packageName.startsWith(`${scope}/`) &&
+        !hasEnvVar(registries[scope])
```
viceice: please filter registry urls before, so we don't need to check again and again for variables

viceice:
please avoid force push

lib/modules/manager/npm/schema.ts
```diff
@@ -39,6 +39,8 @@ export const PnpmWorkspaceFile = Yaml.pipe(
       minimumReleaseAge: z.number().nullish(),
       minimumReleaseAgeExclude: z.array(z.string()).optional(),
       overrides: z.record(z.string(), z.string()).optional(),
+      registry: z.string().optional(),
+      registries: z.record(z.string(), z.string()).optional(),
```
viceice: should we use our nullish helper? if someone set this to null explicit? 🤔

lib/modules/manager/npm/extract/pnpm.ts
```diff
@@ -328,3 +336,68 @@ function pnpmCatalogsToArray({
 
   return result;
 }
+
+// pnpm ignores registry URLs containing `${...}` env-var interpolation since v11.5.3
+function hasEnvVar(value: string): boolean {
+  return value.includes('${');
+}
+
+function withoutEnvVarRegistries(
```
viceice: I think we should move this directly to the schema with a transform 🤔

pataar:
> please avoid force push

Ah I usually avoid merge commits. I'll keep that in mind.

lib/modules/manager/npm/extract/pnpm.ts
pataar: Done!

lib/modules/manager/npm/schema.ts
pataar: Done as well :)

viceice:
> > please avoid force push
> 
> Ah I usually avoid merge commits. I'll keep that in mind.

just leave as is if current main isn't needed. we'll squash merge with merge queue, so ko worries about PR history.

pataar:
> > > please avoid force push
> > 
> > 
> > Ah I usually avoid merge commits. I'll keep that in mind.
> 
> just leave as is if current main isn't needed. we'll squash merge with merge queue, so ko worries about PR history.

Alright, thanks. I'll do that in future PR's too.

lib/modules/manager/npm/extract/index.ts
```diff
@@ -124,6 +127,28 @@ export async function extractPackageFile(
...
+    ? await readLocalFile(pnpmWorkspaceYamlFileName, 'utf8')
+    : null;
+  if (
+    isString(repoPnpmWorkspaceYaml) &&
+    repoPnpmWorkspaceYaml.trim().length > 0
+  ) {
+    const parsed = await PnpmWorkspaceFile.safeParseAsync(
+      repoPnpmWorkspaceYaml,
+    );
+    if (parsed.success) {
```
viceice, line 143: log a debug message on error

lib/modules/manager/npm/extract/index.ts
```diff
@@ -124,6 +127,28 @@ export async function extractPackageFile(
...
+  const pnpmWorkspaceYamlFileName = await findLocalSiblingOrParent(
+    packageFile,
+    'pnpm-workspace.yaml',
+  );
+  const repoPnpmWorkspaceYaml = pnpmWorkspaceYamlFileName
+    ? await readLocalFile(pnpmWorkspaceYamlFileName, 'utf8')
+    : null;
+  if (
+    isString(repoPnpmWorkspaceYaml) &&
+    repoPnpmWorkspaceYaml.trim().length > 0
```
viceice:
```suggestion
    isNonemptyStringAndNotWhitespace(repoPnpmWorkspaceYaml)
```
simplify

viceice (approved):

PR #44092, opened 2026-06-18
fix(datasource/cpanfile): strip v-prefix from CPAN versions

faktas2:
**Problem**: 
CPAN modules that use dotted-decimal versions are published to MetaCPAN with a
leading `v` (e.g. `v1.1.1`), and the `cpan` datasource returns those versions
verbatim.

When the `cpanfile` manager parses a requirement it already strips a leading `v`
(and any comparison operator) from the value it reads out of the file — see
`lib/modules/manager/cpanfile/parser.ts`, where the string handler does
`value.replace(/^(?:\s*(?:==|>=|>))?\s*v?/, '')`. So `currentValue` is stored
without the `v`.

That asymmetry breaks updates. During `autoReplace`, Renovate re-extracts the
file after substituting the new version and checks that the new dependency's
`currentValue` matches the version it intended to write. The datasource supplies
`v1.1.1`, but re-extraction yields `1.1.1`; the two never match, so the update is
considered unverifiable and is silently dropped. The result is that affected
Perl modules are simply never updated.

---
**Solution**:
The `cpanfile` manager now sets `extractVersion: '^v?(?<version>.+)$'` on every
dependency that uses the `cpan` datasource, so a leading `v` is stripped from
versions returned by the datasource before they are compared against the value
written in the file.

### DIFF
--- a/lib/modules/datasource/cpan/__fixtures__/Foo-Bar.json
+++ b/lib/modules/datasource/cpan/__fixtures__/Foo-Bar.json
@@ -1,0 +1,59 @@
+{
+   "took" : 2,
+   "_shards" : {
+      "total" : 3,
+      "successful" : 3,
+      "failed" : 0
+   },
+   "hits" : {
+      "total" : 2,
+      "hits" : [
+         {
+            "_type" : "file",
+            "_index" : "cpan_v1_01",
+            "_id" : "aaaaaaaaaaaaaaaaaaaaaaaaaaa",
+            "_score" : null,
+            "sort" : [
+               1685577600000
+            ],
+            "_source" : {
+               "deprecated" : false,
+               "module" : [
+                  {
+                     "name" : "Foo::Bar",
+                     "version" : "v1.1.1"
+                  }
+               ],
+               "status" : "latest",
+               "date" : "2023-06-01T00:00:00",
+               "maturity" : "released",
+               "distribution" : "Foo-Bar"
+            }
+         },
+         {
+            "_type" : "file",
+            "_index" : "cpan_v1_01",
+            "_id" : "bbbbbbbbbbbbbbbbbbbbbbbbbbb",
+            "_score" : null,
+            "sort" : [
+               1640995200000
+            ],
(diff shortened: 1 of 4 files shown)


lib/modules/manager/cpanfile/parser.ts
```diff
@@ -101,6 +101,11 @@ const moduleMatch = q
       }
 
       dep.datasource = CpanDatasource.id;
+      // The cpan datasource returns versions verbatim from MetaCPAN, which may
+      // be 'v'-prefixed (e.g. `v1.1.1`), but the parser strips a leading 'v'
+      // from `currentValue`. Strip it from the datasource version too so the
+      // autoReplace round-trip does not compare `v1.1.1` against `1.1.1`.
+      dep.extractVersion = '^v?(?<version>.+)$';
```
viceice: no, should be done in code directly before returning from datasource

faktas2: Let me know what you think. Thanks

RahulGautamSingh (changes requested):
changes are in /datasource but the pr title has /manager? typo?

lib/modules/datasource/cpan/schema.spec.ts
RahulGautamSingh, line 1: only the `v1.1.1` case is covered. add a case where an unprefixed version passes through unchanged

RahulGautamSingh (changes requested):

faktas2:
> changes are in /datasource but the pr title has /manager? typo?

I've updated it.

lib/modules/datasource/cpan/__fixtures__/Foo-Bar.json
RahulGautamSingh, line 1: is the fixture necessary since v prefix can be tested via a smaller codeblock?

viceice: inline fixture

viceice (commented):
see open comments

PR #44174, opened 2026-06-22
fix(manager/mise): install resolver tools during mise lock

risu729:
### Problem

When `binarySource=install` is running in Containerbase, Renovate prepends `install-tool ...` commands generated from `toolConstraints` before artifact update commands. The mise lockfile update path runs `mise trust` and `mise lock`, but previously listed only `mise`, so resolver CLIs that `mise lock` may invoke were not provisioned in Renovate's container/sandbox environment.

This is the issue reported in https://github.com/jdx/mise/discussions/10563: if `node`/`npm` are unavailable, `mise lock` can warn and leave stale or missing lock entries for `npm:` tools, including registry shorthands that resolve to the npm backend.

### Fixed Resolver Tools

This change uses a fixed resolver-tool set for every mise lock run: `mise`, `node`, `npm`, `golang`, and `ruby`. All of these tool names are already available through the Renovate/Containerbase image and dynamic-install layer, so this PR only changes Renovate's `toolConstraints`; it does not require a base-image or Containerbase change.

### Mise Lock Call Path

Renovate runs `mise trust <config>` and then `mise lock` or `mise lock <tools>` during the artifact update. Inside `mise lock`, mise resolves the requested tools from project config, aliases, registry entries, and enabled backends, then asks each resolved backend for remote version information before writing lockfile metadata. Some of those backend version-listing paths call external CLIs.

### Why These Tools Are Needed

- `node`/`npm`: [mise's npm backend](https://mise.en.dev/dev-tools/backends/npm.html) lists remote versions with `npm view <package> versions time --json`. It uses the npm CLI so custom registry configuration is honored, for example `.npmrc`, npm config, or npm-related environment variables. `npm` requires `node`, and Renovate's npm manager also includes `node` in its npm lockfile update `toolConstraints` for this reason.
- `golang`: [mise's go backend](https://mise.en.dev/dev-tools/backends/go.html) first queries `$GOPROXY` over HTTP, matching Go module resolution. If no proxy is configured for querying, for example `GOPROXY=direct`, it falls back to `go list -m -versions`, so `go` can be needed during lock-time version resolution. Unset or empty `GOPROXY` is treated as Go's default, not as direct-only.
- `ruby`: [mise's gem backend](https://mise.en.dev/dev-tools/backends/gem.html) uses `gem sources` from the mise-managed Ruby environment to determine the primary RubyGems-compatible source, then queries that source's HTTP API for versions. That makes Ruby needed for source discovery.

### Why This Is Unconditional

A mise tool name is not enough to reliably derive the backend Renovate must provision for.

Examples:

- A registry shorthand can point at a package-manager backend, so a plain tool name like `prettier` can still require the npm backend's `npm view` path.
- A user can configure `tool_alias` so a local name resolves to a different backend than Renovate would infer from the dependency name alone.
- A user can configure `disable_backends`, which can change which backend is eligible for a tool.

Renovate could theoretically invoke `mise` first to resolve backend choices before deciding `toolConstraints`, but that adds another mise execution path and still needs enough tooling available to evaluate project config. A small fixed resolver-tool set is simpler and matches the current lock-time external CLI usage.

### Other Backends

Other backends like [mise's pipx backend](https://mise.en.dev/dev-tools/backends/pipx.html), [mise's cargo backend](https://mise.en.dev/dev-tools/backends/cargo.html), and [mise's SwiftPM backend](https://mise.en.dev/dev-tools/backends/spm.html) do not require external CLI calls to list versions. Their install paths may require tools such as `cargo` or `swift`, but their lock-time version listing is HTTP/API based.

### Npm Package-Manager Settings

mise supports `aube`, `bun`, and `pnpm` for npm-package installation, but version metadata is still queried with `npm view` today. `bun` is not used for this because its info command requires a `package.json`; modern pnpm has related commands, but mise currently avoids depending on package-manager-specific version-query behavior and standardizes on `npm view`.

### Contributor Context

I'm also a mise contributor, and the resolver-tool set above is based on checking mise's current backend behavior/source. This PR addresses Renovate's provisioning gap; it does not change mise's behavior when a resolver CLI is missing.

## Context

## AI assistance disclosure

Used Cursor (AI assistant) to investigate mise lock behavior, review upstream mise source, implement the change, and draft this PR description.

The public repository: https://github.com/risu729/renovate-mise-npm-lock-repro

Ran `pnpm check --all lib/modules/manager/mise/` locally. Added unit tests covering the dynamic-install command sequence for the mise resolver tools before `mise trust`/`mise lock`, plus unconstrained npm/go/gem lockfile update cases that assert `mise.lock` is returned when `mise lock` changes it.

Also ran this PR branch locally inside the official Renovate Containerbase runtime against the public reproduction repository with `binarySource=install` and `allowedUnsafeExecutions=["mise"]`. The released Renovate behavior is shown in https://github.com/risu729/renovate-mise-npm-lock-repro/pull/2: the PR updates only `mise.toml`, leaving `mise.lock` stale. With this fix, the fresh run created https://github.com/risu729/renovate-mise-npm-lock-repro/pull/4, which correctly updates both `mise.toml` and `mise.lock`; the `jdx/mise-action` workflow passed.

### DIFF
--- a/lib/modules/manager/mise/artifacts.spec.ts
+++ b/lib/modules/manager/mise/artifacts.spec.ts
@@ -202,40 +202,6 @@ describe('modules/manager/mise/artifacts', () => {
     ]);
   });
 
-  it('uses constraints.go for the golang resolver tool', async () => {
-    GlobalConfig.set({ ...adminConfig, binarySource: 'install' });
-    fs.readLocalFile
-      .mockResolvedValueOnce('existing content')
-      .mockResolvedValueOnce('existing content');
-    const execSnapshots = mockExecAll();
-
-    await updateArtifacts({
-      packageFileName: 'mise.toml',
-      updatedDeps: [{ depName: 'node' }],
-      newPackageFileContent: '',
-      config: {
-        constraints: {
-          mise: '2026.6.12',
-          node: '24.16.0',
-          npm: '11.4.2',
-          go: '1.24.4',
-          golang: '1.23.0',
-          ruby: '3.4.3',
-        },
-      },
-    });
-
-    expect(execSnapshots).toMatchObject([
-      { cmd: 'install-tool mise 2026.6.12' },
-      { cmd: 'install-tool node 24.16.0' },
-      { cmd: 'install-tool npm 11.4.2' },
-      { cmd: 'install-tool golang 1.24.4' },
-      { cmd: 'install-tool ruby 3.4.3' },
-      { cmd: trustCmd },
-      { cmd: updateToolCmd },
-    ]);
-  });
-
   it('injects GITHUB_TOKEN when host rule found', async () => {
     fs.readLocalFile
       .mockResolvedValueOnce('existing content')
--- a/lib/modules/manager/mise/artifacts.spec.ts
+++ b/lib/modules/manager/mise/artifacts.spec.ts
@@ -208,7 +208,41 @@ describe('modules/manager/mise/artifacts', () => {
     ]);
   });
 
-  it('installs resolver tools before running mise lock with dynamic installs', async () => {
+  it('looks up latest versions for tool versions before `mise lock`, if not set', async () => {
+    GlobalConfig.set({ ...adminConfig, binarySource: 'install' });
+
+    for (const version of ['1.2.3', '2.3.4', '3.4.5', '4.5.6', '5.6.7']) {
+      datasource.getPkgReleases.mockResolvedValueOnce({
+        releases: [{ version: version }],
+      });
+    }
+
+    fs.readLocalFile
+      .mockResolvedValueOnce('existing content')
+      .mockResolvedValueOnce('existing content');
+    const execSnapshots = mockExecAll();
+
+    await updateArtifacts({
+      packageFileName: 'mise.toml',
+      updatedDeps: [{ depName: 'node' }],
+      newPackageFileContent: '',
+      config: {
+        // no constraints
+      },
+    });
+
+    expect(execSnapshots).toMatchObject([
+      { cmd: 'install-tool mise 1.2.3' },
+      { cmd: 'install-tool node 2.3.4' },
+      { cmd: 'install-tool npm 3.4.5' },
+      { cmd: 'install-tool golang 4.5.6' },
+      { cmd: 'install-tool ruby 5.6.7' },
+      { cmd: trustCmd },
+      { cmd: updateToolCmd },
+    ]);
+  });
+
+  it('uses constraints to specify tool versions before `mise lock`, if set', async () => {
     GlobalConfig.set({ ...adminConfig, binarySource: 'install' });
     fs.readLocalFile
       .mockResolvedValueOnce('existing content')
--- a/lib/modules/manager/mise/artifacts.spec.ts
+++ b/lib/modules/manager/mise/artifacts.spec.ts
@@ -1,15 +1,20 @@
 import upath from 'upath';
+import { mockDeep } from 'vitest-mock-extended';
 import { envMock, mockExecAll, mockExecSequence } from '~test/exec-util.ts';
 import { env, fs, hostRules } from '~test/util.ts';
 import { GlobalConfig } from '../../../config/global.ts';
 import type { RepoGlobalConfig } from '../../../config/types.ts';
 import { TEMPORARY_ERROR } from '../../../constants/error-messages.ts';
 import * as docker from '../../../util/exec/docker/index.ts';
+import * as _datasource from '../../datasource/index.ts';
 import type { UpdateArtifactsConfig } from '../types.ts';
 import { updateArtifacts } from './artifacts.ts';
 import * as lockfile from './lockfile.ts';
 import { updateLockedDependency } from './update-locked.ts';
 
+const datasource = vi.mocked(_datasource);
+
+vi.mock('../../datasource/index.ts', () => mockDeep());
 vi.mock('../../../util/exec/env.ts');
 vi.mock('../../../util/fs/index.ts');
 
@@ -45,6 +50,7 @@ describe('modules/manager/mise/artifacts', () => {
     GlobalConfig.set(adminConfig);
     docker.resetPrefetchedImages();
     hostRules.clear();
+    datasource.getPkgReleases.mockReset();
   });
 
   it('returns null if lock file does not exist', async () => {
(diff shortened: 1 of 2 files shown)


lib/modules/manager/mise/artifacts.ts
```diff
@@ -6,12 +6,37 @@ import { TEMPORARY_ERROR } from '../../../constants/error-messages.ts';
...
+ * @see https://mise.jdx.dev/dev-tools/backends/gem.html
+ */
+function getMiseLockToolConstraints(
+  constraints?: Partial<Record<ConstraintName, string>> | null,
+): ToolConstraint[] {
+  return [
+    { toolName: 'mise', constraint: constraints?.mise },
+    { toolName: 'node', constraint: constraints?.node },
+    { toolName: 'npm', constraint: constraints?.npm },
+    { toolName: 'golang', constraint: constraints?.golang },
```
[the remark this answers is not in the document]
risu729: Good catch, fixed in f27810aa0. The `golang` tool constraint now uses `constraints.go`, matching the existing gomod convention.

jamietanna (changes requested):
Can we please add test coverage to cover the change, please? 

I think we'll also need to validate this against a repo before we merge

lib/modules/manager/mise/artifacts.spec.ts
```diff
@@ -202,6 +202,39 @@ describe('modules/manager/mise/artifacts', () => {
     ]);
   });
 
+  it('installs resolver tools before running mise lock with dynamic installs', async () => {
```
jamietanna: Can we also add a test for what happens if none of the options are there?

risu729: I may be misunderstanding what "options" refers to here, but I added coverage for the no-constraints/default case in 474bc9214.

The added tests run the mise artifact update path with `binarySource=install` and no `constraints`, for `npm:renovate`, a `go:` tool, and a `gem:` tool. They assert that Renovate still provisions the resolver tools before `mise trust`/`mise lock`, runs the targeted `mise lock <dep>`, and returns the changed `mise.lock` artifact when the lockfile content changes.

This follows the existing manager artifact test convention of mocking external CLI execution rather than running the real package manager. The real `mise lock` behavior is covered by the public repro linked in the PR body: the released behavior only changed `mise.toml`, while the fixed branch updates both `mise.toml` and `mise.lock`.

risu729:
Thanks for the review! Sorry for not adding tests. I ran Renovate with this PR locally on a real repo and added that reproduction in the PR body.
I added tests, but I didn't really understand what you meant. I guess the test should not run a real `mise lock`?

lib/modules/manager/mise/artifacts.spec.ts
jamietanna: Yeah so I mainly wanted us to cover what happens when we don't specify `constraints` (what I referred to as options, unhelpfully) and to capture what happens to those - would it run `install-tool` for them?

jamietanna: It looks like yes, we'll _always_ install all of those tools, regardless of whether they're needed or not

lib/modules/manager/mise/artifacts.ts
```diff
@@ -6,12 +6,37 @@ import { TEMPORARY_ERROR } from '../../../constants/error-messages.ts';
...
+  ExtraEnv,
+  ToolConstraint,
+} from '../../../util/exec/types.ts';
 import { readLocalFile } from '../../../util/fs/index.ts';
 import * as hostRules from '../../../util/host-rules.ts';
 import type { UpdateArtifact, UpdateArtifactsResult } from '../types.ts';
 import { getConfigType, getLockFileName } from './lockfile.ts';
 
+/**
+ * Resolver tools that mise may invoke while running `mise lock` (version listing),
```
jamietanna: What does the `(version listing)` here mean?

lib/modules/manager/mise/artifacts.ts
```diff
@@ -6,12 +6,37 @@ import { TEMPORARY_ERROR } from '../../../constants/error-messages.ts';
...
+ * regardless of which tools are being locked.
+ *
```
jamietanna: To make it a bit more explicit:

```suggestion
 * Resolver tools that mise may invoke while running `mise lock` (version listing).
 * 
 * NOTE: this will install all relevant tools, regardless of what the given mise configuration uses.
```

lib/modules/manager/mise/artifacts.ts
risu729: mise resolves the version like `1.2.3` against the remote versions list because mise supports fuzzy versions like `1` or `1.2`.
That's why `mise lock` runs `npm view` or other commands.

lib/modules/manager/mise/artifacts.spec.ts
risu729: Sorry, I still don't fully understand what you mean. We can remove `constraints` in the test, then it should break, but since we mock `mise lock` itself, we cannot test that it breaks.

jamietanna (approved):

secustor (commented):
How much more traffic will this generating by downloading these tools in case they are not needed?  
Why can we not detect that needed backends?

risu729:
> How much more traffic will this generating by downloading these tools in case they are not needed? Why can we not detect that needed backends?

I'm not sure for the traffic, but it's hard to avoid. Even if we don't support `tool_alias` and `disable_backends` at all, we still need to have the mise registry data somewhere.

We can detect the required tools if we run `mise ls`, but I guess we cannot defer tool installations.

### Why This Is Unconditional

A mise tool name is not enough to reliably derive the backend Renovate must provision for.

Examples:

- A registry shorthand can point at a package-manager backend, so a plain tool name like `prettier` can still require the npm backend's `npm view` path.
- A user can configure `tool_alias` so a local name resolves to a different backend than Renovate would infer from the dependency name alone.
- A user can configure `disable_backends`, which can change which backend is eligible for a tool.

Renovate could theoretically invoke `mise` first to resolve backend choices before deciding `toolConstraints`, but that adds another mise execution path and still needs enough tooling available to evaluate project config. A small fixed resolver-tool set is simpler and matches the current lock-time external CLI usage.

secustor (approved):
Let's track the telemetry and see how it goes

jamietanna:
Via https://xcancel.com/jdxcode/status/2080451612674347008 we may be able to remove the need to install `npm`

PR #44185, opened 2026-06-23
fix(presets): add new URL for masstransit

mueller-ma:
## Changes

In later versions of MassTransit the source url has been changed.

See https://www.nuget.org/packages/MassTransit.RabbitMQ/9.1.2

## Context

## AI assistance disclosure

### DIFF
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -541,7 +541,10 @@
     "mantine": "https://github.com/mantinedev/mantine",
     "mapstruct": "https://github.com/mapstruct/mapstruct",
     "marten": "https://github.com/JasperFx/marten",
-    "masstransit": "https://github.com/MassTransit/MassTransit",
+    "masstransit": [
+      "https://github.com/MassTransit/MassTransit",
+      "https://github.com/Massient/MassTransit",
+    ]
     "material-components-web": "https://github.com/material-components/material-components-web",
     "material-ui": [
       "https://github.com/mui-org/material-ui",
--- a/lib/data/monorepo.json
+++ b/lib/data/monorepo.json
@@ -543,7 +543,7 @@
     "marten": "https://github.com/JasperFx/marten",
     "masstransit": [
       "https://github.com/MassTransit/MassTransit",
-      "https://github.com/Massient/MassTransit",
+      "https://github.com/Massient/MassTransit"
     ]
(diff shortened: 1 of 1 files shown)


lib/data/monorepo.json
```diff
@@ -541,7 +541,10 @@
     "mantine": "https://github.com/mantinedev/mantine",
     "mapstruct": "https://github.com/mapstruct/mapstruct",
     "marten": "https://github.com/JasperFx/marten",
-    "masstransit": "https://github.com/MassTransit/MassTransit",
+    "masstransit": [
+      "https://github.com/MassTransit/MassTransit",
+      "https://github.com/Massient/MassTransit"
```
jamietanna, line 546: This is a 404 🤔 Might not be intended

jamietanna: I'd suggest raising it upstream with the MassTransit folks

mueller-ma: The latest version is closed source now:

https://massient.com/#faq 

> MassTransit v9 will not be open source under a permissive license like previous versions, but it will remain source-available. The full source code will continue to be hosted on GitHub, accessible to licensed customers and community users with appropriate permissions. This ensures transparency, auditability, and the ability to review or debug the framework while maintaining a commercial licensing model. The codebase, APIs, and development process will remain open for review, but redistribution or modification outside the license terms will require authorization.

I guess the URL links to the private repo.

viceice: we don't accept closed source links

jamietanna: Do we think it's worth accepting it to at least allow users who are using it to keep it? Or our preference not to allow closed source links?

(For users who _are_ using the closed source package, they can set their own config for this)

mueller-ma: Currently the grouping doesn't work for MassTransit version 8 as well.

viceice: > Do we think it's worth accepting it to at least allow users who are using it to keep it? Or our preference not to allow closed source links?
> 
> (For users who _are_ using the closed source package, they can set their own config for this)

if we allow it once here then we'll need to accept all other closed source links which I don't like.

jamietanna: Very fair, and agreed 👍🏼

PR #44182, opened 2026-06-23
feat(datasource/pypi): support json and html simple pypi api

ahippler:
## Changes

Add support for the JSON-based Simple API for Python Package Indexes (PEP 691). See https://peps.python.org/pep-0691/ for the spec.

This JSON-based Simple API contains the `upload-time`, which is used as `releaseTimestamp`. The change is backwards compatible;
the previous HTML parser is used if the response does not contain the JSON content-type header.

## Context

Third party registries that only support the simple api did not contain the `releaseTimestamp`.

Please select one of the following:

- [X] This closes an existing Issue, Closes: #20070

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [X] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation

- [X] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work

I have verified these changes via:

- [ ] Code inspection only, or
- [X] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

Issue #20070, opened 2023-01-27
Support stabilityDays with Python (PyPI) packages


intgr:
### What would you like Renovate to be able to do?

I'm self-hosting Renovate and I notice that `stabilityDays` does not apply to Python packages.

I also have a local mirror of PyPI, using Poetry with custom `tool.poetry.source` configuration.

It's not explicitly documented which datasources support this and which don't. The public PyPI instance seems to know release dates of packages. Is this something that's just not implemented yet, or could this be an issue due to my PyPI mirror?

The `stabilityDays` documentation only mentions:

> Some datasources do not provide a release timestamp (in which case this feature is not compatible), and other datasources may provide a release timestamp but it's not supported by Renovate (in which case a feature request needs to be implemented).

### If you have any ideas on how this should be implemented, please tell us here.

The public PyPI instance seems to return release dates in its JSON API, although not in a very convenient format: https://pypi.org/pypi/django/4.1.5/json

```json
{
    "info": {
        ...
    },
    "urls": [
        {
            "filename": "Django-4.1.5-py3-none-any.whl",
            "upload_time": "2023-01-02T07:34:45",
            "upload_time_iso_8601": "2023-01-02T07:34:45.541744Z",
            ...
        },
        {
            "filename": "Django-4.1.5.tar.gz",
            "upload_time": "2023-01-02T07:34:49",
            "upload_time_iso_8601": "2023-01-02T07:34:49.562776Z",
            ...
        }
    ]
}
```

### Is this a feature you are interested in implementing yourself?

No

rarkins:
It's likely because locally Renovate is using the "simple" PyPI interface, which does not contain timestamps. Take a look at the http statistics at the end of a repository run to see which URLs of your local pypi server are being accessed.

rarkins:
Does your server provide a non simple API with time stamps?

intgr:
Ah sorry, I forgot to reply. I'm using Artifactory, it doesn't support PyPI-like JSON API yet: https://jfrog.atlassian.net/browse/RTFACT-16101

There is, however, an Artifactory-specific API. Not sure if anyone is interested in impementing that: https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API

For my use case, it would also be fine if Renovate used PyPI directly for package metadata.

----

PS: There's even a comment in that ticket regarding Renovate:

[...]

PR #44222, opened 2026-06-24
feat(pypi): add support for JSON-based Simple API (PEP 691)

felipecrs:
## Changes

Renovate has currently two methods for looking up pypi updates:

1. `https://pypi.org/simple/<package>`, which only supports listing versions
2. `https://pypi.org/pypi/<package>/json`, which supports `releaseTimestamp`, and `sourceUrl` detection

The problem is:

For method 2, Renovate relies on the `releases` key in the JSON response. The `releases` key is **deprecated** ([ref](https://docs.pypi.org/api/json/)), and `https://pypi.org/pypi/<package>/json` may drop it at any time. When it does, this method will stop working entirely.

This is already the case for some private registries like Artifactory. For Artifactory remote repositories, the `releases` key is still there since it proxies `https://pypi.org/pypi/<package>/json`. But for **local repositories**, Artifactory already omits `releases` from the response, making Renovate unable to lookup releases in such case.

The only alternative for now is to use `registryUrl=https://example.com/artifactory/api/pypi/example-pypi-local/simple/` instead, but then Renovate will not extract release timestamps (meaning [`minimumReleaseAge`](https://docs.renovatebot.com/configuration-options/#minimumreleaseage) won't work).

I believe the ultimate solution for this is:

1. Support JSON-based Simple API ([docs](https://packaging.python.org/en/latest/specifications/simple-repository-api/#simple-repository-api)) (introduced as [PEP 691](https://peps.python.org/pep-0691/)) when resolving `/simple/<package>` .
   - **This is what this PR does.**
   - It includes `upload-time` which can be used to detect `releaseTimestamp`.
   - `https://pypi.org/simple/<package>` already supports it.
   - [Artifactory supports it on an opt-in basis](https://jfrog.com/help/r/artifactory-enabling-time-based-package-filtering-for-pypi-repositories/artifactory-enabling-time-based-package-filtering-for-pypi-repositories).

2. Refactor PyPI JSON to extract release metadata from `https://pypi.org/pypi/<package>/json/<package>/<version>` ([docs](https://docs.pypi.org/api/json/#get-a-release))
   - I.e. instead of making a single request to `https://pypi.org/pypi/<package>/json`, first retrieve the available versions from the PyPI Simple (as suggested by their [documentation](https://docs.pypi.org/api/json/#get-a-release)), and then query every individual release through `https://pypi.org/pypi/<package>/json/<package>/<version>`
   - Not as great as before but, well, that's life.

As mentioned above, this PR implements solution 1. If the Simple API returns JSON upon request, we use it and extract `releaseTimestamp`s. If not, we fallback to HTML parsing like before.

Relates to #20070

PS: I started working on this PR before realizing https://github.com/renovatebot/renovate/discussions/41916, of which the author stated that he had a working PR (#44182). Apologies!

## Context

- [ ] This closes an existing Issue, Closes: 

## AI assistance disclosure

Claude Opus 4.6 was used to generate the initial code changes.

The public repository: https://github.com/felipecrs/renovate-repro-pip-pep-691

image

> [!IMPORTANT]
> To verify this against https://pypi.org/simple/, this line needs to be changed in lib/modules/datasource/pypi/index.ts:
>
> ```diff
>     const hostUrl = ensureTrailingSlash(
> -    registryUrl!.replace('https://pypi.org/simple', 'https://pypi.org/pypi'),
> +    registryUrl!,
>    );
> ```
> 
> I believe we should consider removing that from Renovate as a separate PR, as that prevents users from setting their own `registryUrl=https://pypi.org/simple/`. As mentioned above, retrieving releases from https://pypi.org/pypi/ may stop working at any time.


nvoq-lblaney:
I swear @felipecrs, everyone I go on GitHub, I run into you =). You might remember me by my non-work user @shadycuz. Thanks for this fix, this issue has been a headache at work.

felipecrs:
Of course I remember you! From [groovy-guru](https://github.com/DontShaveTheYak/groovy-guru)!

psalaberria002:
Looking forward to this one!

ahippler:
same as https://github.com/renovatebot/renovate/pull/44182. I’ll close my PR so @viceice has fewer to review.

RahulGautamSingh (commented):
Thanks for the improvements!

Code looks good to me and approach 1 is the way to go, we have not implemented approach 2 for other datasources as well, as it needs too many API requests.

felipecrs:
Thanks a lot for the review, @RahulGautamSingh!

But I'll incorporate some changes from @ahippler's #44182 and add him as co-author. Marking as draft for now.

EDIT: Done.

lib/modules/datasource/pypi/index.ts
```diff
@@ -198,11 +201,12 @@ export class PypiDatasource extends Datasource {
...
         // There may be multiple releases with different requires_python, so we return all in an array
-        const pythonConstraints = releases
-          .map(({ requires_python }) => requires_python)
-          .filter(isString);
         result.constraints = {
-          python: Array.from(new Set(pythonConstraints)),
+          python: deduplicateArray(
+            releases
+              .map(({ requires_python }) => requires_python)
+              .filter(isString),
```
RahulGautamSingh: nit: isNonEmptyString

RahulGautamSingh (approved):
lgtm otherwise.

lib/modules/datasource/pypi/index.ts
```diff
@@ -303,26 +291,109 @@ export class PypiDatasource extends Datasource {
...
+  }
+
+  private static getSimpleReleasesFromJson(
+    json: string,
+    packageName: string,
+  ): Releases {
+    const releases: Releases = {};
+    const parsed = Json.pipe(PypiSimpleResponse).safeParse(json);
+    if (!parsed.success) {
+      logger.trace(
```
viceice: shouldn't we at least issue a debug message? otherwise this won't be visible if there's any parser error. 🤔

lib/modules/datasource/pypi/index.ts
```diff
@@ -303,26 +291,109 @@ export class PypiDatasource extends Datasource {
...
+          release.requires_python = requiresPython;
+        }
+        const uploadTime = file['upload-time'];
+        if (uploadTime) {
+          release.upload_time = uploadTime;
         }
-        releases[version].push(release);
+        (releases[version] ??= []).push(release);
       }
     }
```
viceice, line 321: i think this should be done inside a zod schema transformer. 🤔

lib/modules/datasource/pypi/schema.ts
```diff
@@ -25,3 +25,21 @@ export const PypiResponse = DeepNullish(
...
+
+// JSON-based Simple API (PEP 691) schemas
+// https://peps.python.org/pep-0691/
+export const PypiSimpleFile = DeepNullish(
+  z.object({
+    filename: z.string(),
+    'requires-python': z.string().optional(),
+    yanked: z.union([z.boolean(), z.string()]).optional().default(false),
+    // `upload-time` is specified by PEP 700
+    'upload-time': z.string().optional(),
```
viceice:
```suggestion
    'upload-time': Timestamp.optional(),
```

why not using `Timestamp` here directly?

lib/modules/datasource/pypi/index.ts
felipecrs: Totally agree. I made it so we `warn` once. Otherwise, it would be hard to spot why no releases are found.

jamietanna (commented):
A couple of things that Claude Opus 4.8 dug out:

````markdown
## 1. Strict `Timestamp` on `upload-time` silently drops file entries (and possibly whole versions) — CONFIRMED
`lib/modules/datasource/pypi/schema.ts:39`

```ts
'upload-time': Timestamp.optional(),
```

`Timestamp` emits `z.NEVER` on an unparseable value, which fails the whole `PypiSimpleFile.safeParse`. Because `PypiSimpleResponse.files` uses `LooseArray(PypiSimpleFile)` (schema.ts:59), a single bad `upload-time` causes `LooseArray` to **silently discard that entire file entry**. If it was the only file for a version, that version vanishes from `releases` and Renovate will never offer it.

Note the asymmetry: the JSON-API path (`PypiRelease`, schema.ts:9) uses `z.string().optional()` + lenient `asTimestamp` downstream, so a bad timestamp there degrades to a *missing* timestamp rather than a dropped version. This is a robustness regression unique to the new Simple-JSON path. **Fix:** parse `upload-time` as `z.string().optional()` and let the existing `asTimestamp` (used at index.ts:368) handle coercion leniently — matching the JSON-API path.

## 2. Content-type dispatch requires the exact vendor string — PLAUSIBLE (registry-dependent)
`lib/modules/datasource/pypi/index.ts:357`

```ts
const releases = contentType?.includes('application/vnd.pypi.simple.v1+json')
  ? getSimpleReleasesFromJson(...)
  : getSimpleReleasesFromHtml(...);
```

A registry or proxy that returns a PEP 691 JSON body but labels it `application/json` / `text/json` (or omits/rewrites the vendor content-type) gets routed into the HTML parser, which finds no `<a>` links and returns **zero releases** → package reported as having no versions, silently. Consider also detecting a JSON body (e.g. `+json` suffix, or trying JSON parse first) rather than requiring the exact vendor media type.

## 3. New `Accept` header can make lookups throw on strict registries — PLAUSIBLE (registry-dependent)
`lib/modules/datasource/pypi/index.ts:341`

The new `accept` header is now sent on the primary simple-index path. A registry doing strict content negotiation that doesn't honor the `text/html; q=0.01` fallback could respond `406 Not Acceptable`; `this.http.getText` rejects, and the primary simple branch (caller at index.ts:64) isn't wrapped in try/catch, so the whole lookup errors — whereas pre-PR it resolved fine. Worth confirming the caller's error handling, though real-world registries honoring q-weighted Accept makes this lower-probability.

## 4. JSON parse failure masks a malformed response as "no versions" — CONFIRMED (cleanup)
`lib/modules/datasource/pypi/index.ts:306`

```ts
if (!parsed.success) {
  logger.once.warn(...);
  return releases; // {}
}
```

On a JSON content-type with an unparseable/malformed body, this returns `{}`, so the caller produces `releases: []` rather than `null`. Renovate can't distinguish "no data / error" from "genuinely empty," and only a deduplicated warn signals the problem. Consider returning `null` so callers can distinguish the cases (or let the error surface for retry).
````

Mind taking a look please?

**Comment by jacobthompsonramsey** [comment]
Hey I would love this feature. Any chance we can get this out?

**Comment by rmacconn** [comment]
We would love this feature.  Ran into this issue with our teams on python

**Comment by joeyweisband** [comment]
Our team would GREATLY benefit from this! Running into this a lot!

**Comment by jamietanna** [comment]
This is on our radar - but please don't add +1 style comments on the PR

**Comment by jamietanna** [comment]
I've noticed (with some pointers from Claude Opus 5) that there's a bug in our existing releaseTimestamp handling for packages - I'll fix that before we carry on here

File: lib/modules/datasource/pypi/index.ts (hunk 1 above)
Comment (viceice) [reply lib/modules/datasource/pypi/index.ts]: ☝️

File: lib/modules/datasource/pypi/index.ts [hunk 5]
```diff
@@ -290,9 +347,39 @@ export class PypiDatasource extends Datasource {
...
-    const response = await this.http.getText(sanitizedUrl, { headers });
+    const headers: OutgoingHttpHeaders = {
+      ...authHeaders,
+      // Request the JSON serialization (PEP 691), falling back to the legacy
+      // HTML serialization (PEP 503) via content negotiation.
+      // https://github.com/pypa/pip/blob/6b0011b49a068c62f65389bf4cea7af5d28cb002/src/pip/_internal/index/collector.py#L128-L134
+      accept:
+        'application/vnd.pypi.simple.v1+json, application/vnd.pypi.simple.v1+html; q=0.1, text/html; q=0.01',
+    };
+    let response: Awaited<ReturnType<typeof this.http.getText>>;
```
viceice:
```suggestion
    let response: HttpResponse;
```

simplify

viceice (commented):
otherwise LGTM

lib/modules/datasource/pypi/index.ts
jamietanna: So, interestingly, I thought https://github.com/renovatebot/renovate/pull/45312 was covering this change, but it may have been something else

jamietanna:
Claude Opus 5 has a few findings to review

lib/modules/datasource/pypi/index.ts
```diff
@@ -290,9 +329,38 @@ export class PypiDatasource extends Datasource {
...
+      // https://github.com/pypa/pip/blob/6b0011b49a068c62f65389bf4cea7af5d28cb002/src/pip/_internal/index/collector.py#L128-L134
+      accept:
+        'application/vnd.pypi.simple.v1+json, application/vnd.pypi.simple.v1+html; q=0.1, text/html; q=0.01',
+    };
+    let response: HttpResponse;
+    try {
+      response = await this.http.getText(sanitizedUrl, { headers });
+    } catch (err) {
+      // An `abortOnError` host rule makes `Http` wrap the original error, so unwrap it before looking at the status code.
+      const httpErr = err instanceof ExternalHostError ? err.err : err;
```
secustor, line 346: I have in back of my head that we had an unwrap function for this 🤔 , but I'm not able to find it

viceice (approved):

felipecrs:
Thank you so much for taking care of this, @jamietanna!

PR #43696, opened 2026-05-30
refactor(misc): remove `getJsonUnchecked`  usages

secustor:
## Changes

13/13

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: # 
- [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [x] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #44227, opened 2026-06-25
fix(datasource/crate): tolerate null metadata fields

Turbo87:
## Changes

`CrateDatasource` reads `sourceUrl` from the crate's `repository` metadata field. The crates.io API returns `null` (not an omitted key) for unset metadata fields, but the `CrateMetadata` schema used `z.string().optional()`, which accepts `undefined` but rejects `null`. A single null field (e.g. `documentation` on the `sentry` crate) therefore failed the whole metadata parse, so `_getCrateMetadata` returned `null` and the valid `repository` was discarded along with it. The result was a missing `sourceUrl`, which broke monorepo grouping (`matchSourceUrls`) for affected crates.

This switches those fields to `.nullish()` so null values are tolerated and `repository` survives.

This resolves https://github.com/renovatebot/renovate/pull/43696#pullrequestreview-4570688879.

## Context

## AI assistance disclosure

Claude Code was used to investigate the root cause from a job log.

The `processes real data: libc` snapshot test now exercises a `null` metadata field (`homepage: null`) and fails without the schema change.


lib/modules/datasource/crate/__fixtures__/libc.json
```diff
@@ -8,7 +8,7 @@
     "documentation": "https://docs.rs/libc/",
     "downloads": 106751998,
     "exact_match": false,
-    "homepage": "https://github.com/rust-lang/libc",
+    "homepage": null,
```
jamietanna: Can we instead use a fixture for a crate that doesn't have a `homepage`?

lib/modules/datasource/crate/schema.ts
```diff
@@ -18,10 +18,10 @@ export const ReleaseTimestamp = z
   .catch(null);
 
 export const CrateMetadata = z.object({
-  description: z.string().optional(),
-  documentation: z.string().optional(),
-  homepage: z.string().optional(),
-  repository: z.string().optional(),
+  description: z.string().nullish(),
```
viceice: use our nullish schema util instead

secustor: The DeepNullish one to be specific

lib/modules/datasource/crate/__fixtures__/sentry
jamietanna, line 1: I think this is meant to be deleted?

Turbo87: it's used by the `Fixtures.get('sentry')` below

lib/modules/datasource/crate/__fixtures__/libc.json
Turbo87: so apparently adding a new test with a similar snapshot is now triggering a CI failure. I'm not sure what exactly it is you want me to do.

lib/modules/datasource/crate/__fixtures__/sentry
jamietanna: Oh my bad, sorry!

secustor (approved):

lib/modules/datasource/crate/index.spec.ts
```diff
@@ -296,6 +296,24 @@ describe('modules/datasource/crate/index', () => {
...
+      httpMock
+        .scope(CRATES_IO_REGISTRY_URL_PARSED)
+        .get('/se/nt/sentry')
+        .reply(200, Fixtures.get('sentry'));
+      const res = await getPkgReleases({
+        datasource,
+        packageName: 'sentry',
+        registryUrls: [CRATES_IO_REGISTRY_URL],
+      });
+      expect(res).toMatchSnapshot();
```
viceice, line 312: no snapshots

Turbo87:
👋  it looks like the GitHub App is still using v43.242.2 without this fix. how long does it usually take for the GitHub App to update? :)

PR #44228, opened 2026-06-25
feat(gerrit): enable push signing when `gitPrivateKey` is set

felipecrs:
## Changes

Amongst the platforms supported by Renovate, Gerrit is the only one that supports [git push signing](https://git-scm.com/docs/git-push#Documentation/git-push.txt---signed).

In fact, [Gerrit does not support commit signing](https://groups.google.com/g/repo-discuss/c/MkzDyK66Oos). It only supports push signing. It is harmless to sign Gerrit commits though, as Gerrit won't reject them. It's still useful in case some additional repository browser is used, like [GerritHub](https://review.gerrithub.io/) which replicates the commits back to GitHub.

This PR enables push signing for Gerrit whenever `gitPrivateKey` is set to a PGP key (Gerrit does not support SSH-signing). Push signing is enabled as `if-asked`, which means that git will only sign the push if the remote signals support for it.

I decided not to add a new config option because no one else beside Gerrit users would be able to use it anyway, and also we already have too many config options.

Because of `if-asked`, it would have been safe to enable it for all platforms, but I decided not to do it because the `git push` command issues a _warning_ when the remote does not support it. That would be confusing for users of other platforms.

The implementation is done so that it can be easily extended to other platforms in the future, if the other platforms ever add support for it.

## Context

## AI assistance disclosure

Claude Opus 4.6 was used to generate the initial changes.

Gerrit Hub does not have push signing enabled. This was tested on an internal Gerrit instance.


docs/usage/self-hosted-configuration.md
```diff
@@ -770,6 +770,7 @@ Before the first commit in a repository, Renovate will:
 
 1. Run `gpg import` (if you haven't before) when using PGP
 1. Run `git config user.signingkey`, `git config commit.gpgsign true` and `git config gpg.format`
+1. Run `git config push.gpgSign if-asked` when using Gerrit
```
RahulGautamSingh:
```suggestion
1. Run `git config push.gpgSign if-asked` when using PGP with Gerrit
```

RahulGautamSingh (commented):
why is SSH excluded?

felipecrs:
> why is SSH excluded?

Gerrit does not support SSH key for push signing. Users can't register an SSH key for signing purposes, only GPG, and if users attempt to push the sign with a SSH key, Gerrit refuses the push.

I added this to the PR description.

RahulGautamSingh (approved):

lib/util/git/private-key.spec.ts
```diff
@@ -487,4 +489,78 @@ some-private-key
...
+      );
+    });
+
+    it('does not set push.gpgSign for SSH key on gerrit', async () => {
+      const sshKey = `\
+-----BEGIN OPENSSH PRIVATE KEY-----
+some-private-key
+some-private-key
+-----END OPENSSH PRIVATE KEY-----
+`;
```
viceice: can you use `codeBlock`  template helper to indent this?

lib/util/git/private-key.ts
```diff
@@ -91,6 +92,13 @@ class GPGKey extends PrivateKey {
     }
   }
 
+  override async configSigningKey(cwd: string): Promise<void> {
+    await super.configSigningKey(cwd);
+    if (GlobalConfig.get('platform') === 'gerrit') {
+      await exec(`git config push.gpgSign if-asked`, { cwd });
```
viceice, line 98: why not using `simple-git` for this?

viceice (commented):
otherwise lgtm

lib/util/git/private-key.ts
felipecrs: No reason except that's how it's currently done:

https://github.com/renovatebot/renovate/blob/a1971afa78f2971bbf330fbf4db3166d07c572b1/lib/util/git/private-key.ts#L74-L76

I think we should refactor in a separate PR.

felipecrs: Also, it would make this PR a lot more complicated and I'm lacking the time to re-test it.

lib/util/git/private-key.spec.ts
felipecrs: Done in e1a77259ddf910e6f1258e4548fde735232826cf.

Issue #44153, opened 2026-06-22
Create a single config option for `vulnerabilityAlerts`


jamietanna:
And migrate OSV vulnerability alerts under this. For instance:

```json
{
  "vulnerabilityAlerts": {
    "enabled": true,
    "source": "osv"
  }
}
```

PR #44262, opened 2026-06-26
feat(pep621): surface uv.lock transitive deps for osvVulnerabilityAlerts

khvn26:
## Changes

In this PR, we let `osvVulnerabilityAlerts` remediate vulnerabilities in transitive (lockfile-only) dependencies for the `uv` manager.

Today the pep621/uv manager only attaches `lockedVersion` to dependencies declared in `pyproject.toml`. Packages that exist solely in `uv.lock` are parsed and then discarded, so `osvVulnerabilityAlerts` can never see a CVE in an indirect dependency. `lockFileMaintenance` is the only current workaround and it isn't security-scoped.

This surfaces those transitive packages so the existing vulnerability machinery can act on them.

- `UvProcessor.extractLockedVersions` now emits each registry-sourced package that isn't already a declared dependency as a `PackageDependency` with `depType: 'uv.lock'`, its `lockedVersion`, and `skipReason: 'lockfile-only'`. Because they're skipped, they produce no routine updates and no datasource lookups.
- Those transitive deps are only extracted when `osvVulnerabilityAlerts` is enabled, so a large `uv.lock` costs nothing for users who don't use the feature. `ExtractConfig` gained the option and it is threaded to `extractLockedVersions`; `generateFingerprintConfig` includes it so toggling the option invalidates the extract cache.
- A new `lockfile-only` `skipReason` marks these deps. Only the OSV processor clears it, and only once a matched advisory has a known fixed version — `Vulnerabilities.unskipLockfileOnlyDependency`. Deps skipped or disabled for any other reason, including the user's own config, are untouched. The generated vulnerability package rule is unchanged apart from this, so there is no cross-ecosystem behavior change.
- The `UvLockfile` schema now captures each package's `source.registry`. The workspace root and any packages not resolved from an index have no registry and are skipped. Non-default indexes are carried through as `registryUrls`.
- Under the default `vulnerabilityFixStrategy=lowest`, `uv lock --upgrade-package` is pinned to the resolved version. A pin held back by another package's constraint surfaces as an artifact error, instead of resolving to a version that doesn't carry the fix.

The same mechanism would generalize to other lockfile managers (poetry, PEP 751, cargo); this PR keeps it to uv as a first implementation.

Some quirks I've noticed and would like to point out/discuss:

- Transitive deps from non-default indexes carry the index URL from `uv.lock` as `registryUrls`. Credentials from those come from `hostRules` the same way it's done for non-transitive deps.
- Clearing the `skipReason` mutates a dep that lives in `cache.scan[baseBranch].packageFiles`, so a later run reusing a valid extract cache sees it unskipped. `malicious-version-in-use` already works this way, so this looks like a pre-existing property of clearing a skip during lookup rather than something new here — happy to address it in this PR if you'd prefer.

## Context

Related discussions: #36793, #22049. I opted to open a PR with a working, tested implementation rather than only a proposal, but I'm very happy to convert this to a discussion-first conversation if you'd rather agree the approach before reviewing code.

## AI assistance disclosure

The code and tests were written by Claude Code (models: Claude Opus 4.8 and Claude Opus 5), with me directing the work, reviewing each change, and running the test suite.

`osvVulnerabilityAlerts` in `configuration-options.md` stated that alerts only cover direct dependencies, which is no longer true for pep621, so that is updated. The new `uv.lock` `depType` carries a description via `dep-types.ts`, which feeds the generated manager docs.

Unit tests: transitive deps are surfaced with `skipReason: 'lockfile-only'` when `osvVulnerabilityAlerts` is enabled and absent when it isn't; the OSV processor clears that skip when a fix exists, keeps it when none does, and leaves other skip reasons alone; `uv lock --upgrade-package` is pinned for `uv.lock` deps.

Reproduction repo and the resulting PR:
- Fixture + observed output: https://github.com/khvn26/renovate-uv-transitive-osv-repro
- The PR it generated: https://github.com/khvn26/renovate-uv-transitive-osv-repro/pull/1

On that repo, the four `uv.lock`-only packages are extracted as skipped with no datasource lookups. `idna 3.6` matches GHSA-65pc-fj4g-8rjx, GHSA-jjg7-2v4v-x38h, PYSEC-2024-60 and PYSEC-2026-215, gets its skip cleared, and is remediated with `uv lock --upgrade-package idna==3.15`. With `osvVulnerabilityAlerts` off, extraction drops from 6 deps to 2 and no branch is created.


RahulGautamSingh (commented):
Thanks for the PR! It has been a widely asked feature which we have implemented and then reverted for npm ie. `transitiveRemediation` so please give the team sometime to build a consensus on this.

Also, due to [summer holidays](https://github.com/renovatebot/renovate/discussions/44212) it could take some time for the reviews and discussions.


I'll in post my review of the implementation soon.

lib/modules/manager/pep621/schema.ts
```diff
@@ -219,17 +219,29 @@ export const PdmLockfile = Toml.pipe(
   )
   .transform((lock) => ({ lock }));
 
+export interface UvLockedPackage {
+  version: string;
+  // The index a package was resolved from, if any. Absent for the workspace
+  // root and for virtual/editable/path/git packages, which cannot be
+  // remediated by `uv lock --upgrade-package`.
```
RahulGautamSingh: use jsdoc comment

lib/workers/repository/process/vulnerabilities.spec.ts
```diff
@@ -1127,6 +1127,9 @@ describe('workers/repository/process/vulnerabilities', () => {
           matchPackageNames: ['stdlib'],
           matchCurrentVersion: '1.7.5',
           allowedVersions: '>= 1.7.6',
+          // remediation must override a disabled dependency, e.g. a transitive
+          // dep surfaced from a lockfile
```
RahulGautamSingh: jsdoc comment

lib/workers/repository/process/vulnerabilities.ts
```diff
@@ -600,6 +600,10 @@ export class Vulnerabilities {
       matchCurrentVersion: depVersion,
       versioning,
       allowedVersions: fixedVersion,
+      // Remediate even when updates are otherwise disabled for the dependency,
+      // e.g. transitive deps surfaced from a lockfile. Consistent with how
+      // vulnerability alerts already override schedule and minimumReleaseAge.
```
RahulGautamSingh: jsdoc comment

lib/modules/manager/pep621/processors/uv.ts
```diff
@@ -32,6 +32,17 @@ import { BasePyProjectProcessor } from './abstract.ts';
...
 const uvUpdateCMD = 'uv lock';
 
+// The public PyPI index, as recorded in `uv.lock` package sources. Transitive
+// deps resolved from it use the datasource default, so no `registryUrls` needed.
+const defaultPypiRegistries = new Set([
+  'https://pypi.org/simple',
+  'https://pypi.org/simple/',
+]);
+
+function isDefaultPypiRegistry(registryUrl: string): boolean {
```
RahulGautamSingh: only used once, inline this

lib/workers/repository/process/vulnerabilities.ts
```diff
@@ -600,6 +600,10 @@ export class Vulnerabilities {
...
+      enabled: true,
```
RahulGautamSingh: enabled: true is a global behavior change, not scoped to uv.

needs maintainers input on this @viceice @secustor

lib/workers/repository/process/vulnerabilities.spec.ts
khvn26: 9a2579f

lib/workers/repository/process/vulnerabilities.ts
khvn26: 9a2579f

lib/modules/manager/pep621/processors/uv.ts
khvn26: 6284112

lib/modules/manager/pep621/schema.ts
khvn26: 9a2579f

lib/workers/repository/process/vulnerabilities.ts
khvn26: Yeah I can see how this can be regarded as too broad.

One way to make this cleaner could be to add a dedicated `lockfile-only` `skipReason` and have the OSV processor clear it when we need to push a fix bump.

RahulGautamSingh (changes requested):
As it stands, the global `enabled: true` in `getRule()` would re-enable any dependency a user has deliberately disabled, as long as a matching OSV alert exists. 

That is a broad behavior change and needs maintainer sign off before we go that route. @viceice @jamietanna @secustor thoughts? 

I think a dedicated `skipReason` that only the OSV processor clears keeps the override scoped to the deps this PR surfaces, instead of flipping `enabled: true` on every vulnerability rule across all managers and datasources .. is a better approach. as @khvn26 suggested [here](https://github.com/renovatebot/renovate/pull/44262#discussion_r3486514515)

One related note: transitive deps are surfaced on every extraction regardless of whether `osvVulnerabilityAlerts` is set, so a large `uv.lock` adds many disabled deps for users who will not use the feature. Please add logic to prevent this.

lib/workers/repository/process/vulnerabilities.ts
khvn26: See 1a61082.

khvn26:
Went ahead and implemented a dedicated `skipReason` in 1a61082. Additionally, noticed that `uv lock --upgrade-package` didn't honour `vulnerabilityFixStrategy`; fixed in 3835995.

Another observation: clearing the `skipReason` mutates the scan cache for the dep, so a later run might see it unskipped. The same behaviour applies to `malicious-version-in-use`, so not sure if this should be fixed (as part of this PR, at least).

RahulGautamSingh:
LGTM, but it needs another set of eyes before merge.

docs/usage/configuration-options.md
```diff
@@ -3020,7 +3020,7 @@ If configured to `true`, it means that any `.npmrc` file in the repo will have `
 Renovate integrates with [OSV](https://osv.dev/), an Open Source vulnerability database, to check if extracted dependencies have known vulnerabilities.
 Set `osvVulnerabilityAlerts` to `true` to get pull requests with vulnerability fixes (once they are available).
 
-You will only get OSV-based vulnerability alerts for _direct_ dependencies.
+You will only get OSV-based vulnerability alerts for _direct_ dependencies. The only exception today is the `pep621` manager.
```
```suggestion
You will only get OSV-based vulnerability alerts for _direct_ dependencies. 
The only exception today is the `pep621` manager.
```
one sentence per line

lib/workers/repository/process/vulnerabilities.ts
```diff
@@ -282,6 +289,40 @@ export class Vulnerabilities {
     }
   }
 
+  private unskipLockfileOnlyDependency(
```
RahulGautamSingh, line 305: add a jsdoc comment for this method

RahulGautamSingh (commented):
LGTM except the nits.

I have asked for maintainers for a review as well.

lib/workers/repository/process/vulnerabilities.ts
khvn26: 6ffd5a01d6

RahulGautamSingh (approved):
Needs a review from the maintainers.

lib/modules/manager/types.ts
```diff
@@ -36,6 +36,7 @@ export interface ExtractConfig extends CustomExtractConfig {
   repository?: string;
   currentDigest?: string;
   newDigest?: string | null;
+  osvVulnerabilityAlerts?: boolean;
```
jamietanna: I think we should make this a generic field:

```suggestion
  vulnerabilityAlerts?: boolean;
```

As we should allow this to happen regardless of the source of vulnerability data

(See also: https://github.com/renovatebot/renovate/issues/44153)

jamietanna (commented):
Will finish my review tomorrow - we'll need to do a bit of refactoring for this to work well (and therefore be usable with other managers in the future)

lib/modules/manager/types.ts
khvn26: Settled on `vulnerabilityAlertsEnabled` since `vulnerabilityAlerts` already exists in `RenovateConfig` as `RenovateSharedConfig`. See 99c5eedb8d.

jamietanna (changes requested):
Thanks for your patience, we'll want to convert this into a set of PRs i.e.:

- refactor(types): add `skipReason: lockfile-only`
- fix(types): add a `hasVulnerabilityAlertsRules`
  - using `vulnerabilityAlerts`, which has matches OR using `osvVulnerabilityAlerts` (which may have matches)
  - `hasVulnerabilityAlertsRules` is also added into the extraction cache - so this needs to be a `fix`
- feat(vulnerabilities): allow performing lockfile only transitive security updates
  - add `unskipLockfileOnlyDependency`/`clearLockfileOnlySkip` in i.e. `lib/workers/repository/process/lockfile-only-deps.ts``
  - this isn't directly in the `vulnerabilities.ts`, but in a shared file so it can be used across both GitHub-native vulnerability alerts and OSV
  - wire it into `vulnerabilities.ts` (for OSV) and `extract-update.ts` (for GitHub)
  - make sure that https://github.com/renovatebot/renovate/blob/f2a8dc2f2800668dc7b0a74676058ebac2e9ffde/lib/workers/repository/process/extract-update.ts#L214 is updated to also include `hasVulnerabilityAlertsRules`
- docs: add metadata to note which managers support lockfile only transitive security updates
  - (I can do this one if needed - it'll not be used right now in the docs, but will be once https://github.com/renovatebot/renovate/issues/46376 is in)
- feat(manager/pep621): extract lockfile if `hasVulnerabilityAlertsRules`
  - the core change
  - and making sure metadata is set so it's clear that this manager now has (partial) support

PR #44323, opened 2026-07-01
fix(workers/branch): preserve caller-set rebaseRequested through rebaseCheck

zeitlinger:
## Changes

`processBranch` unconditionally overwrites `config.rebaseRequested` with the result of `rebaseCheck()`. A caller that pins `rebaseRequested: true` — either via `RENOVATE_FORCE='{"rebaseRequested":true}'` or by setting the field on the config passed in — loses that value whenever the PR has no `"rebase!"` title prefix and no configured `rebaseLabel`, because `rebaseCheck()` returns `false` in that case and clobbers the pin.

This change preserves the caller-set value: only consult `rebaseCheck()` when `config.rebaseRequested` isn't already `true`. `rebaseCheck()`'s existing behavior (auto-detect a rebase intent from PR title/label) is unchanged when the caller hasn't pinned anything.

## Context

## AI assistance disclosure

Claude (Anthropic) drafted the code change and the accompanying test case.

## Documentation

## How I've tested my work

The public repository: n/a

### DIFF
--- a/lib/workers/repository/update/branch/index.spec.ts
+++ b/lib/workers/repository/update/branch/index.spec.ts
@@ -445,6 +445,28 @@ describe('workers/repository/update/branch/index', () => {
       });
     });
 
+    it('does not skip edited PR when caller pinned rebaseRequested', async () => {
+      schedule.isScheduledNow.mockReturnValueOnce(false);
+      scm.branchExists.mockResolvedValue(true);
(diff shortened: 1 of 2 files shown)


viceice:
`rebaseRequested` isn't a public config option. use `checkedBranches`

https://docs.renovatebot.com/self-hosted-configuration/#checkedbranches

PR #43899, opened 2026-06-09
feat(manager/swift): support Swift Package Registry (SE-0292) id-form dependencies

bdolewski:
## Changes

Add support for [SE-0292 Swift Package Registry](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0292-package-registry.md) identity-form dependencies to the `swift` manager.

The manager now recognizes:

```swift
.package(id: "scope.name", from: "1.0.0"),
```

alongside the existing URL form. Identity-form dependencies are resolved against a new `swift-package-registry` datasource that speaks the SE-0292 list-versions protocol. Registry URLs are auto-discovered from `.swiftpm/configuration/registries.json`. `Package.resolved` pins of `kind: "registry"` are updated on version bump (and bridged `kind: "remoteSourceControl"` pins are matched by identity for the same dep).

### Background for non-iOS maintainers

[SE-0292](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0292-package-registry.md) introduced an HTTP package-registry protocol for SwiftPM, parallel to the classic SCM-clone resolution. Adoption is growing: Apple's preview registry, the open [Tuist Registry](https://tuist.dev/blog/2025/11/26/opening-registry), and several enterprise registry servers all speak it. Teams that switch from `.package(url:)` to `.package(id:)` lose Renovate's auto-bump PRs for every migrated dependency today — this PR closes that gap.

### Approach

Strictly additive across the swift manager state machine and the datasource layer:

1. **Datasource** (`lib/modules/datasource/swift-package-registry/`) — calls `GET <registry>/<scope>/<name>` with `Accept: application/vnd.swift.registry.v1+json`. Bearer/Basic auth via standard `hostRules`. Releases carrying an SE-0292 `problem` block are excluded. `registryStrategy: 'hunt'` so multiple discovered registries are tried in order.
2. **Parser** (`lib/modules/manager/swift/extract.ts`) — new state-machine branches for `.package(id:`; downstream version-clause states (`from:`, ranges, `.exact`, `exact:`, trailing `traits:`) are reused unchanged.
3. **Registry discovery** (`lib/modules/manager/swift/registries-json.ts`) — `extractAllPackageFiles` reads `.swiftpm/configuration/registries.json` and attaches found URLs to id-form deps. `customRegistryUrls` / `hostRules` still override.
[...]

PR #44358, opened 2026-07-03
fix(swift): support Package Registry pins in Package.resolved

bdolewski-intellias:
## Summary

Renovate cannot update `Package.resolved` on any Swift project that mixes SwiftPM Package Registry (SE-0292) dependencies with regular source-control dependencies. It silently returns `null` and the lockfile stays stale on every Renovate PR against that project.

## The problem in one sentence

The `swift` manager's Zod schema requires a `revision` field on every pin in `Package.resolved`. That is correct for source-control pins (they have a git commit SHA). Registry pins have no revision, SE-0292 identifies packages by a version string alone. A lockfile with even one registry pin fails validation and `updateArtifacts` bails out.

## Example

A perfectly valid `Package.resolved` fragment that Renovate rejects today:

```json
{
  "pins": [
    {
      "identity": "example.registry-package",
      "kind": "registry",
      "location": "",
      "state": { "version": "1.2.3" }
    }
  ],
  "version": 2
}
```

Failure mode: `PackageResolvedJson.safeParse()` returns an error, `artifacts.ts` logs `"swift: could not parse Package.resolved"` at debug level, `updateArtifacts` returns `null`. The `Package.swift` edit still lands (the manifest manager works fine), but the lockfile stays untouched. Users see a Renovate PR that updates `Package.swift` and produces a lockfile mismatch, with no visible error to explain why.

## Fix

- Make `revision` optional on `PackageResolvedPin.state` in `schema.ts`.
- No other code changes needed: `artifacts.ts` already guards its revision-replace branch with `if (newRevision)`, so registry pins are safely left alone when the resolver is processing a source-control dep.

## Test

`v2MixedFixture` added to `artifacts.spec.ts`: one source-control pin plus one registry pin. New test `parses lockfile with mixed source-control and registry pins` verifies:

- Whole lockfile parses (before this change it panics at parse time)
- Source-control pin's `version` and `revision` get updated as expected
- Registry pin's `version` is preserved
- Registry pin does NOT gain a spurious `revision` field after the update

`pnpm vitest run lib/modules/manager/swift/artifacts.spec.ts` -> 32/32 pass.

## Context

Background:

- Follow-up to #41534 (initial `Package.resolved` support). That PR's test suite only covered source-control pins; the mixed-pins case was never tested.
- Prerequisite for #43899 (still open, adds `.package(id:)` id-form dependency support). id-form deps produce registry-kind pins in `Package.resolved`. #43899 cannot land cleanly until Renovate can parse those pins in the first place.

## AI assistance disclosure

## Documentation

## How I've tested my work


lib/modules/manager/swift/schema.ts
```diff
@@ -6,7 +6,12 @@ const PackageResolvedPin = z.object({
   kind: z.string(),
   location: z.string(),
   state: z.object({
-    revision: z.string(),
+    // `revision` is required for source-control pins (kind: "remoteSourceControl"),
+    // but absent for SwiftPM Package Registry pins (kind: "registry") per SE-0292:
+    // registry packages are identified by version alone, not by a commit SHA. A
+    // Package.resolved file mixing both kinds fails validation if we require
+    // `revision` unconditionally.
+    revision: z.string().optional(),
```
RahulGautamSingh: `.optional()` also makes revision optional for `kind: "remoteSourceControl"` pins, where it's actually mandatory.

 A discriminated union on kind would enforce that.

RahulGautamSingh (commented):
Please follow the PR template we have for the Context section

lib/modules/manager/swift/schema.ts
bdolewski-intellias: @RahulGautamSingh 

Good catch, thanks. You’re right that .optional() makes the schema too permissive for source-control pins, where revision is required.

Proposed discriminated union:
```ts
const PackageResolvedPin = z.discriminatedUnion('kind', [
  z.object({
    identity: z.string(),
    kind: z.literal('remoteSourceControl'),
    location: z.string(),
    state: z.object({
      revision: z.string(),
      version: z.string().nullable().optional(),
      branch: z.string().nullable().optional(),
    }),
  }),
  z.object({
    identity: z.string(),
    kind: z.literal('registry'),
    location: z.string(),
    state: z.object({
      version: z.string().nullable().optional(),
    }),
  }),
]);
```

One question before I push: SwiftPM also has `kind: "localSourceControl"` for `.package(path: ...)` dependencies. I don’t believe it appears in `Package.resolved`, since path dependencies aren’t versioned and there is nothing to pin.

I can add a third variant for defensive parsing, or keep the two-variant union, which covers everything we’ve observed in practice.

RahulGautamSingh: On second thought, let's keep it `.optiona()`. 

`z.discriminatedUnion('kind', [...])` closes the set of kinds. Any pin whose kind isn't one of the listed literals fails the variant → z.array(pins) fails → safeParse fails → updateArtifacts logs "could not parse Package.resolved" and skips the entire file.

RahulGautamSingh (approved):

bdolewski-intellias:
@RahulGautamSingh fixed. Merged main in to resolve the conflict. Approval got auto-dismissed by the merge, ready for another look.

lib/modules/manager/swift/artifacts.spec.ts
```diff
@@ -72,6 +72,37 @@ const v3Fixture = JSON.stringify(
   2,
 );
 
+// Package.resolved that mixes source-control pins and SwiftPM Package
+// Registry (SE-0292) pins. Registry pins have `kind: "registry"`,
+// `location: ""`, and only a `version` in `state` — no `revision`,
+// because registry packages are identified by version alone.
+const v2MixedFixture = JSON.stringify(
```
secustor: Inline the fixture in the test

lib/modules/manager/swift/artifacts.spec.ts
```diff
@@ -199,6 +230,49 @@ describe('modules/manager/swift/artifacts', () => {
...
 
+  it('parses lockfile with mixed source-control and registry pins', async () => {
+    // Regression test: SwiftPM Package Registry (SE-0292) pins have
+    // `kind: "registry"` and no `revision` field, because registry packages
+    // are identified by version alone. Before revision was made optional in
+    // the Zod schema, any Package.resolved mixing source-control and
+    // registry pins failed to parse and updateArtifacts silently returned
+    // null with `swift: could not parse Package.resolved`. Verify the mixed
+    // lockfile parses, the source-control pin gets updated, and the
+    // registry pin is preserved untouched.
```
secustor:
```suggestion
    // Regression test: SwiftPM Package Registry (SE-0292)
```

lib/modules/manager/swift/artifacts.spec.ts
```diff
@@ -199,6 +230,49 @@ describe('modules/manager/swift/artifacts', () => {
...
+    expect(contents).toContain('"version": "5.10.0"');
+    expect(contents).toContain('"revision": "newrevisionsha123"');
+    // Registry pin preserved
+    expect(contents).toContain('"identity": "example.registry-package"');
+    expect(contents).toContain('"kind": "registry"');
+    expect(contents).toContain('"version": "1.2.3"');
+    // Registry pin has no revision — must NOT gain one after the update
+    const registryBlock = contents.slice(
+      contents.indexOf('example.registry-package'),
+    );
```
secustor: Use `toMatchObject`

lib/modules/manager/swift/schema.ts
```diff
@@ -7,7 +7,12 @@ const PackageResolvedPin = DeepNullish(
     kind: z.string(),
     location: z.string(),
     state: z.object({
-      revision: z.string(),
+      // `revision` is required for source-control pins (kind: "remoteSourceControl"),
+      // but absent for SwiftPM Package Registry pins (kind: "registry") per SE-0292:
+      // registry packages are identified by version alone, not by a commit SHA. A
+      // Package.resolved file mixing both kinds fails validation if we require
+      // `revision` unconditionally.
```
secustor, line 14: This should be then a discriminated union with `kind` as key

secustor (changes requested):

lib/modules/manager/swift/artifacts.spec.ts
```diff
@@ -199,6 +199,75 @@ describe('modules/manager/swift/artifacts', () => {
...
+    // Source-control pin updated
+    expect(scmPin).toMatchObject({
+      state: { version: '5.10.0', revision: 'newrevisionsha123' },
+    });
+    // Registry pin preserved, and must NOT gain a revision after the update
+    expect(registryPin).toMatchObject({
+      kind: 'registry',
+      state: { version: '1.2.3' },
+    });
+    expect(registryPin!.state).not.toHaveProperty('revision');
```
secustor:
```suggestion
  await expect(
    updateArtifacts({
      packageFileName: 'Package.swift',
      updatedDeps: [
        {
          depName: 'Alamofire/Alamofire',
          datasource: GithubTagsDatasource.id,
          newVersion: 'v5.10.0',
          newValue: '5.10.0',
        },
      ],
      newPackageFileContent: '',
      config: {},
    }),
  ).resolves.toEqual([
    {
      file: {
        type: 'addition',
        path: 'Package.resolved',
        contents: JSON.stringify(
          {
            pins: [
              {
                identity: 'alamofire',
                kind: 'remoteSourceControl',
                location: 'https://github.com/Alamofire/Alamofire',
                state: {
                  revision: 'newrevisionsha123',
                  version: '5.10.0',
                },
              },
              {
                identity: 'example.registry-package',
                kind: 'registry',
                location: '',
                state: { version: '1.2.3' },
              },
            ],
            version: 2,
          },
          null,
          2,
        ),
      },
    },
  ]);
```

secustor (commented):
Small readability improvment

lib/modules/manager/swift/schema.ts
bdolewski-intellias: Thanks @secustor!
Heads-up that this is the opposite of what @RahulGautamSingh asked for earlier on this same line. He initially suggested a discriminated union too, then reversed and asked me to keep `.optional()`, because `z.discriminatedUnion('kind', ...)` closes the set of kinds.

 Any pin whose kind isn't a listed literal fails the whole  `safeParse`, so updateArtifacts logs "could not parse Package.resolved" and skips the entire file. A future/unknown SwiftPM pin kind would silently disable Renovate for that repo.

Could you two align on the direction you'd prefer? Happy to implement either the discriminated union or keep `.optional()`. If it helps, one middle ground is a discriminated union over the known kinds (remoteSourceControl requiring revision, registry without) plus a permissive fallback variant for any other kind, which enforces the constraint while staying forward-compatible.

bdolewski-intellias:
> Small readability improvment

Applied

RahulGautamSingh (approved):

viceice (approved):

secustor (approved):

PR #44390, opened 2026-07-06
feat(manager): add apm manager for Agent Package Manager

MPV:
## Changes

Adds a new `apm` manager that keeps [APM (Agent Package Manager)](https://github.com/microsoft/apm) dependencies up to date, as requested in [discussion #42507](https://github.com/renovatebot/renovate/discussions/42507).

APM projects declare git-pinned agent dependencies (skills, plugins, agents, full packages) in an `apm.yml` manifest and pin the resolved tree in an `apm.lock.yaml` lockfile. Nothing currently bumps those refs.

The new manager:

- Parses the string entries under `dependencies.apm` and `devDependencies.apm`, of the form `[host/]owner/repo[/subpath]#<ref>`.
- Selects the datasource dynamically from the git host (reusing `detectPlatform`, which also honors `hostRules`) — **no new datasource is required**:
  - `github.com` (and self-hosted GitHub) → `github-tags`
  - `gitlab.com` (and self-hosted GitLab) → `gitlab-tags`
  - any other host (Bitbucket, Azure DevOps, …) → `git-tags`
- Updates entries pinned to a semver tag/range (`#<version>`). Entries **SHA-pinned with a trailing `# <tag>` comment** (`owner/repo#<sha> # v2.0.0`) are updated as **digests** under `pinDigests`, mirroring the `github-actions` `uses: owner/action@<sha> # v4` idiom. Bare SHAs (no recoverable tag) and unpinned entries are skipped with an explicit `skipReason`.
- Regenerates `apm.lock.yaml` (when present) by running `apm install`, wired through a new `apm` tool constraint that resolves the CLI from `microsoft/apm` GitHub releases. `apm install` also re-deploys the harness directories APM consumers commit (`.github/`, `.claude/`, …), so `updateArtifacts` returns **every file the install changed** — not just the lockfile — to keep committed instruction files from going stale after a bump.

### Supported dependency forms

To keep scope explicit for reviewers:

- **Handled:** the shorthand string form `[host/]owner/repo[/subpath]#<ref>`, where `<ref>` is a semver tag/range, or a SHA with a trailing `# <tag>` comment (digest). Includes virtual-package subpaths (`skills/`, `prompts/`, `instructions/`, …) and GitLab nested groups (3+ segment project slugs).
- **Deferred — flagged for a maintainer scope decision:**
  - Full-URL / SSH forms (`https://github.com/owner/repo#v1`, `git@github.com:owner/repo#v1`) are not parsed today.
  - Object / expanded manifest entries — the schema keeps string entries only (`LooseArray(z.string())`), so any object-form dependency is dropped before extraction.
  - Branch refs (`#main`) — parsed but effectively a no-op, since there is no version or digest to track.

  Open question for maintainers: for each of the above, would you prefer explicit `skipReason`s (visibly unsupported) or full support in this PR vs. follow-ups?

MCP server entries under `dependencies.mcp` are intentionally not managed: they are identified by an MCP registry name (e.g. `io.github.github/github-mcp-server`) and carry no version to pin, so there is nothing for Renovate to update.

### Lock file maintenance note

Renovate's [adding-a-package-manager guide](https://github.com/renovatebot/renovate/blob/main/docs/development/adding-a-package-manager.md) prefers a `<pm> update` command over delete-then-`<pm> install`, since the latter can lose platform-specific lock data. APM does not provide an `update` command, so:

- **After a package-file change**, the manager runs `apm install` to sync `apm.lock.yaml` to the updated manifest (no delete).
- **During lock file maintenance**, it deletes `apm.lock.yaml` and runs `apm install` to regenerate it. APM's lockfile records resolved sources and content hashes (no platform-specific fields), so a full regeneration is deterministic and safe here.

### Category

I omitted a manager `category` to keep the diff self-contained (avoiding a change to the shared `categories` constant). I'm happy to add one if preferred — would maintainers like a new `ai` category (APM targets AI coding agents), or is leaving it uncategorized better?

## Context

The New Package Manager Questionnaire is completed in discussion [#42507](https://github.com/renovatebot/renovate/discussions/42507).

## AI assistance disclosure

The manager code, unit tests, and documentation were drafted with Claude (via Claude Code) and reviewed before submission. PR comments/responses on this PR are likewise drafted with Claude Code and reviewed/edited before posting.

100% test coverage on the new files (`pnpm vitest run lib/modules/manager/apm`), plus `pnpm check --all` and `pnpm type-check` passing.

Additionally verified with a local extraction dry-run against a representative `apm.yml`, confirming correct datasource/packageName resolution across hosts:

| Entry | datasource | packageName | currentValue |
| --- | --- | --- | --- |
| `microsoft/apm-sample-package#v1.0.0` | `github-tags` | `microsoft/apm-sample-package` | `v1.0.0` |
| `anthropics/skills/skills/frontend-design#v0.24.0` | `github-tags` | `anthropics/skills` | `v0.24.0` |
| `gitlab.com/acme/agents#v2.3.0` | `gitlab-tags` | `acme/agents` | `v2.3.0` |
| `bitbucket.org/acme/tools#v1.1.0` | `git-tags` | `https://bitbucket.org/acme/tools` | `v1.1.0` |
| `danielmeppiel/design-guide` (unpinned) | — | — | skipped (`unspecified-version`) |
| `owner/repo#v1.2.3` (devDependencies) | `github-tags` | `owner/repo` | `v1.2.3` |

MCP entries under `dependencies.mcp` were correctly ignored.

The public repository: https://github.com/microsoft/apm

🤖 Generated with [Claude Code](https://claude.com/claude-code)


MPV:
(sorry about the force-push, it was necessary to solve the CLA issue, i.e. committing properly as myself)

lib/modules/manager/apm/extract.ts
```diff
@@ -0,0 +1,153 @@
...
+  const hasHost = hostnameRegex.test(segments[0] ?? '');
+  const host = hasHost ? segments[0] : 'github.com';
+  const repoSegments = hasHost ? segments.slice(1) : segments;
+
+  if (repoSegments.length < 2) {
+    logger.debug({ entry }, 'apm: could not determine owner/repo');
+    return {
+      ...base,
+      currentValue,
+      skipReason: 'invalid-dependency-specification' as SkipReason,
```
RahulGautamSingh: drop the cast,

lib/modules/manager/apm/extract.ts
```diff
@@ -0,0 +1,153 @@
...
+  const currentValue = hashIndex === -1 ? '' : entry.slice(hashIndex + 1);
+
+  const base: PackageDependency = {
+    depName: pathPart,
+    depType,
+  };
+
+  if (!currentValue) {
+    // Unpinned dependency (no `#ref`) - nothing for Renovate to update.
+    return { ...base, skipReason: 'unspecified-version' as SkipReason };
```
RahulGautamSingh: drop the case

lib/modules/manager/apm/extract.ts
```diff
@@ -0,0 +1,153 @@
...
+    logger.debug({ packageFile, err }, 'apm: failed to parse YAML');
+    return null;
+  }
+
+  if (!isPlainObject(parsed)) {
+    logger.debug({ packageFile }, 'apm: parsed content is not an object');
+    return null;
+  }
+
+  const manifest = parsed as ApmManifest;
```
RahulGautamSingh: dont cast as such, instead create a zod schema and use with parseSingleYaml as such

`parseSingleYaml(content, { customSchema })`

lib/modules/manager/apm/extract.ts
```diff
@@ -0,0 +1,153 @@
...
+    autoReplaceStringTemplate: '{{depName}}#{{newValue}}',
+  };
+}
+
+function extractSection(
+  deps: ApmDependencies | undefined,
+  depType: string,
+): PackageDependency[] {
+  const result: PackageDependency[] = [];
+  if (!deps || !Array.isArray(deps.apm)) {
```
RahulGautamSingh: You already import from `@sindresorhus/is`; 

`is.array(deps.apm)` is more consistent with the rest of the file.

RahulGautamSingh (changes requested):

lib/modules/manager/apm/extract.ts
MPV: Thanks for the review! Addressed all four points in `3fdb4cf`:

- **Zod schema** — added `schema.ts` with an `ApmManifest` schema and switched to `parseSingleYaml(content, { customSchema: ApmManifest })`, dropping the `as ApmManifest` cast.
- **`as SkipReason` casts** — both removed; the literals infer fine.
- **`is.array(deps.apm)`** — resolved structurally rather than with the `is.array` swap: the schema now models the section as `LooseArray(z.string())`, so it keeps only the string entries and drops MCP objects / non-string / non-array values during parsing. That means the manual `Array.isArray` + `is.string` guards are gone entirely — `extractSection` just maps a typed `string[]`. Let me know if you'd prefer the explicit runtime check kept instead.

Also removed the now-unused `types.ts`. Coverage stays at 100% on the module and `type-check` passes.

lib/modules/manager/apm/schema.ts
```diff
@@ -0,0 +1,22 @@
...
+import { LooseArray } from '../../../util/schema-utils/index.ts';
+
+/**
+ * APM dependencies are declared under `dependencies.apm` / `devDependencies.apm`
+ * as an array of `[host/]owner/repo[/subpath]#<ref>` strings.
+ *
+ * MCP server entries under the sibling `mcp` key are objects without a pinnable
+ * version, so `LooseArray(z.string())` keeps only the string entries and drops
+ * everything else (MCP objects, malformed entries). A non-array `apm` value is
+ * caught and treated as absent.
```
secustor:
```suggestion
```

lib/modules/manager/apm/readme.md
```diff
@@ -0,0 +1,31 @@
...
+  apm:
+    - owner/repo#v1.2.3
+```
+
+The datasource is selected dynamically from the git host:
+
+- `github.com` (the default when no host is given) uses the `github-tags` datasource
+- `gitlab.com` and hosts matching a `gitlab` `hostRules` entry use the `gitlab-tags` datasource
+- any other host uses the generic `git-tags` datasource
+
```
secustor:
```suggestion
```

lib/modules/manager/apm/readme.md
```diff
@@ -0,0 +1,31 @@
...
+Only entries that pin an exact `#<ref>` are updated.
+Entries without a `#<ref>` are skipped because there is no version to bump.
+
+When an `apm.lock.yaml` lockfile is present, Renovate refreshes it by running `apm install` after updating the manifest.
+This requires the `apm` CLI to be available (for example, with `binarySource=global`).
+
+Model Context Protocol (MCP) server entries under `dependencies.mcp` are not managed.
+They are identified by an MCP registry name (for example `io.github.github/github-mcp-server`) and carry no version to pin, so there is nothing for Renovate to update.
```
secustor:
```suggestion
```

lib/modules/manager/apm/schema.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+const ApmDependencySection = z.object({
+  apm: LooseArray(z.string()).optional().catch(undefined),
```
secustor: Your discussion is referring to an object style too. 

Also LooseArray will already discard false items, so default to an empty array to simplify your code.

lib/modules/manager/apm/extract.ts
```diff
@@ -0,0 +1,134 @@
...
+      packageName: repoPath,
+      ...(host === 'gitlab.com' ? {} : { registryUrls: [`https://${host}`] }),
+    };
+  }
+
+  logger.debug({ repoUrl }, 'apm: using git-tags datasource');
+  return { datasource: GitTagsDatasource.id, packageName: repoUrl };
+}
+
+const hostnameRegex = regEx(/\./);
```
secustor: Does APM support local paths where there could be dots in the path?

lib/modules/manager/apm/extract.ts
```diff
@@ -0,0 +1,134 @@
...
+    replaceString: entry,
+    autoReplaceStringTemplate: '{{depName}}#{{newValue}}',
+  };
+}
+
+function extractSection(
+  entries: string[] | undefined,
+  depType: string,
+): PackageDependency[] {
+  return (entries ?? []).map((entry) => parseApmDependency(entry, depType));
```
secustor: Use coerceArray

lib/modules/manager/apm/extract.ts
```diff
@@ -0,0 +1,134 @@
...
+ * the git host. Reuses `detectPlatform` (which also honors `hostRules`) to map
+ * github.com/gitlab.com and their self-hosted variants to the `github-tags` and
+ * `gitlab-tags` datasources, and falls back to the generic `git-tags`
+ * datasource for every other host (Bitbucket, Azure DevOps, etc.).
+ */
+function determineDatasource(host: string, repoPath: string): DatasourceResult {
+  const repoUrl = `https://${host}/${repoPath}`;
+  const platform = detectPlatform(repoUrl);
+
+  if (host === 'github.com' || platform === 'github') {
```
secustor: These host checks shouldn't be necessary

secustor (changes requested):

lib/modules/manager/apm/schema.ts
MPV: Done — `apm: LooseArray(z.string()).catch([])` now, so the field is always a `string[]` (LooseArray already drops non-string/MCP entries) and the downstream `undefined` handling is gone. Trimmed the redundant comment too.

lib/modules/manager/apm/extract.ts
MPV: Done — `extractSection` now uses `coerceArray(entries).map(...)`.

lib/modules/manager/apm/extract.ts
MPV: Agreed — dropped the `host === 'github.com'` / `host === 'gitlab.com'` disjuncts; datasource selection now branches on `detectPlatform` alone. Kept a bare `host === '<default>'` check only to decide whether to set `registryUrls` (self-hosted vs SaaS).

lib/modules/manager/apm/extract.ts
MPV: APM only supports remote git references (`[host/]owner/repo[/subpath]#ref`); local filesystem paths aren't part of the manifest format. The dot check only inspects the first segment to tell a host from an owner — git hostnames always contain a dot, whereas GitHub/GitLab owner names can't, so a dotted first segment unambiguously denotes the host. Dots in repo names or subpaths (later segments) are left untouched; added a regression test covering `owner/repo.js` and `github/awesome-copilot/agents/api-architect.agent.md`.

MPV:
Thanks for your reviews — all points addressed (schema default + `coerceArray`, redundant host checks removed, local-path/dots question answered with a regression test, readme trimmed). CI is green and coverage is 100% on the module. Ready for another look when you have a moment.

lib/modules/manager/apm/extract.ts
secustor: If you have this regex already, than use named capture groups to simplify parseApmDependency

secustor (dismissed):

MPV:
Since the last approval, `main` shifted under this — `localDir` moved into `InternalGlobalConfigOptions`:

- https://github.com/renovatebot/renovate/pull/44635

I pulled `main` in to adapt the spec to it, which dismissed the approval and left `license/cla` pending.

I've traced the CLA one to my merge commit — it landed under a `Claude <noreply@anthropic.com>` author rather than me, so cla-assistant sees an unsigned contributor (every other commit is mine and covered). The cleanest fix I can see is to re-author that commit as me, but that means a force-push, which the guidelines steer away from. Would you rather I do that, or is there a way you'd prefer to handle it?

Anything else I should sort before it's mergeable again?

secustor:
This is an exception for force pushes being acceptable

MPV:
I think the CI failure is a workflow/pnpm issue rather than this PR. The `Schedule test shards` step does:

- `pnpm -s schedule-test-shards >> "$GITHUB_OUTPUT"`

and pnpm 11.15.0 seems to print `? Verifying lockfile against supply-chain policies (N entries)…` to **stdout** (even under `-s`), so that line ends up in `$GITHUB_OUTPUT` → `Unable to process file command 'output'` / `Invalid format`. `test-success` then fails only because `test` got skipped.

It looks cache/lock-state dependent (the entry count changed between my runs), which might be why `main` pushes stay green while PRs hit it. Re-merging `main` didn't help. Anything I can do on my side, or is this one for the CI setup?

MPV:
This had an approval, then CI went red. 

That failure turned out to be the upstream `schedule-test-shards` / pnpm `$GITHUB_OUTPUT` breakage (rather than anything in this PR), which has now been fixed by:

- https://github.com/renovatebot/renovate/pull/44847

I've re-merged `main` to pull it in, so things look green again. 

The original approval got dismissed by the commits added while sorting that out, but should be good to go again now. 🤞

vlsi:
Does this PR support "apm marketplace dependencies"? AFAIK they do not support branch references, and they require explicit sha in apm.yml.

MPV:
Thanks — good question. This manager is intentionally (to begin with) scoped to git-hosted dependencies pinned to a tag/semver (`owner/repo#v1.2.3`), which Renovate keeps current via the github-tags / gitlab-tags / git-tags datasources.

Registry- and marketplace-routed dependencies are a different mechanism: they use the object form (`registry:` / `marketplace:`) and resolve through a registry/marketplace rather than a git host, so updating them would need registry-aware version resolution — not the same tag lookup this manager does. That's a meaningfully larger piece, so we deliberately scoped it out of this first version to keep the change focused and reviewable. Those entries are simply left untouched here rather than partially handled.

Since this one already had a green review, hopefully we can keep it as-is and take marketplace/registry support to a follow-up issue or PR. Happy to dig into the options there, and whether/how it's feasible. Does that sound okay to you too?

---

_Drafted with Claude Code; reviewed and edited by me before posting._

vlsi:
**Please recognize the documented SHA-pinned form; ideally support it via `pinDigests`**

Two facts frame this:

1. APM's docs recommend SHA pinning for reproducibility, keeping the release tag as a trailing comment, and `apm update` maintains that form ([manage-dependencies → Pin a version](https://microsoft.github.io/apm/consumer/manage-dependencies/#pin-a-version)):

   ```yaml
   dependencies:
     apm:
       - acme/playbooks#b1c2d3e4f5a6b7c8d9e0f1234567890abcdef123 # v2.0.0
   ```

2. Renovate already produces exactly this shape for GitHub Actions — `uses: owner/action@<sha> # v4` — through `pinDigests` (enabled by `config:best-practices`). The `currentDigest` + `currentValue` machinery and the pin UX already exist.

Today `parseApmDependency` splits on the first `#`, so the documented form parses as `currentValue = "<sha> # v2.0.0"`, which no `*-tags` datasource can resolve. The entry is a silent no-op.

Two ways forward:

**Option A — support `pinDigests`, keep the SHA in `apm.yml`.** Parse `#<sha> # <ref>` into `currentDigest` + `currentValue`, set `autoReplaceStringTemplate` to `'{{depName}}#{{newDigest}} # {{newValue}}'`, and rely on the datasources' `getDigest` (`github-tags` and `gitlab-tags` have it). Then `pinDigests` behaves as it does for GitHub Actions: it pins a bare `#<tag>` to `#<sha> # <tag>` and keeps both current. The manifest stays the human-reviewed record of what actually changed; `apm.lock.yaml` is treated as the generated, technical artifact.

**Option B — revisit APM's policy and docs first.** If APM would rather keep digests only in `apm.lock.yaml` and keep the manifest on tag or branch names, update the docs to recommend that, and the manager can then treat a clean `#<tag>` as the norm. That is an upstream-APM decision, not something the Renovate manager should assume on its own.

Either way, one minimum bar: the manager should recognize the SHA-pinned form APM documents today and leave it intact — update both fields, or skip it explicitly — rather than silently ignoring it or rewriting it in a way that drops the documented SHA.

rcollette:
If the apm.yml file being discussed here, in which to put the sha digest, is the one that apm consumers would use, there is no way I'm putting a sha digest in it.   There's not a single other package management solution where I have to do this.

I would be fine with a sha digest going in the apm lock file though and actually that seems quite normal.

Very explicitly our internal apm marketplace uses the recommended tag_pattern versioning mechanism.
https://github.com/microsoft/apm/discussions/1332#discussioncomment-16967125

lib/modules/manager/apm/extract.ts
```diff
@@ -0,0 +1,144 @@
...
+/**
+ * APM dependency path `[host/]owner/repo[/subpath...]`. APM only supports remote
+ * git references (no local paths). The optional host prefix is always a hostname
+ * (so it contains a dot), whereas owner names do not - `owner` therefore excludes
+ * dots, which disambiguates the leading segment. Dots in repo names or subpaths
+ * (e.g. `owner/repo.js`, `github/awesome-copilot/agents/api-architect.agent.md`)
+ * are left untouched.
+ */
+const apmDepRegex = regEx(
+  /^(?:(?<host>[^/]+\.[^/]+)\/)?(?<owner>[^/.]+)\/(?<repo>[^/]+)(?:\/.*)?$/,
```
RahulGautamSingh: This captures only owner/repo, so GitLab nested groups will break. 

`gitlab.com/group/subgroup/project#ref` parses repo as subgroup and drops `/project`, so `gitlab-tags` queries the wrong project and finds nothing.

vlsi:
>There's not a single other package management solution where I have to do this.

@rcollette, you don't have to, however, digest pinning is recommended for security reasons:
* GitHub recommends pinning GitHub Action digests, see https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions, and Renovate supports that via `helpers:pinGitHubActionDigests`.
* Docker recommends pinning base image digests as well, see https://docs.docker.com/build/building/best-practices/#pin-base-image-versions
* The current [APM documentation](https://microsoft.github.io/apm/consumer/manage-dependencies/#pin-a-version) suggests going for SHA in apm.yml for immutability and reproducibility reasons (see `For reproducibility, prefer tags or SHAs`)

Renovate includes docker and github "pin digests" in `best-practices` preset, see https://docs.renovatebot.com/presets-config/#configbest-practices.

That is why my "option A" goes for "support digest pinning just like Renovate's GitHub Actions update/pinDigest"

MPV:
**SHA pinning:** the manager now recognizes the documented `owner/repo#<sha> # v2.0.0` form and leaves it untouched (skipped, never rewritten), so nothing is silently mishandled. Full digest updating (like GitHub Actions `pinDigests`) is more involved here — APM keeps the tag in a YAML *comment* that a normal parse strips, so it needs a comment-preserving/line-based extractor, not a small tweak.

I'd be happy to merge this as-is (tags/semver + the safe SHA skip) and add the digest-pinning support in a follow-up PR — hoping to get to it myself before long. But let me know if you'd rather wait longer and have it be part of this one.

---

_Drafted with Claude Code; reviewed and edited by me before posting._

MPV:
On second thought, I made some time, and implemented option A for pinned digest updates — thanks for the concrete steer and the template. The manager now recognizes the documented SHA-pinned form and, with `pinDigests`, keeps the SHA and tag current the same way `github-actions` does (`{{depName}}#{{newDigest}} # {{newValue}}`).

The wrinkle was that APM's tag is a YAML *comment*, which the structured parse drops — so the manager scans the raw manifest to recover the tag (`currentValue`) and the exact text to rewrite, then leans on the datasources' `getDigest`. It's opt-in: without `pinDigests` nothing changes (tag/semver updates as before), so tags-only setups stay untouched. A bare `owner/repo#<sha>` with no tag comment stays skipped (nothing to track), and a quoted `"…#<sha>" # <tag>` is skipped for now since we can't rewrite it verbatim.

Vladimir, two things I'd value your eyes on: the SHA↔`#` whitespace normalizes to a single space on update (matching github-actions' default) — fine, or worth preserving? And `git-tags` (non-github/gitlab hosts) may not expose `getDigest`, so those are best-effort. I'd also appreciate a sanity check that this matches what you had in mind.

---

_Drafted with Claude Code; reviewed and edited by me before posting._

vlsi:
Thanks for adding SHA pinning. Two answers and two findings from a closer read.

**Q1 (whitespace → single space).** Good — matching the github-actions default is the right call.

**Q2 (git-tags `getDigest` is best-effort).** It is actually better than best-effort, and there is no silent-SHA-drop risk:

- `GitTagsDatasource extends GitDatasource` and overrides `getDigest` (resolves the tag's commit via `git ls-remote`), and `getReleases` already sets `newDigest: ref.hash` per release. So for non-GitHub/GitLab hosts (Bitbucket, Azure DevOps, self-hosted) the commit is normally resolved and the SHA is preserved on a bump.
- Even when a digest cannot be resolved, lookup sets `newDigest = null` and filters the update out (`res.updates.filter((u) => u.newDigest !== null)` in `lib/workers/repository/process/lookup/index.ts`) and warns — so the bump is held, not rewritten to a digest-less `owner/repo#v2.1.0`. The template's `{{else}}` branch only renders when there is no `currentDigest` at all, i.e. nothing to drop.

So git-tags is "digest-capable via ls-remote" rather than best-effort. One edge left untested: if `getDigest` throws (auth or host error), that is the error path, not the `null` filter — worth a test, though it aborts rather than drops.

**Finding 1 — `findPinnedTail` scans the whole file and takes the first hit (Medium).** It matches the first line whose item text starts with the entry, ignoring which section the entry came from. When the same `[host/]owner/repo[/sub]#<sha>` text appears twice — the same commit pinned in both `dependencies.apm` and `devDependencies.apm`, or a duplicate — both entries resolve to the **first** line's tag and `replaceString`. The second entry then gets the wrong `currentValue` (if its comment differs), and both updates carry the same `replaceString`, so auto-replace rewrites the first line for both: the second `doAutoReplace` edits the wrong occurrence or fails to find its string. Consider scoping the scan to the section being parsed and matching by position, or detecting duplicates and skipping with a clear reason.

**Finding 2 — `virtualFileRegex` omits `.chatmode.md` (Low).** APM's `VIRTUAL_FILE_EXTENSIONS` is `(.prompt.md, .instructions.md, .chatmode.md, .agent.md)`, but the regex is `/\.(?:prompt|instructions|agent)\.md$/`. On nested-group hosts (GitLab, Azure DevOps) a `*.chatmode.md` virtual-file dependency is not detected as the subpath boundary, so `resolveRepoPath` folds it into `owner/repo`. GitHub is unaffected (always `owner/repo`). Adding `chatmode` fixes it. (`virtualRootSegments` matches APM's `_GITLAB_VIRTUAL_ROOT_SEGMENTS` exactly.)

**Tests worth adding:** a duplicate / same-SHA-in-both-sections case (Finding 1), a `*.chatmode.md` virtual file (Finding 2), and a git-tags-host SHA-pin bump asserting the SHA is updated rather than dropped (locks in the Q2 behavior).

shajz:
Thanks for this PR! ❤️ 

**Finding — `updateArtifacts` omits the deployed harness files (Medium).**

`lockFileNames` is `['apm.lock.yaml']` and `artifacts.ts` returns only that sibling, but the `apm install` it runs also rewrites the harness directories APM deploys into. [APM's install docs](https://microsoft.github.io/apm/consumer/install-packages/#what-to-commit) tell consumers to commit those (`.github/`, `.claude/`, `.cursor/`, `.opencode/`, `.gemini/`, `.windsurf/`, `.kiro/`) and to gitignore only `apm_modules/` as the rebuildable cache.

Renovate commits `updatedPackageFiles` + `updatedArtifacts`, so on a repo following that layout a bump branch carries the new ref and the new lockfile while every committed instruction file keeps the previous version's content. Nothing fails: the manifest says v2, `.github/instructions/*` still holds v1's rules, and the agent reads the stale files. The regenerated files are already in the working tree after the `apm install` call, so they are discarded rather than missing.

The set is derivable without new execution: APM writes only to the harness directories matching the manifest's `targets`, which the schema already parses.

**Test worth adding:** an `updateArtifacts` case on a manifest with `targets: [claude, copilot]` asserting the regenerated harness files appear in the returned artifacts alongside `apm.lock.yaml`, not only the lockfile.

MPV:
Vladimir: Thanks — really useful read. All addressed:

**Q1 (whitespace).** 👍 kept single-space to match the github-actions default.

**Q2 (git-tags digest).** You're right, "best-effort" was my mischaracterisation. Since `GitTagsDatasource.getDigest` resolves via `git ls-remote` and lookup filters out `newDigest === null` (holding the bump rather than rewriting to a digest-less ref), there's no silent-SHA-drop. Added a test asserting a git-tags host (`bitbucket.org/...`) SHA-pin resolves through the `git-tags` datasource with `currentDigest`/`currentValue`, locking in that the SHA is preserved on a bump (3f1b961). The `getDigest`-throws edge lives in shared lookup (it aborts, doesn't drop), so I left it out of the manager's unit tests — happy to add coverage there if you'd rather.

**Finding 1 (findPinnedTail duplicates).** Fixed (40b74dc): tag/`replaceString` recovery now consumes each manifest line at most once, in file order, so the same `owner/repo#<sha>` pinned in both `dependencies.apm` and `devDependencies.apm` maps to distinct lines instead of both resolving to the first. Test added for same-SHA-in-both-sections with differing tag comments.

**Finding 2 (`.chatmode.md`).** Fixed (4ee0c99): added `chatmode` to the virtual-file regex; test for a `*.chatmode.md` subpath boundary on a nested-group host.

**Tests you suggested** — all three are in: duplicate across sections, `.chatmode.md` virtual file, and the git-tags-host SHA-pin.

---

_Drafted with Claude Code; reviewed and edited by me before posting._

MPV:
Richard/Corentin: Good catch — thanks! `updateArtifacts` now captures everything `apm install` regenerates via `getRepoStatus()` (the copier/gomod pattern): the lockfile plus the deployed harness files (`.github/`, `.claude/`, …) as additions/deletions, not just `apm.lock.yaml`. `apm_modules/` is gitignored so it's excluded, and the manifest stays with `updatedPackageFiles`. Added a test on a `targets: [claude, copilot]`-style change asserting the harness files come back alongside the lockfile. (764ee64)

---

_Drafted with Claude Code; reviewed and edited by me before posting._

lib/modules/manager/apm/readme.md
```diff
@@ -0,0 +1,26 @@
...
+devDependencies:
+  apm:
+    - owner/repo#v1.2.3
+```
+
+Only entries that pin an exact `#<ref>` are updated.
+Entries without a `#<ref>` are skipped because there is no version to bump.
+
+APM also documents pinning to a commit SHA with the release tag kept as a trailing comment (`owner/repo#<sha> # v2.0.0`).
+With `pinDigests` enabled (part of the `config:best-practices` preset) Renovate keeps both the SHA and the tag comment current, the same way it does for `github-actions` (`uses: owner/action@<sha> # v4`).
```
RahulGautamSingh, line 22: I wonder if this is generated correctly for the docs or not.

RahulGautamSingh: Looks fine

image

RahulGautamSingh (approved):

rcollette:
> Richard/Corentin: Good catch — thanks! `updateArtifacts` now captures everything `apm install` regenerates via `getRepoStatus()` (the copier/gomod pattern): the lockfile plus the deployed harness files (`.github/`, `.claude/`, …) as additions/deletions, not just `apm.lock.yaml`. `apm_modules/` is gitignored so it's excluded, and the manifest stays with `updatedPackageFiles`. Added a test on a `targets: [claude, copilot]`-style change asserting the harness files come back alongside the lockfile. ([764ee64](https://github.com/renovatebot/renovate/commit/764ee64d5bd8a20f0e7475e2586c836742fdd67d))

We have some teams that are committing APM installed packages to their repo because it gives them a quick way to override if there is an issue (that will eventually feedback into our marketplace).    Ignoring apm_modules here... will that cause apm_modules to be out of sync in the scenario I described?

MPV:
Richard: Good question — and the scenario you describe actually works: `apm_modules/` won't go out of sync for those teams.

The important correction to my earlier wording: we don't hard-code an `apm_modules/` exclusion. `updateArtifacts` returns whatever `getRepoStatus()` reports, which is just `git status` under the hood — so what travels with the PR is decided by **your repo's own `.gitignore`**, not by a path filter in the manager.

- **Default case** (most APM users gitignore `apm_modules/`): git never reports it, so there's nothing to commit — it's purely a local cache. That's the case my earlier comment was describing, and "excluded" was the wrong word for it.
- **Your case** (teams that commit `apm_modules/`): the directory is tracked, so when the manager runs `apm install` it regenerates that tree on disk and git reports the changes as normal modifications / additions / deletions. The manager returns them as artifact changes **alongside** `apm.lock.yaml` and the harness files — so the committed `apm_modules/` is updated in the *same* PR as the bump and stays in sync.

Put differently: whether `apm_modules/` rides along is determined by the consumer repo's git tracking, exactly as it would be for someone running `apm install` locally — which is the behaviour you'd want for the override workflow you described. Track the directory and Renovate keeps it consistent; ignore it and Renovate treats it as a cache.

---

_Drafted with Claude Code; reviewed and edited by me before posting._

jamietanna:
A friendly reminder per [our contribution guidelines](https://github.com/renovatebot/renovate/blob/main/.github/contributing.md#ai-assistance-notice):

> If PR responses are being generated by an AI, disclose that as well.

MPV:
I hope this is now in a mergable state, pre-approves and all. 

Let me know if there's anything amiss and I'm happy to have a look. 🙏🏻

lib/modules/manager/apm/extract.ts
```diff
@@ -0,0 +1,278 @@
...
+  return {
+    datasource: GitTagsDatasource.id,
+    packageName: `https://${host}/${repoPath}`,
+  };
+}
+
+/** A full commit SHA (git sha1 or sha256). */
+const shaRegex = regEx(/^(?:[a-f0-9]{40}|[a-f0-9]{64})$/i);
+
+const newlineRegex = regEx(/\r?\n/);
```
secustor: Use https://github.com/renovatebot/renovate/blob/e0aa2cab3be86dbfd05403a96229af35e2a5ccd9/lib/util/regex.ts#L71

lib/modules/manager/apm/extract.ts
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  return {
+    datasource: GitTagsDatasource.id,
+    packageName: `https://${host}/${repoPath}`,
+  };
+}
+
+/** A full commit SHA (git sha1 or sha256). */
+const shaRegex = regEx(/^(?:[a-f0-9]{40}|[a-f0-9]{64})$/i);
```
secustor: Instead of this use https://github.com/renovatebot/renovate/blob/e0aa2cab3be86dbfd05403a96229af35e2a5ccd9/lib/util/schema-utils/git.ts#L10

lib/modules/manager/apm/extract.ts
```diff
@@ -0,0 +1,278 @@
...
+  };
+
+  if (!ref) {
+    // Unpinned dependency (no `#ref`) - nothing for Renovate to update.
+    return { ...base, skipReason: 'unspecified-version' };
+  }
+
+  // The optional host prefix is a hostname (so it contains a dot); git host
+  // owner names never do, which disambiguates the leading segment.
+  const segments = pathPart.split('/').filter(Boolean);
```
secustor:
```suggestion
  const segments = pathPart.split('/').filter(isTruthy);
```

secustor (commented):
else LGTM

secustor (approved):

PR #44132, opened 2026-06-21
feat(platform/azure): make work item type and states configurable

illay1994:
## Changes

Add the `azureWorkItem` config option so Renovate's Azure Boards issues (such as the Dependency Dashboard) can use a work item type and open/closed state names other than the hard-coded Issue/New/Closed.

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: # 
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [x] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #43786, opened 2026-06-04
fix(platform/azure): Remove 'System.State' field in create issue method

Sander0542:
Removed the operation to add a new state field to the issue.

## Changes

Remove the `System.State: New` field from the Create WorkItem API call. This is currently breaking the Renovate CLI run.

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: # 
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [x] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #44412, opened 2026-07-07
fix(platform/azure): resolve work item states from process

emanuelbesliu:
## Changes

Resolves the hardcoded Azure DevOps work item state that prevents the Dependency Dashboard (and any Renovate issue) from being created/closed on processes without `New`/`Closed` states.

- Autodetect the open/closed `Issue` state names from the project's process via `getWorkItemTypeStates`, grouped by state **category** (`Proposed`/`InProgress` = open; `Completed`/`Resolved`/`Removed` = closed) — categories are stable across processes even when display names differ.
- **Omit `System.State` on create** so Azure DevOps applies the work-item-type's default initial state (valid on every process).
- Classify existing issues open/closed by category instead of the literal `Closed`.
- Fall back to the previous `New`/`Closed` names when the states can't be read (e.g. older Azure DevOps Server). Resolved states are cached per repository.

## Context

Discussion: #43785. This supersedes the approaches in #43786 (drop state on create only) and #44132 (new config option) by fixing create, reopen, and close with zero configuration.

The failure this fixes on affected processes:

> The field 'State' contains the value 'New' that is not in the list of supported values

## Documentation

Added a "Dependency Dashboard work item state" section to the Azure platform readme describing the autodetect behavior and fallback.

## How I've tested my work

Added tests to `lib/modules/platform/azure/issue.spec.ts` (create omits state, category classification, close/reopen using resolved process states, InProgress-only fallback, no-name fallback, error fallback, caching). The full Azure platform suite passes (198 tests) at 100% coverage. `pnpm check` (prettier/biome/oxlint) and `pnpm type-check` pass.

## Disclosure

This change was written primarily with the assistance of GitHub Copilot CLI, then reviewed and verified by me.


lib/modules/platform/azure/issue.ts
```diff
@@ -1,18 +1,101 @@
...
+            )
+            .map((s: WorkItemStateColor) => s.name!);
+
+        const openNames = namesByCategory(openCategories);
+        const closedNames = namesByCategory(closedCategories);
+
+        // First open-category state, else the type's first state overall.
+        states.open = openNames[0] ?? stateColors[0].name ?? states.open;
+        if (closedNames.length) {
+          states.closed = closedNames[0];
```
RahulGautamSingh: `closedNames[0]` picks the first state across `Completed`, `Resolved`, and `Removed` in workflow order, so on Agile this resolves to `Resolved` rather than `Closed`. which is differnt from current behavior for Agile.

Use a Completed-category state for the close target, and fall back to the wider set only if none exists. Keep the full set for `closedNames` classification.

RahulGautamSingh (changes requested):

lib/modules/platform/azure/issue.ts
emanuelbesliu: Good catch, fixed in 2fe966e. The close target is now resolved from the `Completed` category first, falling back to the wider closed set (`Resolved`/`Removed`) only when a type has no `Completed` state. The full set is still used for `closedNames` classification, so already-closed items in any closed-category state are still recognised.

I validated this against our real Azure DevOps org (58 projects) for the `Issue` work item type:
- 56 projects only expose `Active` (InProgress) / `Closed` (Completed) — unaffected.
- One custom process orders states `New, Active, Resolved (Resolved), Closed (Completed), Duplicate (Removed), Rejected (Removed)` — here the old `closedNames[0]` picked **`Resolved`**, exactly the regression you described. With the fix it correctly resolves to `Closed`.

Added two tests covering that ordering: one asserting the close target is `Closed` (not `Resolved`), and one asserting all closed-category states are still classified as closed.

lib/modules/platform/azure/issue.ts
```diff
@@ -1,18 +1,111 @@
...
+
+        // First open-category state, else the type's first state overall.
+        states.open = openNames[0] ?? stateColors[0].name ?? states.open;
+        if (closedNames.length) {
+          // Prefer a `Completed` state as the close target so processes that
+          // order `Resolved` before `Closed` (e.g. some custom Agile
+          // inheritances) still close to `Closed`; fall back to the wider set
+          // only when no `Completed` state exists. Keep the full set for
+          // recognising already-closed items.
+          states.closed = completedNames[0] ?? closedNames[0];
```
RahulGautamSingh: Every closed-state test fixture includes a Completed-category state, so `completedNames[0]` is always defined and the `closedNames[0]` fallback is never exercised .. this is causing the coverage problem. You can update/add test to fix this

lib/modules/platform/azure/issue.ts
```diff
@@ -1,18 +1,111 @@
...
+
+    try {
+      const azureApiWit = await azureApi.workItemTrackingApi();
+      const stateColors = await azureApiWit.getWorkItemTypeStates(
+        this.config.project,
+        workItemType,
+      );
+
+      if (stateColors?.length) {
+        const namesByCategory = (categories: string[]): string[] =>
```
RahulGautamSingh: no arrow functions

lib/modules/platform/azure/readme.md
RahulGautamSingh, line 1: we try to avoid possesives as per the style guide.

rewrite these two:

"the project's process" and "the work item type's default initial state". 

to something like "the process for the project" / "the default initial state of the work item type"

RahulGautamSingh (changes requested):

lib/modules/platform/azure/issue.ts
emanuelbesliu: Fixed in e5aea9f. Added a test where the `Issue` type exposes a `Resolved`-category state but **no** `Completed` state, so `completedNames[0]` is undefined and the `closedNames[0]` fallback is exercised (close target resolves to `Resolved`). Patch coverage is back to 100% (branches 54/54) on the Azure issue suite.

lib/modules/platform/azure/issue.ts
emanuelbesliu: Fixed in e5aea9f. Extracted `namesByCategory` into a module-level function declaration (taking `stateColors` as a parameter) instead of an arrow function assigned to a variable.

lib/modules/platform/azure/readme.md
emanuelbesliu: Fixed in e5aea9f. Reworded to "the process for the project" and "the default initial state of the work item type".

emanuelbesliu:
@RahulGautamSingh thanks for the review — all three points are addressed in e5aea9f:

- **Coverage**: added a test with a `Resolved`-category state but no `Completed` state, exercising the `completedNames[0] ?? closedNames[0]` fallback. Azure issue suite is back to 100% patch coverage (branches 54/54).
- **Arrow function**: `namesByCategory` is now a module-level function declaration.
- **Readme**: reworded to avoid possessives.

`pnpm type-check`, oxlint, biome and prettier all pass locally. Ready for another look.

lib/modules/platform/azure/issue.ts
```diff
@@ -1,18 +1,119 @@
...
+// categories. Which concrete state name maps to a category depends on the
+// project's process (Basic: To Do/Doing/Done, Agile: New/Active/.../Closed,
+// custom inherited processes: anything). Categories are stable, so we resolve
+// the concrete names from them instead of hardcoding process-specific values.
+const openCategories = ['Proposed', 'InProgress'];
+const closedCategories = ['Completed', 'Resolved', 'Removed'];
+// Preferred category for the close target. A `Completed` state (e.g. Agile's
+// `Closed`) is the true "done" state; `Resolved`/`Removed` states also count as
+// closed but must not be picked over `Completed` when moving an item to closed.
+const completedCategories = ['Completed'];
```
RahulGautamSingh: single item array is a bit much for one lookup. a `stateColors.find(s => s.category === 'Completed')` gets you the same thing without the extra `namesByCategory` pass. not a big deal though.

lib/modules/platform/azure/issue.ts
```diff
@@ -1,18 +1,119 @@
...
+      if (stateColors?.length) {
+        const openNames = namesByCategory(stateColors, openCategories);
+        const completedNames = namesByCategory(
+          stateColors,
+          completedCategories,
+        );
+        const closedNames = namesByCategory(stateColors, closedCategories);
+
+        // First open-category state, else the type's first state overall.
+        states.open = openNames[0] ?? stateColors[0].name ?? states.open;
```
RahulGautamSingh, line 96: if a process somehow only has closed states, this picks a closed state name as the reopen target. it's tested and there's nothing better to fall back to, but worth a quick comment here so nobody later reads it as a bug.

lib/modules/platform/azure/issue.ts
```diff
@@ -222,7 +330,6 @@ export class IssueService {
             path: '/multilineFieldsFormat/System.Description',
             value: 'Markdown',
           },
-          { op: 'add', path: '/fields/System.State', value: 'New' },
         ],
         this.config.project,
         'Issue',
```
RahulGautamSingh: you added `workItemType` at the top but this still passes the `'Issue'` literal. use the const so it's not spelled out in two places.

RahulGautamSingh (commented):
nice, reads clean and the tests are thorough. couple of small things inline, none blocking.

lib/modules/platform/azure/issue.ts
emanuelbesliu: Done in 9c89c294d — replaced the single-element `namesByCategory` pass with `stateColors.find((s) => s.category === 'Completed' && s.name)` and use `completed?.name ?? closedNames[0]`. Dropped the now-unused `completedCategories` const.

lib/modules/platform/azure/issue.ts
emanuelbesliu: Added a comment in 9c89c294d noting the open-state fallback can land on a closed state when a process exposes only closed states, and that this is intentional (no better reopen target) rather than a bug.

lib/modules/platform/azure/issue.ts
emanuelbesliu: Fixed in 9c89c294d — the create-issue call, the WIQL filter and the field patch now all reuse the `workItemType` const instead of repeating the `'Issue'` literal.

RahulGautamSingh (approved):

ghost (approved):

lib/modules/platform/azure/issue.ts
```diff
@@ -1,18 +1,120 @@
...
+  /**
+   * Resolve the concrete open/closed state names for the `Issue` work item type
+   * from its process, so Renovate does not depend on hardcoded state names that
+   * only exist in some Azure DevOps processes. Falls back to the historical
+   * `New`/`Closed` values if the states cannot be fetched. Cached per instance.
+   */
+  private async getWorkItemStates(): Promise<WorkItemStates> {
+    if (this.workItemStates) {
+      return this.workItemStates;
+    }
```
viceice: this function could be called twice because it's async and then this condition is not yet true.

use `Lazy` or `Mutex`

https://github.com/renovatebot/renovate/blob/71bb1e2635db0c04aa25c963b1437fc6b12a6a80/lib/util/lazy.ts#L11

https://github.com/renovatebot/renovate/blob/71bb1e2635db0c04aa25c963b1437fc6b12a6a80/lib/util/mutex.ts#L20

emanuelbesliu: Good catch — fixed in 6ba33fd92. Wrapped the resolution in `Lazy<Promise<WorkItemStates>>` (matching the `lazyLoadPackageJson` pattern in `post-update/utils.ts`), so the executor runs once and concurrent callers share the same promise instead of each firing a duplicate `getWorkItemTypeStates` request. The "resolves only once" test still passes.

viceice (approved):

PR #44422, opened 2026-07-07
feat(manager): add nub package manager

colinhacks:
## Changes

Adds a `nub` package-manager manager.

Nub is a Node package manager published as [`@nubjs/nub`](https://www.npmjs.com/package/@nubjs/nub) on npm. It has its own lockfile, `nub.lock`, which is byte-compatible with the pnpm-lock v9 format. This manager gives nub-based projects lockfile-aware update PRs without keeping a foreign lockfile.

The manager is modeled on the existing `bun` manager:

- Detects `nub.lock`, resolves the sibling `package.json`, and extracts dependencies through the shared npm extractor (`extractPackageJson`). The datasource stays `npm`.
- Regenerates the lockfile in `updateArtifacts` by running `nub install --lockfile-only` (with `--ignore-scripts` unless scripts are allowed), then diffs the old and new `nub.lock`.
- Sets `supersedesManagers: ['npm']` so nub takes over from the npm manager only when `nub.lock` is present.
- Registers nub in the Containerbase registry (`toolDefinitions` + `allToolConfig.nub` → npm `@nubjs/nub`), so the default `binarySource=install` runtime can provision it. This depends on the companion Containerbase tool: containerbase/base#7054.

## Context

## AI assistance disclosure

Code, tests, and the manager readme were written with substantive assistance from an AI coding agent (Claude).

Adds `lib/modules/manager/nub/readme.md`, which is auto-rendered into the docs site.

`pnpm vitest run lib/modules/manager/nub --coverage` passes with 100% statement/branch/function/line coverage. The manager, metadata, and fingerprint validators, the `lib/util/exec` and `lib/config` specs, and `pnpm test-schema` all pass.


lib/modules/manager/nub/extract.ts
```diff
@@ -0,0 +1,104 @@
...
+    const packageFile = getSiblingFileName(lockFile, 'package.json');
+    const res = await processPackageFile(packageFile, config);
+    if (res) {
+      packageFiles.push({ ...res, lockFiles: [lockFile] });
+    }
+    // Check if package.json contains workspaces
+    let workspaces = res?.managerData?.workspaces;
+
+    // nub also accepts the object form `workspaces: { packages: [...] }`.
+    if (typeof workspaces === 'object' && 'packages' in workspaces) {
```
RahulGautamSingh: `typeof workspaces === 'object'` is also true for `null`, and `managerData.workspaces` comes straight from the parsed manifest. A `"workspaces": null` value makes `'packages' in workspaces` throw and crashes extract. Use `is.plainObject(workspaces)`, which also matches the prefer-`is` guideline.

lib/modules/manager/nub/extract.ts
```diff
@@ -0,0 +1,104 @@
...
+  );
+}
+
+export async function processPackageFile(
+  packageFile: string,
+  config: ExtractConfig,
+): Promise<PackageFile | null> {
+  const fileContent = await readLocalFile(packageFile, 'utf8');
+  if (!fileContent) {
+    logger.warn({ fileName: packageFile }, 'Could not read file content');
```
RahulGautamSingh: This warns when a `nub.lock` has no readable sibling `package.json`, which is an expected case and will be noisy. Use `logger.debug` here, as the npm manager does.

lib/modules/manager/nub/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+  pwd: string,
+  fileName: string,
+  workspaces: string[],
+): boolean {
+  if (!fileName.startsWith(pwd)) {
+    return false;
+  }
+  const relativeFile = upath
+    .relative(pwd, fileName)
+    .replace(/\/package\.json$/, '');
```
RahulGautamSingh: Use the `regEx()` utility from `lib/util/regex.ts` instead of a regex literal.

lib/modules/manager/nub/artifacts.ts
```diff
@@ -0,0 +1,117 @@
...
+      cmd += ' --ignore-scripts';
+    }
+
+    const execOptions: ExecOptions = {
+      cwdFile: lockFileName,
+      docker: {},
+      toolConstraints: [
+        {
+          toolName: 'nub',
+          constraint: updateArtifact?.config?.constraints?.nub,
```
RahulGautamSingh: `updateArtifact` and `config` are already in scope and always defined. Simplify to `config.constraints?.nub`.

lib/modules/manager/nub/readme.md
```diff
@@ -0,0 +1,6 @@
+Used for updating projects that use [nub](https://nubjs.com) as their package manager.
+nub is a tool for JavaScript projects and therefore an alternative to managers like npm, pnpm and Yarn.
+
+nub's `nub.lock` is the pnpm lockfile v9 format under nub's own filename, so dependencies resolve through the same npm datasource as the `npm` and `bun` managers.
```
RahulGautamSingh: Avoid possessives like `nub's`. Rewrite, for example: the `nub.lock` file uses the pnpm lockfile v9 format under a nub-specific name.

lib/modules/manager/nub/artifacts.spec.ts
```diff
@@ -0,0 +1,285 @@
...
+      ];
+      expect(await updateArtifacts(updateArtifact)).toBeNull();
+    });
+
+    it('returns null if lock content unchanged', async () => {
+      updateArtifact.updatedDeps = [
+        { manager: 'nub', lockFiles: ['nub.lock'] },
+      ];
+      const oldLock = Buffer.from('old');
+      fs.readFile.mockResolvedValueOnce(oldLock as never);
```
RahulGautamSingh: `as never` bypasses the type system. Prefer typing the mock value or `partial<T>()` from `test/util`. The table-driven test lower down also hand-rolls a `for` loop over `testCases`; `it.each` fits better.

colinhacks:
All six addressed in `6d9ea8169`:

1. Replaced `typeof workspaces === 'object'` with `isPlainObject(workspaces)` in `extract.ts` (named import, matching the manager tree's convention), so a `"workspaces": null` manifest no longer throws on the `'packages' in` check. Added a test covering the null case.
2. Demoted the missing sibling `package.json` read to `logger.debug`.
3. Wrapped the `package.json`-suffix regex in `regEx()`.
4. Simplified the tool constraint to `config.constraints?.nub`.
5. Reworded the readme to drop the possessive.
6. Reworked `artifacts.spec.ts` to mock the fs util instead of `fs-extra`, which removes the `as never` casts, and the command-execution cases are now `it.each`. Lockfile reads resolve to a `Buffer`; since `readLocalFile` is overloaded, they go through one typed handle rather than a cast.

Green locally: `pnpm jest lib/modules/manager/nub` (31 tests), `tsc --noEmit`, oxlint, prettier, and biome.

lib/modules/manager/nub/readme.md
```diff
@@ -0,0 +1,6 @@
+Used for updating projects that use [nub](https://nubjs.com) as their package manager.
+nub is a tool for JavaScript projects and therefore an alternative to managers like npm, pnpm and Yarn.
```
RahulGautamSingh, line 2:
```suggestion
nub is a tool for JavaScript projects and therefore an alternative to managers like npm, pnpm and yarn.
```

consistent casing

lib/modules/manager/nub/readme.md
```diff
@@ -0,0 +1,6 @@
...
+
+The `nub.lock` file uses the pnpm lockfile v9 format under a nub-specific name, so dependencies resolve through the same npm datasource as the `npm` and `bun` managers.
+
+If a `package.json` is found to be part of `nub` manager results then the same file will be excluded from the `npm` manager results unless an npm/pnpm/Yarn lock file is also found.
```
RahulGautamSingh, line 6:
```suggestion
If a `package.json` is found to be part of `nub` manager results then the same file will be excluded from the `npm` manager results unless an npm/pnpm/yarn lock file is also found.
```

RahulGautamSingh (commented):
Else LGTM

viceice (commented):
- blocked by containerbase/base#7054

PR #44425, opened 2026-07-07
fix(gerrit): use max available Code-Review vote value for autoApprove

QuentinBtd:
## Changes

The Gerrit platform's `autoApprove` feature hardcoded `Code-Review+2` as the push option used to auto-vote on a newly created/updated change (`lib/modules/platform/gerrit/scm.ts`). This breaks `autoApprove` (and therefore `automerge`) on any Gerrit project whose `Code-Review` label doesn't define a `+2` value (for example, projects that only allow `-1..+1`), since Gerrit rejects/ignores a vote outside of the label's configured range.

This PR makes the auto-approve vote value dynamic:

- `configureScm()` now also receives the project's label definitions (`Record<string, GerritLabelTypeInfo>`), already fetched in `initRepo()` via `client.getProjectInfo()`.
- A new `getAutoApproveLabelValue()` helper computes the maximum value actually available for the `Code-Review` label on the project, reusing the existing `mapBranchStatusToLabel('green', label)` logic (the same mechanism already used for generic `statusCheckNames` label voting).
- If no `Code-Review` label definition is known, it falls back to `+2` (previous hardcoded behavior), so this change is backwards compatible for projects that do support `+2`.

`pushForReview()` now pushes `label=Code-Review+${getAutoApproveLabelValue()}` instead of the hardcoded `label=Code-Review+2`.

## Context

## AI assistance disclosure

This PR (code, tests, and this description) was written with the assistance of OpenCode using the `claude-sonnet-5` model (`github-copilot/claude-sonnet-5`).

The public repository: private Gerrit repository


RahulGautamSingh:
@felipecrs Could you take a look please?

lib/modules/platform/gerrit/scm.ts
```diff
@@ -31,7 +56,7 @@ export async function pushForReview(options: {
 }): Promise<boolean> {
   const pushOptions = ['notify=NONE', 'ready'];
   if (options.autoApprove) {
-    pushOptions.push('label=Code-Review+2');
+    pushOptions.push(`label=Code-Review+${getAutoApproveLabelValue()}`);
   }
```
felipecrs: Nit:

```suggestion
    const value = getAutoApproveLabelValue();
    pushOptions.push(`label=Code-Review+${value}`);
}
```

lib/modules/platform/gerrit/scm.ts
```diff
@@ -9,14 +9,39 @@ import type { LongCommitSha } from '../../../util/schema-utils/git.ts';
...
+}
+
+/**
+ * Determines the numeric vote to apply on the "Code-Review" label when
+ * `autoApprove` is requested, based on the project's actual label
+ * configuration instead of assuming "+2" is always available (some Gerrit
+ * projects only define values up to "+1").
+ */
+function getAutoApproveLabelValue(): number {
+  const codeReviewLabel = projectLabels['Code-Review'];
```
felipecrs: If we are defining a constant for `2`, then perhaps we should define one for `Code-Review` too.

Or we don't define for any.

lib/modules/platform/gerrit/scm.ts
```diff
@@ -9,14 +9,39 @@ import type { LongCommitSha } from '../../../util/schema-utils/git.ts';
 import { isLongCommitSha } from '../../../util/schema-utils/git.ts';
 import { DefaultGitScm } from '../default-scm.ts';
 import { client } from './client.ts';
+import type { GerritLabelTypeInfo } from './schema.ts';
 import type { GerritFindPRConfig } from './types.ts';
-import { convertGerritDateToISO } from './utils.ts';
+import { convertGerritDateToISO, mapBranchStatusToLabel } from './utils.ts';
+
+/** Fallback used when the project's Code-Review label definition is unknown
```
felipecrs: Under which conditions the Code-Review label would not be available? In case users removed it in the project config?

If that's the case, then instead of falling back to voting Code-Review +2, we should actually issue a warning and not vote anything, to avoid failing the entire push operation.

felipecrs (changes requested):
Thanks for the fix. I was wondering when something like this would happen. Next step would be to make the label name configurable, I guess.

lib/modules/platform/gerrit/scm.ts
felipecrs: After thinking more about this, I really believe we should avoid `label=Code-Review` entirely if the label is missing or unknown, and log a warning instead. Otherwise, we risk failing the whole push.

lib/modules/platform/gerrit/scm.ts
QuentinBtd: Good point.

Now defined for both

lib/modules/platform/gerrit/scm.ts
QuentinBtd: I agree with you

Now implemented

In the futur I plan to add an option to customize the label where apply `+1` (or `+2`), but I need to understand more how Renovate works 😄

felipecrs: FYI https://github.com/renovatebot/renovate/discussions/44531.

lib/modules/platform/gerrit/scm.spec.ts
```diff
@@ -573,7 +620,7 @@ describe('modules/platform/gerrit/scm', () => {
       expect(git.pushCommit).not.toHaveBeenCalled();
     });
 
-    it('commitAndPush() - existing change with new changes - auto-approve', async () => {
+    it('commitAndPush() - existing change with new changes - auto-approve without known Code-Review label skips vote', async () => {
```
felipecrs: You already have an exclusive test above for when ` Code-Review` is unset.

I think you should patch this test so that it remains passing `label=Code-Review+2`.

lib/modules/platform/gerrit/scm.ts
```diff
@@ -9,14 +9,44 @@ import type { LongCommitSha } from '../../../util/schema-utils/git.ts';
...
+
+/**
+ * Determines the numeric vote to apply on the "Code-Review" label when
+ * `autoApprove` is requested, based on the project's actual label
+ * configuration instead of assuming "+2" is always available (some Gerrit
+ * projects only define values up to "+1").
+ *
+ * Returns `null` if the project doesn't define a "Code-Review" label at all
+ * (e.g. it was removed from the project config), so that callers can skip
+ * voting instead of risking the whole push being rejected by Gerrit.
```
felipecrs: This is probably more verbose than it needs to be, but ok.

felipecrs (changes requested):
Otherwise LGTM. I tested this patch via https://github.com/renovatebot/renovate/pull/42828 and it's working as expected.

lib/modules/platform/gerrit/scm.spec.ts
QuentinBtd: Good catch, fixed.

This test now configures a Code-Review label with a +2 value so it still exercises the "vote applied" path, and the "label missing" scenario stays exclusively covered by the other dedicated test.

felipecrs (approved):
LGTM.

QuentinBtd:
@felipecrs I rebased the branch and resolved the conflicts. Could you please review it again? 🙏

felipecrs (approved):

QuentinBtd:
Up @viceice 🙏

viceice (commented):
looks good. 

@felipecrs can you review please?

QuentinBtd:
> **[felipecrs](/felipecrs) ** approved these changes

[felipecrs](https://github.com/felipecrs) approved these changes [last week](https://github.com/renovatebot/renovate/pull/44425#pullrequestreview-4729078305) ;)

felipecrs (approved):
LGTM

lib/modules/platform/gerrit/scm.ts
```diff
@@ -17,8 +21,30 @@ import { client } from './client.ts';
  */
 
 let repository: string;
-export function configureScm(repo: string): void {
+let projectLabels: Record<string, GerritLabelTypeInfo> = {};
```
RahulGautamSingh: nit:

extract this type .. Record<string, GerritLabelTypeInfo>

RahulGautamSingh (commented):
lgtm

viceice (approved):

lib/modules/platform/gerrit/scm.ts
felipecrs: @QuentinBtd please mark this as resolved.

PR #44452, opened 2026-07-09
fix(config): normalize migrated package rule matchers

jonathanmorley:
## Changes

This addresses https://github.com/renovatebot/renovate/discussions/43936.

In summary, the current matcher normalization causes some configurations to fail when evaluated against the recently tightened matcher validations.
This changes the normalization of matchers to comply with the new requirements.

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/config/migrate-validate.spec.ts
+++ b/lib/config/migrate-validate.spec.ts
@@ -53,6 +53,31 @@ describe('config/migrate-validate', () => {
       });
     });
 
+    it('validates a legacy package match-all combined with an exclusion', async () => {
+      const input = {
+        packageRules: [
+          {
+            packagePatterns: ['*'],
+            excludePackageNames: ['some-package'],
+            automerge: true,
+          },
+        ],
+      } as RenovateConfig;
+
+      const res = await migrateAndValidate(config, input);
+
+      expect(res).toEqual({
+        packageRules: [
+          {
+            automerge: true,
+            matchPackageNames: ['!some-package'],
+          },
+        ],
+        errors: [],
+        warnings: [],
+      });
+    });
+
     it('handles invalid', async () => {
       // @ts-expect-error -- invalid option
       const input: RenovateConfig = { foo: 'none' };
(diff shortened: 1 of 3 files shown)


RahulGautamSingh (commented):
Nice fix on the core bug, but I think the normalization pass is over-scoped.

The real defect is narrow: 

The old migration wrapped match-all patterns as regexes (`/*/`, `!/*/`), and validation rejects those because they fail to compile. 

The two-line fix in `migratePattern` / `migrateExcludePattern` (emit `*` and `!**` instead) resolves that completely, since both are globs and skip the regex check.

Everything in `normalizeRegexOrGlobMatchers` / `normalizeMatchers` and the supporting Sets and predicates only collapses combinations that already pass validation:

```
['*', '!/foo/'] to ['!/foo/']
['**', 'foo'] to ['**']
['!**', 'foo'] to ['!**']
```

None of those trigger a validation error, so none of it is needed to fix the reported issue.

jonathanmorley:
I think there are two migration failures at play here, that are potentially getting conflated:

1. Legacy match-all patterns can become invalid regex tokens, such as `"*"` becoming `"/*/"`. The `migratePattern` / `migrateExcludePattern` changes fix this by emitting `"*"` and `"!**"`.
2. Migration can produce individually valid matchers that form an invalid array. The original reproduction is:
```
{
  "packagePatterns": ["*"],
  "excludePackagePatterns": ["some-package"]
}
```
Without normalization, that becomes:
```
{
  "matchPackageNames": ["*", "!/some-package/"]
}
```
`regex-glob-matchers.ts` rejects this because it contains `*` alongside another matcher.

RahulGautamSingh:
Can we fix this by adding a helper function which after migration checks the failing case ie. that * pattern is not combined with others?

The current solution looks very bloated to me, and I don;t want it to be a maintenance issue in future.

jonathanmorley:
This should be ready for a re-review. It has the requested helper function, called after all the config merges are completed. There is a regression test covering the need for it to run after merges, rather than during.

RahulGautamSingh (approved):

jonathanmorley:
@viceice how can we keep this PR moving?

PR #44469, opened 2026-07-10
fix(cache): send TLS SNI when connecting to rediss:// package cache

### LINKED ISSUE #44461: fix(cache/redis): set servername for rediss:// so SNI-routed endpoints connect
### Discussed in https://github.com/renovatebot/renovate/discussions/44460

<sup>Originally posted by **lauraseidler** July 10, 2026</sup>

Self-hosted Renovate CLI

GitHub.com

43.251.3

With `RENOVATE_REDIS_URL=rediss://user:pass@privatelink-valkey-...aivencloud.com:9706/1`, every run aborts during Redis cache init:

```
{"err":{"code":"ECONNRESET","host":"privatelink-valkey-...aivencloud.com","port":9706, "message":"Client network socket disconnected before secure TLS connection was established", ... "msg":"Renovate is exiting with a non-zero code due to the following logged errors"}
```

The endpoint itself is healthy: a Go client (`valkey-go`, which sets ServerName from the URL host) connects to the same host:port from the same network without issue, and `valkey-cli --tls` fails with the same mid-handshake EOF unless `--sni` is passed.

**Root cause**

Aiven privatelink (and similar setups) exposes multiple hostnames on one shared IP:port and routes connections by TLS SNI; handshakes without SNI are dropped. Renovate's cache client never sends SNI:

  - Node's automatic servername-from-host behavior lives in the https agent, not in raw `tls.connect` — verified empirically on Node v22.18: `tls.connect({host: 'valkey.example.com', ...})` produces a ClientHello with no SNI extension (server-side
  SNICallback never fires, `socket.servername === false`).
  - `@redis/client` 5.12.1 passes its socket options straight into `tls.connect` without setting servername (`socket.ts`), and Renovate's createClient call in `lib/util/cache/package/impl/redis.ts` doesn't set it either — so no SNI is ever sent.
  - Setting socket: `{ servername: '<hostname>' }` explicitly makes `@redis/client` 5.12.1 send SNI and the handshake succeed (verified against a local TLS server with an SNI-logging SNICallback).

Most managed Redis offerings (ElastiCache classic, Azure Cache) give each instance a dedicated endpoint, so the missing SNI goes unnoticed — it only bites SNI-routed frontends, which is presumably why this hasn't come up much.

**Proposed fix**

In `PackageCacheRedis.create`, derive `servername` from the parsed URL for `rediss://` schemes:

```
const parsedUrl = parseUrl(rewrittenUrl);
const config = {
  url: rewrittenUrl,
  socket: {
    reconnectStrategy: (retries: number) => Math.min(retries * 100, 3000),
    ...(parsedUrl?.protocol === 'rediss:' && !isIP(parsedUrl.hostname)
      ? { servername: parsedUrl.hostname }
      : {}),
  },
  pingInterval: 30000,
};
```

This matches what every HTTPS client already does, is a no-op for existing non-TLS/dedicated-endpoint users, and needs no new configuration option. The same applies to the createCluster root nodes. Happy to submit a PR if the approach sounds right.

### Logs (if relevant)

```
{"err":{"code":"ECONNRESET","host":"privatelink-valkey-...aivencloud.com","port":9706, "message":"Client network socket disconnected before secure TLS connection was established", ... "msg":"Renovate is exiting with a non-zero code due to the following logged errors"}

```

</details>
</div>

lauraseidler:
## Changes

On Node.js, `tls.connect({host})` does not automatically set the SNI servername from the host. `@redis/client` spreads caller socket options directly into `tls.connect` without injecting `servername`, so a plain `createClient({url: 'rediss://...'})` sends no SNI.

On SNI-routed endpoints (e.g. Aiven Privatelink, where many hostnames share a single IP:port), the TLS handshake is rejected and the client gets `ECONNRESET before secure TLS connection was established`.

Fix: set `socket.servername` to the URL hostname when the scheme is `rediss:` and the hostname is not an IP literal (RFC 6066 forbids IP addresses in SNI). Bracketed IPv6 literals (`[::1]`) are correctly excluded by stripping the brackets before the `isIP` check.

The `parsedUrl` call is hoisted above the config object so the cluster path can reuse it without a second parse. Placing `servername` in `config.socket` means it is also passed to cluster `rootNodes`; note that dynamically-discovered cluster nodes may still lack per-node SNI (out of scope for this fix).

## Context

- [x] This closes an existing Issue, Closes: #44461 

## AI assistance disclosure

The public repository: not a public repository

### DIFF
--- a/lib/util/cache/package/impl/redis.spec.ts
+++ b/lib/util/cache/package/impl/redis.spec.ts
@@ -72,6 +72,46 @@ describe('util/cache/package/impl/redis', () => {
         );
       });
 
+      it('sets socket.servername for rediss:// with a hostname', async () => {
+        await PackageCacheRedis.create('rediss://cache.example.com:6380', '');
+
+        expect(createClient).toHaveBeenCalledWith(
+          expect.objectContaining({
+            socket: expect.objectContaining({
+              servername: 'cache.example.com',
+            }),
+          }),
+        );
+      });
+
+      it('does not set socket.servername for rediss:// with an IPv4 address', async () => {
+        await PackageCacheRedis.create('rediss://127.0.0.1:6380', '');
+
+        const [config] = vi.mocked(createClient).mock.calls[0];
+        expect(config?.socket).not.toHaveProperty('servername');
+      });
+
+      it('does not set socket.servername for rediss:// with an IPv6 address', async () => {
+        await PackageCacheRedis.create('rediss://[::1]:6380', '');
+
+        const [config] = vi.mocked(createClient).mock.calls[0];
+        expect(config?.socket).not.toHaveProperty('servername');
+      });
+
+      it('does not set socket.servername for redis:// (non-TLS)', async () => {
+        await PackageCacheRedis.create('redis://cache.example.com:6379', '');
+
+        const [config] = vi.mocked(createClient).mock.calls[0];
+        expect(config?.socket).not.toHaveProperty('servername');
+      });
+
+      it('does not set socket.servername when URL cannot be parsed', async () => {
+        await PackageCacheRedis.create('not-a-valid-url', '');
+
+        const [config] = vi.mocked(createClient).mock.calls[0];
+        expect(config?.socket).not.toHaveProperty('servername');
+      });
+
       it('initializes cluster client', async () => {
         await PackageCacheRedis.create('redis+cluster://host', '');
 
--- a/lib/util/cache/package/impl/redis.ts
+++ b/lib/util/cache/package/impl/redis.ts
@@ -1,3 +1,4 @@
+import { isIP } from 'node:net';
 import type { RedisClusterOptions } from '@redis/client';
 import { RESP_TYPES, createClient, createCluster } from '@redis/client';
 import { logger } from '../../../../logger/index.ts';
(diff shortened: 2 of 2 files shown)


lib/util/cache/package/impl/redis.ts
```diff
@@ -29,10 +30,15 @@ export class PackageCacheRedis extends PackageCacheBase {
     const rewrittenUrl = normalizeRedisUrl(url);
     const clusteredMode = rewrittenUrl !== url;
 
+    const parsedUrl = parseUrl(rewrittenUrl);
+    const rawHostname = parsedUrl?.hostname.replace(/^\[(.+)\]$/, '$1') ?? '';
```
RahulGautamSingh, line 34:
```suggestion
    const rawHostname = parsedUrl?.hostname.replace(regEx(/^\[(.+)\]$/), '$1') ?? '';
```

RahulGautamSingh (changes requested):
Else LGTM

grakshith (approved):

viceice (commented):
looks good when suggestion is merged

lib/util/cache/package/impl/redis.ts
```diff
@@ -29,10 +30,16 @@ export class PackageCacheRedis extends PackageCacheBase {
...
+    const parsedUrl = parseUrl(rewrittenUrl);
+    const rawHostname =
+      parsedUrl?.hostname.replace(regEx(/^\[(.+)\]$/), '$1') ?? '';
     const config = {
       url: rewrittenUrl,
       socket: {
         reconnectStrategy: (retries: number) => Math.min(retries * 100, 3000),
+        ...(parsedUrl?.protocol === 'rediss:' && !isIP(rawHostname)
+          ? { servername: parsedUrl.hostname }
+          : {}),
```
RahulGautamSingh, line 42: ...(parsedUrl?.protocol === 'rediss:' &&
  !isIP(rawHostname) && { servername: rawHostname }),

lib/util/cache/package/impl/redis.spec.ts
RahulGautamSingh, line 1: `expect(undefined).not.toHaveProperty('servername')` passes, so these four tests would still go green if `createClient` were never called. 

The file already uses mock.calls[0][0]! elsewhere; use ! here too.


----

And, a missing test: `rediss+cluster://`

Only `redis+cluster:// `is covered. TLS-plus-cluster is the common managed-Redis setup and it goes down the rootNodes: [config] path, so a test asserting servername lands in rootNodes[0].socket is needed.

RahulGautamSingh (changes requested):

viceice (commented):
see open comment

PR #44502, opened 2026-07-11
fix(platform/gitlab): retry MR creation on transient source_branch race.

### LINKED ISSUE #44498: GitLab: MR creation race `400 source_branch does not exist` immediately after a successful push (no retry)
### Discussed in https://github.com/renovatebot/renovate/discussions/44231

<sup>Originally posted by **vincentjames501** June 25, 2026</sup>

Self-hosted Renovate CLI

GitLab (.com or self-hosted)

43.242.2

### Describe the bug

When Renovate pushes a branch and immediately creates the merge request, GitLab
occasionally rejects the MR-create call with:

    POST /api/v4/projects/:id/merge_requests
    => 400  {"message": {"source_branch": ["does not exist"]}}

even though the preceding `git push` returned success and Renovate logged
`INFO: Branch created`. The MR `POST` is sent only tens of milliseconds after the
push completes, before GitLab (Gitaly) has made the new ref visible to the API —
a classic eventual-consistency race.

The request is **not retried** (`retryCount=0`), so Renovate logs
`Pull request creation error` and continues, leaving an **orphaned branch with no
merge request**. The branch is created on the remote and self-heals on the *next*
Renovate run (the ref is visible by then), confirming the failure is purely transient.

This is intermittent and load/timing dependent: in a single run, 10 of 24 branches
failed this way while the identical branch names succeeded in other repositories
processed earlier in the same run.

We only wish to run this biweekly and not everyday.

### Expected behavior
A `400 {"source_branch": ["does not exist"]}` immediately after a successful push
should be treated as a transient/eventual-consistency error and **retried a few
times with a short backoff** before being surfaced as a failure, so the MR is
created in the same run instead of leaving an orphaned branch until the next run.

### Proposed solution
In the GitLab platform's `createPr`, detect this specific transient response
(HTTP 400 with body `{"message":{"source_branch":["does not exist"]}}`) and retry
the MR creation with a bounded exponential backoff (e.g. 3 attempts, ~0.5–2 s).
Optionally expose the attempt count / delay as a self-hosted config option. A
generic retry is not required — scoping it to this known GitLab response keeps it safe.

### Additional context
- `gitUrl: 'ssh'` and `prCreation: 'not-pending'` can reduce the window but don't
  address the root race for `prCreation: 'immediate'` (the default).
- Affects self-managed GitLab where Gitaly ref propagation lags the HTTPS push.

### Logs (if relevant)

<details><summary>logs</summary>

2026-06-25T07:11:34.090Z DEBUG: git push
2026-06-25T07:11:34.139Z INFO:  Branch created
2026-06-25T07:11:34.146Z DEBUG: Creating Merge Request: Update dependency internal-lib to v1
2026-06-25T07:11:34.215Z DEBUG: POST .../merge_requests = (code=ERR_NON_2XX_3XX_RESPONSE, statusCode=400 retryCount=0, duration=67)
2026-06-25T07:11:34.216Z        "statusCode": 400,
2026-06-25T07:11:34.216Z        "body": {"message": {"source_branch": ["does not exist"]}}
2026-06-25T07:11:34.217Z DEBUG: Pull request creation error
(~56 ms between `git push` returning and the MR POST; whole window ~125 ms)
</details></div>

vincentjames501:
## Changes

GitLab occasionally rejects a merge request POST with `400 {"message":{"source_branch":["does not exist"]}}` immediately after a branch push succeeds, because the freshly pushed ref is not yet visible to the API due to Gitaly eventual consistency. The request was not retried, leaving an orphaned branch until the next run.

Extend the existing `calculateRetryDelay` hook (which already retries transient `409 Resource lock` POST errors) to also retry this `400 source_branch does not exist` case, reusing got's bounded exponential backoff.

Closes #44498

Note: Original suggestion was to wrap with `wrapWithRetry`. This just extends the existing precedent but I'm happy to switch to `wrapWithRetry` if preferred.

## Context

- [ X ] This closes an existing Issue, Closes: #44498

## AI assistance disclosure

Did you use AI tools to create any part of this pull request? Yes

- [ X ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).

- [ X ] No documentation update is required

- [ X ] Newly added/modified unit tests, or

### DIFF
--- a/lib/util/http/gitlab.spec.ts
+++ b/lib/util/http/gitlab.spec.ts
@@ -201,4 +201,68 @@ describe('util/http/gitlab', () => {
       );
     });
   });
+
+  describe('handles 400 source_branch errors', () => {
+    let NODE_ENV: string | undefined;
+
+    beforeAll(() => {
+      // Unset NODE_ENV so that we can test the retry logic
+      NODE_ENV = process.env.NODE_ENV;
+      delete process.env.NODE_ENV;
+    });
+
+    afterAll(() => {
+      process.env.NODE_ENV = NODE_ENV;
+    });
+
+    it('retries the request when source branch does not yet exist', async () => {
+      const body = { message: { source_branch: ['does not exist'] } };
+      httpMock.scope(gitlabApiHost).post('/api/v4/some-url').reply(400, body);
+      httpMock.scope(gitlabApiHost).post('/api/v4/some-url').reply(200, {});
+      const res = await gitlabApi.postJson('some-url', {});
+      expect(res.statusCode).toBe(200);
+    });
+
+    it('does not retry more than twice when source branch does not exist', async () => {
+      const body = { message: { source_branch: ['does not exist'] } };
+      httpMock.scope(gitlabApiHost).post('/api/v4/some-url').reply(400, body);
+      httpMock.scope(gitlabApiHost).post('/api/v4/some-url').reply(400, body);
+      httpMock.scope(gitlabApiHost).post('/api/v4/some-url').reply(400, body);
+      await expect(gitlabApi.postJson('some-url', {})).rejects.toThrow(
+        HTTPError,
+      );
+    });
+
+    it('does not retry for unrelated 400 errors', async () => {
+      const body = { message: { base: ['Another error'] } };
+      httpMock.scope(gitlabApiHost).post('/api/v4/some-url').reply(400, body);
+      await expect(gitlabApi.postJson('some-url', {})).rejects.toThrow(
+        HTTPError,
+      );
+    });
+
+    it('does not retry for other source_branch 400 errors', async () => {
+      const body = { message: { source_branch: ['is invalid'] } };
+      httpMock.scope(gitlabApiHost).post('/api/v4/some-url').reply(400, body);
+      await expect(gitlabApi.postJson('some-url', {})).rejects.toThrow(
+        HTTPError,
+      );
+    });
+
+    it('does not retry other POST error codes', async () => {
+      const body = { message: 'Forbidden' };
+      httpMock.scope(gitlabApiHost).post('/api/v4/some-url').reply(403, body);
+      await expect(gitlabApi.postJson('some-url', {})).rejects.toThrow(
+        HTTPError,
+      );
+    });
+
+    it('does not retry non-POST requests', async () => {
+      const body = { message: { source_branch: ['does not exist'] } };
+      httpMock.scope(gitlabApiHost).get('/api/v4/some-url').reply(400, body);
+      await expect(gitlabApi.getJsonUnchecked('some-url')).rejects.toThrow(
+        HTTPError,
+      );
+    });
+  });
 });
--- a/lib/util/http/gitlab.ts
+++ b/lib/util/http/gitlab.ts
@@ -119,8 +119,7 @@ export class GitlabHttp extends HttpBase<GitlabHttpOptions> {
     if (
       attemptCount <= retryOptions.limit &&
       error.options.method === 'POST' &&
-      error.response?.statusCode === 409 &&
-      error.response.rawBody.toString().includes('Resource lock')
+      isRetryablePostError(error)
     ) {
       const noise = Math.random() * 100;
       return 2 ** (attemptCount - 1) * 1000 + noise;
@@ -128,4 +127,29 @@ export class GitlabHttp extends HttpBase<GitlabHttpOptions> {
 
     return super.calculateRetryDelay(retryObject);
   }
+}
+
+/**
+ * Detects transient GitLab errors on POST requests that are safe to retry:
+ *
+ * - `409` with a `Resource lock` message, which happens when concurrent
+ *   requests conflict.
+ * - `400` with a `source_branch does not exist` message, which happens when a
+ *   freshly pushed branch is not yet visible to the API due to Gitaly eventual
+ *   consistency.
+ */
+function isRetryablePostError(error: RequestError): boolean {
+  const { response } = error;
+  if (response?.statusCode === 409) {
+    return response.rawBody.toString().includes('Resource lock');
+  }
+
+  if (response?.statusCode === 400) {
+    const rawBody = response.rawBody.toString();
+    return (
+      rawBody.includes('source_branch') && rawBody.includes('does not exist')
+    );
+  }
+
+  return false;
 }


RahulGautamSingh (commented):
@viceice @secustor Do you think `wrapWithRetry` or `got` approach which should be preferred here?

secustor:
`wrapWithRetry` is more explicit, so lets use that.

viceice (commented):
agreed use the wrap helper

vincentjames501:
Ok, swapped to `wrapWithRetry`. Just some general comments:

- The current impl of `wrapWithRetry` is focused mostly around handling 429s.
- It doesn't do exponential backoffs like the original comment suggested and has a fixed retry count of 2 no matter what.
- I'd personally change the signature a bit given this function is only used in one spot in the application right now

From this:

```
export async function wrapWithRetry<T>(
  task: Task<T>,
  url: string,
  getRetryAfter: (err: unknown) => number | null,
  maxRetryAfter: number,
): Promise<T> {
```

To something more like:

```
export interface RetryContext {
  /** The error from the attempt that just failed. */
  err: unknown;
  /** 1-based index of the retry about to happen (1 = first retry). */
  attempt: number;
  /** ms elapsed since the first attempt started. */
  elapsedMs: number;
}

export interface RetryOptions {
  /** Retries after the initial attempt. Default: 2. */
  maxRetries?: number;

  /** Should this error be retried at all? Return false to rethrow immediately. */
  retryable: (err: unknown) => boolean;

  /** Delay in ms before the next attempt. Gets attempt # → enables backoff. */
  backoff?: (ctx: RetryContext) => number;

  /** Hard cap; if a computed/returned delay exceeds it, stop and rethrow. */
  maxDelayMs?: number;

  /** Opt-in cross-call coordination keyed by host (the Retry-After use case). */
  hostKey?: string;

  /** Cancellation. */
  signal?: AbortSignal;
}

export async function wrapWithRetry<T>(
  task: () => Promise<T>,
  options: RetryOptions,
): Promise<T>;

With small reusable backoff builders:

export const fixedBackoff = (ms: number) => () => ms;

export const exponentialBackoff = (
  baseMs: number,
  { factor = 2, jitterMs = 0, maxMs = Infinity } = {},
) => ({ attempt }: RetryContext) =>
  Math.min(baseMs * factor ** (attempt - 1) + Math.random() * jitterMs, maxMs);

How the two current callers map:

// GitLab source_branch race — 2 retries, 1s→2s exponential:
await wrapWithRetry(() => gitlabApi.postJson(url, { body }), {
  retryable: isSourceBranchDoesNotExistError,
  backoff: exponentialBackoff(1000, { factor: 2, jitterMs: 100 }),
});

// Retry-After header — delay is server-driven, coordinated per host:
await wrapWithRetry(task, {
  retryable: (err) => getRetryAfter(err) !== null,
  backoff: ({ err }) => getRetryAfter(err)! * 1000,
  maxDelayMs: maxRetryAfter * 1000,
  hostKey: parseUrl(url)?.host ?? url,
});
```

@viceice @RahulGautamSingh @secustor

viceice:
please use the re-request review function instead of pinging reviewers

vincentjames501:
> please use the re-request review function instead of pinging reviewers

Apologies. My intention was more to start a discussion around enhancements to your suggestion to use `wrapWithRetry` especially if the intention is to use it in more places like this (i.e. continuation of previous review rather than a new one). I re-requested review after the changes though.

lib/modules/platform/gitlab/index.ts
```diff
@@ -732,19 +744,28 @@ export async function createPr({
...
-        labels: (labels ?? []).join(','),
-        squash: config.squash,
-      },
-    },
+  const mergeRequestsUrl = `projects/${config.repository}/merge_requests`;
+  // GitLab can briefly reject MR creation with `400 source_branch does not
+  // exist` when a freshly pushed branch is not yet visible to the API due to
+  // Gitaly eventual consistency. Retry so the MR is created instead of leaving
+  // the branch orphaned until the next run.
+  const res = await wrapWithRetry(
```
RahulGautamSingh: `wrapWithRetry` is already applied to every request inside `http.ts`, so this nests it, and the outer call writes a 2s entry into the shared module-global `hostDelays` map keyed by the GitLab host. 

That delay then leaks onto the next unrelated GitLab request. The existing place for GitLab transient-retry is `calculateRetryDelay` in `util/http/gitlab.ts`, which already retries POST 409 Resource lock.

lib/modules/platform/gitlab/index.ts
```diff
@@ -717,6 +719,16 @@ async function approveMr(mrNumber: number): Promise<void> {
   }
 }
 
+const mrCreationRetryDelaySeconds = 2;
```
RahulGautamSingh: This value is passed as both the retry delay and the `maxRetryAfter` cap, so it works only because the guard is `delay > maxRetryAfter`. 

Two separate values would make the intent clear. Also, the existing Resource lock retry uses exponential backoff with jitter while this is a flat 2s. For an eventual-consistency race, matching that backoff is gentler on the API.

RahulGautamSingh (changes requested):

viceice (commented):
see open conversations

vincentjames501:
> see open conversations

I believe this is what the original implementation was but was asked by all three reviewers to swap to `wrapWithRetry`. Are you asking me to change this back to what the original commits were? When I swapped back to `wrapWithRetry` I did a force push to keep the commits clean.

Can you please confirm if https://github.com/vincentjames501/renovate/commit/0e4d0798d9a133b1f5917cc72e26378bc04dd84a is the direction you'd prefer? If so I can revert back to this approach.

RahulGautamSingh:
Yes, `0e4d0798` is exactly the direction. Sorry for the confusion.

`calculateRetryDelay` is already used for transient GitLab POST retries (the 409 Resource lock case), and it gives you `got`'s bounded exponential backoff + jitter for free, which is what we want for an eventual-consistency race. 

`wrapWithRetry` is purposefully built for Retry-After (429/403) header coordination: it wraps around the `http` layer that already retries, so it double-nests, and it writes the delay into a module-global `hostDelays` map keyed by host meaning a `source_branch` retry would stall the next, unrelated GitLab request.

Please revert to `0e4d0798` (the `isRetryablePostError` version). 

@secustor @viceice flagging so we're aligned before Vincent redoes the work: I think the `calculateRetryDelay` approach is the better fit here for the reasons above.

viceice:
agreed

lib/modules/platform/gitlab/index.ts
vincentjames501: Reverted to original commit.

lib/modules/platform/gitlab/index.ts
vincentjames501: Reverted to original commit.

viceice (approved):

RahulGautamSingh (approved):

secustor (approved):

PR #44505, opened 2026-07-11
fix(post-upgrade): preserve executable file modes

### LINKED ISSUE #44501: postUpgradeTasks fileFilters drop the executable bit
### Discussed in https://github.com/renovatebot/renovate/discussions/44214

<sup>Originally posted by **richard-kramer** June 24, 2026</sup>

Self-hosted Renovate CLI

GitLab (.com or self-hosted)

43.234.0

When `postUpgradeTasks` runs a command that sets the executable bit on a file (e.g. `chmod +x` or a generator script that creates files with mode `755`), Renovate detects the file as changed via `fileFilters`, but commits it with mode `644`. The execute bit is silently dropped.

### Root cause (suspected)

In `execute-post-upgrade-commands.js`, files picked up from `git status` after the post-upgrade command runs are pushed onto `updatedArtifacts` with only `contents`:

```js
const status = await getRepoStatus();
for (const relativePath of status.modified.concat(status.not_added)) {
  for (const pattern of fileFilters) {
    if (minimatch(pattern, { dot: true }).match(relativePath)) {
      const existingContent = await readLocalFile(relativePath);
      updatedArtifacts.push({
        type: 'addition',
        path: relativePath,
        contents: existingContent,
        // isExecutable is never set here
      });
    }
  }
}
```

`readLocalFile` only reads file content — it doesn't stat the file, so there's no way to know the mode at this point.

Later, in `prepareCommit` (`util/git/index.js`), the artifact is written and staged using that missing flag:

```js
await fs.outputFile(path, contents, {
  mode: file.isExecutable ? 0o777 : 0o666,  // always 0o666 here, since isExecutable is undefined
});
await git.add(addParams);
if (file.isExecutable) {                    // never true for post-upgrade artifacts
  await git.raw(['update-index', '--chmod=+x', fileName]);
}
```

Since `isExecutable` is never set for artifacts collected from `postUpgradeTasks`, the file is always written/staged as non-executable, regardless of its actual mode on disk after the task ran.

This is despite `isExecutable` already being a supported mechanism elsewhere in the codebase (e.g. `hermit/artifacts.js`, `npm/post-update/index.js` both set it explicitly when needed) — `execute-post-upgrade-commands.js` is just never wired up to populate it.

### Minimal reproduction

1. Repo contains a tracked script with mode `644`, e.g. `scripts/run.sh`.
2. `renovate.json`:

```json
{
  "packageRules": [
    {
      "matchManagers": ["npm"],
      "postUpgradeTasks": {
        "commands": ["chmod +x scripts/run.sh"],
        "fileFilters": ["scripts/run.sh"],
        "executionMode": "branch"
      }
    }
  ]
}
```

3. Trigger any update matched by this rule.
4. Renovate commits `scripts/run.sh` — but the commit still has mode `100644`, not `100755`.
5. After the branch is checked out / merged, `scripts/run.sh` loses its executable bit relative to what the post-upgrade command actually produced in the working tree.

### Expected behavior

If a file picked up via `fileFilters` has the executable bit set on disk after the post-upgrade command runs, Renovate should commit it with that mode preserved (`isExecutable: true`), the same way it already does for hermit/npm artifacts.

### Suggested fix

Stat the file when collecting it instead of relying solely on `readLocalFile`:

```js
const stat = await statLocalFile(relativePath); // already exported from util/fs
const isExecutable = !!(stat?.mode & 0o111);

updatedArtifacts.push({
  type: 'addition',
  path: relativePath,
  contents: existingContent,
  ...(isExecutable && { isExecutable: true }),
});
```

`statLocalFile` is already implemented and exported in `util/fs/index.js`, so this would not require a new dependency.

### Environment

- Renovate versions checked: v37.x through v43.x (same code path, behavior unchanged)

### Workaround attempted

Setting `core.fileMode=true` globally before Renovate runs does **not** fix this — it only affects whether `git status` reports mode changes as `modified`, not whether Renovate's `prepareCommit` actually preserves the mode when writing/staging the artifact. The bug is in the artifact layer, not in git's diff detection.

### Logs (if relevant)

_No response_</div>

zharinov:
## Changes

`postUpgradeTasks.fileFilters` now preserves executable files produced by post-upgrade commands.

### Why the executable bit was lost

1. A post-upgrade command made a matched file executable, either through `chmod +x` or by generating it with mode `0755`.
2. `fileFilters` copied the file into `updatedArtifacts`, but recorded only its contents.
3. `prepareCommit()` reset the worktree and recreated the file without `isExecutable: true`.
4. Git could therefore commit the file as `100644` instead of `100755`.

### How executable state is collected

Renovate first reads `core.fileMode`. If the key is absent, this PR substitutes `true`. That fallback matches Git's documented default for an unspecified setting.

- If the resulting value is `true` and the owner execute bit is set, Renovate records `isExecutable: true`.
- Otherwise, Renovate leaves the artifact's existing `isExecutable` value unchanged. A new artifact therefore remains unset.

This interpretation follows Git's mode handling. `core.fileMode=false` means Git does not trust worktree execute bits. Git also uses the owner execute bit to choose between `100644` and `100755`, so group or other execute permissions alone should not make the committed file executable.

### Scope

This fixes both cases reported in #44501: an existing file changed with `chmod +x` and a newly generated executable file.

The collector only adds a positive executable signal. It never writes `isExecutable: false`, so this PR does not establish a general executable-removal contract. Supporting that contract would require the commit path to distinguish three states: force executable, force non-executable, and preserve an unknown or existing mode. The current path does not distinguish `false` from unset.

## Context

- [x] This closes an existing Issue, Closes: #44501

## AI assistance disclosure

AI assistance covered investigation, implementation, tests, review, and this PR description.

Model used: GPT-5.6 Sol

The real-Git regression starts with a tracked `100644` file, makes it executable through the post-upgrade command path, collects it through `fileFilters`, passes it to `prepareCommit()`, and verifies that Git records `100755`.

Focused tests cover new and existing artifacts, every `core.fileMode` state, owner versus group or other execute bits, unavailable file metadata, and preservation of an existing `isExecutable: true` value.

The public repository: Not applicable.


viceice (approved):

jamietanna (commented):
This looks to have caused a regression: https://github.com/renovatebot/renovate/discussions/44875

PR #44560, opened 2026-07-14
feat(logging): do not use dynamic fields in log messages with severity WARN or higher

secustor:
## Changes

Adds a custom oxlint rule that enforces a static msg component for logger.warn/error/fatal calls (including logger.once variants) so the messages can be grouped in metrics and error catching services, per docs/development/best-practices.md. It also flags metadata properties named 'error' that hold error-ish values, because the bunyan serializer only handles the 'err' key.

The rule is enabled as an error for lib/ production code and disabled for spec files. All 33 pre-existing violations are fixed by moving interpolated values into the metadata object, making messages static, and renaming 'error' metadata keys to 'err'. Affected log-message assertions in co-located specs are updated accordingly.

We could follow up this with a general standardization around log meta data

## Context

## AI assistance disclosure

The public repository: <URL>


lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
```diff
@@ -161,7 +161,7 @@ export async function postUpgradeCommandsExecutor(
                       validTool,
                       validConstraint,
                     },
-                    `Skipping ${validConstraint ? 'valid' : 'invalid'} constraint that is not a tool that Containerbase knows`,
+                    'Skipping constraint that is not a tool that Containerbase knows',
```
jamietanna, line 164: Should we instead make this an `if` around the log? IMO, it's useful to have this

jamietanna (commented):
cc @nabeelsaabna for visibility. AFAICT, this shouldn't cause any issues for us

lib/workers/repository/update/branch/execute-post-upgrade-commands.ts
secustor: You mean also to log in case the `tool` is invalid?

jamietanna: As in to log whether it's a `valid constraint` or an `invalid constraint`

secustor: `validConstraint` is part of the metadata, it is a boolean instead of a string with `valid` and `invalid` though. 

I don't see the value of switching this to a string union or duplicating the same info 🤔

lib/config-validator.ts
```diff
@@ -94,7 +94,7 @@ async function validate(
           );
       })
       .join('\n');
-    logger.warn(`Config migration diff:\n${msg}`);
+    logger.warn({ diff: msg }, 'Config migration diff');
```
jamietanna: Can we confirm the output looks like we expect?

viceice: this will look weird because it's json stringified. 🤔

secustor: Yeah, it has been indeed mangled. I have simply added a exemption.

jamietanna (approved):

PR #44567, opened 2026-07-14
fix: honour dashboard rebase for edited branch with no open PR

pose:
## Changes

The Dependency Dashboard's **PR Edited (Blocked)** checkbox is silently ignored for an edited branch with no open PR: the `else if (branchIsModified)` block in `processBranch()` returns `pr-edited` without consulting `dependencyDashboardCheck`, unlike every other early return in that function. The dashboard re-renders the branch in the same list with the box unticked, so the request can never take effect.

Skipping that block when a dashboard check is present lets the flow reach `userRebaseRequested`, which sets `reuseExistingBranch = false` and recreates the branch from base, as the section's text promises.

## Context

## AI assistance disclosure

Claude Opus 4.8, via Claude Code: the code change, the two unit tests, and this description. Human-reviewed before submission.


pose:
For transparency: I force-pushed this branch a few times while it was still in draft, to fix the commit message and author email. No commits have been amended since it was marked ready for review, and I won't rebase from here.

RahulGautamSingh (commented):
I understand the edge case this is handling, but also curious to know the real case in which it is affecting you. ie. can you please create a reproduction and share how you noticed this.

pose:
Sure. Real case: on pulumi/pulumi our `renovate/pulumi` and `renovate/major-pulumi` branches got wedged. Both had been hand-edited by a human, and their original PRs (pulumi/pulumi#21161 and pulumi/pulumi#21081) had since been closed, so each was a modified branch with a closed PR rather than an open one. Ticking their boxes in the **PR Edited (Blocked)** section did nothing, run after run.

Minimal reproduction:

1. Let Renovate open an update PR so its branch exists.
2. Push a manual commit to that branch, so `isBranchModified()` becomes true.
3. Close (or merge/supersede) that PR, so there's no open PR but the modified branch remains. Its old closed PR is still findable via `findPr(state: '!open')` with a SHA that no longer matches the branch head.
4. On the Dependency Dashboard, tick that branch's box in **PR Edited (Blocked)** ("discard all commits and start over").

Expected: the branch is recreated from base. Actual: nothing happens, and it re-renders in the same section with the box unticked, so the click can never take effect.

The give-away in our logs was, for the exact branch, `dependencyDashboardCheck=rebase` immediately followed by `Found old PR but the SHA is different` and then no further action — `processBranch()` returns `pr-edited` from the `else if (branchIsModified)` block before reaching the `userRebaseRequested` handling ~120 lines down. The sibling "no old PR found" sub-case returns the same way. Reproduced against `main` (2026-07-13).

The two regression tests in this PR (`index.spec.ts`) exercise both sub-cases — modified branch with no PR, and modified branch with a SHA-mismatched old PR — each with `dependencyDashboardChecks` set, so they're a runnable reproduction of the fix.

pose:
Hey @RahulGautamSingh, sorry to bother you again. I was wondering if you had some time to check my previous response. Thanks! 🙏

RahulGautamSingh (changes requested):
Thanks for the detailed reply!

The PR LGTMs nnly the test setup is slightly misleading: the `findPr` mocks are never consumed because the new guard short-circuits before that call, so the "sha mismatch" test isn't testing a sha mismatch. 

Drop the `findPr` mocks (they add nothing)

pose:
Confirmed: I instrumented the tests and `platform.findPr` is called 0 times, so both mocks were dead. Dropped them. That also made the two tests byte-identical, so they're now a single test covering "branch edited, no open PR, dashboard rebase ticked". Re-verified it fails against the unguarded `else if (branchIsModified)` and passes with the fix.

RahulGautamSingh (commented):
tests failing 😕

pose:
Fixed in bd9317bd6, and merged main in. Main had switched the shared mock to `fakeSha('123test')` while my test still asserted the old literal, so it only broke on the merge ref. Now `expect.any(String)`. All checks green.

RahulGautamSingh (approved):
lgtm

viceice (approved):

PR #44585, opened 2026-07-14
fix(config): make `platformCommit` global-only

felipecrs:
## Changes

This PR changes `platformCommit` so that it's a global-only option rather than repository-level.

I believe this only makes sense.

/cc @zharinov

## Context

## AI assistance disclosure

Used Grok Build 4.5 for implementing this.

The public repository: N/A

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -4033,17 +4033,6 @@ Please check platform specific docs for version requirements.
 
 To learn how to use GitHub's Merge Queue feature with Renovate, read our [GitHub Merge Queue](./key-concepts/automerge.md#github-merge-queue) docs.
 
-## `platformCommit`
-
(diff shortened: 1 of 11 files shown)


zharinov:
👍

jonathanmorley:
We have a use-case for `platformCommit` in `packageRules`:

GitHub has a (from what I can find) undocumented limit on the number of files that can be created with the https://docs.github.com/en/rest/git/trees?apiVersion=2026-03-10 endpoint.

Normally, we don't hit that limit, and we can use `platformCommit: true`.

However, one of our internal packages decided that they wanted to add around 900 files to consuming repositories when the package was updated (using `postUpgradeTasks`). This hit the limit, and so we used `platformCommit: false` in a `packageRule` scoped to that package to allow them to proceed.

While I don't necessarily agree with that package's decisions, this is a configuration workaround that allows them to do it.

felipecrs:
@jonathanmorley, that's super useful feedback, thank you!

I'm closing this PR as I don't want to cause any regression.

PR #44592, opened 2026-07-15
fix(platform/azure): don't create Dependency Dashboard for projects without an `Issue` type

emanuelbesliu:
## Changes

`IssueService.ensureIssue` dereferences `newWorkItem.id` immediately after `createWorkItem`, assuming a work item is always returned. The underlying `azure-devops-node-api` / `typed-rest-client` layer, however, resolves `createWorkItem` to `null` **without throwing** for:

- HTTP `404 Not Found` (returned as `{ result: null }`), and
- any `2xx` response with an empty or non-JSON body — e.g. a `203` sign-in page from an expired/invalid token, or an HTML page injected by an SSO/proxy in front of the instance (JSON parsing fails silently and `result` stays `null`).

In those cases the run crashed with:

```
WARN: Error ensuring issue (repository=...)
TypeError: Cannot read properties of null (reading 'id')
    at IssueService.ensureIssue (.../azure/issue.ts:344:54)
```

This guards the result: if no work item (with an `id`) is returned, we log a warning and skip the issue instead of crashing the Dependency Dashboard for the whole repository. Reported in discussion #43785 as a regression-adjacent edge case after the Azure Dependency Dashboard state autodetection (#44412) fixed the original `System.State` `400` errors.

## Context

## AI assistance disclosure

- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation): investigation, the null guard, and the unit test were drafted with GitHub Copilot CLI (Claude Opus 4.8) and reviewed by me.

Added a unit test asserting that `ensureIssue` returns `null` and logs a warning when `createWorkItem` resolves to `null`. `pnpm vitest run lib/modules/platform/azure/issue.spec.ts` passes (93 tests).

### DIFF
--- a/lib/modules/platform/azure/index.spec.ts
+++ b/lib/modules/platform/azure/index.spec.ts
@@ -2210,6 +2210,7 @@ describe('modules/platform/azure/index', () => {
       partial<IWorkItemTrackingApi>({
         queryByWiql: vi.fn().mockResolvedValue({ workItems: [] }),
         createWorkItem: createWorkItemMock,
+        getWorkItemTypes: vi.fn().mockResolvedValue([{ name: 'Issue' }]),
       }),
     );
     const result = await azure.ensureIssue({
--- a/lib/modules/platform/azure/issue.spec.ts
+++ b/lib/modules/platform/azure/issue.spec.ts
@@ -309,6 +309,7 @@ describe('modules/platform/azure/issue', () => {
       azureApi.workItemTrackingApi.mockResolvedValue(
         partial<IWorkItemTrackingApi>({
           createWorkItem: createWorkItemMock,
+          getWorkItemTypes: vi.fn().mockResolvedValue([{ name: 'Issue' }]),
         }),
       );
 
@@ -328,6 +329,7 @@ describe('modules/platform/azure/issue', () => {
       azureApi.workItemTrackingApi.mockResolvedValue(
         partial<IWorkItemTrackingApi>({
           createWorkItem: createWorkItemMock,
+          getWorkItemTypes: vi.fn().mockResolvedValue([{ name: 'Issue' }]),
         }),
       );
 
@@ -340,6 +342,29 @@ describe('modules/platform/azure/issue', () => {
       expect(result).toBeNull();
       expect(logger.warn).toHaveBeenCalledWith(
         'Azure: work item creation returned no result; skipping issue',
+      );
+    });
+
+    it('should skip issue when the Issue work item type does not exist', async () => {
+      vi.spyOn(issueService, 'getIssueList').mockResolvedValue([]);
+
+      const createWorkItemMock = vi.fn();
+      azureApi.workItemTrackingApi.mockResolvedValue(
+        partial<IWorkItemTrackingApi>({
+          createWorkItem: createWorkItemMock,
+          getWorkItemTypes: vi.fn().mockResolvedValue([{ name: 'Bug' }]),
+        }),
+      );
+
+      const result = await issueService.ensureIssue({
+        title: 'Test Issue',
+        body: 'Test body content',
+      });
+
+      expect(createWorkItemMock).not.toHaveBeenCalled();
+      expect(result).toBeNull();
+      expect(logger.warn).toHaveBeenCalledWith(
+        expect.stringContaining("work item type 'Issue' does not exist"),
       );
     });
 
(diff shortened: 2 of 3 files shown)


jamietanna:
I guess we've not been able to test this against a real repo, to find out exactly why this issue occurs?

emanuelbesliu:
Correct that we haven't reproduced this *specific* null-return on a live instance ourselves — but the reporter's debug log in [#43785](https://github.com/renovatebot/renovate/discussions/43785#discussioncomment-17646035) narrows the root cause down convincingly:

```
DEBUG: getIssueList() no work items found
 WARN: Error ensuring issue → Cannot read properties of null (reading 'id')
```

The WIQL read (`getIssueList` → `queryByWiql`) succeeds, so auth, project and the Work Item Tracking API are all fine — it's specifically `createWorkItem(..., type: 'Issue')` that resolves to `null`. They're on hosted `dev.azure.com` and their PAT can create PRs, so this isn't auth or an on-prem route issue.

Leading cause: **their project's process has no `Issue` work item type** (Basic/Agile/CMMI define one; **Scrum does not**). An unknown type → 404 → `typed-rest-client` returns `null` without throwing → the crash; meanwhile `getIssueList`'s `WHERE [System.WorkItemType] = 'Issue'` matches nothing → "no work items found". That fits the log exactly. I've asked them to confirm the process.

Either way this PR is the defensive fix: it turns a hard crash of the whole Dependency Dashboard step into a skipped issue + warning, and it's covered by a unit test that mocks the null return. If the missing-`Issue`-type theory is confirmed, I'll follow up with up-front detection so those users get an actionable message instead of a silent skip — or I can fold that into this PR if you'd prefer it handled together.

jamietanna:
(Aside: https://github.com/renovatebot/renovate/blob/main/.github/contributing.md#ai-assistance-notice)

jamietanna:
Are you able to see if you can create a Work Item via the API against the project you're seeing this issue with? (via the API, or running Renovate locally)

That may help confirm the case this is happening with, so we can understand the root cause + make sure this is the right fix

emanuelbesliu:
Reproduced the mechanism against live Azure DevOps Services (hosted), using the `azure-devops-node-api` dependency, `validateOnly=true` (nothing persisted):

- `createWorkItem` with a work item type **the project's process doesn't define** → **HTTP 404** (`WorkItemTypeNotFoundException`), which `typed-rest-client` silently turns into `null` (no throw) → `newWorkItem.id` throws the reported `Cannot read properties of null (reading 'id')`.
- A valid type → returns a work item, no crash.

ADO resolves the type by name against the project's process and returns that same 404 for any name not present, so this is the exact code path the reporter hits. The 404 fires when the `Issue` type doesn't exist in the process (Scrum defines none; Basic/Agile/CMMI do) **or** the token lacks permission to it — the message covers both, so I can't say which applies to the reporter without their process/permission details. Either way this PR's null guard is the correct defensive fix; I'll follow up with explicit 404 handling + an actionable log.

*AI assistance disclosure: I used GitHub Copilot CLI to help with the investigation and drafting; the analysis, decisions, and this contribution are my own.*

jamietanna:
OK great, thanks! 

Is there an API where we can check if that Work Item type (`Issue`) actually exists, on top of what we're doing to find if there are currently any of that type?

jamietanna:
I guess via https://github.com/renovatebot/renovate/discussions/43785#discussioncomment-17646646 we can at least see that this error is the Work Item type not existing, so let's see if we can use that to improve the error message

emanuelbesliu:
Yes — `getWorkItemTypes(project)` lists the project's types, or `getWorkItemType(project, 'Issue')` (404 if absent). Both are project-scoped, so we can check `Issue` exists before creating and log a clear message if not. I'll add that to this PR.

emanuelbesliu:
Pushed: `ensureIssue` now calls `getWorkItemTypes(project)` and, if `Issue` isn't defined, logs an actionable warning and skips instead of attempting the create. The null guard stays as a fallback for other empty-body cases. 100% coverage.

*AI assistance disclosure: implemented with GitHub Copilot CLI; reviewed and authored by me.*

jamietanna (dismissed):
~~Thoughts @viceice~~ Ignore :)

jamietanna:
Looks like tests are failing - once that's resolved, we'll be good to go

jamietanna (approved):

PR #44621, opened 2026-07-15
feat(gitlab): add gitlab codeowners role support

sbreakey:
Discussion: https://github.com/renovatebot/renovate/discussions/44620

## Changes

GitLab supports the `@@role` syntax in its CODEOWNERS specification (for example `@@developer`, `@@maintainer`, `@@owner`), which assigns a code owner by project role rather than by a specific user or group. See the [GitLab docs](https://docs.gitlab.com/user/project/codeowners/reference/#add-a-role-as-a-code-owner).

Previously, when resolving reviewers/assignees, `expandGroupMembers` did not recognise the `@@role` syntax. It stripped the leading `@` and passed the remainder (for example `@developer`) to the GitLab groups API as if it were a group name. No such group exists, so the request returned a `404`, the role handle was pushed through unchanged as a fallback, and the intended reviewers were never assigned.

This PR detects `@@role` handles in `expandGroupMembers` before the group-expansion path. When a handle maps to a known GitLab role, it resolves to the project's members at the matching access level (via the project members API, filtered by `access_level`) and expands to their usernames, mirroring how group handles are expanded today. Handles that are not recognised roles continue through the existing group/user logic unchanged.

## Context

## AI assistance disclosure

Claude Code (Opus 4.8) was used to help implement the role-resolution logic, write unit tests, and draft documentation. All changes were reviewed by me.

The public repository: N/A

### DIFF
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -2654,7 +2654,7 @@ const options: Readonly<RenovateOptions>[] = [
   {
     name: 'expandCodeOwnersGroups',
     description:
-      'Expand the configured code owner groups into a full list of group members. On GitLab, also expands role handles (`@@developer`, `@@maintainer`, `@@owner`) into the project members holding that role.',
+      'Expand the configured code owner groups into a full list of group members.',
     type: 'boolean',
     default: false,
     supportedPlatforms: ['gitlab'],
--- a/lib/config/options/index.ts
+++ b/lib/config/options/index.ts
@@ -2654,7 +2654,7 @@ const options: Readonly<RenovateOptions>[] = [
   {
     name: 'expandCodeOwnersGroups',
     description:
-      'Expand the configured code owner groups into a full list of group members.',
+      'Expand the configured code owner groups into a full list of group members. On GitLab, also expands role handles (`@@developer`, `@@maintainer`, `@@owner`) into the project members holding that role.',
     type: 'boolean',
     default: false,
     supportedPlatforms: ['gitlab'],
--- a/lib/modules/platform/gitlab/http.ts
+++ b/lib/modules/platform/gitlab/http.ts
@@ -21,7 +21,7 @@ export async function getUserID(username: string): Promise<number> {
   return userInfo[0].id;
 }
 
-async function getMembers(group: string): Promise<GitLabUser[]> {
+async function getGroupMembers(group: string): Promise<GitLabUser[]> {
   const groupEncoded = encodeURIComponent(group);
   return (
     await gitlabApi.getJsonUnchecked<GitLabUser[]>(
@@ -32,7 +32,7 @@ async function getMembers(group: string): Promise<GitLabUser[]> {
 
 export async function getMemberUserIDs(group: string): Promise<number[]> {
   try {
-    const members = await getMembers(group);
+    const members = await getGroupMembers(group);
     return members.map((u) => u.id);
   } catch (err) {
     logger.once.warn(
@@ -44,7 +44,7 @@ export async function getMemberUserIDs(group: string): Promise<number[]> {
 }
 
 export async function getMemberUsernames(group: string): Promise<string[]> {
-  const members = await getMembers(group);
+  const members = await getGroupMembers(group);
   return members.map((u) => u.username);
 }
 
--- a/lib/modules/platform/gitlab/http.ts
+++ b/lib/modules/platform/gitlab/http.ts
@@ -48,25 +48,22 @@ export async function getMemberUsernames(group: string): Promise<string[]> {
   return members.map((u) => u.username);
 }
 
-async function getProjectMembersByRole(
-  repo: string,
-  accessLevel: number,
-): Promise<GitLabUser[]> {
-  const members = (
+async function getProjectMembers(repo: string): Promise<GitLabUser[]> {
+  const repoEncoded = encodeURIComponent(repo);
+  return (
     await gitlabApi.getJsonUnchecked<GitLabUser[]>(
-      `projects/${encodeURIComponent(repo)}/members`,
+      `projects/${repoEncoded}/members`,
       { paginate: true },
     )
   ).body;
-  return members.filter((m) => m.access_level === accessLevel);
 }
 
-export async function getRoleMemberUsernames(
+export async function getProjectMembersByRole(
   repo: string,
   accessLevel: number,
-): Promise<string[]> {
-  const members = await getProjectMembersByRole(repo, accessLevel);
-  return members.map((u) => u.username);
+): Promise<GitLabUser[]> {
+  const members = await getProjectMembers(repo);
+  return members.filter((m) => m.access_level === accessLevel);
 }
 
 export async function isUserBusy(user: string): Promise<boolean> {
--- a/lib/modules/platform/gitlab/http.ts
+++ b/lib/modules/platform/gitlab/http.ts
@@ -48,6 +48,27 @@ export async function getMemberUsernames(group: string): Promise<string[]> {
   return members.map((u) => u.username);
 }
 
+async function getProjectMembersByRole(
+  repo: string,
+  accessLevel: number,
+): Promise<GitLabUser[]> {
+  const members = (
+    await gitlabApi.getJsonUnchecked<GitLabUser[]>(
+      `projects/${encodeURIComponent(repo)}/members`,
+      { paginate: true },
+    )
+  ).body;
+  return members.filter((m) => m.access_level === accessLevel);
+}
+
+export async function getRoleMemberUsernames(
+  repo: string,
+  accessLevel: number,
+): Promise<string[]> {
+  const members = await getProjectMembersByRole(repo, accessLevel);
+  return members.map((u) => u.username);
+}
+
 export async function isUserBusy(user: string): Promise<boolean> {
   try {
     const url = `/users/${user}/status`;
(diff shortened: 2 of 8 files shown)


lib/modules/platform/gitlab/roles.ts
```diff
@@ -0,0 +1,26 @@
...
+  minimal_access: 5,
+  guest: 10,
+  planner: 15,
+  reporter: 20,
+  developer: 30,
+  maintainer: 40,
+  owner: 50,
+};
+
+const roleRegex = regEx(/^@@(\w+)$/);
```
RahulGautamSingh: Use `/^@@(?<role>\w+)$/` and `match.groups?.role`

RahulGautamSingh (changes requested):
lgtm apart from this nitpick

lib/modules/platform/gitlab/roles.ts
sbreakey: Thanks, much nicer! Applied that.

RahulGautamSingh (approved):

lib/config/options/index.ts
```diff
@@ -2655,7 +2654,7 @@ const options: Readonly<RenovateOptions>[] = [
   {
     name: 'expandCodeOwnersGroups',
     description:
-      'Expand the configured code owner groups into a full list of group members.',
+      'Expand the configured code owner groups into a full list of group members. On GitLab, also expands role handles (`@@developer`, `@@maintainer`, `@@owner`) into the project members holding that role.',
```
secustor:
```suggestion
      'Expand the configured code owner groups into a full list of group members.',
```

That description is getting too long.

lib/modules/platform/gitlab/http.ts
```diff
@@ -48,6 +48,27 @@ export async function getMemberUsernames(group: string): Promise<string[]> {
...
+async function getProjectMembersByRole(
+  repo: string,
+  accessLevel: number,
+): Promise<GitLabUser[]> {
+  const members = (
+    await gitlabApi.getJsonUnchecked<GitLabUser[]>(
+      `projects/${encodeURIComponent(repo)}/members`,
+      { paginate: true },
+    )
+  ).body;
```
secustor: Use the existing getMembers function 

https://github.com/renovatebot/renovate/pull/44621/changes#diff-01672374522bf63a4e5e74360b5e1cb1e7f13edbd3aa293f13fc5a39c1aeae5eR24

lib/modules/platform/gitlab/roles.ts
```diff
@@ -0,0 +1,26 @@
...
+const roleAccessLevels: Record<string, number> = {
+  no_access: 0,
+  minimal_access: 5,
+  guest: 10,
+  planner: 15,
+  reporter: 20,
+  developer: 30,
+  maintainer: 40,
+  owner: 50,
+};
```
secustor:
```suggestion
} as const;
```

secustor (changes requested):

lib/modules/platform/gitlab/http.ts
sbreakey: @secustor thanks for the review. If I have understood you correctly, [getMembers](https://github.com/renovatebot/renovate/blob/969d1a1c49ff273218fe10a094bbd9bbf9a3ae6b/lib/modules/platform/gitlab/http.ts#L24-L31) doesn't appear appropriate here.

It calls `groups/{group}/members`, whereas I need to call `projects/{repo}/members` to get project level members as opposed to group specific members.

Please let me know if I have misunderstood.

An alternative tidy up may be:
- rename `getMembers` to `getGroupMembers` for clarity
- factor out a `getProjectMembers` to do the pure api call (retaining the pagination which `getMembers` incidentally lacks)
- keep `getRoleMemberUsernames` as a thin wrapper that delegates to `getProjectMembers` and does the `access_level` filtering and mapping to `username` itself

secustor: Ah, I have missed that detail. 

Let's do it like this: 
- Rename getMembers to getGroupMembers 
- Extract `getProjectMembers` out of `getProjectMembersByRole`
- Move the logic of `getRoleMemberUsernames` to https://github.com/renovatebot/renovate/pull/44621/changes#diff-bd7375b4be9976d8d1ee7d683faf833d11b283e65518c59be09ace09271b90e8R1485

Having only the usernames there reduces the functions re usability.

lib/modules/platform/gitlab/index.spec.ts
```diff
@@ -4340,5 +4254,63 @@ These updates have all been created already. To force a retry/rebase of any, cli
...
+        .reply(200, [
+          { username: 'dev-one', access_level: 30 },
+          { username: 'maintainer', access_level: 40 },
+          { username: 'dev-two', access_level: 30 },
+          { username: 'owner', access_level: 50 },
+        ]);
+      const expandedGroupMembers = await gitlab.expandGroupMembers?.([
+        '@@developer',
+      ]);
+      expect(expandedGroupMembers).toEqual(['dev-one', 'dev-two']);
```
secustor:
```suggestion
      await expect(gitlab.expandGroupMembers?.([
        '@@developer',
      ])).resolves.toEqual(['dev-one', 'dev-two']);
```
Chain this directly together, so there are better errors in case of crashes.

Same for the other tests.

sbreakey: Thanks, I applied this locally and have pushed.

To point out, I changed other tests that I initially hadn't touched. Seemed sensible to remain consistent.

lib/modules/platform/gitlab/http.ts
sbreakey: Thanks, thats done.

secustor (approved):

secustor:
Thanks for contributing!

lib/modules/platform/gitlab/http.ts
```diff
@@ -44,10 +44,28 @@ export async function getMemberUserIDs(group: string): Promise<number[]> {
...
 export async function getMemberUsernames(group: string): Promise<string[]> {
-  const members = await getMembers(group);
+  const members = await getGroupMembers(group);
   return members.map((u) => u.username);
 }
 
+async function getProjectMembers(repo: string): Promise<GitLabUser[]> {
+  const repoEncoded = encodeURIComponent(repo);
+  return (
+    await gitlabApi.getJsonUnchecked<GitLabUser[]>(
```
viceice, line 54: new http requests should use zod schema validation 😔

lib/modules/platform/gitlab/roles.ts
```diff
@@ -0,0 +1,26 @@
...
+const roleAccessLevels: Record<string, number> = {
+  no_access: 0,
+  minimal_access: 5,
+  guest: 10,
+  planner: 15,
+  reporter: 20,
+  developer: 30,
+  maintainer: 40,
+  owner: 50,
+} as const;
```
viceice, line 13: const males no sense when the variable is explicit typed, so this or the explicit type should be removed

lib/modules/platform/gitlab/roles.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const roleRegex = regEx(/^@@(?<role>\w+)$/);
+
+/**
+ * Parse a GitLab CODEOWNERS role handle (`@@developer`, `@@maintainer`,
+ * `@@owner`, etc) into its access level. Returns `null` when the handle is not a
+ * recognized role, so `getRoleAccessLevel(x) !== null` doubles as a role check.
+ */
+export function getRoleAccessLevel(handle: string): number | null {
+  const role = roleRegex.exec(handle)?.groups?.role;
```
viceice, line 23: I would use simple string ops (starts with, substting, ...)instead of regex for performance reasons.

lib/modules/platform/gitlab/roles.ts
```diff
@@ -0,0 +1,26 @@
...
+  const level = role ? roleAccessLevels[role.toLowerCase()] : undefined;
```
viceice, line 24: does gitlab also ignore the case?

sbreakey: [The api](https://docs.gitlab.com/api/project_members/) unfortunately doesn't accept any filtering on the role, so a lowercased role never reaches the API. It's only lowercased for local matching, and filtering is then local too and based on the mapped `access_level` number not the role text.

lib/modules/platform/gitlab/http.ts
sbreakey: @viceice I'm happy to submit a follow up for this, and the below comments too. I will do so in a day or two.

sbreakey: PR created here: https://github.com/renovatebot/renovate/pull/44796

lib/modules/platform/gitlab/roles.ts
viceice: I mean: does gitlab accept any casing in code owners file for roles or names?

sbreakey: doh. I can't find any concrete evidence in the docs one way or another.

I did stumble upon [this issue](https://gitlab.com/gitlab-org/gitlab/-/work_items/553323) which incidentally mentions (in [this comment](https://gitlab.com/gitlab-org/gitlab/-/work_items/553323#note_2645474939)) that usernames are case insensitive when it comes to approval rights. So I assume roles are too.

viceice: ok

PR #44651, opened 2026-07-16
feat: add `commitTrailers` option

felipecrs:
## Changes

Currently, the only way to add git trailers today is via `commitBody`. But that doesn't work for Gerrit because Renovate appends its own `Renovate-Branch`/`Change-Id` block **after** the body, so user-defined trailers end up in a non-final message block and are not real trailers per [git interpret-trailers](https://git-scm.com/docs/git-interpret-trailers) rules.

This PR adds a dedicated `commitTrailers: string[]` option. Each entry should be a full `Key: value` line, e.g.:

```json
{
  "commitTrailers": ["Signed-off-by: {{{gitAuthor}}}"]
}
```

It also standardizes Gerrit's own `Renovate-Branch`/`Change-Id` trailers through the same mechanism, so user trailers and platform trailers now merge into one final block.

And migrates the `:gitSignOff` preset from `commitBody` to `commitTrailers`, so sign-off works correctly on Gerrit too.

I believe a dedicated option like this makes sense because:

1. [Git commit trailers](https://git-scm.com/docs/git-interpret-trailers) are a standardized way to store **structured** data through the commit message. `commitBody` is too flexible and therefore cannot really ensure user-defined trailers are treated as actual trailers.
2. This option is **mergeable**. If a given Renovate global config requires some specific trailer, overriding it per repository does not require re-defining the global one. For example, this allow users to extend `gitSignOff` preset while still defining their own trailers.

> [!IMPORTANT]
> Previously the `:gitSignOff` preset set `commitBody`, so configuring your own `commitBody` overrode the preset and the sign-off was dropped. The preset now uses `commitTrailers` instead, so a custom `commitBody` and the `Signed-off-by` trailer are both kept.

## Context

Related:

- https://github.com/renovatebot/renovate/discussions/33306
- https://github.com/renovatebot/renovate/discussions/28047

## AI assistance disclosure

Claude Sonnet 5 was used to implement this.

The public repository:

- https://review.gerrithub.io/c/felipecrs/renovate-repro-commit-trailers/+/1242297
- https://github.com/felipecrs/renovate-repro-commit-trailers/pull/3/commits


lib/util/git/index.ts
```diff
@@ -1359,6 +1376,33 @@ export async function prepareCommit({
...
+  const missing = trailers.filter(
+    (trailer) => !parsedTrailers.includes(trailer),
+  );
+  if (isNonEmptyArray(missing)) {
+    logger.warn(
+      { missing },
+      'Some commit trailers were not parsed as git trailers',
+    );
+  }
+}
```
felipecrs: I thought this would be nice to have, but considering it bypasses simple-git and it's only meant to add a proactive warning on logs, I think I'll drop it from this PR.

lib/config/validation.ts
```diff
@@ -450,6 +451,19 @@ export async function validateConfig(
                   }
                 }
 
+                if (key === 'commitTrailers') {
+                  for (const subval of val) {
+                    if (!isString(subval) || !commitTrailerRe.test(subval)) {
```
RahulGautamSingh: this checks the raw template, not the compiled value. `Signed-off-by: {{{gitAuthor}}}` passes even if `gitAuthor` is empty or expands to a newline, which then breaks the trailer block. validate the compiled output too.

lib/config/presets/internal/default.preset.ts
```diff
@@ -282,7 +282,7 @@ export const presets: Record<string, Preset> = {
     ],
   },
   gitSignOff: {
-    commitBody: 'Signed-off-by: {{{gitAuthor}}}',
+    commitTrailers: ['Signed-off-by: {{{gitAuthor}}}'],
```
RahulGautamSingh, line 285: this changes commit output for anyone on `:gitSignOff` who also sets `commitBody`. the sign-off moves from inline to the final trailer block. flag it in the PR description.

docs/usage/configuration-options.md
```diff
@@ -805,6 +797,30 @@ We recommend you use `matchManagers` and `commitMessageTopic` in a `packageRules
 }
 ```
 
+## `commitTrailers`
+
+Use this option to add [git trailers](https://git-scm.com/docs/git-interpret-trailers) to Renovate's commits.
```
```suggestion
Use this option to add [git trailers](https://git-scm.com/docs/git-interpret-trailers) to the commits Renovate creates.
```

lib/workers/repository/onboarding/branch/create.ts
```diff
@@ -36,6 +36,11 @@ export async function createOnboardingBranch(
     logger.trace(`commitMessage: ${commitMessage}`);
   }
 
+  // only allow gitAuthor template value in the commitTrailers
+  const trailers = config.commitTrailers?.map((trailer) =>
+    compile(trailer, { gitAuthor: config.gitAuthor }),
```
RahulGautamSingh, line 43: only `gitAuthor` is compiled here, so `{{{updateType}}}` from the docs example silently drops during onboarding. same as `commitBody` today, but the docs should say so.

lib/config/presets/internal/default.preset.ts
felipecrs: Well, in fact if someone sets `:gitSignOff` and also sets `commitBody`, then `:gitSignOff` gets overridden and doesn't do anything.

Whereas now both the custom `commitBody` plus the `:gitSignOff` trailer will be kept.

I'll flag that in the PR description.

lib/config/validation.ts
felipecrs: At config validation phase only the raw template can be validated. But I've added a runtime validation which checks after the compilation as well.

https://github.com/renovatebot/renovate/pull/44651/commits/0ac5f8d3df04aa6b9743619e76cf3938828c3cfe

lib/workers/repository/onboarding/branch/create.ts
felipecrs: I've reworked the example so that this problem is not surfaced. I prefer that over adding extra documentation for this corner case, which as you mentioned, is pre-existing and already applies to `commitBody`.

https://github.com/renovatebot/renovate/pull/44651/commits/aad8cb2a85895c7bd7867dfb0b7591885f71cf5f

felipecrs:
@RahulGautamSingh, thanks a lot for your review. My AI agent replied to some of your comments on my behalf without asking for permission, and I've deleted such comments. I apologize for it.

lib/config/validation.ts
RahulGautamSingh: Yeah, sorry I missed we can;t have compiled value during validation stage.

lib/util/git/index.ts
```diff
@@ -1452,7 +1454,16 @@ export async function prepareCommit({
       commitOptions['--no-verify'] = null;
     }
 
-    const commitRes = await git.commit(message, [], commitOptions);
+    let commitMessage = message;
+    if (isNonEmptyArray(trailers)) {
+      // simple-git joins message array elements with blank lines, so the
+      // trailers become the final block of the commit message
+      commitMessage = (
+        typeof message === 'string' ? [message] : message
```
RahulGautamSingh: isString from sindresorhus

RahulGautamSingh (approved):
nit

RahulGautamSingh (approved):

viceice (dismissed):

lib/util/git/index.ts
```diff
@@ -1452,7 +1454,16 @@ export async function prepareCommit({
...
 
-    const commitRes = await git.commit(message, [], commitOptions);
+    let commitMessage = message;
+    if (isNonEmptyArray(trailers)) {
+      // simple-git joins message array elements with blank lines, so the
+      // trailers become the final block of the commit message
+      commitMessage = (isString(message) ? [message] : message).concat(
+        trailers.join('\n'),
+      );
+    }
```
secustor:
```suggestion
    const commitMessage = isString(message) ? [message] : message
    if (isNonEmptyArray(trailers)) {
      commitMessage.push(trailers.join('\n'))
    }
```
Simplify

lib/workers/repository/onboarding/branch/create.ts
```diff
@@ -36,6 +37,15 @@ export async function createOnboardingBranch(
...
   }
 
+  // only allow gitAuthor template value in the commitTrailers
+  const trailers = config.commitTrailers
+    ? filterValidCommitTrailers(
+        config.commitTrailers.map((trailer) =>
+          compile(trailer, { gitAuthor: config.gitAuthor }),
+        ),
+      )
+    : undefined;
```
secustor:
```suggestion
  // only allow gitAuthor template value in the commitTrailers
  const compiledTrailers = coerceArray(config.commitTrailers).map((trailer) =>
    compile(trailer, { gitAuthor: config.gitAuthor }),
  );
  const trailers = filterValidCommitTrailers(compiledTrailers);
```
Make more readable

secustor (approved):

jamietanna:
There might have been a slight regression here: https://github.com/renovatebot/renovate/discussions/44895

felipecrs:
> There might have been a slight regression here: #44895

Interesting! I'll work on it today.

PR #44659, opened 2026-07-17
feat(presets): add `security:minimumReleaseAgeCrate` preset

Turbo87:
## Changes

This adds a `security:minimumReleaseAgeCrates` preset, based on the existing `security:minimumReleaseAgeNpm` preset, to delay crate updates for three days. The delay overlaps the initial 72-hour crates.io deletion window and gives malware scanners time to identify suspicious releases. Pin updates remain exempt because Renovate does not currently attach release timestamps to pin updates.

## Context

## AI assistance disclosure


lib/config/presets/internal/security.preset.ts
```diff
@@ -23,6 +23,26 @@ export const presets: Record<string, Preset> = {
...
+      {
+        internalChecksFilter: 'strict',
+        matchDatasources: ['crate'],
+        minimumReleaseAge: '3 days',
+      },
+      {
+        description: 'Do not require Minimum Release Age for package pinning',
+        matchDatasources: ['crate'],
+        matchUpdateTypes: ['pin'],
+        minimumReleaseAge: null,
```
viceice, line 59: why not using the update types. like go indirect dep updates? so we only need a single rule (perf)

Turbo87: this is copied over from `minimumReleaseAgeNpm` which does the same thing 🤔

jamietanna (commented):
Similar to the npm datasource, we've got a few other update types that are common in the Rust ecosystemt that would need to be excluded here

jamietanna:
> The delay overlaps the initial 72-hour crates.io deletion window

I didn't realise y'all had this, similar to npm. TIL :)

lib/config/presets/internal/security.preset.ts
Turbo87: I've updated this to use `matchUpdateTypes: ['major', 'minor', 'patch']` instead. let me know if this is what you had in mind.

jamietanna:
@viceice LMK your thoughts - I was happy with the previous setup, but interested to see if we need to improve the https://docs.renovatebot.com/presets-security/#securityminimumreleaseagenpm

Turbo87:
> I didn't realise y'all had this, similar to npm. TIL :)

yeah, we introduced crate deletions about two years ago (see https://github.com/rust-lang/rfcs/pull/3660) and modeled the rules roughly after the npm policy, which seemed like a good tradeoff between lowering the support team burden and avoiding leftpad incidents :)

viceice (dismissed):

jamietanna (commented):
@viceice do you think we should refactor `minimumReleaseAgeNpm` to do something similar?

Or maybe we refactor out a `minimumReleaseAgeBase` that goes through common `updateTypes` that don't have a `releaseTimestamp` and add the warning?

My concern is that folks enable this but don't then understand why they get hit by a supply chain attack via `lockFileMaintenance`

Turbo87:
is there anything more needed from me before we can merge this? :)

viceice:
> @viceice do you think we should refactor `minimumReleaseAgeNpm` to do something similar?
> 
> Or maybe we refactor out a `minimumReleaseAgeBase` that goes through common `updateTypes` that don't have a `releaseTimestamp` and add the warning?
> 
> My concern is that folks enable this but don't then understand why they get hit by a supply chain attack via `lockFileMaintenance`

concers are reasonable, maybe we should add those PR body warnings to all those rules. 🤔

jamietanna:
> to all those rules

You mean re-add the warnings, in a similar way to how we're doing it with npm?

Turbo87:
I don't want to be pushy, but does this cleanup of the existing presets need to block the introduction of this new one with semantics matching the existing ones?

jamietanna (changes requested):
Let's make sure we have the warnings for `updateTypes` that aren't supported at this time:

```typescript
// example via Claude Opus 4.8
minimumReleaseAgeCrates: {
  description:
    "Wait until the crates.io crate version is three days old before raising the update. This a) introduces a short delay to allow for malware researchers and scanners to (possibly) detect any malicious behaviour in crates, and b) reduces the risk of upgrading to a newly published crate that its owner deletes during crates.io's initial 72-hour deletion window.",
  packageRules: [
    {
      internalChecksFilter: 'strict',
      matchDatasources: ['crate'],
      minimumReleaseAge: '3 days',
    },
    {
      description:
        'Do not require Minimum Release Age for update types that are controlled by the package manager',
      matchDatasources: ['crate'],
      matchUpdateTypes: ['lockFileMaintenance'],
      minimumReleaseAge: null,
      prBodyNotes: [
        "⚠️ Renovate's lock file maintenance functionality does not support validating Minimum Release Age, as the package manager performs the required changes to update package(s). Confirm whether your package manager perform its own validation for the Minimum Release Age of packages.",
      ],
    },
    {
      description: 'Do not require Minimum Release Age for package replacements',
      matchDatasources: ['crate'],
      matchUpdateTypes: ['replacement'],
      minimumReleaseAge: null,
      prBodyNotes: [
        "⚠️ Renovate's replacement functionality [does not currently](https://github.com/renovatebot/renovate/issues/39400) wire in the release age for a package, so the Minimum Release Age checks can apply. You will need to manually validate the Minimum Release Age for these package(s).",
      ],
    },
    {
      description: 'Do not require Minimum Release Age for package pinning',
      matchDatasources: ['crate'],
      matchUpdateTypes: ['pin'],
      minimumReleaseAge: null,
      prBodyNotes: [
        "⚠️ Renovate's pin functionality [does not currently](https://github.com/renovatebot/renovate/issues/40288) wire in the release age for a package, so the Minimum Release Age checks can apply. You will need to manually validate the Minimum Release Age for these package(s).",
      ],
    },
  ],
},
```

It's important we keep these warnings - I'll follow-up to do a bit more of a refactor, which will look at making the warnings "on by default" for each of the different presets

(Related: https://github.com/renovatebot/renovate/discussions/40553 https://github.com/renovatebot/renovate/discussions/42610)

lib/config/presets/internal/security.preset.ts
```diff
@@ -23,6 +23,18 @@ export const presets: Record<string, Preset> = {
       },
     ],
   },
+  minimumReleaseAgeCrates: {
```
jamietanna: Also, please use singular:

```suggestion
  minimumReleaseAgeCrate: {
```

This then matches the datasource name exactly - thanks!

Turbo87: I'll do it, but it's called crates.io, not crate.io 😅

jamietanna (approved):
Thank you!

lib/config/presets/internal/security.preset.ts
jamietanna: For context, it's to stay in line with [the Datasource name, with is `crate`](https://docs.renovatebot.com/modules/datasource/crate/)

jamietanna (dismissed):

jamietanna:
@Turbo87 just to confirm, you're referring to `crates.io` - this change will affect _all_ users of the Crate datasource (including i.e. Artifactory, etc) - is that OK?

lib/config/presets/internal/security.preset.ts
```diff
@@ -23,6 +23,46 @@ export const presets: Record<string, Preset> = {
       },
     ],
   },
+  minimumReleaseAgeCrate: {
+    description:
+      "Wait until the crates.io crate version is three days old before raising the update. This a) introduces a short delay to allow for malware researchers and scanners to (possibly) detect any malicious behaviour in crates, and b) reduces the risk of upgrading to a newly published crate that its owner deletes during crates.io's initial 72-hour deletion window.",
```
jamietanna: To remove the `crates.io` reference, as this affects anything using this Datasource

```suggestion
      "Wait until the Crate version is three days old before raising the update. This a) introduces a short delay to allow for malware researchers and scanners to (possibly) detect any malicious behaviour in crates, and b) reduces the risk of upgrading to a newly published crate that its owner deletes during crates.io's initial 72-hour deletion window.",
```

jamietanna (approved):

Turbo87:
> just to confirm, you're referring to `crates.io` - this change will affect _all_ users of the Crate datasource (including i.e. Artifactory, etc) - is that OK?

that is a good question. for crates.io we have `pubtime` on all index entries now. from what I understand, if someone would enable this preset and use a custom registry that does not supply `pubtime` then they wouldn't get updates anymore due to https://docs.renovatebot.com/configuration-options/#minimumreleaseagebehaviour defaulting to `timestamp-required`.

do you think we should scope the preset to just crates.io?

jamietanna:
So the way we planned this for npm was similar - a user of the preset wants to be confident that they're requiring timestamps everywhere (rather than "only" npmjs.org) so we should make this apply to _all_ registries - any that don't have the release timestamp (`pubtime` in this case) will now have updates skipped, with relevant debug logs (+ as seen in Dependency Dashboard)

It's - IMO - a safer best practice than only opting in specific registries

Turbo87:
yeah, that sounds reasonable. `minimumReleaseAgeBehaviour` should allow people to override the behavior for custom registries, so it's not blocking them either.

would it be viable to enable this new preset by default in the `config:best-practices` preset? happy to send a PR :)

jamietanna:
If I'm honest - no, not at this point 😅 

We'd most likely only want to do that in a major version, given a significant change in behaviour

We'd also want to get some feedback - generally through user "I think I found a bug" reports - as to whether it's working as expected before rolling it out to everyone, as we did with `security:minimumReleaseAgeNpm`

We may change our stance on that at some point, but it's likely to want to wait until next major (maybe August/September)

Turbo87:
okay, thanks. I'll explicitly enable it for https://github.com/rust-lang/crates.io itself and see how it goes.

PR #44667, opened 2026-07-17
fix(gerrit): resolve relative submodule URLs for nested project paths

Tej128:
## Changes

`createHttpUrl()` in the Gerrit platform was using `encodeURIComponent(repository)` when constructing the git clone URL. This encoded `/` as `%2F`, making the URL constructor treat the entire project path as a single segment.

When the git-submodules manager later resolves relative URLs from `.gitmodules` using `new URL("../sibling", remoteUrl + "/")`, the `../` traversal fails because `%2F` is not treated as a path separator.

This removes `encodeURIComponent` and passes the repository string directly to `joinUrlParts`, matching how `createSshUrl` and other platforms handle it.

## Context

## AI assistance disclosure

The public repository: <URL>


felipecrs (approved):
@Tej128 is a colleague of mine and we are using this patch for a while now internally. I can confirm the issue is real and this solves it.

There will also be a follow-up patch on the git-submodules manager which is not parsing `./` (not `../`) submodules correctly.

felipecrs (approved):

viceice (approved):

Issue #39304, opened 2025-11-14
feat(globalExtends): allow `file://` paths


jamietanna:
As a means to provide self-hosting operators the ability to wire in additional configuration (separate to [the additional config file](https://docs.renovatebot.com/config-overview/#additional-file-config)) we will implement a `file://` URI handler for the `globalExtends` presets to resolve a preset from a given file.

This will only be available for self-hosting, and will be marked as experimental.

This is an alternative to the work in https://github.com/renovatebot/renovate/pull/39238

Implementation details of note:

- This **must** be an absolute path to a filename
- This is only allowed in `globalExtends`
- Using `file://` outside of `globalExtends` is a configuration _error_
- This will be marked as an experimental feature.

PR #44673, opened 2026-07-17
feat: allow to define custom presets in self-hosted config

felipecrs:
## Changes

Adds a `customPresets` global-only option, so self-hosted admins can define presets directly in their global config. Repositories can then extend from them with a `custom:` prefix:

```js
// config.js
module.exports = {
  customPresets: {
    myPreset: {
      description: 'My custom preset',
      labels: ['custom-label'],
    },
  },
};
```

```json
// renovate.json
{
  "extends": ["custom:myPreset"]
}
```

This was previously proposed in #40117 and discussed in #40119. This PR reproposes it with a few improvements on top.

### Why not just use `file://` presets (#39304)?

`globalExtends`/`extends` gaining `file://` support would get partway there, but `customPresets` has some advantages that are still worth having on top of it:

1. Simpler deployments: A single `config.js` can house both the global config and all custom presets. This eliminates the need to mount multiple files into the Renovate container, prevents repos from hardcoding fragile server-specific paths (`file:///...`), and bypasses filesystem permission and sandboxing issues entirely.

2. Dynamic generation: Since `customPresets` can be defined in the `config.js`, presets can be computed, composed, or fetched programmatically at startup, whereas `file://` can only return static file contents.

## Context

Related:

- #40117 (original PR, closed)
- #40119 (discussion where this was re-proposed)
- #39304 (related `file://` support for `globalExtends`)

## AI assistance disclosure

Used Claude Sonnet 5 to reimplement #40117 against the current codebase, adapt/extend the tests, and draft the docs.


felipecrs:
I believe this is in good shape. Tested via https://github.com/renovatebot/renovate/pull/44673/commits/7692ec6a7296795b43ddd47e7953b263ec002c4a.

docs/usage/self-hosted-configuration.md
```diff
@@ -487,6 +487,31 @@ The above configuration approach will mean the values are redacted in logs like
...
+  "customPresets": {
+    "myPreset": {
+      "description": ["My custom preset"],
+      "labels": ["custom-label"]
+    }
+  }
+}
+```
+
+Then in a repository's `renovate.json`:
```
RahulGautamSingh:
```suggestion
Then in the `renovate.json` of a repository:"
```

RahulGautamSingh (commented):
`customPresets` is excluded from secret interpolation, so `{{ secrets.X }}` inside a custom preset is left as a literal and only resolves later in repo context (repo secrets, not global secrets). 

This is similar to how `onboardingConfig` and maybe that's why you chose it, but it means a self-hosted admin cannot use global secrets inside a custom preset. 

Confirming if that's the intended behavior and if so we should mention it in the docs, since a global-config option that can't use global secrets is a little surprising.

felipecrs:
Thanks a lot for the review, @RahulGautamSingh!

> `customPresets` is excluded from secret interpolation, so `{{ secrets.X }}` inside a custom preset is left as a literal and only resolves later in repo context (repo secrets, not global secrets).
> 
> This is similar to how `onboardingConfig` and maybe that's why you chose it, but it means a self-hosted admin cannot use global secrets inside a custom preset.
> 
> Confirming if that's the intended behavior and if so we should mention it in the docs, since a global-config option that can't use global secrets is a little surprising.

In fact, global secrets do resolve at repository-time expansion. I've confirmed this via https://github.com/renovatebot/renovate/pull/44673/commits/3095e128aac56ba9ef3f5985c38b833d13580882.

Maybe I'm missing something?

felipecrs:
What I've [intentionally chosen not to have](https://github.com/renovatebot/renovate/pull/44673/commits/671ee301e1da613af6635aecb44e5fd6db1ee44a) is support for global-only options because I believe that can be done as a follow-up discussion/PR.

---

Anyway, I'm happy to discuss that:

If `customPresets` supported global-only options, this would expand the functionality of presets as nowadays no shareable preset can define global-only options.

It would fulfill a similar role as repository-objects in the global `repositories` array, where global-only options can also be set.

Imagine for example:

1. An administrator can define a preset like:

	```json
	{
	  "customPresets": {
	    "trustGradleWrapper": {
	      "allowedUnsafeExecutions": ["gradleWrapper"]
	    }
	  }
	}
	```
	
2. Users can then opt into that via `extends: ["custom:trustGradleWrapper"]`.

Which means: as an administrator, I am okay with giving users the choice of running gradle wrapper. But I want to make it so users need to explicitly opt-in so they hopefully understand the security implications.

Maybe there are other, more interesting, use cases.

Though there could be downsides or additional security implications which I haven't thought of yet.

And perhaps @jamietanna would be interested in joining this discussion. 😊

docs/usage/self-hosted-configuration.md
```diff
@@ -487,6 +487,31 @@ The above configuration approach will mean the values are redacted in logs like
...
+    }
+  }
+}
+```
+
+Then in the `renovate.json` of a repository:
+
+```json
+{
+  "extends": ["custom:myPreset"]
```
jamietanna: I would prefer this be denoted much more differently to an internal preset, to make it clear that it's _not_ a Renovate preset called `custom:` but that it's maybe `admin>` or `self-hosted>` or `global>` or similar

jamietanna (changes requested):
I'll review when I get some more time

I'm still not quite sure about the shape and interface (see comment) and this will lead to additional complexity for folks who want to use `renovate-config-validator` that we should consider

I'd probably want this marked as `experimental: true`, too

docs/usage/self-hosted-configuration.md
felipecrs: That makes sense. I'm okay with your suggestion, but how about `custom>`?

Reason is to make a reference to the word "custom" in `customPresets`. Similar to `custom.regex`, `custom.jsonata` (`customManagers`) and `custom.k3s` (`customDatasources`).

Or perhaps even `custom.`? E.g. `custom.myPreset`.

I guess that would be even more consistent, and would be denoted very differently from regular presets.

felipecrs:
> this will lead to additional complexity for folks who want to use `renovate-config-validator` that we should consider

Just FYI, this PR accounts for `renovate-config-validator`. I already use it internally to validate my `customPresets` before merging.

felipecrs:
> I'd probably want this marked as experimental: true, too

Done. Completely agree with that.

docs/usage/self-hosted-configuration.md
felipecrs: Implemented `custom.` via https://github.com/renovatebot/renovate/pull/44673/commits/43a895b5d499d5f4a09c441513b8f51c41afc546.

Not a problem if you'd like something different. Just let me know.

viceice:
conflicted

PR #44683, opened 2026-07-17
feat(report): surface additional repository metadata in report

justfalter:
## Changes

Surfaces additional repository metadata in the report generated via `reportType`
(`logging`/`file`/`s3`), to make reports more self-describing without needing to
cross-reference the run that produced them.

Per repository, the report now includes `platform`, `endpoint`, `defaultBranch`,
`dependencyDashboardIssue` (or `null`), `configFileName`, `repoIsOnboarded`, and
`onboardingPrNumber` (or `null`). Each branch entry now also carries its
`baseBranch`.

Rough shape of the additions (marked with `// +`):

```jsonc
{
  "repositories": {
    "some-org/some-repo": {          // the key is already the repository identifier
      "problems": [],
      "branches": [
        {
          "baseBranch": "main",                       // +
          "branchName": "renovate/actions-checkout-6.x",
          "prNo": 105,
          "prTitle": "Update actions/checkout action to v6.0.3",
          "result": "pr-created",
          "upgrades": [ /* unchanged */ ]
        }
      ],
      "packageFiles": { /* unchanged */ },
      "libYearsWithStatus": { /* unchanged */ },
      "platform": "github",                            // +
      "endpoint": "https://api.github.com/",           // +
      "defaultBranch": "main",                         // +
      "dependencyDashboardIssue": 6,                   // + (null when none exists)
      "configFileName": ".github/renovate.json5",      // +
      "repoIsOnboarded": true,                         // +
      "onboardingPrNumber": null                       // + (PR number when an unmerged onboarding PR exists)
    }
  }
}
```

Onboarding status is derived in `finalizeRepo` from the already-fetched PR list,
so no additional platform API calls are made. `platform`/`endpoint` are read from
`GlobalConfig`; the remaining values come from the repo config, cache, and PR list.

## Context

## AI assistance disclosure

AI (Claude Code, model Claude Opus 4.8) was used to author the implementation,
unit tests, and this PR description, under human direction and review.

The public repository: https://github.com/justfalter/renovate-test-repo

Verified in this run (report visible in the job logs under `Printing report`):
https://github.com/justfalter/renovate-test-repo/actions/runs/29619043257

### DIFF
--- a/lib/instrumentation/types.ts
+++ b/lib/instrumentation/types.ts
@@ -1,6 +1,7 @@
 import type { Attributes, SpanKind, SpanOptions } from '@opentelemetry/api';
 import type { ATTR_CODE_FUNCTION_NAME } from '@opentelemetry/semantic-conventions';
 import type { RenovateSplit } from '../config/types.ts';
+import type { PlatformId } from '../constants/index.ts';
 import type { BunyanRecord } from '../logger/types.ts';
 import type { PackageFile } from '../modules/manager/types.ts';
 import type { BranchCache } from '../util/cache/repository/types.ts';
@@ -54,6 +55,35 @@ interface RepoReport {
   branches: Partial<BranchCache>[];
   packageFiles: Record<string, PackageFile[]>;
   libYearsWithStatus?: LibYearsWithStatus;
+  /**
+   * The platform type of the repository (ex: `github`, `gitlab`).
+   */
+  platform?: PlatformId;
+  /**
+   * The platform API endpoint used for the repository.
+   */
+  endpoint?: string;
+  /**
+   * The default branch name of the repository.
+   */
+  defaultBranch?: string;
+  /**
+   * The issue number of the Dependency Dashboard, or `null` if none exists.
+   */
+  dependencyDashboardIssue?: number | null;
+  /**
+   * Path to the repository's Renovate config file (ex: `.github/renovate.json5`).
+   * Empty when the repository has no in-repo config file.
+   */
+  configFileName?: string;
+  /**
+   * Whether the repository has already been onboarded (has a merged Renovate config).
+   */
+  repoIsOnboarded?: boolean;
+  /**
+   * The number of the open (unmerged) onboarding PR, or `null` if none exists.
+   */
+  onboardingPrNumber?: number | null;
 }
 
 export interface LibYearsWithStatus {
--- a/lib/instrumentation/types.ts
+++ b/lib/instrumentation/types.ts
@@ -73,7 +73,7 @@ interface RepoReport {
   dependencyDashboardIssue?: number | null;
   /**
    * Path to the repository's Renovate config file (ex: `.github/renovate.json5`).
-   * Empty when the repository has no in-repo config file.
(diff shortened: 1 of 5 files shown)


jamietanna:
Out of interest, how are you using this? What additional functionality does this give you?

justfalter:
I've been working on some reporting tooling, inspired a bit by the existing Backstage-Renovate integration. It's mostly to help our security folks, and also give me a way to automate the hounding of repo owners. 

All of our Renovate scans kick off via GHA, but I've wired them up to send reports to an s3 bucket. The additional metadata will allow me to link to dependency dashboards, link to the renovate config file (defaultBranch), and catch situations where repos have not been onboarded completely. The baseBranch allows me to know where the changes are going to merge into, and the platform/endpoint just seemed like the right thing to do for all of humanity.

lib/instrumentation/types.ts
```diff
@@ -54,6 +55,35 @@ interface RepoReport {
...
+   * The default branch name of the repository.
+   */
+  defaultBranch?: string;
+  /**
+   * The issue number of the Dependency Dashboard, or `null` if none exists.
+   */
+  dependencyDashboardIssue?: number | null;
+  /**
+   * Path to the repository's Renovate config file (ex: `.github/renovate.json5`).
+   * Empty when the repository has no in-repo config file.
```
jamietanna:
```suggestion
   * Omitted when the repository has no in-repo config file.
```

lib/instrumentation/types.ts
```diff
@@ -54,6 +55,35 @@ interface RepoReport {
...
   packageFiles: Record<string, PackageFile[]>;
   libYearsWithStatus?: LibYearsWithStatus;
+  /**
+   * The platform type of the repository (ex: `github`, `gitlab`).
+   */
+  platform?: PlatformId;
+  /**
+   * The platform API endpoint used for the repository.
+   */
+  endpoint?: string;
```
jamietanna, line 65: This shouldn't include secrets, but I wonder if it's worth creating a prefactor PR (before this one goes in) to make sure that we call `sanitize()` before we write out the result of the report, just in case 🤔

justfalter: You're imagining that someone could have basic auth (or something else) configured as part of their platform endpoint URL a la `https://myusername:mypassword@some.host.com/api/v3`?


While `https://**redacted**@some.host.com/api/v3` might work for logging, it can really goof things up for machine-readable data structures. For reporting purposes, we really don't need to care about whether something is redacted or not.

Would it be OK if I just focus on the sanitization of the `endpoint` property for this PR? I don't think there's anything else at the moment that needs it...

I would take the following approach:
1. Parse the `endpoint` as a URL, retaining only the scheme, hostname, port, and path (ex: `https://some.host.com:1234/api/v3`.
2. Run it through  stuff in `util/sanitize.ts` `sanitize(...)`
3. If we detect that the value has changed as a result of sanitization, we emit a warning message to the log and omit the endpoint value. 

----

I'm also open to excluding `endpoint` entirely  - it is really just a nice-to-have.

PR #44736, opened 2026-07-20
feat(azure): make issue work item type configurable

### LINKED ISSUE #44643: Azure: make Work Item configurable
As found in recent Azure-related Discussions, the creation of the Dependency Dashboard depends on the use of the Work Item `Issue` (https://github.com/renovatebot/renovate/pull/44592)

Via https://github.com/renovatebot/renovate/blob/b32c6c62faac353e54d87d251dc31e3b23f01c59/lib/modules/platform/azure/issue.ts#L14 we're hardcoding `Issue` type, but should make it configurable for projects that use a different Work Item type.

This would be repo-level config, as it's up to the repository owner themselves to know what they're using.

RahulGautamSingh:
## Changes

Azure DevOps has no native "issue" concept, so Renovate creates a Work Item for its issues (e.g. the Dependency Dashboard). The Work Item type was hardcoded to `Issue`, which only exists in the `Basic` process. Projects using another process (`Agile`, `Scrum`, or a custom inherited process) may not have an `Issue` type, so creation failed and the Dependency Dashboard could not be created.

This adds a repo-level `azureWorkItemType` config option (default `Issue`) so users can tell Renovate which Work Item type to create its issues as.

- New `azureWorkItemType` option (`supportedPlatforms: ['azure']`), threaded to the Azure platform via `EnsureIssueConfig` — the same carrier pattern already used for `azureWorkItemId`.
- Azure now creates the Work Item, and resolves its open/closed states, using the configured type.
- The Work Item lookup query no longer filters by `[System.WorkItemType]`. A repo's Renovate issues have a unique title, so matching on title + project finds them regardless of which type they were created as — this keeps find/close working without changing the shared `ensureIssueClosing` interface.

## Context

- [x] This closes an existing Issue, Closes: #44643

## AI assistance disclosure

AI (Claude, Opus 4.8) was used to help write the code, unit tests, and documentation. All changes were reviewed by a human.

The public repository: N/A


jamietanna (commented):
Looks reasonable - can we please make sure we test this in an Azure DevOps project that doesn't have an `Issue`?

RahulGautamSingh:
> Looks reasonable - can we please make sure we test this in an Azure DevOps project that doesn't have an `Issue`?

https://dev.azure.com/reno-test/reno-one/_workitems/edit/1/

Created a repo, and in the repo config added `azureWorkItemType:"Task"`. As expected, Dependency Dashboard was created with `Task` instead of `Issue`

viceice (dismissed):
I really want to group those platform options below a single platform option object.

jamietanna:
In user-facing config, internal config, or both?

jamietanna (dismissed):
Thanks!

lib/modules/platform/azure/issue.ts
```diff
@@ -52,6 +52,9 @@ function namesByCategory(
 export class IssueService {
   private config: Config;
   private readonly workItemStates: Lazy<Promise<WorkItemStates>>;
+  // Work item type to create issues as. Repo-configurable via `azureWorkItemType`
+  // and set from `ensureIssue`; defaults to the historical `Issue` type.
+  private workItemType = defaultWorkItemType;
```
jamietanna: did we think about whether this could get set in `Config`?

RahulGautamSingh: checking

jamietanna (commented):
Via Claude Opus 4.8:

````markdown
### Main concern — mutable `workItemType` + cached `Lazy` state resolution (ordering fragility)

`this.workItemType` is only ever assigned inside `ensureIssue` (issue.ts:230). But `this.workItemStates` is a `Lazy` that resolves **once** against whatever `this.workItemType` is at the *first* `getValue()` call (issue.ts:81-84), and every subsequent caller reuses that cached result.

- `findIssue` / `getIssueList` / `ensureIssueClosing` never receive or set the configured type.
- `ensureIssueClosing` is called in flows that can run *before* `ensureIssue` in a given repo pass — e.g. `onboarding/branch/check.ts:136,142` and `finalize/index.ts:47-48`, and the dashboard-disabled path (`dependency-dashboard.ts:376,442`) which calls `ensureIssueClosing` *instead of* `ensureIssue`.

**Failure scenario:** a repo configured with `azureWorkItemType: Task` where `ensureIssueClosing` (or `getIssueList`) triggers `workItemStates.getValue()` first → states are resolved against the default `Issue` type and cached. A later `ensureIssue` sets `this.workItemType = 'Task'` but the stale `Issue` states are reused for open/close operations. Real-world impact is limited because state resolution is category-based (Proposed/InProgress/Completed) and most types within a process share the same category→name mapping — but it is a genuine latent inconsistency, and it means "close the Task dashboard when disabled" resolves the close state against `Issue`.

Since `azureWorkItemType` is repo-constant (the comment at issue.ts:229 even says so), a mutable field set on one method is the fragile way to carry it. Cleaner options:
- Add `workItemType` to the Azure `Config` and set it in `initRepo` (issue.ts:59 constructor reads it once) — then all methods resolve against the same value regardless of call order; or
- At minimum, also pass/set the type on the closing path so `ensureIssueClosing`'s state resolution matches.

Not a blocker, but I'd raise it with the author.
````

Thoughts?

File: lib/modules/platform/azure/issue.ts [hunk 2]
```diff
@@ -331,13 +345,14 @@ export class IssueService {
...
+      // The configured work item type (default `Issue`) only exists in some
+      // processes (Basic, Agile, CMMI) but not others (e.g. Scrum). Creating one
+      // on a process without it returns a 404 that the REST client surfaces as
+      // `null`, so check first and log an actionable message instead of failing
+      // cryptically.
       if (!(await this.hasWorkItemType(azureApiWit))) {
         logger.warn(
-          { workItemType, project: this.config.project },
+          { workItemType: this.workItemType, project: this.config.project },
           'Azure: work item type does not exist in project (or the token lacks permission to it); skipping issue. The Dependency Dashboard needs a process that defines this work item type.',
```
jamietanna: Can we please update this to note that this can be configured using `azureWorkItemType` in repo config? and/or link to `documentationUrl: https://docs.renovatebot.com/configuration-options/#azureworkitemtype`

RahulGautamSingh: Can;t add the docs URL, `no-hardcoded-docs-url` lint rule forbids it

lib/modules/platform/azure/issue.ts
RahulGautamSingh: done

lib/modules/platform/azure/issue.ts
jamietanna: Can't you use `productLinks` to inject the docs site URL?

lib/modules/platform/azure/issue.ts
```diff
@@ -331,14 +335,18 @@ export class IssueService {
...
+      // cryptically.
       if (!(await this.hasWorkItemType(azureApiWit))) {
         logger.warn(
-          { workItemType, project: this.config.project },
-          'Azure: work item type does not exist in project (or the token lacks permission to it); skipping issue. The Dependency Dashboard needs a process that defines this work item type.',
+          {
+            workItemType: this.config.workItemType,
+            project: this.config.project,
+          },
+          'Azure: work item type does not exist in project (or the token lacks permission to it); skipping issue. The Dependency Dashboard needs a process that defines this work item type. Set one your project defines via the `azureWorkItemType` repo config option.',
```
jamietanna, line 352: Do we have the list of Work Items available in the project here? We could always log them out here too, so it's more actionable?

lib/modules/platform/azure/index.ts
```diff
@@ -226,6 +227,7 @@ export async function initRepo({
 
   config.project = repo.project!.name!;
   config.projectId = repo.project!.id!;
+  config.workItemType = azureWorkItemType ?? defaultWorkItemType;
```
jamietanna: Should we instead get the default for the config option? Then we can make sure that it stays in sync with the actual default

lib/modules/platform/azure/issue.ts
RahulGautamSingh: done

RahulGautamSingh:
Need to re-test on a reproduction, after recent changes. Please do not merge.

RahulGautamSingh:
Tested again and found an edge case: changing `azureWorkItemType` on an existing Dependency Dashboard

**Repro:**
1. Create a DD with `azureWorkItemType: "Task"` → work item created as `Task`.
2. Close it, then change config to `azureWorkItemType: "Bug"`.
3. Next run: instead of creating a new `Bug` work item, Renovate finds and reopens the existing `Task` one.

**Cause:** 
`getIssueList` intentionally queries by `title + project` only, not by `[System.WorkItemType]`, because the type is now configurable, and matching on type would miss existing issues created under a different type. 

The flip side is that a type change on an existing DD reuses the old work item rather than creating one of the new type.

So the type is effectively fixed at DD-creation time; changing it later has no effect until the old work item is gone.

**Two ways to handle it:**

- Document it: treat `azureWorkItemType` as set-once per DD. If a user wants the new type, they delete/close-and-remove the old work item and Renovate recreates it. 
- Handle it internally: when the found work item's type doesn't match the configured one, close the old one and create a fresh work item of the new type. Better, but it adds churn and an extra field (`System.WorkItemType`) to the query results, and edges toward "Renovate manages work-item-type migrations" which may be more than we want.

I prefer documenting it. It's an uncommon action, self-correcting once the old item is removed, and the alternative solution adds extra handling for a case.

Thoughts?? @jamietanna @secustor 

_PS. AI used for re-wording_

jamietanna:
Yep, IMO document it as an edge case 👍🏼 If it really is a concern we're running into a lot, we can look at working around it at that point

jamietanna:
Happy for me to merge?

lib/modules/platform/azure/issue.ts
```diff
@@ -128,27 +127,31 @@ export class IssueService {
...
+   * that *are* available) instead of a cryptic error.
    */
-  private async hasWorkItemType(
+  private async getWorkItemTypeNames(
     azureApiWit: IWorkItemTrackingApi,
-  ): Promise<boolean> {
+  ): Promise<string[]> {
     const types = await azureApiWit.getWorkItemTypes(this.config.project);
-    return types.some((t) => t.name === workItemType);
+    return types.map((t) => t.name).filter((name): name is string => !!name);
```
secustor:
```suggestion
    return types.map((t) => t.name).filter(isNonEmptyString);
```

secustor (commented):
Else LGTM

jamietanna:
Will merge once CI is fixed 👍🏼

granjacours697-netizen (approved):
Arbitrary command

viceice:
> In user-facing config, internal config, or both? 

both

docs/usage/configuration-options.md
```diff
@@ -344,6 +344,15 @@ If you're not already using `bors-ng` or similar, don't worry about this option.
 When creating a PR in Azure DevOps, some branches can be protected with branch policies to [check for linked work items](https://learn.microsoft.com/azure/devops/repos/git/branch-policies#check-for-linked-work-items).
 Creating a work item in Azure DevOps is beyond the scope of Renovate, but Renovate can link an already existing work item when creating PRs.
 
+## `azureWorkItemType`
```
viceice, line 347: document the edge case resolutions 

- rename work item and close
- close and delete

viceice (approved):

jamietanna (approved):

PR #44744, opened 2026-07-20
feat(manager/nuget): pass defaultRegistryUrls to NuGet artifact updates to support private feeds

loganprice:
## Changes

This PR modifies the NuGet manager's artifact/lockfile update process to respect the existing `defaultRegistryUrls` configuration option instead of falling back to the hardcoded `api.nuget.org` registry feed.
Specifically:

- Exposes `defaultRegistryUrls` to manager-level artifact updaters in `UpdateArtifactsConfig` (`lib/modules/manager/types.ts`).
- Passes down the resolved `defaultRegistryUrls` from the first updated dependency when invoking manager-level restores in `get-updated.ts`.
- Updates `createCachedNuGetConfigFile` in `lib/modules/manager/nuget/artifacts.ts` to check for `config.defaultRegistryUrls`. If configured, it will map these to NuGet package sources, completely bypassing the default fallback to the public `api.nuget.org` registry.
This allows self-hosted Renovate deployments to run cleanly in airgapped environments (using internal mirrors like Artifactory) without encountering slow network connection timeouts on `api.nuget.org` during `dotnet restore` execution, and without requiring any new configuration options to be added.

## Context

## AI assistance disclosure

**Details**:
This pull request was co-authored with **Antigravity**, a coding assistant developed by Google DeepMind running the **Gemini 3.5 Flash (Medium)** model. The AI was used to research the codebase structure, design the non-breaking implementation flow, draft the type/worker changes, write the new unit tests, and format documentation.

The public repository: <URL>

### DIFF
--- a/lib/modules/manager/nuget/artifacts.ts
+++ b/lib/modules/manager/nuget/artifacts.ts
@@ -40,9 +40,14 @@ async function createCachedNuGetConfigFile(
   nugetCacheDir: string,
   packageFileName: string,
   updatedDeps: Upgrade[],
+  config: UpdateArtifactsConfig,
 ): Promise<string> {
+  const fallbackRegistries = config.defaultRegistryUrls
+    ? config.defaultRegistryUrls.map((url) => ({ url }))
+    : getDefaultRegistries();
+
   const registries =
-    (await getConfiguredRegistries(packageFileName)) ?? getDefaultRegistries();
+    (await getConfiguredRegistries(packageFileName)) ?? fallbackRegistries;
 
   const updatedDepsRegistries: Registry[] = Array.from(
     new Set(
@@ -76,6 +81,7 @@ async function runDotnetRestore(
     nugetCacheDir,
     packageFileName,
     updatedDeps,
+    config,
   );
 
   const dotnetVersion =
--- a/lib/modules/manager/nuget/artifacts.spec.ts
+++ b/lib/modules/manager/nuget/artifacts.spec.ts
@@ -515,4 +515,33 @@ describe('modules/manager/nuget/artifacts', () => {
     ]);
     expect(execSnapshots).toBeEmptyArray();
   });
+
+  it('respects defaultRegistryUrls', async () => {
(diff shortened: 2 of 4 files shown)


viceice (commented):
@renovatebot/maintainers Opinions? I'm usure about this change

lib/modules/manager/nuget/artifacts.ts
```diff
@@ -40,9 +40,14 @@ async function createCachedNuGetConfigFile(
   nugetCacheDir: string,
   packageFileName: string,
   updatedDeps: Upgrade[],
+  config: UpdateArtifactsConfig,
 ): Promise<string> {
+  const fallbackRegistries = config.defaultRegistryUrls
+    ? config.defaultRegistryUrls.map((url) => ({ url }))
+    : getDefaultRegistries();
```
secustor, line 47: How is this different from 

https://github.com/renovatebot/renovate/pull/44744/changes#diff-a731713d229124948c6029489e95159163df1422e03d6804a2b7f3e6fe499940R52-R59 ? 

Outside of the current implementation is checking ALL dependencies.

loganprice: From looking at the existing code they serve two different purposes and levels of configuration:

- fallbackRegistries (lines 45–47)  is the baseline default registry used only when the repository does not have its own nuget.config file. It replaces the hardcoded api.nuget.org fallback with config.defaultRegistryUrls. If the repo already defines a nuget.config, fallbackRegistries is ignored.

- updatedDepsRegistries (lines 52–59) is for dependency-specific feeds defined on individual packages via Renovate packageRules using registryUrls. These are always merged into the generated nuget.config so dotnet restore can locate private feeds configured for specific dependencies being upgraded, regardless of whether the repo has an existing nuget.config. Which is great if the lockfile is being updated during a version upgrade but  in the case of a standalone lockFileMaintenance run, updatedDeps is passed as an empty array so no registries would get added.

PR #44781, opened 2026-07-22
perf(package-rules): avoid re-cloning invariant packageRules array on every rule match

### LINKED ISSUE #44459: perf(package-rules): avoid re-cloning invariant packageRules array on every rule match in applyPackageRules
### Discussed in https://github.com/renovatebot/renovate/discussions/44444

<sup>Originally posted by **eyalzek** July  9, 2026</sup>

Self-hosted Renovate CLI (official `renovate/renovate` image) on Kubernetes.

### Which platform are you running Renovate on?

GitHub.com

43.251.3 (behavior observed consistently across 43.200.0 → 43.255.x)

**Summary:** on a large monorepo, ~88% of the whole Renovate run's CPU goes into `applyPackageRules` — dominated by `klona` deep-clones via `mergeChildConfig` — executed once per *matching rule* per *dependency instance*. Because `vulnerabilityAlerts` compiles every open Dependabot alert into force package rules (with full advisory markdown in `prBodyNotes`) that join the per-dependency config, the effective cost scales roughly **O(open alerts × package rules × dependency instances)**. On our repo this makes the lookup phase run for hours, and — worse than just being slow — the sustained synchronous clone/match churn **starves the event loop for 60–90s at a stretch**, so in-flight sockets get closed by peers (`socket hang up` on GraphQL POSTs, TLS handshakes timing out). A single such failure on a github datasource escalates to `ExternalHostError`, which aborts the whole repository run before any PRs are created.

**Repo shape:** monorepo with ~400 package files (~180 `go.mod`, ~200 `package.json`, plus Dockerfiles/workflows), i.e. tens of thousands of dependency *instances* (the same transitive Go dep appears in ~180 files and is fully re-processed per file). Several thousand open Dependabot alerts (large majority non-actionable transitive go-module advisories).

**CPU profile** (3-minute capture via inspector during the lookup phase, totals of self-time):

| self time | share | function | location |
|---|---|---|---|
| ~120s | 66% | `klona` | `klona/json` via `clone()` |
| ~35s | 19% | `matchesRule` | `util/package-rules` |
| ~4s | 2.3% | `applyPackageRules` | `util/package-rules` |
| ~8s | 4.6% | (garbage collector) | |

Every hot `klona` call site has the same stack: `applyPackageRules → mergeChildConfig → clone (klona)`.

**Observed impact measurements:**

- Disabling `vulnerabilityAlerts` alone made dependency processing **~20× faster** on this repo (same run otherwise: caches, node size, config) — consistent with the alert-derived rules being the biggest object in every clone.
- With alerts enabled, throughput was ~1 dependency-instance/second of pure local CPU (single core pinned, network mostly idle) — even for deps that end up `enabled: false` and need no lookups at all.
- It's self-amplifying: when runs abort, no update/security PRs merge, so open alerts accumulate, so the per-dep cost grows, so the next run is slower. Our alert count roughly 2.5×'d in the weeks after runs started failing.
- Raising the heap (`--max-old-space-size`) removed GC pressure (we measured `mu = 0.006` when running against the default 4GB cap) but the event-loop stalls persist — it's compute, not memory.

**Why it's fatal and not just slow:** the multi-minute event-loop stalls kill unrelated in-flight requests; a failed `POST /graphql` (not retried by got) from `github-tags`/`github-releases` becomes `ExternalHostError`, and for onboarded repos `lib/workers/repository/process/fetch.ts` propagates that as a repo abort (`external-host-error`) with no config opt-out. So past a certain repo size × alert count, runs abort ~100% of the time.

**Suggested directions** (happy to help with a PR if maintainers agree on an approach):

1. **Memoize/deduplicate rule application per identical input** — the same dependency (name, datasource, versioning, essentially identical file-level config) is fully re-evaluated for each of the ~180 package files that contain it. A cache keyed on the inputs of `applyPackageRules` would cut our instance count ~two orders of magnitude.
2. **Avoid deep-cloning the full config per matching rule** — `mergeChildConfig(config, rule)` clones the entire accumulated config (which itself contains the complete `packageRules` array, including the alert-derived rules) once per matching rule. Options: strip `packageRules` (and other large, invariant keys) before the per-rule merge loop and reattach after; or accumulate matched rules and merge once.
3. **Keep advisory bodies out of the per-dependency config path** — the alert-derived force rules embed full advisory markdown (`prBodyNotes`) that is only needed at PR-creation time; storing bodies out-of-band and referencing them by id would shrink the cloned object dramatically for repos with many alerts.

I can share the full `.cpuprofile` and put together a synthetic public reproduction (N go.mod files sharing M deps + a large generated packageRules array) if that helps triage.

### Logs (if relevant)

Representative sequence during a stall (log output drops from ~5 lines/s to ~1 line/20-30s, then all failures land in one burst):

```
DEBUG: Dependency: go.opentelemetry.io/otel/metric, is disabled   (per-instance evaluations 20-30s apart)
...
DEBUG: POST https://api.github.com/graphql = (code=ECONNRESET, statusCode=-1 retryCount=0, duration=128272)
DEBUG: Shrinking GitHub GraphQL page size after error
WARN:  Error obtaining docker token: Timeout awaiting 'request' for 60000ms
WARN:  Host error (github-tags): Client network socket disconnected before secure TLS connection was established
INFO:  Repository finished  "result": "external-host-error"
```</div>

### PR DESCRIPTION
**Pull request by eyalzek**
## Changes

`applyPackageRules()` merges each matching rule into the working config via `mergeChildConfig()`, which deep-clones the entire accumulated config on every call — including the invariant `packageRules` array itself. On repos with many dependency instances and large rule arrays (e.g. `vulnerabilityAlerts` compiling every open advisory into a forced rule), this cloning dominates the whole Renovate run — observed at ~66% of total CPU in a production profile.

This PR excludes `packageRules` from the working config while rules are applied and restores it afterwards, preserving the previous merge semantics for:
- rules that themselves carry nested `packageRules` (still merged/concatenated onto the restored array),
- `null` / empty / absent `packageRules` inputs (round-tripped unchanged),
- external callers that observe the returned config (the array reference survives the round-trip).

No behavior change; strictly a hot-path allocation removal.

## Context

- [x] This closes an existing Issue, Closes: #44459

## AI assistance disclosure

Claude Fable 5 (via Claude Code) was used to co-author the implementation and the added unit tests, based on a CPU profile and hypothesis I supplied. All final wording, review, and validation are mine.

New tests in `lib/util/package-rules/index.spec.ts` cover the round-trip guarantee (rules preserved), nested-rules concatenation, and null/empty input handling. The change was also validated on a large private monorepo where it eliminated the applyPackageRules hot path from the CPU profile (issue #44459 has the original profile).

The public repository: n/a (private profile source; behavior fully covered by the new specs)

### DIFF
--- a/lib/util/package-rules/index.ts
+++ b/lib/util/package-rules/index.ts
@@ -38,6 +38,12 @@ export async function applyPackageRules<T extends PackageRuleInputConfig>(
 ): Promise<T> {
   let config = { ...inputConfig };
   const packageRules = config.packageRules ?? [];
+  // The `packageRules` array is invariant while rules are being applied, and
+  // can be very large (e.g. vulnerability alerts append rules embedding full
+  // advisory texts). Remove it from the working config so mergeChildConfig()
+  // does not deep-clone the whole array once per matched rule, then restore
+  // it afterwards.
+  delete config.packageRules;
   logger.trace(
     { dependency: config.depName, packageRules },
     `Checking against ${packageRules.length} packageRules`,
@@ -106,6 +112,14 @@ export async function applyPackageRules<T extends PackageRuleInputConfig>(
       delete toApply.overridePackageName;
       config = mergeChildConfig(config, toApply);
     }
+  }
+  // Restore the rules. If any applied rule carried nested `packageRules`
+  // (e.g. from a resolved preset), preserve the concat-merge semantics that
+  // mergeChildConfig() would previously have applied.
+  if (config.packageRules) {
+    config.packageRules = packageRules.concat(config.packageRules);
+  } else if ('packageRules' in inputConfig) {
+    config.packageRules = inputConfig.packageRules;
   }
   return config;
 }
--- a/lib/util/package-rules/index.ts
+++ b/lib/util/package-rules/index.ts
@@ -39,6 +39,12 @@ export async function applyPackageRules<T extends PackageRuleInputConfig>(
 ): Promise<T> {
   let config = { ...inputConfig };
   const packageRules = coerceArray(config.packageRules);
+  // The `packageRules` array is invariant while rules are being applied, and
+  // can be very large (e.g. vulnerability alerts append rules embedding full
+  // advisory texts). Remove it from the working config so mergeChildConfig()
+  // does not deep-clone the whole array once per matched rule, then restore
+  // it afterwards.
+  delete config.packageRules;
   logger.trace(
     { dependency: config.depName, packageRules },
     `Checking against ${packageRules.length} packageRules`,
@@ -107,6 +113,14 @@ export async function applyPackageRules<T extends PackageRuleInputConfig>(
       delete toApply.overridePackageName;
       config = mergeChildConfig(config, toApply);
     }
+  }
+  // Restore the rules. If any applied rule carried nested `packageRules`
+  // (e.g. from a resolved preset), preserve the concat-merge semantics that
+  // mergeChildConfig() would previously have applied.
+  if (config.packageRules) {
+    config.packageRules = packageRules.concat(config.packageRules);
+  } else if ('packageRules' in inputConfig) {
+    config.packageRules = inputConfig.packageRules;
   }
   return config;
 }
--- a/lib/util/package-rules/index.spec.ts
+++ b/lib/util/package-rules/index.spec.ts
@@ -1530,4 +1530,56 @@ describe('util/package-rules/index', () => {
     const res = await applyPackageRules(config);
     expect(res.sourceUrl).toBe('https://github.com/hashicorp/aws');
   });
+
+  describe('packageRules array handling', () => {
+    it('returns the input packageRules array without re-cloning it', async () => {
+      const config: TestConfig = {
+        packageName: 'a',
+        packageRules: [
+          {
+            matchPackageNames: ['a'],
+            // @ts-expect-error -- testing
+            x: 2,
+          },
+        ],
+      };
+      const res = await applyPackageRules(config);
+      expect(res.x).toBe(2);
+      expect(res.packageRules).toBe(config.packageRules);
+    });
+
+    it('does not add a packageRules key when the input has none', async () => {
+      const config: PackageRuleInputConfig = { packageName: 'a' };
+      const res = await applyPackageRules(config);
+      expect(res.packageRules).toBeUndefined();
+    });
+
+    it('keeps an empty packageRules array', async () => {
+      const config: PackageRuleInputConfig = {
+        packageName: 'a',
+        packageRules: [],
+      };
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by eyalzek** [comment]
Force-pushed once to remove a `Closes #44459` line from the commit message — the `closes-issue-or-discussion` required check rejects closing keywords in commits (they belong in the PR description, which still carries the reference). The diff is byte-identical; only the commit message changed. Sorry for the force-push against the template's guidance — was the only way to clear the required check.

File: lib/util/package-rules/index.ts [hunk 1]
```diff
@@ -107,6 +113,14 @@ export async function applyPackageRules<T extends PackageRuleInputConfig>(
       config = mergeChildConfig(config, toApply);
     }
   }
+  // Restore the rules. If any applied rule carried nested `packageRules`
+  // (e.g. from a resolved preset), preserve the concat-merge semantics that
+  // mergeChildConfig() would previously have applied.
+  if (config.packageRules) {
+    config.packageRules = packageRules.concat(config.packageRules);
```
viceice, line 120:
```suggestion
  if (config.packageRules.length && packageRules.length) {
    packageRules.push(...config.packageRules);
    config.packageRules = packageRules;
```
this seems to be better? so we don't create another array?

eyalzek: Good catch on avoiding the extra allocation, but I want to flag one thing before switching to `push`: `packageRules` here is the same array reference as `inputConfig.packageRules` (since `config = {...inputConfig}` is only a shallow copy). Mutating it with `push` mutates the caller's array in place, not just a local copy.

I traced every current call site (`flatten.ts`, `fetch.ts`, `filter-checks.ts`) and in each one the config passed in is freshly deep-cloned via `mergeChildConfig`/`klona` immediately before calling `applyPackageRules`, so mutating in place is safe *today*. But that's an implicit invariant — if a future caller ever passes a config without cloning first, `push` would silently corrupt that caller's shared `packageRules` array across separate dependencies/calls, which would be a nasty bug to track down.

Are you comfortable relying on that "callers always clone first" invariant holding indefinitely, or would you rather I keep the `concat()` (new array, no shared-mutation risk) and only reach for `push` if we want to formalize/document that callers must always pass an already-cloned config into this function?

eyalzek: @viceice any input on how you'd like to proceed? I tested this patch in our massive monorepos that would never complete a run before the GH token expires (1 hour) and with the fix it takes under 30 minutes.

eyalzek: @viceice ping

viceice (approved):
let's not asume the fresh cloned package rules yet. we can improve later by moving the clone to relevant places and add function docs about what happens / needs to be done (eg packages rules array  can / will be mutated)

PR #44827, opened 2026-07-24
fix(bun): use --lockfile-only when updating artifacts

paul-godhouse:
## Changes

During Bun lockfile updates (especially `lockFileMaintenance`, which deletes the lockfile then regenerates it), Renovate currently runs a full `bun install`. That materializes `node_modules` even though Renovate only reads the updated lockfile afterward.

On large Bun workspaces (e.g. Expo / React Native monorepos), a full install can exceed Mend Community runner memory (3GB) and fail with kernel OOM. Measured cold `bun install --lockfile-only --ignore-scripts` for a Bun workspace used ~460MB RSS, while a full install of an Expo monorepo OOMs the same runners.

This change always passes `--lockfile-only`, matching how Renovate already treats pnpm (`--lockfile-only`) and npm (`--package-lock-only`).

- `lib/modules/manager/bun/artifacts.ts`: `bun install` → `bun install --lockfile-only` (optional `--ignore-scripts` unchanged)
- Update unit expectations in `artifacts.spec.ts`

## Context

## AI assistance disclosure

Used Cursor (model: Cursor Grok 4.5) to implement the command change, update unit expectations, draft the PR, and measure lockfile-only memory usage.

The public repository: N/A

### DIFF
--- a/lib/modules/manager/bun/artifacts.spec.ts
+++ b/lib/modules/manager/bun/artifacts.spec.ts
@@ -106,15 +106,18 @@ describe('modules/manager/bun/artifacts', () => {
           },
         ]);
 
-        expect(exec).toHaveBeenCalledWith('bun install --ignore-scripts', {
-          cwdFile: 'bun.lockb',
-          docker: {},
-          toolConstraints: [
-            {
-              toolName: 'bun',
-            },
-          ],
-        });
+        expect(exec).toHaveBeenCalledWith(
+          'bun install --lockfile-only --ignore-scripts',
+          {
+            cwdFile: 'bun.lockb',
+            docker: {},
+            toolConstraints: [
+              {
+                toolName: 'bun',
+              },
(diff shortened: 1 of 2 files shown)


viceice (commented):
does this cause errors on older bun versions?

paul-godhouse:
@viceice 
No. --lockfile-only was added in Bun 1.1.43, but older versions don't reject unknown bun install flags — they ignore them and do a full install.

I checked this against real binaries:

1.0.30 / 1.1.0 / 1.1.42: bun install --lockfile-only exits 0 and still writes node_modules (same as today's command)
1.1.43+: the flag is honored, lockfile is updated, node_modules is not created
So this isn't a hard error on old Bun. Self-hosted binarySource=global users on <1.1.43 keep the previous full-install behavior; everyone else gets the lockfile-only path.

PR #44834, opened 2026-07-24
feat(versioning): support ranges for the regex scheme

MayCXC:
## Changes

The `regex` versioning scheme orders versions by the release array assembled from its captured groups, but it matches every range by exact equality (inherited from `GenericVersioningApi`). So a comparator range in `allowedVersions` (for example `<=1.2.3`, a documented constraint) never resolves against a `regex`-versioned dependency: `isValid` rejects the range, and filter.ts falls back to `semver.satisfies`, which cannot compare versions the scheme parses but semver does not.

This adds comparator range support to the `regex` scheme:

- `lib/modules/versioning/generic-range.ts`: a small, scheme-independent comparator parser. It strips the operator (`<`, `<=`, `>`, `>=`, `=`, `==`; whitespace/comma-separated comparators are ANDed) and defers operand validation and comparison to the caller, so it never coerces or normalizes operands the way npm `semver` does.
- The `regex` scheme overrides `isValid`/`matches`/`getSatisfyingVersion`/`minSatisfyingVersion` (and `isVersion`/`isSingleVersion`, to keep a range distinct from a version) to use the helper, evaluating each comparator through the scheme's own `_compare`.

Because the comparison uses the scheme's own ordering, a numeric suffix captured as a `build` group orders correctly, where semver cannot: as a prerelease it breaks the range, as build metadata it is ignored, and a four-part release is invalid semver. The same comparator ranges also work in `matchCurrentVersion`, which goes through the same `versioningApi.matches`.

Dependency values remain exact tags, so `supportsRanges` stays `false` (the scheme does not update range dependencies). `generic.ts` and the other generic-based schemes are untouched. The `regex` readme documents the capability with an example.

## Context

## AI assistance disclosure

AI was used substantively for the code, tests, and documentation. Multiple models were used.

The public repository: N/A (verified with unit tests; coverage is 100% on both new and changed files, and the full versioning suite passes)

### DIFF
--- a/lib/modules/versioning/regex/readme.md
+++ b/lib/modules/versioning/regex/readme.md
@@ -83,3 +83,22 @@ Here is another example, this time for handling `ghcr.io/linuxserver/openssh-ser
   ]
 }
 ```
+
+The `regex` scheme also accepts a comparator range wherever a version is matched against a constraint, such as `allowedVersions` and `matchCurrentVersion`, evaluated through the same capture-group ordering.
+The supported operators are `<`, `<=`, `>`, `>=`, `=` and `==`, and several comparators separated by whitespace or a comma are combined with logical AND.
+Because the comparison uses the scheme's own ordering rather than npm `semver`, a numeric suffix captured as `build` is ordered correctly, where `semver` would treat it as a prerelease and reject the range.
+
+For example, to hold a GitLab EE image at or below a required upgrade stop, capture the `-ee` suffix as a numeric `build` group and cap it with `allowedVersions`:
+
+```json
+{
+  "packageRules": [
+    {
+      "matchDatasources": ["docker"],
+      "matchPackageNames": ["gitlab/gitlab-ee"],
+      "versioning": "regex:^(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)-ee\\.(?<build>\\d+)$",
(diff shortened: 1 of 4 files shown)


viceice (dismissed):
interesting

lib/modules/versioning/regex/readme.md
```diff
@@ -83,3 +83,22 @@ Here is another example, this time for handling `ghcr.io/linuxserver/openssh-ser
   ]
 }
 ```
+
+The `regex` scheme also accepts a comparator range wherever a version is matched against a constraint, such as `allowedVersions` and `matchCurrentVersion`, evaluated through the same capture-group ordering.
```
secustor: I'm worried that will break use cases where the order is not strictly following a decreasing order e.g. calender based versioning `YYYY.DD.MM`

MayCXC: Ordering is by the `major`/`minor`/`patch` names in precedence, not capture-group position, so `YYYY.DD.MM` sorts correctly by naming the month `minor` and the day `patch` (year, then month, then day). Reworded the readme to say that, and added a test.

secustor (approved):

MayCXC:
thank you @viceice and @secustor 😎 more contributions are available at the automatically closed https://github.com/renovatebot/renovate/pull/45908 and the as of yet still open
https://github.com/renovatebot/renovate/pull/45909

PR #3582, opened 2019-04-23
fix(azure-devops): Sanitize azure devops project name

hillerstorm:
Closes #3581

PR #44854, opened 2026-07-26
fix(platform/azure): Encode all path segments

rm-code:
The changes introduced in #3582 and #3620 only encoded the project and repository names for the manual fallback url.

This uses the already existent parseUrl utility wrapped in an azure specific encodeUrlPathSegments util to make sure the full url is encoded correctly.

See discussion #44818

## Changes

Replaced the manual encoding of project and repository path segments with a function to encode all path segments (so we don't have to make any assumptions about the path structure).

## Context

## AI assistance disclosure

- [x] Yes — other (please describe): I wasn't familiar with the test framework and had AI help with understanding the structure, mocking and how to override the before-each setup in `index.spec.ts` where necessary.

The public repository: n/a


rm-code:
I added additional (regression) tests to verify this change doesn't break existing URLs: baseline tests confirming plain project/repo names (no special characters) still produce unchanged URLs via both the manualUrl fallback and repo.remoteUrl paths, plus a test for on-prem Azure DevOps Server collection paths.

While validating these, I found a quirk with `encodeUrlPathSegments()`. It threw URI malformed for a project/repo name containing a literal % (e.g. "50% off"), since decodeURIComponent() doesn't tolerate a bare % and the fallback path no longer pre-encodes the name before calling this function. 

This shouldn't be an issue for any real ADO paths, because -afaict- ADO doesn't allow bare % characters in its paths. It's just something the more lenient URL-handling inside of Node causes.

As a defensive measure we now sanitizing stray % characters (% not followed by two hex digits → %25) before decoding, to restore the previous encodeURIComponent()-equivalent behavior. 

FWIW we've been running Renovate with this fix against our internal ADO Server without any issues.

rm-code:
@viceice I think this PR was falsely flagged:

> Error: Merging this PR would make semantic-release publish a new major version (43.281.0 → 44.0.0), because a commit carries a breaking-change marker (`!` in the header, or a `BREAKING CHANGE:`/`BREAKING-CHANGE:` note). Remove the marker, or — if a major bump is intended (e.g. preparing the next major) — override this check manually.

Should I rebase?

lib/modules/platform/azure/util.ts
```diff
@@ -224,9 +228,18 @@ export function encodeUrlPathSegments(inputUrl: string): string {
...
+  // percent-encoded sequence (e.g. a project named "50% off"). The WHATWG
+  // URL parser accepts such a bare '%' in a path without error, but
+  // decodeURIComponent() throws "URI malformed" on it. Escape any such
+  // stray '%' first so decoding never throws, then decode and re-encode
+  // as usual.
   url.pathname = url.pathname
     .split('/')
-    .map((segment) => encodeURIComponent(decodeURIComponent(segment)))
+    .map((segment) => {
+      const sanitized = segment.replace(invalidPercentEncoding, '%25');
```
viceice: inline variable

viceice (commented):
merge main if you think it will fix your ci issue

lib/modules/platform/azure/util.ts
rm-code: Inlined the `sanitized` variable. I assume the regex is fine as a separate const outside of the map statement?

lib/modules/platform/azure/index.ts
```diff
@@ -244,10 +245,8 @@ export async function initRepo({
     url: defaults.endpoint,
   });
   // TODO: types (#22198)
-  const manualUrl = `${defaults.endpoint!}${encodeURIComponent(
-    projectName,
-  )}/_git/${encodeURIComponent(repoName)}`;
-  const url = repo.remoteUrl ?? manualUrl;
+  const manualUrl = `${defaults.endpoint!}${projectName}/_git/${repoName}`;
+  const url = encodeUrlPathSegments(repo.remoteUrl ?? manualUrl);
```
RahulGautamSingh: Keep `encodeURIComponent` on the manual segments and apply `encodeUrlPathSegments` only to `repo.remoteUrl`.

lib/modules/platform/azure/util.ts
```diff
@@ -213,3 +214,34 @@ export function getWorkItemTitle(rawTitle: string, repository: string): string {
   }
   return rawTitle;
 }
+
+// Matches a '%' that is NOT the start of a valid percent-encoded triplet
+// (i.e. not followed by two hex digits), e.g. the literal '%' in "50% off".
+const invalidPercentEncoding = /%(?![0-9a-fA-F]{2})/g;
```
RahulGautamSingh: RE2 has no lookahead, so `regEx()` would throw `CONFIG_VALIDATION` at module load. 

Add a comment saying so, otherwise someone will "fix" this later
and remove the regEx wrapper

lib/modules/platform/azure/util.ts
```diff
@@ -213,3 +214,34 @@ export function getWorkItemTitle(rawTitle: string, repository: string): string {
...
+
+export function encodeUrlPathSegments(inputUrl: string): string {
```
RahulGautamSingh: This is not Azure specific. Move it to util/url.ts next to parseUrl.

lib/modules/platform/azure/util.ts
```diff
@@ -213,3 +214,34 @@ export function getWorkItemTitle(rawTitle: string, repository: string): string {
...
+  // Parsing the URL is not enough. URL doesn't encode
+  // every reserved character.
+  const url: URL | null = parseUrl(inputUrl);
```
RahulGautamSingh: Drop the annotation, parseUrl already returns URL | null.

lib/modules/platform/azure/util.ts
```diff
@@ -213,3 +214,34 @@ export function getWorkItemTitle(rawTitle: string, repository: string): string {
...
+    return inputUrl;
+  }
+
+  // Decode each segment, re-encode it and join them together.
+  // A segment may contain a literal '%' that isn't part of a valid
+  // percent-encoded sequence (e.g. a project named "50% off"). The WHATWG
+  // URL parser accepts such a bare '%' in a path without error, but
+  // decodeURIComponent() throws "URI malformed" on it. Escape any such
+  // stray '%' first so decoding never throws, then decode and re-encode
+  // as usual.
```
RahulGautamSingh: Seven lines of comment for six lines of code, and the same paragraph is pasted again into both spec files. 

One line covers it: 
`escape stray % so decodeURIComponent cannot throw.`

lib/modules/platform/azure/util.spec.ts
```diff
@@ -268,6 +269,105 @@ describe('modules/platform/azure/util', () => {
...
+
+    it('encodes a space in a path segment (repository)', () => {
+      expect(
+        encodeUrlPathSegments(
+          'https://dev.azure.com/my-org/my-project/_git/my repo/',
+        ),
+      ).toBe('https://dev.azure.com/my-org/my-project/_git/my%20repo/');
+    });
+
+    it('encodes a spaces in multiple path segments', () => {
```
RahulGautamSingh:
```suggestion
    it('encodes spaces in multiple path segments', () => {
```

lib/modules/platform/azure/util.spec.ts
```diff
@@ -268,6 +269,105 @@ describe('modules/platform/azure/util', () => {
...
+      expect(
+        encodeUrlPathSegments(
+          'https://dev.azure.com/my org/my project/_git/my repo/',
+        ),
+      ).toBe('https://dev.azure.com/my%20org/my%20project/_git/my%20repo/');
+    });
+
+    it('leaves the origin untouched', () => {
```
RahulGautamSingh: The origin is not untouched, :443 is dropped, so the name is misleading. 

Also expect(x.startsWith(...)).toBe(true) hides the actual value on failure. Assert the full string with toBe.

lib/modules/platform/azure/util.spec.ts
```diff
@@ -268,6 +269,105 @@ describe('modules/platform/azure/util', () => {
...
+      const origin = 'https://dev.azure.com:443/';
+      const encoded = encodeUrlPathSegments(
+        `${origin}my org/my project/_git/my repo`,
+      );
+      expect(encoded.startsWith('https://dev.azure.com/')).toBe(true);
+    });
+
+    it('prevserves non-default ports', () => {
```
RahulGautamSingh:
```suggestion
    it('preserves non-default ports', () => {
```

and  same startsWith problem as above, assert the full string with toBe.

lib/modules/platform/azure/index.spec.ts
```diff
@@ -239,6 +239,275 @@ describe('modules/platform/azure/index', () => {
     });
   });
 
+  describe('initRepo URL encoding', () => {
```
RahulGautamSingh, line 242: These all test cases re-assert what the `encodeUrlPathSegments` unit tests already cover. 

One manualUrl case and one remoteUrl case are enough at this layer.

RahulGautamSingh (changes requested):

lib/modules/platform/azure/index.ts
rm-code: @RahulGautamSingh I'm not sure if I understand correctly. This would mean that the endpoint for `manualUrl` isn't encoded.
```ts
  const manualUrl = `${defaults.endpoint!}${encodeURIComponent(
    projectName,
  )}/_git/${encodeURIComponent(repoName)}`;
  const url = repo.remoteUrl
    ? encodeUrlPathSegments(repo.remoteUrl)
    : manualUrl;
```

Is that what you meant?

lib/modules/platform/azure/index.spec.ts
rm-code: Following up on my other comment, here we only need a manualUrl case, if we don't encode the remoteUrl at all.

lib/modules/platform/azure/util.ts
rm-code: I dropped the lookahead and used a regEx wrapper. lmwyt

romanett:
@RahulGautamSingh any update on this? I would need this fix to start using renovate

PR #44876, opened 2026-07-27
feat(github-actions): update mise checksums

zeitlinger:
## Changes

Teach the GitHub Actions manager to treat a literal `jdx/mise-action` `with.sha256` input as the digest paired with `with.version`.

The manager now uses the `github-release-attachments` datasource, so Renovate maps the checksum of the currently selected mise binary to the corresponding asset in the new release and updates both fields together. Non-literal checksums and configurations without `sha256` continue to update only the version.

This follows up on #44660. Its native `with.version` support overlaps checksum-aware custom managers: the native update can change the version while leaving the old checksum behind, as seen in https://github.com/grafana/oats/pull/444.

## Context

## AI assistance disclosure

- [x] Yes — substantive assistance (OpenAI Codex helped investigate the regression and implement the code and tests).

Commands:

- `pnpm vitest lib/modules/manager/github-actions/extract.spec.ts lib/modules/manager/github-actions/integration.spec.ts --run`
- `pnpm check --all lib/modules/manager/github-actions/community.ts lib/modules/manager/github-actions/extract.spec.ts lib/modules/manager/github-actions/integration.spec.ts`

### DIFF
--- a/lib/modules/manager/github-actions/community.ts
+++ b/lib/modules/manager/github-actions/community.ts
@@ -2,6 +2,7 @@ import { z } from 'zod/v4';
 
 import { escapeRegExp, regEx } from '../../../util/regex.ts';
 import { DockerDatasource } from '../../datasource/docker/index.ts';
+import { GithubReleaseAttachmentsDatasource } from '../../datasource/github-release-attachments/index.ts';
 import { GithubReleasesDatasource } from '../../datasource/github-releases/index.ts';
 import { NpmDatasource } from '../../datasource/npm/index.ts';
 import { PypiDatasource } from '../../datasource/pypi/index.ts';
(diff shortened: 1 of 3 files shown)


lib/modules/manager/github-actions/community.ts
```diff
@@ -92,6 +93,17 @@ const InstallBinaryWith = z
   .object({ repo: z.string(), tag: z.string() })
   .transform(({ repo, tag }) => ({ packageName: repo, val: tag }));
 
+const sha256Regex = regEx(/^[a-f0-9]{64}$/);
```
jamietanna, line 96: Will we see any capitalised SHAs?

viceice (dismissed):

jamietanna:
I've moved from `fix` to `feat` as I believe this is new support, not a regression/bug

lib/modules/manager/github-actions/community.ts
zeitlinger: No. mise-action generates a lowercase hex digest and compares it case-sensitively with the `sha256` input, so a capitalized checksum would fail verification.

jamietanna: Worth adding a test to document that it's not supported? Is it worth us supporting it so - at some point in the future, if it's changed - that we support that?

zeitlinger:
Makes sense—thanks for updating it.

jamietanna (approved):

lib/modules/manager/github-actions/community.ts
zeitlinger: Yes, documenting the current behavior makes sense. I added a test showing that uppercase checksums aren’t extracted because mise-action currently compares the supplied checksum case-sensitively.

PR #44878, opened 2026-07-27
fix(manager/azure-pipelines): support template expression nesting

scrocquesel-ml150:
## Changes

Recurse template conditional expression. Related to #38821

## Context

## AI assistance disclosure

The public repository: <URL>

### DIFF
--- a/lib/modules/manager/azure-pipelines/extract.spec.ts
+++ b/lib/modules/manager/azure-pipelines/extract.spec.ts
@@ -517,5 +517,148 @@ describe('modules/manager/azure-pipelines/extract', () => {
       });
       expect(res).toBeNull();
     });
+
+    it('should extract steps wrapped in a template expression conditional', () => {
+      const packageFile = [
+        'steps:',
+        "- ${{ if eq(variables.foo, 'bar') }}:",
+        '  - task: Bash@3',
+        '    inputs:',
+        "      script: 'echo Hello World'",
+      ].join('\n');
+      const res = extractPackageFile(packageFile, azurePipelinesFilename, {
+        repository: 'repo',
+      });
+      expect(res?.deps).toEqual([
+        {
+          depName: 'Bash',
+          currentValue: '3',
+          datasource: AzurePipelinesTasksDatasource.id,
+        },
+      ]);
+    });
+
+    it('should extract jobs wrapped in a template expression conditional', () => {
+      const packageFile = [
+        'jobs:',
+        "- ${{ if eq(parameters.environment, 'prod') }}:",
+        '  - job: build',
+        '    steps:',
+        '      - task: Bash@3',
+      ].join('\n');
+      const res = extractPackageFile(packageFile, azurePipelinesFilename, {
+        repository: 'repo',
+      });
+      expect(res?.deps).toEqual([
+        {
+          depName: 'Bash',
+          currentValue: '3',
+          datasource: AzurePipelinesTasksDatasource.id,
+        },
+      ]);
+    });
+
+    it('should extract stages wrapped in a template expression conditional', () => {
+      const packageFile = [
+        'stages:',
+        "- ${{ if eq(parameters.environment, 'prod') }}:",
+        '  - stage: deploy',
+        '    jobs:',
+        '      - job: build',
+        '        steps:',
+        '          - task: Bash@3',
+      ].join('\n');
+      const res = extractPackageFile(packageFile, azurePipelinesFilename, {
+        repository: 'repo',
+      });
+      expect(res?.deps).toEqual([
+        {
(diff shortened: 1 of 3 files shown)


lib/modules/manager/azure-pipelines/extract.ts
```diff
@@ -162,16 +168,54 @@ function extractDeploy(deploy: Deploy | undefined): PackageDependency[] {
...
   const deps: PackageDependency[] = [];
   for (const jobOrDeployment of coerceArray(jobs)) {
-    const deployment = jobOrDeployment as Deployment;
-    if (deployment.strategy) {
-      deps.push(...extractDeploy(deployment.strategy.canary));
-      deps.push(...extractDeploy(deployment.strategy.rolling));
-      deps.push(...extractDeploy(deployment.strategy.runOnce));
+    if (isTemplateExprNode(jobOrDeployment)) {
+      for (const nested of Object.values(jobOrDeployment)) {
+        deps.push(...extractJobs(nested as Jobs));
```
zharinov: I don't think Zod is being used properly: nested loops, `coerceArray`, `Object.values` and type coercion are symptoms. But I guess it's not the fault of this PR, so I wouldn't insist to refactor it.

lib/modules/manager/azure-pipelines/extract.spec.ts
```diff
@@ -517,5 +517,148 @@ describe('modules/manager/azure-pipelines/extract', () => {
...
     });
+
+    it('should extract steps wrapped in a template expression conditional', () => {
+      const packageFile = [
+        'steps:',
+        "- ${{ if eq(variables.foo, 'bar') }}:",
+        '  - task: Bash@3',
+        '    inputs:',
+        "      script: 'echo Hello World'",
+      ].join('\n');
```
zharinov: Please use the `codeBlock`

zharinov (changes requested):

lib/modules/manager/azure-pipelines/extract.ts
scrocquesel-ml150: I think you right, give me time to find something better. I never use zod and pretty newbie on typescript type system, so I need to look into it carefully.

zharinov: Look at how `.transform(...)` is used somewhere else. The key idea is Zod first matches the schema, then transforms the data using the data inferred from the schema. In the end of this pipeline, your data has the most convenient shape and minimal processing is needed at the business logic layer.

scrocquesel-ml150: I didn't use transform. The transformation is about an array item that can transform to istelf or many and I didn't know how to express that. Kept the type matching function to do if/else in the extract.

scrocquesel-ml150: Finally, got it working with transform and with minor modification of extract

lib/modules/manager/azure-pipelines/schema.ts
```diff
@@ -1,13 +1,57 @@
...
+    z.union([
+      z
+        .record(z.string(), Items)
+        .refine((node) => Object.keys(node).every(isTemplateExpression))
+        .transform((node) => Object.values(node).flat()),
+      Elem.transform((item: z.TypeOf<Schema>) => [item]),
+    ]),
+  );
+
+  const Items: z.ZodType<z.TypeOf<Schema>[], any> = LooseArray(Item).transform(
```
viceice, line 38:
```suggestion
  const Items = LooseArray(Item).transform(
```

lib/modules/manager/azure-pipelines/schema.ts
```diff
@@ -1,13 +1,57 @@
...
+ * @returns Schema for a flat array of elements
+ */
+function TemplateExpressionArray<Schema extends z.ZodTypeAny>(
+  Elem: Schema,
+): z.ZodType<z.TypeOf<Schema>[], any> {
+  // An element expands to zero or more items: a template expression node
+  // expands to the items of its nested lists, anything else to itself.
+  // The `refine` is required to not mistake an element whose every value is a
+  // list, such as `- jobs: [...]`, for a template expression node.
+  const Item: z.ZodType<z.TypeOf<Schema>[], any> = z.lazy(() =>
```
viceice, line 28:
```suggestion
  const Item = z.lazy(() =>
```

are the explicit types necessary?

viceice (commented):
otherwise LGTM

lib/modules/manager/azure-pipelines/schema.ts
scrocquesel-ml150: Yes, because Item and Items are recursively referencing themself

PR #44894, opened 2026-07-27
fix(config-validator): warn when `RE2` is not in use

silverwind:
## Changes

`renovate-config-validator` now warns when `RE2` is unavailable (or ignored via `RENOVATE_X_IGNORE_RE2`) and it falls back to the JS `RegExp` engine, which accepts syntax `RE2` rejects, such as lookahead. Exit code is unchanged.

Reported in https://github.com/renovatebot/renovate/discussions/44873.

## Context

## AI assistance disclosure

Code and test written by Claude (Opus 5) via Claude Code.


lib/config-validator.ts
```diff
@@ -174,6 +174,18 @@ If you have specified global self-hosted configuration (https://docs.renovatebot
...
+    if (regexEngineStatus.type === 'unavailable') {
+      logger.warn(
+        { err: regexEngineStatus.err },
+        'RE2 not usable, falling back to RegExp: regex validation may be inaccurate',
+      );
+    } else if (regexEngineStatus.type === 'ignored') {
+      logger.warn(
+        'RE2 ignored via RENOVATE_X_IGNORE_RE2: regex validation may be inaccurate',
+      );
+    }
```
viceice, line 187: I'm not sure if we want to warn here.

@secustor @jamietanna wdyt?

viceice (commented):
please sign the cla

jamietanna (dismissed):

lib/config-validator.ts
jamietanna: I think it's a reasonable thing to do - but maybe dropping it down to DEBUG is better. If you've set the env var you're _likely_ aware of what it'll do

silverwind: My uses case: I was running the validator via Claude and if this DEBUG requires non-standard usage, it would not have passed it.

I would like this visible with regular usage.

silverwind:
cla done

lib/config-validator.ts
secustor: I would switch the second logger to debug and keep the first on warn. 

In Claude sets the env variable to skip the test it would have also ignored warning log.

silverwind: Agree, someone setting `RENOVATE_X_IGNORE_RE2` knows what they are doing. Done in https://github.com/renovatebot/renovate/pull/44894/commits/291835e19191a28268a6dd5de64c306e241802ac.

Alternatively could also remove that debug log and test. As I see it it was primarily only added so that this codepath is testable when `re2` is available.

secustor (approved):

jamietanna (approved):

PR #44004, opened 2026-06-15
feat(managers/pep621): update pixi.lock if present

nielspardon:
## Changes

When a `pyproject.toml` uses both `[tool.pixi]` configuration and standard PEP 621 dependencies (e.g. `[project.dependencies]` or `[dependency-groups]`), the `pep621` manager now also updates `pixi.lock` whenever those standard dependencies change.

This is implemented as a new `PixiProcessor` in the `pep621` manager's processor pipeline, following the same pattern as `PdmProcessor` and `UvProcessor`. The processor:

- Detects pixi usage by checking for a sibling `pixi.lock` file
- Runs `pixi lock` after `pep621` dependencies are updated
- Handles lock file maintenance (deletes the lock file first to force full regeneration)
- Respects `requires-pixi` from `[tool.pixi.project]`, `[tool.pixi.workspace]` and `config.constraints.pixi` for tool version pinning

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: # 
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

Relates to discussion: https://github.com/renovatebot/renovate/discussions/43260

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

New unit tests added in lib/modules/manager/pep621/processors/pixi.spec.ts.

The public repository: https://github.com/nielspardon/substrait/pulls

PR #44939, opened 2026-07-29
feat(manager/pixi): gate pixi.lock updates behind `allowedUnsafeExecutions`

nielspardon:
## Changes

The `pixi` manager runs `pixi lock` to refresh `pixi.lock`. That command can execute arbitrary code from conda package hooks (see [pixi's security docs](https://pixi.prefix.dev/latest/security/#4-treat-package-hooks-as-code-execution)), so it should be treated as an unsafe execution.

This PR gates `pixi.lock` updates in the `pixi` manager behind the global [`allowedUnsafeExecutions`](https://docs.renovatebot.com/self-hosted-configuration/#allowedunsafeexecutions) option, mirroring the existing `mise` gate:

- Adds `pixi` to the allowed values of `allowedUnsafeExecutions`.
- The `pixi` manager now only runs `pixi lock` when `pixi` is included in `allowedUnsafeExecutions`. When it is not, the package file is still updated but `pixi.lock` is left unchanged (a `logger.once.warn` is emitted).
- Documents the trust model in the manager readme, the `allowedUnsafeExecutions` table, and the security & permissions page.

This is split out from #44004 as a standalone change, as requested by maintainers, so it can land independently of the pep621 `pixi.lock` work. Per maintainer guidance this is a regular `feat` rather than a breaking change.

## Context

## AI assistance disclosure

Claude Code (model: Claude Opus 4.8) was used, under my direction, to write the gate, unit test, and documentation.

Added `returns null when pixi is not in allowedUnsafeExecutions` to the pixi artifacts test suite; ran `oxlint`, `prettier`, `tsc --noEmit`, and the pixi manager Vitest suite locally (all green).


jamietanna (approved):

jamietanna:
Thank you! I appreciate this may be a breaking change for some users, but this is security hardening to ensure that we reduce risk of remote code execution - we're releasing it as a `feat` to make it a _little bit_ clearer, without requiring a full major version

PR #44941, opened 2026-07-29
refactor(manager/pixi): extract shared pixi.lock update helper

nielspardon:
## Changes

Pure refactor, no behavior change. Extracts the duplicated `pixi.lock` update logic — shared by the standalone `pixi` manager (`lib/modules/manager/pixi/artifacts.ts`) and the `pep621` pixi processor (`lib/modules/manager/pep621/processors/pixi.ts`) — into a single shared helper `updatePixiLockfile` in `lib/modules/manager/pixi/lockfile.ts`.

Both call sites now only derive the `pixi` version constraint from their own config source and delegate to the helper. The helper owns `commandLock`, the updated-deps/lockfile existence checks, the `allowedUnsafeExecutions` (`pixi`) gate, and the `pixi lock` exec flow.

Behavior is preserved for both managers:

- The `allowedUnsafeExecutions` gate stays before the write/exec, so both managers still no-op the lockfile refresh (package file still updated) when `pixi` isn't allowed.
- The package file is written before running `pixi lock` only when the caller passes `newPackageFileContent` (the `pixi` manager). The `pep621` manager writes the package file itself and leaves it unset, matching its prior behavior.

Follow-up to #44004, which landed the `allowedUnsafeExecutions` gate and `pep621` `pixi.lock` support. This extraction was originally part of that PR but was reverted on maintainer request to keep it focused; this re-lands it as a standalone refactor.

## Context

## AI assistance disclosure

Claude Code (model: Claude Opus 4.8) performed the extraction and delegation, then ran type-check, the existing `pixi` + `pep621` unit tests, and lint/format. The approach and every change were reviewed by me before opening.

The existing `pixi` and `pep621` pixi unit tests exercise the shared helper through both call sites and pass unchanged, giving 100% coverage of the three touched files. No new tests were needed since behavior is unchanged.

### DIFF
--- a/lib/modules/manager/pep621/processors/pixi.spec.ts
+++ b/lib/modules/manager/pep621/processors/pixi.spec.ts
@@ -6,7 +6,6 @@ import type {
   InternalGlobalConfigOptions,
   RepoGlobalConfig,
 } from '../../../../config/types.ts';
-import { TEMPORARY_ERROR } from '../../../../constants/error-messages.ts';
 import * as docker from '../../../../util/exec/docker/index.ts';
 import { getPkgReleases as _getPkgReleases } from '../../../datasource/index.ts';
 import type { UpdateArtifactsConfig } from '../../types.ts';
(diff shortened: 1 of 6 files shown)


jamietanna (dismissed):
Nice! I guess we're intentionally not duplicating the tests / moving them into `lockfile.spec.ts`? (From your comment, yes)

Or making sure that there's a spec file that maybe indicates we already have coverage in those two files?

jamietanna (approved):

PR #28275, opened 2024-04-08
fix(cache): Trigger cacache to remove expired contents

gg-kialo:
## Changes

Verify the cacache after removing index entries at the end of a run to remove cache contents on disk.

The existing cacache cleanup is only removing cache entries from the index, not the contents.
This results in the cache contents being forgotten about and continuing to accrue on disk.

Verifying with `cacache.verify(...)` will garbage collect any content entries not referenced by the index as well as cleaning up corrupted or invalid index entries.

## Context

Closes #28169.

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

PR #44987, opened 2026-07-31
fix(cache): re-add cacache.verify() to garbage collect orphaned content from put() overwrites

bbodenmiller:
## Changes

Re-add `cacache.verify()` to the `destroy()` method to garbage collect content blobs orphaned by `cacache.put()` overwrites.

When `cacache.put()` is called with an existing key but different data (e.g., a version lookup result changes because a new release was published), cacache writes a new content blob and updates the index entry to point to it, but does **not** remove the previous content blob. Over many scheduled runs with many dependencies, these orphaned blobs accumulate indefinitely in `content-v2/`.

`rm.content` (added in #29860 when `verify()` was removed) only removes content when an entry
      **expires**. It does not handle orphans created by overwrites of still-live keys.

On a self-hosted instance running scheduled scans of ~100 repos over several months, the cache grew to multiple GB of orphaned content while the actual referenced data was only ~16 MB. Running `cacache.verify()` reclaimed ~5 GB by removing ~565k unreferenced content blobs. After GC, Renovate's cache behavior was unchanged (verified: identical package cache hits, HTTP request counts, and lookup results before and after).

### Performance

The concern in #29795 was that `verify()` is slow. However:

- On a healthy cache (after initial cleanup), `verify()` completes in under 1 second (~260ms for ~500 entries in testing)
- The slow times previously reported were caused by the orphan accumulation itself — scanning hundreds of thousands of orphaned blobs is slow, but that only happens once
- Without `verify()`, the cache grows until it causes CI cache upload failures (exceeding S3 size limits), pod evictions from ephemeral storage exhaustion, and minutes-long cache restore times — all worse than a sub-second GC pass

## Context

- [ ] This closes an existing Issue, Closes: 

Related:
- #28275 — originally added `cacache.verify()` to fix #28169
- #29795 / #29860 — removed `verify()` for performance, replaced with `rm.content` per expired entry (does not cover put() orphans)
- #42543 — open PR with alternative GC approach (tracks digests during sweep, but does not handle pre-existing orphans)

## AI assistance disclosure

Investigation and drafting assisted by Claude. Root cause analysis, testing on real self-hosted instance, and code change were done collaboratively.

Tested `cacache.verify()` on a real self-hosted GitLab instance across three job types (local scan, 50-repo dry run, 100+ repo dry run). For each: ran cold (no cache), warm (with cache), GC, then warm again. Cache metrics (package cache gets/sets, HTTP requests) were identical between warm and warm-after-GC runs, confirming GC does not affect cache functionality.

### DIFF
--- a/lib/util/cache/package/impl/file.ts
+++ b/lib/util/cache/package/impl/file.ts
@@ -56,6 +56,15 @@ export class PackageCacheFile extends PackageCacheBase {
     if (errorCount > 0) {
       logger.debug(`Error count cleaning up cache: ${errorCount}`);
     }
+
+     // Garbage collect content blobs orphaned by put() overwrites.
+     // rm.content only handles expiry; when put() writes new data for an existing
+     // key, the previous content blob is left on disk unreferenced.
+     const gcStats = await cacache.verify(this.cacheFileName);
+     logger.debug(
+       `Cache GC: kept ${gcStats.verifiedContent} content entries (${gcStats.keptSize} bytes), removed ${gcStats.reclaimedCount} orphaned (${gcStats.reclaimedSize} bytes) in ${gcStats.runTime.total}ms`,
+     );
+    
     const durationMs = Date.now() - startTime;
     logger.debug(
       `Deleted ${deletedCount} of ${totalCount} file cached entries in ${durationMs}ms`,
--- a/lib/util/cache/package/impl/file.ts
+++ b/lib/util/cache/package/impl/file.ts
@@ -57,13 +57,13 @@ export class PackageCacheFile extends PackageCacheBase {
       logger.debug(`Error count cleaning up cache: ${errorCount}`);
     }
 
-     // Garbage collect content blobs orphaned by put() overwrites.
-     // rm.content only handles expiry; when put() writes new data for an existing
-     // key, the previous content blob is left on disk unreferenced.
-     const gcStats = await cacache.verify(this.cacheFileName);
-     logger.debug(
-       `Cache GC: kept ${gcStats.verifiedContent} content entries (${gcStats.keptSize} bytes), removed ${gcStats.reclaimedCount} orphaned (${gcStats.reclaimedSize} bytes) in ${gcStats.runTime.total}ms`,
-     );
+    // Garbage collect content blobs orphaned by put() overwrites.
+    // rm.content only handles expiry; when put() writes new data for an existing
+    // key, the previous content blob is left on disk unreferenced.
+    const gcStats = await cacache.verify(this.cacheFileName);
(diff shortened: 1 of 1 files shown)


zharinov (approved):
I approve but needs prettier fix

RahulGautamSingh (approved):
Nice

lib/util/cache/package/impl/file.ts
```diff
@@ -56,6 +56,15 @@ export class PackageCacheFile extends PackageCacheBase {
...
     }
+
+    // Garbage collect content blobs orphaned by put() overwrites.
+    // rm.content only handles expiry; when put() writes new data for an existing
+    // key, the previous content blob is left on disk unreferenced.
+    const gcStats = await cacache.verify(this.cacheFileName);
+    logger.debug(
+      `Cache GC: kept ${gcStats.verifiedContent} content entries (${gcStats.keptSize} bytes), removed ${gcStats.reclaimedCount} orphaned (${gcStats.reclaimedSize} bytes) in ${gcStats.runTime.total}ms`,
+    );
+
```
bbodenmiller:
```suggestion

```

lib/util/cache/package/impl/file.ts
```diff
@@ -56,6 +56,14 @@ export class PackageCacheFile extends PackageCacheBase {
...
       logger.debug(`Error count cleaning up cache: ${errorCount}`);
     }
+
+    // Garbage collect content blobs orphaned by put() overwrites.
+    // rm.content only handles expiry; when put() writes new data for an existing
+    // key, the previous content blob is left on disk unreferenced.
+    const gcStats = await cacache.verify(this.cacheFileName);
+    logger.debug(
+      `Cache GC: kept ${gcStats.verifiedContent} content entries (${gcStats.keptSize} bytes), removed ${gcStats.reclaimedCount} orphaned (${gcStats.reclaimedSize} bytes) in ${gcStats.runTime.total}ms`,
+    );
```
bbodenmiller, line 66:
```suggestion
    );

```

bbodenmiller:
> I approve but needs prettier fix

@zharinov thanks, fixed

zharinov (approved):

viceice (approved):
we'll revert if it's causing too much trouble / time

bbodenmiller:
> we'll revert if it's causing too much trouble / time

@viceice if that becomes case please let me know so we can find other solutions to keep cache under control

viceice:
> > we'll revert if it's causing too much trouble / time
> 
> @viceice if that becomes case please let me know so we can find other solutions to keep cache under control

sure, we're doing that almost always 🤗

PR #45000, opened 2026-07-31
fix(nix): correctly qualify flake input refs as tags or branches

mglazer:
## Changes

When the nix manager rewrites a flake input's `ref=` query param (e.g. bumping `ref=some-branch` to a newer value), it substring-replaces the old ref value with the new one while keeping whatever `refs/heads/`/`refs/tags/` qualifier (or lack of one) the old ref happened to have.

This breaks when the new value resolves to a different ref namespace than the old one — for example, bumping an unqualified branch ref to a value that only exists as a tag upstream. Nix's git fetcher resolves a bare or mis-qualified ref via `refs/heads/<ref>` only, with no fallback to `refs/tags/`, so `nix flake lock`/`nix flake update` then fails with:

```
fatal: couldn't find remote ref refs/heads/<ref>
```

even though the ref exists as a tag (or vice versa).

This PR:
- Adds `gitRefType` (`'tags' | 'heads'`) to `Release`, populated by the git-refs datasource from the actual remote refs it already fetches (preferring `'tags'` when a value is ambiguous, matching Nix's own resolution order).
- Threads `gitRefType` through `generateUpdate` → `LookupUpdate` → `Upgrade`, so managers can see how a candidate new version actually resolves upstream.
- Reintroduces a nix-specific `updateDependency` (it didn't previously exist; nix relied on the generic auto-replace fallback). Without it, auto-replace can't reliably locate and rewrite a flake input URL's `ref=`/`rev=` query params together and throws for the common case of an input pinning both a ref and a rev.
- The new `updateDependency` uses `gitRefType` to decide the correct qualifier when rewriting a flake input URL, falling back to the old ref's own shape (qualified or bare) when that information isn't available, so behavior for datasources/managers that don't set `gitRefType` is unchanged.

## Context / repro

Repro'd locally with real `nix` (Determinate Nix 2.33 / upstream 2.18+): a flake input pinned via `ref=v20.0.0` where `v20.0.0` exists only as a tag on the remote fails identically:

```
fatal: couldn't find remote ref refs/heads/v20.0.0
```

while `ref=refs/tags/v20.0.0` resolves correctly. Also confirmed that without a nix-specific `updateDependency`, the generic auto-replace path throws `WORKER_FILE_UPDATE_FAILED` for a flake input pinning both `ref=` and `rev=` when bumping to a new value, since it can't locate a single `replaceString`/`currentValue` covering both.

## Tests

- Added `lib/modules/manager/nix/update.spec.ts` (12 cases): bare ref → tag, bare ref → branch, stale-qualifier correction, tag→tag preservation, no-`gitRefType` fallback, `github:` shorthand handling, digest-only updates, and not-found/unparseable-URL edge cases. 100% statement/line coverage on the new file.
- Updated `lib/modules/datasource/git-refs/index.spec.ts` snapshot for the new `gitRefType` field.
- Ran the full existing suites for `lib/modules/manager/nix/`, `lib/modules/datasource/git-refs/`, `lib/workers/repository/process/lookup/`, and `lib/workers/repository/update/branch/` — all passing, no regressions.

### DIFF
--- a/lib/modules/datasource/git-refs/__snapshots__/index.spec.ts.snap
+++ b/lib/modules/datasource/git-refs/__snapshots__/index.spec.ts.snap
@@ -5,31 +5,37 @@ exports[`modules/datasource/git-refs/index > getReleases > returns versions filt
   "releases": [
     {
       "gitRef": "v1.0.0",
+      "gitRefType": "tags",
       "newDigest": "7b756026fb2de270240a889a413e7e3a9d4d4d85",
       "version": "v1.0.0",
     },
     {
       "gitRef": "v1.0.1",
+      "gitRefType": "tags",
       "newDigest": "e173183f932ba8a31d0e4f23cc1070e8ebfa59d6",
       "version": "v1.0.1",
     },
     {
       "gitRef": "v1.0.2",
+      "gitRefType": "tags",
       "newDigest": "3936a6bced3587dc9fd464b0a910e0dfd4cfe10d",
       "version": "v1.0.2",
     },
     {
       "gitRef": "v1.0.3",
+      "gitRefType": "tags",
       "newDigest": "125ca9f3df4151e50046e5327ecb29ec4c13efab",
       "version": "v1.0.3",
     },
     {
       "gitRef": "v1.0.4",
+      "gitRefType": "tags",
       "newDigest": "3ed9e7d7094fd4ee7751c24a3e6b706060f461ff",
       "version": "v1.0.4",
     },
     {
       "gitRef": "v1.0.5",
+      "gitRefType": "tags",
       "newDigest": "6d7a933c2e6b7b39e992b1f93b6b42de083b28f0",
       "version": "v1.0.5",
     },
--- a/lib/modules/datasource/git-refs/index.ts
+++ b/lib/modules/datasource/git-refs/index.ts
@@ -43,6 +43,17 @@ export class GitRefsDatasource extends GitDatasource {
 
     const uniqueRefs = [...new Set(refs)];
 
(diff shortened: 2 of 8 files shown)


jbw976:
I think we may be hitting this same gap on https://github.com/crossplane/crossplane, and some investigation found this PR, so sharing our scenario here in case it's useful.

disclaimer: i'm still new to nix, so this is absolutely claude assisted 😇 

Our nixpkgs input is a plain branch ref with no rev, `nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.11"`, e.g. https://github.com/crossplane/crossplane/blob/1765f0e3914d934cda3f6cea12d7f86f441e6068/flake.nix#L8

Renovate wants to move it to `nixos-26.05` but it keeps encountering an error every run, without ever reaching the nix manager's `updateArtifacts`:

```
DEBUG: Value is not updated
       depName: nixpkgs
       expectedValue: nixos-26.05
       foundValue: nixos-25.11
WARN: Error updating branch: update failure
```

I think what's happening for us is a little different from the `ref=`/`rev=` case in your description. Auto-replace does rewrite our input correctly, but then it confirms itself by re-extracting, and the nix manager's `extractPackageFile` reads `flake.lock` rather than the `flake.nix` content it was handed. Since `flake.lock` still has the old `ref` at that point, the check always fails and the rewrite gets reverted.

Your `updateDependency` looks like it could possibly cover this scenario too - if you think that's reasonable, then it might be worth adding a test case for the bare `github:owner/repo/<branch>` shape - what do you think? 🤔 

Thanks! 🙇‍♂️

Issue #26937, opened 2024-01-30
Detect JsDelivr links in html manager


rarkins:
### Describe the proposed change(s).

Add support for JSDelivr links in Renovate's [html manager](https://github.com/renovatebot/renovate/tree/main/lib/modules/manager/html)

Blocked by https://github.com/renovatebot/renovate/issues/26936

waldyrious:
This would also make sense in the [cdnurl manager](https://docs.renovatebot.com/modules/manager/cdnurl/).

PR #45015, opened 2026-08-02
feat(datasource/jsdelivr): add new jsdelivr datasource

### LINKED ISSUE #26936: jsDelivr datasource

Add new datasource for jsDelivr (npm) modeled on CDNJS

Ref: https://www.jsdelivr.com/docs/data.jsdelivr.com#tag--Package-metadata--npm

rm-code:
## Changes

This PR adds a new datasource for the [jsdelvir](https://www.jsdelivr.com/) cdn as discussed in #26936, which is a blocker for #26937 and also could be used for a later libman manager implementation (see #41893).

The datasource queries the jsdelivr api: https://data.jsdelivr.com/v1

For fetching releases it queries three different endpoints for:
- [scoped npm](https://www.jsdelivr.com/docs/data.jsdelivr.com#get-/v1/packages/npm/@-scope-/-package-) packages
- [unscoped npm](https://www.jsdelivr.com/docs/data.jsdelivr.com#get-/v1/packages/npm/-package-) packages
- [gh](https://www.jsdelivr.com/docs/data.jsdelivr.com#get-/v1/packages/gh/-user-/-repo-) packages

To distinguish between `gh` and `npm` packages, it expects the manager to prefix the package name with a the respective `type`. As far as I understood, this should be the responsibility of the caller (extracting the URL from whatever source it's dealing with). Strictly speaking this goes against the outline given in the `datasource/README.md`:

> - `packageName`: the package's full name including scope if present (e.g. `@foo/bar`)

If this is a violation against the coding guidelines, I would need some guidance to how we'd deal with the different endpoint types instead and I'll happily update the PR. 

For digest it fetches a flat file structure via the respective API endpoints and searches for the asset in the list of files. I haven't found a better way of dealing with this, because the API doesn't offer any alternative afaict.

## Context

- [x] This closes an existing Issue, Closes: #26936

## AI assistance disclosure

- [ ] No documentation update is required (unless I missed it)

I added a `readme.md` documenting the type-prefix requirement for this datasource.

The public repository: n/a


lib/modules/datasource/jsdelivr/__fixtures__/gh_bootstrap.json
viceice, line 1: do we really need so many new fixtures? can't we inline them are reduce as much as possible?

looks like they can easily be generated from test code

lib/modules/datasource/jsdelivr/index.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  private async _getReleases(
+    config: GetReleasesConfig,
+  ): Promise<ReleaseResult | null> {
+    const result = Result.parse(config, ReleasesConfig)
+      .transform(({ packageName, registryUrl }) => {
+        const { type, package: parsedPackageName } =
+          parseJsDelivrPackageName(packageName);
+        const url = `${ensureTrailingSlash(registryUrl)}packages/${type}/${parsedPackageName}`;
+        return this.http.getJsonSafe(
```
viceice, line 56:
```suggestion
        return this.http.getJson(
```
simplify. we should drop the obsolete safe variants 😅

lib/modules/datasource/jsdelivr/index.ts
```diff
@@ -0,0 +1,163 @@
...
+    const {
+      type,
+      package: parsedPackageName,
+      asset,
+    } = parseJsDelivrPackageName(packageName);
+
+    const result = Result.parse(config, DigestsConfig).transform(
+      ({ registryUrl }) => {
+        const url = `${ensureTrailingSlash(registryUrl)}packages/${type}/${parsedPackageName}@${newValue}?structure=flat`;
+        return this.http.getJsonSafe(url, JsDelivrDigestResponse);
```
viceice, line 121:
```suggestion
        return this.http.getJson(url, JsDelivrDigestResponse);
```

lib/modules/datasource/jsdelivr/index.ts
```diff
@@ -0,0 +1,163 @@
...
+      },
+    );
+
+    const { val, err } = await result.unwrap();
+    if (err instanceof ZodError) {
+      logger.debug({ err }, 'jsdelivr: validation error');
+      return null;
```
viceice, line 128: should we return `undefined`? so a temporary error isn't cached?

rm-code: Sounds good👍 I'll also update `_getReleases` tp return `undefined` to avoid caching there.

rm-code:
The change from getJsonSafe to getJson was a bit more involved than I originally expected. I had modeled this after the cdnjs datasource which also still uses the `Result` approach. I now moved it towards the try / catch approach that (iiuc) is preferrable (e.g. how it's done in the JSR datasource).

I also carried over the guard against an empty `registryUrl` at the top. This returns `undefined` as well to avoid caching. I hope that's correct.

PR #45041, opened 2026-08-03
fix(workers/repository): don't bump versions when `rangeStrategy=pin` and `pinDigests` on pinned digests

jamietanna:
## Changes

If a repository has `rangeStrategy=pin` as their top-level config, and
i.e.

```
- name: Setup Allure TestOps
  uses: allure-framework/setup-allurectl@6878686fce27aeb6f49009173b2947a400e282b5 # v1
```

Previously, when generating a `renovate/pin-dependencies` branch, this
could also include the dependency bumps for a given `rangeStrategy=pin`
dependencies, which would result in the "pin dependencies" branch
including updates that can make it a little less safe.

Instead, we shouldn't attempt to perform `rangeStrategy=pin` when there
is already a digest pinned.

## Context

## AI assistance disclosure

Who answers review comments:

- [x] @jamietanna will read and reply directly. 

The public repository: <URL>

i.e. before this change: https://github.com/JamieTanna-Mend-testing/pindigest-no-bump/pull/4

After this change, no longer raised

### DIFF
--- a/lib/workers/repository/process/lookup/index.spec.ts
+++ b/lib/workers/repository/process/lookup/index.spec.ts
@@ -4,6 +4,7 @@ import * as httpMock from '~test/http-mock.ts';
 import { fakeSha, logger, partial } from '~test/util.ts';
 import * as hostRules from '../../../../../lib/util/host-rules.ts';
 import { getConfig } from '../../../../config/defaults.ts';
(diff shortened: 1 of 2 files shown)


secustor (commented):
I wonder if we should create separate PRs for pinning digests vs pinning versions. 

To allow a both at the same time 🤔

viceice (approved):

PR #40699, opened 2026-01-26
feat(sbt): support anonymous objects and dotted symbols for version declaration

Thooms:
## Changes

Support anonymous object version definition in the sbt plugin.

## Context

The following is a commonly used pattern:

```scala
val versions = new {
  example = "0.0.8"
}
libraryDependencies += "org.example" % "foo" % versions.example
```

This PR brings support to this, only on one level of nesting.

Please select one of the following:

- [ ] This closes an existing Issue, Closes: # 
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [x] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

The public repository: https://github.com/Thooms/renovate-example-scala-repository

Example of a created PR: https://github.com/Thooms/renovate-example-scala-repository/pull/4

PR #45082, opened 2026-08-05
feat(sbt): support singleton object with versions and dotted symbols

simonjscott:
## Context

PR #40699 (merged v43.113.0) added support for anonymous object assignment and dotted symbol resolution:

```scala
val versions = new {
  http4s = "0.23.35"
}
// resolved as versions.http4s
```

However, this anonymous object syntax (`new { ... }`) is not valid in sbt v2 build files — the Scala compiler rejects member access on structural types without `import scala.language.reflectiveCalls`, and sbt's build file compiler does not enable that feature. The pattern therefore cannot be used in practice in `project/*.scala` files targeting sbt v2.

A common — and compiler-safe — sbt pattern uses a named singleton object instead:

```scala
object Dependencies {
  object Version {
    val http4s    = "0.23.35"
    val circe     = "0.14.16"
    val log4cats  = "2.8.0"
  }

  val http4s = Seq(
    "org.http4s" %% "http4s-ember-server" % Version.http4s,
    ...
  )
}
```

This pattern is semantically equivalent but currently not extracted. The dotted reference resolution introduced in #40699 would handle `Version.http4s` correctly if the variable were registered — the only missing piece is extracting `object Name { val field = "..." }` into the shared variables map as `Name.field`.

**Suggested approach**

Run a regex pre-pass over the file before the main parser query, matching leaf-level named objects (bodies without nested braces):

```
/object\s+(\w+)\s*\{([^{}]*)\}/g
```

For each match, extract inner `val <field> = "<version>"` pairs and seed the parser context's `vars` map with `<ObjectName>.<field>` entries before the query runs. The existing dotted symbol resolver in `versionMatch` and `scalaVersionMatch` then resolves `Version.http4s` references without any further changes.

The regex naturally skips outer wrapper objects such as `object Dependencies { ... }` (whose bodies contain nested braces) and only captures the leaf version objects — so it does not interfere with dep extraction in the main query pass.

Two test cases cover the feature: a top-level `object Version { ... }` and one nested inside `object Dependencies { ... }`.

## AI assistance disclosure

Who answers review comments:

- [X] @simonjscott will read and reply directly.
- [X] An agent will draft replies and @simonjscott will read them before they are posted.

The public repository: <URL>

### DIFF
--- a/lib/modules/manager/sbt/extract.ts
+++ b/lib/modules/manager/sbt/extract.ts
@@ -393,6 +393,33 @@ export function extractProxyUrls(
   return extractedProxyUrls;
 }
 
+// Finds named singleton objects of the form:
+//   object Name {
+//     val field = "version"
+//   }
+// and returns a map of `Name.field -> version` entries.
+// Uses a regex that only matches object bodies without nested braces, so
+// outer wrapper objects (e.g. `object Dependencies { ... }`) are naturally
+// skipped — only leaf-level version objects are captured.
+const namedSingletonObjectRegex = regEx(/object\s+(\w+)\s*\{([^{}]*)\}/g);
+const namedSingletonObjectFieldRegex = regEx(
+  /\bval\s+(\w+)(?:\s*:\s*String)?\s*=\s*"([^"]+)"/g,
+);
+
+function extractNamedSingletonObjectVars(content: string): Vars {
+  const vars: Vars = {};
+  namedSingletonObjectRegex.lastIndex = 0;
+  for (const objectMatch of content.matchAll(namedSingletonObjectRegex)) {
+    const objectName = objectMatch[1];
+    const body = objectMatch[2];
+    namedSingletonObjectFieldRegex.lastIndex = 0;
+    for (const fieldMatch of body.matchAll(namedSingletonObjectFieldRegex)) {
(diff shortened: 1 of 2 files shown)


lib/modules/manager/sbt/extract.ts
```diff
@@ -393,6 +393,33 @@ export function extractProxyUrls(
...
+//     val field = "version"
+//   }
+// and returns a map of `Name.field -> version` entries.
+// Uses a regex that only matches object bodies without nested braces, so
+// outer wrapper objects (e.g. `object Dependencies { ... }`) are naturally
+// skipped — only leaf-level version objects are captured.
+const namedSingletonObjectRegex = regEx(/object\s+(\w+)\s*\{([^{}]*)\}/g);
+const namedSingletonObjectFieldRegex = regEx(
+  /\bval\s+(\w+)(?:\s*:\s*String)?\s*=\s*"([^"]+)"/g,
+);
```
viceice: do not store regex with global flag

simonjscott: apologies

simonjscott: I'm sorry my last comment wasn't very useful. Thank you for the feedback, I have changed the scope of the regex values and added a commit.

PR #45085, opened 2026-08-05
fix(manager/npm): massage out lockstep siblings pinning the replaced version

jeffreyzant:
## Changes

The npm post-update lockfile "massage" deletes `node_modules/<depName>` from `package-lock.json` before running `npm install`. When the updated dependency is the entry point of a lockstep-versioned monorepo (vue, react, @angular/*), siblings that exact-pin it (e.g. `@vue/server-renderer` with `peerDependencies: { "vue": "3.5.39" }`) stay in the lock at the old version. npm places the fresh dependency into the hole (`placeDep ... OK`, no `REPLACE`), keeps the stale sibling because it still satisfies its own demanding edges, and hard-fails with `ERESOLVE` on the sibling's exact peer pin — a deterministic artifact error.

This PR extends the massage: lock entries whose `dependencies`/`peerDependencies` pin the updated dep to **exactly the version being replaced** (`upgrade.lockedVersion ?? upgrade.currentVersion`) are massaged out as well, so npm re-resolves the whole family in lockstep. Per the discussion feedback, the version guard keeps this precise — range pins (`@vue/test-utils`'s `vue: "3.x"`) and unrelated exact pins are untouched.

Also adds the missing `peerDependencies` field to `LockFilePackage`.

## Context

Discussion #44664 (maintainer-approved direction: pin-scan variant with version guard).

Repro without Renovate (mirrors what the massage produces):

```bash
# deps: { "vue": "3.5.39" } + devDeps: { "@vue/test-utils": "^2.4.3" }
npm install --package-lock-only
# bump range to ^3.5.40 in package.json, delete node_modules/vue from the lock
npm install --package-lock-only    # ERESOLVE: peer vue@"3.5.39" from @vue/server-renderer@3.5.39
```

With this fix the stranded sibling is removed alongside the dep, and the same install resolves cleanly with the whole `@vue/*` family cascaded.

## Documentation (please check one)

- [ ] No new tests but ran on a real repository, or

New test: `massages out lockstep siblings pinning the old version` in `lib/modules/manager/npm/post-update/index.spec.ts` with fixture `update-lockfile-massage-2` (vue-style lock: exact-pinned sibling removed, range-pinned `@vue/test-utils` and unpinned `@vue/shared` kept). Also validated the resolution behavior end-to-end with npm 11.16.0 against a real repository lock exhibiting the failure.


lib/modules/manager/npm/post-update/index.ts
```diff
@@ -197,15 +197,35 @@ export async function writeExistingFiles(
             widens.push(upgrade.depName!);
           }
           const { depName } = upgrade;
+          const oldVersion = upgrade.lockedVersion ?? upgrade.currentVersion;
           for (const packageName of packageNames) {
+            if (!('packages' in npmLockParsed)) {
+              continue;
```
RahulGautamSingh: this continue is unreachable, so `codecov/patch` fails on it. 

`packageNames` is only filled when packages exists, so the loop never runs otherwise. wrap the deletion loop in a single `if ('packages' in npmLockParsed)` block to keep the type narrowed and drop this dead guard.

lib/modules/manager/npm/post-update/index.ts
```diff
@@ -197,15 +197,35 @@ export async function writeExistingFiles(
...
-              'packages' in npmLockParsed &&
-              (packageName === `node_modules/${depName}` ||
-                packageName.startsWith(`node_modules/${depName}/`))
+              packageName === `node_modules/${depName}` ||
+              packageName.startsWith(`node_modules/${depName}/`)
             ) {
               logger.trace({ packageName }, 'Massaging out package name');
               lockFileChanged = true;
               delete npmLockParsed.packages[packageName];
+            } else if (depName && oldVersion) {
```
RahulGautamSingh: flatten this. 

after the node_modules match does its delete, add continue and drop the else, so the sibling check is not nested a level deeper.

lib/modules/manager/npm/post-update/index.ts
```diff
@@ -197,15 +197,35 @@ export async function writeExistingFiles(
...
+              // Lockstep monorepo siblings (e.g. vue -> @vue/server-renderer)
+              // pin the updated dep to its exact old version. Left in the lock
+              // they cannot resolve against the new version and npm fails with
+              // ERESOLVE, so massage them out too and let npm re-resolve them.
+              const entry = npmLockParsed.packages[packageName];
+              const pin =
+                entry?.peerDependencies?.[depName] ??
```
RahulGautamSingh: `optionalDependencies` can exact-pin the updated dep too, and `LockFilePackage` already types it. add it to this lookup or say why it stays out.

lib/modules/manager/npm/post-update/index.spec.ts
```diff
@@ -203,6 +203,46 @@ describe('modules/manager/npm/post-update/index', () => {
       expect(git.getFile).toHaveBeenCalledExactlyOnceWith('package-lock.json');
     });
 
+    it('massages out lockstep siblings pinning the old version', async () => {
```
RahulGautamSingh, line 206: the fixture only hits the peerDependencies pin. 

add a sibling that pins through dependencies so the `?? entry?.dependencies?.[depName]` delete path is proven.

also is it possible to remove the fixture and instead use the necessary part directly in the test file

RahulGautamSingh (changes requested):

lib/modules/manager/npm/post-update/index.ts
jeffreyzant: Didn't know that branch existed, added it aswell!

RahulGautamSingh (approved):

viceice (approved):

jamietanna:
This may have caused a regression - we're investigating in https://github.com/renovatebot/renovate/discussions/45681

PR #45157, opened 2026-08-08
fix: keep HTTPS when resolving same origin URLs

### LINKED ISSUE #45134: "Ignoring cross-origin or invalid NuGet feed pagination link" should remap HTTPS -> HTTP downgrade
When i.e.

```json
{
  "name": "renovate",
  "hostname": "renovate-org-repo-df27891ewzdg8-2k5pw",
  "pid": 20,
  "level": 40,
  "logContext": "6cbb198f-f1f5-46f0-8032-2889c8f5aed5",
  "repository": "org/repo",
  "baseBranch": "main",
  "feedUrl": "https://community.chocolatey.org/api/v2",
  "nextUrl": "http://community.chocolatey.org/api/v2/FindPackagesById?id='chocolatey-core.extension'&$select=Version,IsLatestVersion,ProjectUrl,Published&$skiptoken='1757694056117','0.1.2','chocolatey-core.extension'",
  "msg": "Ignoring cross-origin or invalid NuGet feed pagination link",
  "time": "2026-08-07T10:16:31.215Z",
  "v": 0,
  "trace_id": "740c2621cba91797ae2d90231cd748f2",
  "span_id": "3805eec92f517a15",
  "trace_flags": "01"
}
```

We should convert the `nextUrl` to use HTTPS - if so, this would work as expected.

This need to apply to all datasources modified in 44.11.2.

### LINKED ISSUE #45135: Don't allow HTTPS downgrade
Similar to https://github.com/renovatebot/renovate/issues/45134, we should make sure that an HTTPS -> HTTP downgrade is always rejected.

If we can keep it at HTTPS, that's best, but if the HTTPS connection fails, do not proceed with HTTP.

palakkhinvasara:
#451335

* Update same-origin pagination URL resolution to allow HTTP → HTTPS upgrades when the hostname remains the same.
* Keep cross-origin pagination blocked by default to prevent SSRF.
* Preserve the existing experimental opt-in for following cross-origin pagination links.
* Add/adjust URL resolution tests covering same-host HTTP → HTTPS pagination and cross-origin rejection.
* Verify NuGet v2 pagination behavior for both the default restricted mode and the explicit cross-origin opt-in.

## Context

* [x] This closes an existing Issue, 
        Closes: #45135 
        Closes: #45134

* [ ] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

* [ ] No — I did not use AI for this contribution.
* [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
* [x] Yes — substantive assistance (AI-generated non-trivial portions of code, tests, or documentation).

AI was used to help analyze the existing URL/pagination behavior, reason about the SSRF protection, and develop/refine the implementation and tests.

* [x] An agent will draft replies and @username will read them before they are posted. **Name the account: @palakkhinvasara**
* [ ] @username will read and reply directly. **Name the account.**
* [ ] Nobody has explicitly committed to replying.

## Documentation

* [x] No documentation update is required

## How I've tested my work

* [x] Newly added/modified unit tests
* [ ] Code inspection only
* [ ] No unit tests, but ran on a real repository
* [ ] Both unit tests + ran on a real repository

### Test results

* `lib/util/url.spec.ts`: **84/84 tests passed**
* `lib/modules/datasource/nuget/index.spec.ts`: **40/40 tests passed**
* Combined: **124/124 tests passed**

The tests specifically verify:

* same-origin pagination URLs
* relative pagination URLs
* HTTP → HTTPS upgrade for the same hostname
* cross-origin pagination rejection by default
* cross-origin pagination when explicitly enabled
* invalid URL handling

### DIFF
--- a/lib/util/url.ts
+++ b/lib/util/url.ts
@@ -97,13 +97,23 @@ export function resolveSameOriginUrl(
   if (!base) {
     return null;
   }
+
   let resolved: URL;
   try {
     resolved = new URL(nextUrl.toString(), base);
   } catch {
     return null;
   }
-  return resolved.origin === base.origin ? resolved.href : null;
+
+  if (resolved.hostname !== base.hostname) {
+    return null;
+  }
+
+  if (base.protocol === 'https:' && resolved.protocol === 'http:') {
+    resolved.protocol = 'https:';
+  }
+
+  return resolved.href;
 }
 
 export function getQueryString(params: Record<string, any>): string {
--- a/lib/util/url.ts
+++ b/lib/util/url.ts
@@ -105,6 +105,7 @@ export function resolveSameOriginUrl(
     return null;
   }
 
+  // If the base URL is HTTPS and the resolved URL is HTTP, but has no port specified, we can assume that the server intended to use HTTPS. This is a common misconfiguration in some registries.
   if (
     base.protocol === 'https:' &&
     resolved.protocol === 'http:' &&
--- a/lib/util/url.ts
+++ b/lib/util/url.ts
@@ -109,11 +109,15 @@ export function resolveSameOriginUrl(
     return null;
   }
 
-  if (base.protocol === 'https:' && resolved.protocol === 'http:') {
+  if (
+    base.protocol === 'https:' &&
+    resolved.protocol === 'http:' &&
+    resolved.port === ''
+  ) {
     resolved.protocol = 'https:';
   }
 
-  return resolved.href;
+  return resolved.origin === base.origin ? resolved.href : null;
 }
 
 export function getQueryString(params: Record<string, any>): string {
--- a/lib/util/url.ts
+++ b/lib/util/url.ts
@@ -111,6 +111,7 @@ export function resolveSameOriginUrl(
     resolved.protocol === 'http:' &&
     resolved.port === ''
   ) {
+    logger.debug(`Detected protocol downgrade and fixed it: ${resolved.href}`);
     resolved.protocol = 'https:';
   }
 
--- a/lib/util/url.ts
+++ b/lib/util/url.ts
@@ -105,10 +105,6 @@ export function resolveSameOriginUrl(
     return null;
   }
 
-  if (resolved.hostname !== base.hostname) {
-    return null;
-  }
-
   if (
     base.protocol === 'https:' &&
     resolved.protocol === 'http:' &&
--- a/lib/util/url.spec.ts
+++ b/lib/util/url.spec.ts
@@ -247,6 +247,24 @@ describe('util/url', () => {
       ).toBe('https://community.chocolatey.org/api/v2/FindPackagesById?page=2');
     });
 
+    it('does not upgrade HTTP to HTTPS when the next URL has a non-standard port', () => {
+      expect(
+        resolveSameOriginUrl(
+          'https://community.chocolatey.org/api/v2/FindPackagesById',
+          'http://community.chocolatey.org:8080/api/v2/FindPackagesById?page=2',
+        ),
+      ).toBeNull();
+    });
+
+    it('rejects a different port on the same host', () => {
+      expect(
+        resolveSameOriginUrl(
+          'https://registry.example.com:8443/v2/foo',
+          'https://registry.example.com/v2/foo?page=2',
+        ),
+      ).toBeNull();
+    });
+
     it('resolves a relative next URL against the base', () => {
       expect(
         resolveSameOriginUrl(
--- a/lib/util/url.spec.ts
+++ b/lib/util/url.spec.ts
@@ -238,6 +238,15 @@ describe('util/url', () => {
       ).toBe('https://registry.example.com/v2/foo/tags/list?n=10&last=z');
     });
 
+    it('upgrades an HTTP URL to HTTPS when the host is the same', () => {
+      expect(
+        resolveSameOriginUrl(
+          'https://community.chocolatey.org/api/v2/FindPackagesById',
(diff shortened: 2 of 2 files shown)


palakkhinvasara:
hi @viceice  , can you pls review this pr

Thanks 
palak

lib/util/url.ts
```diff
@@ -97,13 +97,23 @@ export function resolveSameOriginUrl(
...
+
+  if (resolved.hostname !== base.hostname) {
+    return null;
+  }
+
+  if (base.protocol === 'https:' && resolved.protocol === 'http:') {
+    resolved.protocol = 'https:';
+  }
+
+  return resolved.href;
```
mjpieters: This now ignores the port of the URL, which is also part of the origin. Next, if the port is explicitly listed, then you can't just change the scheme from http to https either, as the protocols generally use different ports!

Specifically, if the resolved address set a non-standard port, then you can't map it from http to https anyway and you should not adjust. And if the base set a non-standard port, you can't know if the change from `https:` to `http:` can be mapped straight back to `https:` with the custom port. It's best to just ignore such URLs.

I think the above overcomplicated the fix, and there is no need to remove the check for origins. The following works for any http base that doesn't set the port (so uses the standard port, 80), but explicitly listed port would cause the origin not to match:


```suggestion

  // keep communication encrypted, stay on https. This only works
  // for URLs that do not explicitly set a port
  if (base.protocol === 'https:' && resolved.protocol === 'http:' && resolved.port === "") {
    resolved.protocol = 'https:';
  }

  return resolved.origin === base.origin ? resolved.href : null;
```

palakkhinvasara: added the changes as suggested by viceice .
Thanks for the review , kindly review the changes and let me kow.

Thanks & Regards 
Palak

lib/util/url.ts
```diff
@@ -97,12 +97,26 @@ export function resolveSameOriginUrl(
...
   try {
     resolved = new URL(nextUrl.toString(), base);
   } catch {
     return null;
   }
+
+  if (resolved.hostname !== base.hostname) {
+    return null;
+  }
+
```
mjpieters: This check is redundant now; it can be dropped entirely.

lib/util/url.ts
mjpieters: > added the changes as suggested by viceice .

I'm not viceice 😃

Note: I am not a maintainer of this project, I am just an interested bystander with domain expertise.

lib/util/url.ts
palakkhinvasara: it is a defensive mechanism .

mjpieters: To defend against what? The origin check at the bottom includes the hostname already.

lib/util/url.ts
```diff
@@ -97,12 +97,26 @@ export function resolveSameOriginUrl(
...
+  if (
```
mjpieters, line 109: This check really needs a comment explaining why this is here. My original suggestion had such a comment:

```suggestion
  // keep communication encrypted, stay on https. This only works
  // for URLs that do not explicitly set a port
  if (
```

lib/util/url.ts
palakkhinvasara: understood

lib/util/url.ts
mjpieters: I'm not sure why you marked this as resolved?

Are you saying that this check here is self-explanatory and doesn't need a clarifying comment for future maintainers?

mjpieters:
There is no need for requesting a review from me.

- I am subscribed to this PR so I already am notified of further changes.
- I am not a maintainer for this project. My approval or rejection of your changes has no official status, and would likely be dismissed as noise.

I am just an interested bystander; I maintain projects that rely on renovate to keep the dependencies updated.

lib/util/url.ts
palakkhinvasara: sorry my mistake 

if (
    base.protocol === 'https:' &&
    resolved.protocol === 'http:' &&
    resolved.port === ''
  ) {

resolved.port==' '  has been added so that when the port is not explicitly mentioned it should upgrade to https
when the port is mentioned , the code should not become https://example.com:8080

mjpieters: I think you misunderstood me. I am not asking you to explain the code to me. I am saying that there needs to be a comment _in the code_ for anyone else that tries to read and understand this function in the future.

I think this specific block needs a comment because the stated goal of this function is to resolve a same-origin URL, and adjusting the scheme here _deviates_ from that goal.

So, we need a comment here to document why it is here, to justify why the normal same-origin security measures are being ignored.

palakkhinvasara: yeah i got it now ,, added the comment as asked by you

viceice (dismissed):

lib/util/url.ts
```diff
@@ -97,12 +97,23 @@ export function resolveSameOriginUrl(
...
     return null;
   }
+
+  // If the base URL is HTTPS and the resolved URL is HTTP, but has no port specified, we can assume that the server intended to use HTTPS. This is a common misconfiguration in some registries.
+  if (
+    base.protocol === 'https:' &&
+    resolved.protocol === 'http:' &&
+    resolved.port === ''
+  ) {
+    resolved.protocol = 'https:';
```
viceice, line 115: log a message that a downgrade was detected and fixed.

palakkhinvasara: added

viceice (approved):

secustor (approved):

PR #45162, opened 2026-08-08
fix(platform/forgejo): retry automerge until mergeable + reattempt on update

silenthooligan:
## Changes

Mirrors #43449 (the same two fixes for the Gitea adapter) onto the Forgejo platform, which was split out of Gitea in #37075 and carries the same two gaps.

### Bug 1: race on PR creation (single attempt, no retry)

`createPr` issues the merge API call right after the branch is pushed. Forgejo computes `mergeable` asynchronously, so hitting the merge endpoint before that computation finishes fails. The code as of `main` makes one attempt and swallows the error:

```ts
try {
  await helper.mergePR(config.repository, gpr.number, {
    Do: ..., merge_when_checks_succeed: true, delete_branch_after_merge: true,
  });
  logger.debug({ prNumber: gpr.number }, 'Forgejo-native automerge: success');
} catch (err) {
  logger.warn({ err, prNumber: gpr.number }, 'Forgejo-native automerge: fail');
}
```

`tryPrAutomerge` now polls the PR until `mergeable` is true before calling merge, with exponential backoff, then merges anyway on the last attempt so behaviour is never worse than today. This follows the existing GitLab `tryPrAutomerge` pattern. Tunable via `RENOVATE_X_FORGEJO_AUTO_MERGEABLE_CHECK_ATTEMPTS` (default 5) and `RENOVATE_X_FORGEJO_AUTO_MERGEABLE_CHECK_DELAY` (default 250 ms), matching the GitLab and Gitea equivalents.

### Bug 2: no retroactive flagging on existing PRs

The branch worker calls `platform.reattemptPlatformAutomerge` after every branch update. The Forgejo adapter does not implement the method, so the optional-chained call is a silent no-op, and PRs created before a `packageRules` change that enables automerge never get flagged.

The version gate and all log message wording are unchanged, including the `10.0.0-0` lower bound that keeps Codeberg's `11.0.1-99-c504062+gitea-1.22.0` style versions working.

## Context

## AI assistance disclosure

Claude Fable 5, via Claude Code, ported the Gitea implementation from #43449 to the Forgejo adapter and wrote the accompanying tests and docs. The original Gitea implementation being mirrored here is my own earlier work. Everything was run locally against the repo's own gates before submission (see below).

Who answers review comments:

- [x] An agent will draft replies and @silenthooligan will read them before they are posted. **Name the account.**

Both new environment variables are documented in `docs/usage/self-hosted-experimental.md`, alongside the existing Gitea and GitLab entries.

`pnpm vitest run lib/modules/platform/forgejo` passes at 268 tests. Coverage of `lib/modules/platform/forgejo` is 100% of statements, functions and lines, and 304/306 branches. The two uncovered branches are pre-existing on `main` (verified by running the same coverage report against an unmodified tree, which gives 300/302 on the same two spots); this PR adds four new branches and covers all four.

`pnpm check --all lib/modules/platform/forgejo`, `pnpm type-check`, `pnpm markdown-lint` and `pnpm doc-fence-check` are all clean.


lib/modules/platform/forgejo/index.ts
```diff
@@ -198,6 +202,105 @@ async function fetchRepositories({
...
+ */
+export async function reattemptPlatformAutomerge({
+  number,
+  platformPrOptions,
+}: ReattemptPlatformAutomergeConfig): Promise<void> {
+  try {
+    await tryPrAutomerge(number, platformPrOptions);
+
+    logger.debug(`PR platform automerge re-attempted...prNo: ${number}`);
+  } catch (err) /* v8 ignore next -- defensive: automerge re-attempt failures are logged and swallowed, not simulated in specs */ {
```
viceice: add a test instead

lib/modules/platform/forgejo/index.spec.ts
```diff
@@ -2025,6 +2033,154 @@ describe('modules/platform/forgejo/index', () => {
         });
       },
     );
+
+    it('retries until mergeable is true before calling merge', async () => {
+      process.env.RENOVATE_X_FORGEJO_AUTO_MERGEABLE_CHECK_DELAY = '1';
```
viceice: mock `setTimeout`, so we don't slow down tests.

lib/modules/platform/forgejo/index.ts
silenthooligan: Done. `helper.getPR` in the poll loop isn't wrapped, so a rejecting mergeable check lands here. Ignore removed, test added. Module coverage unchanged at 100%.

lib/modules/platform/forgejo/index.spec.ts
silenthooligan: Done, same as the gitlab/github specs: `vi.mock('timers/promises')` + `vi.mocked(_timers)`. Dropped the `RENOVATE_X_FORGEJO_AUTO_MERGEABLE_CHECK_DELAY = '1'` override, so the tests run the real default and assert the schedule: `[[250], [1000]]` retry case, `[[250]]` exhausted case. Env cleanup moved to `beforeEach`.

lib/modules/platform/forgejo/index.spec.ts
```diff
@@ -2025,6 +2040,172 @@ describe('modules/platform/forgejo/index', () => {
...
+        platformPrOptions: { usePlatformAutomerge: true },
+      });
+
+      expect(res).toMatchObject({ number: 42 });
+      expect(logger.logger.debug).toHaveBeenCalledWith(
+        expect.objectContaining({ prNumber: 42 }),
+        'Forgejo-native automerge: success',
+      );
+      // exponential backoff between polls, no sleep before the first poll
+      expect(timers.setTimeout.mock.calls).toMatchObject([[250], [1000]]);
```
viceice: use `toHaveBeenCalledWith` and alike, then you don't need this `const timers = vi.mocked(_timers);` dummy code

lib/modules/platform/forgejo/index.spec.ts
```diff
@@ -2025,6 +2040,172 @@ describe('modules/platform/forgejo/index', () => {
...
+        prBody: mockNewPR.body,
+        platformPrOptions: { usePlatformAutomerge: true },
+      });
+
+      expect(res).toMatchObject({ number: 42 });
+      expect(logger.logger.debug).toHaveBeenCalledWith(
+        'PR not mergeable after 2 attempts, merging anyway...prNo: 42',
+      );
+      // no sleep after the final poll
+      expect(timers.setTimeout.mock.calls).toMatchObject([[250]]);
```
viceice: same

viceice (commented):
otherwise LGTM

viceice (approved):

lib/modules/platform/forgejo/index.ts
```diff
@@ -198,6 +202,105 @@ async function fetchRepositories({
...
+    logger.debug(
+      { prNumber },
+      `Forgejo-native automerge: not supported on this version of Forgejo. Use 10.0.0 or newer.`,
+    );
+    return;
+  }
+
+  const env = getEnv();
+  const retryTimes = parseInteger(
+    env.RENOVATE_X_FORGEJO_AUTO_MERGEABLE_CHECK_ATTEMPTS,
```
secustor, line 241: Instead of those env variables create experimental config options, which is independent of the platform (`prMergeabilityCheckAttempts`). That way we can reuse it for Gitea and Gitlab

viceice:
conflicted

silenthooligan:
Closing in favor of #43449. Forgejo runs on the Gitea code since #45825. A separate Forgejo fix is no longer needed. @secustor, #43449 adds `prMergeabilityCheckAttempts` as you suggested. Thanks @viceice for the review.

PR #45176, opened 2026-08-09
fix(changelog): match release notes for host-qualified names

wittdennis:
## Changes

`getExactReleaseMatch` builds its pattern by interpolating `packageName` and `depName` directly:

```ts
const exactReleaseReg = regEx(
  `(?:^|/)(?:${packageName}|${depName})[@_/-]v?${version}`,
);
```

The `(?:^|/)` anchor allows a tag prefix, but it anchors the *entire* name. For anything distributed through a registry, that name is host-qualified (`<registry>/<org>/<repo>/<chart>`), while the Git tag is scoped by the trailing name alone (`<chart>-<version>`). The host can never appear in a tag, so the match is structurally impossible and the PR body is rendered without release notes.

This affects every OCI-distributed Helm chart in a multi-chart repository, which is an increasingly common publishing layout. Go module paths in monorepos are qualified the same way and hit the same gap (`github.com/<org>/<repo>/service/s3` vs. tag `service/s3/v1.2.3`).

There is no config workaround today. `extractVersion` is the only tag-shaping knob and it is dual-purpose: `applyExtractVersion` runs at the datasource layer and *drops* every release that does not match, before release-note matching ever sees them. Using it to describe the tag format therefore breaks version detection for the package.

### What this PR does

`getNamePatterns()` now derives the alternatives for the match:

| name | patterns tried | matches tag |
| --- | --- | --- |
| `<registry>/<org>/<repo>/<chart>` | full, host-stripped, trailing segment | `<chart>-1.2.3` |
| `<registry>/<org>/<chart>` | full, host-stripped, trailing segment | `<org>/<chart>-1.2.3` |
| `@scope/pkg` | full only (unchanged) | — |

Host detection follows the OCI distribution reference rule: the first path segment is a host when it contains a `.` or a `:` (plus `localhost`). That deliberately excludes scoped npm names, so the widening only reaches names that are registry- or module-path-qualified, and existing tag prefixes stay exactly as specific as they are today.

False-positive risk is small: `candidateReleases` is already narrowed to `tag.endsWith(version)`, so a wrong hit would need a tag shaped `*<trailing-name>[@_/-]v?<that exact version>`.

Two smaller fixes ride along:

- The interpolated names and version are now escaped with `RegExp.escape`. Previously `some.registry/dep` matched the tag `someXregistry/dep-1.0.0`, and a name containing regex metacharacters could throw out of `regEx()`.
- `getExactReleaseMatch` takes `string | undefined` and returns early when no name is available, removing two `!` assertions that contradicted `ChangeLogProject`'s optional `packageName`/`depName`.

## Context

## AI assistance disclosure

Claude Opus 5, via Claude Code. It diagnosed the root cause from debug logs of a real Renovate run, wrote the implementation, the unit tests and this description. All of it was reviewed before submitting.

Who answers review comments:

- [x] @wittdennis will read and reply directly.

Five tests added to `release-notes.spec.ts`: trailing-name match (with a decoy release at the adjacent version, to prove the right one is picked), host-stripped-name match, scoped-npm negative, regex-metacharacter negative, and the no-name-set case. `release-notes.ts` keeps 100% statement, function and line coverage; the two remaining branch misses are pre-existing.

The originating case was found on a self-hosted Forgejo instance, so there is no public repository to link.

### DIFF
--- a/lib/workers/repository/update/pr/changelog/release-notes.spec.ts
+++ b/lib/workers/repository/update/pr/changelog/release-notes.spec.ts
@@ -1556,6 +1556,270 @@ describe('workers/repository/update/pr/changelog/release-notes', () => {
       });
     });
 
+    it('matches a host-qualified name on its trailing name', async () => {
+      githubReleasesMock.mockResolvedValueOnce([
+        {
+          id: 1,
+          version: 'exampleChart-1.0.0',
+          releaseTimestamp: '2020-01-01' as Timestamp,
+          url: 'wrong/url/tag.com',
+          name: 'some/dep',
+          description: 'some old body',
+        },
+        {
+          id: 2,
+          version: 'exampleChart-1.0.1',
+          releaseTimestamp: '2020-01-02' as Timestamp,
+          url: 'correct/url/tag.com',
+          name: 'some/dep',
+          description: 'some body',
+        },
+      ]);
+
+      const res = await getReleaseNotes(
+        {
+          ...githubProject,
+          repository: 'some/other-repository',
+          packageName: 'some.registry/some/charts/exampleChart',
+          depName: 'some.registry/some/charts/exampleChart',
+        },
+        partial<ChangeLogRelease>({
+          version: '1.0.1',
+          gitRef: '1.0.1',
+        }),
+        partial<BranchUpgradeConfig>(),
+      );
+
+      expect(res).toEqual({
+        url: 'correct/url/tag.com',
+        notesSourceUrl:
+          'https://api.github.com/repos/some/other-repository/releases',
+        id: 2,
+        tag: 'exampleChart-1.0.1',
+        name: 'some/dep',
+        body: 'some body\n',
+      });
+    });
+
+    it('matches a host-qualified name on its unqualified name', async () => {
+      githubReleasesMock.mockResolvedValueOnce([
+        {
+          id: 1,
+          version: 'charts/podinfo-1.0.0',
+          releaseTimestamp: '2020-01-01' as Timestamp,
+          url: 'correct/url/tag.com',
+          name: 'some/dep',
+          description: 'some body',
+        },
+      ]);
+
+      const res = await getReleaseNotes(
+        {
+          ...githubProject,
+          repository: 'some/other-repository',
+          packageName: 'ghcr.io/charts/podinfo',
+        },
+        partial<ChangeLogRelease>({
+          version: '1.0.0',
+          gitRef: '1.0.0',
+        }),
+        partial<BranchUpgradeConfig>(),
+      );
+
+      expect(res).toEqual({
+        url: 'correct/url/tag.com',
+        notesSourceUrl:
+          'https://api.github.com/repos/some/other-repository/releases',
+        id: 1,
+        tag: 'charts/podinfo-1.0.0',
+        name: 'some/dep',
+        body: 'some body\n',
+      });
+    });
+
+    it('matches a host-qualified name with a port', async () => {
+      githubReleasesMock.mockResolvedValueOnce([
+        {
+          id: 1,
+          version: 'exampleChart-1.0.0',
+          releaseTimestamp: '2020-01-01' as Timestamp,
+          url: 'correct/url/tag.com',
+          name: 'some/dep',
+          description: 'some body',
+        },
+      ]);
+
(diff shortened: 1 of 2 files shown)


viceice (dismissed):

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -206,6 +217,36 @@ function getExactReleaseMatch(
...
+    names.add(name);
+
+    const unqualifiedName =
+      hostQualifiedNameRegex.exec(name)?.groups?.unqualifiedName;
+    if (isNonEmptyString(unqualifiedName)) {
+      names.add(unqualifiedName);
+      names.add(unqualifiedName.split('/').pop()!);
+    }
+  }
+  return [...names].map((name) => RegExp.escape(name));
```
secustor, line 256: Use `regexEscape` util

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -206,6 +217,36 @@ function getExactReleaseMatch(
...
+    if (!isNonEmptyString(name)) {
+      continue;
+    }
+    names.add(name);
+
+    const unqualifiedName =
+      hostQualifiedNameRegex.exec(name)?.groups?.unqualifiedName;
+    if (isNonEmptyString(unqualifiedName)) {
+      names.add(unqualifiedName);
+      names.add(unqualifiedName.split('/').pop()!);
```
secustor: A trailing slash will lead to an empty string being added

lib/workers/repository/update/pr/changelog/release-notes.spec.ts
```diff
@@ -1355,6 +1355,170 @@ describe('workers/repository/update/pr/changelog/release-notes', () => {
...
+          version: '1.0.0',
+          gitRef: '1.0.0',
+        }),
+        partial<BranchUpgradeConfig>(),
+      );
+
+      expect(res).toEqual({
+        url: 'correct/url/tag.com',
+        notesSourceUrl:
+          'https://api.github.com/repos/some/other-repository/releases',
```
secustor, line 1431: Add a test with ports

lib/workers/repository/update/pr/changelog/release-notes.ts
```diff
@@ -191,13 +197,18 @@ export async function getReleaseNotes(
...
   releases: ChangeLogNotes[],
 ): ChangeLogNotes | undefined {
+  const namePatterns = getNamePatterns(packageName, depName);
+  if (!namePatterns.length) {
+    return undefined;
+  }
+
   const exactReleaseReg = regEx(
-    `(?:^|/)(?:${packageName}|${depName})[@_/-]v?${version}`,
+    `(?:^|/)(?:${namePatterns.join('|')})[@_/-]v?${RegExp.escape(version)}`,
```
secustor, line 216: Same here

lib/workers/repository/update/pr/changelog/release-notes.ts
wittdennis: escapeRegExp in lib/util/regex.ts carries `@deprecated use RegExp.escape instead`. 

Thats why I used it here. Happy to switch but it seems kinda weird to me to use a deprecated function.

lib/workers/repository/update/pr/changelog/release-notes.ts
wittdennis: Added an empty string guard

lib/workers/repository/update/pr/changelog/release-notes.spec.ts
wittdennis: Added a test with ports

lib/workers/repository/update/pr/changelog/release-notes.ts
wittdennis: Same as on the other spot. Used RegExp.escape because escapeRegExp is marked deprecated

lib/workers/repository/update/pr/changelog/release-notes.ts
wittdennis: Does this suffice or do you have anything other in mind?

viceice (approved):

secustor (approved):

PR #45199, opened 2026-08-10
feat(manager/github-actions): add support for Actions Lockfiles

### LINKED ISSUE #45190: Add support for GitHub Actions lockfile
Related to https://github.com/renovatebot/renovate/issues/44753, but for the private preview: https://github.com/github/actions-lockfile

jamietanna:
## Changes

In June, GitHub announced a public preview for the GitHub Actions
lockfile functionality[0].

To add support for this, we can extend our support for GitHub Actions to
handle the presence of an `actions.lock` file, and if so, update it
after updating other manager files.

Although our ideal case would be to have a pure Typescript-based parsing
and updating of the `actions.lock`, due to the early nature of the
Actions Lockfile syntax, and its lack of stability, we can invoke the
`gh actions-lock` extension when updating.

After the schema has reached a v1 release, we can then migrate to
Typescript.

To make sure that the version of the `actions-lock` extension is always
version pinned, we can expose a `constraints` to control the version,
but allow Renovate to keep the pinned default updated as needed.

This doesn't imply that we have support for `lockFileMaintenance`, as we
are only updating the lockfile when there is a dependency update, not
periodically recreating the lockfile from the current ranges.

`gh actions-lock` re-reads every workflow from disk and rewrites the
whole lockfile, so it does not fit `updateArtifacts`, which is called
once per updated package file.

We can instead follow the shape the `npm` manager uses for its
lock files, and have `processBranch` invoke this once per branch, after
every updated package file has been written to disk.

We skip workflows that aren't onboarded to the lockfile, but still run
for any other package file, as a local composite action can be a
transitive dependency of a workflow which is.

`gh actions-lock` rewrites the workflows it manages back to plain refs
when it regenerates them, which strips out any inline digest pin:

```diff
-        uses: actions/setup-java@1bcf9fb12cf4aa # v5.4.0
+        uses: actions/setup-java@v5.7.0
```

This would lead to a repository with `pinDigests` set "flip flopping",
and would cause a lot of erroneous noise for repository owners.

Instead, we can make sure that:

- we don't suggest a `pinDigest` for workflows that are managed by the
	lockfile
- digest pinning is still enabled for `docker://` references
- already-committed digests are updated when they're present - we only
	avoid suggesting /new/ digests to be pinned

[0]: https://github.com/github/actions-lockfile

## Context

- [x] This closes an existing Issue, Closes: #45190

## AI assistance disclosure

Who answers review comments:

- [x] @jamietanna will read and reply directly. **Name the account.**

The public repository:

- https://github.com/JamieTanna-Mend-testing/assertj-generator/pull/2
  - vs Dependabot: https://github.com/assertj/assertj-generator/pull/328
  - Repo has `Require actions to be pinned to a full-length commit SHA `
- https://github.com/JamieTanna-Mend-testing/jessehouwing-actions-dependency-submission/pull/8


lib/modules/manager/github-actions/__fixtures__/actions.lock
jamietanna, line 1: Are we OK with a fixture, or should this be inline?

lib/util/exec/types.ts
```diff
@@ -172,6 +172,14 @@ export function isToolName(value: unknown): value is ToolName {
  * Additional constraints that can be specified for some Managers, but are **not** tools that Containerbase supports, with optional description.
  */
 export const additionalConstraintDefinitions = [
+  {
+    name: 'ghActionsLock',
+    description: `Used in the \`github-actions\` manager to specify a release tag for the [\`github/gh-actions-lock\`](https://github.com/github/gh-actions-lock) \`gh\` CLI extension, which regenerates \`.github/workflows/actions.lock\`.
+
+Must be a full release tag, prefixed with \`v\`, such as \`v0.1.7\`. Set it to an empty string to always install the latest release.
+
+Unlike the other additional constraints, this has no effect when using [\`binarySource=global\`](./self-hosted-configuration.md#binarysource): Renovate then leaves the \`gh\` installation alone, and the extension must be installed and pinned by the administrator.`,
```
jamietanna:
```suggestion
```

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,268 @@
...
+
+  const env: ExtraEnv = {};
+
+  if (isEnterprise) {
+    const enterpriseToken = findToken(endpoint);
+    if (enterpriseToken) {
+      env.GH_HOST = hostname;
+      env.GH_ENTERPRISE_TOKEN = enterpriseToken;
+    }
+  }
```
jamietanna, line 57: Hasn't been tested "for real"

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,268 @@
...
+  // Narrowing rewrites a freshly pinned `@v4` to `@v4.2.1`, which would contradict the `newValue` in the PR we are raising, and would quietly defeat a deliberately floating major tag.
+  //
+  // `--no-migrate-local-actions` stops it rewriting same-repo `uses: ./…` refs to the `uses: $/…` form, which has nothing to do with the update at hand.
+  const commands = [
+    'gh actions-lock --no-interactive --no-narrow --no-migrate-local-actions',
+  ];
+
+  // `binarySource=global` means the administrator provisions tooling, so we must not reach into their `gh` install to add or re-pin an extension.
+  // They own the extension version, and `constraints.ghActionsLock` does not apply.
+  if (GlobalConfig.get('binarySource') === 'global') {
```
jamietanna, line 181: Is this unnecessary? Or only once `install-tool` knows about it in https://github.com/containerbase/base/issues/7279 ?

lib/modules/manager/github-actions/__fixtures__/actions.lock
secustor: Should be inline

lib/config/options/index.ts
```diff
@@ -2801,12 +2801,13 @@ const options: Readonly<RenovateOptions>[] = [
     description:
       'Configuration object to define language or manager version constraints.',
     type: 'object',
-    default: {},
+    default: { ghActionsLock: 'v0.1.6' },
```
secustor, line 2805: Why has to be this set here?

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,274 @@
...
+import { ActionsLockfile } from './schema.ts';
+
+/** The `gh` CLI extension that generates and verifies the lockfile. */
+const extensionName = 'github/gh-actions-lock';
+
+/** The `uses:` dep types the lockfile graph is built from. */
+const lockedDepTypes = ['action', 'docker'];
+
+/** Mirrors the shape `processBranch` already concatenates onto its config for lock file updates. */
+export interface ActionsLockfileResult {
```
secustor: Should be in types.ts

lib/modules/manager/github-actions/__fixtures__/actions.lock
viceice: better inline if possible

renovate.json
```diff
@@ -131,6 +131,13 @@
       "datasourceTemplate": "github-releases",
       "packageNameTemplate": "containerbase/erlang-prebuild",
       "versioningTemplate": "loose"
+    },
+    {
+      "customType": "regex",
+      "managerFilePatterns": ["lib/config/options/index.ts"],
+      "matchStrings": ["ghActionsLock: '(?<currentValue>.*)'"],
+      "datasourceTemplate": "github-releases",
+      "packageNameTemplate": "github/gh-actions-lock"
```
viceice, line 145: we should change `semanticCommitType` to  `fix` in a package rule, so it doesn't hide in a default `chore`

lib/workers/repository/update/branch/index.ts
```diff
@@ -634,6 +635,18 @@ export async function processBranch(
       config.updatedArtifacts = (config.updatedArtifacts ?? []).concat(
         additionalFiles.updatedArtifacts,
       );
+      // `gh actions-lock` rewrites the whole lockfile from the workflows on disk, so like the lock files above it runs once here, after every updated package file has been written, rather than per package file.
+      const actionsLockfile = await updateActionsLockfile(
```
viceice, line 639: we should abstract it to a optional manager function, which is run on branch, so we don't need special manager handling here

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,274 @@
...
+  updatedArtifacts: FileChange[];
+  artifactErrors: ArtifactError[];
+}
+
+/**
+ * An `ExecError` always carries a `stderr` property, so nullish coalescing would keep an empty string and render an artifact error with no message at all.
+ */
+function execErrorMessage(err: {
```
secustor: I think this should be an util if we have it not already

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  if (isEnterprise) {
+    const enterpriseToken = findToken(endpoint);
+    if (enterpriseToken) {
+      env.GH_HOST = hostname;
+      env.GH_ENTERPRISE_TOKEN = enterpriseToken;
+    }
+  }
+
+  const token = findToken(isEnterprise ? githubComEndpoint : endpoint);
```
viceice: isn't endpoint always `api.github.com` here? 🤔

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,274 @@
...
+}
+
+/**
+ * `gh` takes the token from a different variable for GitHub Enterprise Server, and has to be told which host to talk to.
+ *
+ * Public actions are still resolved against github.com from there, so pass a token for both hosts.
+ */
+function getTokenEnv(): ExtraEnv {
+  const githubComEndpoint = 'https://api.github.com/';
+  const endpoint =
```
secustor, line 40: use detectPlatform here

lib/modules/manager/github-actions/extract.ts
```diff
@@ -354,11 +356,33 @@ function extractWithYAMLParser(
...
+
+  const parsed = ActionsLockfile.safeParse(content);
+  if (!parsed.success) {
+    // `updateActionsLockfile` warns about this once per branch, so stay quiet.
+    // It refuses to regenerate a lock file which it cannot parse, so treat everything as managed: pinning inline here would raise a PR which pins the digests the tool owns and leaves the lock file stale.
+    logger.debug(`Failed to parse ${actionsLockFile}`);
+    return true;
+  }
+
+  return isLockfileManaged(packageFile, parsed.data.workflows ?? {});
```
viceice: `coerceObject`

lib/modules/manager/github-actions/extract.ts
```diff
@@ -354,11 +356,33 @@ function extractWithYAMLParser(
...
+    return false;
+  }
+
+  const parsed = ActionsLockfile.safeParse(content);
+  if (!parsed.success) {
+    // `updateActionsLockfile` warns about this once per branch, so stay quiet.
+    // It refuses to regenerate a lock file which it cannot parse, so treat everything as managed: pinning inline here would raise a PR which pins the digests the tool owns and leaves the lock file stale.
+    logger.debug(`Failed to parse ${actionsLockFile}`);
+    return true;
+  }
```
viceice, line 372: we should cache this, so it's not reloaded and parsed again and again for each package file (can be a lot files)

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,274 @@
...
+      ' ' +
+      // we don't have a terminal to answer with
+      '--no-interactive' +
+      ' ' +
+      // keep the ref that we wrote, to avoid a `@v4` being rewritten as `@v4.2.1`
+      '--no-narrow' +
+      ' ' +
+      // on-by-default, but unrelated to our lockfile updates
+      '--no-migrate-local-actions',
+  ];
```
secustor, line 176: Rather join by space

lib/modules/manager/github-actions/schema.ts
```diff
@@ -68,6 +68,19 @@ export const Workflow = Yaml.pipe(
...
+/**
+ * The `actions.lock` schema is owned by the `gh actions-lock` CLI, and is still unstable, so don't rely on the structure too heavily.
+ *
+ * If `workflows` is reshaped/removed, we fail loudly to silently prevent lockfile updates.
+ *
+ * TODO #45191: after v1 of the schema, make this Typescript only
+ */
+export const ActionsLockfile = Yaml.pipe(
+  z.object({
+    workflows: z.record(z.string(), z.unknown()).optional(),
```
viceice: with the optional we don't fail if it's missing because they renamed / moved it. so should not be optional

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,274 @@
...
+  const parsedLockfile = ActionsLockfile.safeParse(existingLockFileContent);
+  if (!parsedLockfile.success) {
+    logger.warn(
+      { err: parsedLockfile.error, lockFile: actionsLockFile },
+      'Failed to parse the GitHub Actions lockfile, so not regenerating it',
+    );
+    return empty;
+  }
+
+  if (!hasLockedUpdate(config, parsedLockfile.data.workflows ?? {})) {
```
viceice: `coerceObject`

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,274 @@
...
+    logger.debug(`No locked dependencies updated for ${actionsLockFile}`);
+    return empty;
+  }
+
+  logger.debug(`Regenerating ${actionsLockFile}`);
+
+  // Ensure that in-memory file contents are flushed to disk, so `actions-lock` can run against it
+  const writtenContent = new Map<string, string>();
+  for (const file of config.updatedPackageFiles ?? []) {
```
viceice: `coerceArray`

lib/modules/manager/github-actions/schema.ts
```diff
@@ -68,6 +68,19 @@ export const Workflow = Yaml.pipe(
   z.union([WorkFlowJobs, Actions, z.null()]),
 ).catch(withDebugMessage(null, 'Does not match schema'));
 
+/**
+ * The `actions.lock` schema is owned by the `gh actions-lock` CLI, and is still unstable, so don't rely on the structure too heavily.
+ *
+ * If `workflows` is reshaped/removed, we fail loudly to silently prevent lockfile updates.
+ *
+ * TODO #45191: after v1 of the schema, make this Typescript only
```
secustor: ?

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,274 @@
...
+    if (file.type === 'addition') {
+      const contents = file.contents!.toString();
+      await writeLocalFile(file.path, contents);
+      writtenContent.set(file.path, contents);
+    }
```
viceice: we should check that we only write yaml files, file contents can be buffer / binary if some user groups this with other managers like gradle (jar update)

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  const tokenEnv = getTokenEnv();
+
+  const execOptions: ExecOptions = {
+    // The tool rescans every workflow, so it runs from the repository root
+    cwd: GlobalConfig.get('localDir'),
+    // `extraEnv` is what forwards a variable into the Docker sidecar, but on its own it is only a default: `getChildEnv` lets a same-named variable in Renovate's own environment win.
```
viceice, line 163: another reason to remove docker mode soon 😁

lib/modules/manager/github-actions/artifacts.ts
```diff
@@ -0,0 +1,274 @@
...
+    // The token we resolved from the host rules must not be shadowed by an ambient `GH_TOKEN`, so force it as well.
+    extraEnv: tokenEnv,
+    env: tokenEnv,
+    toolConstraints: [{ toolName: 'gh', constraint: config.constraints?.gh }],
+    docker: {},
+  };
+
+  // TODO #45191: after v1 of the schema, make this Typescript only
```
viceice: if possible 😁

lib/modules/manager/github-actions/artifacts.ts
viceice: is needed until the mentioned containerbase issue is done

lib/modules/manager/github-actions/artifacts.ts
jamietanna: Good shout - I've captured https://github.com/renovatebot/renovate/issues/45214 for a linting rule to catch this - there are some types across other files we may want to extract out too

lib/workers/repository/update/branch/index.ts
jamietanna: I've not done it yet as I prefer the "rule of 3" for refactoring. I can do so if we're comfortable with the shape of this and npm being how we want to do it for all managers going forwards

lib/config/options/index.ts
jamietanna: I've kept it here to make it much clearer that it's a default constraint - it'll show up in docs + JSON Schema

lib/modules/manager/github-actions/artifacts.ts
jamietanna: https://github.com/renovatebot/renovate/pull/45218

secustor (approved):

viceice (approved):

PR #45200, opened 2026-08-10
[merged] ci: improve performance, caching + harden GitHub Actions config

jamietanna:
## Context

## AI assistance disclosure

Who answers review comments:

- [x] @jamietanna will read and reply directly. **Name the account.**

The public repository: <URL>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -297,9 +297,6 @@ jobs:
       - name: Lint documentation
         run: pnpm lint-documentation
 
-      - name: Markdown lint
-        run: pnpm markdown-lint
-
   lint-other:
     needs:
       - setup-build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -499,7 +499,7 @@ jobs:
         with:
           use_oidc: true
           directory: coverage/lcov
-          fail_ci_if_error: github.event_name != 'pull_request'
+          fail_ci_if_error: ${{ github.event_name != 'pull_request' }}
           verbose: true
 
   coverage-threshold:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -321,7 +321,9 @@ jobs:
         with:
           path: .cache/tsbuildinfo
           key: tsbuildinfo-${{ hashFiles('pnpm-lock.yaml') }}
-          restore-keys: tsbuildinfo-${{ hashFiles('pnpm-lock.yaml') }}-
+          restore-keys: |
+            tsbuildinfo-${{ hashFiles('pnpm-lock.yaml') }}-
+            tsbuildinfo-
 
       - name: Type check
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -297,9 +297,6 @@ jobs:
       - name: Lint documentation
         run: pnpm lint-documentation
 
-      - name: Markdown lint
-        run: pnpm markdown-lint
-
```
secustor, line 302: Why has been this been deleted?

jamietanna: Because it's already there in that job - https://github.com/renovatebot/renovate/blob/3f20dca6e0f397c7aaba678e232f0acddb0e9f32/.github/workflows/build.yml#L288-L301 (on `main`) shows that we have two executions of it

.github/workflows/build.yml
```diff
@@ -845,32 +882,42 @@ jobs:
         with:
           node-version: ${{ needs.setup-build.outputs.node-version }}
           os: ${{ runner.os }}
+          # ensure that the release job installs fresh, with installed pacakges verified against the lockfile's integrity hashes
```
viceice, line 885:
```suggestion
          # ensure that the release job installs fresh, with installed packages verified against the lockfile's integrity hashes
```

why not running on install on cache, so we can use cache and also verify? 🤔

jamietanna: > running on install on cache

what do you mean, sorry?

.github/workflows/build.yml
```diff
@@ -603,6 +617,30 @@ jobs:
           needs.lint-actions.result == 'skipped'
         run: exit 1
 
+      - name: Fail for failed or cancelled linting
```
jamietanna, line 620: These are required by GitHub Status Check - but we could relax it after this (if we wanted to)

.github/workflows/build.yml
viceice: i mean wouldn't an `pnpm install` catch any cache inconsistencies? so we can restore cache and do a `pnpm install` to ensure validation instead always skip cache restore?

jamietanna: I guess we want to avoid any potential cache poisoning. We can look at the before/after for how much a cache miss costs us time wise, and then look at whether we want to add an additional stage to check the cache is correct before we then use it in the release job?

viceice (approved):

jamietanna:
I'll rebase-merge once https://github.com/renovatebot/renovate/queue/main is clear

jamietanna:
Hmm, not sure why this isn't showing as closed - I've manually merged this (via rebase) onto `main`

jamietanna:
~16s taken (on https://github.com/renovatebot/renovate/actions/runs/32124905460/job/95675812572) to run the `pnpm install` from no cache

PR #45224, opened 2026-08-11
feat(manager/mix): support private hex repos via registryAliases

### LINKED ISSUE #29622: Support hex v2 registry

indrekj:
## Changes

Adds support for private/third-party hex registries in the `mix` manager.

Dependencies declared with a `repo:` option (`{:my_dep, "~> 1.0", repo: "acme"}`) come from a third-party hex registry rather than hex.pm. The registry URL is not discoverable from `mix.exs`, so the repo name is mapped to its URL through `registryAliases`, and authenticated with `hostRules`:

```json
{
  "registryAliases": {
    "acme": "https://hex.acme.com"
  },
  "hostRules": [
    {
      "matchHost": "hex.acme.com",
      "token": "...",
      "authType": "Token-Only"
    }
  ]
}
```

Details:

- **Extract** sets `registryUrls` for deps with a `repo:` option. The hex datasource already speaks the V2 registry protocol for any non-default registry, so no datasource change is needed. `packageName` is deliberately left as the bare app name, so the existing `app:organization` format used by the datasource is untouched.
- **Artifacts** runs `mix hex.repo add` for each aliased repo before updating the lockfile. `mix` refuses to fetch from a repo that has no stored registry public key — without one, `:mix_hex_registry.key(nil)` raises and the fetch hangs until the registry timeout. `registryAliases` is a name-to-URL map with nowhere to put a key or its fingerprint, so the key is fetched from the registry and passed to `mix hex.repo add` on disk. This is the same trust-on-first-use model the hex datasource already applies when verifying V2 payloads. A registry that does not serve `/public_key` is skipped rather than added in a state that would crash `mix`.

Based on the approach in #29775 by @bryannaegele, which stalled on a `packageName` format change that is no longer required.

Verified end to end against a self-signed registry built with `mix hex.registry build`, served both locally and from S3 behind a CloudFront function enforcing the hex `authorization` header. In both cases the dep resolves to its private `registryUrl`, the update is found, and the generated `mix hex.repo add` command authenticates and updates `mix.lock` inside a container with the cache dir mounted. The datasource returns both published versions with a valid token and fails closed with a wrong or absent one, and an origin that rejects `Bearer`-prefixed tokens confirms `authType: 'Token-Only'` sends the bare token hex expects.

## Context

- [x] This closes an existing Issue, Closes: #29622

## AI assistance disclosure

Claude Code (Claude Opus 5) was used for non-trivial portions of the code, tests, and documentation, as well as for the end-to-end verification setup described above.

Who answers review comments:

- [x] @indrekj will read and reply directly.

`docs/usage/configuration-options.md` (registryAliases manager list) and `lib/modules/manager/mix/readme.md` (private registry setup).

The public repository: not public — verified against a local repository plus a self-hosted hex registry (local and S3/CloudFront), since a private registry cannot be exercised from a public repo.

### DIFF
--- a/lib/modules/manager/mix/artifacts.ts
+++ b/lib/modules/manager/mix/artifacts.ts
@@ -1,5 +1,6 @@
 import { isEmptyArray, isString } from '@sindresorhus/is';
 import { quote } from 'shlex';
+import upath from 'upath';
 import { GlobalConfig } from '../../../config/global.ts';
 import { TEMPORARY_ERROR } from '../../../constants/error-messages.ts';
 import { logger } from '../../../logger/index.ts';
@@ -8,21 +9,32 @@ import type { ExecOptions } from '../../../util/exec/types.ts';
 import {
   deleteLocalFile,
   ensureCacheDir,
+  ensureDir,
   findLocalSiblingOrParent,
   getSiblingFileName,
   localPathExists,
+  outputCacheFile,
+  privateCacheDir,
   readLocalFile,
   writeLocalFile,
 } from '../../../util/fs/index.ts';
 import * as hostRules from '../../../util/host-rules.ts';
+import { memCacheProvider } from '../../../util/http/cache/memory-http-cache-provider.ts';
+import { Http } from '../../../util/http/index.ts';
+import { coerceObject } from '../../../util/object.ts';
 import { regEx } from '../../../util/regex.ts';
+import { joinUrlParts } from '../../../util/url.ts';
+import { HexDatasource } from '../../datasource/hex/index.ts';
 
 import type { UpdateArtifact, UpdateArtifactsResult } from '../types.ts';
+
+const http = new Http(HexDatasource.id);
 
 const hexRepoUrl = 'https://hex.pm/';
 const hexRepoOrgUrlRegex = regEx(
   `^https://hex\\.pm/api/repos/(?<organization>[a-z0-9_]+)/$`,
 );
+const repoOptionRegex = regEx(/repo:\s*"(?<name>[^"]+)"/g);
 
 export async function updateArtifacts({
   packageFileName,
@@ -145,6 +157,13 @@ export async function updateArtifacts({
     return acc;
   }, [] as string[]);
 
+  preCommands.push(
+    ...(await getRepoAddCommands(
+      newPackageFileContent,
+      config.registryAliases,
+    )),
+  );
+
   // renovate: will update this
   const erlangVersion = '26';
 
@@ -153,6 +172,11 @@ export async function updateArtifacts({
       // https://hexdocs.pm/mix/1.15.0/Mix.Tasks.Archive.html
       // TODO: should include a version constraint
       MIX_ARCHIVES: await ensureCacheDir('mix_archives'),
+      // `mix hex.organization auth` and `mix hex.repo add` persist registry
+      // URLs and their credentials in `$HEX_HOME/hex.config`. Keep that under
+      // privateCacheDir, which is wiped between repositories, so credentials
+      // cannot leak into the next repository of the same run.
+      HEX_HOME: await hexHomeDir(),
     },
     cwdFile: packageFileName,
     docker: {},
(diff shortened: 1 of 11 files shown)


lib/modules/manager/mix/artifacts.ts
```diff
@@ -224,6 +247,97 @@ export async function updateArtifacts({
...
+ * stored for it — with none, `:mix_hex_registry.key(nil)` raises and the fetch
+ * hangs until the registry timeout. `registryAliases` is a name-to-URL map with
+ * nowhere to put a key or its fingerprint, so we fetch the key from the registry
+ * ourselves. This is the same trust-on-first-use model the hex datasource
+ * already applies when it verifies V2 payloads.
+ *
+ * `registryAliases` is inherited from top-level config, so it may hold entries
+ * belonging to other managers. Only the repos `mix.exs` declares are added.
+ */
+async function getRepoAddCommands(
```
RahulGautamSingh, line 270: this fn scans the raw `newPackageFileContent`, but `extract.ts` strips `#` comments before matching repo:. 

A commented-out repo: line will be treated as declared here and trigger a stray public key fetch and `hex.repo` add. 

Strip comments the same way extract does, or reuse a shared helper, so both paths agree on what is declared.

lib/modules/manager/mix/artifacts.ts
```diff
@@ -8,21 +9,31 @@ import type { ExecOptions } from '../../../util/exec/types.ts';
...
 
 import type { UpdateArtifact, UpdateArtifactsResult } from '../types.ts';
 
+const http = new Http(HexDatasource.id);
+
 const hexRepoUrl = 'https://hex.pm/';
 const hexRepoOrgUrlRegex = regEx(
   `^https://hex\\.pm/api/repos/(?<organization>[a-z0-9_]+)/$`,
 );
+const repoOptionRegex = regEx(/repo:\s*"(?<name>[^"]+)"/g);
```
RahulGautamSingh: This is the same pattern as `repoRegexp` in `extract.ts`, differing only in the group name and the global flag. Share one regex so the two do not drift

RahulGautamSingh (changes requested):
Thanks for the PR!

Nice work. The `HEX_HOME` isolation to `privateCacheDir` is a good improvement, and the edge-case test coverage is good as well.

lib/modules/manager/mix/artifacts.ts
indrekj: 7884ea88215b82550035c555b65ca50532f7f97c

lib/modules/manager/mix/artifacts.ts
indrekj: 7884ea88215b82550035c555b65ca50532f7f97c

indrekj:
I think the build failure is not related to my PR

RahulGautamSingh:
> I think the build failure is not related to my PR

Can you add the below, to the tets file? I would like to see if it works, before creating a PR with these changes. (suggestion from github co-pilot session attached to the build test CI workflow)

```ts
// Around line 51-57, enhance the function:
async function disableGitAutoMaintenance(
  repo: ReturnType<typeof simpleGit>,
): Promise<void> {
  await repo.addConfig('gc.auto', '0');
  await repo.addConfig('maintenance.auto', 'false');
  await repo.addConfig('receive.autogc', 'false');
  // Add these additional configurations to prevent background processes
  await repo.addConfig('gc.autodetach', 'false');
  await repo.addConfig('pack.writeReverseIndex', 'false');
}

// In afterEach (around line 204), add a delay before cleanup:
afterEach(async () => {
  // Give Git operations time to fully complete and release file handles
  await new Promise(resolve => setTimeout(resolve, 100));
  await tmpDir?.cleanup();
  await origin?.cleanup();
  vi.restoreAllMocks();
});
```

lib/modules/manager/mix/artifacts.ts
```diff
@@ -224,6 +247,92 @@ export async function updateArtifacts({
   ];
 }
 
+async function hexHomeDir(): Promise<string> {
```
RahulGautamSingh, line 253: `hexHomeDir` runs `ensureDir` on every call, once for `HEX_HOME` and once per repo. 

It is idempotent so this is harmless, but the dir only needs creating once. Compute the path once and `ensureDir` it a single time.

RahulGautamSingh (changes requested):

lib/modules/manager/mix/artifacts.ts
indrekj: f99c7f347fa58c6934337d4d79320f979b8343eb

RahulGautamSingh (approved):
lgtm

indrekj:
There were some merge conflicts, so rebased it to the latest main.

PR #29602, opened 2024-06-11
feat(manager/gradle): Use dependencies task when generating verification metadata

raphiz:
## Changes

As [discussed here](https://github.com/renovatebot/renovate/discussions/29259#discussioncomment-9735907), use `dependencies` task when generating verification metadata.

## Context

Currently, the[ `help` task is called to generate the `verification-metadata.xml`](https://github.com/renovatebot/renovate/blob/e6b04dab1b48097c96f89661ef03a1d1bef3f0b4/lib/modules/manager/gradle/artifacts.ts#L131) (e.g. `./gradlew --write-verification-metadata sha256 help`). Gradle is lazy and [may not evaluate the required dependencies](https://docs.gradle.org/current/userguide/dependency_verification.html#sec:bootstrapping-verification):

> [...] Gradle may discover more dependencies and artifacts depending on the tasks you execute. [...] 

In my tests, using the built-in `dependencies` task instead of `help` works quite well (`./gradlew --write-verification-metadata sha256 dependencies`). The `dependencies` task is also called for lockfile generation.

## Documentation (please check one with an [x])

- [x] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [x] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

PR #45228, opened 2026-08-11
fix(manager/gradle): add --refresh-dependencies when writing verification metadata

derekslenk:
## Changes

Appends `--refresh-dependencies` to the command built by `buildUpdateVerificationMetadataCmd`.

**Why:** with a warm dependency metadata cache, Gradle skips the parent-POM / platform-BOM hierarchy walk and omits their entries from the written `verification-metadata.xml` (gradle/gradle#19228, gradle/gradle#20194 — both open for years). Renovate warms that cache *systematically*: multiple repositories processed through one `GRADLE_USER_HOME` means every repository after the first regenerates against a warm cache. The written file then reliably misses exactly the parent-POM/BOM `.pom`/`.module` entries, and the repository's own CI fails dependency verification on the update PR. The fingerprint is always a recorded `.module` with a missing `.pom`, never jars.

Measured with a controlled comparison against a real-world incomplete file (five known-missing entries; every arm reset to the same committed baseline and asserted to start at 0-of-5; warm arms restored from one snapshotted `GRADLE_USER_HOME`; details in [this gradle/gradle#19228 comment](https://github.com/gradle/gradle/issues/19228#issuecomment-5256639663)):

| Arm | Cache | Added (of 5 missing) |
|---|---|---|
| current command | warm | **0** |
| current command + `--refresh-dependencies` | warm | **5** |
| current command | cold | 5 |
| `build` instead of `dependencies` | warm | **0** |

The last arm matters for this repo's history: the task token was previously changed to `dependencies` (#29602, from discussion #29259) — but the writer resolves all resolvable configurations regardless of the named task, so task choice cannot fix or cause this class. Cache state is the whole variable, and only `--refresh-dependencies` addresses it.

**Cost:** module descriptors are re-fetched on each regeneration rather than served from cache. In our production use (15 repositories, 4 runs/day, several weeks of PR churn including a 13-branch rebase sweep) the write step runs in 11–22 seconds and we observed no registry rate limiting; the correctness failure it prevents has repeatedly cost human debugging sessions, because an incomplete-but-plausible file fails CI in a way that looks like a repo problem rather than a generation problem.

## Context

## AI assistance disclosure

Code change, test updates, and this PR text were drafted with Claude (Fable 5) under my direction; I reviewed the change and we validated it in production across a 15-repository fleet before proposing it here.

Who answers review comments:

- [x] An agent will draft replies and @derekslenk will read them before they are posted. **Name the account.**

The public repository: the validating fleet is private (15 Gradle repositories, self-hosted Renovate 43.288.0); the measurement methodology and results are public in the gradle/gradle#19228 comment linked above. The equivalent preset-level workaround (`postUpgradeTasks` re-running the same command with the flag) has been in production there since 2026-08-11 with zero recurrences.

### DIFF
--- a/lib/modules/manager/gradle/artifacts.ts
+++ b/lib/modules/manager/gradle/artifacts.ts
@@ -154,7 +154,11 @@ async function buildUpdateVerificationMetadataCmd(
     return null;
   }
 
-  return `${baseCmd} --write-verification-metadata ${hashTypes.join(',')} dependencies`;
+  // --refresh-dependencies: with a warm dependency metadata cache, Gradle
+  // skips the parent POM / platform BOM walk and omits their entries from the
+  // written file (https://github.com/gradle/gradle/issues/19228), breaking
+  // consumers that verify metadata. The flag forces the full walk.
+  return `${baseCmd} --write-verification-metadata ${hashTypes.join(',')} dependencies --refresh-dependencies`;
 }
 
 export async function updateArtifacts({
--- a/lib/modules/manager/gradle/artifacts.spec.ts
+++ b/lib/modules/manager/gradle/artifacts.spec.ts
@@ -722,7 +722,7 @@ describe('modules/manager/gradle/artifacts', () => {
       ]);
       expect(execSnapshots).toMatchObject([
         {
-          cmd: './gradlew -Dorg.gradle.jvmargs="-Xms512m -Xmx512m" --console=plain --dependency-verification lenient -q --write-verification-metadata sha256 dependencies',
+          cmd: './gradlew -Dorg.gradle.jvmargs="-Xms512m -Xmx512m" --console=plain --dependency-verification lenient -q --write-verification-metadata sha256 dependencies --refresh-dependencies',
           options: {
             cwd: '/tmp/github/some/repo',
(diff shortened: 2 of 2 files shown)


RahulGautamSingh (approved):
LGTM

lib/modules/manager/gradle/artifacts.ts
```diff
@@ -154,7 +154,11 @@ async function buildUpdateVerificationMetadataCmd(
     return null;
   }
 
-  return `${baseCmd} --write-verification-metadata ${hashTypes.join(',')} dependencies`;
+  // --refresh-dependencies: with a warm dependency metadata cache, Gradle
+  // skips the parent POM / platform BOM walk and omits their entries from the
+  // written file (https://github.com/gradle/gradle/issues/19228), breaking
+  // consumers that verify metadata. The flag forces the full walk.
+  return `${baseCmd} --write-verification-metadata ${hashTypes.join(',')} dependencies --refresh-dependencies`;
```
viceice, line 161: will this cause a lot new requests? maven central is already known to be overloaded by renovate

viceice (commented):
otherwise LGTM

PR #45235, opened 2026-08-11
test(bun): mock util/fs instead of fs-extra in artifacts spec

RahulGautamSingh:
## Changes

Refactor the bun manager artifacts test to mock the `util/fs` wrapper (via the shared `fs` mock from `~test/util.ts`) instead of mocking `fs-extra` directly.

The bun spec was the only manager artifacts spec that mocked `fs-extra`. Every other manager artifacts spec, including the sibling `deno` manager, mocks `util/fs`. This brings bun in line with that convention and removes most of the `as never` casts on the mock return values. Buffer reads keep `as never`, matching the `deno` spec, because the `readLocalFile` overloads resolve the mock return type to `string | null`.

There are no production code changes and the runtime behavior is unchanged.

## Context

## AI assistance disclosure

The test refactor was produced with Claude Code (Claude Opus 4.8), directed and reviewed by @RahulGautamSingh.

Who answers review comments:

- [x] @RahulGautamSingh will read and reply directly. **Name the account.**

The public repository: N/A (test-only refactor, verified via `pnpm vitest run lib/modules/manager/bun/artifacts.spec.ts`, `pnpm check --all`, and `pnpm tsc --noEmit`)

### DIFF
--- a/lib/modules/manager/bun/artifacts.spec.ts
+++ b/lib/modules/manager/bun/artifacts.spec.ts
@@ -1,4 +1,4 @@
-import _fs from 'fs-extra';
+import { fs } from '~test/util.ts';
 import { GlobalConfig } from '../../../config/global.ts';
 import type {
   InternalGlobalConfigOptions,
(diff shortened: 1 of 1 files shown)


viceice (approved):
i still don't like the `as never`, we should try to find an alternate 🤔

PR #45241, opened 2026-08-12
fix(http/github): support cursor pagination

goda6565:
## Changes

GitHub removed offset-based pagination (`page`, `first`, and `last`) from the Dependabot alerts API. Dependabot alert endpoints now support cursor-based pagination using `before`, `after`, and `per_page`.

Sources:
- [GitHub REST API: List Dependabot alerts for a repository](https://docs.github.com/en/rest/dependabot/alerts#list-dependabot-alerts-for-a-repository)

GitHub HTTP pagination previously constructed subsequent URLs using `page=N`, so Renovate could not retrieve Dependabot alerts beyond the first page.

This change:

- follows cursor-based `rel="next"` links sequentially
- preserves existing parallel pagination for page-number-based endpoints
- respects `pageLimit` and `paginate: "all"`
- retains cross-origin pagination protection
- updates the Dependabot alerts integration test to use an `after` cursor
- adds unit tests for cursor traversal, page limits, full pagination, and cross-origin links

Validated with `pnpm check`.

## Context

## AI assistance disclosure

OpenAI Codex (gpt-5.6-sol) was used to inspect the existing pagination implementation, implement cursor pagination, add tests.

Who answers review comments:

- [x] @goda6565 will read and reply directly. **Name the account.**

The public repository: N/A

### DIFF
--- a/lib/modules/platform/github/index.spec.ts
+++ b/lib/modules/platform/github/index.spec.ts
@@ -5392,11 +5392,11 @@ describe('modules/platform/github/index', () => {
             },
           ],
           {
-            link: `<${githubApiHost}/repos/some/repo/dependabot/alerts?state=open&direction=asc&per_page=100&page=2>; rel="next", <${githubApiHost}/repos/some/repo/dependabot/alerts?state=open&direction=asc&per_page=100&page=2>; rel="last"`,
+            link: `<${githubApiHost}/repos/some/repo/dependabot/alerts?state=open&direction=asc&per_page=100&after=cursor-1>; rel="next"`,
           },
         )
         .get(
-          '/repos/some/repo/dependabot/alerts?state=open&direction=asc&per_page=100&page=2',
+          '/repos/some/repo/dependabot/alerts?state=open&direction=asc&per_page=100&after=cursor-1',
         )
         .reply(200, [
           {
--- a/lib/modules/platform/github/index.spec.ts
+++ b/lib/modules/platform/github/index.spec.ts
@@ -5392,6 +5392,103 @@ describe('modules/platform/github/index', () => {
             },
           ],
           {
+            link: `<${githubApiHost}/repos/some/repo/dependabot/alerts?state=open&direction=asc&per_page=100&page=2>; rel="next", <${githubApiHost}/repos/some/repo/dependabot/alerts?state=open&direction=asc&per_page=100&page=2>; rel="last"`,
+          },
+        )
+        .get(
+          '/repos/some/repo/dependabot/alerts?state=open&direction=asc&per_page=100&page=2',
+        )
+        .reply(200, [
+          {
+            security_advisory: {
+              ghsa_id: 'GHSA-1234-5678-9012',
+              summary: 'summary',
+              description: 'description',
+              identifiers: [{ type: 'type', value: 'value' }],
+              references: [],
+              severity: 'low',
+            },
+            security_vulnerability: {
+              package: {
+                ecosystem: 'npm',
+                name: 'center-pad',
+              },
+              severity: 'low',
+              vulnerable_version_range: '0.0.3',
+              first_patched_version: { identifier: '0.0.4' },
+            },
+            dependency: {
+              manifest_path: 'bar/foo',
+            },
+          },
+        ]);
+
+      await github.initRepo({ repository: 'some/repo' });
+      const res = await github.getVulnerabilityAlerts();
+
+      expect(res).toHaveLength(3);
+      expect(res[0].security_vulnerability!.package.name).toBe('left-pad');
+      expect(res[1].security_vulnerability!.package.name).toBe('right-pad');
+      expect(res[2].security_vulnerability!.package.name).toBe('center-pad');
+    });
+
+    it('handles cursor pagination correctly', async () => {
+      const scope = httpMock.scope(githubApiHost);
+      initRepoMock(scope, 'some/repo');
+
+      scope
+        .get(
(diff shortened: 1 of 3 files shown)


jamietanna:
Do you have a changelog link to when they removed it? We'll still need to support it on older GitHub Enterprise Server (GHES) instances if they have the old API paths

goda6565:
Thanks! The GitHub.com changelog is here:
https://github.blog/changelog/2025-10-14-dependabot-alerts-api-pagination-parameters-deprecated/

GHES compatibility is preserved: page-based Link headers still use the existing offset-pagination path.

lib/util/http/github.ts
```diff
@@ -423,36 +423,73 @@ export class GithubHttp extends HttpBase<GithubHttpOptions> {
...
+            );
+            pages = await p.all(queue);
+          } else {
+            pages = [];
+            const paginateAll =
+              !!env.RENOVATE_PAGINATE_ALL || httpOptions.paginate === 'all';
+            let nextUrl: URL | null = firstPageUrl;
+            for (
+              let pageNumber = 2;
+              nextUrl && (paginateAll || pageNumber <= pageLimit);
```
Churro: When `paginateAll` is true this loop has no upper bound. It ends only when the server stops sending `rel="next"`. The offset branch can't do that, since `lastPage` comes from the header and is always finite. A self-hosted or proxied instance that emits a self-referential next, or keeps emitting next on an empty final page, hangs the run. `RENOVATE_PAGINATE_ALL` is a documented self-hosted flag and `getRepos` already hardcodes `paginate: 'all'`, so this could actually be an issue in practice. 

Could we keep a hard safety cap even under paginateAll?

lib/util/http/github.ts
```diff
@@ -423,36 +423,73 @@ export class GithubHttp extends HttpBase<GithubHttpOptions> {
...
+              }
+              const nextPage: HttpResponse<T> =
+                await super.requestJsonUnsafe<T>(method, {
+                  ...opts,
+                  url: nextUrl,
+                });
+              pages.push(nextPage);
+              const nextLink = parseLinkHeader(nextPage.headers.link)?.next;
+              nextUrl = nextLink?.url
+                ? resolvePaginationUrl(nextLink.url)
```
Churro: You could `break` when the page body is empty, which would also help against the unbounded-loop case above.

lib/util/http/github.ts
```diff
@@ -423,36 +423,73 @@ export class GithubHttp extends HttpBase<GithubHttpOptions> {
...
-        const parsedUrl = new URL(next.url, baseUrl);
-        const rebasePagination =
-          !!baseUrl &&
-          !!env.RENOVATE_X_REBASE_PAGINATION_LINKS &&
-          // Preserve github.com URLs for use cases like release notes
-          parsedUrl.origin !== 'https://api.github.com';
-        const firstPageUrl = rebasePagination
-          ? replaceUrlBase(parsedUrl, baseUrl)
-          : parsedUrl;
+        function resolvePaginationUrl(url: string): URL {
```
Churro: oxlint passes but IMHO block-scoped function declarations inside an `if` are ugly to maintain. Surrounding code uses arrow consts, and this is re-created per request, so perhaps use `const resolvePaginationUrl = (url: string): URL => …` instead?

lib/modules/platform/github/index.spec.ts
```diff
@@ -5392,11 +5392,11 @@ describe('modules/platform/github/index', () => {
             },
           ],
           {
-            link: `<${githubApiHost}/repos/some/repo/dependabot/alerts?state=open&direction=asc&per_page=100&page=2>; rel="next", <${githubApiHost}/repos/some/repo/dependabot/alerts?state=open&direction=asc&per_page=100&page=2>; rel="last"`,
+            link: `<${githubApiHost}/repos/some/repo/dependabot/alerts?state=open&direction=asc&per_page=100&after=cursor-1>; rel="next"`,
```
Churro, line 5492: Since your change adds support for `cursor` but keeps the old approach, it would make sense to add a new test for cursor but keep the old one. Otherwise, the deprecated but still supported way won't have coverage.

Churro (commented):
Generally LGTM. Can't test it with GHES but since the old code is still included verbatim, backward compatibility should be warranted. Or in other words:

Before:
```ts
if (next?.url && linkHeader?.last?.page) { ...offset... }
```

After:
```ts
if (next?.url) {
     …resolvePaginationUrl / firstPageUrl / origin check...
     if (linkHeader?.last?.page) { ...offset... }   ← old code
     else                        { ...cursor... }   ← new
}
```

lib/util/http/github.ts
goda6565: Thanks! I tried the arrow function, but it violates `func-style: ["error", "declaration"]` oxlint rule. 
I moved the helper to a top-level function declaration instead, which avoids the block-scoped declaration and per-request recreation while keeping oxlint green.

goda6565:
@Churro 
Thanks for the review.
I've addressed all the comments.

Churro (approved):

jamietanna (dismissed):

jamietanna:
Thanks - re-reviewing then I'll merge

jamietanna (commented):
Would we be able to add a DEBUG log to indicate whether we're trying `next` or `cursor` based pagination? Doesn't need too much information adding, but an indication of which one we're using would be useful!

goda6565:
@jamietanna 
I've added DEBUG logs indicating whether offset-based or cursor-based pagination is being used.

jamietanna (approved):

PR #42024, opened 2026-03-20
fix(manager/npm): pass correct arguments to pnpm dedupe

mestriga:
Run pnpm dedupe with the same options that pnpm install, to:
- suppress unecessary dependency download;
- respect Renovate allowScripts and ignoreScripts options.

## Changes

This PR changes the command line options passed to pnpm-dedupe to mirror those passed to pnpm-install.

Before this PR, pnpm-dedupe was always passed only the --ignore-scripts option, Thsi caused  it to needlessly download all dependencies, always run any .pnpmfile.cjs and ignore scripts even when running them is allowed by the  allowScripts and ignoreScripts Renovate options.

After this PR, pnpm-dedupe always receives the --lockfile-only option and the --ignore-scripts and --ignore-pnpmfile options are set according to the allowScripts and ignoreScripts Renovate options, just as is done for pnpm-install.

The problem that this PR solves was reported in https://github.com/renovatebot/renovate/discussions/41965

## Context

Please select one of the following:

- [ ] This closes an existing Issue, Closes: # 
- [X] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [X] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes — substantive assistance (AI-generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [X] I have updated the documentation, or
- [ ] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [X] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests, but ran on a real repository, or
- [ ] Both unit tests + ran on a real repository

The public repository: <URL>

PR #45243, opened 2026-08-12
fix: Revert "fix(manager/npm): pass correct arguments to pnpm dedupe"

viceice:
Reverts renovatebot/renovate#42024

it breaks on monorepos because of `--recursive` flag

```
[ERROR] Unknown option: 'recursive'
Did you mean 'runtime'? Use "--config.unknown=value" to force an unknown option.
For help, run: pnpm help dedupe
```

### DIFF
--- a/docs/usage/configuration-options.md
+++ b/docs/usage/configuration-options.md
@@ -4158,7 +4158,7 @@ Run `npm install` commands _twice_ to work around bugs where `npm` generates inv
 
 ### `pnpmDedupe`
 
-Run `pnpm dedupe` after `pnpm-lock.yaml` updates.
+Run `pnpm dedupe --ignore-scripts` after `pnpm-lock.yaml` updates.
 
 ### `yarnDedupeFewer`
 
--- a/lib/modules/manager/npm/post-update/pnpm.spec.ts
+++ b/lib/modules/manager/npm/post-update/pnpm.spec.ts
@@ -316,7 +316,7 @@ describe('modules/manager/npm/post-update/pnpm', () => {
         cmd: 'pnpm install --lockfile-only --ignore-scripts --ignore-pnpmfile',
       },
       {
-        cmd: 'pnpm dedupe --lockfile-only --ignore-scripts --ignore-pnpmfile',
+        cmd: 'pnpm dedupe --ignore-scripts',
       },
     ]);
   });
--- a/lib/modules/manager/npm/post-update/pnpm.ts
+++ b/lib/modules/manager/npm/post-update/pnpm.ts
@@ -149,7 +149,7 @@ export async function generateLockFile(
 
     // postUpdateOptions
     if (config.postUpdateOptions?.includes('pnpmDedupe')) {
-      commands.push(`pnpm dedupe ${args}`);
+      commands.push('pnpm dedupe --ignore-scripts');
     }
 
     if (upgrades.find((upgrade) => upgrade.isLockFileMaintenance)) {


viceice:
@mestriga will try to fix it before merging this.

viceice:
```
Version 10.33.2 (compiled to binary; bundled Node.js v20.11.1)
Usage: pnpm dedupe

Perform an install removing older dependencies in the lockfile if a newer version can be used.

Options:
      --[no-]color               Controls colors in the output. By default, output is always colored when it goes directly to a terminal
      --aggregate-output         Aggregate output from child processes that are run in parallel, and only print output when child process is finished. It makes reading large logs after running `pnpm recursive` with `--parallel` or with `--workspace-concurrency` much
                                 easier (especially on CI). Only `--reporter=append-only` is supported.
      --check                    Check if running dedupe would result in changes without installing packages or editing the lockfile. Exits with a non-zero status code if changes are possible.
  -C, --dir <dir>                Change to directory <dir> (default: C:\Users\kriese\projects\visualon\xfab\sx.worktrees\main)
      --global-dir               Specify a custom directory to store global packages
  -h, --help                     Output usage information
      --ignore-scripts           Don't run lifecycle scripts
      --loglevel <level>         What level of logs to report. Any logs at or higher than the given level will be shown. Levels (lowest to highest): debug, info, warn, error. Or use "--silent" to turn off all logging.
      --offline                  Trigger an error if any required dependencies are not available in local store
      --prefer-offline           Skip staleness checks for cached data, but request missing data from the server
      --store-dir <dir>          The directory in which all the packages are saved on the disk
      --stream                   Stream output from child processes immediately, prefixed with the originating package directory. This allows output from different packages to be interleaved.
      --use-stderr               Divert all output to stderr
      --virtual-store-dir <dir>  The directory with links to the store (default is node_modules/.pnpm). All direct and indirect dependencies of the project are linked into this directory
  -w, --workspace-root           Run the command on the root workspace project

Visit https://pnpm.io/10.x/cli/dedupe for documentation about this command.
```

`--lockfile-only` also doesn't work on pnpm v10 and we need to be compatible

viceice:
```
pnpm dedupe --help                                                                                                                                                                                                                                                                 ─╯
Version 11.20.0 (compiled to binary; bundled Node.js v26.0.0)
Usage: pnpm dedupe

Perform an install removing older dependencies in the lockfile if a newer version can be used.

Options:
      --[no-]color               Controls colors in the output. By default, output is always colored when it goes directly to a terminal
      --aggregate-output         Aggregate output from child processes that are run in parallel, and only print output when child process is finished. It makes reading large logs after running `pnpm recursive` with `--parallel` or with `--workspace-concurrency` much easier
                                 (especially on CI). Only `--reporter=append-only` is supported.
      --check                    Check if running dedupe would result in changes without installing packages or editing the lockfile. Exits with a non-zero status code if changes are possible.
  -C, --dir <dir>                Change to directory <dir> (default: C:\Users\kriese\projects\visualon\xfab\sx)
      --global-dir               Specify a custom directory to store global packages
  -h, --help                     Output usage information
      --ignore-scripts           Don't run lifecycle scripts
      --loglevel <level>         What level of logs to report. Any logs at or higher than the given level will be shown. Levels (lowest to highest): debug, info, warn, error. Or use "--silent" to turn off all logging.
      --offline                  Trigger an error if any required dependencies are not available in local store
      --prefer-offline           Skip staleness checks for cached data, but request missing data from the server
      --store-dir <dir>          The directory in which all packages are saved on disk. Use a shared store only with trusted users and jobs
      --stream                   Stream output from child processes immediately, prefixed with the originating package directory. This allows output from different packages to be interleaved.
      --use-stderr               Divert all output to stderr
      --virtual-store-dir <dir>  The directory with links to the store (default is node_modules/.pnpm). All direct and indirect dependencies of the project are linked into this directory
  -w, --workspace-root           Run the command on the root workspace project
  -y, --yes                      Automatically answer yes to prompts and run non-interactively. Will abort if an undesirable situation occurs and user input is strictly necessary.

Visit https://pnpm.io/11.x/cli/dedupe for documentation about this command.
```

`--lockfile-only` is also not available on pnpm v11

secustor (approved):

viceice:
- #45244

PR #45252, opened 2026-08-12
fix(platform/github): abort branch update when PR is in merge queue

viceice:
## Changes

`GithubScm.commitAndPush()` now checks whether the branch's open PR is currently in a merge queue (via the GraphQL `isInMergeQueue` field), and if so throws a new `PR_ALREADY_IN_MERGE_QUEUE` error instead of pushing. `processBranch()` handles this error and returns gracefully (`result: 'done'`), so Renovate does not update a branch whose PR is queued for merging.

Pushing does not kick the PR out of the merge queue, but the final merge uses the PR title / commit message at merge time - so updating a queued PR could produce a merged commit message that no longer matches the changes that actually got merged. See #41903.

Details:

- New `prIsInMergeQueueQuery` GraphQL query (`graphql.ts`), since merge queue status is not exposed via the REST API; `isPrInMergeQueue()` in `pr.ts` wraps it.
- New `assertPrNotInMergeQueue(branchName, baseBranch)` in `index.ts`, called by `GithubScm.commitAndPush()`. It looks up the branch's open PR from the PR cache and throws `PR_ALREADY_IN_MERGE_QUEUE` if the PR is queued.
  - Merge queue enablement is checked per base branch and cached in `config.mergeQueueEnabled`, following the `getBranchForceRebase` pattern - repos/branches without a merge queue skip the per-PR check entirely.
  - For the default branch, enablement comes for free: `repoInfoQuery` now includes `mergeQueue { id }`, so `initRepo` seeds the cache without any extra API call. Only non-default base branches use the lazy `repoMergeQueueQuery` (`Repository.mergeQueue(branch:)`), once per branch per run.
  - The check is skipped on GHES <3.12.0 (no merge queue support): the `mergeQueue` field is stripped from `repoInfoQuery` like other unsupported fields, and the per-branch check short-circuits, following the existing GHE version guard patterns.
  - It fails open: on GraphQL errors or request failures (in both the enablement and PR status checks) it logs at debug level and continues as if a merge queue may exist, so the safety check is not skipped by mistake.
- New `PR_ALREADY_IN_MERGE_QUEUE` constant in `lib/constants/error-messages.ts`.
- `processBranch()` (`lib/workers/repository/update/branch/index.ts`) catches the error and returns `{ result: 'done' }` instead of treating it as a branch error.

## Context

- [x] This closes an existing Issue, Closes: #41903 

## AI assistance disclosure

Code, tests, and this PR description were written by Claude Code (model: Claude Fable 5) under the direction of @viceice.

Who answers review comments:

- [x] @viceice will read and reply directly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/modules/platform/github/index.ts
+++ b/lib/modules/platform/github/index.ts
@@ -77,15 +77,13 @@ import { smartTruncate } from '../utils/pr-body.ts';
 import { remoteBranchExists } from './branch.ts';
 import { coerceRestPr, githubApi, mapMergeStartegy } from './common.ts';
 import {
-  dequeuePullRequestMutation,
   enableAutoMergeMutation,
   getIssuesQuery,
-  prIsInMergeQueueQuery,
   repoInfoQuery,
 } from './graphql.ts';
 import { GithubIssueCache } from './issue.ts';
 import { massageMarkdownLinks } from './massage-markdown-links.ts';
-import { getPrCache, updatePrCache } from './pr.ts';
+import { dequeuePr, getPrCache, updatePrCache } from './pr.ts';
 import {
   GithubBranchProtection,
   GithubBranchRulesets,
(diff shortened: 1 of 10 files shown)


viceice:
causes one additional request when renovate wan#t to update the PR

lib/modules/platform/github/index.ts
```diff
@@ -1984,6 +2022,10 @@ export async function updatePr({
   targetBranch,
 }: UpdatePrConfig): Promise<void> {
   logger.debug(`updatePr(${prNo}, ${title}, body)`);
+  if (await isPrInMergeQueue(prNo)) {
+    logger.debug(`PR #${prNo} is in the merge queue - skipping update`);
+    return;
+  }
```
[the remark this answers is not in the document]
viceice: 🤔  Ideas?

jamietanna (changes requested):
I'm not sure we want to do this - should we not instead _not_ push or update the PR at all when it's in the Merge Queue?

jamietanna:
Can we add the "these branches require merge queue" to the repo cache, so we know whether we do need to do the request?

viceice:
> I'm not sure we want to do this - should we not instead _not_ push or update the PR at all when it's in the Merge Queue?

if we simply return `null` then `processBranch` didn't know it and still sets new commit statuses and calls `updatePr` so this would make the handling much more complicated.

> Can we add the "these branches require merge queue" to the repo cache, so we know whether we do need to do the request?

should be doable if we've the rights to fetch that infos. some self-hosted users don't have admin access, so can't read branch protections. Should we do the check if we can't determine?

secustor (commented):
I think this would be disruptive for the CI system. 
If the PR is in the middle of the queue we restart all the workflows behind it.

I would rather simply defer updating it to the next run 🤔

lib/modules/platform/github/index.ts
```diff
@@ -1974,6 +1977,83 @@ export async function createPr({
   return result;
 }
 
+async function isMergeQueueEnabled(baseBranch: string): Promise<boolean> {
```
secustor, line 1995: In case of multiple baseBranches we should query this together? We can template the query string 🤔 
Would use save an API call per each additional base branch. 

Though this is maybe over optimization at that point

secustor (commented):
else LGTM

lib/modules/platform/github/index.ts
viceice: that needs bigger refactor as the base branches are not available on initRepo. currently they're never send to platform code, so we don't know them. so a single request per base branch should be acceptable. one improvement could be to fetch the default branch merge queue in `initRepo` call, so most user never see the additional call.

secustor: Let's create an issue for that and add a TODO in the code for reference

viceice: I'll let Claude implement it here after a short lunch break 😅

secustor (approved):
LGTM, no new API requests for nearly every user.

jamietanna (approved):

PR #37899, opened 2025-09-09
fix(nix): Fix Nix support

joshuaspence:
## Changes

Builds on top of https://github.com/renovatebot/renovate/pull/37222 to fix (I think?) Nix support. The current implementation seems to be primarily focussed on updating `nixpkgs` dependencies and, from [my testing](https://github.com/joshuaspence/renovate-nix-test/issues/6), doesn't support the following dependency types:

- `git` dependency with `ref` parameter (see joshuaspence/renovate-nix-test#7)
- `git` dependency with `rev` parameter (see joshuaspence/renovate-nix-test#4)
- `git` dependency with `ref` and `rev` parameters (see joshuaspence/renovate-nix-test#5)
- `github` dependency pointing to a commit hash (see joshuaspence/renovate-nix-test#13)
- `github` dependency with `ref` parameter (see joshuaspence/renovate-nix-test#12)

The main change implemented in order to fix this is the implementation of `updateDependency`. This was necessary because in order to fully support Nix dependencies it is necessary for Renovate to update both the current version (i.e. the `ref` parameter) _and_ the current digest (i.e. the `rev` parameter).

NOTE: As this change builds on top of #37222, it also includes the changes from that PR. My recommendation would be to merge that PR before this one.

## Context

Please select one of the below:

- [ ] This closes an existing Issue: #
- [x] This doesn't close an Issue, but I accept the risk that this PR may be closed if maintainers disagree with its opening or implementation

## AI assistance disclosure

Did you use AI tools to create any part of this pull request?

Please select one option and, if yes, briefly describe how AI was used (e.g., code, tests, docs) and which tool(s) you used.

- [ ] No — I did not use AI for this contribution.
- [ ] Yes — minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [x] Yes — substantive assistance (AI generated non‑trivial portions of code, tests, or documentation).
- [ ] Yes — other (please describe):

## Documentation (please check one with an [x])

- [ ] I have updated the documentation, or
- [x] No documentation update is required

## How I've tested my work (please select one)

I have verified these changes via:

- [ ] Code inspection only, or
- [ ] Newly added/modified unit tests, or
- [ ] No unit tests but ran on a real repository, or
- [x] Both unit tests + ran on a real repository

The public repository: https://github.com/joshuaspence/renovate-nix-test

PR #45258, opened 2026-08-13
fix(nix): support flake input updates

tpansino:
## Changes

Continues and supersedes #37899 by @joshuaspence, which was itself built upon earlier work performed by @SuperSandro2000.

Nix flake inputs may track both a version or ref in `flake.nix` and a digest in `flake.lock`. Renovate previously could not reliably update both, and digest-only updates did not consistently trigger lock-file regeneration.

This PR:

- updates supported Git, GitHub, GitLab, SourceHut, Nix channel, and HTTP tarball inputs
- supports combined version/ref and digest updates
- introduces a manager-neutral `updateDependency` result for artifact-only updates
- regenerates `flake.lock` without committing an unchanged `flake.nix`
- maps lock-file node names back to their user-facing root input names
- limits replacements to the matched input URL and recognized `ref` or `rev` locations
- avoids parsing arbitrary complex Nix attribute sets with regular expressions
- uses the current `nix flake update` command
- documents supported declaration forms and limitations

FlakeHub inputs are detected and skipped instead of being treated as ordinary Git references. Full FlakeHub support requires registry-specific version lookup and is substantial enough for a separate contribution.

The root-input mapping also covers the behavior addressed by #40282 and #40867.

## Context

## AI assistance disclosure

OpenAI Codex (GPT-5) helped review the previous implementation and review feedback, implement fixes, write tests and documentation, and run validation.

Who answers review comments:

- [x] @tpansino will read and reply directly.
- [ ] An agent will draft replies and @tpansino will read them before they are posted.


tpansino:
CC @RahulGautamSingh. I tried to address all of your feedback from the other PR via Codex, let me know if you see anything else you would like corrected. Thanks.

lib/modules/manager/nix/update.ts
```diff
@@ -0,0 +1,183 @@
...
+  logger.trace({ depName, currentValue, newValue }, 'nix.updateDependency()');
+
+  if (!depName) {
+    logger.debug('No depName provided');
+    return null;
+  }
+
+  const inputPrefix = '(?:inputs\\.)?';
+  const directPattern = regEx(
+    `^\\s*${inputPrefix}${RegExp.escape(depName)}\\.url\\s*=\\s*"([^"]+)"`,
```
RahulGautamSingh: this regex also matches an override nested inside another input's attrset. 

ie.

```ts
inputs = {
  foo = {
    url = "github:o/foo";
    inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.05";   # <- this line matches first
  };
  nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.05";            # <- the real root input
};
```

exec returns the nested line. The override gets rewritten, the root input stays stale, and nix flake update nixpkgs then relocks against the old ref. 

Fix: collect all matches and pick the one whose URL actually corresponds to the dep (packageName / currentValue / currentDigest), not the first hit.

lib/modules/manager/nix/update.ts
```diff
@@ -0,0 +1,183 @@
...
+    `^\\s*${inputPrefix}${RegExp.escape(depName)}\\s*=\\s*\\{\\s*url\\s*=\\s*"([^"]+)"`,
+    'gm',
+  );
+  const match =
+    directPattern.exec(fileContent) ?? attrSetPattern.exec(fileContent);
+
+  if (!match) {
+    logger.debug(`Could not find URL for dependency ${depName}`);
+    return null;
+  }
```
RahulGautamSingh: return null lands in `get-updated.ts` at `if (!newContent)`, which throws `WORKER_FILE_UPDATE_FAILED`. 

The patterns are deliberately narrow (readme documents "url must be the first member"), so a flake with `inputs.nixpkgs.follows` listed before url, or a let-bound URL, now fails the whole branch. Before this PR nix had no `updateDependency` and `autoReplace` handled these. There needs to be a non-fatal path for "I recognise the dep but cannot rewrite it".

lib/modules/manager/nix/update.ts
```diff
@@ -0,0 +1,183 @@
...
+  ) {
+    updatedUrl.searchParams.set('rev', newDigest);
+    queryModified = true;
+  }
+
+  if (queryModified) {
+    const query = updatedUrl.search.replace(regEx(/%2F/g), '/');
+    updatedUrl.search = '';
+    newUrl = `${updatedUrl.toString()}${query}`;
+  }
```
RahulGautamSingh: `git+https://example.com/o/r?ref=v1.0&dir=sub#frag` → `git+https://example.com/o/r#frag?ref=v2.0&dir=sub`. 

`updateGithubPath` explicitly captures a `[?#]` suffix, so fragments are expected input here. The manual `%2F` un-escaping is fragile for the same reason; a targeted replace of just the ref=/rev= value inside the original string avoids the whole round-trip.

lib/modules/manager/nix/artifacts.ts
RahulGautamSingh: If every depName is filtered out by isNonEmptyStringAndNotWhitespace, inputs is '' and the command becomes nix flake update, relocking every input. Return null when there are no inputs.

lib/modules/manager/nix/index.ts
```diff
@@ -12,9 +13,5 @@ export const url = 'https://nix.dev';
 
 export const defaultConfig = {
   managerFilePatterns: ['/(^|/)flake\\.nix$/'],
-  commitMessageTopic: 'nix',
-  commitMessageExtra: 'to {{newValue}}',
```
RahulGautamSingh, line 16: dropping commitMessageTopic: 'nix' and commitMessageExtra changes branch names and PR titles for every existing nix user

RahulGautamSingh (changes requested):

lib/modules/manager/types.ts
```diff
@@ -383,7 +391,7 @@ interface ManagerApiBase extends ModuleApi {
 
   updateDependency?(
     updateDependencyConfig: UpdateDependencyConfig,
-  ): MaybePromise<string | null>;
+  ): MaybePromise<string | UpdateDependencyResult | null>;
```
viceice: this should be done in a separate PR. Why not stick with range strategy update-lockfile? then this change is probably not needed at all. like npm / bun / deno is doing

viceice (commented):
looks good so far beside the range / update dep change

lib/modules/manager/types.ts
tpansino: > Why not stick with range strategy update-lockfile? then this change is probably not needed at all.

Great point, thank you. Updated.

tpansino:
Apologies for rebasing unnecessarily, it was a force of habit 🤦

PR #45287, opened 2026-08-14
fix(linters): forbid inert counts in v8 coverage-ignore comments

### LINKED ISSUE #44728: Add linting support which prevents `next N` in v8 code coverage comments
We should adapt an existing rule or add a new one which prevents usage of comments like `next N` as these are not honored by V8. Users should rather make use of 
- `if`
- `else`
- `next`
- `file`

```
/* v8 ignore next 4 -- foo */
```

secustor:
## Changes

Adds a new custom oxlint rule `renovate/v8-ignore-no-count` that forbids the count form of coverage-ignore comments, e.g. `/* v8 ignore next 4 -- foo */`.

The count is not honored by the coverage provider: vitest's `ast-v8-to-istanbul` hint parser captures only the keyword (`if`/`else`/`next`/`file`), so `next` always exempts exactly the next AST node, however many lines it spans. A trailing count therefore suggests a line range that is never applied and misleads readers about what is actually exempted.

Details:

- New rule `tools/lint/rules/v8-ignore-no-count.js`, enabled as `error` in `.oxlintrc.json`. It is autofixable: the fixer drops the count (e.g. `next 4` → `next`), which preserves the actual keyword-only behavior, so coverage results are unchanged.
- The autofix was applied to all 78 existing occurrences across `lib/` and `tools/`.
- Updated the note in `v8-ignore-reason.js` that previously documented counts as "allowed inert noise" to point to the new companion rule.
- Added a bullet to `docs/development/best-practices.md` advising against counts.

## Context

- [x] This closes an existing Issue, Closes: #44728

## AI assistance disclosure

The rule implementation, config wiring, and docs update were written by Claude Code (model: Claude Fable 5), which also ran the autofix over the existing occurrences. Verified via `pnpm oxlint`, `pnpm type-check`, `pnpm biome`, `pnpm prettier`, and targeted vitest runs.

Who answers review comments:

- [x] @secustor will read and reply directly. **Name the account.**


jamietanna (dismissed):
Conflicted otherwise LGTM

viceice (approved):
we should also add a lint rule to capture c8 or istanbul ignore comments and forbid them, auto fix to v8 ignore

secustor:
> we should also add a lint rule to capture c8 or istanbul ignore comments and forbid them, auto fix to v8 ignore

Maybe I don't understand you correctly, though if it would fail on c8 ignore comments too than it wouldn't be an autofix 😕

PR #45298, opened 2026-08-15
fix(github): set x-github-api-version

RonaldPhilipsen:
## Changes

Explicitly sets the X-Github-API-Version header to `22-11-28` and define the constant so its easy to change in the future

The minimum GHES version that uses this API version is GHES3.7. but that has been EOL since `2024-01-04`

## Context

- [x] This closes an existing Issue, Closes: #41883 

## AI assistance disclosure

Who answers review comments:

The public repository: <URL>


viceice (commented):
this can be problematic for older ghes versions. please update PR description with the required ghes version for that API

RonaldPhilipsen:
> this can be problematic for older ghes versions. please update PR description with the required ghes version for that API

GHES3.7 already [mentioned](https://docs.github.com/en/enterprise-server@3.7/rest/using-the-rest-api/getting-started-with-the-rest-api#headers) this api version in its docs, not sure which version is the lowest supported, updated the PR description

lib/util/http/github.spec.ts
```diff
@@ -327,6 +327,20 @@ describe('util/http/github', () => {
       );
     });
 
+    it('sets the x-github-api-version header)', async () => {
```
RahulGautamSingh:
```suggestion
    it('sets the x-github-api-version header', async () => {
```

RahulGautamSingh (changes requested):

jamietanna (dismissed):
I can see the earliest GHES we have handling in the code for is [`<3.3.0`](https://github.com/renovatebot/renovate/blob/da01a2f63d365c983b3e214c31d87675c4f59e44/lib/modules/platform/github/index.ts#L548), which [via](https://github.com/github/docs/blob/a44e4a97b17b47505bb3e5b1946e0e816c4bcd49/src/ghes-releases/lib/enterprise-dates.json#L164-L169) ([via](https://github.com/endoflife-date/endoflife.date/issues/5388))

I think given the GHES docs from 3.7 (GA on `2022-11-08`) explicitly documents this, we should be safe to have it

@viceice we could always scoping this based on GHES version being `>= 3.7`?

PR #45303, opened 2026-08-16
feat(platform/local): support dryRun=full to preview changes

virzak:
## Changes

The `local` platform previously only allowed `dryRun=lookup` or `dryRun=extract`, so users could never see the actual file changes Renovate would make. Attempting `dryRun=full` crashed with "Cannot sync git when platform=local" because `getFile` always synced git, which the local platform never initializes.

This PR:

- reads files straight from the working directory in `getFile` when `platform=local`, instead of syncing git (fixes the crash discussed in #37869)
- allows opting in to `dryRun=full` (default stays `lookup`), which applies the computed changes to the working directory so they can be inspected with `git diff`, without committing, pushing, or opening a PR
- guards `dryRun=full` against clobbering work: it refuses to run when the git work tree is dirty (so changes are revertable with `git checkout .`) and warns when run outside a git repository

This implements the `dryRun=full` roadmap item from discussion #40970, where the approach was described and has since drawn interest from other users wanting a fast local config-iteration loop: put files + config in a clean checkout, run `renovate --platform=local --dry-run=full`, inspect with `git diff`, revert, tweak, repeat.

Happy to rework this toward the `local` / `local-git` platform split suggested by @secustor in that discussion if that is the preferred direction; the dirty-work-tree guard slots into whichever platform ends up owning "writes to the local directory".

## Context

## AI assistance disclosure

Code, tests, and docs were developed with Claude Code (Claude Opus 4.8 during initial development; Claude Fable 5 for the rebase and PR preparation), directed and reviewed by @virzak, and verified end to end locally and against real repositories (see below).

Who answers review comments:

- [x] An agent will draft replies and @virzak will read them before they are posted.

The public repository: https://github.com/lareeth/choco-podman-cli

This branch's `--platform=local --dry-run=full` loop was used to develop and validate the Renovate config for that repo (regex manager + `github-release-attachments` checksums); the local `git diff` output matched what the hosted bot later produced on its first real update PRs. It was also used the same way to validate https://github.com/zompinc/lattice-ecp5-toolchain's config (git-refs digest tracking) before onboarding.

### DIFF
--- a/lib/modules/platform/local/index.spec.ts
+++ b/lib/modules/platform/local/index.spec.ts
@@ -1,6 +1,17 @@
+import { partial } from '~test/util.ts';
+import { GlobalConfig } from '../../../config/global.ts';
+import { rawExec as _rawExec } from '../../../util/exec/common.ts';
+import type { ExecResult } from '../../../util/exec/types.ts';
 import * as platform from './index.ts';
 
+vi.mock('../../../util/exec/common.ts');
+const rawExec = vi.mocked(_rawExec);
+
 describe('modules/platform/local/index', () => {
+  afterEach(() => {
+    GlobalConfig.reset();
+  });
+
   describe('initPlatform', () => {
     it('returns input', async () => {
       expect(await platform.initPlatform({})).toMatchInlineSnapshot(`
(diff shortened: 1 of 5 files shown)


jamietanna:
Thanks for this - would prefer we finalise the thinking on how we want to proceed with this before we go further. We'll discuss in the above Discussion :+1:

PR #45321, opened 2026-08-17
feat(manager): add pants manager

jasonwbarnett:
## Changes

Adds a `pants` manager that extracts Python third-party dependencies from [Pants](https://www.pantsbuild.org) build files.

Pants evaluates build files as Python, so they are parsed with `good-enough-parser`'s Python grammar rather than regular expressions. Only the `name`, `requirements` and `source` fields are read, so values in fields like `module_mapping` or `overrides` cannot be mistaken for requirements. Values that an expression builds or chooses are not reported.

Four targets are supported:

- [`python_requirement`](https://www.pantsbuild.org/stable/reference/targets/python_requirement): `requirements=[...]` entries, updated in place in the build file.
- [`python_requirements`](https://www.pantsbuild.org/stable/reference/targets/python_requirements): the file named by `source`, default `requirements.txt`.
- [`poetry_requirements`](https://www.pantsbuild.org/stable/reference/targets/poetry_requirements): the file named by `source`, default `pyproject.toml`.
- [`uv_requirements`](https://www.pantsbuild.org/stable/reference/targets/uv_requirements): the file named by `source`, default `pyproject.toml`.

Generator sources are resolved relative to the build file, extracted once however many targets name them, and delegated by content rather than by the target that pointed there: a `tool.poetry` table means `poetry`, any other TOML means `pep621`, everything else means `pip_requirements`. `poetry_requirements` is the reason this matters, since a pure Poetry `pyproject.toml` extracts nothing through `pep621`.

`extractPackageFile` is also the auto-replace confirmation, and it has to reach the same answer extraction did or every update fails. `managerFilePatterns` cannot tell it, being a repository-stage option that `filterConfig(update, 'branch')` strips, so extraction records how it read each file in `managerData.pantsReadAs`, which is not a config option and survives. `ExtractConfig` gains `packageFile`, `managerFilePatterns` and `managerData` to describe values the confirmation is already passed.

A source this manager cannot maintain, meaning one with a lock file beside it or a requirements file carrying `--hash=` entries, is not reported at all, so the manager that owns the format keeps it. Where a `source=` gives such a file a name no other manager matches, nothing reports it; that trade is documented in the readme.

A file whose extension is a document format is refused whatever the patterns say. `BUILD.md` is a real name that Pants' own default `build_patterns` match, and reading one meant proposing edits to its fenced examples, including claiming whatever file an example named as a source.

**Known limitation:** this PR does not touch Pants lockfiles, so repositories using `pants generate-lockfiles` get a correct package-file update with a stale lockfile. I would rather land extraction first and follow up with `updateArtifacts`, but happy to fold that in if you would prefer it before a `pants` manager exists.

#45455 adds a `cannotUpdate` field to `PackageFileContent` so that such a source can be reported without taking the file from the manager that owns it. It is stacked on this PR rather than folded into it, since it changes shared `supersedes.ts` behaviour and deserves its own review.

## Context

## AI assistance disclosure

This PR was written primarily by Claude Code (Claude Opus), including the manager code, its unit tests and its readme, under my direction and review.

Who answers review comments:

- [x] @jasonwbarnett will read and reply directly. **Name the account.**

`lib/modules/manager/pants/readme.md` covers the supported targets, source-format routing, the file-pattern default and how to extend it, and the overlap with `pip_requirements` / `pep621` / `poetry`.

81 unit tests, with 100% coverage of the new files on statements, branches, functions and lines.

End-to-end against a public repository with one directory per code path: https://github.com/jasonwbarnett/pants-renovate-e2e

Its workflow runs Renovate from this branch, asserts one expectation per path from the JSON report, and applies an update to every extracted dependency through `doAutoReplace`, so a regression fails a check rather than hiding in a log. The latest run checks 83 expectations against 50 dependencies and applies 37 updates across 25 package files.

Extraction was also run against https://github.com/pantsbuild/example-python, where it matches all four `BUILD` files and extracts the default `requirements.txt` source. With `poetry_requirements` and `uv_requirements` targets added there, extraction returns 4 package files and 10 dependencies; running the same repository with `RENOVATE_ENABLED_MANAGERS=pep621` extracts zero dependencies from the Poetry file, which is the case for supporting `poetry_requirements`.

Finally, over a private Pants monorepo of 2,577 build files: 46 inline requirement strings across 42 `python_requirement` targets, 31 `python_requirements` sources, and 29 `poetry_requirements` sources, all 29 of which are pure Poetry files that `pep621` reads as empty. The counts match those computed independently with Python's `ast` module.


jasonwbarnett:
Hey @viceice 👋

Proof of humanity: I have a pulse, and I once spent 45 minutes debugging a BUILD file before realizing I was in the wrong directory.

Full disclosure: this PR is 100% Opus 5 generated. I did the reviewing, the swearing, and the caring. What I want is Pants support in Renovate, built up iteratively, because today the story is "manually bump your python_requirement target and hope."

My ask: can this merge in a clearly labeled experimental state? Feature flag, docs warning, whatever shape you prefer, so I can dogfood it on real repos and iterate in follow-up PRs. Shipping it behind a big scary sign beats it rotting on a branch while I guess at what real usage looks like.

Tell me what hoops to jump through and I'll jump: tests, docs, splitting this into smaller PRs, any of it.

Context for the unfamiliar: https://www.pantsbuild.org/

🙏

lib/modules/manager/pants/readme.md
RahulGautamSingh, line 1: we prefer to not have em-dashes in user-facing docs.

lib/modules/manager/pants/readme.md
```diff
@@ -0,0 +1,36 @@
...
+- [`python_requirements`](https://www.pantsbuild.org/stable/reference/targets/python_requirements): the file named by its `source` field, default `requirements.txt`.
+- [`poetry_requirements`](https://www.pantsbuild.org/stable/reference/targets/poetry_requirements): the file named by its `source` field, default `pyproject.toml`.
+- [`uv_requirements`](https://www.pantsbuild.org/stable/reference/targets/uv_requirements): the file named by its `source` field, default `pyproject.toml`.
+
+The generator targets are handled the same way: the source is resolved relative to the build file, extracted, and updated in that file, and a source referenced by several targets is extracted once. The source's own format picks the extractor — Poetry for a `pyproject.toml` with a `[tool.poetry...]` table, PEP 621 for any other `pyproject.toml`, and a pip requirements file otherwise — so each dependency keeps the `depType` its format gives it, including Poetry and uv dependency groups.
```
RahulGautamSingh: one sentence per line

lib/modules/manager/pants/parser.ts
```diff
@@ -0,0 +1,116 @@
...
+
+// Only the attributes we consume. Anything else — `module_mapping`,
+// `overrides`, `resolve` — is skipped, so their string values can never be
+// mistaken for requirements.
+const attrNameRegex = regEx(/^(?:name|requirements|source)$/);
+
+function startTarget(ctx: Ctx, name: string): Ctx {
+  return {
+    ...ctx,
+    target: { type: name as PantsTargetType, requirements: [] },
```
RahulGautamSingh: possible to remove the type assertion?

lib/modules/manager/pants/extract.ts
```diff
@@ -0,0 +1,164 @@
...
+ * from ever disagreeing.
+ */
+function extractSourceFile(
+  content: string,
+  packageFile: string,
+): Promise<PackageFileContent | null> | PackageFileContent | null {
+  if (upath.basename(packageFile) !== 'pyproject.toml') {
+    return extractRequirementsFile(content);
+  }
+  return content.includes('[tool.poetry')
```
RahulGautamSingh: includes('[tool.poetry') also matches tables like [tool.poetry-dynamic-versioning]. a PEP 621 file carrying one would route to the poetry extractor and miss its [project] deps. 

match [tool.poetry] and [tool.poetry. instead.

RahulGautamSingh (changes requested):
fly-by review

jasonwbarnett:
@RahulGautamSingh thanks for the fly by 🛩️. Addressed the feedback. Let me know if there is anything else.

lib/modules/manager/pants/extract.ts
```diff
@@ -0,0 +1,198 @@
...
+          { packageFile, source, target: target.type },
+          'pants: generator source not found',
+        );
+        continue;
+      }
+
+      const extracted = await extractSourceFile(sourceContent, source);
+      if (extracted?.deps?.length) {
+        result.push({
+          ...extracted,
```
RahulGautamSingh, line 457: `extractPoetryFile` sets `res.lockFiles = ['poetry.lock']`, and `{ ...extracted }` carries that onto the pants package file. 

The pants manager exports no `updateArtifacts`, so Renovate opens PRs that bump `pyproject.toml` and leave `poetry.lock` stale. Same for `uv.lock` via `pep621`.

lib/modules/manager/pants/index.ts
RahulGautamSingh, line 1: Missing supersedesManagers

The readme tells users to disable `pip_requirements` to stop duplicate PRs. Renovate already solves this: via `export const supersedesManagers = ['pip_requirements', 'pep621', 'poetry']`. 

`processSupersedesManagers` drops the secondary manager's entry for any packageFile pants also produced, and it has a lockFiles branch that rejects the primary instead when the secondary carries a lock file.

lib/modules/manager/pants/extract.ts
```diff
@@ -0,0 +1,198 @@
...
+
+const defaultSources: Record<PantsTargetType, string> = {
+  python_requirement: '',
+  python_requirements: 'requirements.txt',
+  poetry_requirements: 'pyproject.toml',
+  uv_requirements: 'pyproject.toml',
+};
+
+function isBuildFile(packageFile: string): boolean {
+  return upath.basename(packageFile).startsWith('BUILD');
```
RahulGautamSingh: Pants `build_patterns` allows any name, and the readme encourages setting `managerFilePatterns`. 

With `build_patterns = ["pants_targets.py"]`, extraction still works (it goes through `extractAllPackageFiles`), but `autoReplace` re-extraction routes the build file into `extractRequirementsFile` and every inline `python_requirement` update silently fails. 

Parse first and treat the file as a build file when parse(content) yields targets.

lib/modules/manager/pants/extract.ts
```diff
@@ -0,0 +1,198 @@
...
+function toDep(requirement: string): PackageDependency | null {
+  const dep = extractRequirementsFile(requirement)?.deps?.[0];
+  if (!dep?.depName) {
+    return null;
+  }
+  return {
+    ...dep,
+    depType: 'python_requirement',
+    // The bare version range repeats across targets in a big BUILD file; the
+    // whole requirement string is what makes the replacement unambiguous.
```
RahulGautamSingh: The comment claims the whole requirement string makes the replacement unambiguous. `doAutoReplace` uses `existingContent.indexOf(replaceString)`, first match only, no `depIndex` offset. 

Pants repos routinely pin the same requirement in several targets in one BUILD file, one per resolve. The second occurrence never updates, and `confirmIfDepUpdated` then fails on "Value is not updated". Correct the comment.

RahulGautamSingh (changes requested):

lib/modules/manager/pants/readme.md
jamietanna, line 1: This is incredibly verbose - I assume because an LLM wrote it

Please look at condensing this to what's most important and/or use section headers to break things up

jamietanna:
Some commits on this branch have the wrong author

lib/modules/manager/pants/readme.md
jasonwbarnett: yep, it's a bunch of LLM verbose output junk. I'm fixing it up. Thanks for the review. I can only imagine how painful it is having to see this crap day in and day out from random "contributors" on the internet 😅

jamietanna: Yeah I think I'll mark this as draft for now - when it's ready for a re-review open it back up

If you can also fix up the history - I'd prefer not to have a ~50 commit PR getting squash-merged

PR #45334, opened 2026-08-18
fix(vulnerabilities): scope Go stdlib OSV remediation to the `toolchain` directive

jalaziz:
## Changes

When `osvVulnerabilityAlerts` fires for a Go `toolchain` directive, Renovate also bumped the unrelated module `go` directive (`depType: golang`) onto a non-vulnerability branch.

The OSV lookup already skipped the module `go` line (`depType !== 'toolchain'`), but `vulnerabilityToPackageRules()` emitted a remediation rule with only `matchDatasources` and `matchPackageNames`. Both directives share `datasource: golang-version` and `packageName: go`, so the toolchain's `allowedVersions` also applied to the module line.

`matchCurrentVersion` of the toolchain (e.g. `1.26.5`) does not exclude `go 1.26.0` because `go-mod-directive` versioning treats `1.26.0` as `^1.26.0`.

Vulnerability rules are generated per dependency. This PR copies that dep's `depType` onto the rule as `matchDepTypes` when present, so the rule cannot match a different dep that happens to share datasource and package name. The Go toolchain still updates on the vulnerability branch; the module `go` directive is left alone.

### Reproduction

`go.mod` / `go.work`:

```
go 1.26.0
toolchain go1.26.5
```

with an active Go stdlib OSV advisory affecting 1.26.5, fixed in 1.26.6.

**Before:** three updates (`go`, `go`, `go`) — toolchain 1.26.5→1.26.6 on `renovate/…-go-vulnerability`, plus the module `go` directive 1.26.0→1.26.6 on `renovate/go-1.x`.

**After:** only the toolchain update on the vulnerability branch.

Reproduced on Renovate 44.29.5 and 44.32.6 (present on current main).

## Context

## AI assistance disclosure

This PR was written primarily by Cursor Grok 4.6. The bug, root cause, and fix direction were specified by the author; the agent implemented the change, added the unit tests, and drafted this description.

### DIFF
--- a/lib/workers/repository/process/types.ts
+++ b/lib/workers/repository/process/types.ts
@@ -7,6 +7,7 @@ export interface Vulnerability {
   packageFileConfig: RenovateConfig & PackageFile;
   packageName: string;
   osvPackageName: string;
+  depType?: string;
   depVersion: string;
   fixedVersion: string | null;
   datasource: string;
--- a/lib/workers/repository/process/vulnerabilities.spec.ts
+++ b/lib/workers/repository/process/vulnerabilities.spec.ts
@@ -4,7 +4,9 @@ import { mockFn } from 'vitest-mock-extended';
 import type { RenovateConfig } from '~test/util.ts';
 import { logger } from '~test/util.ts';
 import { getConfig } from '../../../config/defaults.ts';
+import type { PackageRuleInputConfig } from '../../../config/types.ts';
 import type { PackageFile } from '../../../modules/manager/types.ts';
+import { applyPackageRules } from '../../../util/package-rules/index.ts';
 import { Vulnerabilities } from './vulnerabilities.ts';
 
 const getVulnerabilitiesMock =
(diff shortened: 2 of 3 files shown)


jamietanna (commented):
This looks reasonable - @renovatebot/maintainers would this potentially break other ecosystems, where we may want to catch on multiple `depTypes` i.e. `devDependencies` + `dependencies` in a given project?

Churro (changes requested):
I'm not concerned about `devDependencies` vs `dependencies` because they each get their own package rule. However, at the moment in npm a rule for `dependencies` implicitly also covers e.g., `overrides`. This would break with the changes proposed in this PR.

Hence, I'd suggest scoping it to Go to avoid side effects for other managers. It would also simplify the change because there is already special handling for `toolchain`:
https://github.com/renovatebot/renovate/blob/98988e1b2a4fc0dc9f165d5709288ad462eb79f2/lib/workers/repository/process/vulnerabilities.ts#L178-L184

Simply set `depType = dep.depType;` after `osvPackageName = 'stdlib';` and then
```ts
vulnerabilities.push({ 
   ...
   depType (instead of dep.depType)
   ... 
})
```

jalaziz (commented):
> Simply set depType = dep.depType; after osvPackageName = 'stdlib'; and then

Done. `depType` should only be set for Go dependencies now.

Churro (approved):

jamietanna (approved):

PR #45335, opened 2026-08-18
fix(exec): supress unhandled promise rehection from execa timeout

Tej128:
## Changes

When a command (like pipenv lock in my case) exceeds the configured executionTimeout, Renovate already handles this correctly by marking the repository as temporary-error and moves on. However, the execa library also throws a separate "Timed out" error internally. This second error is not caught by anyone, so it triggers the global unhandledRejection handler, which
logs it at ERROR level and causes the entire Renovate run to exit with a failure code even though every other repository was
processed successfully.

This fix catches and discards that duplicate error, since the timeout is already handled through the normal code path.

## Context

## AI assistance disclosure

Who answers review comments:

- [x] @username will read and reply directly. **Name the account.**

The public repository: <URL>

### DIFF
--- a/lib/util/exec/common.ts
+++ b/lib/util/exec/common.ts
@@ -136,9 +136,10 @@ export function exec(
     // Suppress execa's internal promise rejection (e.g., from timeout).
     // We handle all exit scenarios via 'exit' and 'error' event listeners below,
     // so the promise rejection would otherwise surface as an unhandledRejection.
-    // v8 ignore next -- execa returns a thenable in production but not in mocked tests
     if (isFunction(cp.catch)) {
-      cp.catch(() => undefined);
+      cp.catch((err) =>
+        logger.warn({ err }, 'execa promise rejection suppressed'),
+      );
     }
 
     // handle streams
--- a/lib/util/exec/common.ts
+++ b/lib/util/exec/common.ts
@@ -1,6 +1,6 @@
 import type { ChildProcess } from 'node:child_process';
 import type { Readable } from 'node:stream';
-import { isNullOrUndefined } from '@sindresorhus/is';
+import { isFunction, isNullOrUndefined } from '@sindresorhus/is';
 import { execa } from 'execa';
 import { join, split } from 'shlex';
 import { instrument } from '../../instrumentation/index.ts';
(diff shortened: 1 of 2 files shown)


Tej128:
@felipecrs

felipecrs:
Not sure about this. Would be nice to have a [minimum reproduction](https://github.com/renovatebot/renovate/blob/da01a2f63d365c983b3e214c31d87675c4f59e44/docs/development/minimal-reproductions.md).

Tej128:
Hello,
Here is a minimum reproduction of the issue: https://github.com/Tej128/renovate-repro-exec-timeout.git

viceice (dismissed):

lib/util/exec/common.ts
```diff
@@ -144,6 +133,14 @@ export function exec(
       extendEnv: false,
     });
 
+    // Suppress execa's internal promise rejection (e.g., from timeout).
+    // We handle all exit scenarios via 'exit' and 'error' event listeners below,
+    // so the promise rejection would otherwise surface as an unhandledRejection.
+    // v8 ignore next -- execa returns a thenable in production but not in mocked tests
+    if (isFunction(cp.catch)) {
+      cp.catch(() => undefined);
```
secustor:
```suggestion
      cp.catch((err) => logger.warn({ err }, 'execa promise rejection suppressed'));
```
We should log this

lib/util/exec/common.ts
```diff
@@ -144,6 +133,14 @@ export function exec(
       extendEnv: false,
     });
 
+    // Suppress execa's internal promise rejection (e.g., from timeout).
+    // We handle all exit scenarios via 'exit' and 'error' event listeners below,
+    // so the promise rejection would otherwise surface as an unhandledRejection.
+    // v8 ignore next -- execa returns a thenable in production but not in mocked tests
```
secustor: I think this would be solved by adding a `catch` function to he mock https://github.com/renovatebot/renovate/blob/8d08b51db0f782c3efad5ea09220759051bdd4a1/lib/util/exec/common.spec.ts#L149

lib/util/exec/common.ts
Tej128: Applied - now logs the suppressed rejection at warn level.

lib/util/exec/common.ts
Tej128: Done - added catch to the mock in common.spec.ts so the code path is covered without v8 ignore.

secustor (approved):

